PR #30, opened 2018-03-22
support spark struct directly to thrift type

yaooqinn:


yaooqinn:
8d05c98 SparkEnv.get called in an non doas wrapping code, which will cause NPE

PR #75, opened 2018-05-24
support incrementally get result from executors to reduce OOM risks

yaooqinn:
As a multi tenant server, kyuubi plays the role of many "spark apps"'s driver/client, it might need more memory than the spark thrift server. I think that `Dataset.toLocalIterator` api can receive partial results from executors incrementally, which helps us to reduce the server OOM risk


yaooqinn:
when i test ffb2ad0  with beeline connected to kyuubi, the query results seem to show before part of the log, while without this patch results always shows after

Issue #124, opened 2018-11-26
[Configuration] A configuration error for expiryInterval


hustfeiwang:
**Describe the bug**
There is an error when set the expiryInterval for KyubbiAppMaster.
The relative code in KyubbiAppMaster.scala is shown below.
```
val expiryInterval = yarnConf.getInt(RM_AM_EXPIRY_INTERVAL_MS, 120000)
interval = math.max(100, math.min(expiryInterval / 2, 3000))
```
Maybe you want to express
```
 interval = math.max(100 * 1000, math.min(expiryInterval / 2, 3000))
```

PR #126, opened 2018-11-27
KYUUBI-124 fix the error when set interval #124

hustfeiwang:


yaooqinn:
as to your change, the interval can not be less than 100 seconds, right?
the original code is just to prevent the value being negative.

Issue #129, opened 2018-12-06
kyuubi throws  java.lang.UnsupportedOperationException when  starting


hustfeiwang:
**Describe the bug**
In KyuubiMain.scala,     val env = Map.empty[String, String].asJava
The type of env is EmptyMap which does not implement the put method of abstractMap.

PR #130, opened 2018-12-06
Kyuubi-129 kyuubi throws java.lang.UnsupportedOperationException when starting #129

hustfeiwang:


yaooqinn:
nice catch, pending travis

PR #146, opened 2019-01-20
修复可能存在并发opensession的时候，realUser 可能会获取另一个session 的user

Neilxzn:
FrontService  getUserName method is not thread safe  because of the variable ` realUser`


Neilxzn:
I didn't change the opensessionhandle file, why it can't pass checks
@yaooqinn

PR #184, opened 2019-05-09
Cast to int wc jar count

gnieto:
While using Kyuubi on my local environment (OSX) I'm having issues with
`wc` returning the result with a prefixed tab character.

Due to this issue, it fails to compare "	1" with "1" and script
is aborted due to duplicated JARs found.

To solve that, I'm casting `wc` result to integer and perform integer
comparisons, which seems to solve the issue.


yaooqinn:
LGTM, thanks. Merging to master. Give Kyuubi a star, maybe. Thanks again.

PR #222, opened 2019-10-24
[KYUUBI-221] Optimise FetchResults speed when maxRows is large

### LINKED ISSUE #221: Optimise FetchResults speed when maxRows is large
**Is your feature request related to a problem? Please describe.**
When maxRows in FetchResults is large, FetchResults would run very slow. 
For example, Tableau implements the thrift client with maxRows = 10000, usually it will cost
few seconds to get the response.
Because ColumnBasedSet.toTRowSet is very slow. The input rows: Seq[Row] is actually something like Stream[Row]. toTRowSet will get each element by its index, this will take 
linear time.

**Describe the solution you'd like**
Use map function instead of index accessing.

qccash:
To Resolve #221


yaooqinn:
LGTM.  merge to master

PR #250, opened 2020-11-16
add SparkSQLEngine idle check

xjzrc:
1.新增SparkSQLEngine空闲定时检测任务 
2.修复SparkSQLEngine关闭session时未移除SparkSQLOperationManager中缓存的sparkSession


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -18,25 +18,47 @@
...
+  override def start(): Unit = {
+    val interval = conf.get(KyuubiConf.ENGINE_CHECK_INTERVAL)
+    val idleTimeout = conf.get(KyuubiConf.ENGINE_IDLE_TIMEOUT)
+    val checkTask = new Runnable {
+      override def run(): Unit = {
+        val current = System.currentTimeMillis
+        val sessionManager = backendService.sessionManager.asInstanceOf[SparkSQLSessionManager]
+        if (sessionManager.getOpenSessionCount <= 0 &&
+          current - sessionManager.getLogoutTime > idleTimeout) {
+          sys.exit(0)
```
yaooqinn, line 55: this seems incorrect. we need to deregister from zk too

yaooqinn: Can we print some log here？

PR #254, opened 2020-11-23
[KYUUBI#253] Support for configuring the scope range of SparkSQLEngine.

xjzrc:
Support for configuring the scope range of SparkSQLEngine.


yaooqinn:
Hi @xjzrc any update on this？

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -378,6 +379,18 @@ object KyuubiConf {
     .timeConf
     .createWithDefault(Duration.ofMinutes(30L).toMillis)
 
+  val ENGINE_SCOPE: ConfigEntry[String] = buildConf("session.engine.scope")
+    .doc("The engine session scope.<ul>" +
+      " <li>S: One engine per kyuubi session in kyuubi cluster.</li>" +
+      " <li>U: One engine per user in kyuubi cluster.</li>" +
+      " <li>G: One engine per group in kyuubi cluster.</li>" +
+      " <li>K: One engine per kyuubi server in kyuubi cluster.</li>")
```
yaooqinn, line 387: Missing </ul>" here

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/EngineAppName.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.kyuubi.engine.EngineScope.{GROUP, SERVER, SESSION, USER}
+
+class EngineAppName(user: String, sessionId: String, conf: KyuubiConf) {
+
+  import EngineAppName._
+
+  private val engineScope = EngineScope.withName(conf.get(ENGINE_SCOPE))
+  private val serverHost = conf.get(FRONTEND_BIND_HOST)
+    .getOrElse(InetAddress.getLocalHost.getHostName)
+  private val serverPort = conf.get(FRONTEND_BIND_PORT)
```
yaooqinn, line 35: How about using the connection URL of the server after the port is actually resoved？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -17,71 +17,69 @@
...
 import org.apache.kyuubi.service.Serverable
-import org.apache.kyuubi.util.{SignalRegister, ThreadUtils}
+import org.apache.kyuubi.util.SignalRegister
 
 private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
   extends Serverable(name) {
 
   def this(spark: SparkSession) = this(classOf[SparkSQLEngine].getSimpleName, spark)
 
-  private val timeoutChecker =
```
yaooqinn, line 39: Can we make a separate PR for this issue?

xjzrc: ok

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
xjzrc: ok, i have been added

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/EngineAppName.scala
xjzrc: this is the kyuubiServer's host and port ,It comes from the kyuubiServer configuration.

yaooqinn: we support 0 as a random port, which need to be resolved to actual port

xjzrc: if the kyuubiServer use a random port, how does the client connect to the service

yaooqinn: say way how the server connect engines

xjzrc: no, It's port number for Kyuubi, not engine. If we parse to the actual port number, then write back to the kyuubiServer 
 configuration.How about this？

yaooqinn: yes this is the port for kyuubi server and it supports 0 for a random pick, we have a HA layer that support dynamic service discovery

yaooqinn: So, I have found another issue for you, the zookeeper directories that you designed so far, may never consider the conflict with the server `znode`

xjzrc: It‘s HA server `znode`?

PR #260, opened 2020-12-10
ZkClient should close when the session is opened.

xjzrc:


kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -84,28 +84,35 @@ class KyuubiSessionImpl(
...
-        val started = System.currentTimeMillis()
-        while (sh.isEmpty) {
-          if (process.waitFor(1, TimeUnit.SECONDS)) {
-            throw builder.getError
-          }
-          if (started + timeout <= System.currentTimeMillis()) {
-            process.destroyForcibly()
-            throw KyuubiSQLException(s"Timed out($timeout ms) to launched Spark with $builder",
-              builder.getError)
+    try {
```
yaooqinn, line 87: Seems that we don't need this try-catch here as all are handled by the session manager.

How about just add this to the end?

```scala
try {
   zkClient.close()
} catch {
   case e: IOException => error("Failed to release the zkClient after session established", e)
}
```

xjzrc: How about all kyuubi's sessions share one ZkClient?

yaooqinn: I don‘t trust zk’s stability very much :)
For registering server or engines, we have a de-register handler there.

yaooqinn: here, I don't think it is a right way to go

xjzrc: i see

yaooqinn (approved):

yaooqinn:
LGTM, pending CI

yaooqinn:
thanks, merged to master

PR #264, opened 2020-12-15
Support Spark YARN cluster deployment mode

xjzrc:
Support Spark YARN cluster deployment mode

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -66,6 +66,13 @@ case class KyuubiConf(loadSysDefault: Boolean = true) extends Logging {
     this
   }
 
+  def setIfMissing(key: String, value: String): KyuubiConf = {
+    require(key != null)
+    require(value != null)
+    settings.putIfAbsent(key, value)
+    this
+  }
+
   def setIfMissing[T](entry: ConfigEntry[T], value: T): KyuubiConf = {
     settings.putIfAbsent(entry.key, entry.strConverter(value))
     this
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -66,13 +66,6 @@ case class KyuubiConf(loadSysDefault: Boolean = true) extends Logging {
     this
   }
 
-  def setIfMissing(key: String, value: String): KyuubiConf = {
-    require(key != null)
-    require(value != null)
-    settings.putIfAbsent(key, value)
-    this
-  }
-
   def setIfMissing[T](entry: ConfigEntry[T], value: T): KyuubiConf = {
     settings.putIfAbsent(entry.key, entry.strConverter(value))
     this
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -19,7 +19,6 @@ package org.apache.kyuubi.session
 
 import java.io.IOException
 import java.nio.charset.StandardCharsets
-import java.util.Locale
 import java.util.concurrent.TimeUnit
 
 import scala.collection.JavaConverters._
@@ -66,14 +65,6 @@ class KyuubiSessionImpl(
   private var client: TCLIService.Client = _
   private var remoteSessionHandle: TSessionHandle = _
 
-  private val isClusterMode = sessionConf.getOption(SPARK_SUBMIT_DEPLOY_MODE)
-    .map(_.toLowerCase(Locale.ROOT)).contains("cluster")
-
-  if (isClusterMode) {
-    sessionConf.set(SPARK_YARN_MAX_APP_ATTEMPTS, "1")
-    sessionConf.set(SPARK_YARN_SUBMIT_WAIT_APP_COMPLETION, "false")
-  }
-
   private def getServerHost: Option[(String, Int)] = {
     try {
       val hosts = zkClient.getChildren.forPath(zkPath)
@@ -103,9 +94,13 @@ class KyuubiSessionImpl(
         info(s"Launching SQL engine: $builder")
         var sh = getServerHost
         val started = System.currentTimeMillis()
+        var exitValue: Option[Int] = None
         while (sh.isEmpty) {
-          if (!isClusterMode && process.waitFor(1, TimeUnit.SECONDS)) {
-            throw builder.getError
+          if (exitValue.isEmpty && process.waitFor(1, TimeUnit.SECONDS)) {
+            exitValue = Some(process.exitValue())
+            if (exitValue.get != 0) {
+              throw builder.getError
+            }
           }
           if (started + timeout <= System.currentTimeMillis()) {
             process.destroyForcibly()
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -29,7 +29,7 @@ import org.apache.thrift.TException
 import org.apache.thrift.protocol.TBinaryProtocol
 import org.apache.thrift.transport.{TSocket, TTransport}
 
-import org.apache.kyuubi.{KyuubiSQLException, ThriftUtils}
+import org.apache.kyuubi._
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf._
 import org.apache.kyuubi.engine.spark.SparkProcessBuilder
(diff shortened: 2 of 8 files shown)


kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -65,6 +66,14 @@ class KyuubiSessionImpl(
   private var client: TCLIService.Client = _
   private var remoteSessionHandle: TSessionHandle = _
 
+  private val isClusterMode = sessionConf.getOption(SPARK_SUBMIT_DEPLOY_MODE)
+    .map(_.toLowerCase(Locale.ROOT)).contains("cluster")
+
+  if (isClusterMode) {
+    sessionConf.set(SPARK_YARN_MAX_APP_ATTEMPTS, "1")
```
yaooqinn, line 73: Why should it be 1?

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -65,6 +66,14 @@ class KyuubiSessionImpl(
...
+    sessionConf.set(SPARK_YARN_SUBMIT_WAIT_APP_COMPLETION, "false")
```
yaooqinn, line 74: Is it necessary?

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -95,7 +104,7 @@ class KyuubiSessionImpl(
         var sh = getServerHost
         val started = System.currentTimeMillis()
         while (sh.isEmpty) {
-          if (process.waitFor(1, TimeUnit.SECONDS)) {
+          if (!isClusterMode && process.waitFor(1, TimeUnit.SECONDS)) {
```
yaooqinn, line 107: does this mean that cluster mode never fails with client exceptions/errors?

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: Prevent AM restarts when the engine reaches its maximum idle time and terminated.

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: In cluster mode, clients should not be resident.

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: yes, the engine will start in AM.

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
yaooqinn: AM will restart when it terminated w/ FINISHED state?

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
yaooqinn: You did not get my point

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
yaooqinn: what is wrong if we keep the client alive?

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: enen

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: No, but it consumes kyuubi service's resources.

xjzrc: so i think set `spark.yarn.submit.waitAppCompletion=false` is necessary.

yaooqinn: Let us keep it configurable in case of any regression happen?

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: In client mode, we can use `process.waitFor` to return true to determine engine startup exceptions. But in cluser mode, if   'spark.yarn.submit.waitAppCompletion=false' is set , `process.waitFor` always return true.

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
yaooqinn: What makes it different from other regular spark applications? I guess they are the same

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
yaooqinn: The client exception is useful. for example, your YARN cluster is crashed, or maybe the server side is configured wrong or the user overwrites some setting, then Kyuubi failed to start the engine proc, we need to capture the err.

yaooqinn:
we need some end to end case to test this feature

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: I made a mistake,  AM didn't w/ FINISHED state when it terminated. I will fix it.

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: ok

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: @yaooqinn I've fixed it.

xjzrc:
> we need some end to end case to test this feature?
I packaged this for testing in our YARN cluster. How do we do unit testing locally.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -66,6 +66,13 @@ case class KyuubiConf(loadSysDefault: Boolean = true) extends Logging {
     this
   }
 
+  def setIfMissing(key: String, value: String): KyuubiConf = {
```
yaooqinn, line 69: unnecessary change？

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -96,9 +96,13 @@ class KyuubiSessionImpl(
         info(s"Launching SQL engine: $builder")
         var sh = getServerHost
         val started = System.currentTimeMillis()
+        var exitValue: Option[Int] = None
```
yaooqinn, line 99: unnecessary definition？

yaooqinn:
> > we need some end to end case to test this feature?
> 
> I packaged this for testing in our YARN cluster. How do we do unit testing locally.

That is OK, but I didn't know that before you told me.

BTW, if you verified locally in your yarn cluster, could you please provide the screenshot or anything else that can help me confirm that.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
xjzrc: sorry, I forgot to remove it.

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
xjzrc: necessary.  there is no need to execute `process.waitFor`, when the process exits normally, eg: cluster mode and 'spark.yarn.submit.waitAppCompletion=false'

yaooqinn: If the process exits already, it won't wait.

xjzrc:
> > > we need some end to end case to test this feature?
> > 
> > 
> > I packaged this for testing in our YARN cluster. How do we do unit testing locally.
> 
> That is OK, but I didn't know that before you told me.
> 
> BTW, if you verified locally in your yarn cluster, could you please provide the screenshot or anything else that can help me confirm that.
kyuubi server log:
image
client log:
image
AM log:
image
image

PR #268, opened 2021-01-08
Improve pom.xml

pan3793:
### What changes in this PR?

- correct `log4j.version` to` slf4j.version`
- extract versions to properties

And, we usually pin dependency version and exclude transitive dependencies in `<dependencyManagement>` block, but define the `scope` in specify `<dependency>` block, do you agree with that? If yes, I'd like to do it in another PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,6 @@
         <maven.version>3.6.3</maven.version>
         <spark.version>3.0.1</spark.version>
         <slf4j.version>1.7.30</slf4j.version>
-        <mockito.verison>1.10.19</mockito.verison>
 
         <maven.plugin.antrun.version>3.0.0</maven.plugin.antrun.version>
         <maven.plugin.compiler.version>3.8.1</maven.plugin.compiler.version>
@@ -292,12 +291,6 @@
                 </exclusions>
             </dependency>
 
-            <dependency>
-                <groupId>org.mockito</groupId>
-                <artifactId>mockito-core</artifactId>
-                <version>${mockito.verison}</version>
-                <scope>test</scope>
-            </dependency>
             <dependency>
                 <groupId>org.apache.hadoop</groupId>
                 <artifactId>hadoop-minicluster</artifactId>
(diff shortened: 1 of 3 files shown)


yaooqinn:
Thanks, @pan3793  for your first PR, and cc @ulysses-you

pom.xml
```diff
@@ -281,7 +295,7 @@
             <dependency>
                 <groupId>org.mockito</groupId>
                 <artifactId>mockito-core</artifactId>
-                <version>1.10.19</version>
+                <version>${mockito.verison}</version>
```
yaooqinn, line 298: Hi, this dependency is useless anymore, can we remove it?

pom.xml
```diff
@@ -66,7 +66,21 @@
         <scala.binary.version>2.12</scala.binary.version>
         <maven.version>3.6.3</maven.version>
         <spark.version>3.0.1</spark.version>
-        <log4j.version>1.7.30</log4j.version>
+        <slf4j.version>1.7.30</slf4j.version>
+        <mockito.verison>1.10.19</mockito.verison>
+
+        <maven.plugin.antrun.version>3.0.0</maven.plugin.antrun.version>
```
yaooqinn, line 72: The above version change looks fine. But I am -1 with these plugin changes. Not that straightforward as we only use them once each

pom.xml
pan3793: Sound good, removed.

pom.xml
pan3793: In my view, there are 2 benefits:
1. unify both dependencies and plugins versions to one place
2. if we want to try another version of plugin, just build with `-Dmaven.plugin.xxx.version=1.2.3`

yaooqinn: thanks for clarification

yaooqinn (approved):

pan3793:
@yaooqinn 
> And, we usually pin dependency version and exclude transitive dependencies in `<dependencyManagement>` block, but define the `scope` in specify `<dependency>` block, do you agree with that? If yes, I'd like to do it in another PR.

How about this?

yaooqinn:
ok

yaooqinn:
thanks @pan3793  merged to master and feel free to open PRs

PR #269, opened 2021-01-08
Support launch SparkProcessBuilder with keytab and principal

yaooqinn:
Temporary support for long-running engine instances in kerberized clusters. 

When the engine launched via proxy-user and the token expired, currently spark does not support update tokens for proxy users. we can enable session-level keytabs and principals

They can be  configured on the system side with default values in the form of user-defaults, like https://kyuubi.readthedocs.io/en/stable/deployment/settings.html#user-defaults

Or via JDBC Connection URL accordingly, like https://kyuubi.readthedocs.io/en/stable/deployment/settings.html#via-jdbc-connection-url

Both of them need to ensure the keytab is valid and reachable for now. Otherwise, we will fallback to the proxy user way.

### DIFF
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
@@ -17,9 +17,12 @@
 
 package org.apache.kyuubi.engine.spark
 
+import java.io.IOException
 import java.nio.file.{Files, Path, Paths}
 
 import scala.collection.mutable.ArrayBuffer
+
+import org.apache.hadoop.security.UserGroupInformation
 
 import org.apache.kyuubi._
 import org.apache.kyuubi.config.KyuubiConf
@@ -108,8 +111,11 @@ class SparkProcessBuilder(
       buffer += CONF
       buffer += s"$k=$v"
     }
-    buffer += PROXY_USER
-    buffer += proxyUser
+    // iff the keytab is specified, PROXY_USER is not supported
+    if (!useKeytab()) {
+      buffer += PROXY_USER
+      buffer += proxyUser
+    }
 
     mainResource.foreach { r => buffer += r }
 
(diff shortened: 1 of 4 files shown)


yaooqinn:
cc @ulysses-you

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -119,11 +125,32 @@ class SparkProcessBuilder(
...
+  private def useKeytab(): Boolean = {
+    val principal = conf.get(PRINCIPAL)
+    val keytab = conf.get(KEYTAB)
+    if (principal.isEmpty || keytab.isEmpty) {
+      false
+    } else {
+      try {
+        val ugi = UserGroupInformation
+          .loginUserFromKeytabAndReturnUGI(principal.get, keytab.get)
+        ugi.getShortUserName == proxyUser
```
yaooqinn, line 138: If the username are not satisfied, we will fail intentionally

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -119,11 +125,32 @@ class SparkProcessBuilder(
   override def toString: String = commands.mkString(" ")
 
   override protected def module: String = "kyuubi-spark-sql-engine"
+
+  private def useKeytab(): Boolean = {
+    val principal = conf.get(PRINCIPAL)
+    val keytab = conf.get(KEYTAB)
+    if (principal.isEmpty || keytab.isEmpty) {
```
ulysses-you, line 132: Do we need a exception or error msg if just specify one ? like what Spark do:
```
  require((principal == null) == (keytab == null),
    "Both principal and keytab must be defined, or neither.")
```

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -119,11 +125,32 @@ class SparkProcessBuilder(
...
+      false
+    } else {
+      try {
+        val ugi = UserGroupInformation
+          .loginUserFromKeytabAndReturnUGI(principal.get, keytab.get)
+        ugi.getShortUserName == proxyUser
+      } catch {
+        case e: IOException =>
+          error(s"Failed to login for ${principal.get}", e)
+          false
```
ulysses-you, line 142: Why here is `false` ? That means if the keytab has some problem we will ignore this and use proxy implictly.

yaooqinn: Like I described in the description, this feature is targeting end-users that from the non-spark world.  Currently, I prefer not to break their jobs. We can change the strategy whenever we want

yaooqinn: Maybe after we support get user specifed keytab from the client-side or somewhere else

PR #274, opened 2021-01-09
Handle InvocationTargetException and UndeclaredThrowableException in KSE

yaooqinn:
In Kyuubi, we use KSE to wark with TStatus.

This PR adds a function to extract the root cause from InvocationTargetException and UndeclaredThrowableException, this mostly happens when spark talks to Hive Metastore where has a lot of innovation by  java reflection calls .

Otherwise, the end-users may get error like this
```sql
Error: Error opening session SessionHandle [b9d43f01-4e8d-4cc9-a69a-9b492b565c4c] for username due to Error opening session SessionHandle [2b00f3d2-448b-432b-a470-08321b21535d] for username: null (state=,code=0)
```


yaooqinn:
cc @ulysses-you

PR #279, opened 2021-01-13
Fix high cpu load due to log capture thread not release

ulysses-you:
We will create a new log capture thread when open a new session, but we forgot release the thread after session opened success or failure.

Note that, we don't need to wait sub process finished because the log capture lifecycle is `SESSION_CREATE`  -> `SESSION_OPEN`. After `SESSION_OPEN` the log is controled by session.

### DIFF
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -106,28 +106,33 @@ class KyuubiSessionImpl(
         sessionConf.set(SparkProcessBuilder.APP_KEY, boundAppName.toString)
         sessionConf.set(HA_ZK_NAMESPACE, appZkNamespace)
         val builder = new SparkProcessBuilder(appUser, sessionConf.toSparkPrefixedConf)
-        val process = builder.start
-        info(s"Launching SQL engine: $builder")
-        var sh = getServerHost
-        val started = System.currentTimeMillis()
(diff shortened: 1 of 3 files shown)


ulysses-you:
cc @yaooqinn

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -106,28 +106,35 @@ class KyuubiSessionImpl(
...
-            throw KyuubiSQLException(s"Timed out($timeout ms) to launched Spark with $builder",
-              builder.getError)
-          }
-          sh = getServerHost
+          val Some((host, port)) = sh
+          openSession(host, port)
+        } finally {
+          // we must close the process builder whether session open is success or failure since
+          // we have a log capture thread in process builder.
+          builder.close()
```
yaooqinn, line 134: Shall we give it a more specific name here like stopLoggingThread

yaooqinn (approved):

kyuubi-main/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
ulysses-you: I think user(like `KyuubiSessionImpl`) does not need to care about whether `ProcessBuilder` use a log thread. `close` is a more common method that can remind user execute it.

yaooqinn:
thanks! merge to master for 1.1 and branch 1.0 for 1.0.2

PR #282, opened 2021-01-13
Downgrade libthrift:0.9.3 due to THRIFT-4805

pan3793:
because of THRIFT-4805, we don't upgrade to libthrift:0.12.0,
because of THRIFT-5274, we don't upgrade to libthrift:0.13.0,
so just keep hive-service-rpc:2.3.7 transitive dependency libthrift:0.9.3

### DIFF
--- a/kyuubi-common/pom.xml
+++ b/kyuubi-common/pom.xml
@@ -70,16 +70,6 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


yaooqinn (approved):

yaooqinn:
thanks~ merged to master

PR #284, opened 2021-01-14
Sesson submit log should be overwritten

ulysses-you:
We will create and write a log during submit a new session (like spark-submit) but we don't delete or overwrite it. This PR aims to add a time config to decide if we need to overwrite a session submit log.

The log path template is `KYUUBI_WORD_DIR/user/$engine-log.$index`. We will increase the `index` automatically if a user create some session concurrently.

After this PR. The log path template is `KYUUBI_WORD_DIR/user/$engine.log.$index`.

PR #289, opened 2021-01-15
[KYUUBI #284][FOLLOWUP] Fix a race condition in engine redirect log file initialization

ulysses-you:
ulysses-you [PR 289](https://github.com/yaooqinn/kyuubi/pull/289) [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

### Please add issue ID here?

fixes #284
closes #289

Fix the problem if some thread overwrite the same file.

### Test Plan:
- Add some test cases that check the changes thoroughly including negative and positive cases if possible
- Add screenshots for manual tests if appropriate
- [Run test](https://kyuubi.readthedocs.io/en/latest/tools/testing.html#running-tests) locally before make a pull request

### DIFF
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
@@ -60,32 +60,8 @@ trait ProcBuilder {
   // Visible for test
   private[kyuubi] var logCaptureThread: Thread = null
 
-  private def getRollAppendProcessLogFile: File = {
-    val processLogPath = workingDir
-    var index = 0
-    while (true) {
-      val file = new File(processLogPath.toFile, s"$module.log.$index")
-      if (file.exists()) {
-        val lastModified = file.lastModified()
-        if (lastModified < System.currentTimeMillis() - processLogRetainTimeMillis) {
-          return file
-        }
-        // retry if exists file has been modified recently
-      } else {
-        Files.createDirectories(processLogPath)
-        if (file.createNewFile()) {
-          return file
-        }
-        // retry if create failed due to create file concurrently
-      }
-      index = index + 1
-    }
-    // never reach here.
-    null
-  }
-
   final def start: Process = synchronized {
-    val procLogFile = getRollAppendProcessLogFile
+    val procLogFile = getRollAppendProcessLogFile(workingDir, module, processLogRetainTimeMillis)
     processBuilder.redirectError(procLogFile)
     processBuilder.redirectOutput(procLogFile)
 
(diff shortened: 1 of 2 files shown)


yaooqinn:
Please follow the PR template to add

```
fixes #289 
fixes #284
```
in the PR description

the help the maintainer automatically link and close related PRs and Issues

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -158,4 +134,30 @@ object ProcBuilder {
     }
   }
 
+  private def getRollAppendProcessLogFile(
+      workingDir: Path, module: String, processLogRetainTimeMillis: Long): File = synchronized {
```
yaooqinn, line 138: In the worst case, does this block to visit all files during `processLogRetainTimeMillis`(maybe a day or even longer?)

ulysses-you: Logically it is. I make a simple stress test offline. Seems it's OK.
```
(1 to 100).foreach { _ =>
          pool.execute(new Runnable {
            override def run(): Unit = {
              val pb = new FakeSparkProcessBuilder(config) {
                override val workingDir: Path = fakeWorkDir
              }
              try {
                val p = pb.start
                p.waitFor()
              } finally {
                pb.close()
              }
            }
          })
        }
```
```
val start = System.currentTimeMillis()
atomicTest()
println(System.currentTimeMillis() - start)
```

That shows only 500ms.

yaooqinn: Can we just move forward when an error occurs? and track the error count to be under the threshold. Listing plenty of files can increase random errors? through?

ulysses-you: We won't get a exception since it's a overwrite file. Update the new implementation.

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -158,4 +134,25 @@ object ProcBuilder {
...
+    val currentTime = System.currentTimeMillis()
+    val processLogPath = workingDir
+    val totalExistsFile = processLogPath.toFile.listFiles()
+    val sorted = totalExistsFile.sortBy(_.getName.split("\\.").last.toInt)
```
yaooqinn, line 142: this will throw an exception when the file is an invalid file

It's better to list files only with a filter that starts w/ `module` and skip those files that are invalid

ulysses-you: update it.

yaooqinn (approved):

yaooqinn:

PR #297, opened 2021-01-18
Remove unused property from build properties file

turboFei:
turboFei [PR 297](https://github.com/yaooqinn/kyuubi/pull/297) Target Issue Test Plan [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Now, there is no module named `kyuubi-server`.
And the property `jar` written into kyuubi-version-info.properties is not correct, it is also not used in `package` object.

https://github.com/yaooqinn/kyuubi/blob/6f8564ce595740feef1c6a0da85c06cd331cd753/kyuubi-common/src/main/scala/org/apache/kyuubi/package.scala#L37-L50

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,8 +23,6 @@
 build/apache-maven*
 build/scala*
 build/test
-kyuubi-server/build
-kyuubi-server/*example*
 target/
 dist/
 kyuubi-*-bin-*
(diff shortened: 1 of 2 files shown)


yaooqinn (approved):

yaooqinn:

PR #307, opened 2021-01-28
[KYUUBI #307]GetCatalogs supports DSv2 and keeps its backward compatibility

yaooqinn:
yaooqinn [PR 307](https://github.com/yaooqinn/kyuubi/pull/307) Feature [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

### Please add issue ID here?

Fixes #307

GetCatalogs supports DSv2 and keeps its backward compatibility

### Test Plan:
image

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetCatalogs.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetCatalogs.scala
@@ -17,7 +17,7 @@
 
 package org.apache.kyuubi.engine.spark.operation
 
-import org.apache.spark.sql.{Row, SparkSession}
+import org.apache.spark.sql.{AnalysisException, Row, SparkSession}
 import org.apache.spark.sql.types.StructType
 
 import org.apache.kyuubi.operation.OperationType
@@ -33,8 +33,12 @@ class GetCatalogs(spark: SparkSession, session: Session)
   }
 
   override protected def runInternal(): Unit = {
-    iter = Seq(
-      Row(spark.sessionState.catalogManager.currentCatalog.name())
-    ).toList.iterator
+     try {
+       iter = try {
+        spark.sql("SELECT CURRENT_CATALOG()").collect().toList.toIterator
+      } catch {
+        case _: AnalysisException => Seq(Row("spark_catalog")).toIterator
+      }
+    } catch onError()
   }
 }
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetTableTypes.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetTableTypes.scala
@@ -18,7 +18,6 @@
 package org.apache.kyuubi.engine.spark.operation
 
 import org.apache.spark.sql.{Row, SparkSession}
-import org.apache.spark.sql.catalyst.catalog.CatalogTableType
 import org.apache.spark.sql.types.StructType
 
 import org.apache.kyuubi.operation.OperationType
@@ -33,6 +32,6 @@ class GetTableTypes(spark: SparkSession, session: Session)
   }
 
   override protected def runInternal(): Unit = {
-    iter = CatalogTableType.tableTypes.map(t => Row(t.name)).toList.iterator
+    iter = Seq("EXTERNAL", "MANAGED", "VIEW").map(Row(_)).toList.iterator
   }
 }
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/Shim_v2_4.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/Shim_v2_4.scala
@@ -15,27 +15,12 @@
  * limitations under the License.
  */
 
-package org.apache.kyuubi.engine.spark.operation
+package org.apache.kyuubi.engine.spark.shim
 
-import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.{Row, SparkSession}
 
-import org.apache.kyuubi.engine.spark.shim.SparkShim
-import org.apache.kyuubi.operation.OperationType
-import org.apache.kyuubi.operation.meta.ResultSetSchemaConstant.TABLE_CAT
-import org.apache.kyuubi.session.Session
-
-class GetCatalogs(spark: SparkSession, session: Session)
-  extends SparkOperation(spark, OperationType.GET_CATALOGS, session) {
-
-  override protected def resultSchema: StructType = {
-    new StructType()
-      .add(TABLE_CAT, "string", nullable = true, "Catalog name. NULL if not applicable.")
-  }
-
-  override protected def runInternal(): Unit = {
-   try {
-     iter = SparkShim().getCatalogs(spark).toIterator
-    } catch onError()
+class Shim_v2_4 extends SparkShim {
+  override def getCatalogs(ss: SparkSession): Seq[Row] = {
+    Seq(Row("spark_catalog"))
   }
 }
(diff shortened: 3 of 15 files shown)


yaooqinn:
cc @ulysses-you @turboFei @pan3793

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetCatalogs.scala
```diff
@@ -33,8 +33,12 @@ class GetCatalogs(spark: SparkSession, session: Session)
...
 
   override protected def runInternal(): Unit = {
-    iter = Seq(
-      Row(spark.sessionState.catalogManager.currentCatalog.name())
-    ).toList.iterator
+     try {
+       iter = try {
+        spark.sql("SELECT CURRENT_CATALOG()").collect().toList.toIterator
+      } catch {
+        case _: AnalysisException => Seq(Row("spark_catalog")).toIterator
```
pan3793, line 40: How about fallback to previous action rather than hardcode?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetTableTypes.scala
```diff
@@ -33,6 +32,6 @@ class GetTableTypes(spark: SparkSession, session: Session)
   }
 
   override protected def runInternal(): Unit = {
-    iter = CatalogTableType.tableTypes.map(t => Row(t.name)).toList.iterator
+    iter = Seq("EXTERNAL", "MANAGED", "VIEW").map(Row(_)).toList.iterator
```
pan3793, line 35: ditto

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetCatalogs.scala
yaooqinn: The previous one calling unstable API of spark, we can improve this later(like retrieving the default catalog from conf, but none of these are stable yet)

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetTableTypes.scala
yaooqinn: Avoiding non-public API calling

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetCatalogs.scala
pan3793: I'm afraid -1 for the current approach. I'm using Spark 3.0.1 with Iceberg catalog, the change will broken our use case.
```
spark.sql.defaultCatalog=spark_catalog
spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions
spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog
spark.sql.catalog.iceberg=org.apache.iceberg.spark.SparkCatalog
spark.sql.catalog.iceberg.type=hadoop
spark.sql.catalog.iceberg.warehouse=hdfs://nameservice1/iceberg-warehouse
```

yaooqinn: Oh I see, but according to your configuration the catalogname is still spark_catalog?

pan3793: There are 2 catalogs: `spark_catalog` and `iceberg`

yaooqinn: OK, I just notice that the CURRENT_CATALOG does not get merged into Spark 3.0

turboFei:
Is it possible that create a util class for each in-compatible method with different spark version.


For example:
        <profile>
            <id>spark-3.0</id>
            <properties>
                <spark.version>3.0.0</spark.version>
                <src.version>3.0</src.version>
            </properties>
        </profile>

source code path: /v3.0/src/main/scala

            <plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>build-helper-maven-plugin</artifactId>
                <executions>
                    <execution>
                        <id>add-sources</id>
                        <phase>generate-sources</phase>
                        <goals>
                            <goal>add-source</goal>
                        </goals>
                        <configuration>
                            <sources>
                                <source>v${src.version}/src/main/scala</source>
                                <source>v${src.version}/src/main/java</source>
                            </sources>
                        </configuration>
                    </execution>
                    <execution>
                        <id>add-scala-test-sources</id>
                        <phase>generate-test-sources</phase>
                        <goals>
                            <goal>add-test-source</goal>
                        </goals>
                        <configuration>
                            <sources>
                                <source>v${src.version}/src/test/scala</source>
                                <source>v${src.version}/src/test/java</source>
                            </sources>
                        </configuration>
                    </execution>
                </executions>
            </plugin>

pan3793:
> Is it possible that create a util class for each in-compatible method with different spark version.

I recommend introduce a `SparkShim` rather than create different source folders.

yaooqinn:
> > Is it possible that create a util class for each in-compatible method with different spark version.
> 
> I recommend introduce a `SparkShim` rather than create different source folders.

I prefer this way

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/Shim_v3_0.scala
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.spark.shim
+
+import org.apache.spark.sql.{Row, SparkSession}
+
+class Shim_v3_0 extends Shim_v2_4 {
```
turboFei, line 22: Why not extend SparkShim?

yaooqinn: this is involving, we can add more catalog operation APIs here. and the higher shims can inherit or override the previous one.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/Shim_v2_4.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.spark.shim
+
+import org.apache.spark.sql.{Row, SparkSession}
+
+class Shim_v2_4 extends SparkShim {
+  override def getCatalogs(ss: SparkSession): Seq[Row] = {
+    Seq(Row("spark_catalog"))
```
yaooqinn, line 24: Does Spark v2.4 have the concept of catalog? Maybe here we shall return `""`, anyway we can check this later when supporting Spark v2.4

pom.xml
```diff
@@ -507,6 +509,13 @@
                 <artifactId>netty</artifactId>
                 <version>${netty3.version}</version>
             </dependency>
+
+            <dependency>
+                <groupId>org.apache.iceberg</groupId>
+                <artifactId>${iceberg.name}</artifactId>
+                <version>${iceberg.version}</version>
+                <scope>test</scope>
```
pan3793, line 517: How about define `scope` at sub modules `<dependency>` block? 
IMO, define `version` in  `<dependencyManagement>` and define `scope` at sub modules `<dependency>` make things more clear.

pan3793 (approved):
> I add a screenshot in the PR desc, is that what you expected @pan3793

Nice! LGTM, just few comments.

pom.xml
yaooqinn: can we have a lint tool  for this？

pan3793: Unfortunately, I haven't seen those tools.

PR #334, opened 2021-02-01
GetSchemas supports DSv2 multipart namespaces

yaooqinn:
yaooqinn [Closes #334](https://github.com/yaooqinn/kyuubi/pull/334)    Target Issue [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

This PR adds support for using Spark DSv2 to get Schemas with multipart namespaces under multiple catalogs.

The current works are all based on Apache Iceberg.

TODOS:
    next step we will support get tables operation for multiple catalogs and namespaces.

image

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/Shim_v3_0.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/Shim_v3_0.scala
@@ -52,8 +52,22 @@ class Shim_v3_0 extends Shim_v2_4 {
     case catalog: SupportsNamespaces =>
       // TODO: 1. We need explode here based on the impl of DSv2
       // TODO: 2. we need ensure how BI tools support multipart namespaces
-      val schemas = (catalog.defaultNamespace() ++ catalog.listNamespaces().map(_.head)).distinct
-      schemas.filter(_.matches(schemaPattern))
+      val rootSchema = catalog.listNamespaces()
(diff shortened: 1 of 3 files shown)


yaooqinn:
cc @cloud-fan, It seems to work fine with Dbeaver and Apache Iceberg, the `hive-jdbc` module also feed `.` to the client-side as a valid character.

We can make this happen in the Spark thrift server too.

@rdblue @aokolnychyi It would be nice if we could get help from the Apache Iceberg community

rdblue:
@yaooqinn, what would you like our help with?

yaooqinn:
> @yaooqinn, what would you like our help with?

@rdblue thanks for your response, I'd like to know whether the `listNamespaces` or `listTables` will support SQL wildcards `%` and `_`?

rdblue:
I'm not sure if there is a good justification for adding that to the interface. It would be simple enough to add, but there is the extra complexity in the interface to consider. I don't think that pushing the pattern into the interface adds much value because there aren't many cases where you actually need to push the pattern down to the underlying store. Databases probably have at most a few thousand tables, which is isn't a high cost to send as identifiers or a high cost to filter in Spark. So keeping the complexity in Spark seems like a good idea to me, rather than adding an extra method that implementations need to deal with.

yaooqinn:
Thanks for the clarification.  I see your point. +1 for keeping the simplification for DSv2 APIs. My biggest concern is the stability of the APIs, as they are mainly designed for the underlying systems to implement and for Spark to call, not for other readers like what we do here. So the stability and simplicity are good to me.

And thanks again.

PR #351, opened 2021-02-08
[KYUUBI #350] Avoid using HiveResult

### LINKED ISSUE #350: Avoid using HiveResult
**Describe the bug**
`HiveResult` is not spark public API, and it not compatible in spark 2.4, 3.0, 3.1 

**To Reproduce**
1. Configurations
2. Environments
3. Operations

**Expected behavior**

**Screenshots**

**Additional context**

pan3793:
pan3793 [Closes #351](https://github.com/yaooqinn/kyuubi/pull/351)    [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

close #350 


pan3793:
`HiveResult` has been replaced in `RowSet`, but still used in UT, let's wait CI report to ensure that `Formatter` will generate same result with `HiveResult`

yaooqinn (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -62,7 +63,15 @@ class ExecuteStatement(
       Thread.currentThread().setContextClassLoader(spark.sharedState.jarClassLoader)
       spark.sparkContext.setJobGroup(statementId, statement)
       result = spark.sql(statement)
-      iter = result.collect().toList.iterator
+      val castCols = result.schema.map { field =>
+        field.dataType match {
+          case BooleanType | ByteType | ShortType | IntegerType | LongType |
```
yaooqinn, line 68: `ByteType | ShortType | IntegerType | LongType | FloatType | DoubleType` to  _: NumericType for simplicity

yaooqinn: also add StringType to avoid casting

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/schema/RowSet.scala
```diff
@@ -102,11 +102,7 @@ object RowSet {
       case _ =>
```
yaooqinn, line 102: we can remove the `case StringType =>` branch,

and simply do

```
case _ =>
        val values = getOrSetAsNull[java.lang.String](rows, ordinal, nulls, "")
        TColumn.stringVal(new TStringColumn(values, nulls))
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
pan3793: `DecimalType` also is `NumericType` but not present here.

pan3793: I think `IntegralType | FloatType | DoubleType` is more precise since `DecimalType` is `NumericType` but should not be included.

pan3793: Oops! We can't use `IntegralType` because it is `private[sql]`
```
private[sql] abstract class IntegralType extends NumericType
```

yaooqinn: `IntegralType` isn't public, Let's leave it

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/schema/RowSet.scala
pan3793: updated

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
pan3793: updated

yaooqinn:
this is a huge behavior change
I will merge this first but still expecting more test cases

Issue #347, opened 2021-02-04
Does session have a timeout cleanup mechanism?


Honglun:
**Is your feature request related to a problem? Please describe.**
I can't find any configuration or code for this. If the user forgets to close the session, the Engine lives forever.

yaooqinn:
https://kyuubi.readthedocs.io/en/latest/deployment/settings.html#session

Honglun:
This will not work if the user does not actively close the session. See 'getOpenSessionCount' down here:
`override def run(): Unit = {
        while (getOpenSessionCount > 0 ||
          System.currentTimeMillis - latestLogoutTime < idleTimeout) {
          TimeUnit.MILLISECONDS.sleep(interval)
        }
        info(s"Idled for more than $idleTimeout ms, terminating")
        sys.exit(0)
      }`

yaooqinn:
kyuubi.session.timeout？

> This will not work if the user does not actively close the session. See 'getOpenSessionCount' down here:
> `override def run(): Unit = { while (getOpenSessionCount > 0 || System.currentTimeMillis - latestLogoutTime < idleTimeout) { TimeUnit.MILLISECONDS.sleep(interval) } info(s"Idled for more than $idleTimeout ms, terminating") sys.exit(0) }`

Honglun:
Yes, I missed this configuration. But there is still a problem, 'session.closeExpiredOperations' will constantly update the values of '_lastAccessTime' and '_lastIdleTime', so the session can't be closed. Method  'closeExpiredOperations' should remove 'withAcquirerease'.
https://github.com/yaooqinn/kyuubi/blob/e4d9390a74a33f09eb25b7f6ab9174c25d00b0e2/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala#L162

yaooqinn:
nice catch！ would you like to fix  that？

Honglun:
OK

Honglun:
@yaooqinn I didn't think it through last time. After the last expired Handle has been cleaned, the 'lastIdleTime' needs to be updated. Can you help re-open the issue?

PR #354, opened 2021-02-09
[KYUUBI #347] The 'lastIdleTime' needs to be updated after the last expired Handle has been cleaned

Honglun:
Honglun [Closes #354](https://github.com/yaooqinn/kyuubi/pull/354)    [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

The 'lastIdleTime' needs to be updated after the last expired Handle has been cleaned.

mvn test -Dtest=none -DwildcardSuites=org.apache.kyuubi.session.SessionManagerSuit

### DIFF
--- a/kyuubi-common/src/test/scala/org/apache/kyuubi/session/SessionManagerSuite.scala
+++ b/kyuubi-common/src/test/scala/org/apache/kyuubi/session/SessionManagerSuite.scala
@@ -20,29 +20,19 @@ package org.apache.kyuubi.session
 import java.time.Duration
 
 import org.apache.hive.service.rpc.thrift._
+import org.scalatest.concurrent.Eventually
+import org.scalatest.time.{Seconds, Span}
 
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.service.FrontendServiceSuite
 
-class SessionManagerSuite extends FrontendServiceSuite {
+class SessionManagerSuite extends FrontendServiceSuite with Eventually {
 
   override val conf = KyuubiConf()
     .set(KyuubiConf.FRONTEND_BIND_PORT, 0)
     .set("kyuubi.test.server.should.fail", "false")
     .set(KyuubiConf.SESSION_CHECK_INTERVAL, Duration.ofSeconds(5).toMillis)
     .set(KyuubiConf.SESSION_TIMEOUT, Duration.ofSeconds(5).toMillis)
-
-  def waitFor(waitForTime: Long)(f: () => Boolean): Unit = {
-    val startTime = System.currentTimeMillis()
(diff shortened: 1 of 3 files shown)


kyuubi-common/src/test/scala/org/apache/kyuubi/session/SessionManagerSuite.scala
```diff
@@ -0,0 +1,155 @@
...
+    server.initialize(conf)
+    server.start()
+  }
+
+  override def afterAll(): Unit = {
+    server.getServices.foreach(_.stop())
+    super.afterAll()
+  }
+
+  protected def withThriftClient(f: TCLIService.Iface => Unit): Unit = {
```
yaooqinn, line 51: can we reuse the existing code instead of copying them everywhere？

kyuubi-common/src/test/scala/org/apache/kyuubi/session/SessionManagerSuite.scala
```diff
@@ -0,0 +1,155 @@
...
+        try {
+           client.CloseSession(tCloseSessionReq)
+        } catch {
+          case e: Exception => error(s"Failed to close $handle", e)
+        }
+      }
+    }
+  }
+
+  def waitFor(waitForTime: Long)(f: () => Boolean): Unit = {
```
yaooqinn, line 93: can't we simply use `eventually`？

kyuubi-common/src/test/scala/org/apache/kyuubi/session/SessionManagerSuite.scala
Honglun: OK，I will solve this by extending 'FrontendServiceSuite'.

kyuubi-common/src/test/scala/org/apache/kyuubi/session/SessionManagerSuite.scala
Honglun: I don't know what you mean. Can you give me an example?

Honglun:
@yaooqinn I've simplified the common code for test cases.

kyuubi-common/src/test/scala/org/apache/kyuubi/session/SessionManagerSuite.scala
yaooqinn: I mean org.scalatest.concurrent.Eventually

Honglun: Thanks.

yaooqinn:

PR #370, opened 2021-02-25
[KYUUBI #360] Correct handle getNextRowSet with FETCH_PRIOR FETCH_FIRST

### LINKED ISSUE #360: FetchOrientation.FETCH_FIRST is not handled properly
**Describe the bug**

In `SparkOperation`, all `order` actually  are treat as `FETCH_NEXT`

```
  override def getNextRowSet(order: FetchOrientation, rowSetSize: Int): TRowSet = {
    validateDefaultFetchOrientation(order)
    assertState(OperationState.FINISHED)
    setHasResultSet(true)
    val taken = iter.take(rowSetSize)
    RowSet.toTRowSet(taken.toList, resultSchema, getProtocolVersion)
  }
```

**To Reproduce**
Use HUE to connect Kyuubi server and run a query, then export the result to csv, we found the records showed in web page is lost in exported data.

**Expected behavior**
Hue can export all records of query result.

**Screenshots**

**Additional context**

pan3793:
pan3793 [Closes #370](https://github.com/yaooqinn/kyuubi/pull/370)    Feature Bug [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

close #360

Ref: https://github.com/apache/spark/pull/30600


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/FetchIterator.scala
```diff
@@ -0,0 +1,107 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.spark
+
+private[engine] sealed trait FetchIterator[A] extends Iterator[A] {
```
yaooqinn, line 20: Let's honor Spark libraries and add some comment about it

pan3793: Updated

yaooqinn (approved):

yaooqinn:

PR #379, opened 2021-02-26
Improve engine log capture

yaooqinn:
yaooqinn [Closes #379](https://github.com/yaooqinn/kyuubi/pull/379)    Target Issue [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

1. limit error size with `kyuubi.session.engine.startup.error.max.size` to reduce long time pause or heap overhead
2. update `error` immediately when detecting errors at the very first line ASAP
3. Using `Exception:` instead of `Exception` to search each line to filter better
4. add the engine log address in the exception message in  case of error bootstraping


yaooqinn:
cc @pan3793 @ulysses-you  thanks for reviewing

Issue #386, opened 2021-03-03
Clean up discovery service before engine stop in connection share level


ulysses-you:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**
In connection level, we will release engine if the session connection is closed but we don't clean up the discovery service (i.e. the namespace in zookeeper). That may cause the disk stress after running a long time.

**Additional context**

yaooqinn:
Looks cool！

yaooqinn:
Mind take a lot at the flaky test https://travis-ci.org/github/yaooqinn/kyuubi/jobs/761385865, seems related

PR #387, opened 2021-03-03
[KYUUBI #386] Clean up discovery service before engine stop in connection share level

ulysses-you:
ulysses-you [Closes #387](https://github.com/yaooqinn/kyuubi/pull/387)    Test Plan [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

In connection level, we will release engine if the session connection is closed but we don't clean up the discovery service (i.e. the namespace in zookeeper). That may cause the disk stress after running a long time.

Add some new suites in spark engine moudle with zk emmbedded.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -90,6 +90,12 @@
             <artifactId>${iceberg.name}</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-test</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -20,8 +20,6 @@ package org.apache.kyuubi.engine.spark
 import java.time.Instant
 import java.util.concurrent.CountDownLatch
 
-import scala.util.control.NonFatal
-
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.SparkSession
 
@@ -30,7 +28,7 @@ import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf.ENGINE_SHARED_LEVEL
 import org.apache.kyuubi.engine.spark.SparkSQLEngine.countDownLatch
 import org.apache.kyuubi.ha.HighAvailabilityConf._
-import org.apache.kyuubi.ha.client.{RetryPolicies, ServiceDiscovery}
+import org.apache.kyuubi.ha.client.{EngineServiceDiscovery, RetryPolicies, ServiceDiscovery}
 import org.apache.kyuubi.service.{Serverable, ServiceState}
 import org.apache.kyuubi.util.SignalRegister
 
@@ -40,7 +38,7 @@ private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
   def this(spark: SparkSession) = this(classOf[SparkSQLEngine].getSimpleName, spark)
 
   override private[kyuubi] val backendService = new SparkSQLBackendService(spark)
-  private val discoveryService = new ServiceDiscovery(this)
+  private val discoveryService = new EngineServiceDiscovery(this)
 
   override def initialize(conf: KyuubiConf): Unit = {
     val listener = new SparkSQLEngineListener(this)
@@ -60,9 +58,6 @@ private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
   }
 
   override def stop(): Unit = {
-    if (getServiceState == ServiceState.STOPPED) {
-      return
-    }
     conf.get(ENGINE_SHARED_LEVEL) match {
       case "CONNECTION" =>
         info("Clean up discovery service due to this is connection share level.")
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -34,7 +34,6 @@ import org.apache.kyuubi.ha.client.{RetryPolicies, ServiceDiscovery}
 import org.apache.kyuubi.service.Serverable
 import org.apache.kyuubi.util.SignalRegister
 
-
 private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
   extends Serverable(name) {
 
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -31,7 +31,7 @@ import org.apache.kyuubi.config.KyuubiConf.ENGINE_SHARED_LEVEL
 import org.apache.kyuubi.engine.spark.SparkSQLEngine.countDownLatch
 import org.apache.kyuubi.ha.HighAvailabilityConf._
 import org.apache.kyuubi.ha.client.{RetryPolicies, ServiceDiscovery}
-import org.apache.kyuubi.service.Serverable
+import org.apache.kyuubi.service.{Serverable, ServiceState}
 import org.apache.kyuubi.util.SignalRegister
 
 private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
@@ -60,6 +60,9 @@ private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
   }
 
   override def stop(): Unit = {
+    if (getServiceState == ServiceState.STOPPED) {
+      return
+    }
     try {
       conf.get(ENGINE_SHARED_LEVEL) match {
         case "CONNECTION" =>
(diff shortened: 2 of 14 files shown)


externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -90,6 +90,12 @@
             <artifactId>${iceberg.name}</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-test</artifactId>
+            <scope>test</scope>
+        </dependency>
```
ulysses-you, line 98: This dependency is for zk embedded test.

ulysses-you:
cc @yaooqinn @pan3793

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/SparkSQLEngineListenerSuite.scala
```diff
@@ -20,7 +20,7 @@ package org.apache.kyuubi.engine.spark
 import org.apache.kyuubi.service.ServiceState
 
 class SparkSQLEngineListenerSuite extends WithSparkSQLEngine {
-  override def conf: Map[String, String] = Map.empty
+  override def withKyuubiConf: Map[String, String] = Map.empty
```
ulysses-you, line 23: I believe this name is more readable.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/WithSparkSQLEngine.scala
```diff
@@ -56,6 +59,11 @@ trait WithSparkSQLEngine extends KyuubiFunSuite {
 
   override def afterAll(): Unit = {
     super.afterAll()
+    // we need to clean up conf since it's the global config in same jvm.
+    withKyuubiConf.foreach { case (k, _) =>
+      System.clearProperty(k)
+      kyuubiConf.unset(k)
```
ulysses-you, line 65: Avoid affect other unit test, we should reset the config. The behavior like `withConf("" -> "") {}` but this is for the whole class.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -56,6 +59,22 @@ private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
...
+  override def stop(): Unit = {
+    try {
+      conf.get(ENGINE_SHARED_LEVEL) match {
+        case "CONNECTION" =>
+          info("Clean up discovery service due to this is connection share level.")
+          discoveryService.cleanup()
+        case _ =>
+      }
+    } catch {
+      case NonFatal(e) =>
```
yaooqinn, line 71: can we move it closer to where the exception comes out?

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/SparkSQLEngineListenerSuite.scala
yaooqinn: It's not only Kyuubi confs, right?

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/ShareLevelSparkEngineSuite.scala
```diff
@@ -0,0 +1,81 @@
...
+  override def withKyuubiConf: Map[String, String] = {
+    super.withKyuubiConf ++ Map(ENGINE_SHARED_LEVEL.key-> sharedLevel.toString)
+  }
+  override protected def jdbcUrl: String = getJdbcUrl
+  override val namespace: String = {
+    // for test, we always use uuid as namespace
+    s"/kyuubi/${sharedLevel.toString}/${UUID.randomUUID().toString}"
+  }
+
+  test("check discovery service is clean up with different share level") {
```
yaooqinn, line 47: can we just use a` foreach` here?

ulysses-you: a little diffcult since the conf is global. Otherhand, use different class is more simple if we add other ut in future.

yaooqinn: it is ambiguous that we need to start/stop engines for non-connection mode

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -56,6 +57,16 @@ private[spark] final class SparkSQLEngine(name: String, spark: SparkSession)
     backendService.sessionManager.startTerminatingChecker()
   }
 
+  override def stop(): Unit = {
```
yaooqinn, line 60: If we have EngineServiceDiscovery, do we need to hack here?

ulysses-you: Seems it's better to move it to `EngineServiceDiscovery`.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/ShareLevelSparkEngineSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+    withZkClient { zkClient =>
+      assert(engine.getServiceState == ServiceState.STARTED)
+      assert(zkClient.checkExists().forPath(namespace) != null)
+      withJdbcStatement() {_ => }
+      sharedLevel match {
+        // Connection level, we will cleanup namespace since it's always a global unique value.
+        case ShareLevel.CONNECTION =>
+          eventually(Timeout(120.seconds)) {
```
yaooqinn, line 55: What is this for?

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
```diff
@@ -0,0 +1,61 @@
...
+      case "CONNECTION" =>
+        cleanup()
+        info("Clean up discovery service due to this is connection share level.")
+
+      case _ =>
+    }
+    super.stop()
+  }
+
+  private def cleanup(): Unit = {
```
yaooqinn, line 49: can we simply put this in stop? it's simple and used only in one place

ulysses-you: That will make code look bad due to so many nested condition..
```scala
override def stop(): Unit = {
    conf.get(ENGINE_SHARED_LEVEL) match {
      // For connection level, we should clean up the namespace in zk in case the disk stress.
      case "CONNECTION" =>
        if (namespace != null) {
          try {
            zkClient.delete().deletingChildrenIfNeeded().forPath(namespace)
          } catch {
            case NonFatal(e) =>
              warn("Failed to clean up Spark engine before stop.", e)
          }
        }
```

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/SparkSQLEngineListenerSuite.scala
ulysses-you: The config mush with the prefix `kyuubi.` so that kyuubi could load.

yaooqinn: kyuubiConf.set is different, it accepts all pattern.
FYI, the iceberg configs are not `kyuubi.`-prefixed
https://github.com/yaooqinn/kyuubi/pull/387/files#diff-ebc4d984b839029afebf66e9c0fed77837f88b9a32fc81831dc681d6fc001dfaR25

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
yaooqinn:
```
case "CONNECTION" if (namespace != null) =>
```

yaooqinn: and here to use a single `if` instead of match looks neater

ulysses-you: OK, merge this.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/SparkSQLEngineListenerSuite.scala
ulysses-you: ah, yes we can load that. `withKyuubiConf` mean the conf will be loaded into kyuubi conf.

yaooqinn:
this change looks ok to me but seems to have a behavior change that the engine will not terminate by `NodeDelete` event. It's kind of a `backdoor` if we want it safely terminate an engine

yaooqinn (approved):

yaooqinn:
cc @pan3793 @zwangsheng

zwangsheng (approved):

yaooqinn:

ulysses-you:
thanks all !

Issue #397, opened 2021-03-04
Error running 'select function' command


Honglun:
**Describe the bug**

**To Reproduce**
1. Configurations
2. Environments
3. Operations

**Expected behavior**

select tmp_db.hello('mike')
Run the above SQL causes an error:
Error: Error operating EXECUTE_STATEMENT: org.apache.spark.sql.AnalysisException: cannot resolve '`tmp_db.hello(mike)`' given input columns: [tmp_db.hello(mike)];;
And SQL like this: select 'tmp_db.hello'

We see a problem here
https://github.com/yaooqinn/kyuubi/blob/master/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala#L78

**Screenshots**

**Additional context**

yaooqinn:
can it pass on Kyuubi 1.0.3?

Honglun:
> can it pass on Kyuubi 1.0.3?

Yes

yaooqinn:
could you help to verify https://github.com/yaooqinn/kyuubi/tree/397 for https://github.com/yaooqinn/kyuubi/pull/400

PR #400, opened 2021-03-04
Implement a simpler toHiveString to convert spark row

yaooqinn:
yaooqinn [Closes #400](https://github.com/yaooqinn/kyuubi/pull/400)    Target Issue Bug [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Hive compatibility is essential! This PR tries to restore the previous behavior.

1. Not use Spark's internal class HiveResult to convert spark catalyst type-values to JDBC results, the interval APIs are not APIs and changed unexpectedly. It involves a lot of overhead. (For this reason, @pan3793 and I propose a Casting way to solve it.)

2. Unfortunately, the casting way breaks things like #397. So this is a reverting PR.

3. I checked the spark `HiveResult` carefully, I noticed that the `nested` logic is not needed in Kyuubi as we already matched primitives ahead.
4. Also the date time formatter used in `HiveResult` is excessively packaged with massively complicated logic inside spark. It's not clear and simple enough to call.
5. (tiny breaking change)the trailing spaces of a timestamp in milliseconds may not be trimmed. But this is a correctness issue and acceptable.
 

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/schema/RowSet.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/schema/RowSet.scala
@@ -216,7 +216,7 @@ object RowSet {
       .withChronology(IsoChronology.INSTANCE)
   }
 
-  private lazy val simpleDateFormat = new SimpleDateFormat("yyyy-MM-dd", Locale.US)
+  private lazy val simpleDateFormatter = new SimpleDateFormat("yyyy-MM-dd", Locale.US)
 
   private lazy val timestampFormatter: DateTimeFormatter = {
     createBuilder().appendPattern("yyyy-MM-dd HH:mm:ss")
@@ -239,7 +239,7 @@ object RowSet {
         "null"
 
       case (d: Date, DateType) =>
-        simpleDateFormat.format(d)
+        simpleDateFormatter.format(d)
 
       case (ld: LocalDate, DateType) =>
         dateFormatter.format(ld)
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/schema/RowSet.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/schema/RowSet.scala
@@ -19,6 +19,13 @@ package org.apache.kyuubi.schema
 
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
+import java.sql.Timestamp
+import java.text.SimpleDateFormat
+import java.time.{Instant, LocalDate}
+import java.time.chrono.IsoChronology
+import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder, ResolverStyle}
+import java.time.temporal.ChronoField
+import java.util.{Date, Locale}
 
 import scala.collection.JavaConverters._
 import scala.language.implicitConversions
@@ -88,6 +95,10 @@ object RowSet {
         val values = getOrSetAsNull[java.lang.Double](rows, ordinal, nulls, 0.toDouble)
         TColumn.doubleVal(new TDoubleColumn(values, nulls))
 
+      case StringType =>
+        val values = getOrSetAsNull[java.lang.String](rows, ordinal, nulls, "")
+        TColumn.stringVal(new TStringColumn(values, nulls))
+
       case BinaryType =>
         val values = getOrSetAsNull[Array[Byte]](rows, ordinal, nulls, Array())
           .asScala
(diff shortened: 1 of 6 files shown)


yaooqinn:
The Projection-Cast way seems to introduce a regression of https://github.com/yaooqinn/kyuubi/issues/397, cc @pan3793 

In this PR, I re-implement the `toHiveString` partially from Spark

also cc @Honglun, can you help to verify this and thanks for the bug report

Honglun:
I don't understand the purpose of this code( from line 69 to 78). Why don't we just use 'result' like this 
'iter = new ArrayFetchIterator(result.collect())' at line 80 ?

Honglun:
Missing the file I mentioned above:
https://github.com/yaooqinn/kyuubi/blob/master/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala#L78

yaooqinn:
> I don't understand the purpose of this code( from line 69 to 78). Why don't we just use 'result' like this
> 'iter = new ArrayFetchIterator(result.collect())' at line 80 ?

the deleted code  is used for converting spark value to JDBC value

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/JDBCTests.scala
```diff
@@ -215,8 +215,8 @@ trait JDBCTests extends BasicJDBCTests {
       val resultSet = statement.executeQuery(
         "SELECT struct('1', '2') AS col1, named_struct('a', 2, 'b', 4) AS col2")
       assert(resultSet.next())
-      assert(resultSet.getObject("col1") === "[1, 2]")
-      assert(resultSet.getObject("col2") === "[2, 4]")
+      assert(resultSet.getObject("col1") === "{\"col1\":\"1\",\"col2\":\"2\"}")
+      assert(resultSet.getObject("col2") === "{\"a\":2,\"b\":4}")
```
pan3793, line 219: use """ to avoid back slash

pan3793 (approved):

yaooqinn:
Hi, @Honglun, can you tell us whether the patch works for you or not? If yes, can you make a follow-up PR to add your test case to the master branch to ensure we don't break anything related again?

also cc @ulysses-you @zwangsheng

Honglun:
> Hi, @Honglun, can you tell us whether the patch works for you or not? If yes, can you make a follow-up PR to add your test case to the master branch to ensure we don't break anything related again?
> 
> also cc @ulysses-you @zwangsheng

Is this patch already available for branch '397' ?

yaooqinn:
> > Hi, @Honglun, can you tell us whether the patch works for you or not? If yes, can you make a follow-up PR to add your test case to the master branch to ensure we don't break anything related again?
> > also cc @ulysses-you @zwangsheng
> 
> Is this patch already available for branch '397' ?

yes, I'd like you to help check your case locally first or I can just merge it to master then you start your investigation?

Honglun:
> > > Hi, @Honglun, can you tell us whether the patch works for you or not? If yes, can you make a follow-up PR to add your test case to the master branch to ensure we don't break anything related again?
> > > also cc @ulysses-you @zwangsheng
> > 
> > 
> > Is this patch already available for branch '397' ?
> 
> yes, I'd like you to help check your case locally first or I can just merge it to master then you start your investigation?

OK, Let me test branch 397 first.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/schema/RowSet.scala
```diff
@@ -164,19 +191,90 @@ object RowSet {
...
+   * A simpler impl of Spark's toHiveString
+   */
+  def toHiveString(dataWithType: (Any, DataType), timeZone: ZoneId): String = {
+    dataWithType match {
+      case (null, _) =>
+        // Only match nulls in nested type values
+        "null"
+
+      case (d: Date, DateType) =>
+        simpleDateFormat.format(d)
```
ulysses-you, line 242: nit: `dateFormatter.format(d)`

ulysses-you:
If I understand correctly, this PR try to fix two problem ?
1. result which is not alias can not be cast. 
2. timezone not respect the config `spark.sql.session.timeZone`.

yaooqinn:
> If I understand correctly, this PR try to fix two problem ?
> 
> 1. result which is not alias can not be cast.
> 2. timezone not respect the config `spark.sql.session.timeZone`.

I updated the PR description a bit for you guys to understand and review

Honglun:
I tested 'select function' locally, and it is normal.

yaooqinn:
> I tested 'select function' locally, and it is normal.

can you submit a PR with your test case later, and thanks for the help

Honglun:
> > I tested 'select function' locally, and it is normal.
> 
> can you submit a PR with your test case later, and thanks for the help

OK, I'll submit it later.

ulysses-you (approved):

ulysses-you:
thanks, merging to master for 1.1.0

PR #408, opened 2021-03-09
Fix engine log does not be overwrite

ulysses-you:
ulysses-you [Closes #408](https://github.com/yaooqinn/kyuubi/pull/408)    Target Issue Test Plan Bug [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Bug fix, otherwise the engine log will always increase size with append mode.

Add new test.

### DIFF
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
@@ -65,7 +65,7 @@ trait ProcBuilder {
   // Visible for test
   private[kyuubi] var logCaptureThread: Thread = _
 
-  private lazy val engineLog: File = ProcBuilder.synchronized {
+  private[kyuubi] lazy val engineLog: File = ProcBuilder.synchronized {
     val engineLogTimeout = conf.get(KyuubiConf.ENGINE_LOG_TIMEOUT)
     val currentTime = System.currentTimeMillis()
     val processLogPath = workingDir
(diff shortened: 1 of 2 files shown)


ulysses-you:
cc @pan3793 @yaooqinn

kyuubi-main/src/test/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilderSuite.scala
```diff
@@ -179,6 +180,27 @@ class SparkProcessBuilderSuite extends KerberizedTestHelper {
...
+    val fakeWorkDir = Files.createTempDirectory("fake")
+    val builder1 = new FakeSparkProcessBuilder(KyuubiConf()) {
+      override val workingDir: Path = fakeWorkDir
+    }
+    val file1 = builder1.engineLog
+    Files.write(file1.toPath, "a".getBytes(), StandardOpenOption.APPEND)
+    assert(file1.length() == 1)
+    Files.write(file1.toPath, "a".getBytes(), StandardOpenOption.APPEND)
+    assert(file1.length() == 2)
+    file1.setLastModified(System.currentTimeMillis() - Duration.ofDays(1).toMillis - 1000)
```
pan3793, line 194: Should we explicitly set ENGINE_LOG_TIMEOUT rather than depends on default value?

ulysses-you: Yeah, use the config exsplicitly.

yaooqinn:
image

let's encourage people to fulfill these fields, it can help a lot to reduce the burden of the release process

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -65,7 +65,7 @@ trait ProcBuilder {
   // Visible for test
   private[kyuubi] var logCaptureThread: Thread = _
 
-  private lazy val engineLog: File = ProcBuilder.synchronized {
+  private[kyuubi] lazy val engineLog: File = ProcBuilder.synchronized {
     val engineLogTimeout = conf.get(KyuubiConf.ENGINE_LOG_TIMEOUT)
     val currentTime = System.currentTimeMillis()
     val processLogPath = workingDir
```
pan3793, line 71: Scala support SAM since 2.12, we can simplify above code as
```
val totalExistsFile = processLogPath.toFile.listFiles { (_, name) => name.startsWith(module) }
```

yaooqinn (approved):

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
ulysses-you: magic !

ulysses-you:
thanks merging to branch-1.1 and master.

PR #409, opened 2021-03-09
[KYUUBI #410] Add Spark-3.1 trouble shooting

ulysses-you:
ulysses-you [Closes #409](https://github.com/yaooqinn/kyuubi/pull/409)    Test Plan Doc [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

More details could see the config `spark.sql.legacy.setCommandRejectsSparkCoreConfs`


yaooqinn:
this needs to be fixed. a trouble-shooting is necessary but not enough. for the first session that creates the engine, it's very like that the queue will be passed here

ulysses-you:
Yes, I considered to fix it, but it's a little hard to list the total illegal configs(need to collect the config in all module). We could fix it in next major release since it's have enough time.

yaooqinn:
let's create an issue to track this first

PR #416, opened 2021-03-11
Detected Dependency List Change for better dependency management

yaooqinn:
yaooqinn [Closes #416](https://github.com/yaooqinn/kyuubi/pull/416)    Target Issue [&#10088;?&#10089;](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Detect Dependency List Change, to let us and users know what actually release with Kyuubi

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -55,6 +55,8 @@ jobs:
         run: |
           mvn clean install ${{ matrix.profiles }} -Dmaven.javadoc.skip=true -V
           bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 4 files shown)


yaooqinn:
cc @pan3793 @ulysses-you please take a look

pan3793:
It's pretty good! And I think we also need to update document to tell contributors what they should do when change dependencies?

yaooqinn:
good point~ Let me update the doc

pan3793 (approved):

yaooqinn:

Issue #449, opened 2021-03-23
Add Basic Metrics Service Platform


yaooqinn:
### 1. Describe the feature

add engine, connection, statement and threads usage, JVM metrics etc., at the server side.

### 2. Motivation

monitoring

PR #450, opened 2021-03-24
[KYUUBI #449] Add Basic Metrics Service framework

yaooqinn:
yaooqinn [Closes #450](https://github.com/yaooqinn/kyuubi/pull/450)    [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

This PR adds a future-proofing Metrics Service framework close #499

add some reporters, like json file, jmx, etc..
add some basic metrics, like statement.error

WIP


yaooqinn:
Cc @zwangsheng @pan3793 @ulysses-you, can you check this proposal?

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -133,6 +133,18 @@
...
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-scala_2.12</artifactId>
```
pan3793, line 145: replace 2.12 by `${scala_binary_version}`

yaooqinn: nice catch~

pan3793:
Can we also support prometheus?

yaooqinn:
> Can we also support prometheus?

the current way works for prometheus, e.g. using jmx. 

Do you have any better idea?

kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.metrics.MetricsConf.METRICS_REPORTERS
+import org.apache.kyuubi.metrics.MetricsSystem.maybeSystem
+import org.apache.kyuubi.metrics.ReporterType._
+import org.apache.kyuubi.service.CompositeService
+
+class MetricsSystem extends CompositeService("MetricsSystem") {
+
+  private val registry = new MetricRegistry
+  private var counters: LoadingCache[String, Counter] = _
```
pan3793, line 36: Is the `LoadingCache` necessary? What if just use `registry.counter(key)`?

pan3793:
> > Can we also support prometheus?
> 
> the current way works for prometheus, e.g. using jmx.
> 
> Do you have any better idea?

I mean a `PrometheusReporter` which export dropwizard metrics to a HTTP restful api and return prometheus syntax data. But it seems not officially supported by dropwizard metrics framework.

yaooqinn:
Hi @pan3793 I am going the merge this, and I am also happy to see your PR with `micrometer`

PR #451, opened 2021-03-25
Support query auto timeout cancel on thriftserver

ulysses-you:
ulysses-you [Closes #451](https://github.com/yaooqinn/kyuubi/pull/451)    Target Issue Test Plan Feature [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Manual cherry-pick some Spark patch into Kyuubi.
1. [Support query auto timeout cancel on thriftserver](https://github.com/apache/spark/pull/29933)
2. [Add config to control if cancel invoke interrupt task on thriftserver](https://github.com/apache/spark/pull/30481)

In order to keep backward with early Spark version, we hard code the config key instead of refer to Spark SQLConf.

Note that, the exists timeout of operator (`kyuubi.operation.idle.timeout`) is to cancel that client has no access with engine. That said if a query run a long time and the client is alive, the query would not be cancelled. Then the new added config `spark.sql.thriftServer.queryTimeout` can handle this case.

Add new test.


ulysses-you:
cc @pan3793 @yaooqinn

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/Constants.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.spark
+
+object Constants {
```
yaooqinn, line 20: We should define `kyuubi.server.query.timeout` at server-side rather engine-side.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
```diff
@@ -746,4 +751,96 @@ class SparkOperationSuite extends WithSparkSQLEngine with JDBCTests {
       assert(tFetchResultsResp.getStatus.getStatusCode === TStatusCode.SUCCESS_STATUS)
     }
   }
+
+  test("SPARK-26533: Support query auto timeout cancel on thriftserver - setQueryTimeout") {
```
yaooqinn, line 755: we need to move these test to the superclass to test both the server and engine.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -32,9 +34,25 @@ class ExecuteStatement(
...
     session: Session,
     protected override val statement: String,
-    override val shouldRunAsync: Boolean)
+    override val shouldRunAsync: Boolean,
+    queryTimeout: Long)
   extends SparkOperation(spark, OperationType.EXECUTE_STATEMENT, session) with Logging {
 
+  // If a timeout value `queryTimeout` is specified by users and it is smaller than
+  // a global timeout value, we use the user-specified value.
+  // This code follows the Hive timeout behaviour (See #29933 for details).
```
pan3793, line 43: how about replace `#29933` by `SPARK-26533`?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/Constants.scala
pan3793: Users can't infer time unit from `kyuubi.server.query.timeout`, and there are some configurations defined in second and others defined in millisecond, should we make the configuration name more clear?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -502,6 +502,27 @@ object KyuubiConf {
       .timeConf
       .createWithDefault(Duration.ofSeconds(5).toMillis)
 
+  val OPERATION_FORCE_CANCEL: ConfigEntry[Boolean] =
+    buildConf("operation.interruptOnCancel")
```
yaooqinn, line 506: nit: we don't do camel style for configs in Kyuubi just `interrupt.on.cancel` for consistiency

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -502,6 +502,27 @@ object KyuubiConf {
...
+      .doc("Set a query duration timeout in seconds in Kyuubi. If the timeout is set to " +
+        "a positive value, a running query will be cancelled automatically when the timeout is " +
+        "exceeded, otherwise the query continues to run till completion. If timeout values are " +
+        "set for each statement via `java.sql.Statement.setQueryTimeout` and they are smaller " +
+        "than this configuration value, they take precedence. If you set this timeout and prefer " +
+        "to cancel the queries right away without waiting task to finish, consider enabling " +
+        s"${OPERATION_FORCE_CANCEL.key} together.")
+      .version("1.2.0")
+      .timeConf
+      .createWithDefault(Duration.ofSeconds(0).toMillis)
```
yaooqinn, line 524: is it just 0L?

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
```diff
@@ -71,12 +73,24 @@ class ExecuteStatement(
...
       val req = new TExecuteStatementReq(remoteSessionHandle, statement)
       req.setRunAsync(shouldRunAsync)
+      req.setQueryTimeout(getQueryTimeout)
       val resp = client.ExecuteStatement(req)
       verifyTStatus(resp.getStatus)
       _remoteOpHandle = resp.getOperationHandle
     } catch onError()
   }
 
+  private def getQueryTimeout: Long = {
```
yaooqinn, line 83: move this to sessionManager?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -32,9 +35,13 @@ class ExecuteStatement(
     spark: SparkSession,
     session: Session,
     protected override val statement: String,
-    override val shouldRunAsync: Boolean)
+    override val shouldRunAsync: Boolean,
+    queryTimeout: Long)
   extends SparkOperation(spark, OperationType.EXECUTE_STATEMENT, session) with Logging {
 
+  private val forceCancel =
+    session.sessionManager.getConf.get(KyuubiConf.OPERATION_FORCE_CANCEL)
```
yaooqinn, line 43: I guess we should get it from  `spark.conf`

yaooqinn:
a local test run is needed when new configs are comming

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
ulysses-you: It's a global config, can only passed by jdbc url.

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
ulysses-you: We only do the timeout check only in `ExecteStatement` operation, no stronge opinion to move that..

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/IndividualSparkSuite.scala
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.spark.scheduler.{SparkListener, SparkListenerTaskEnd}
+import org.apache.spark.sql.SparkSession
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
+import org.scalatest.time.SpanSugar._
+
+import org.apache.kyuubi.KyuubiFunSuite
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.operation.JDBCTestUtils
+
+class SparkEngineSuites extends KyuubiFunSuite {
```
ulysses-you, line 33: This new test class is for easily test which need modify the global config.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -502,6 +502,27 @@ object KyuubiConf {
...
+  val OPERATION_FORCE_CANCEL: ConfigEntry[Boolean] =
+    buildConf("operation.interrupt.on.cancel")
+      .doc("When true, all running tasks will be interrupted if one cancels a query. " +
+        "When false, all running tasks will remain until finished.")
+      .version("1.2.0")
+      .booleanConf
+      .createWithDefault(true)
+
+  val OPERATION_QUERY_TIMEOUT: ConfigEntry[Long] =
+    buildConf("operation.query.timeout")
```
ulysses-you, line 514: Decide to hide the Spark thing that in `SparkThriftServer,` so define the config with Kyuubi style @pan3793

ulysses-you:
After the refector, provide some summary for reviewer:

1. Add the query timeout at Kyuubi Server and that can be modified by session global config and statement query tiemout.
2. Engine side just receive the query timeout and to do the timeout check.
3. Add a interrupt flag to Spark local property with `true` default that is also a session global config.
4. In order to add engine test in one class eaily, add a new class `IndividualSparkSuite` and change some related test code place.

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
```diff
@@ -49,6 +50,17 @@ class KyuubiOperationManager private (name: String) extends OperationManager(nam
     tSessionHandle
   }
 
+  private def getQueryTimeout(statementTimeout: Long): Long = {
+    // If a timeout value `queryTimeout` is specified by users and it is smaller than
+    // a session timeout value, we use the user-specified value.
+    val sessionTimeout = getConf.get(KyuubiConf.OPERATION_QUERY_TIMEOUT)
+    if (sessionTimeout > 0 && (statementTimeout <= 0 || sessionTimeout < statementTimeout)) {
```
yaooqinn, line 57: what if sessionTimeout=-1 and statementTimeout=-1?

ulysses-you: nothing happened, that only add timeout checkout when timeout > 0.

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
```diff
@@ -49,6 +50,17 @@ class KyuubiOperationManager private (name: String) extends OperationManager(nam
     tSessionHandle
   }
 
+  private def getQueryTimeout(statementTimeout: Long, session: Session): Long = {
+    // If a timeout value `queryTimeout` is specified by users and it is smaller than
+    // a session timeout value, we use the user-specified value.
+    val sessionTimeout = session.sessionConf.get(KyuubiConf.OPERATION_QUERY_TIMEOUT)
```
yaooqinn, line 56: getConf.get(KyuubiConf.OPERATION_QUERY_TIMEOUT)?

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
```diff
@@ -49,6 +50,17 @@ class KyuubiOperationManager private (name: String) extends OperationManager(nam
     tSessionHandle
   }
 
+  private def getQueryTimeout(statementTimeout: Long, session: Session): Long = {
```
yaooqinn, line 53: we don't need session as parameter here

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
```diff
@@ -49,6 +50,17 @@ class KyuubiOperationManager private (name: String) extends OperationManager(nam
...
+    // If a timeout value `queryTimeout` is specified by users and it is smaller than
+    // a session timeout value, we use the user-specified value.
+    val sessionTimeout = getConf.get(KyuubiConf.OPERATION_QUERY_TIMEOUT)
```
yaooqinn, line 56: nit: rename session timeout to `systemQueryTimeout` and `statementTimeout` to `clientQueryTimeout`

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
ulysses-you: ah, good catch ~

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -100,4 +108,26 @@ class ExecuteStatement(
       executeStatement()
     }
   }
+
+  private def addTimeoutMonitor(): Unit = {
+    if (queryTimeout > 0) {
+      val timeoutExecutor = Executors.newSingleThreadScheduledExecutor()
```
pan3793, line 114: Use `NamedThreadFactory` to create thread

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
```diff
@@ -49,6 +50,18 @@ class KyuubiOperationManager private (name: String) extends OperationManager(nam
     tSessionHandle
   }
 
+  private def getQueryTimeout(clientQueryTimeout: Long): Long = {
+    // If a timeout value `queryTimeout` is specified by users and it is smaller than
+    // a session timeout value, we use the user-specified value.
```
yaooqinn, line 55: nit: also update comments

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
yaooqinn: org.apache.kyuubi.util.ThreadUtils$#newDaemonSingleThreadScheduledExecutor

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
pan3793: I think Kyuubi style timeConf defined in ms? The comments here is in seconds.

pan3793: Actually, I think we don't need specific time unit in kyuubi timeConf style, user should not care about what we use internal, they just define it via ISO-8601 duration string

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -502,6 +502,27 @@ object KyuubiConf {
...
+      .doc("Set a query duration timeout in seconds in Kyuubi. If the timeout is set to " +
+        "a positive value, a running query will be cancelled automatically when the timeout is " +
+        "exceeded, otherwise the query continues to run till completion. If timeout values are " +
+        "set for each statement via `java.sql.Statement.setQueryTimeout` and they are smaller " +
+        "than this configuration value, they take precedence. If you set this timeout and prefer " +
+        "to cancel the queries right away without waiting task to finish, consider enabling " +
+        s"${OPERATION_FORCE_CANCEL.key} together.")
+      .version("1.2.0")
+      .timeConf
+      .createWithDefault(0L)
```
pan3793, line 524: `Duration.ZERO.toMillis` has better semantic

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/Constants.scala
ulysses-you: @pan3793 
I see the point, but the exists conf already named with `*.timeout`.. How about enhance the `kyuubi-defaults.conf.template` ? I guess it's useful if we write some common config into it.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -502,6 +502,27 @@ object KyuubiConf {
...
+      .doc("When true, all running tasks will be interrupted if one cancels a query. " +
+        "When false, all running tasks will remain until finished.")
+      .version("1.2.0")
+      .booleanConf
+      .createWithDefault(true)
+
+  val OPERATION_QUERY_TIMEOUT: ConfigEntry[Long] =
+    buildConf("operation.query.timeout")
+      .doc("Set a query duration timeout in seconds in Kyuubi. If the timeout is set to " +
+        "a positive value, a running query will be cancelled automatically when the timeout is " +
```
yaooqinn, line 516: when the timeout is exceeded -> `when timeout` or `if timeout`

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -502,6 +502,27 @@ object KyuubiConf {
...
+        "exceeded, otherwise the query continues to run till completion. If timeout values are " +
```
yaooqinn, line 517: , otherwise -> `. Otherwise, `

yaooqinn (approved):
LGTM, if the test flakiness fixed

yaooqinn:
thanks, merged to master for v1.2.0 and branch 1.1 for v1.1.1

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
yaooqinn: Sorry, I missed this comment. What @pan3793 pointed out is true, we don't need to tell the unit of a time conf, it's Duration  ISO-8601 strings..

Here is a bug at caller side! the client timeout is conf, and the system timeout is ms... @ulysses-you

PR #461, opened 2021-03-26
[KYUUBI #456] Can not cancel SQL job if the timeout value is small

### LINKED ISSUE #456: Can not cancel SQL job if the timeout value is small
### 1. Describe the bug

We kown, in short Spark will compile the SQL to RDD then use `DAGScheduler` to execute the each partiiton. If the SQL string is large (probable over hundred lines), the SQL compilation may take a few time(several seconds). At this time with small query timeout limit(also several seconds), `sc.cancelJobGroup` may not find the SQL job in `DAGScheduler` since it's still in SQL compilation. As the result, the SQL job leak.

A simple pipeline:
```
a thread: 1) SQL string -> 2) SQL compilation -> 3) DAGScheduler -> ...

b thread: 1) cancelJobGroup -> 2) DAGScheduler
```
When thread b at step 2, the thread a may stay at step 2, so can not cancel the job correctly.

The reason about the slow SQL compilation can be:
1. the SQL it is large enough
2. the slow rpc with Hive
3. since the Hive catalog in Spark driver use the global lock, it might be a reason in some case.

Some option to solve such case:
1. interrupt the SQL compliation thread and then do cancelJobGroup
2. check job is exits for a memont until job exists and then do cancelJobGroup

ulysses-you:
ulysses-you [Closes #461](https://github.com/yaooqinn/kyuubi/pull/461)    Test Plan Feature [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

* avoid condition race
* avoid job leak

pass `org.apache.kyuubi.engine.spark.SparkEngineSuites`
close #456

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -115,7 +115,6 @@ class ExecuteStatement(
       timeoutExecutor.schedule(new Runnable {
         override def run(): Unit = {
           cleanup(OperationState.TIMEOUT)
-          Option(getBackgroundHandle).foreach(_.cancel(true))
           timeoutExecutor.shutdown()
         }
       }, queryTimeout, TimeUnit.SECONDS)
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -116,15 +116,8 @@ class ExecuteStatement(
         ThreadUtils.newDaemonSingleThreadScheduledExecutor("query-timeout-thread")
       timeoutExecutor.schedule(new Runnable {
         override def run(): Unit = {
-          try {
-            cleanup(OperationState.TIMEOUT)
-          } catch {
-            case NonFatal(e) =>
-              setOperationException(KyuubiSQLException(e))
-              error(s"Error cancelling the query after timeout: $queryTimeout seconds")
-          } finally {
-            timeoutExecutor.shutdown()
-          }
+          cleanup(OperationState.TIMEOUT)
+          timeoutExecutor.shutdown()
         }
       }, queryTimeout, TimeUnit.SECONDS)
     }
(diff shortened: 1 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -47,6 +45,7 @@ class ExecuteStatement(
     OperationLog.createOperationLog(session.handle, getHandle)
   override def getOperationLog: Option[OperationLog] = Option(operationLog)
   private var result: DataFrame = _
+  private var statementFuture: Option[Future[_]] = _
```
yaooqinn, line 48: not neccesary

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -116,18 +114,9 @@ class ExecuteStatement(
...
-            }
-          } catch {
-            case NonFatal(e) =>
-              setOperationException(KyuubiSQLException(e))
-              error(s"Error cancelling the query after timeout: $queryTimeout seconds")
-          } finally {
-            timeoutExecutor.shutdown()
-          }
+          cleanup(OperationState.TIMEOUT)
+          getBackgroundHandle.cancel(true)
```
yaooqinn, line 118: avoid potential npe

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -116,18 +114,9 @@ class ExecuteStatement(
...
-            }
-          } catch {
-            case NonFatal(e) =>
-              setOperationException(KyuubiSQLException(e))
-              error(s"Error cancelling the query after timeout: $queryTimeout seconds")
-          } finally {
-            timeoutExecutor.shutdown()
-          }
+          cleanup(OperationState.TIMEOUT)
+          Option(getBackgroundHandle).foreach(_.cancel(true))
```
yaooqinn, line 118: is this should be inside cleanup? as Non-timeout cancelation also need this?

yaooqinn:
also cc @pan3793

yaooqinn (approved):

yaooqinn:
thanks, merged to master for v1.2.0 and branch 1.1 for v1.1.1

PR #469, opened 2021-03-28
Using Optional Config for query timeout

yaooqinn:
yaooqinn [Closes #469](https://github.com/yaooqinn/kyuubi/pull/469)    Target Issue Bug [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

1. bugfix for comparing a client-side value with a second unit to a server-side value with a millisecond unit
2. using OptionalConfigEntry for query timeout, it's more simple, it will take effect only when set with a valid value
3. add tests for system query timeout
4. move timeout conf getter to `initialize` to detect during Kyuubi server startup, the current behavior is a bit dangerous for deploying. 


yaooqinn:
cc @turboFei @ulysses-you @pan3793

turboFei:
IIUC, client get the timeout from server side with Second Unit.

Then it submit execute query request with this Second UNIT timeout.

This timeout transferred to kyuubi operation and I do not find that is was transferred to millsecond unit.

But for the operation `isTimeout` method,  it's still compared with millsecond.
https://github.com/yaooqinn/kyuubi/blob/5684407ce064a26286871abbdaaf68a0cf5c2612/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala#L156-L160

yaooqinn:
> IIUC, client get the timeout from server side with Second Unit.
> 
> Then client submits ExecuteQuery request with this Second UNIT timeout.
> 
> This it is transferred to kyuubi operation and I do not find that is was transferred to millsecond unit.
> 
> But for the operation `isTimeout` method, it's still compared with millsecond.
> https://github.com/yaooqinn/kyuubi/blob/5684407ce064a26286871abbdaaf68a0cf5c2612/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala#L156-L164

the query timeout is different from the operation timeout. 

the query timeout is a client setting but not all client-tools have such feature supported. So for server-robust, we introduced a query timeout threshold at the server-side. In case a query hangs forever at the engine side or finishes execution in time, we should `CANCEL` it, and then the client will receive a error status for it's own failover logic or something else.

the timeout that you point out is not the query timeout, but the operation idle timeout. the idle one is used to tell the monitor whether to `CLOSE` if the client remains radio-silence or zombied during the timeout. in this case, the client might need to know nothing. 

> This it is transferred to kyuubi operation and I do not find that is was transferred to millsecond unit.

All timeConfs in Kyuubi are based on `java.util.Duration` and be converted to milliseconds. FYI, you can check  the impl of `timeConf`'

turboFei (approved):

PR #471, opened 2021-03-30
Separate Embedded Zookeeper server from ha

yaooqinn:
yaooqinn [Closes #471](https://github.com/yaooqinn/kyuubi/pull/471)    Target Issue [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

1. Separated embedded zk from ha module, as it's only used for setup a zk from Kyuubi server for now if there is not an external one
2. Support more configuration for the embedded zookeeper, for example, HOSTNAME, then it's more close to prod usage.
3. TODO: Make the embedded zookeeper of our own, which should support as quorum peers


yaooqinn:
cc @turboFei @pan3793 @Honglun @ulysses-you

pan3793 (approved):

kyuubi-zookeeper/src/main/scala/org/apache/kyuubi/zookeeper/EmbeddedZookeeper.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  override def start(): Unit = synchronized {
+    server.start()
+    info(s"$getName is started at $getConnectString")
+    sys.addShutdownHook(server.close())
+    super.start()
+  }
+
+  override def stop(): Unit = synchronized {
+    if (getServiceState == ServiceState.STARTED) {
```
yaooqinn, line 79: with L74, can we bypass the race issue in the test with zk?  @ulysses-you

ulysses-you: We have already added ShutdownHook in KyuubiServer ? The state check can reduce the possibility of race .. but I think it's still here.

yaooqinn: I shall remove the hook there, but we will implement a zk of our own

turboFei (approved):

Honglun (approved):

yaooqinn:

Issue #304, opened 2021-01-26
Driver is not terminated immediately after Kyuubi server is stopped


Honglun:
when we modify the configuration of kyuubi and then restart the kyuubi server，we have to manually kill the yarn app or the SparkSubmit process, otherwise, the driver will use the old configuration, which is not what we expect.

yaooqinn:
https://kyuubi.readthedocs.io/en/latest/quick_start/quick_start.html#stopping-kyuubi
this is in doc

Or do you have a better idea?

yaooqinn:
For the case of failover, recreating a Kyuubi server is light

Honglun:
When Kyuubi server, can all drivers be terminated at the same time? Or add an option to terminate all drivers at once

yaooqinn:
In HA mode or just a failover case, a server terminated does not mean that the engines should stop too, they can still be reached by other peers or a recovered server.

we have provided several configs in the doc for an engine's lifecycle, you can check that.

BTW, I am quite open with a better idea for managing engines

Honglun:
I see what you mean.  I'm not talking about changing the current design, I mean add a new feature that will allow us to stop all the engines in time.

yaooqinn:
> I see what you mean. I'm not talking about changing the current design, I mean add a new feature that will allow us to stop all the engines in time.

The new feature would be planned into v1.1, it would be better if you happen to have a solution

yaooqinn:
This is partially supported now for the `connection` share level egnines

yaooqinn:
Keep it open and retarget it to v1.2.0 to see if we can get a better idea for this. cc @ulysses-you as the release manager of v1.1.0

ulysses-you:
Sorry for the vert late reply. The idea looks like useful.

The behavior more like `decommission`which is a liitle different with `stop`, that means the whole Kyuubi is not available and the all reousrce (not only the application but also some tmp file if exists) of Kyuubi will be released.

Maybe we can expose a new script, and user can execute like `sh decommission.sh`. I hope it could be done in v1.2.0 but it would be great if you can make a PR for this. : )

Honglun:
'ServiceDiscovery' create a 'stopTag' node in zookeeper during initialization, and add watchers on the 'stopTag' node for all the servers and engines. When we want to stop all the servers, we only need to remove the 'stopTag' node.
How about this one?

yaooqinn:
[...]

PR #481, opened 2021-04-03
[KYUUBI #304] Validate specified host and port in arguments parse phase

turboFei:
turboFei [Closes #481](https://github.com/yaooqinn/kyuubi/pull/481)    Test Plan [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Validate specified host and port in arguments parse phase.

UT

### DIFF
--- a/kyuubi-ctl/src/test/scala/org/apache/kyuubi/ctl/KyuubiCtlArgumentsSuite.scala
+++ b/kyuubi-ctl/src/test/scala/org/apache/kyuubi/ctl/KyuubiCtlArgumentsSuite.scala
@@ -56,15 +56,17 @@ class KyuubiCtlArgumentsSuite extends KyuubiFunSuite {
           "--port", port,
           "--version", KYUUBI_VERSION
         )
-        val opArgs = new KyuubiCtlArguments(args)
-        assert(opArgs.action.toString.equalsIgnoreCase(command))
-        assert(opArgs.service.toString.equalsIgnoreCase(service))
-        assert(opArgs.zkQuorum == zkQuorum)
-        assert(opArgs.nameSpace == namespace)
-        assert(opArgs.user == user)
-        assert(opArgs.host == host)
-        assert(opArgs.port == port)
-        assert(opArgs.version == KYUUBI_VERSION)
+        if (!(command == "create" && service == "engine")) {
+          val opArgs = new KyuubiCtlArguments(args)
+          assert(opArgs.action.toString.equalsIgnoreCase(command))
+          assert(opArgs.service.toString.equalsIgnoreCase(service))
+          assert(opArgs.zkQuorum == zkQuorum)
+          assert(opArgs.nameSpace == namespace)
+          assert(opArgs.user == user)
+          assert(opArgs.host == host)
+          assert(opArgs.port == port)
+          assert(opArgs.version == KYUUBI_VERSION)
+        }
       }
     }
   }
@@ -133,8 +135,8 @@ class KyuubiCtlArgumentsSuite extends KyuubiFunSuite {
     assert(opArgs.action == KyuubiCtlAction.LIST)
   }
 
-  test("test create/get/delete action arguments") {
-    Seq("create", "get", "delete").foreach { op =>
+  test("test get/delete action arguments") {
+    Seq("get", "delete").foreach { op =>
       val args = Array(
         op
       )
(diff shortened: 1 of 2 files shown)


kyuubi-ctl/src/test/scala/org/apache/kyuubi/ctl/KyuubiCtlArgumentsSuite.scala
```diff
@@ -201,4 +201,35 @@ class KyuubiCtlArgumentsSuite extends KyuubiFunSuite {
     assert(opArgs3.verbose)
     assert(opArgs3.action == KyuubiCtlAction.HELP)
   }
+
+  test("test with unknown host") {
+    val args = Array(
+      "create", "server",
+      "-zk", zkAddress,
+      "-ns", namespace,
```
pan3793, line 209: I'm not sure if it's a good arg style. personally, I prefer
- single-letter Unix-style, i.e. `-v -p 1234` or `-v -p1234` or `-vp1234`
- all-word  GNU-style, i.e. `--verbose --port 1234` or `--verbose --port=1234`

But `java -version` dose not use that style

turboFei:
```
Arguments:
  --zkAddress, -zk            one of the zk ensemble address, using kyuubi-defaults/conf
                              if absent
  --namespace, -ns            the namespace, using kyuubi-defaults/conf if absent
  --host, -h                  hostname or IP address of a service
  --port, -p                  listening port of a service
  --version, -V               using the compiled KYUUBI_VERSION default, change it if the
                              active service is running in another
  --user, -u                  for engine service only, the user name this engine belong to
  --help, -I                  Show this help message and exit.
  --verbose, -v               Print additional debug output.
```

pan3793: Yes, I saw that, but we haven't release v1.2.0, if we have a better choice, the change is safe.

turboFei: do you mean that, align `--version` with `-v`?

pan3793: I suggest 

- change `--zkAddress, -zk` to `--zk-addr, -z`
- change `--help, -I` to `--help, -I`

to make it match more command line tool arg styles.

also cc @yaooqinn

turboFei: could you clear it for `--help`? @pan3793

yaooqinn: Well, I am not a fan of abbreviations, while it always confusing me such as `-h` here is for help or short. Shall we only use fully qualified parameters for v1.2.0?

pan3793: @turboFei sorry for typo, fixed

turboFei: sure, we can use full qualified parameters at first.

pan3793: @yaooqinn I'm OK with only GNU-style, especially in shell script, the single-letter arg is not clear sometimes.

turboFei: should we change `--zkAddress` to `--zk-addr`?
cc @yaooqinn

yaooqinn: on second thought, it's better to use `--zk-quorum` to match with `kyuubi.ha.zookeeper.quorum`

pan3793: +1 for `--zk-quorum`, please also change the text, `one of the zk ensemble address` is not properly here

yaooqinn:
And also for the `CREATE` action, let's limit its use case only work for the `Kyuubi Server` that created in the same directory that the `kyuubi-ctl` runs, not for creating engines and a custom instance of a server. As an instance for exposing needs more verification than just a host and port, it's very likely for users to abuse. Consequently, we only allow `--zk-quorum` and namespace for `CREATE` action to expose the current server instance to another domain. Maybe that's enough for potential use cases that I can imagine so far.

turboFei:
> And also for the `CREATE` action, let's limit its use case only work for the `Kyuubi Server` that created in the same directory that the `kyuubi-ctl` runs, not for creating engines and a custom instance of a server. As an instance for exposing needs more verification than just a host and port, it's very likely for users to abuse it. Consequently, we only allow `--zk-quorum` and namespace for `CREATE` action to expose the current server instance to another domain. Maybe that's enough for potential use cases that I can imagine so far.

ok， do you means that copy all the kyuubi server zk nodes to another namespace?

yaooqinn:
> because kyuubi-ctl is only a client, and can not get the current server instance but get all the server instances under same namespace.

A I see, then let's not support engines for now

yaooqinn (approved):

yaooqinn:

PR #488, opened 2021-04-06
Add kyuubi-shaded module to avoid conflicts with external jars

turboFei:
turboFei [Closes #488](https://github.com/yaooqinn/kyuubi/pull/488)    Target Issue Feature [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Shade curator dependency in kyuubi-ha module to avoid impaction by external jars.

Existing UT.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -68,7 +68,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-io-
       - name: install kyuubi-shaded jars
-        run: cd kyuubi-shaded && mvn clean install && cd -
+        run: mvn clean install -pl :kyuubi-shaded-curator,:kyuubi-shaded-hive-service-rpc -am -Pkyuubi-shaded
       - name: Build with Maven
         run: |
           mvn clean install ${{ matrix.profiles }} -Dmaven.javadoc.skip=true -V
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -67,6 +67,8 @@ jobs:
           key: ${{ runner.os }}-maven-io-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-io-
+      - name: install kyuubi-shaded jars
+        run: cd kyuubi-shaded && mvn clean install && cd -
       - name: Build with Maven
         run: |
           mvn clean install ${{ matrix.profiles }} -Dmaven.javadoc.skip=true -V
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,7 @@ matrix:
 
 install:
   - mvn --version
-  - cd kyuubi-shaded
-  - mvn clean install
-  - cd -
+  - mvn clean install -pl :kyuubi-shaded-curator,:kyuubi-shaded-hive-service-rpc -am -Pkyuubi-shaded
 
 script:
   - mvn clean install $PROFILE -Dmaven.plugin.scalatest.exclude.tags=$EXCLUDE_TAGS -Dmaven.javadoc.skip=true -V
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,9 @@ matrix:
 
 install:
   - mvn --version
+  - cd kyuubi-shaded
+  - mvn clean install
+  - cd -
 
 script:
   - mvn clean install $PROFILE -Dmaven.plugin.scalatest.exclude.tags=$EXCLUDE_TAGS -Dmaven.javadoc.skip=true -V
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -170,6 +170,7 @@
                             <include>org.apache.curator:curator-client</include>
                             <include>org.apache.curator:curator-framework</include>
                             <include>org.apache.curator:curator-recipes</include>
+                            <include>org.apache.hive:hive-service-rpc</include>
                         </includes>
                     </artifactSet>
                     <relocations>
@@ -178,6 +179,13 @@
                             <shadedPattern>${kyuubi.shade.packageName}.curator</shadedPattern>
                             <includes>
                                 <include>org.apache.curator.**</include>
+                            </includes>
+                        </relocation>
+                        <relocation>
+                            <pattern>org.apache.hive</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.hive</shadedPattern>
+                            <includes>
+                                <include>org.apache.hive.**</include>
                             </includes>
                         </relocation>
                     </relocations>
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -100,6 +100,12 @@
 
         <dependency>
             <groupId>org.apache.hive</groupId>
+            <artifactId>hive-service-rpc</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hive</groupId>
             <artifactId>hive-llap-client</artifactId>
             <scope>test</scope>
         </dependency>
@@ -125,12 +131,6 @@
         <dependency>
             <groupId>org.apache.iceberg</groupId>
             <artifactId>${iceberg.name}</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.curator</groupId>
-            <artifactId>curator-test</artifactId>
             <scope>test</scope>
         </dependency>
 
@@ -167,28 +167,10 @@
                         <includes>
                             <include>org.apache.kyuubi:kyuubi-common</include>
                             <include>org.apache.kyuubi:kyuubi-ha</include>
-                            <include>org.apache.curator:curator-client</include>
-                            <include>org.apache.curator:curator-framework</include>
-                            <include>org.apache.curator:curator-recipes</include>
-                            <include>org.apache.hive:hive-service-rpc</include>
+                            <include>org.apache.kyuubi:kyuubi-shaded-curator</include>
+                            <include>org.apache.kyuubi:kyuubi-shaded-hive-service-rpc</include>
                         </includes>
                     </artifactSet>
-                    <relocations>
-                        <relocation>
-                            <pattern>org.apache.curator</pattern>
-                            <shadedPattern>${kyuubi.shade.packageName}.curator</shadedPattern>
-                            <includes>
-                                <include>org.apache.curator.**</include>
-                            </includes>
-                        </relocation>
-                        <relocation>
-                            <pattern>org.apache.hive</pattern>
-                            <shadedPattern>${kyuubi.shade.packageName}.hive</shadedPattern>
-                            <includes>
-                                <include>org.apache.hive.**</include>
-                            </includes>
-                        </relocation>
-                    </relocations>
                 </configuration>
                 <executions>
                     <execution>
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -167,7 +167,8 @@
                         <includes>
                             <include>org.apache.kyuubi:kyuubi-common</include>
                             <include>org.apache.kyuubi:kyuubi-ha</include>
-                            <include>org.apache.curator:*</include>
+                            <include>org.apache.curator:curator-framework</include>
+                            <include>org.apache.curator:curator-recipes</include>
                         </includes>
                     </artifactSet>
                     <relocations>
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -167,8 +167,18 @@
                         <includes>
                             <include>org.apache.kyuubi:kyuubi-common</include>
                             <include>org.apache.kyuubi:kyuubi-ha</include>
+                            <include>org.apache.curator:*</include>
                         </includes>
                     </artifactSet>
+                    <relocations>
+                        <relocation>
+                            <pattern>org.apache.curator</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.curator</shadedPattern>
+                            <includes>
+                                <include>org.apache.curator.**</include>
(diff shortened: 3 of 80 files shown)


pom.xml
```diff
@@ -105,6 +105,8 @@
         <maven.plugin.download.version>1.6.0</maven.plugin.download.version>
         <maven.plugin.silencer.version>1.6.0</maven.plugin.silencer.version>
         <maven.plugin.rat.version>0.13</maven.plugin.rat.version>
+        <!-- Package to use when relocating shaded classes. -->
+        <kyuubi.shade.packageName>org.kyuubi_project</kyuubi.shade.packageName>
```
yaooqinn, line 109: `org.apache.kyuubi.shade` looks more clear to me

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -167,8 +167,18 @@
                         <includes>
                             <include>org.apache.kyuubi:kyuubi-common</include>
                             <include>org.apache.kyuubi:kyuubi-ha</include>
+                            <include>org.apache.curator:*</include>
```
yaooqinn, line 170: can we avoid using wildcard here?

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -167,8 +167,18 @@
...
                         </includes>
                     </artifactSet>
+                    <relocations>
+                        <relocation>
+                            <pattern>org.apache.curator</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.curator</shadedPattern>
+                            <includes>
+                                <include>org.apache.curator.**</include>
```
yaooqinn, line 178: what about the transitive dependencies, shall we explicitly exclude them or add them here too? cc @pan3793

externals/kyuubi-spark-sql-engine/pom.xml
turboFei: I tried to include curator-framework and curator-recipes here.

Let me wait the ut result and determine whether also include curator-test here.

turboFei: ok，include curator-client

pan3793:
Do shade on `kyuubi-spark-sql-engine` looks like not safe for me.
If we define a different version of curator and hive from spark, it will overwrite transitive dependencies of spark in `kyuubi-spark-sql-engine` compile and test scopes. 

For example, kyuubi uses curator4 and hive3, but spark30 depends curator2 and hive2.
In `kyuubi-spark-sql-engine` compile and test scopes, the actually effected version is curator4 and hive3.

pan3793:
I recommend to pre-shade the `hive-services-rpc` and `curator-*` in separated modules, and let `kyuubi-spark-engine` depends it. (or we should let `kyuubi-common` depends it, and other modules just use the transitive dependencies).

AFAK, Apache Flink use this approach, and they maintains the pre-shaded dependencies in a separated repo https://github.com/apache/flink-shaded . Apache Hadoop has a plan to migrate such approach like Flink.

pan3793:
If we chose the pre-shade way, we need change `import` in codebase to reference the shaded versions.

turboFei:
thanks for the proposal, it sounds good.

yaooqinn:
Does pre-shade way has any adverse impact, such as users want another curator client?

pan3793:
> Does pre-shade way has any adverse impact, such as users want another curator client?

If user want to extend kyuubi original functions, i.e. custom the auth of curator, the should use kyuubi shaded curator api.

If user want to add individual functions to kyuubi, i.e. integration kyuubi to corp micro services discovery framework, and it happens to use curator, in such scenario normal curator will co-exist will kyuubi shaded curator.

yaooqinn:
SGTM

pan3793:
Apache Drill https://github.com/apache/drill pre-shade guava in a module, let's follow it.

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -98,6 +98,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.hive</groupId>
+            <artifactId>hive-service-rpc</artifactId>
+            <scope>test</scope>
+        </dependency>
```
turboFei, line 105: we need it here, because it could provide org.apache.thrift.classes for the test.

turboFei:
ci passed finally @pan3793

.travis.yml
```diff
@@ -47,6 +47,9 @@ matrix:
 
 install:
   - mvn --version
+  - cd kyuubi-shaded
+  - mvn clean install
+  - cd -
```
pan3793, line 52: Can it be replaced by `mvn clean install -pl <modules> -am -Pkyuubi-shaded`?

kyuubi-shaded/kyuubi-shaded-curator/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+    <parent>
+        <artifactId>kyuubi-shaded</artifactId>
+        <groupId>org.apache.kyuubi</groupId>
+        <version>1.2.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>kyuubi-shaded-curator</artifactId>
+    <name>Kyuubi Project Shaded curator</name>
```
yaooqinn, line 31: Kyuubi Project Shade Apache Curator

kyuubi-shaded/kyuubi-shaded-curator/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+                <artifactId>maven-shade-plugin</artifactId>
+                <configuration>
+                    <shadedArtifactAttached>false</shadedArtifactAttached>
+                    <artifactSet>
+                        <includes>
+                            <include>org.apache.curator:curator-client</include>
+                            <include>org.apache.curator:curator-framework</include>
+                            <include>org.apache.curator:curator-recipes</include>
+                            <include>org.apache.curator:curator-test</include>
+                            <include>org.javassist:javassist</include>
```
yaooqinn, line 112: is there any transitive we need?

kyuubi-shaded/kyuubi-shaded-curator/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+                        </includes>
+                    </artifactSet>
+                    <relocations>
+                        <relocation>
+                            <pattern>org.apache.curator</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.curator</shadedPattern>
```
yaooqinn, line 118: `${kyuubi.shade.packageName}.org.apache.curator` looks more future-proofing, as maybe there is another project called `curator` :)

kyuubi-shaded/kyuubi-shaded-curator/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+                            <includes>
+                                <include>org.apache.curator.**</include>
+                            </includes>
+                        </relocation>
+                        <relocation>
+                            <pattern>org.javassist</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.javassist</shadedPattern>
```
yaooqinn, line 125: ditto

kyuubi-shaded/kyuubi-shaded-hive-service-rpc/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+    <parent>
+        <artifactId>kyuubi-shaded</artifactId>
+        <groupId>org.apache.kyuubi</groupId>
+        <version>1.2.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>kyuubi-shaded-hive-service-rpc</artifactId>
+    <name>Kyuubi Project Shaded Hive Service Rpc</name>
```
yaooqinn, line 31: Kyuubi Project Shade Hive Service RPC

pan3793:
https://github.com/yaooqinn/kyuubi/pull/488/files#L1038

As we have shade module now, it's not suitable at parent `pom.xml`

.travis.yml
pan3793: `mvn clean install -pl :kyuubi-shaded-curator,:kyuubi-shaded-hive-service-rpc -am -Pkyuubi-shaded`

pom.xml
```diff
@@ -1440,5 +1376,11 @@
                 <spark.archive.download.skip>true</spark.archive.download.skip>
             </properties>
         </profile>
+        <profile>
+            <id>kyuubi-shaded</id>
+            <modules>
+                <module>kyuubi-shaded</module>
```
yaooqinn, line 1382: why it is a profile?

.github/workflows/master.yml
```diff
@@ -67,6 +67,8 @@ jobs:
           key: ${{ runner.os }}-maven-io-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-io-
+      - name: install kyuubi-shaded jars
+        run: cd kyuubi-shaded && mvn clean install && cd -
```
yaooqinn, line 71: can it not be installed in a single `mvn install`?

pom.xml
pan3793: After some investigation with @turboFei, we found that maven has some limitation of shade module reference, when run `mvn package`, the downstream module can't see shaded classes of upstream module which in same maven project.

There is a solution https://stackoverflow.com/a/50274253 but introduced another issue, we must run `mvn clean xxx` everytime, w/o `mvn clean` will cause compile failed.

Maybe that's why Apache Flink and Apache Hadoop choose shade deps in separated module, and Apache Iceberg which use Gradle doesn't suffer this issue.

yaooqinn: So this do related to the releasing process and a doc is also needed

yaooqinn: So this is related to the releasing process, and a documentation is also needed

kyuubi-shaded/kyuubi-shaded-curator/pom.xml
turboFei: we need javassist, otherwise exception would be thrown.

pom.xml
turboFei: is it ok to add some notes into building doc?

turboFei: https://github.com/yaooqinn/kyuubi/pull/488/files?file-filters%5B%5D=.md&file-filters%5B%5D=.scala&file-filters%5B%5D=.xml#diff-434e619c9343e148d9ef7e85b044633eb86f72c6f1137eac287a97708f18c86cR11

turboFei:
after remove
```
    <parent>
        <artifactId>kyuubi</artifactId>
        <version>1.2.0-SNAPSHOT</version>
        <relativePath>../pom.xml</relativePath>
    </parent>
```
in kyuubi-shaded module, there is no transaction dependencies.
https://github.com/yaooqinn/kyuubi/pull/488/commits/224cc38b45d5a02acd9fcc45d2902b60113f2abf

pan3793 (approved):
LGTM, thanks!

pan3793:
There are some things we need consider after switch to shaded.

- License declaration.
- Add rule to force use shaded version in Kyuubi codebase.

And we can do those in following PRs.

pan3793:
After this change, `kyuubi-spark-sql-engine` expect to work with spark build by without `-Phive-thriftserver`, right? @turboFei

yaooqinn:
Can we have a module like `test/e2e` to verify `./build/dist --tgz` still works in followups?

yaooqinn:
> After this change, `kyuubi-spark-sql-engine` is expected to work with spark build without `-Phive-thriftserver`, right? @turboFei

Hmm? Do you mean that we are going to drop support for official spark releases?

pan3793:
> > After this change, `kyuubi-spark-sql-engine` is expected to work with spark build without `-Phive-thriftserver`, right? @turboFei
> 
> Hmm? Do you mean that we are going to drop support for official spark releases?

Kyuubi always works with spark official release. And if we use custom spark dist, I want to minimize the deps.

turboFei:
> > After this change, `kyuubi-spark-sql-engine` is expected to work with spark build without `-Phive-thriftserver`, right? @turboFei
> 
> Hmm? Do you mean that we are going to drop support for official spark releases?

I think he means that we can support the spark release without hive-thritfserver after this change.

turboFei:
> Can we have a module like `test/e2e` to verify `./build/dist --tgz` still works in followups?

sure, will try

yaooqinn:
I see. I misunderstood. LGTM

yaooqinn (approved):

yaooqinn:
note that, we have an auto pipeline for GA release, we also need to verify the scripts and all the steps are still well

turboFei:
ok, I will update ./build/dist to add install shaded modules part.
and verify it in ci/GA

yaooqinn:

yaooqinn:
>mvn clean install -pl :kyuubi-shaded-curator,:kyuubi-shaded-hive-service-rpc -am -Pkyuubi-shaded

is this warning expected?

[WARNING] The requested profile "kyuubi-shaded" could not be activated because it does not exist.

pan3793:
> > mvn clean install -pl :kyuubi-shaded-curator,:kyuubi-shaded-hive-service-rpc -am -Pkyuubi-shaded
> 
> is this warning expected?
> 
> [WARNING] The requested profile "kyuubi-shaded" could not be activated because it does not exist.

Add to `kyuubi-shaded/pom.xml` will suppress this warning, but without it impact nothing.
```
   <profiles>
        <profile>
            <id>kyuubi-shaded</id>
        </profile>
    </profiles>
```

pan3793:
Although we have reached a consensus that pre-shade is a right direction, but such build step confuse developers new to Kyuubi, so I temporarily revert it and let us do pre-shade in separated repo or find some way more friendly for developers.

yaooqinn:
Let keep the old school way by shading curator to the engine module as a46f00f396c8191a3e45254b50fcf8979372c909, thanks @pan3793, and sorry for the back and forth @turboFei

PR #494, opened 2021-04-08
[MINOR] Fix the properties on spark-hadoop-3.2

Yikun:
Yikun [Closes #494](https://github.com/yaooqinn/kyuubi/pull/494)    Target Issue Test Plan Bug [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Change the `spark.hadoop.binary.version` to `hadoop.binary.version`

spark-hadoop-3.2 CI passed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1430,7 +1430,7 @@
         <profile>
             <id>spark-hadoop-3.2</id>
             <properties>
-                <spark.hadoop.binary.version>3.2</spark.hadoop.binary.version>
+                <hadoop.binary.version>3.2</hadoop.binary.version>
             </properties>
         </profile>
 


pan3793:
Good catch! Thanks for your contribute! 
And we have a workflow use `-Pspark-3.1 -Dhadoop.binary.version=3.2`, can you also change it to use maven profile?

yaooqinn (approved):

yaooqinn (approved):

Yikun:
@pan3793 We didn't use this spark-hadoop-3.2 profile in anywhere. So, I think we should fix it first, and then enable the
`-Pspark-hadoop-3.2` test, that means we use the [default spark 3.0.2](https://github.com/yaooqinn/kyuubi/blob/0a3862a9d119aef615788d6e6804fb7dc72a457a/pom.xml#L76) and then 3.2 hadoop

pan3793:
> We didn't use this `spark-hadoop-3.2` profile in anywhere. 

You are right, because we use `-Dhadoop.binary.version=3.2` instead, that why we didn't catch this issue.

> I think we should fix it first.

It's fine. Let limit this PR scope into only fix the profile definition, and replace `-Dhadoop.binary.version=3.2` by `-Pspark-hadoop-3.2` in following PR.

pan3793 (approved):

yaooqinn:

PR #495, opened 2021-04-08
[KYUUBI #491] Add Spark nightly build CI

Yikun:
Yikun [Closes #495](https://github.com/yaooqinn/kyuubi/pull/495)    Test Plan [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

This patch add Spark nightly build CI

github action CI passed

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -75,6 +75,7 @@ jobs:
           mvn clean install ${{ matrix.profiles }} -Dmaven.javadoc.skip=true -V
           bash <(curl -s https://codecov.io/bash)
       - name: Detected Dependency List Change
+        if: ${{ ! contains(matrix.profiles, 'spark-3.2-snapshot') }}
         run: build/dependency.sh
       - name: Upload unit tests log
         if: failure()
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -37,6 +37,7 @@ jobs:
           - '-Pspark-3.0 -Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.1.1 -Dspark.archive.name=spark-3.1.1-bin-hadoop2.7.tgz -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest'
           - '-Pspark-3.1'
           - '-Pspark-3.1 -Dhadoop.binary.version=3.2'
+          - '-Pspark-3.2-snapshot -pl :kyuubi-spark-sql-engine,:kyuubi-common,:kyuubi-ha,:kyuubi-zookeeper -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest'
     steps:
       - uses: actions/checkout@v2
       - name: Setup JDK 1.8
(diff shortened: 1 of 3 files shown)


.github/workflows/master.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           - '-Pspark-3.0 -Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.1.1 -Dspark.archive.name=spark-3.1.1-bin-hadoop2.7.tgz -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest'
           - '-Pspark-3.1'
           - '-Pspark-3.1 -Dhadoop.binary.version=3.2'
+          - '-Pspark-3.2-snapshot -Dspark.archive.download.skip=true'
```
yaooqinn, line 40: `-pl :kyuubi-spark-sql-engine -Pspark-3.2-snapshot ` might be enough to verify API consistency with Spark

yaooqinn:
Thanks for your first contribution and welcome! Just left some minor comments.

.github/workflows/master.yml
Yikun: ok, will fix in next PS.

Yikun:
@yaooqinn Thanks! And I noticed that we have two CI configuration: `github action` and `travis`, I'm not sure why we need all them.

Maybe we will remove the travis( or github action) config in future? So I'm not sure should I also add this CI in travis?

yaooqinn:
> Maybe we will remove the travis( or github action) config in future? 

I've been asked many times about this. Personally, I am not fully trusting the stability of free CI providers(with a grateful heart off course). As we don't want to block PRs because of CI flakiness, as long as one side passes, we can merge the PR.

As you can see the Github Action environment is not stable these days, I also see these errors in the Spark project too.

> So I'm not sure should I also add this CI in travis?

yes. As I am not 100% sure you can pass Github Action for now

Yikun:
Greate, and thanks for your explanation, make sense.

> As I am not 100% sure you can pass Github Action for now

now I'm 100% sure I can't pass the github action : ( raise the issue here: https://github.com/yaooqinn/kyuubi/issues/496

yaooqinn:
My bad, I forgot to let you know about this property `-Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest`

yaooqinn:
seems we do need `:kyuubi-common,:kyuubi-ha` too

Yikun:
> seems we do need `:kyuubi-common,:kyuubi-ha` too

and `kyuubi-zookeeper`, the `./build/mvn clean install -pl :kyuubi-spark-sql-engine,:kyuubi-common,:kyuubi-ha,:kyuubi-zookeeper -Pspark-3.2-snapshot -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest` had been passed in my loca env.

Yikun:
```
Using `mvn` from path: /usr/bin/mvn
awk: cmd. line:2: (FILENAME=- FNR=1) fatal: attempt to access field -1
Error: Process completed with exit code 1.
```
[1] https://github.com/yaooqinn/kyuubi/pull/495/checks?check_run_id=2303522554#step:10:45

Looks like some github action ci error happens in [here](https://github.com/yaooqinn/kyuubi/blob/354f55a0a5ec4d9c806bba5319505dfe0109815c/build/dependency.sh#L38), I'm not familar with it, could you take a look?

Note: the lastest code still passed in my locl env.

turboFei:
seems caused by https://github.com/yaooqinn/kyuubi/pull/488, will check it.

yaooqinn:
we can skip dependency check for this job with `if` statement  or set it like `Check License ` separately

yaooqinn (approved):

pan3793:
Thanks, merged to master for v1.2.0

PR #512, opened 2021-04-12
[KYUUBI #304]  Add kyuubi-ctl shell script

turboFei:
turboFei [Closes #512](https://github.com/yaooqinn/kyuubi/pull/512)    [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Add shell script `kyuubi-ctl` to launch kyuubi service control client.

Run test locally.
image

### DIFF
--- a/bin/kyuubi-ctl
+++ b/bin/kyuubi-ctl
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+## Kyuubi Service Control Client Entrance
+CLASS="org.apache.kyuubi.ctl.ServiceControlCli"
+
+export KYUUBI_HOME="$(cd "$(dirname "$0")"/..; pwd)"
+
+echo "Starting Kyuubi Service Control Client from ${KYUUBI_HOME}"
+
+. "${KYUUBI_HOME}/bin/load-kyuubi-env.sh"
+
+if [[ -z ${JAVA_HOME} ]]; then
+  echo "Error: JAVA_HOME IS NOT SET! CANNOT PROCEED."
+  exit 1
+fi
+
+RUNNER="${JAVA_HOME}/bin/java"
+
+## Find the Kyuubi Jar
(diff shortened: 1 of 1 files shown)


turboFei:
list:
image

turboFei:
get

turboFei:
create:
image

pan3793:
The script LGTM. One question, I saw you run `sh bin/kyuubi-ctl` in screenshot, have you set `x` permission on `bin/kyuubi-ctl`? The git will track the unix permissions of file.

I have a look on `build/dist`, we don't run `chmod +x bin/*` when make dist, and suppose the `dist` only running on UNIX-like OS, so the permission is retained.

turboFei:
delete:
image

turboFei:
Its permission is same with `bin/kyuubi`
```
-rwxr-xr-x  1 fwang12  110525255   1.6K Apr 12 14:15 bin/kyuubi-ctl
```

pan3793 (approved):
LGTM

turboFei:
Other question,
For the embedded zookeeper use case,
I have to set below two parameters.

```
kyuubi.zookeeper.embedded.data.dir /kyuubi
kyuubi.ha.zookeeper.namespace /kyuubi
```

yaooqinn:
> Will follow it and try to combine these two parameters.

I don't get it, what do you mean about `combine`?

yaooqinn:
Can you omit `sh` and test locally too?

turboFei:
> > Will follow it and try to combine these two parameters.
> 
> I don't get it, what do you mean about `combine`?


At first, I started kyuubi server with `kyuubi.zookeeper.embedded.data.dir=/kyuubi` and did not define the value of kyuubi.ha.zookeeper.namespace.


For create action, the namespace transferred is toNamespace, and the fromNamepace is loaded by the key of `kyuubi.ha.zookeeper.namespace`, which was not defined in kyuubi-defaults.conf.
So, we need some fallback for the default namespace.

yaooqinn:
image

This command usage hint looks confusing, we may improve it later. @pan3793 do you think we should use `common-cli` or something as simple as it to do the command parse?

turboFei:
It works well without `sh`.
image

yaooqinn:
> > > Will follow it and try to combine these two parameters.
> > 
> > 
> > I don't get it, what do you mean about `combine`?
> 
> At first, I started kyuubi server with `kyuubi.zookeeper.embedded.data.dir=/kyuubi` and did not define the value of kyuubi.ha.zookeeper.namespace.
> 
> For create action, the namespace transferred is toNamespace, and the fromNamepace is loaded by the key of `kyuubi.ha.zookeeper.namespace`, which was not defined in kyuubi-defaults.conf.
> So, we need some fallback for the default namespace.

image

image

yaooqinn:
the get/list looks exactly the same.

For followups,
1. I think the `get` should return the znode.toString, as something like `serverUrl......version=...seq...`
2. the result wrapping seems to make the command pipeline a bit hard? `list foreach delete` for `delete-all`?
3. I am not sure these commands are working well when ACL or SSL enabled,
4. Shall we also setACL/getACL actions?

yaooqinn (approved):

pan3793:
> image
> 
> This command usage hint looks confusing, we may improve it later. @pan3793 do you think we should use `common-cli` or something as simple as it to do the command parse?

+1 for leveraging library to build cli parser.

Issue #526, opened 2021-04-14
Support `kyuubi.operation.scheduler.pool` for queries


yaooqinn:
### 1. Describe the feature

http://spark.apache.org/docs/latest/job-scheduling.html#scheduling-within-an-application

### 2. Motivation

Better leverage spark capabilities, fine-grained queries scheduler within an app

PR #528, opened 2021-04-14
[KYUUBI #526] Support kyuubi.operation.scheduler.pool for queries

ulysses-you:
ulysses-you [Closes #528](https://github.com/yaooqinn/kyuubi/pull/528)    Test Plan Feature [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

Support scheduler pool which is native supported by Spark.

Pass CI
+
Manual tets

The default pool name is `default`, aflter execute `set kyuubi.operation.scheduler.pool = default1;`, the pool changed.

pool


ulysses-you:
cc @pan3793 @turboFei @yaooqinn

pan3793 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -533,4 +533,11 @@ object KyuubiConf {
     .version("1.2.0")
     .stringConf
     .createWithDefault("SHOW DATABASES")
+
+  val OPERATION_SCHEDULER_POOL: OptionalConfigEntry[String] = buildConf("operation.scheduler.pool")
+    .doc("The scheduler pool of job. Note that, this config should be used after change Spark " +
+      "config spark.scheduler.mode=FAIR.")
+    .version("1.1.1")
```
turboFei, line 540: should be 1.2.0?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -41,6 +41,10 @@ class ExecuteStatement(
   private val forceCancel =
     session.sessionManager.getConf.get(KyuubiConf.OPERATION_FORCE_CANCEL)
 
+  private val schedulerPool =
+    spark.conf.getOption(KyuubiConf.OPERATION_SCHEDULER_POOL.key).orElse(
+      session.sessionManager.getConf.get(KyuubiConf.OPERATION_SCHEDULER_POOL))
```
turboFei, line 46: should we fallback to spark.scheduler.pool?

turboFei: For ad-hoc use case,  I think kyuubi should be transparent for customer.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
ulysses-you: Decide to make 1.1.1 support this minor feature for our requirement : )

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
ulysses-you: `spark.scheduler.pool` just a local property that user can not set it even using `SparkThriftServer` so we don not need to respect it at config level.

turboFei: make sense

turboFei (approved):

ulysses-you:
thanks @pan3793 @turboFei  merging to master for v1.2.0. Seems here some conflicts with branch-1.1, let me do a backport in another PR.

yaooqinn:
can have uts? it's not that hard to test this

PR #547, opened 2021-04-20
[BUILD] Add build timestamp and support scala.binary.version specific code

pan3793:

This change added `module.build.timestamp` and `local.build.timestamp` to maven properties, so we can use it in later for make dist.

This change also add `src/main/scala-${scala.binary.version}` to maven sources files, so we can write `scala.binary.version` specific code to the codebase.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,7 @@
         <jars.target.dir>${project.build.directory}/scala-${scala.binary.version}/jars</jars.target.dir>
 
         <maven.plugin.antrun.version>3.0.0</maven.plugin.antrun.version>
+        <maven.plugin.build.helper>3.2.0</maven.plugin.build.helper>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -89,6 +89,7 @@
         <jars.target.dir>${project.build.directory}/scala-${scala.binary.version}/jars</jars.target.dir>
 
         <maven.plugin.antrun.version>3.0.0</maven.plugin.antrun.version>
+        <maven.plugin.build.helper>3.2.0</maven.plugin.build.helper>
```
yaooqinn, line 92: maven.plugin.build.helper.version

pan3793: oops, fixed.

yaooqinn (approved):

turboFei (approved):

turboFei:
thanks, merged to master

PR #555, opened 2021-04-20
[INFRA] Backport Using from Scala 2.13

### LINKED ISSUE #553: Backport Using from Scala 2.13
### 1. Describe the feature
Back port Using from Scala 2.13 to simplify `try with resources` code.
https://github.com/scala/scala/blob/v2.13.5/src/library/scala/util/Using.scala

### 2. Motivation
Scala doesn't support Java style `try with resource` in syntax. But Scala provided `Using` since 2.13 to simulate such behavior.

### 3. Describe the solution
Back port Using from Scala 2.13 into Kyuubi codebase.

pan3793:

close #553

### DIFF
--- a/kyuubi-common/src/main/scala-2.12/scala/util/Using.scala
+++ b/kyuubi-common/src/main/scala-2.12/scala/util/Using.scala
@@ -108,7 +108,8 @@ import scala.util.control.{ControlThrowable, NonFatal}
  *   - `java.lang.VirtualMachineError`
  *   - `java.lang.LinkageError`
  *   - `java.lang.InterruptedException` and `java.lang.ThreadDeath`
- *   - [[scala.util.control.NonFatal fatal exceptions]], excluding `scala.util.control.ControlThrowable`
+ *   - [[scala.util.control.NonFatal fatal exceptions]],
+ *   excluding `scala.util.control.ControlThrowable`
  *   - `scala.util.control.ControlThrowable`
  *   - all other exceptions
  *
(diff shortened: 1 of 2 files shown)


turboFei (approved):

kyuubi-common/src/main/scala-2.12/scala/util/Using.scala
```diff
@@ -0,0 +1,415 @@
...
+   * @param resource2 the second resource
+   * @param body      the operation to perform using the resources
+   * @tparam R1 the type of the first resource
+   * @tparam R2 the type of the second resource
+   * @tparam A  the return type of the operation
+   * @return the result of the operation, if neither the operation nor
+   *         releasing the resources throws
+   */
+  def resources[R1: Releasable, R2: Releasable, A](
+                                                    resource1: R1,
```
yaooqinn, line 314: Is scala source so ugly looking?

yaooqinn:
can we refine one place in the main code in case of dead code?

pan3793:
> can we refine one place in the main code in case of dead code?

Will refactor `try ... finally` by `Using.resource() { ... } ` in following PR.

kyuubi-common/src/main/scala-2.12/scala/util/Using.scala
pan3793: updated codesytle. the original source code at https://raw.githubusercontent.com/scala/scala/v2.13.5/src/library/scala/util/Using.scala

yaooqinn (approved):

turboFei (approved):

turboFei:
thanks, merging to master

PR #559, opened 2021-04-21
[KYUUBI #551] TEST JDK 11 in travis

jackMintao:

test againest jdk11 in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ language: scala
 scala:
   - 2.12.10
 jdk:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,7 +20,7 @@ language: scala
 scala:
   - 2.12.10
 jdk:
-  - openjdk8
+  - openjdk11
```
pan3793, line 23: We should test against both jdk8 and jdk11 :)

Issue #508, opened 2021-04-11
Enable `env:` type configuration in JDBC connection url to switch env settings during engine bootstrap


yaooqinn:
### 1. Describe the feature

Now we set hiveconf/sparkconf/systemconf etc to `sessionConf`, then `env:` prefixed shall go engines environments

pan3793:
Let me take this task.

pan3793:
Have a quick look on SparkThriftServer and HiveServer2 implementation, those two doesn't support set env neither, and beeline and JDBC url doesn't expose the env setting place, the only function of env vars is for variable substitution. In other words, env vars are read only.

To match the HiveServer2 behavior, we need to add the check on `set:env:xxxx` configurations on creating session. And seems `set env:TERM=abc` should handle by Spark itself?

PR #561, opened 2021-04-21
Forbid set env:* in thrift connection configurations

pan3793:

This change only forbid the `set:env:*` from thrift connection configurations, to match hive behavior, forbid `set env:*` in SQL, we need to change it in Spark.
#508
https://issues.apache.org/jira/browse/HIVE-6175


yaooqinn:

PR #582, opened 2021-04-24
[TEST][TPCDS] Introduce schema check for q15

turboFei:

Introduce schema check for q15


turboFei:
it failed for some tpcds queries.

Skip them at first.

PR #591, opened 2021-04-24
[KYUUBI #550] Test jdk11 in GA

turboFei:

Test jdk11 in GA


kyuubi-common/src/test/scala/org/apache/kyuubi/KyuubiSQLExceptionSuite.scala
```diff
@@ -35,7 +35,6 @@ class KyuubiSQLExceptionSuite extends KyuubiFunSuite {
 
     val e3 = KyuubiSQLException(e2.toTStatus)
     assert(e3.getMessage === e2.getMessage)
-    assert(e3.getStackTrace === e2.getStackTrace)
```
turboFei, line 38: image
In jdk11, the stack trace is not absolute same.

turboFei:
ok,

PR #601, opened 2021-04-24
[METRICS] Change metrics config keys

pan3793:

Change the config keys of metrics to get a consistent name style, `metrics.{reporter_type}.{property}`.

The change breaks the compatibility with master, but consider v1.2.0 is not out yet, so changes should be safe. 


yaooqinn (approved):

kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConf.scala
```diff
@@ -34,26 +34,39 @@ object MetricsConf {
...
+      " <li>JMX - JmxReporter which listens for new metrics and exposes them as MBeans.</li> " +
+      " <li>JSON - JsonReporter which outputs measurements to json file periodically.</li>" +
+      " <li>SLF4J - Slf4jReporter which outputs measurements to system log periodically.</li>" +
+      "</ul>")
     .version("1.2.0")
     .stringConf
     .transform(_.toUpperCase())
     .toSequence
-    .createWithDefault(Seq(JSON.toString))
+    .createWithDefault(Seq(SLF4J.toString))
```
yaooqinn, line 48: this is a bit dangerous change. the server log might increase exponentially

pan3793: Got it, will revert it.

.gitignore
```diff
@@ -42,8 +42,8 @@ derby.log
 ldap
 */dependency-reduced-pom.xml
 */*/dependency-reduced-pom.xml
-metrics/report.json
-metrics/.report.json.crc
+report.json
```
yaooqinn, line 45: please check the scala-test plugin settings in pom.xml, you should be able to skip these changes

yaooqinn (approved):

PR #602, opened 2021-04-24
[TPCDS][TEST] Add query output schema check for TPCDS 2.7

turboFei:

Add query output schema check for TPCDS 2.7


ulysses-you:
Seems we need to make `kyuubi-spark-sql-engine` test `log4j.properties` loglevel to INFO.

Current log length is to large.
```
The job exceeded the maximum log length, and has been terminated.
```

turboFei:
thanks, have adjusted log level

turboFei:
Now GA is slow, close it temporarily.

yaooqinn:
we can tag this test and test it in a separate GA job or even workflow

turboFei:
yes, I also considered using g a tag to include tpcds tests only. will add tpcds2.7 test after it is ready.

PR #604, opened 2021-04-25
[KYUUBI #552] Sync build jobs between GA and travis

turboFei:

Sync build jobs between GA and travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,10 +28,14 @@ cache:
 
 matrix:
   include:
-    - name: Test Kyuubi w/ -Pspark-3.0
+    - name: Test Kyuubi w/ Default Profile
       env:
-        - PROFILE="-Pspark-3.0"
+        - PROFILE=""
         - EXCLUDE_TAGS=""
+    - name: Test Kyuubi w/ -Pspark-3.0 and Spark 3.1 binary
+      env:
+        - PROFILE="-Pspark-3.0 -Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.1.1 -Dspark.archive.name=spark-3.1.1-bin-hadoop2.7.tgz"
+        - EXCLUDE_TAGS="org.apache.kyuubi.tags.DataLakeTest"
     - name: Test Kyuubi w/ -Pspark-3.1
       env:
         - PROFILE="-Pspark-3.1"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,20 +30,16 @@ matrix:
...
     - name: Test Kyuubi w/ -Pspark-3.0
       env:
-        - PROFILE="-Pspark-3.0"
-        - EXCLUDE_TAGS=""
+        - PROFILE="-Pspark-3.0 -Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.1.1 -Dspark.archive.name=spark-3.1.1-bin-hadoop2.7.tgz"
+        - EXCLUDE_TAGS="org.apache.kyuubi.tags.DataLakeTest"
     - name: Test Kyuubi w/ -Pspark-3.1
       env:
         - PROFILE="-Pspark-3.1"
-        - EXCLUDE_TAGS="org.apache.kyuubi.tags.DataLakeTest"
```
turboFei, line 38: it is excluded by default, and keep it sync with GA

yaooqinn: iceberg does not support spark 3.1 yet

turboFei: I see. We need specify it here, otherwise, it will be overwritten by an empty string.

.travis.yml
```diff
@@ -30,8 +30,8 @@ matrix:
   include:
```
pan3793, line 30: GA also test against `PROFILE=''`

.travis.yml
```diff
@@ -30,8 +30,8 @@ matrix:
...
     - name: Test Kyuubi w/ -Pspark-3.0
```
pan3793, line 31: Test Kyuubi w/ -Pspark-3.0 and Spark 3.1 binary

.travis.yml
turboFei: so,  add tags into profile instead of EXECLUDE_TAGS?

yaooqinn: PROFILE='' == PROFILE="-Pspark-3.0", this pr seems to delete the default job?

turboFei: ok, got it

.travis.yml
```diff
@@ -28,10 +28,14 @@ cache:
 
 matrix:
   include:
-    - name: Test Kyuubi w/ -Pspark-3.0
+    - name: Test Kyuubi w/ Default Profile
```
turboFei, line 31: rename it to default profile to keep consistent with GA

pan3793 (approved):
LGTM

turboFei:
The travis ci job was failed because its execution time was too long.

turboFei:
merged to master, and will create a pr to split the build job.

PR #606, opened 2021-04-25
Introduce ExtendedSQLTest tag to split build jobs

turboFei:

Involve ExtendedSQLTest tag to split build jobs

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -34,10 +34,11 @@ jobs:
       matrix:
         profiles:
           - ''
-          - '-Pspark-3.0 -Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.1.1 -Dspark.archive.name=spark-3.1.1-bin-hadoop2.7.tgz -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest'
+          - '-Pspark-3.0 -Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.1.1 -Dspark.archive.name=spark-3.1.1-bin-hadoop2.7.tgz -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.ExtendedSQLTest,org.apache.kyuubi.tags.DataLakeTest'
           - '-Pspark-3.1'
           - '-Pspark-3.1 -Pspark-hadoop-3.2'
-          - '-Pspark-3.2-snapshot -pl :kyuubi-spark-sql-engine,:kyuubi-common,:kyuubi-ha,:kyuubi-zookeeper -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest'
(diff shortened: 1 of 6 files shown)


turboFei:
maybe we can remove some jobs.

For ExtendedSQLTest, just keep:
- spark-3.0
- spark-3.1
- spark-3.2.0-SNAPSHOT

.github/workflows/master.yml
```diff
@@ -33,11 +33,14 @@ jobs:
     strategy:
       matrix:
         profiles:
-          - ''
-          - '-Pspark-3.0 -Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.1.1 -Dspark.archive.name=spark-3.1.1-bin-hadoop2.7.tgz -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest'
-          - '-Pspark-3.1'
-          - '-Pspark-3.1 -Pspark-hadoop-3.2'
-          - '-Pspark-3.2-snapshot -pl :kyuubi-spark-sql-engine,:kyuubi-common,:kyuubi-ha,:kyuubi-zookeeper -Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.DataLakeTest'
+          - '-Dmaven.plugin.scalatest.exclude.tags=org.apache.kyuubi.tags.ExtendedSQLTest'
```
yaooqinn, line 36: How about making it default in pom?

yaooqinn:
maybe enabling 3.1 and 3.2 is enough?

ulysses-you:
Can we move these tag to `pom.xml` is possible ? not only for GA but also travis.

turboFei:
> Can we move these tag to `pom.xml` if possible ? not only for GA but also travis.

move  EXCLUDE_TAGS/INCLUDE_TAGS into PROFILE?

cc @yaooqinn

yaooqinn:
I am ok with the current setting, but let's remove some duplicated workflows

kyuubi-main/src/test/scala/org/apache/kyuubi/operation/datalake/IcebergOperationSuite.scala
```diff
@@ -19,8 +19,9 @@ package org.apache.kyuubi.operation.datalake
 
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.operation.{BasicIcebergJDBCTests, WithKyuubiServer}
-import org.apache.kyuubi.tags.DataLakeTest
+import org.apache.kyuubi.tags.{DataLakeTest, ExtendedSQLTest}
 
+@ExtendedSQLTest
```
yaooqinn, line 24: we can remove this tag to make the latter usage simple?

yaooqinn:
Let's mark the test suites with one tag only to make the caller side simpler

For now, `ExtendedSQLTest` should be marked to TPCDS tests, and defined as `exclude` for all workflows in pom directly. So I guess the current workflow stays the same as it is now.

Add one or two separate jobs to test the TPCDS test cases only, SHOULD NOT use the tag `ExtendedSQLTest` here

yaooqinn:
THAT said, the newly added jobs can simply do something like `build/mvn -Dtest=none -DwildcardSuites=the test name test`

kyuubi-main/src/test/scala/org/apache/kyuubi/operation/datalake/IcebergOperationSuite.scala
turboFei: ok

yaooqinn (approved):

yaooqinn:

PR #613, opened 2021-04-28
[TEST] [DELTA] Add test for deltalake

pan3793:

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/SparkCatalogShim.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/SparkCatalogShim.scala
@@ -123,7 +123,7 @@ trait SparkCatalogShim extends Logging {
       methodName: String,
       args: (Class[_], AnyRef)*): Any = {
     val (types, values) = args.unzip
-    val method = obj.getClass.getDeclaredMethod(methodName, types: _*)
+    val method = obj.getClass.getMethod(methodName, types: _*)
     method.setAccessible(true)
     method.invoke(obj, values.toSeq: _*)
   }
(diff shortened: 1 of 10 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/SparkCatalogShim.scala
```diff
@@ -123,7 +123,7 @@ trait SparkCatalogShim extends Logging {
       methodName: String,
       args: (Class[_], AnyRef)*): Any = {
     val (types, values) = args.unzip
-    val method = obj.getClass.getDeclaredMethod(methodName, types: _*)
+    val method = obj.getClass.getMethod(methodName, types: _*)
```
pan3793, line 126: `getDeclaredMethods` includes all methods declared by the class itself, whereas `getMethods` returns only public methods, but also those inherited from a base class.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkDeltaOperationSuite.scala
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.spark.operation
+
+import org.apache.kyuubi.engine.spark.WithSparkSQLEngine
+import org.apache.kyuubi.operation.BasicDeltaJDBCTests
+import org.apache.kyuubi.tags.DataLakeTest
+
+@DataLakeTest
```
pan3793, line 24: Delta 0.8 doesn't support Spark 3.1.1 yet. https://github.com/delta-io/delta/issues/630

yaooqinn: Hmm, so slow...

yaooqinn (approved):

PR #616, opened 2021-04-28
[TEST] [TPCDS] [DELTA] Migrate TPCDS test to Delta

pan3793:

Finally, delta 1.0.0 is out, with Spark 3.1.1 support.

- In Spark 3.0, column of CHAR type is not allowed in non-Hive-Serde tables, and CREATE/ALTER TABLE commands will fail if CHAR type is detected. Please use STRING type instead. In Spark version 2.4 and below, CHAR type is treated as STRING type and the length parameter is simply ignored.

- Since Spark 3.1, CHAR/CHARACTER and VARCHAR types are supported in the table schema. Table scan/insertion will respect the char/varchar semantic. If char/varchar is used in places other than table schema, an exception will be thrown (CAST is an exception that simply treats char/varchar as string like before). To restore the behavior before Spark 3.1, which treats them as STRING types and ignores a length parameter, e.g. `CHAR(4)`, you can set `spark.sql.legacy.charVarcharAsString` to `true`.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -123,5 +123,5 @@ jobs:
             ${{ runner.os }}-maven-io-
       - name: Run TPC-DS Tests
         run: |
-          mvn clean install --no-transfer-progress -Pspark-3.0 -DskipTests -pl :kyuubi-spark-sql-engine,:kyuubi-common,:kyuubi-ha,:kyuubi-zookeeper
-          mvn test --no-transfer-progress -Pspark-3.0 -Dtest=none -DwildcardSuites=*TPCDSSuite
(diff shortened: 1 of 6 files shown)


pan3793:
@yaooqinn Besides switching TPCDS suites to datalake format, what do you think the refactor of `*SuiteMixin`? Maybe we can get the refactor in at first but do not touch TPCDS suites now.

yaooqinn:
Why we need `DeltaSuiteMixin` and `BasicDeltaJDBCTests` both？

pan3793:
`BasicDeltaJDBCTests` define the Delta specific test cases, but `DeltaSuiteMixin` define the spark relative configs, for `TPCDSSuite`, only `DeltaSuiteMixin` is needed but not `BasicDeltaJDBCTests`

yaooqinn:
please send a separate PR

pan3793:
see #619

yaooqinn (approved):

PR #618, opened 2021-04-30
Add lock while creating engine with non-CONNECTION mode

Honglun:

We may create multiple SparkContext in non-CONNECTION ShareLevel.


kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
```diff
@@ -99,4 +99,11 @@ object HighAvailabilityConf {
     .timeConf
     .checkValue(_ > 0, "Must be positive")
     .createWithDefault(Duration.ofSeconds(120).toMillis)
+
+  val HA_ZK_ENGINE_LOCK_TIMEOUT: ConfigEntry[Long] =
+    buildConf("ha.zookeeper.engine.lock.timeout")
+      .doc("The timeout(s) of a lock when a engine created")
+      .version("1.0.0")
+      .timeConf
+      .createWithDefault(Duration.ofSeconds(30).toMillis)
```
ulysses-you, line 108: The default timeout number I think should be same with `session.engine.initialize.timeout` which is the max timeout before fail.

ulysses-you (commented):
@Honglun thank you for working on this. I left some comments.

It is a real issue if the same user open a new session more than once at the same time using `USER` mode. cc @yaooqinn

Honglun:
@ulysses-you Thanks, I will update the code latter.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -342,6 +343,31 @@ object ServiceDiscovery extends Logging {
     }
     serviceNode
   }
+
+  def withLock(
+      zkClient: CuratorFramework,
+      lockPath: String,
+      lockTimeOutSeconds: Long)(f: => Unit): Unit = {
```
yaooqinn, line 350: it's milliseconds

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -342,6 +343,31 @@ object ServiceDiscovery extends Logging {
...
+    var lock: InterProcessSemaphoreMutex = null
+    try {
+      try {
+        lock = new InterProcessSemaphoreMutex(zkClient, ZKPaths.makePath(lockPath, "lock"))
+        lock.acquire(lockTimeOutSeconds, TimeUnit.SECONDS)
+      } catch {
+        case e: Exception => throw new KyuubiException(s"Lock failed on path [$lockPath]", e)
```
yaooqinn, line 357: KyuubiSQLException

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -342,6 +343,31 @@ object ServiceDiscovery extends Logging {
     }
     serviceNode
   }
+
+  def withLock(
+      zkClient: CuratorFramework,
+      lockPath: String,
+      lockTimeOutSeconds: Long)(f: => Unit): Unit = {
+    var lock: InterProcessSemaphoreMutex = null
+    try {
```
yaooqinn, line 352: Can we simply use `scala.util.Using`? @pan3793

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/SQLEngineAppName.scala
```diff
@@ -32,6 +32,12 @@ case class SQLEngineAppName(
       case _ => ZKPaths.makePath(s"${prefix}_$sharedLevel", user)
     }
   }
+
+  def getZkLockPath(prefix: String): String = {
+    sharedLevel match {
```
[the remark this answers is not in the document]
yaooqinn: remove this `match-case`

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
```diff
@@ -99,4 +99,11 @@ object HighAvailabilityConf {
     .timeConf
     .checkValue(_ > 0, "Must be positive")
     .createWithDefault(Duration.ofSeconds(120).toMillis)
+
+  val HA_ZK_ENGINE_LOCK_TIMEOUT: ConfigEntry[Long] =
+    buildConf("ha.zookeeper.engine.lock.timeout")
+      .doc("The timeout(s) of a lock when a engine created")
```
pan3793, line 105: **when a engine created** => **when creating an engine**

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
pan3793: yeap

Honglun: `scala.util.Using` was introduced in scala-2.13 ?

pan3793: we back ported it

Honglun: Can we use this instead ?
  `def using[A <: {def close(): Unit}](r: => A)(f: A => Unit): Unit = {`
    `val resource = r`
    `try { f(resource) } finally { if (resource != null) r.close() }`
  `}`

Honglun: `InterProcessSemaphoreMutex` is not a AutoCloseable, so  we need to add this in object `Releasable` ? 
`type AutoReleasableIsReleasable = {def release() : Unit}`
`implicit object AutoReleasableIsReleasable extends Releasable[{def release() : Unit}] {`
      `def release(resource: AutoReleasableIsReleasable): Unit = resource.release()`
`    }`
Or is there a better way? @pan3793

pan3793: Oops, I didn't realize that `InterProcessSemaphoreMutex` is not a `AutoCloseable`, the extension is feasible, but don't do it(at least don't modify in `Using.scala` file), keep the `Using` compatible with future Scala versions.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -342,6 +343,31 @@ object ServiceDiscovery extends Logging {
     }
     serviceNode
   }
+
+  def withLock(
+      zkClient: CuratorFramework,
+      lockPath: String,
+      lockTimeOutMilliseconds: Long)(f: => Unit): Unit = {
```
pan3793, line 350: Rename `lockTimeOutMilliseconds` => `lockTimeout`.
Personally, I agree with that any time variable should indicate the unit of values, but kyuubi already choose another style, all time should represent by `Long` in milliseconds, let's keep same style.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
pan3793: Just leave it as it is.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
Honglun: Update

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/SQLEngineAppName.scala
```diff
@@ -32,6 +32,13 @@ case class SQLEngineAppName(
       case _ => ZKPaths.makePath(s"${prefix}_$sharedLevel", user)
     }
   }
+
+  def getZkLockPath(prefix: String): String = {
+    assert(sharedLevel != CONNECTION)
+    sharedLevel match {
+      case _ => ZKPaths.makePath(s"${prefix}_$sharedLevel", "lock", user)
```
yaooqinn, line 39: `/kyuubi/kyuubi_USER/lock/userName/lock/lockes`, is the first  above `lock` neccessary?

Honglun: Yes. `GetServiceNodesInfo` cannot touch the `lock` node.
https://github.com/NetEase/kyuubi/blob/master/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala#L266

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -342,6 +343,31 @@ object ServiceDiscovery extends Logging {
...
+        case e: Exception => throw KyuubiSQLException(s"Lock failed on path [$lockPath]", e)
+      }
+      f
+    } finally {
+      try {
+        if (lock != null) {
+          lock.release()
+        }
+      } catch {
+        case _: Exception =>
```
yaooqinn, line 366: What if we fail to release the lock？So it locks forever？

Honglun: Yes, unless we restart `Kyuubi` Server. In this case, the new process acquires the lock after waiting for `HA_ZK_ENGINE_LOCK_TIMEOUT`.

yaooqinn: I see. Seems we can use `ENGINE_INIT_TIMEOUT` for the lock timeout?

Honglun: Yes, the default value of `HA_ZK_ENGINE_LOCK_TIMEOUT` is the same as `ENGINE_INIT_TIMEOUT`. Using 'ENGINE_INIT_TIMEOUT' should work. Do we need to modify this?

yaooqinn: yes, let's use `ENGINE_INIT_TIMEOUT ` to make it simple

Honglun: Update

yaooqinn: thanks

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/SQLEngineAppName.scala
yaooqinn, line 39: `ZKPaths.makePath(s"${prefix}_$sharedLevel", "lock", user) `？

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
```diff
@@ -178,6 +178,7 @@ class KyuubiOperationManager private (name: String) extends OperationManager(nam
         if (remoteHandle != null) {
           val or = FetchOrientation.toTFetchOrientation(order)
           val req = new TFetchResultsReq(remoteHandle, or, maxRows)
+          req.setFetchType(1.toShort)
```
yaooqinn, line 181: Is this related? can we rebase master branch?

yaooqinn (approved):
LGTM only minor comments

yaooqinn:
Hi @Honglun  can you address the last few comments? I'd like to merge this

Honglun (commented):
Hi @yaooqinn Any other questions about this?

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
yaooqinn: I mean you need rebase master branch

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/SQLEngineAppName.scala
yaooqinn: Hi @Honglun didn't you miss this comment? why do we need to match with _ here?

Honglun: I see, I will update this.

kyuubi-main/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
Honglun: @yaooqinn help, can we have a chat in Gitter?

Honglun:
Create a new PR

PR #619, opened 2021-05-06
[TEST] Refactor data lake test suite hierarchical

pan3793:

Make `*SuiteMixin` resuable.

### DIFF
--- a/kyuubi-common/src/test/scala/org/apache/kyuubi/DataLakeSuiteMixin.scala
+++ b/kyuubi-common/src/test/scala/org/apache/kyuubi/DataLakeSuiteMixin.scala
@@ -19,6 +19,8 @@ package org.apache.kyuubi
 
 import java.nio.file.Path
 
+import org.apache.kyuubi.config.KyuubiConf
+
 trait DataLakeSuiteMixin {
 
   protected val format: String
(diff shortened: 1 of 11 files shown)


yaooqinn:
Can we also add `DataLakeMixin` to gather duplicates around?

pan3793:
Updated

kyuubi-common/src/test/scala/org/apache/kyuubi/DataLakeSuiteMixin.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  protected val catalog: String
+
+  protected val warehouse: Path
+
+  protected val extraJars: String
+
+  protected val extraConfigs: Map[String, String]
+
+  protected val conf: KyuubiConf = {
```
pan3793, line 36: Not sure if I should put `conf` here

kyuubi-common/src/test/scala/org/apache/kyuubi/DataLakeSuiteMixin.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  protected val format: String
+
+  protected val catalog: String
+
+  protected val warehouse: Path
+
+  protected val extraJars: String
+
+  protected val extraConfigs: Map[String, String]
```
yaooqinn, line 32: val -> def

yaooqinn (approved):

pan3793:
Thanks for reviewing, merged to master for v1.2.0.

Issue #625, opened 2021-05-10
Kyuubi server-side does not support "operation.idle. timeout"


Honglun:
### 1. Describe the bug

"operation.idle. timeout" does not work in Kyuubi's configuration. Before Kyuubi-1.x it works. Sometimes we want to control "operation.idle. timeout" on the server-side. Can we support this?

#### 2.1 Versions

- Kyuubi
  - [ ] 1.1.0
  - [ ] 1.0.3
  - [ ] 0.8.x and earlier 

- Spark
  - [ ] 3.1.x
  - [ ] 3.0.x
  - [ ] 2.4.x and earlier 

#### 4.2 Log of the corresponding engine

yaooqinn:
can you detail this？Why does `kyuubi.operation.idle.timeout` not work

Honglun:
@yaooqinn We only get `operation.idle.timeout` from user request.
https://github.com/NetEase/kyuubi/blob/master/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala#L38

yaooqinn:
```
  private final val operationTimeout: Long = {
    session.sessionManager.getConf.get(OPERATION_IDLE_TIMEOUT)
  }
```

Can this change work?

Honglun:
We should use user configuration first.
```
private final val operationTimeout: Long = session.conf.get(OPERATION_IDLE_TIMEOUT.key)
    .map(_.toLong).getOrElse(session.sessionManager.getConf.get(OPERATION_IDLE_TIMEOUT))
```

yaooqinn:
Make sense to me. But also we should check this client-side setting is not greater than the server-side limit

yaooqinn:
> We should use user configuration first.
> 
> ```
> private final val operationTimeout: Long = session.conf.get(OPERATION_IDLE_TIMEOUT.key)
>     .map(_.toLong).getOrElse(session.sessionManager.getConf.get(OPERATION_IDLE_TIMEOUT))
> ```



> Make sense to me. But also we should check this client-side setting is not greater than the server-side limit.

On second thought, is the client-side more likely to use query timeout? The operation timeout is used by the server-side to clean errored/closed/canceled/finished/timeouted operations? These operations will have nothing to do with the client-side. So `session.sessionManager.getConf.get(OPERATION_IDLE_TIMEOUT)` shall be enough

Honglun:
[...]

PR #626, opened 2021-05-10
[KYUUBI #625]Kyuubi server-side support 'operation.idle. timeout'

Honglun:

Now we only get 'operation.idle.timeout' from user request. Kyuubi server-side should support "operation.idle. timeout".


yaooqinn (approved):

yaooqinn:

PR #628, opened 2021-05-10
solve the folder naming problem

zhang1002:
For starting kyuubiServer, the script need to get 'SPARK_HOME' from the file: RELEASE. The rule is: spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}. But when we package project from source, the script named dist will create the directory about 'SPARK_HOME' by the rule: spark-${SPARK_VERSION}-bin-hadoop${SPARK_HADOOP_VERSION}. So it's the differences between two rules. This change let ${HADOOP_VERSION} as same as ${SPARK_HADOOP_VERSION}.


pan3793:
Thanks for your first contribution and welcome! It's definitely a bug, and the change make sense to me, but Kyuubi Server build against on fixed version (3.2.2) of `hadoop-client-api` and `hadoop-client-runtime`, with this change should we also reflect this in `RELEASE`? cc @yaooqinn

pan3793 (approved):
LGTM, thanks!

pan3793:
Thanks, merged to master for v1.2.0

PR #632, opened 2021-05-13
solve the timeout problem when stop the kyuubiServer

zhang1002:
When you stop the KyuubiServiceDiscovery, you need to close node in zk and delete localNodePath at first. But when EmbeddedZookeeper is closed before KyuubiServiceDiscovery, the operation that stop kyuubiServer will timeout. 
Now use 'org.apache.hadoop.util.ShutdownHookManager' to close each service by different priority.


kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -165,4 +166,15 @@ private[kyuubi] object Utils extends Logging {
...
+
+  /**
+    * Add some operations that you want into ShutdownHook
+    * @param hook
+    * @param priority: 0~100
+    */
+  def addHook(hook: Runnable, priority: Int): Unit = {
+    ShutdownHookManager.get().addShutdownHook(
+      hook, priority
+    )
```
yaooqinn, line 178: inline these

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -123,8 +123,7 @@ abstract class ServiceDiscovery private (
   }
 
   // close the EPHEMERAL_SEQUENTIAL node in zk
-  protected def closeServiceNode(): Unit = {
-    if (_serviceNode != null) {
+  protected def closeServiceNode(): Unit = {if (_serviceNode != null) {
```
yaooqinn, line 126: unnecessary change

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -165,4 +166,15 @@ private[kyuubi] object Utils extends Logging {
   def isTesting: Boolean = {
     System.getProperty(IS_TESTING.key) != null
   }
+
+  /**
+    * Add some operations that you want into ShutdownHook
+    * @param hook
+    * @param priority: 0~100
+    */
+  def addHook(hook: Runnable, priority: Int): Unit = {
```
yaooqinn, line 175: rename it to addShutdownHook to be more specfic

yaooqinn (approved):

yaooqinn:

PR #635, opened 2021-05-13
Fix stage level config miss ReusedExchange

ulysses-you:

Avoid the reused exchange cause the wrong stage config.

Add test.


ulysses-you:
cc @yaooqinn

yaooqinn (approved):

ulysses-you:
Thanks, GA passed. Merged to master/v1.2.0.

PR #637, opened 2021-05-13
[METRICS] Remove unnecessary Guava Cache

pan3793:

`MetricRegistry` hold counters in `ConcurrentHashMap`, it's unnecessary to use Guava Cache.

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
@@ -32,16 +32,14 @@ class MetricsSystem extends CompositeService("MetricsSystem") {
 
   private val registry = new MetricRegistry
 
-  def incAndGetCount(key: String): Long = synchronized {
+  def incCount(key: String): Unit = {
     val counter = registry.counter(key)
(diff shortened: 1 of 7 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
```diff
@@ -33,29 +31,17 @@ import org.apache.kyuubi.service.CompositeService
...
-      counter.inc(1L)
-      counter.getCount
-    } finally {
-      countersLock.unlock()
-    }
+
+  def incAndGetCount(key: String): Long = synchronized {
+    val counter = registry.counter(key)
+    counter.inc(1L)
+    counter.getCount
```
pan3793, line 38: The return value never used, if return `Unit`, we can remove `synchronized`

yaooqinn: sgtm

ulysses-you (approved):

yaooqinn (approved):

yaooqinn:
thanks, merged to master for v1.3.0 and branch 1.2 for v1.2.0

PR #646, opened 2021-05-26
Capture exceptions in async ExecuteStatement

pan3793:

When ExecuteStatement running in async, if engine broken(e.g. killed by yarn), the thread just crash without `setOperationException(ke)`, then client(like `beeline`) will hang.

Manual tested on our yarn cluster: kill yarn app when executing statement.
1. `beeline` connect hiveserver2. Result: exit with error
2. `beeline` connect kyuubi(before this patch). Result: hang
3. `beeline` connect kyuubi(after this patch). Result:  exit with error

### DIFF
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
@@ -91,8 +91,10 @@ class ExecuteStatement(
       isComplete = true
       remoteState match {
         case INITIALIZED_STATE | PENDING_STATE | RUNNING_STATE =>
-          isComplete = false
-          statusResp = client.GetOperationStatus(statusReq)
+          try {
+            statusResp = client.GetOperationStatus(statusReq)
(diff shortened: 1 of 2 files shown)


kyuubi-main/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
```diff
@@ -91,8 +91,10 @@ class ExecuteStatement(
       isComplete = true
       remoteState match {
         case INITIALIZED_STATE | PENDING_STATE | RUNNING_STATE =>
-          isComplete = false
-          statusResp = client.GetOperationStatus(statusReq)
+          try {
```
yaooqinn, line 94: shall we wrap the whole method？ there are still other places calling the `client`

yaooqinn:
cc @ulysses-you, we shall fail 1.2.0-rc3 for this

yaooqinn:
please link the target issue

ulysses-you:
nice catch !

yaooqinn (approved):
LGTM

PR #654, opened 2021-05-31
[GA] Enable Kyuubi nightly build against Spark master

pan3793:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,56 @@
+name: Kyuubi Nightly
+
+on:
+  schedule:
+    - cron: 0 23 * * *
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        profiles:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,56 @@
+name: Kyuubi Nightly
+
+on:
+  schedule:
+    - cron: 0 23 * * *
```
yaooqinn, line 5: What is the time for spark nightly build?

yaooqinn (approved):

.github/workflows/nightly.yml
pan3793: midnight, and should we change it to 1:00 am?

pan3793: Have look on Spark GA list, nightly build usually cost 1~2 hours, and sometime over 3 hours, maybe we should start at 4:00 am.

yaooqinn:
will merge this affer 6:00 pm, Asia/Shanghai

yaooqinn:

Issue #662, opened 2021-06-03
Fix run on k8s


hddong:
### 1. Describe the bug

When run on k8s, engine write hostname to zk. But, server can not connect pod by hostname, we should use ip instead.

#### 2.1 Versions

- Kyuubi
  - [x] 1.1.0
  - [ ] 1.0.3
  - [ ] 0.8.x and earlier 

- Spark
  - [x] 3.1.x
  - [ ] 3.0.x
  - [ ] 2.4.x and earlier 

#### 4.1 Log of the server
```
org.apache.kyuubi.KyuubiSQLException: Error opening session SessionHandle [4567b833-5690-48e4-b384-052a7bcf2daa] for test due to java.net.UnknownHostException: kyuubiusertest4567b833-5690-48e4-b384-052a7bcf2daa-03cd9b79cbb7
```
#### 4.2 Log of the corresponding engine

PR #663, opened 2021-06-03
[KYUUBI #662]Fix run on k8s

hddong:

When run on k8s, engine write hostname to zk. But, server can not connect pod by hostname, we should use ip instead.


yaooqinn:
cc @zwangsheng

yaooqinn:
@hddong

Can we control this on the Kyuubi side not based on the cluster manager where engines running against?

We can add a conf directly e.g. `kyuubi.engine.connection.url.use.hostname=true`, then whether underlying platform change or not, we can use this to switch hostname or IP?

zwangsheng:
Hi @hddong , What mode did you submit ？

hddong:
> Hi @hddong , What mode did you submit ？
cluster mode, dirver as a pod.

yaooqinn (approved):

yaooqinn:

Issue #665, opened 2021-06-04
Skip exist check when main resource jar stored remote


hddong:
### 1. Describe the feature
When main resource specified by user, we always check exists.  Not need check when jar stored on HDFS or in k8s images.

### 3. Describe the solution
Skip exist check when main resource jar stored remote

PR #666, opened 2021-06-04
[KYUUBI #665]Skip exist check when main resource jar stored remote

hddong:

When main resource specified by user, we always check exists. Not need check when jar stored on HDFS or in k8s images.


kyuubi-main/src/test/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilderSuite.scala
```diff
@@ -204,6 +205,23 @@ class SparkProcessBuilderSuite extends KerberizedTestHelper {
     Files.write(file2.toPath, "a".getBytes(), StandardOpenOption.APPEND)
     assert(file2.length() == 1)
   }
+
+  test("main resource jar should not check when is not a local file") {
```
yaooqinn, line 209: can we also test relative path？

PR #667, opened 2021-06-06
Better error msg if yarn queue has no reousrce during submitting app

ulysses-you:

Give more readable error msg if the queue has no resource after submitting app timeout.

before
```
Failed to detect the root cause, please check $engineLog at server side if necessary
```

after
```
The last status of Spark App is ACCEPTED, please check your cluster resource
```

Add new test.

### DIFF
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
@@ -63,6 +63,7 @@ trait ProcBuilder {
   }
 
   @volatile private var error: Throwable = UNCAUGHT_ERROR
+  private var lastRowOfLog: String = _
   // Visible for test
   private[kyuubi] var logCaptureThread: Thread = _
 
@@ -101,6 +102,14 @@ trait ProcBuilder {
     file
   }
 
+  private def checkAcceptedStatus: Option[Throwable] = {
+    if (lastRowOfLog != null && lastRowOfLog.contains("state: ACCEPTED")) {
+      Some(ACCEPTED_ERROR)
+    } else {
+      None
+    }
+  }
+
   final def start: Process = synchronized {
 
     val proc = processBuilder.start()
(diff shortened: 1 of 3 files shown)


kyuubi-main/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -101,6 +102,14 @@ trait ProcBuilder {
     file
   }
 
+  private def checkAcceptedStatus: Option[Throwable] = {
+    if (lastRowOfLog != null && lastRowOfLog.contains("state: ACCEPTED")) {
```
turboFei, line 106: This  log pattern seems bound to yarn + spark?

Shall we move it into SparkProcessBulder?

ulysses-you: yea, I created a issue [#668](https://github.com/NetEase/kyuubi/issues/668) to track this. This PR has not finished yet.

ulysses-you: sounds good

yaooqinn (approved):

kyuubi-main/src/test/scala/org/apache/kyuubi/engine/EngineSubmitSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+
+class EngineSubmitSuite extends WithKyuubiServerWithMiniYarnService with JDBCTestUtils {
+  override protected def jdbcUrl: String = getJdbcUrl
+  override protected val kyuubiServerConf: KyuubiConf = KyuubiConf()
+  override protected val connectionConf: Map[String, String] = {
+    Map("spark.yarn.queue" -> "two_cores_queue",
+      "spark.master" -> "yarn",
+      "spark.executor.instances" -> "1",
+      "spark.driver.cores" -> "1",
+      "spark.executor.cores" -> "1",
```
yaooqinn, line 32: Can we simply request an executor or AM with exceeded cores to make the app hang?

ulysses-you: `spark.executor.instances` cann't be 0 so executor always exists.. If we want to use am to hold the resource, we need to tune other config to make app available without executor, otherwise app will wait until executor available.

yaooqinn:

PR #676, opened 2021-06-09
Refactor engine process

ulysses-you:

Refactor engine process with follow mainly change:

* Add new class `EngineProcess` which is a wrapped of `Process`
* Move code from `ProcessBuilder` to `EngineProcess`

Pass CI


ulysses-you:
cc @yaooqinn @pan3793

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/EngineProcess.scala
```diff
@@ -0,0 +1,172 @@
...
+
+import org.apache.commons.lang3.StringUtils.containsIgnoreCase
+
+import org.apache.kyuubi.KyuubiSQLException
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.engine.EngineProcess.{PROC_BUILD_LOGGER, UNCAUGHT_ERROR}
+import org.apache.kyuubi.util.NamedThreadFactory
+
+/**
+ * A wrapper of [[java.lang.Process]]
```
yaooqinn, line 33: why it's a wrapper for [[java.lang.Process]] but we do not touch any of [[java.lang.Process]] in this trait?

kyuubi-main/src/main/scala/org/apache/kyuubi/engine/EngineProcess.scala
```diff
@@ -0,0 +1,172 @@
...
+}
+object EngineProcess {
+  val UNCAUGHT_ERROR = new RuntimeException("Uncaught error")
+  val PROC_BUILD_LOGGER = new NamedThreadFactory("process-logger-capture", daemon = true)
+}
+
+/**
+ * Note that: This class is not thread safe.
+ */
+class EngineProcess(processBuilder: ProcessBuilder, conf: KyuubiConf, engineLog: File)
```
yaooqinn, line 80: Will we have another concrete class for IEngineProcess

yaooqinn:
TBH, I think this brings extra overhead for developers to create a new type of engine. Before they only need implement a new ProcBuilder by mimicking SparkProcBuilder with `java.lang.Process`, now they have to learn How EngineProcess works.

How about we only extract the log capturing part to a new trait? e.g. `EngineLogCapturer`？

ulysses-you:
Seems a little overkill, close this.

Issue #678, opened 2021-06-10
kyuubi ctl use default property if missing


hddong:
#### 2.1 Versions

- Kyuubi
  - [ ] 1.1.0
  - [ ] 1.0.3
  - [ ] 0.8.x and earlier 

- Spark
  - [ ] 3.1.x
  - [ ] 3.0.x
  - [ ] 2.4.x and earlier 

#### 3.1 Operations
`bin/kyuubi-ctl list` throw:
```
Exception in thread "main" org.apache.kyuubi.KyuubiException: Zookeeper namespace is not specified and no default value to load
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.fail(ServiceControlCliArguments.scala:339)
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.validateZkArguments(ServiceControlCliArguments.scala:136)
```

#### 3.2 Expected behavior
list engine nodes with default namespace `kyuubi`

#### 4.2 Log of the corresponding engine

PR #679, opened 2021-06-10
[KYUUBI #678]Fix: kyuubi ctl use default property if missing

hddong:

when use `bin/kyuubi-ctl list` throw:
```
Exception in thread "main" org.apache.kyuubi.KyuubiException: Zookeeper namespace is not specified and no default value to load
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.fail(ServiceControlCliArguments.scala:339)
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.validateZkArguments(ServiceControlCliArguments.scala:136)
```
we expect: list engine nodes with default namespace `kyuubi`

### DIFF
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
@@ -63,11 +63,13 @@ class ServiceControlCliArguments(args: Seq[String], env: Map[String, String] = s
     // for create action, it only expose Kyuubi service instance to another domain,
     // so we do not use namespace from default conf
     if (action != ServiceControlAction.CREATE && namespace == null) {
-      Option(conf.get(HA_ZK_NAMESPACE)).foreach { v =>
+      namespace = conf.getOption(HA_ZK_NAMESPACE.key).getOrElse{
+        val defaultNamespace = conf.get(HA_ZK_NAMESPACE)
         if (verbose) {
-          info(s"Zookeeper namespace is not specified, use value from default conf:$v")
+          info(s"Zookeeper namespace is not specified," +
+            s" use value from default conf:$defaultNamespace")
         }
-        namespace = v
+        defaultNamespace
       }
     }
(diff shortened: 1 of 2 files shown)


yaooqinn:
thank you very much @hddong, 
LGTM, only minor revisions.

hddong:
@yaooqinn: thanks for your review, make some modifications to keep `verbose` log output.

yaooqinn (approved):

yaooqinn:
LGTM

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
```diff
@@ -63,11 +63,10 @@ class ServiceControlCliArguments(args: Seq[String], env: Map[String, String] = s
...
     if (action != ServiceControlAction.CREATE && namespace == null) {
-      conf.getOption(HA_ZK_NAMESPACE.key).foreach { v =>
-        if (verbose) {
-          info(s"Zookeeper namespace is not specified, use value from default conf:$v")
-        }
-        namespace = v
+      namespace = conf.get(HA_ZK_NAMESPACE)
+      if (verbose) {
+        info(s"Zookeeper namespace is not specified," +
+          s" use value from default conf:$namespace")
```
yaooqinn, line 69: NOT related to the PR, can we avoid mess up verbose with log4j?

yaooqinn:
thanks, merged to master

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
hddong: > NOT related to the PR, can we avoid mess up verbose with log4j?

IMO, we can have a try, use a custom log level.

yaooqinn: Yea

PR #687, opened 2021-06-13
Enable transfer env variables to engine launcher process

turboFei:
Enable transfer some env variables to engine launcher process


pan3793 (approved):
LGTM

yaooqinn:
A list may be hard to use and cause some parsing issues for  the JDBC URL, can we make it simpler? Users can use `kyuubi.engineEnv.VAR_NAME`, then provides a `getEnvs` in `KyuubiConf` to retrieve all environment vars

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -101,6 +101,13 @@ case class KyuubiConf(loadSysDefault: Boolean = true) extends Logging {
     settings.entrySet().asScala.map(x => (x.getKey, x.getValue)).toMap[String, String]
   }
 
+  /** Get all engine envs as map */
+  def getEngineEnvs: Map[String, String] = {
```
yaooqinn, line 105: use getAllWithPrefix?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -101,6 +101,13 @@ case class KyuubiConf(loadSysDefault: Boolean = true) extends Logging {
...
+    settings.entrySet().asScala.filter(_.getKey.startsWith(KYUUBI_ENGINE_ENV_PREFIX))
+      .map(entry => entry.getKey.stripPrefix(KYUUBI_ENGINE_ENV_PREFIX) -> entry.getValue)
```
yaooqinn, line 107: use getAllWithPrefix?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
yaooqinn, line 105: `getEnvs` can be more general name

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
yaooqinn: this should also lookup `sys.env` first and apply `KYUUBI_ENGINE_ENV_PREFIX`s in the rightmost way

yaooqinn (approved):
LGTM

yaooqinn:

PR #696, opened 2021-06-15
Save Hadoop conf file instead of appending all of them to connection url

turboFei:

Save Hadoop conf file instead of appending all of them to connection url

### DIFF
--- a/kyuubi-main/src/test/scala/org/apache/kyuubi/server/MiniYarnService.scala
+++ b/kyuubi-main/src/test/scala/org/apache/kyuubi/server/MiniYarnService.scala
@@ -18,14 +18,12 @@
 package org.apache.kyuubi.server
 
 
-import java.io.File
+import java.io.{File, FileWriter}
 import java.net.InetAddress
-import java.nio.charset.StandardCharsets
 
 import scala.collection.JavaConverters._
 
-import com.google.common.io.Files
-import org.apache.commons.lang.StringEscapeUtils
+import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.hadoop.yarn.conf.YarnConfiguration
 import org.apache.hadoop.yarn.server.MiniYARNCluster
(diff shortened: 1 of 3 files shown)


kyuubi-main/src/test/scala/org/apache/kyuubi/server/MiniYarnService.scala
```diff
@@ -104,6 +109,24 @@ class MiniYarnService(name: String) extends AbstractService(name) {
     }.toMap
   }
 
+  private def saveHadoopConf(): Unit = {
```
yaooqinn, line 112: Can org.apache.hadoop.conf.Configuration#writeXml(java.io.OutputStream) be used here？

turboFei: thanks

kyuubi-main/src/test/scala/org/apache/kyuubi/server/MiniYarnService.scala
```diff
@@ -97,11 +100,14 @@ class MiniYarnService(name: String) extends AbstractService(name) {
...
+    val configToWrite = new Configuration(false)
     val hostName = InetAddress.getLocalHost.getHostName
-    yarnCluster.getConfig.iterator().asScala.map { kv =>
-      kv.getKey -> kv.getValue.replaceAll(hostName, "localhost")
-    }.toMap
+    yarnCluster.getConfig.iterator().asScala.foreach { kv =>
+      configToWrite.set(kv.getKey, kv.getValue.replaceAll(hostName, "localhost"))
+    }
+    val writer = new FileWriter(new File(hadoopConfDir, "yarn-site.xml"))
+    configToWrite.writeXml(writer)
```
pan3793, line 110: will the `writeXml` close writer?

kyuubi-main/src/test/scala/org/apache/kyuubi/server/MiniYarnService.scala
```diff
@@ -97,11 +100,14 @@ class MiniYarnService(name: String) extends AbstractService(name) {
     super.stop()
```
ulysses-you, line 100: Shall we delete the temporary file after stop ?

turboFei: thanks, added.

The file created by Utils.createTempDir should be deleted on exit.

kyuubi-main/src/test/scala/org/apache/kyuubi/server/MiniYarnService.scala
turboFei: thanks, fixed.

pan3793 (approved):
LGTM, thanks!

yaooqinn (approved):

ulysses-you:

PR #700, opened 2021-06-17
Add class description for DataGenerator

ulysses-you:

Add usage description so we can use this tpcds data generator easily.

Not need.

### DIFF
--- a/dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/DataGenerator.scala
+++ b/dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/DataGenerator.scala
@@ -21,6 +21,13 @@ import org.apache.spark.SparkConf
 import org.apache.spark.sql.SparkSession
 import org.slf4j.LoggerFactory
 
+// scalastyle:off line.size.limit
+/**
+ * Usage:
+ * If you want to generate 10GB data with new database `tpcds_sf10`, you can run the cmd:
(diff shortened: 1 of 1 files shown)


yaooqinn (approved):

dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/DataGenerator.scala
```diff
@@ -21,6 +21,13 @@ import org.apache.spark.SparkConf
 import org.apache.spark.sql.SparkSession
 import org.slf4j.LoggerFactory
 
+// scalastyle:off line.size.limit
+/**
+ * Usage:
+ * If you want to generate 10GB data with new database `tpcds_sf10`, you can run the cmd:
+ *   `$SPARK_HOME/bin/spark-submit --conf spark.sql.tpcds.scale.factor=10 --conf spark.sql.tpcds.database=tpcds_sf10 --class org.apache.kyuubi.tpcds.DataGenerator kyuubi-tpcds-1.3.0-SNAPSHOT.jar`
```
pan3793, line 28: is the command valid if use `kyuubi-tpcds-*.jar` instead of a specific version?

ulysses-you: use * instead

pan3793 (approved):
LGTM, thanks!

pan3793:
Thanks, merged for v1.3.0.

ulysses-you:
thanks all

PR #707, opened 2021-06-22
Respect spark.sql.catalogImplementation in engine

pan3793:

Respect `spark.sql.catalogImplementation` in engine.


yaooqinn:
`spark.sql.catalogImplementation` should be the first choice for this feature.

yaooqinn:
How about?
```
val defaultCat = if (hiveClassesArePresent()) "hive" else "in-memory"

conf.setIfMissing("spark.sql.catalogImplementation", defaultCat)
```

yaooqinn:
FYI, spark uses this for detecting hive classes

```
  /**
   * @return true if Hive classes can be loaded, otherwise false.
   */
  private[spark] def hiveClassesArePresent: Boolean = {
    try {
      Utils.classForName(HIVE_SESSION_STATE_BUILDER_CLASS_NAME)
      Utils.classForName("org.apache.hadoop.hive.conf.HiveConf")
      true
    } catch {
      case _: ClassNotFoundException | _: NoClassDefFoundError => false
    }
  }
```

pan3793:
make sense, will switch to `spark.sql.catalogImplementation`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/EngineUtils.scala
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.kyuubi
+
+import org.apache.spark.sql.SparkSession
+
+object EngineUtils {
+
+  def hiveClassesArePresent: Boolean = SparkSession.hiveClassesArePresent
```
pan3793, line 24: Since `private[spark] SparkSession.hiveClassesArePresent`, we need to export it for Kyuubi using.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/EngineUtils.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.kyuubi
+
+import org.apache.spark.sql.SparkSession
+
+object EngineUtils {
```
yaooqinn, line 22: org.apache.kyuubi.engine.spark.KyuubiSparkUtil

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/EngineUtils.scala
yaooqinn: we can simply fork it

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -96,7 +96,9 @@ object SparkSQLEngine extends Logging {
       }
     }
 
-    val session = SparkSession.builder().config(sparkConf).enableHiveSupport().getOrCreate()
+    val defaultCat = if (EngineUtils.hiveClassesArePresent) "hive" else "in-memory"
+    sparkConf.setIfMissing("spark.sql.catalogImplementation", defaultCat)
```
yaooqinn, line 100: can we move these to to L83

yaooqinn (approved):

pan3793:
Thanks for reviewing, merged into master for v1.3.0.

Issue #708, opened 2021-06-22
[BUILD] Upgrade scala-maven-plugin to 4.5.3


pan3793:
### 1. Describe the feature
Upgrade the `scala-maven-plugin` version to 4.5.3.

### 2. Motivation
This will upgrade `sbt-compiler-bridge` from 1.3.1 to 1.5.5 in order to bring the latest bug fixes.

### 3. Describe the solution
Upgrade the `scala-maven-plugin` version to 4.5.3.

### 4. Additional context
SPARK-35850

pan3793:
I'm afraid that we can't do this upgrade, confirmed that #441 still existed with `scala-maven-plugin` 4.5.3.
```
[ERROR] Failed to execute goal net.alchim31.maven:scala-maven-plugin:4.5.3:testCompile (scala-test-compile-first) on project kyuubi-common: wrap: java.io.IOException: 
Cannot run program "/usr/java/jdk1.8.0_221-amd64/jre/bin/javac" (in directory "/data/ansible/ci/kyuubi"): error=2, No such file or directory -> [Help 1]
```

PR #441, opened 2021-03-16
[BUILD] Downgrade scala-maven-plugin to 4.3.0

pan3793:
pan3793 [Closes #441](https://github.com/yaooqinn/kyuubi/pull/441)    Target Issue [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=yaooqinn&utm_campaign=badge_info)

### _Why are the changes needed?_

There are several known issues with `scala-maven-plugin:4.4.0` about `zinc`.

On CentOS 7, with Oracle JDK 8 installed via RPM, `/bin/java` links to `jdk/jre/bin/java` but `/bin/javac` links to `jdk/bin/javac`
```
[ops@bd-ck-uat-30007-vm ~]$ ll -l /bin/java
lrwxrwxrwx 1 root root 22 Mar 12 14:03 /bin/java -> /etc/alternatives/java
[ops@bd-ck-uat-30007-vm ~]$ ll -l /etc/alternatives/java
lrwxrwxrwx 1 root root 41 Mar 12 14:03 /etc/alternatives/java -> /usr/java/jdk1.8.0_251-amd64/jre/bin/java
[ops@bd-ck-uat-30007-vm ~]$ ll -l /bin/javac
lrwxrwxrwx 1 root root 23 Mar 12 14:03 /bin/javac -> /etc/alternatives/javac
[ops@bd-ck-uat-30007-vm ~]$ ll -l /etc/alternatives/javac
lrwxrwxrwx 1 root root 38 Mar 12 14:03 /etc/alternatives/javac -> /usr/java/jdk1.8.0_251-amd64/bin/javac
```
when run `build/dist --tgz` cause
```
[ERROR] Failed to execute goal net.alchim31.maven:scala-maven-plugin:4.4.0:testCompile (scala-test-compile-first) on project kyuubi-common: wrap: java.io.IOException: Cannot run program "/usr/java/jdk1.8.0_221-amd64/jre/bin/javac" (in directory "/data/ansible/ci/kyuubi"): error=2, No such file or directory -> [Help 1]
```
And there is another issue reported by Apache Spark: https://github.com/apache/spark/pull/31031

Please also merge it into branch-1.1

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [x] Add screenshots for manual tests if appropriate

- [x] [Run test](https://kyuubi.readthedocs.io/en/latest/tools/testing.html#running-tests) locally before make a pull request

PR #720, opened 2021-06-26
[KYUUBI #708] [BUILD] Upgrade scala-maven-plugin to 4.5.3

wuchunfu:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,6 +96,7 @@
         <maven.plugin.build.helper.version>3.2.0</maven.plugin.build.helper.version>
         <maven.plugin.compiler.version>3.8.1</maven.plugin.compiler.version>
         <!-- DO NOT bump 4.4.0, see https://github.com/NetEase/kyuubi/pull/441 -->
+        <!-- DO NOT bump 4.5.3, see https://github.com/NetEase/kyuubi/issues/708 -->
         <maven.plugin.scala.version>4.5.3</maven.plugin.scala.version>
         <maven.plugin.surefire.version>2.22.0</maven.plugin.surefire.version>
         <maven.plugin.scalatest.version>2.0.0</maven.plugin.scalatest.version>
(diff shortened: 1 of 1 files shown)


yaooqinn (approved):

pom.xml
```diff
@@ -96,7 +96,7 @@
         <maven.plugin.build.helper.version>3.2.0</maven.plugin.build.helper.version>
         <maven.plugin.compiler.version>3.8.1</maven.plugin.compiler.version>
         <!-- DO NOT bump 4.4.0, see https://github.com/NetEase/kyuubi/pull/441 -->
```
pan3793, line 98: Have a verification that #441 still exists in 4.5.3, I think we need to figure out what happens in scala maven plugin before doing this upgrading.

wuchunfu: Now I will upgrade the version of `scala-maven-plugin` to `4.5.3` and pass the local verification, and run the `build/dist --tgz` command to package normally

pan3793: Yes, it also passed ci but not the case reported by #441

wuchunfu: What should I do?

pan3793: We need to figure out what happens in scala maven plugin cause #441 first. 
If it's an accident change, fix it in scala maven plugin;
if scala maven plugin team confirme the change is by design and they don't think #441 should be supported, we can do upgrade and ignore #441.

yaooqinn: Let's add a new comment here and skip this version too.

```
<!-- DO NOT bump 4.4.0, see https://github.com/NetEase/kyuubi/pull/441 -->
<!-- DO NOT bump 4.5.3, see https://github.com/NetEase/kyuubi/issues/708 -->
```

pom.xml
```diff
@@ -96,7 +96,8 @@
...
-        <maven.plugin.scala.version>4.3.0</maven.plugin.scala.version>
+        <!-- DO NOT bump 4.5.3, see https://github.com/NetEase/kyuubi/issues/708 -->
+        <maven.plugin.scala.version>4.5.3</maven.plugin.scala.version>
```
pan3793, line 100: revert it to `4.3.0`

turboFei (approved):

pom.xml
yaooqinn: @wuchunfu

pan3793 (approved):
LGTM

pom.xml
wuchunfu: @yaooqinn I has now been revert and add a new comment here and skip this version too.

pan3793:
Thanks, merged for v1.3.0.

Issue #694, opened 2021-06-13
Upgrade Scala to 2.12.14


yaooqinn:
### 1. Describe the feature
https://www.scala-lang.org/download/2.12.14.html is now available

PR #721, opened 2021-06-26
[KYUUBI #694] Upgrade Scala to 2.12.14

wuchunfu:


pan3793 (approved):
LGTM

pan3793 (changes requested):
We should also update `dev/dependenyList`

yaooqinn (approved):
LGTM

PR #722, opened 2021-06-27
[TEST] Use in-memory catalog in most test suites

pan3793:

Followup #707, use in-memory catalog in most test suites to avoid derby lock to speed up test suites.


externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
```diff
@@ -33,10 +33,10 @@ import org.apache.spark.sql.types._
 import org.apache.kyuubi.Utils
 import org.apache.kyuubi.engine.spark.WithSparkSQLEngine
 import org.apache.kyuubi.engine.spark.shim.SparkCatalogShim
-import org.apache.kyuubi.operation.JDBCTests
+import org.apache.kyuubi.operation.HiveJDBCTests
 import org.apache.kyuubi.operation.meta.ResultSetSchemaConstant._
 
-class SparkOperationSuite extends WithSparkSQLEngine with JDBCTests {
+class SparkOperationSuite extends WithSparkSQLEngine with HiveJDBCTests {
```
turboFei, line 39: Do you set catalog implementation to hive in this test suite?

turboFei: For HiveJDBCTests, it is only test for hive catalog?

pan3793: hive is the default catalogImplementation
```
    val defaultCat = if (KyuubiSparkUtil.hiveClassesArePresent) "hive" else "in-memory"
    sparkConf.setIfMissing("spark.sql.catalogImplementation", defaultCat)
```

pan3793: Actually, hive catalog has more functionalities than in-memory catalog, I'm not sure if we need in-memory catalog test in both engine side and kyuubi server side.

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/HiveJDBCTests.scala
```diff
@@ -0,0 +1,288 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+import org.apache.kyuubi.Utils
+import org.apache.kyuubi.operation.meta.ResultSetSchemaConstant._
+
+trait HiveJDBCTests extends JDBCTestUtils {
```
yaooqinn, line 23: this base trait mainly duplicated, can we make it neater?

pan3793: updated, let trait HiveJDBCTests extends BasicJDBCTests

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/HiveJDBCTests.scala
```diff
@@ -0,0 +1,100 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+import org.apache.kyuubi.Utils
+import org.apache.kyuubi.operation.meta.ResultSetSchemaConstant._
+
+trait HiveJDBCTests extends BasicJDBCTests {
```
yaooqinn, line 23: HiveCatalogTests

yaooqinn (approved):

Issue #728, opened 2021-06-30
0A000: feature_not_supported


yaooqinn:
### 1. Describe the feature

Follow https://github.com/NetEase/kyuubi/issues/670 to: 
Apply `errorCode`  `0A000: feature_not_supported` to `val errStatus = KyuubiSQLException("Feature is not available").toTStatus`

### 3. Describe the solution

Add this to `KyuubiSQLException` object and use it wherever needed
```scala

  def featureNotSupported(): KyuubiSQLException = {
    KyuubiSQLException("feature not supported", sqlState = "0A000")
  }

```

yaooqinn:
cc @zhaomin1423

zhaomin1423:
> cc @zhaomin1423
I would like to do it.

Issue #670, opened 2021-06-08
Support SQLSTATE and venderCode in KyuubiSQLException 


yaooqinn:
### 1. Describe the feature

https://www.postgresql.org/docs/12/errcodes-appendix.html
https://docs.oracle.com/cd/E15817_01/appdev.111/b31228/appd.htm
https://www.ibm.com/docs/en/db2-for-zos/11?topic=codes-sqlstate-values-common-error
https://en.wikipedia.org/wiki/SQLSTATE
https://wiki.midrange.com/index.php/SQLSTATE_Codes
https://docs.teradata.com/r/CeAGk~BNtx~axcR0ed~5kw/TONkXZQG19hcmPLSw~O36w

### 3. Describe the solution

We can make KyuubiSQLException inherit

 `public SQLException(String reason, String sqlState, int vendorCode, Throwable cause)` first.

yaooqinn:
track https://github.com/apache/spark/pull/32850

PR #729, opened 2021-06-30
[KYUUBI #728] Support SQL State: 0A000 - feature_not_supported

zhaomin1423:

Follow #670 to:
Apply errorCode 0A000: feature_not_supported to val errStatus = KyuubiSQLException("Feature is not  @available").toTStatus

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/KyuubiSQLException.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/KyuubiSQLException.scala
@@ -81,6 +81,10 @@ object KyuubiSQLException {
     }
   }
 
+  def featureNotSupported(): KyuubiSQLException = {
+    KyuubiSQLException("feature not supported", sqlState = "0A000")
+  }
+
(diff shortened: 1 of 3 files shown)


yaooqinn:
we shall call this at FrontendService for。unsupported features

kyuubi-common/src/main/scala/org/apache/kyuubi/KyuubiSQLException.scala
```diff
@@ -81,6 +81,10 @@ object KyuubiSQLException {
     }
   }
 
+  def featureNotSupported(): KyuubiSQLException = {
+    KyuubiSQLException("feature not supported", sqlState = "0A000")
+  }
```
turboFei, line 86: could you invoke it in the FrontendService for unsupported operation?

zhaomin1423: done

yaooqinn (approved):

ulysses-you:

PR #741, opened 2021-07-02
[TEST] Refine test suites

pan3793:

Rename `KyuubiOperationHiveSuite` to `KyuubiOperationHiveCatalogSuite`;
Simplify redundant case in `KyuubiOperationPerConnectionSuite`;

### DIFF
--- a/kyuubi-main/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala
+++ b/kyuubi-main/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala
@@ -26,7 +26,7 @@ import org.apache.kyuubi.config.KyuubiConf
 /**
  * UT with Connection level engine shared cost much time, only run basic jdbc tests.
  */
-class KyuubiOperationPerConnectionSuite extends WithKyuubiServer with BasicJDBCTests {
+class KyuubiOperationPerConnectionSuite extends WithKyuubiServer with JDBCTestUtils {
 
   override protected def jdbcUrl: String = getJdbcUrl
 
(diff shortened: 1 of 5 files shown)


kyuubi-main/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala
```diff
@@ -26,7 +26,7 @@ import org.apache.kyuubi.config.KyuubiConf
 /**
  * UT with Connection level engine shared cost much time, only run basic jdbc tests.
  */
-class KyuubiOperationPerConnectionSuite extends WithKyuubiServer with BasicJDBCTests {
+class KyuubiOperationPerConnectionSuite extends WithKyuubiServer with JDBCTestUtils {
```
ulysses-you, line 29: We don't need to test basic tests for connection share level ?

yaooqinn (approved):

kyuubi-main/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala
pan3793: The cases in `JDBCTestUtils` got tested on Engine side, and several times on Kyuubi Server side, it's quite redundant. Thus just keep connection share level specific cases in `KyuubiOperationPerConnectionSuite` to simplify the test suites.

ulysses-you:

kyuubi-main/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala
ulysses-you: make sense

Issue #659, opened 2021-06-02
Add UDF engine_name


yaooqinn:
### 1. Describe the feature

1. engine_name means the sparkContext.conf.get("spark.app.name")
2. add doc too

### 2. Motivation

system information functions

zhaomin1423:
I would like to do it.

zhaomin1423:
Done

PR #747, opened 2021-07-04
[KYUUBI #659] Add UDF engine_name

zhaomin1423:

Add UDF engine_name to show the runtime information about execution engine


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
```diff
@@ -36,6 +37,14 @@ object KDFRegistry {
     "string",
     "1.3.0")
 
+  val engine_name: KyuubiDefinedFunction = create(
+    "engine_name",
+    udf(() => SparkContext.getOrCreate().getConf.get("spark.app.name")).asNonNullable(),
```
yaooqinn, line 42: SparkSession.active.sparkContext.appName

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
```diff
@@ -36,6 +37,14 @@ object KDFRegistry {
...
+    "Return the engine_name of Kyuubi Server",
```
yaooqinn, line 43: Return the  spark application name for the associated query engine

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
zhaomin1423: Thanks, I will update it later.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
zhaomin1423: Thansk for the suggestion.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
zhaomin1423: 'SparkSession.active' can't get the SparkSession when it is running in the executor, so the 'SparkSession.active.sparkContext.appName' have an exception of 'No active or default Spark session found.'

turboFei: How about `SparkEnv.get.conf`?

zhaomin1423: SparkEnv.get.conf can work well,  what is the difference between SparkEnv.get and SparkContext.getOrCreate().getConf, It looks like that their output is the same about appName.

turboFei: SparkContext only existing in driver.

For sparkEnv, it is about seria/lizerManager/mapOutputTracker/blockManager.
And there are driverEnv and executorEnv.
So, you can access it both on driver and executor.

turboFei: FYI:
https://github.com/apache/spark/blob/554d5fef13579cefea083e69b98bce32c0c99126/core/src/main/scala/org/apache/spark/SparkEnv.scala#L197

zhaomin1423: Thanks, I will update it later.

yaooqinn: can we get `val appName = SparkSession.active.sparkContext.appName` at driver side ahead and then pass  the `appName` into the udf?

zhaomin1423: Make sense, getting it directly is more concise unless there are some advantages.

turboFei (approved):
Thanks

yaooqinn (approved):

yaooqinn:
it conflicts, can you resolve it? @zhaomin1423

yaooqinn (approved):

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/JDBCTests.scala
```diff
@@ -360,6 +361,14 @@ trait JDBCTests extends BasicJDBCTests {
     }
   }
 
+  test("kyuubi defined function - engine_name") {
```
pan3793, line 364: The test will fail in the MiniKube integration test due to the docker image issue. Please temporarily ignore it just like #938.

pan3793:
I'm going to merge it and send a following PR to ignore the test until we fix the docker issue.

PR #758, opened 2021-07-06
[KYUUBI 661] Add UDF system_user

zhaomin1423:

add udf system_user.

### DIFF
--- a/docs/sql/functions.md
+++ b/docs/sql/functions.md
@@ -13,4 +13,5 @@ Kyuubi provides several auxiliary SQL functions as supplement to Spark's [Built-
 Name | Description | Return Type | Since
 --- | --- | --- | ---
 kyuubi_version | Return the version of Kyuubi Server | string | 1.3.0
+system_user | Return the system name for the associated query engine | string | 1.3.0
 
--- a/docs/sql/functions.md
+++ b/docs/sql/functions.md
@@ -13,5 +13,5 @@ Kyuubi provides several auxiliary SQL functions as supplement to Spark's [Built-
 Name | Description | Return Type | Since
 --- | --- | --- | ---
 kyuubi_version | Return the version of Kyuubi Server | string | 1.3.0
-system_user | Return the system name for the associated query engine | string | 1.3.0
+system_user | Return the system user name for the associated query engine | string | 1.3.0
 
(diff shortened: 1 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
```diff
@@ -36,6 +36,13 @@ object KDFRegistry {
     "string",
     "1.3.0")
 
+  val system_user: KyuubiDefinedFunction = create(
+    "system_user",
+    udf(() => System.getProperty("user.name")).asNonNullable(),
+    "Return the system name for the associated query engine",
```
turboFei, line 42: system user name

docs/sql/functions.md
```diff
@@ -13,4 +13,5 @@ Kyuubi provides several auxiliary SQL functions as supplement to Spark's [Built-
 Name | Description | Return Type | Since
 --- | --- | --- | ---
 kyuubi_version | Return the version of Kyuubi Server | string | 1.3.0
+system_user | Return the system name for the associated query engine | string | 1.3.0
```
turboFei, line 16: system user name

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
zhaomin1423: Thanks for the suggestion.

ulysses-you:
Just curious, where do you want to use the system user instead of `UserGroupInformation.getCurrentUser` ? For me, it's more like a administrator function. And then maybe we can have some other useful functions `current_machine`, `current_session_number`, `current_statement_number` ?

zhaomin1423:
@ulysses-you  Should the system_user return the UserGroupInformation.getCurrentUser?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
```diff
@@ -36,6 +36,13 @@ object KDFRegistry {
     "string",
     "1.3.0")
 
+  val system_user: KyuubiDefinedFunction = create(
+    "system_user",
+    udf(() => System.getProperty("user.name")).asNonNullable(),
```
turboFei, line 41: should we use a lazy val for `System.getProperty("user.name")` and return it directly? @yaooqinn

turboFei (approved):
LGTM

yaooqinn:
thanks @zhaomin1423 @turboFei , merged to master

PR #763, opened 2021-07-06
add TPC-DS Tests for spark-monitor module

zhang1002:
add TPC-DS Tests for spark-monitor module


externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -43,7 +43,13 @@
             <artifactId>kyuubi-ha</artifactId>
             <version>${project.version}</version>
         </dependency>
-
```
ulysses-you, line 46: unnecessary change

ulysses-you:
thanks, merging to master

PR #764, opened 2021-07-08
Add Customized Configuration Item: kyuubi.statement.id

zhang1002:
For broadcast, Spark will introduce a new runId as SPARK_JOB_GROUP_ID. That cause we can't follow up this statement's whole life by SPARK_JOB_GROUP_ID.
This configuration item will be set when this statement was submitted and it will not be changed.
You can get the dependency relationship from job or stage to statement by this configuration item in this statement's whole life.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -122,16 +122,17 @@ class ExecuteStatement(
   private def withLocalProperties[T](f: => T): T = {
     try {
       spark.sparkContext.setJobGroup(statementId, statement, forceCancel)
-      spark.sparkContext.setLocalProperty("kyuubi.statement.id", statementId)
+      spark.sparkContext.setLocalProperty(ExecutionStatement.KYUUBI_STATEMENT_ID_KEY, statementId)
       schedulerPool match {
         case Some(pool) =>
-          spark.sparkContext.setLocalProperty("spark.scheduler.pool", pool)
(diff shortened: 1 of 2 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -122,6 +122,7 @@ class ExecuteStatement(
   private def withLocalProperties[T](f: => T): T = {
     try {
       spark.sparkContext.setJobGroup(statementId, statement, forceCancel)
+      spark.sparkContext.setLocalProperty("kyuubi.statement.id", statementId)
```
yaooqinn, line 125: do we need to clean it in the `finally` block？

yaooqinn:
Maybe we can introduce a companion object to define the key
```scala
object ExecuteStatement {
  final val KYUUBI_STATEMENT_ID_KEY = "kyuubi.statement.id"
}

File: externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala [hunk 2]
```diff
@@ -122,16 +122,17 @@ class ExecuteStatement(
   private def withLocalProperties[T](f: => T): T = {
     try {
       spark.sparkContext.setJobGroup(statementId, statement, forceCancel)
-      spark.sparkContext.setLocalProperty("kyuubi.statement.id", statementId)
+      spark.sparkContext.setLocalProperty(ExecutionStatement.KYUUBI_STATEMENT_ID_KEY, statementId)
```
yaooqinn, line 125: use `KYUUBI_STATEMENT_ID_KEY` for better readability

yaooqinn (approved):
LGTM

yaooqinn:
thanks, merged to master

PR #794, opened 2021-07-13
Refactor Kyuubi kubernetes block cleaner

ulysses-you:

* fix bug as a blockMgr dir is created with no file at the begining.
* simplify code
* simplify log
* add test

the release directory structure
image


ulysses-you:
cc @zwangsheng @yaooqinn

tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
```diff
@@ -0,0 +1,205 @@
...
+      info(s"start all clean job")
+      val startTime = System.currentTimeMillis()
+      val hasFinished = new CountDownLatch(cacheDirs.length)
+      cacheDirs.foreach { dir =>
+        threadPool.execute(() => {
+          doCleanJob(dir)
+          hasFinished.countDown()
+        })
+      }
+      hasFinished.await()
```
zwangsheng, line 186: Here waiting for all the threads finsh jobs， means if usedtime greater than scheduleInterval, we can not start next cleaner on time.

yaooqinn (dismissed):

tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
ulysses-you: yes, see the warn log. I think it'ok and it's no meaning if schedule interval is too small.

zwangsheng: I got the idea if two thread cleaner one specific dir may trigger thread problem.LGTM.

zwangsheng (approved):

yaooqinn (dismissed):

yaooqinn (approved):

zwangsheng (approved):

ulysses-you:

PR #797, opened 2021-07-14
[INFRA] Disable branches protection

pan3793:

The branch protection breaks our pr merge tool

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -34,13 +34,6 @@ github:
     squash: true
     merge: false
     rebase: false
-  protected_branches:
-    master:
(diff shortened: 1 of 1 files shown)


yaooqinn (approved):

yaooqinn:
```

GitHub Protected Branches has set requires signature setting on branch 'master' to 'False'

With regards,
ASF Infra.
```

Ideally, we shall receive a message in our private mailing list like the above from the INFRA team. But it seems that this is not triggered but will be triggered if we send another PR to disable/enable the Protected Branches

Issue #786, opened 2021-07-13
Engine load balancing strategy


wForget:
### 1. Describe the feature

The last engine is acquired by default. Is there any plan to support other engine load balancing strategies?

Such as load balancing according to the current number of sessions of the engine.

ulysses-you:
Good idea, actually we have some discussion about this internal.

Maybe we can have something like engine pool to provide the ability of load balancing. And about the strategies, we can support `round robin`, `session/statment-number-weight-based` or other user defined strategy which is used the interface provided by this framework.

As you konw, Kyuubi is stateless in a cluster so it should be considered first how to share the engine pool status between several Kyuubi services. So we need more details and  some design of this.

It's also very helpful if you can share some use case of this feature.

yaooqinn:
For engine, we'd call it service discovery, not load balancing. The engine discovery logic is better to bind with client requests, a.k.a, JDBC connections, as an engine is stateful with cache tables, udf, etc. We already have implemented 3 share levels, CONNECTION/USER/SERVER for different kinds of usage scenarios.
- For CONNECTION mode, the engine and all states are within a JDBC connection.
- For USER mode, the engine and all states might/could share across all JDBC connections of a user with a `subDomain` if any.

For `The last engine is acquired by default`, it's a potential bug since `1.2.0-rc`s as a distributed lock for engine creation. When an engine is created but killed forcefully and externally, the service uri of this engine will not get cleared. And another thread might get a malformed engine after it gains the lock. For this part, I think we need to introduce a retry and recreation of engine

wForget:
Hi @ulysses-you @yaooqinn 
A simple implementation of this feature in #802 , can you help me review it and give some suggestions?

PR #802, opened 2021-07-14
Enhance service discovery

wForget:

Enhance service discovery #786

Refer to the curator-x-discovery module of the curator project for ServiceDiscovery enhancement. 

Some Interfaces:

+ `InstanceProvider`: provide instance that meet some conditions
+ `ProviderStrategy`: select an instance from the instances provided by the InstanceProvider
+ `ProviderFilter`: used in InstanceProvider to filter some instances
+ `InstanceSerializer`: serialize and deserialize instance object
+ `ServiceDiscoveryClient`: service discovery client, used to obtain the final instance

Feature：
+ support version filter
+ support tag filter
+ support random strategy
+ support strategy based on session number
+ support configure core size

TODO：

+ statistics session number
+ persistence engine version, tags and other information
+ adapt to EngineRef

Test Case: kyuubi-ha/src/test/scala/org/apache/kyuubi/ha/client/v2/EngineServiceDiscoveryClientSuite.scala


kyuubi-ha/src/test/scala/org/apache/kyuubi/ha/client/v2/EngineServiceDiscoveryClientSuite.scala
```diff
@@ -0,0 +1,158 @@
...
+        .creatingParentsIfNeeded()
+        .withMode(PERSISTENT)
+        .forPath(ns)
+    } catch {
+      case _: NodeExistsException => // do nothing
+      case e: KeeperException =>
+        throw new KyuubiException(s"Failed to create namespace '$ns'", e)
+    }
+    val pathPrefix = ZKPaths.makePath(
+      engine.namespace,
```
yaooqinn, line 113: Currently, the EngineSpace(namespace) and serviceNode(engine instance) are 1 vs 1

yaooqinn: Then, the ` EngineSpace` could achieve the isolation purpose for a tenant.

```scala
  /**
   * The EngineSpace used to expose itself to the KyuubiServers in `serverSpace`
   *
   * For `CONNECTION` share level:
   *   /$serverSpace_CONNECTION/$user/$sessionId
   * For `USER` share level:
   *   /$serverSpace_USER/$user[/$subDomain]
   *
   */
```

yaooqinn: The issue I mentioned in #786 is that for now we can not ensure the `serviceNode` here is alive or not

wForget: > Currently, the EngineSpace(namespace) and serviceNode(engine instance) are 1 vs 1

It doesn't seem to be one-to-one, engin namespace like: `/kyuubi_USER/test`, instance path like `/kyuubi_USER/test/serviceUri=127.0.0.1:1234;version=1.3.0;sequence=`

wForget: > The issue I mentioned in #786 is that for now we can not ensure the serviceNode here is alive or not

Sorry, I haven't considered it. This PR is mainly to choose the more appropriate one for all engines under the same shared user.

yaooqinn: In this test, yes, it's not one-to-one. but in real-world usage, we have a distributed lock for engine creations at the server-side.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/engine/EngineInstance.scala
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.ha.v2.engine
+
+case class EngineInstance(
```
yaooqinn, line 20: Can we make a separate PR for this first?

kyuubi-ha/src/test/scala/org/apache/kyuubi/ha/client/v2/EngineServiceDiscoveryClientSuite.scala
wForget: I added a `CoreSizeStrategy` to create more instances.  When the number of engines does not reach `CoreSize`, do not return to the existing engine, but to create a new engine.

yaooqinn: I get your point. How about when the client-side uses `DBCP`? the connections will point to different engines, will this result in surprise?

yaooqinn:
why can we just use ` curator-x-discovery`?

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/engine/EngineProvider.scala
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+}
+
+class EngineProviderBuilder() extends InstanceProviderBuilder[EngineInstance] {
+
+  var namespace: String = _
+  var zkClient: CuratorFramework = _
+  var version: Option[String] = None
+  var tags: ArrayBuffer[String] = ArrayBuffer()
```
yaooqinn, line 71: why this both a var and a mutable buffer?

yaooqinn, line 71: why this both a var and a mutable buffer?

kyuubi-ha/src/test/scala/org/apache/kyuubi/ha/client/v2/EngineServiceDiscoveryClientSuite.scala
wForget: > I get your point. How about when the client-side uses DBCP? the connections will point to different engines, will this result in surprise?

Yes, this may be problematic, but we can use a unique tag to adapt it.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/ServiceInstance.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.ha.v2
+
+case class ServiceInstance(host: String, port: Int) {
```
yaooqinn, line 20: what's the relation between a ServiceInstance and EngineInstance？

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/engine/EngineProvider.scala
```diff
@@ -0,0 +1,118 @@
...
+  val filters: ArrayBuffer[ProviderFilter[EngineInstance]] = ArrayBuffer()
+
+  def addFilter(filter: ProviderFilter[EngineInstance]): Unit = {
+    filters += filter
+  }
+
+  override def getInstances(): Seq[EngineInstance] = {
+    val enginePaths = zkClient.getChildren().forPath(namespace)
+    enginePaths.asScala
+      .map(getEngineInstance(_))
```
yaooqinn, line 42: we don't try to get as much as available engine instances as we can but fail directly if any error occurs

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/engine/EngineInstance.scala
```diff
@@ -0,0 +1,31 @@
...
+    namespace: String,
+    nodeName: String,
+    host: String,
+    port: Int,
+    version: Option[String],
+    sessionNum: Int = 0,
```
yaooqinn, line 26: what is the EngineInstance. sessionNum related to? a client session?

kyuubi-ha/src/test/scala/org/apache/kyuubi/ha/client/v2/EngineServiceDiscoveryClientSuite.scala
yaooqinn: Does it allow only one engine for a unique tag?

yaooqinn: then the tag is a bit redundant with `[/$subDomain]`, we need to remove one of them

wForget: Sorry, it does not seem to be the case, it may also need to limit core size.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/engine/EngineProvider.scala
wForget: Sorry, I don't quite understand this reply. The `EngineProvider.getInstances` method is to get all available engines of a shared user, and then get the most appropriate engine through `ProviderStrategy`.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/engine/EngineProvider.scala
wForget: This is a problem, I will fix it.

kyuubi-ha/src/test/scala/org/apache/kyuubi/ha/client/v2/EngineServiceDiscoveryClientSuite.scala
yaooqinn: Can we minimize the PR scope and do these separately?

Maybe we can start by adding a pool size for engines, and randomly pick it. then we apply picking rules to it separately in a series of PRs

yaooqinn: another thing is that we don't want the engine discovery to be bind to zookeeper too much in the near future, we are seeking a more light way to do this

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/v2/engine/EngineProvider.scala
yaooqinn: you mention in getEngineInstance that it will throw exceptions when can not parse the data to an engine instance

kyuubi-ha/src/test/scala/org/apache/kyuubi/ha/client/v2/EngineServiceDiscoveryClientSuite.scala
wForget: OK, thank you. I will minimize it and implement it gradually.

wForget:
needs to be improved, close it.

Issue #283, opened 2021-01-14
Error cancelling EXECUTE_STATEMENT: Read a negative frame size (-2147418110)


yaooqinn:
**Describe the bug**

```logtalk
21/01/14 10:10:36 WARN service.FrontendService: Error cancelling operation:
org.apache.kyuubi.KyuubiSQLException: Error cancelling EXECUTE_STATEMENT: Read a negative frame size (-2147418110)!
	at org.apache.kyuubi.KyuubiSQLException$.apply(KyuubiSQLException.scala:49)
	at org.apache.kyuubi.operation.KyuubiOperation$$anonfun$onError$1.applyOrElse(KyuubiOperation.scala:54)
	at org.apache.kyuubi.operation.KyuubiOperation$$anonfun$onError$1.applyOrElse(KyuubiOperation.scala:44)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:38)
	at org.apache.kyuubi.operation.KyuubiOperation.cancel(KyuubiOperation.scala:82)
	at org.apache.kyuubi.operation.OperationManager.cancelOperation(OperationManager.scala:96)
	at org.apache.kyuubi.session.AbstractSession.$anonfun$cancelOperation$1(AbstractSession.scala:189)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at org.apache.kyuubi.session.AbstractSession.withAcquireRelease(AbstractSession.scala:74)
	at org.apache.kyuubi.session.AbstractSession.cancelOperation(AbstractSession.scala:189)
	at org.apache.kyuubi.service.AbstractBackendService.cancelOperation(AbstractBackendService.scala:149)
	at org.apache.kyuubi.service.FrontendService.CancelOperation(FrontendService.scala:420)
	at org.apache.hive.service.rpc.thrift.TCLIService$Processor$CancelOperation.getResult(TCLIService.java:1657)
	at org.apache.hive.service.rpc.thrift.TCLIService$Processor$CancelOperation.getResult(TCLIService.java:1642)
	at org.apache.thrift.ProcessFunction.process(ProcessFunction.java:39)
	at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:39)
	at org.apache.kyuubi.service.authentication.HadoopThriftAuthBridgeServer$TUGIAssumingProcessor.process(HadoopThriftAuthBridgeServer.scala:163)
[...]

PR #819, opened 2021-07-16
[KYUUBI #283] Add a sync client to ensure all requests sent to engine within a session are in order

yaooqinn:

Add a sync client to ensure all requests sent to engine within a session are in order

### DIFF
--- a/kyuubi-main/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
+++ b/kyuubi-main/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
@@ -1,0 +1,202 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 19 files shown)


pan3793 (approved):

kyuubi-main/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -0,0 +1,201 @@
...
+   * Lock every rpc call to send them sequentially
+   */
+  private def withLockAcquired[T](block: => T): T = {
+    try {
+      lock.lock()
+      if (!protocol.getTransport.isOpen) {
+        throw KyuubiSQLException.connectionDoesNotExist()
+      }
+      block
+    } finally lock.unlock()
```
pan3793, line 48: seems the explicit `ReentrantLock` can be simplified by `synchronized { }` here.

yaooqinn:
thanks, merged to master

Issue #817, opened 2021-07-16
Fix KyuubiDriverSuite UT failure because of NPE


yanghua:
### 1. Describe the bug

Run `KyuubiDriverSuite` in my local IDE, but failed because of NPE(missing kyuubi-version-info.properties).

### 2. Environments

- Mac OS X 10.14.5
- Intellij IDEA 19.2
- Java 1.8

#### 2.1 Versions

- Kyuubi
  - [x] 1.13.0(master)
  - [ ] 1.0.3
  - [ ] 0.8.x and earlier 

- Spark
  - [ ] 3.1.x
  - [ ] 3.0.x
  - [x] 2.4.x and earlier 

#### 2.2 Kyuubi Configurations

none(local UT)

#### 2.3 Spark Configurations

none(local UT)

### 4. Additional context

```
An exception or error caused a run to abort: Failed to initialize frontend service on /127.0.0.1:0. 
org.apache.kyuubi.KyuubiException: Failed to initialize frontend service on /127.0.0.1:0.
	at org.apache.kyuubi.service.FrontendService.initialize(FrontendService.scala:102)
	at org.apache.kyuubi.service.CompositeService.$anonfun$initialize$1(CompositeService.scala:40)
	at org.apache.kyuubi.service.CompositeService.$anonfun$initialize$1$adapted(CompositeService.scala:40)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at org.apache.kyuubi.service.CompositeService.initialize(CompositeService.scala:40)
	at org.apache.kyuubi.service.Serverable.initialize(Serverable.scala:44)
	at org.apache.kyuubi.engine.spark.SparkSQLEngine.initialize(SparkSQLEngine.scala:50)
	at org.apache.kyuubi.engine.spark.SparkSQLEngine$.startEngine(SparkSQLEngine.scala:108)
	at org.apache.kyuubi.engine.spark.WithSparkSQLEngine.startSparkEngine(WithSparkSQLEngine.scala:57)
	at org.apache.kyuubi.engine.spark.WithSparkSQLEngine.startSparkEngine$(WithSparkSQLEngine.scala:39)
[...]

PR #820, opened 2021-07-17
[KYUUBI 817] Fix KyuubiDriverSuite UT failure because of NPE

yanghua:

See issue #817 , when we pull the project and do not invoke any build action, e.g. `mvn package` or `mvn generate-resources`, we would miss the `kyuubi-version-info.properties` file, it caused the NPE. 

As discussed under issue #817 , this PR tries to check it and enhance the exception message to make it more friendly to newbie developers.

- [ ] Verified in the local env and no need to add test cases.


yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #824, opened 2021-07-18
[GA] Remove spark-bundled binary distributions in GA release

### LINKED ISSUE #798: Make distribution with --spark-provided as default an only in release.yml of GA

pan3793:

close #798

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,14 +34,6 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: '1.8'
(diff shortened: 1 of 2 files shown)


yaooqinn:
BTW, the release is still invalid according to apache policy after this change

yaooqinn (approved):

pan3793:
I see, there are still lots of works to do to meet apache policies. The biggest one I think is that apache cares about the source release but we only have binary release now.

yaooqinn:
spark-*-bin-*.tgz rm this in license

yaooqinn:
thanks, merged to master

PR #841, opened 2021-07-20
Kyuubi Spark External Shuffle Service

yaooqinn:
#### Kyuubi Spark External Shuffle Service

This enables the missing feature of Dynamic Resource Allocation(a.k.a, DRA, DA) for running Spark on Kubernetes, which provides the auto-scaling capability for Spark application, especially downscaling.

Although there is a feature called shuffle tracking to track the shuffle data and tell Spark to remove executor while the data exceeds the time to live, these removals are coarse-grained.

##### Prerequisites

- Everything you need to know about Spark on Kubernetes
- Running Spark on Kubernetes apps with hostNetwork enabled
- Running this module as a DaemonSet with hostNetwork enabled
- Self-managed shuffle data cleaner as the state of apps will not be posted to this service
  - See: tools/spark-block-cleaner/kubernetes/spark-block-cleaner.yml

##### Usage

Check `tools/kyuubi-shuffle-service/kubernetes/example.yaml` for more information.

#### cmd
```
bin/spark-sql 
```

#### config
```
spark.kubernetes.file.upload.path=./
spark.master=k8s://https://kubernetes.docker.internal:6443
spark.kubernetes.container.image=yaooqinn/spark:v20210619
spark.kubernetes.context=docker-for-desktop_1
spark.dynamicAllocation.enabled=true
spark.dynamicAllocation.minExecutors=2
spark.dynamicAllocation.maxExecutors=10
spark.driver.memory 5g
spark.kubernetes.executor.podNamePrefix=sparksql
spark.shuffle.service.enabled=true

spark.kubernetes.executor.podTemplateFile=/Users/kentyao/Downloads/executor-template.yml
spark.kubernetes.executor.volumes.hostPath.spark-local-dir-1.mount.path=/opt/spark/data1

spark.kubernetes.executor.volumes.hostPath.spark-local-dir-1.options.path=/Users/kentyao/Downloads/data/1

spark.kubernetes.executor.volumes.hostPath.spark-local-dir-2.mount.path=/opt/spark/data2

spark.kubernetes.executor.volumes.hostPath.spark-local-dir-2.options.path=/Users/kentyao/Downloads/data/2
spark.blockManager.port=0
```

#### query
```
SELECT /*+ REPARTITION */ 1 as pid, id from range(1, 10000, 1, 500);
```

### DIFF
--- a/tools/kyuubi-shuffle-service/kubernetes/example.yaml
+++ b/tools/kyuubi-shuffle-service/kubernetes/example.yaml
@@ -34,7 +34,7 @@ spec:
       containers:
         # the image built base on ./docker/Dockerfile
         - image: yaooqinn/kyuubi-shuffle-service:latest
-          name: exterbal-shuffle-server
+          name: external-shuffle-server
           resources:
             limits:
               memory: 1024Mi
--- a/tools/kyuubi-shuffle-service/kubernetes/example.yaml
+++ b/tools/kyuubi-shuffle-service/kubernetes/example.yaml
@@ -1,0 +1,81 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# kubectl apply -f tools/kyuubi-shuffle-service/kubernetes/example.yaml
+
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kyuubi-shuffle-service
+  namespace: default
+spec:
+  selector:
+    matchLabels:
+      name: kyuubi-shuffle-service
+  template:
+    metadata:
+      labels:
+        name: kyuubi-shuffle-service
+    spec:
+      containers:
+        # the image built base on ./docker/Dockerfile
+        - image: yaooqinn/kyuubi-shuffle-service:latest
+          name: exterbal-shuffle-server
+          resources:
+            limits:
+              memory: 1024Mi
+            requests:
+              cpu: 100m
+              memory: 512Mi
+          volumeMounts:
+            - name: spark-local-dir-1
+              mountPath: /opt/spark/data1
+            - name: spark-local-dir-2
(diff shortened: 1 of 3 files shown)


tools/kyuubi-shuffle-service/kubernetes/example.yaml
```diff
@@ -0,0 +1,81 @@
...
+              mountPath: /opt/spark/data2
+          env:
+            # Change this in case of OOM based on you spec.containers.resources.* for
+            # the heap memory of shuffle service proc
+            - name: SPARK_DAEMON_MEMORY
+              value: 512M
+            # SPARK_SHUFFLE_OPTS allows us to pass spark configurations to the shuffle service
+            # spark.local.dir points to spec.containers.volumeMounts.*, shall be same as Spark
+            # app side.
+            # spark.shuffle.cleaner.interval=30s, this currently not work because only can a
```
yaooqinn, line 57: If an `ExternalBlockStoreClient` can also post ShuffleServiceHeartbea, we can reach much more fine grained disk/shuffle data management

yaooqinn:
cc @holdenk @dongjoon-hyun to see if we have an opportunity to make the current external shuffle service work for all cluster managers, including k8s, standalone, etc.

holdenk:
So I don't know anything about this project but certainly does seem like it could be interesting. This provides an alternative shuffle service for Spark to use on Kube? Is there a design doc? I notice hostNetwork has to be set to true which is something I'm generally speaking not comfortable with for security reasons, is that a temporary design decision or something more permanent?

yaooqinn:
Thanks @holdenk 

> So I don't know anything about this project but certainly does seem like it could be interesting. 

Do worry about the project itself. The project relies on Spark who can achieve better scalability to use computing resources more effectively on Kubernetes. And this is the motivation of this PR.

> This provides an alternative shuffle service for Spark to use on Kube? Is there a design doc?

Yes, this provides an external shuffle service for Spark to use on Kube. The current implementation is codeless as you can see in this PR. We only need to write a Dockerfile(3 LOC only) for the shuffle service(`org.apache.spark.deploy.ExternalShuffleService`) endpoint based on the official Spark image and then deploy it on k8s as a DaemonSet. This can work with official releases(verified 3.1.2) without any modification. The only thing a bit tricky here is that users need to enable the `hostNetwork` to let the client successfully establish the connection `LOCAL` shuffle server during an executor initialization.

> I notice hostNetwork has to be set to true which is something I'm generally speaking not comfortable with for security reasons, is that a temporary design decision or something more permanent?

I agree that using hostNetwork is not secure and incomprehensive for this feature. I will try to figure out a solution with the Kubernetes pod network only, maybe in the next few weeks with a SPIP or a POC PR in the Spark community.

holdenk:
I like the simplicity of this approach, but I think there are some pretty serious limitations we should think through. It would let us scale up and down executor pods but not the shuffle services and we'd probably end up doing a lot of remote fetches. Can you give me some more context around your plans here? Have you looked at the current work on introducing the new shuffle APIs for Kube in Spark?

Is the goal just to support dynamic scaling? If so have you tried enabling migrations?

jerqi:
@yaooqinn Is there any progress or further discuss about K8S ESS?

PR #857, opened 2021-07-22
Support init SQL for each session

pan3793:


pan3793:
@turboFei @yaooqinn would you please take a look? thx

docs/deployment/settings.md
```diff
@@ -147,6 +147,7 @@ kyuubi\.engine<br>\.deregister\.exception<br>\.messages|<div style='width: 65pt;
...
 kyuubi\.engine<br>\.deregister\.job\.max<br>\.failures|<div style='width: 65pt;word-wrap: break-word;white-space: normal'>4</div>|<div style='width: 170pt;word-wrap: break-word;white-space: normal'>Number of failures of job before deregistering the engine.</div>|<div style='width: 30pt'>int</div>|<div style='width: 20pt'>1.2.0</div>
 kyuubi\.engine<br>\.initialize\.sql|<div style='width: 65pt;word-wrap: break-word;white-space: normal'>SHOW DATABASES</div>|<div style='width: 170pt;word-wrap: break-word;white-space: normal'>SemiColon-separated list of SQL statements to be initialized in the newly created engine before queries.</div>|<div style='width: 30pt'>string</div>|<div style='width: 20pt'>1.2.0</div>
+kyuubi\.engine\.session<br>\.initialize\.sql|<div style='width: 65pt;word-wrap: break-word;white-space: normal'>SHOW DATABASES</div>|<div style='width: 170pt;word-wrap: break-word;white-space: normal'>SemiColon-separated list of SQL statements to be initialized in the newly created engine session before queries.</div>|<div style='width: 30pt'>string</div>|<div style='width: 20pt'>1.3.0</div>
```
yaooqinn, line 150: SemiColon-separated seems not quite friendly to jdbc url

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -99,7 +99,10 @@ object SparkSQLEngine extends Logging {
     }
 
     val session = SparkSession.builder.config(sparkConf).getOrCreate
-    kyuubiConf.get(KyuubiConf.ENGINE_INITIALIZE_SQL).split(";").foreach(session.sql(_).show)
+    kyuubiConf.get(KyuubiConf.ENGINE_INITIALIZE_SQL).split(";").foreach { sql =>
+      info(s"execute engine init sql: $sql")
```
yaooqinn, line 103: Execute engine initializing SQL

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -69,6 +71,14 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
       setSession(handle, sessionImpl)
       info(s"$user's session with $handle is opened, current opening sessions" +
       s" $getOpenSessionCount")
+
+      if (!this.conf.get(ENGINE_SINGLE_SPARK_SESSION)) {
+        kyuubiConf.get(KyuubiConf.ENGINE_SESSION_INITIALIZE_SQL).split(";").foreach { sql =>
+          info(s"execute session init sql: $sql")
```
yaooqinn, line 77: ditto

docs/deployment/settings.md
pan3793: Any suggestion? I don't know if any other RDMS support init sql in Jdbc url.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
pan3793: updated

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
pan3793: updated

docs/deployment/settings.md
yaooqinn: We shall notify user that it a system only configuation

yaooqinn (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -608,11 +608,21 @@ object KyuubiConf {
...
     .createWithDefault("SHOW DATABASES")
 
+  val ENGINE_SESSION_INITIALIZE_SQL: ConfigEntry[String] =
+    buildConf("engine.session.initialize.sql")
+      .doc("SemiColon-separated list of SQL statements to be initialized in the newly created " +
+        "engine session before queries. This configuration can not be used in JDBC url due to " +
+        "the limitation of Beeline/JDBC driver.")
+      .version("1.3.0")
+      .stringConf
+      .createWithDefault("SHOW DATABASES")
```
turboFei, line 624: should we use `.fallbackConf(ENGINE_INITIALIZE_SQL)`?

turboFei, line 624: should we use `.fallbackConf(ENGINE_INITIALIZE_SQL)`?

kyuubi-main/src/test/scala/org/apache/kyuubi/engine/spark/InitializeSQLSuite.scala
```diff
@@ -40,12 +45,31 @@ class InitializeSQLSuite extends WithKyuubiServer with JDBCTestUtils {
...
+      assert(!result.next())
+    }
+    // new session
+    withJdbcStatement() { statement =>
+      val result = statement.executeQuery("SELECT COUNT(*) FROM INIT_DB.test WHERE a = 2")
+      assert(result.next())
+      // use great than or equals to support concurrent test
+      assert(result.getLong(1) >= currentSessionCnt + 1)
+      assert(!result.next())
+    }
```
turboFei, line 71: should we add a test for single spark session mode?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
pan3793: i think there are two things, fallback doesn't make sense to me.

turboFei: ok

kyuubi-main/src/test/scala/org/apache/kyuubi/engine/spark/InitializeSQLSuite.scala
pan3793: yes, will update

pan3793: updated, please review again

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -69,6 +75,18 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
       setSession(handle, sessionImpl)
       info(s"$user's session with $handle is opened, current opening sessions" +
       s" $getOpenSessionCount")
+
+      if (singleSparkSessionInitialized.compareAndSet(false, true)
```
yaooqinn, line 79: 1. We need to move this up to L74 or higher.

2. nit:
```scala
val singleSparkSession = this.conf.get(ENGINE_SINGLE_SPARK_SESSION)
```
3. we can execute `ENGINE_SESSION_INITIALIZE_SQL` once when the engine startup instead of hacking here

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -69,6 +75,18 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
...
+        || !this.conf.get(ENGINE_SINGLE_SPARK_SESSION)) {
+        kyuubiConf.get(KyuubiConf.ENGINE_SESSION_INITIALIZE_SQL)
```
yaooqinn, line 81: this.conf is kyuubiConf

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
pan3793: Addressed 1 and 2.
Have some thoughts on 3, one difference between single session mode and isolate session mode is that the former uses the engine startup session but the latter does not. Is it necessary also to execute `ENGINE_SESSION_INITIALIZE_SQL` following engine startup for the engine startup session?

yaooqinn: I think it's ok, as the main purpose of init sqls is always to bring up some external stuff alive or visible

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -59,6 +63,17 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
         spark.newSession()
       }
 
+      if (singleSparkSessionInitialized.compareAndSet(false, true)
```
yaooqinn, line 66: if singleSparkSessionInitialized then we need always updating?

pan3793: no, `compareAndSet(expect = false, update = true)` only return `true` in first round, but the if statement break the semantic of `singleSparkSessionInitialized ` here, it should be
```
if (!this.conf.get(ENGINE_SINGLE_SPARK_SESSION)) || singleSparkSessionInitialized.compareAndSet(false, true))
```
anyway, removed `singleSparkSessionInitialized` and partial move the init logic to engine startup

ulysses-you:
use `.show` seems not gentle, how about `.count` ?

pan3793:
> use `.show` seems not gentle, how about `.count` ?

make sense

ulysses-you (approved):

pan3793:
> > use `.show` seems not gentle, how about `.count` ?
> 
> make sense

`count` introduced new `job`, which broken UTs depends on JobId, reverted.

yaooqinn:
thanks, merged to master

PR #864, opened 2021-07-26
Block cleaner should catch NonFatal exception

ulysses-you:

The daemonset may encounter some pontential issue during long-time running. It should be robust to catch `NonFatal` exception with it.

### DIFF
--- a/tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
+++ b/tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
@@ -21,6 +21,8 @@ import java.io.File
 import java.nio.file.{Files, Paths}
 import java.util.concurrent.{CountDownLatch, Executors}
 
+import scala.util.control.NonFatal
+
 import org.apache.log4j.PropertyConfigurator
 
 import org.apache.kyuubi.Logging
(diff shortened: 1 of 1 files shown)


ulysses-you:
cc @yaooqinn @zwangsheng

tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
```diff
@@ -185,8 +187,14 @@ object KubernetesSparkBlockCleaner extends Logging {
       val hasFinished = new CountDownLatch(cacheDirs.length)
       cacheDirs.foreach { dir =>
         threadPool.execute(() => {
-          doCleanJob(dir)
-          hasFinished.countDown()
+          try {
+            doCleanJob(dir)
+          } catch {
+            case NonFatal(e) =>
+              error("failed to clean", e)
```
yaooqinn, line 194: print the dir too?

ulysses-you: added

yaooqinn (approved):

zwangsheng:
Can we figure out where cause this NonFatal Exception ？

ulysses-you:
> Can we figure out where cause this NonFatal Exception ？

@zwangsheng I see a NPE error in `doClean` code and seems `blockManagerDir` become a null value. Not sure how this happened, but we need catch this exception.

Issue #859, opened 2021-07-22
Add kubernetes  configmap


hddong:
### 1. Describe the feature
Now, we need to rebuild docker image when config changed. Use `configmap` to save kyuubi config file, then we can change config by edit configmap and restart pod.

yaooqinn:
make sense

PR #866, opened 2021-07-26
[KYUUBI #859] Add kubernetes configmap

hddong:

Now, we need to rebuild docker image when config changed. Use configmap to save kyuubi config file, then we can change config by edit configmap and restart pod.

- [X] [Run test](https://kyuubi.readthedocs.io/en/latest/tools/testing.html#running-tests) locally before make a pull request

### DIFF
--- a/docker/kyuubi-pod.yaml
+++ b/docker/kyuubi-pod.yaml
@@ -59,7 +59,7 @@ spec:
           protocol: TCP
       volumeMounts:
         - name: kyuubi-defaults
-          mountPath: /opt/kyuubi/conf/
+          mountPath: /opt/kyuubi/conf
   volumes:
     - name: kyuubi-defaults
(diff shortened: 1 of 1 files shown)


docker/kyuubi-pod.yaml
```diff
@@ -49,3 +57,10 @@ spec:
         - name: frontend-port
           containerPort: 10009
           protocol: TCP
+      volumeMounts:
+        - name: kyuubi-defaults
+          mountPath: /opt/kyuubi/conf/
```
yaooqinn, line 62: nit: /opt/kyuubi/conf

yaooqinn (approved):

ulysses-you (approved):

yaooqinn:
thanks, merged to master

PR #868, opened 2021-07-27
Avoid calling a stopped SparkContext

yaooqinn:

We shall avoid calling cancelJobGroup if the sc is already stopped.

We shall stop the engine to stop serving before we stop the sc during the shutdown hook

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -168,12 +168,16 @@ private[kyuubi] object Utils extends Logging {
     System.getProperty(IS_TESTING.key) != null
   }
 
+  val DEFAULT_SHUTDOWN_PRIORITY = 100
+  val SERVER_SHUTDOWN_PRIORITY = 75
+  val SPARK_CONTEXT_SHUTDOWN_PRIORITY = 50
+
   /**
    * Add some operations that you want into ShutdownHook
    * @param hook
    * @param priority: 0~100
    */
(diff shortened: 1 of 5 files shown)


yaooqinn:
cc @zhang1002 @ulysses-you @pan3793 @turboFei

yaooqinn:
cc @hammertank

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -168,12 +168,16 @@ private[kyuubi] object Utils extends Logging {
     System.getProperty(IS_TESTING.key) != null
   }
 
+  val DEFAULT_SHUTDOWN_PRIORITY = 100
+  val SERVER_SHUTDOWN_PRIORITY = 75
+  val SPARK_CONTEXT_SHUTDOWN_PRIORITY = 50
```
ulysses-you, line 173: We can add some comment that this priority depends on `org.apache.spark.util.ShutdownHookManager.SPARK_CONTEXT_SHUTDOWN_PRIORITY`

ulysses-you (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
pan3793: a small or big one has a higher priority?

yaooqinn: big

pan3793 (approved):

yaooqinn:
thanks, merged to master

PR #875, opened 2021-07-30
Enhance kubernetes cluster diagnostics

ulysses-you:

Add kubernetes context and namespace info if Spark is running on k8s.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
@@ -19,13 +19,39 @@ package org.apache.kyuubi.engine.spark
 
 import java.time.{Instant, LocalDateTime, ZoneId}
 
+import org.apache.spark.SparkContext
 import org.apache.spark.sql.SparkSession
 
 object KyuubiSparkUtil {
+  // Regular expression for connecting to kubernetes clusters
+  val KUBERNETES_REGEX = """k8s://(.*)""".r
 
   lazy val diagnostics: String = {
-    val spark = SparkSession.active
-    val sc = spark.sparkContext
+    val sc = SparkSession.active.sparkContext
+
+    sc.master match {
+      case KUBERNETES_REGEX(_) => getKubernetesClusterInfo(sc)
+      case _ => defaultClusterInfo(sc)
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -19,13 +19,39 @@ package org.apache.kyuubi.engine.spark
...
+
+  def getKubernetesClusterInfo(sc: SparkContext): String = {
+    // scalastyle:off line.size.limit
+    s"""
+       |           Spark application name: ${sc.appName}
+       |                 application ID:  ${sc.applicationId}
+       |                 application web UI: ${sc.uiWebUrl}
+       |                 master: ${sc.master}
+       |                 deploy mode: ${sc.deployMode}
+       |                 context: ${sc.getConf.getOption("spark.kubernetes.context").getOrElse("default")}
```
yaooqinn, line 46: get("spark.kubernetes.context", "default")

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -19,13 +19,39 @@ package org.apache.kyuubi.engine.spark
...
+       |                 namespace: ${sc.getConf.getOption("spark.kubernetes.namespace").getOrElse("default")}
```
yaooqinn, line 47: ditto

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -19,13 +19,39 @@ package org.apache.kyuubi.engine.spark
...
+       |                 version: ${sc.version}
+       |           Start time: ${LocalDateTime.ofInstant(Instant.ofEpochMilli(sc.startTime), ZoneId.systemDefault)}
+       |           User: ${sc.sparkUser}""".stripMargin
+    // scalastyle:on line.size.limit
+  }
+
+  def defaultClusterInfo(sc: SparkContext): String = {
```
yaooqinn, line 54: can we reduce the duplicated code?

PR #887, opened 2021-08-02
[GA] Daily publish snapshot

pan3793:

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -34,4 +34,6 @@ jobs:
       env:
         ASF_USERNAME: ${{ secrets.NEXUS_USER }}
         ASF_PASSWORD: ${{ secrets.NEXUS_PW }}
-      run: ./build/mvn clean deploy -DskipTests -Pspark-provided -s ./build/release/asf-settings.xml
+      run: |
+        rm -rf ~/.m2/repository/org/apache/kyuubi
+        ./build/mvn clean deploy -DskipTests -Pspark-provided -s ./build/release/asf-settings.xml
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -1,0 +1,37 @@
+name: Publish Snapshot
+
+on:
+  schedule:
+  - cron: '0 0 * * *'
+
+jobs:
+  publish-snapshot:
+    if: ${{ startsWith(github.repository, 'apache/') }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        branch:
+          - master
+          - branch-1.3
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@master
+      with:
+        ref: ${{ matrix.branch }}
+    - name: Cache Maven local repository
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: snapshot-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          snapshot-maven-
+    - name: Install Java 8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    - name: Publish snapshot
+      env:
+        ASF_USERNAME: ${{ secrets.NEXUS_USER }}
+        ASF_PASSWORD: ${{ secrets.NEXUS_PW }}
+      run: ./build/mvn clean deploy -DskipTests -Pspark-provided -s ./build/release/asf-settings.xml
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  publish-snapshot:
+    if: ${{ startsWith(github.repository, 'apache/') }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        branch:
+          - master
+          - branch-1.3
```
ulysses-you, line 16: do we need `branch-1.3` snapshot ?

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,37 @@
...
+          snapshot-maven-
+    - name: Install Java 8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    - name: Publish snapshot
+      env:
+        ASF_USERNAME: ${{ secrets.NEXUS_USER }}
+        ASF_PASSWORD: ${{ secrets.NEXUS_PW }}
+      run: ./build/mvn clean deploy -DskipTests -Pspark-provided -s ./build/release/asf-settings.xml
```
ulysses-you, line 37: `-Pkyuubi-extension-spark-3-1` ?

ulysses-you: not need to specify `-Papache-release` ?

ulysses-you (commented):
thank you @pan3793  for the great work

.github/workflows/publish-snapshot.yml
pan3793: Yes, plan to cut `branch-1.3` soon

.github/workflows/publish-snapshot.yml
pan3793: > `-Pkyuubi-extension-spark-3-1` ?

@yaooqinn Should we publish `kyuubi-extension-spark-3-1` to maven central ?

pan3793: > not need to specify `-Papache-release`?

There is no GPG info in GA, I'm planing use `-Papache-release` in RC release.

pan3793:
BTW, what do you think of renaming `kyuubi-hive-jdbc` to `kyuubi-jdbc`?  @yaooqinn @ulysses-you

yaooqinn:
we may add other protocol support

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,37 @@
+name: Publish Snapshot
+
+on:
+  schedule:
+  - cron: '0 0 * * *'
+
+jobs:
+  publish-snapshot:
+    if: ${{ startsWith(github.repository, 'apache/') }}
+    runs-on: ubuntu-latest
```
yaooqinn, line 10: shall we use the latest LTS？

pan3793:
> we may add other protocol support

got it.

.github/workflows/publish-snapshot.yml
yaooqinn, line 37: shall we rm all the kyuubi artifacts from cache？

.github/workflows/publish-snapshot.yml
pan3793: Make sense, I'm going to switch all image to `ubuntu-20.04`. BTW, Apache Spark mix uses `ubuntu-20.04` and `ubuntu-latest`.

.github/workflows/publish-snapshot.yml
pan3793: Updated.

yaooqinn: I’AM OK to publish it as long as the snapshot has nothing to do with final releases

pan3793: Just realized that Spark 3.2 will support both Scala 2.12 and 2.13. But `kyuubi-extension-spark-3-1` only build against Scala 2.12, and the package name lack the `_2.12` suffix, it may confuse users who want to use it with Spark 3.2. (I guess only few users will run Spark 3.2 with Scala 2.13 in production?)

ulysses-you: > There is no GPG info in GA, I'm planing use -Papache-release in RC release.

ah I see, thanks

yaooqinn (approved):

pan3793:
Thanks for reviewing. I'm going to merge it. Then will send a following PR to add `_2.12` suffix in Scala modules and also enable `-Pkyuubi-extension-spark-3-1` in daily snapshot publish.

PR #892, opened 2021-08-04
Make JsonEventLogger compatible with HDFS-like FileSystems

yaooqinn:

FSDataOutputStream is used for this

### DIFF
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -68,6 +68,19 @@
             <scope>provided</scope>
         </dependency>
 
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client-runtime</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
         <dependency>
             <groupId>org.apache.kyuubi</groupId>
             <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
@@ -130,18 +143,6 @@
         <dependency>
             <groupId>org.apache.hive</groupId>
             <artifactId>hive-llap-common</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-client-api</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-client-runtime</artifactId>
             <scope>test</scope>
         </dependency>
 
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -68,16 +68,9 @@
             <scope>provided</scope>
         </dependency>
 
-
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client-api</artifactId>
-            <scope>provided</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-client-runtime</artifactId>
             <scope>provided</scope>
         </dependency>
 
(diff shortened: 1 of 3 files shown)


yaooqinn:
cc @pan3793 @ulysses-you @zhang1002

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -68,6 +68,19 @@
...
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client-runtime</artifactId>
+            <scope>provided</scope>
```
pan3793, line 81: keep `hadoop-client-runtime` in test scope

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -68,6 +68,19 @@
             <scope>provided</scope>
         </dependency>
 
+
```
pan3793, line 71: nit: keep one blank line

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
```diff
@@ -33,32 +36,51 @@ import org.apache.kyuubi.service.AbstractService
  * The {eventType} is based on core concepts of the Kyuubi systems, e.g. engine/session/statement
  * @param logName the engine id formed of appId + attemptId(if any)
  */
-class JsonEventLogger(logName: String)
+class JsonEventLogger(logName: String, hadoopConf: Configuration)
   extends AbstractService("JsonEventLogger") with EventLogger with Logging {
 
-  private var logRoot: Path = _
-  private val writers = new scala.collection.mutable.HashMap[String, PrintWriter]()
+  type Logger = (PrintWriter, Option[FSDataOutputStream])
```
pan3793, line 42: How about `type Logger = Either[PrintWriter, FSDataOutputStream]`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
```diff
@@ -33,32 +36,51 @@ import org.apache.kyuubi.service.AbstractService
...
+
+  private var logRoot: URI = _
+  private var fs: FileSystem = _
+  private val writers =
+    new scala.collection.mutable.HashMap[String, Logger]()
```
pan3793, line 47: `scala.collection.mutable.HashMap` => `mutable.HashMap`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
yaooqinn: PrintWriter is not optional

pan3793 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
```diff
@@ -33,32 +39,50 @@ import org.apache.kyuubi.service.AbstractService
...
+      val eventPath = new Path(new Path(logRoot), event.eventType)
+      FileSystem.mkdirs(fs, eventPath, JSON_LOG_DIR_PERM)
+      val logFile = new Path(eventPath, logName + ".json")
+      var hadoopDataStream: FSDataOutputStream = null
+      val rawStream = if (logFile.toUri.getScheme == "file") {
+        new FileOutputStream(logFile.toUri.getPath)
+      } else {
+        hadoopDataStream = fs.create(logFile)
+        hadoopDataStream
+      }
```
ulysses-you, line 62: why we need distinguish the scheme here, we can't just do this ?
```
val stream = fs.create(logFile)
val bStream = new BufferedOutputStream(stream)
```

yaooqinn: avoid hdfs bug

ulysses-you: Can we add that bug link here

Issue #897, opened 2021-08-05
Support stop Spark engine through web ui


ulysses-you:
### 1. Describe the feature
Enhance Spark web ui to allow user stop the engine.

### 2. Motivation
In **user** shared level, an engine could exist a long time then it's useful to provide an entrance to alow user stop the engine.

PR #898, opened 2021-08-05
[KYUUBI #897] Support stop Spark engine through web ui

ulysses-you:

Closes https://github.com/apache/incubator-kyuubi/issues/897

The following two images show the stop feature is enabled:
image
image

The following image show the stop feature is disabled:
image


ulysses-you:
cc @pan3793 @yaooqinn

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/ui/KyuubiUIUtils.scala
```diff
@@ -0,0 +1,78 @@
...
+      (basePath + path).stripSuffix("/")
+    }
+    val contextHandler = new ServletContextHandler
+    val holder = new ServletHolder(servlet)
+    contextHandler.setContextPath(prefixedPath)
+    contextHandler.addServlet(holder, "/")
+    contextHandler
+  }
+
+  def createRedirectHandler(
```
yaooqinn, line 43: Why not just call org.apache.spark.ui.JettyUtils. createRedirectHandler

ulysses-you: We also need to use reflect if we want to use `org.apache.spark.ui.JettyUtils.createRedirectHandler`, and I'm afriad Spark may change the code in future.

yaooqinn: this can certainly prevent  the impact of future spark code change？

ulysses-you: OK removed the forked code, only use reflect

pan3793 (changes requested):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/ui/EngineTab.scala
```diff
@@ -17,22 +17,49 @@
...
   engine.spark.sparkContext.ui.foreach { ui =>
     this.attachPage(EnginePage(this))
     ui.attachTab(this)
     Utils.addShutdownHook(() => ui.detachTab(this))
   }
+
+  engine.spark.sparkContext.ui.foreach { ui =>
+    try {
+      // Spark shade the jetty package so here we use reflect
+      Class.forName("org.apache.spark.ui.SparkUI")
```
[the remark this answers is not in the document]
ulysses-you: it's not about access, the reason using refection is Spark shade the jetty package. There is some compile error if we invoke that method directly.

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -50,6 +50,19 @@
...
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-server</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-core_${scala.binary.version}</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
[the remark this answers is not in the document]
ulysses-you: removed and it works. I copied some code before so added these dependencies, but now not needed.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -702,4 +702,11 @@ object KyuubiConf {
       .checkValue(_.toSet.subsetOf(Set("SPARK", "JSON", "JDBC", "CUSTOM")),
         "Unsupported event loggers")
       .createWithDefault(Nil)
+
+  val ENGINE_UI_STOP_ENABLED: ConfigEntry[Boolean] =
+    buildConf("engine.ui.stop.enabled")
+      .doc("When true, allows Kyuubi engine to be killed from the web UI.")
```
yaooqinn, line 708: the Spark Web UI

pan3793 (approved):

yaooqinn (approved):

pan3793:
Thanks, merging to master for v1.3.0.

PR #899, opened 2021-08-05
Avoid RejectedExecutionException if engine is shutdown

ulysses-you:

Engine can be shutdown but not stopped when an operation is submitted , then we might get such exception:

```
2021-08-05 14:19:33.787 WARN service.FrontendService: Error executing statement:
org.apache.kyuubi.KyuubiSQLException: Error submitting query in background, query rejected
	at org.apache.kyuubi.KyuubiSQLException$.apply(KyuubiSQLException.scala:68)
	at org.apache.kyuubi.engine.spark.operation.ExecuteStatement.runInternal(ExecuteStatement.scala:124)
	at org.apache.kyuubi.operation.AbstractOperation.run(AbstractOperation.scala:130)
	at org.apache.kyuubi.session.AbstractSession.runOperation(AbstractSession.scala:95)
	at org.apache.kyuubi.session.AbstractSession.$anonfun$executeStatement$1(AbstractSession.scala:123)
	at org.apache.kyuubi.session.AbstractSession.withAcquireRelease(AbstractSession.scala:78)
	at org.apache.kyuubi.session.AbstractSession.executeStatement(AbstractSession.scala:120)
	at org.apache.kyuubi.service.AbstractBackendService.executeStatement(AbstractBackendService.scala:61)
	at org.apache.kyuubi.service.FrontendService.ExecuteStatement(FrontendService.scala:256)
	at org.apache.hive.service.rpc.thrift.TCLIService$Processor$ExecuteStatement.getResult(TCLIService.java:1557)
	at org.apache.hive.service.rpc.thrift.TCLIService$Processor$ExecuteStatement.getResult(TCLIService.java:1542)
	at org.apache.thrift.ProcessFunction.process(ProcessFunction.java:38)
	at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:39)
	at org.apache.kyuubi.service.authentication.TSetIpAddressProcessor.process(TSetIpAddressProcessor.scala:36)
	at org.apache.thrift.server.TThreadPoolServer$WorkerProcess.run(TThreadPoolServer.java:310)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.FutureTask@1202ad6e rejected from java.util.concurrent.ThreadPoolExecutor@a50ca2b[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]

```

It's noisy to print this excption.

Pass CI

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -112,17 +112,21 @@ class ExecuteStatement(
         }
       }
 
+      val sparkSQLSessionManager = session.sessionManager
       try {
-        val sparkSQLSessionManager = session.sessionManager
         val backgroundHandle = sparkSQLSessionManager.submitBackgroundOperation(asyncOperation)
         setBackgroundHandle(backgroundHandle)
       } catch {
         case rejected: RejectedExecutionException =>
-          setState(OperationState.ERROR)
-          val ke = KyuubiSQLException("Error submitting query in background, query rejected",
-            rejected)
-          setOperationException(ke)
-          throw ke
+          if (sparkSQLSessionManager.isShutdown) {
+            // ignore this exception since the engine is shutdown
+          } else {
+            setState(OperationState.ERROR)
+            val ke = KyuubiSQLException("Error submitting query in background, query rejected",
+              rejected)
+            setOperationException(ke)
+            throw ke
+          }
       }
     } else {
       executeStatement()
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -117,16 +117,14 @@ class ExecuteStatement(
         val backgroundHandle = sparkSQLSessionManager.submitBackgroundOperation(asyncOperation)
         setBackgroundHandle(backgroundHandle)
       } catch {
-        case rejected: RejectedExecutionException =>
-          if (sparkSQLSessionManager.isShutdown) {
-            // ignore this exception since the engine is shutdown
-          } else {
-            setState(OperationState.ERROR)
-            val ke = KyuubiSQLException("Error submitting query in background, query rejected",
-              rejected)
-            setOperationException(ke)
-            throw ke
-          }
+        case rejected: RejectedExecutionException if !sparkSQLSessionManager.isShutdown =>
+          setState(OperationState.ERROR)
+          val ke = KyuubiSQLException("Error submitting query in background, query rejected",
+            rejected)
+          setOperationException(ke)
+          throw ke
+        // ignore this exception since the engine is shutdown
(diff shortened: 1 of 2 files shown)


timothy65535:
hi @ulysses-you

> Engine can be shutdown but not stopped 

How can I encounter this situation? The `shutdown` variable will be set to true only when the service is stopped in the codebase.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -114,17 +114,21 @@ class ExecuteStatement(
...
         val backgroundHandle = sparkSQLSessionManager.submitBackgroundOperation(asyncOperation)
         setBackgroundHandle(backgroundHandle)
       } catch {
         case rejected: RejectedExecutionException =>
-          setState(OperationState.ERROR)
-          val ke = KyuubiSQLException("Error submitting query in background, query rejected",
-            rejected)
-          setOperationException(ke)
-          throw ke
+          if (sparkSQLSessionManager.isShutdown) {
```
pan3793, line 123:
```
case rejected: RejectedExecutionException if !sparkSQLSessionManager.isShutdown =>
```

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -198,6 +198,8 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
     }
   }
 
+  def isShutdown: Boolean = shutdown
```
SteNicholas, line 201: IMO, here is unnecessary. You could directly use the `shutdown` to check.

ulysses-you: `shutdown` is a private mutable field, it should be good to add a method here

SteNicholas (approved):
LGTM. cc @yaooqinn @pan3793

yaooqinn:
I am -1 with this.

Instead, we still need to report the operation status as ERROR even the engine is shutting down.

But, we can improve the error for this case.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -112,17 +112,19 @@ class ExecuteStatement(
...
+      val sparkSQLSessionManager = session.sessionManager
       try {
-        val sparkSQLSessionManager = session.sessionManager
         val backgroundHandle = sparkSQLSessionManager.submitBackgroundOperation(asyncOperation)
         setBackgroundHandle(backgroundHandle)
       } catch {
-        case rejected: RejectedExecutionException =>
+        case rejected: RejectedExecutionException if !sparkSQLSessionManager.isShutdown =>
           setState(OperationState.ERROR)
           val ke = KyuubiSQLException("Error submitting query in background, query rejected",
```
yaooqinn, line 122: how about we refine the error message to show that the engine is in shutting down phase.

PR #914, opened 2021-08-10
Use isEmpty to trigger default init queries

yaooqinn:

1 use isEmpty instead show for the init SQLs
2 add job desc for SQL tab of Spark UI
3 make it simple.


yaooqinn:
cc @pan3793 @turboFei @ulysses-you

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -35,9 +35,9 @@ object Utils extends Logging {
 
   import org.apache.kyuubi.config.KyuubiConf._
 
-  def strToSeq(s: String): Seq[String] = {
+  def strToSeq(s: String, sp: String = ","): Seq[String] = {
```
pan3793, line 38: what if sql contains `,` ?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -53,20 +55,15 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
...
-          .split(";")
-          .filter(_.trim.nonEmpty)
-          .foreach { sql =>
-            info(s"Execute session initializing sql: $sql")
-            sparkSession.sql(sql).show
-          }
+        val ss = spark.newSession()
+        this.conf.get(ENGINE_SESSION_INITIALIZE_SQL).foreach { sqlStr =>
+          ss.sparkContext.setJobGroup(handle.identifier.toString, sqlStr, interruptOnCancel = true)
+          ss.sql(sqlStr).isEmpty
```
pan3793, line 64: can we keep log here?

yaooqinn: the UI is more straightforward

pan3793 (approved):

PR #920, opened 2021-08-12
Add COMPILED State for ExecuteStatement af engine side

yaooqinn:

RUNNING -> COMPILED can be used to calculate to the duration of a statement being parsed, analyzed, and optimized

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -93,6 +93,7 @@ class ExecuteStatement(
       // TODO: Make it configurable
       spark.sparkContext.addSparkListener(operationListener)
       result = spark.sql(statement)
+      // TODO( #921): COMPILED need consider eagerly executed commands
       setState(OperationState.COMPILED)
       kyuubiStatementInfo.queryExecution = result.queryExecution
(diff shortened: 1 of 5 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationState.scala
```diff
@@ -42,24 +52,14 @@ object OperationState extends Enumeration {
     case UNKNOWN => UKNOWN_STATE
   }
 
-  implicit def fromTOperationState(from: TOperationState): OperationState = from match {
```
yaooqinn, line 45: dead code

yaooqinn:
cc @zhang1002 @ulysses-you @pan3793 @turboFei @yanghua

pan3793 (approved):
LGTM

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -93,6 +93,7 @@ class ExecuteStatement(
       // TODO: Make it configurable
       spark.sparkContext.addSparkListener(operationListener)
       result = spark.sql(statement)
+      // TODO( #921): COMPILED need consider eagerly executed commands
```
yaooqinn, line 96: cc @ulysses-you

ulysses-you (approved):

ulysses-you:
thanks, merging to master for 1.3.0

Issue #930, opened 2021-08-13
Extract the zookeeper part from ServiceDiscovery


timothy65535:
### Description

`ServiceDiscovery` contains `discovery` and `zookeeper` part, `ServiceDiscovery` should only has one responsibility.

![image](https://user-images.githubusercontent.com/86483005/129339811-9cfaeabf-10dc-4962-9d59-4ca7835ccce5.png)

PR #931, opened 2021-08-13
[KYUUBI #930] Extract the zookeeper part from ServiceDiscovery

timothy65535:
### Description

`ServiceDiscovery` contains `discovery` and `zookeeper` part, `ServiceDiscovery` should only has one responsibility.

image


yaooqinn (approved):
looks OK to me

Issue #933, opened 2021-08-16
Enhance the detection mechanism for engine startup


timothy65535:
### 1. Description

EngineRef use exists znode to determine whether engine started successfully or not(use the last znode), there are inconsistencies in this. So, take an idea to improve the detection mechanism by adding the started sessionid to the znone. `EngineRef` will check the sessionid in a loop after a engine started.

### 2. Here are two options
**Option I, add sesssionid to the data of znode**
![image](https://user-images.githubusercontent.com/86483005/128994276-5c9c7839-0a21-4253-8989-110b88ac0e74.png)

**Option II, add sessionid to the path of znode**
![image](https://user-images.githubusercontent.com/86483005/129527282-46f7f726-e110-4d87-9272-dda9b9f325bf.png)

### 3. Solution

In order to be consistent with the design of server znode, prefer to choose _Option II_ .

PR #935, opened 2021-08-16
[KYUUBI #933] Enhance the detection mechanism for engine startup

timothy65535:
### 1. Description

EngineRef use exists znode to determine whether engine started successfully or not(use the last znode), there are inconsistencies in this. So, take an idea to improve the detection mechanism by adding the started sessionid to the znone. `EngineRef` will check the sessionid in a loop after a engine started.

### 2. Here are two options
**Option I, add sesssionid to the data of znode**
image

**Option II, add sessionid to the path of znode**
image

### 3. Solution

In order to be consistent with the design of server znode, prefer to choose _Option II_ . 

**Demo**
```
/kyuubi
- serviceUri=bigdata:10009;version=1.3.0-SNAPSHOT;sequence=0000000000
/kyuubi_USER/test
- serviceUri=bigdata:39869;version=1.3.0-SNAPSHOT;session=8178069f-0b22-4d06-b1c0-908094769397;sequence=0000000000]
```

### DIFF
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
@@ -103,7 +103,7 @@ object HighAvailabilityConf {
   // This option will not be exposed to the user.
   val HA_ZK_ENGINE_SESSION_ID: OptionalConfigEntry[String] =
     ConfigBuilder("kyuubi.ha.engine.session.id")
-    .version("1.3.0")
+    .version("1.4.0")
     .stringConf
     .createOptional
 }
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
@@ -21,7 +21,7 @@ import java.time.Duration
 
 import org.apache.hadoop.security.UserGroupInformation
 
-import org.apache.kyuubi.config.{ConfigBuilder, ConfigEntry, KyuubiConf}
+import org.apache.kyuubi.config.{ConfigBuilder, ConfigEntry, KyuubiConf, OptionalConfigEntry}
 import org.apache.kyuubi.ha.client.RetryPolicies
 
 object HighAvailabilityConf {
@@ -101,7 +101,8 @@ object HighAvailabilityConf {
     .createWithDefault(Duration.ofSeconds(120).toMillis)
 
   // This option will not be exposed to the user.
-  val HA_ZK_ENGINE_SESSION_ID: ConfigEntry[String] = ConfigBuilder("kyuubi.ha.engine.session.id")
+  val HA_ZK_ENGINE_SESSION_ID: OptionalConfigEntry[String] =
+    ConfigBuilder("kyuubi.ha.engine.session.id")
     .version("1.3.0")
     .stringConf
     .createOptional
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
@@ -99,4 +99,10 @@ object HighAvailabilityConf {
     .timeConf
     .checkValue(_ > 0, "Must be positive")
     .createWithDefault(Duration.ofSeconds(120).toMillis)
+
+  // This option will not be exposed to the user.
+  val HA_ZK_ENGINE_SESSION_ID: ConfigEntry[String] = ConfigBuilder("kyuubi.ha.engine.session.id")
+    .version("1.2.0")
+    .stringConf
+    .createWithDefault("-1")
 }
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
@@ -102,7 +102,7 @@ object HighAvailabilityConf {
 
   // This option will not be exposed to the user.
   val HA_ZK_ENGINE_SESSION_ID: ConfigEntry[String] = ConfigBuilder("kyuubi.ha.engine.session.id")
-    .version("1.2.0")
+    .version("1.3.0")
     .stringConf
-    .createWithDefault("-1")
+    .createOptional
 }
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
@@ -100,9 +100,11 @@ object HighAvailabilityConf {
     .checkValue(_ > 0, "Must be positive")
     .createWithDefault(Duration.ofSeconds(120).toMillis)
 
-  // This option will not be exposed to the user.
   val HA_ZK_ENGINE_SESSION_ID: OptionalConfigEntry[String] =
-    ConfigBuilder("kyuubi.ha.engine.session.id")
+    buildConf("ha.engine.session.id")
+    .doc("The sessionId will be attached to zookeeper node when engine started, " +
+      "and the kyuubi server will check it cyclically.")
+    .internal
     .version("1.4.0")
     .stringConf
     .createOptional
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
@@ -161,11 +161,11 @@ object ServiceDiscovery extends Logging {
     zkEnsemble != null && zkEnsemble.nonEmpty
   }
 
-  def getServerHost(zkClient: CuratorFramework, namespace: String): Option[ServiceNodeInfo] = {
+  def getServerHost(zkClient: CuratorFramework, namespace: String): Option[(String, Int)] = {
     // TODO: use last one because to avoid touching some maybe-crashed engines
     // We need a big improvement here.
     getServiceNodesInfo(zkClient, namespace, Some(1), silent = true) match {
-      case Seq(sn) => Some(sn)
+      case Seq(sn) => Some((sn.host, sn.port))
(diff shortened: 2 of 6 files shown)


kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
```diff
@@ -99,4 +99,10 @@ object HighAvailabilityConf {
     .timeConf
     .checkValue(_ > 0, "Must be positive")
     .createWithDefault(Duration.ofSeconds(120).toMillis)
+
+  // This option will not be exposed to the user.
+  val HA_ZK_ENGINE_SESSION_ID: ConfigEntry[String] = ConfigBuilder("kyuubi.ha.engine.session.id")
+    .version("1.2.0")
```
ulysses-you, line 105: now the master branch is `1.3.0` but will be cut soon cc @pan3793

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
```diff
@@ -99,4 +99,10 @@ object HighAvailabilityConf {
...
+    .stringConf
+    .createWithDefault("-1")
```
ulysses-you, line 107: `createOptional` ?

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -258,6 +255,7 @@ case class ServiceNodeInfo(
     nodeName: String,
     host: String,
     port: Int,
-    version: Option[String]) {
+    version: Option[String],
+    sessionId: Option[String]) {
```
ulysses-you, line 259: how about naming this and zk node `createSessionId` FYI @yaooqinn

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
timothy65535: Nice!!

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
timothy65535: Good catch.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
timothy65535: make sense.

yaooqinn:
Is UT broken？ The newly added ut does not trigger failure？

timothy65535:
> Is UT broken？ The newly added ut does not trigger failure？

Already rebase the latest master branch to the patch.

yaooqinn:
Can you check the UT, adding a new configuration need update the doc by default

timothy65535:
> Can you check the UT, adding a new configuration need update the doc by default

Hi @yaooqinn, you mean `HA_ZK_ENGINE_SESSION_ID` option? this option will be never exposed to end-user, so we don't need update the doc. I use ConfigBuilder to define `HA_ZK_ENGINE_SESSION_ID` instead of `buildConf`, so the UT will not check it.

yaooqinn:
> > Can you check the UT, adding a new configuration need update the doc by default
> 
> Hi @yaooqinn, you mean `HA_ZK_ENGINE_SESSION_ID` option? this option will be never exposed to end-user, so we don't need update the doc. I use ConfigBuilder to define `HA_ZK_ENGINE_SESSION_ID` instead of `buildConf`, so the UT will not check it.

this is a bit tricky, we'd better add an `isInternal` flag like spark

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -124,7 +126,9 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
   }
 
   private def get(zkClient: CuratorFramework): Option[(String, Int)] = {
```
yaooqinn, line 128: `getEngineBySessionId`

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -124,7 +126,9 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
-    getServerHost(zkClient, engineSpace)
+    getServiceNodesInfo(zkClient, engineSpace, silent = true)
```
yaooqinn, line 129: https://github.com/apache/incubator-kyuubi/pull/935/files#diff-6c518a53d95d41547d6677ef6f3b7557f1e52a93a8c81ab547d95ca19d601f53R135-R137

we shouldn't filter by id here

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: Done.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: Two method calls to ServiceDiscovery
```
(No filter )    KyuubiSessionImpl -> getOrCreate -> getServerHost
(Filter by id)  EngineRef -> create -> getEngineBySessionId -> getServiceNodesInfo
```

yaooqinn: `EngineRef -> create -> (blocked by lock) -> (obtain lock) -> (try to get engine by another success) -> get OR actually create`

timothy65535: right, got your point, we need filter id here, and already rename `get` to `getEngineBySessionId`.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -215,26 +216,22 @@ object ServiceDiscovery extends Logging {
       case e: KeeperException =>
         throw new KyuubiException(s"Failed to create namespace '$ns'", e)
     }
+
+    val session = conf.getOption(HA_ZK_ENGINE_SESSION_ID.key)
```
yaooqinn, line 220: conf.get(HA_ZK_ENGINE_SESSION_ID)

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -123,12 +125,14 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
-    getServerHost(zkClient, engineSpace)
+  private def getEngineBySessionId(zkClient: CuratorFramework): Option[(String, Int)] = {
+    getServiceNodesInfo(zkClient, engineSpace, silent = true)
+      .find(_.createSessionId.exists(_.equals(sessionId)))
+      .map(data => (data.host, data.port))
   }
 
   private def create(zkClient: CuratorFramework): (String, Int) = tryWithLock(zkClient) {
-    var engineRef = get(zkClient)
+    var engineRef = getEngineBySessionId(zkClient)
```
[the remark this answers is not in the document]
timothy65535: we can'n use `getServerHost` here, only use `getEngineBySessionId` is right.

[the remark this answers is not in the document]
yaooqinn: We should use getServerHost, so we don't get a engine by session id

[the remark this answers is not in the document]
yaooqinn: https://github.com/apache/incubator-kyuubi/pull/935/files#diff-6c518a53d95d41547d6677ef6f3b7557f1e52a93a8c81ab547d95ca19d601f53R130

Please use `getServerHost`, this is for a concurrent connection, which uses an engine created by `another sessionId`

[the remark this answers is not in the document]
timothy65535: hi, `getEngineBySessionId` is under lock, will not meet concurrent issue, let me think how to explain it to you.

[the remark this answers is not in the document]
timothy65535: _`getServerHost`_: used to choose suitable engine, where we can use pool here in the future.
_`getEngineBySessionId`_:  used to check the session.id, actually it doesn’t need to be under the lock, it is stateless.


image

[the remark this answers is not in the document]
yaooqinn: I see what you mean. Let's wait for #958

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -123,12 +125,9 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
       }
   }
 
-  private def get(zkClient: CuratorFramework): Option[(String, Int)] = {
-    getServerHost(zkClient, engineSpace)
-  }
-
   private def create(zkClient: CuratorFramework): (String, Int) = tryWithLock(zkClient) {
-    var engineRef = get(zkClient)
+    // TODO: improve this after support engine pool
```
yaooqinn, line 129: can we also add the issue id？

timothy65535: issue id? ❓

timothy65535: you mean add https://github.com/apache/incubator-kyuubi/issues/918 ?

yaooqinn (approved):

yaooqinn:
thanks, merged to master

timothy65535:
> thanks, merged to master

Welcome.

PR #936, opened 2021-08-16
Add date info for json event path

ulysses-you:

Add date info helps collect metrics across some days.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
@@ -36,7 +36,7 @@ import org.apache.kyuubi.service.AbstractService
 /**
  * This event logger logs Kyuubi engine events in JSON file format.
  * The hierarchical directory structure is:
- *   ${ENGINE_EVENT_JSON_LOG_PATH}/event=${eventType}/day=${date}/${logName}.json
+ *   ${ENGINE_EVENT_JSON_LOG_PATH}/${eventType}/day=${date}/${logName}.json
  * The ${eventType} is based on core concepts of the Kyuubi systems, e.g. engine/session/statement
  * The ${date} is based on the time of events, e.g. engine.startTime, statement.startTime
  * @param logName the engine id formed of appId + attemptId(if any)
@@ -53,7 +53,7 @@ class JsonEventLogger(logName: String, hadoopConf: Configuration)
   private def getOrUpdate(event: KyuubiEvent): Logger = synchronized {
     writers.getOrElseUpdate(event.eventType + event.datePartition, {
       val eventPath = new Path(
-        new Path(new Path(logRoot), s"event=${event.eventType}"), s"day=${event.datePartition}")
+        new Path(new Path(logRoot), s"${event.eventType}"), s"day=${event.datePartition}")
       FileSystem.mkdirs(fs, eventPath, JSON_LOG_DIR_PERM)
       val logFile = new Path(eventPath, logName + ".json")
       var hadoopDataStream: FSDataOutputStream = null
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
@@ -36,8 +36,9 @@ import org.apache.kyuubi.service.AbstractService
 /**
  * This event logger logs Kyuubi engine events in JSON file format.
  * The hierarchical directory structure is:
- *   ${ENGINE_EVENT_JSON_LOG_PATH}/day=${date}/event=${eventType}/${logName}.json
+ *   ${ENGINE_EVENT_JSON_LOG_PATH}/event=${eventType}/day=${date}/${logName}.json
  * The ${eventType} is based on core concepts of the Kyuubi systems, e.g. engine/session/statement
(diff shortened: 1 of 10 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
```diff
@@ -27,30 +27,33 @@ import org.apache.hadoop.conf.Configuration
...
+ *   ${ENGINE_EVENT_JSON_LOG_PATH}/day=${date}/event=${eventType}/${logName}.json
+ * The ${eventType} is based on core concepts of the Kyuubi systems, e.g. engine/session/statement
  * @param logName the engine id formed of appId + attemptId(if any)
  */
 class JsonEventLogger(logName: String, hadoopConf: Configuration)
   extends AbstractService("JsonEventLogger") with EventLogger with Logging {
 
   type Logger = (PrintWriter, Option[FSDataOutputStream])
 
+  private val currentDate = Utils.getCurrentDate
```
timothy65535, line 48: if we define `currentDate` as a member variable of the class, it will not change after initialization.

ulysses-you: yes, it's OK since this logger is used at engine side. That said, one engine has a certain date.

yaooqinn:
I have not catch up withthis idea，so let me ask some general questions.

1. Why need we duplicate the date info we already have in all kinds of events？
2. How does it possible to let event be the sub partition column of date，while they contains totally different schemas
3. Why does itonly affect the json writer？
4. How a static date handle session or statement events that might be created in the same engine but in different days？

ulysses-you:
The origin requirements is to collect some days event. Let's say we need to collect last 7 days statement event, after this PR, we can index those files easily using date partition path.

> Why need we duplicate the date info we already have in all kinds of events？

yeah, We can use `sc.startTime` instead of `CurrentDate` here.

> How does it possible to let event be the sub partition column of date，while they contains totally different schemas

It depends on how to use. I didn't try to create a table on all events since they have different schema, but it can help index one event files with the certain date.

> Why does it only affect the json writer？

Currently, We only support write metrics using json format(exclude the Spark Event logger), it can be a quick start. 

> How a static date handle session or statement events that might be created in the same engine but in different days？

I think it's not a problem, all events for the certain engine can be written in one date partition and it also help to index event log.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
```diff
@@ -35,8 +35,10 @@ import org.apache.kyuubi.service.AbstractService
 
 /**
  * This event logger logs Kyuubi engine events in JSON file format.
- * The hierarchical directory structure is {ENGINE_EVENT_JSON_LOG_PATH}/{eventType}/{logName}.json
- * The {eventType} is based on core concepts of the Kyuubi systems, e.g. engine/session/statement
+ * The hierarchical directory structure is:
+ *   ${ENGINE_EVENT_JSON_LOG_PATH}/event=${eventType}/day=${date}/${logName}.json
```
timothy65535, line 39: Maybe is more friendly : )
`${ENGINE_EVENT_JSON_LOG_PATH}/day=${date}/event=${eventType}/${logName}.json`

ulysses-you: Since the events have different schema, we can create view on `${ENGINE_EVENT_JSON_LOG_PATH}/event=${eventType}`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
```diff
@@ -49,8 +51,9 @@ class JsonEventLogger(logName: String, hadoopConf: Configuration)
   private val writers = HashMap.empty[String, Logger]
 
   private def getOrUpdate(event: KyuubiEvent): Logger = synchronized {
-    writers.getOrElseUpdate(event.eventType, {
-      val eventPath = new Path(new Path(logRoot), event.eventType)
+    val partitions = event.partitions.map(kv => s"${kv._1}=${kv._2}").mkString("/")
+    writers.getOrElseUpdate(event.eventType + partitions, {
+      val eventPath = new Path(new Path(new Path(logRoot), event.eventType), partitions)
```
yaooqinn, line 56: new Path(new Path("abc"), "")
java.lang.IllegalArgumentException: Can not create a Path from an empty string

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JsonEventLogger.scala
```diff
@@ -49,8 +51,9 @@ class JsonEventLogger(logName: String, hadoopConf: Configuration)
   private val writers = HashMap.empty[String, Logger]
 
   private def getOrUpdate(event: KyuubiEvent): Logger = synchronized {
-    writers.getOrElseUpdate(event.eventType, {
-      val eventPath = new Path(new Path(logRoot), event.eventType)
+    val partitions = event.partitions.map(kv => s"${kv._1}=${kv._2}").mkString("/")
```
yaooqinn, line 54: Please do use `Path.SEPARATOR`

yaooqinn (approved):

ulysses-you:
thanks, merging to master/v1.3.0

Issue #942, opened 2021-08-17
Support for adding internal attributes to ConfigBuilder


timothy65535:
### Description

`ConfigBuilder` currently only supports public configuration by default. In practice, there are also some configurations, which are internal configurations and do not need to be reserved for users.

refer to https://github.com/apache/incubator-kyuubi/pull/935#issuecomment-900117832

### Modification

- Add `internal` to ConfigBuilder
- Update `AllKyuubiConfiguration`

yaooqinn:
👍

PR #949, opened 2021-08-18
[KYUUBI #942] Support for adding internal attributes to ConfigBuilder

timothy65535:
### Description

`ConfigBuilder` currently only supports public configuration by default. In practice, there are also some configurations, which are internal configurations and do not need to be reserved for users.

refer to https://github.com/apache/incubator-kyuubi/pull/935#issuecomment-900117832

### Modification

- Add `internal` to ConfigBuilder
- Update `AllKyuubiConfiguration`

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/config/AllKyuubiConfiguration.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/config/AllKyuubiConfiguration.scala
@@ -119,7 +119,7 @@ class AllKyuubiConfiguration extends KyuubiFunSuite {
 
     KyuubiConf.kyuubiConfEntries.values().asScala
       .toSeq
-      .filter(cfg => !cfg.internal)
+      .filterNot(_.internal)
       .groupBy(_.key.split("\\.")(1))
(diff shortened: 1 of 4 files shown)


ulysses-you:
Shall we skip generating configuration docs in `AllKyuubiConfiguration` if config tag is internal ?

timothy65535:
hi @ulysses-you 

> Shall we skip generating configuration docs in AllKyuubiConfiguration if config tag is internal ?

The internal optional is not exposed to end-user, so we don't need to update `settings.md`. We will never assign values manually.

kyuubi-server/src/test/scala/org/apache/kyuubi/config/AllKyuubiConfiguration.scala
```diff
@@ -119,6 +119,7 @@ class AllKyuubiConfiguration extends KyuubiFunSuite {
 
     KyuubiConf.kyuubiConfEntries.values().asScala
       .toSeq
+      .filterNot(_.internal)
```
ulysses-you, line 122: ah, I see thank you for the filter

Issue #954, opened 2021-08-18
Refactor dockerfile


cfmcgrady:
### 1. Describe the feature

1. As the binary distributions contain all things that we want, we don't need a pre-build spark as a docker base image, use java as the base image instead.

2. Build kyuubi from source code.

zwangsheng:
Thanks for your issue. My opinion is that if we don't use kyuubi externals spark, we need to set SPARK_HOME in kyuubi-env.sh. This requires us to package spark together in the image. It may better to use FROM Spark image.

PR #955, opened 2021-08-18
[KYUUBI #954] Refactor Dockerfile

cfmcgrady:

1. As the binary distributions contain all things that we want, we don't need a pre-build spark as a docker base image, use java as the base image instead.

2. Build kyuubi from source code.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -171,7 +171,7 @@ jobs:
       - name: kyuubi pod check
         run: kubectl get pods
       - name: integration tests
-        run: ./build/mvn clean test -Pkubernetes -Dtest=none -DwildcardSuites=org.apache.kyuubi.kubernetes.test
+        run: ./build/mvn clean test -pl :kyuubi-common_2.12,:kubernetes-integration-tests_2.12 -Pkubernetes -Dtest=none -DwildcardSuites=org.apache.kyuubi.kubernetes.test
       - name: Upload test logs
         if: failure()
         uses: actions/upload-artifact@v2
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -175,11 +175,11 @@ jobs:
           kubectl get serviceaccount
       - name: start kyuubi
         run: kubectl apply -f kubernetes/integration-tests/test-k8s.yaml
-      - name: kyuubi pod check
-        run: kubectl get pods
       - name: Sleeep for 2 minutes (only for debug GA)
         run: sleep 120s
         shell: bash
+      - name: kyuubi pod check
+        run: kubectl get pods
       - name: integration tests
         run: ./build/mvn clean test -Pkubernetes -Dtest=none -DwildcardSuites=org.apache.kyuubi.kubernetes.test
       - name: Upload test logs
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -145,6 +145,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      # https://github.com/docker/build-push-action
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Build Kyuubi Docker Image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: docker/Dockerfile
+          tags: kyuubi:latest
       # from https://github.com/marketplace/actions/setup-minikube-kubernetes-cluster
       - name: Setup Minikube
         uses: manusa/actions-setup-minikube@v2.4.2
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -146,22 +146,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       # https://github.com/docker/build-push-action
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
-        with:
-          driver-opts: network=host
       - name: Build Kyuubi Docker Image
         uses: docker/build-push-action@v2
         with:
           context: .
           file: docker/Dockerfile
-          push: true
-          tags: localhost:5000/apache/kyuubi:latest
-      - name: Inspect
-        run: |
-          docker buildx imagetools inspect localhost:5000/apache/kyuubi:latest
+          load: true
+          tags: apache/kyuubi:latest
       # from https://github.com/marketplace/actions/setup-minikube-kubernetes-cluster
       - name: Setup Minikube
         uses: manusa/actions-setup-minikube@v2.4.2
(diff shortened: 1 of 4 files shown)


cfmcgrady:
cc @ulysses-you @yaooqinn

docker/Dockerfile
```diff
@@ -16,18 +16,26 @@
...
+
+ADD . /workspace/kyuubi
+WORKDIR /workspace/kyuubi
+
+RUN ./build/dist ${MVN_ARG} && \
+    mv /workspace/kyuubi/dist /opt/kyuubi && \
+    rm -rf ~/.m2 && \
+    rm -rf /workspace/kyuubi
+
+FROM openjdk:8-jdk
```
timothy65535, line 38: this line is useless, already use `FROM maven:3.6-jdk-8 as builder`
`FROM openjdk:8-jdk`

cfmcgrady: The first one `FROM maven:3.6-jdk-8 as builder` provides build Kyuubi runtime(mvn/java), and the second one `FROM openjdk:8-jdk` is for Kyuubi server runtime(java only). In order to reduce the final docker image size, we removed MAVEN here.

yaooqinn (approved):

ulysses-you:
How do you consider about if user package Kyuubi with `--spark-provided` or some private Spark image ? Do you plan to support that case ?

zwangsheng:
My opinion is that if we use spark docker image as the underlying image, we can easily manage the version of Kyuubi engine.

cfmcgrady:
How about make the base image configurable? 

```
....
# set this build argument if the user wants to customize Spark runtime
ARG BASE_IMAGE=openjdk:8-jdk

FROM ${BASE_IMAGE}
```

build with a custom Spark runtime.

```
docker build --build-arg MVN_ARG="--spark-provided" --build-arg BASE_IMAGE="your-private-spark-image-name:tag" -f docker/Dockerfile -t apache/kyuubi:tagname .
```
Note that the provided base image should include the base environment(java/spark)
@ulysses-you @zwangsheng

ulysses-you:
If user package with `--spark-provided` that said Kyuubi would not generate Spark binary package during building. So we need give user a way to copy their self Spark binary package into Kyuubi dockerfile.

How about add ARG `SPARK_HOME` in dockerfile:
```
COPY ${SPARK_HOME} ${KYUUBI_HOME}/externals/spark-provided
```

And we should update the `SPARK_HOME` in kyuubi-env.sh to make it can be found by Kyuubi.

docker/Dockerfile
```diff
@@ -16,18 +16,26 @@
...
+
+ARG MVN_ARG
+
+ADD . /workspace/kyuubi
+WORKDIR /workspace/kyuubi
+
+RUN ./build/dist ${MVN_ARG} && \
+    mv /workspace/kyuubi/dist /opt/kyuubi && \
+    rm -rf ~/.m2 && \
+    rm -rf /workspace/kyuubi
```
pan3793, line 36: why should clean them?

docker/Dockerfile
```diff
@@ -16,18 +16,26 @@
 #
 
 # Usage:
-#   1. use ./build/dist to make binary distributions of Kyuubi or download a release
-#   2. Untar it and run the docker command below
-#      docker build -f docker/Dockerfile -t yaooqinn/kyuubi:tagname .
+#   Run the docker command below
+#      docker build --build-arg MVN_ARG="-Pmirror-cn" -f docker/Dockerfile -t apache/kyuubi:tagname .
```
pan3793, line 20: I think it's better to change the demo value of `MVN_ARG` from `-Pmirror-cn` to `-Pspark-3.1,spark-hadoop-3.2`

docker/Dockerfile
```diff
@@ -16,18 +16,26 @@
...
-# TODO: REPALCE it with offical spark image iff Apache Spark community deploy
-# Or make one after kyuubi to be setup under ASF
 
-ARG SPARK_IMAGE=yaooqinn/spark:3.0.3
-FROM ${SPARK_IMAGE}
+FROM maven:3.6-jdk-8 as builder
+
+ARG MVN_ARG
+
+ADD . /workspace/kyuubi
```
pan3793, line 30: please also add a `.dockerignore` at the project root folder

cfmcgrady:
It seems hard to do this, isn't it? As I know, Dockerfile doesn't support optional copy/add, this means we should programming generate Dockerfile. So I think using BASE_IMAGE as a build argument is a simpler approach. WDYT? @ulysses-you

docker/Dockerfile
cfmcgrady: Removing stuff saves time because docker creates a temporary layer. I will add comments here and explain why we clean them.

docker/Dockerfile
cfmcgrady: make sense

docker/Dockerfile
cfmcgrady: Which file/directory should we ignore now?

ulysses-you:
@cfmcgrady I aggre that. Can we add a new dockerfile instead of overwrite the existed ? So we can use this new dockerfile to do kubernetes integration test.

For the kubernetes integration test, you can use github action that add a new step to build docker image  in https://github.com/apache/incubator-kyuubi/blob/32b18ba19c969206d3ce03637560ff84f88009d2/.github/workflows/master.yml#L143

Then change the image name in existed test-k8s.yaml
https://github.com/apache/incubator-kyuubi/blob/32b18ba19c969206d3ce03637560ff84f88009d2/kubernetes/integration-tests/test-k8s.yaml#L29

docker/Dockerfile
pan3793: The `.git` and `target` folder is quite large and not required for `build/dist`, let's add these two first.

cfmcgrady: Good catch. Updated.

pan3793 (approved):

cfmcgrady:
GA failed, I'm testing GA with my forked repository.

.github/workflows/master.yml
```diff
@@ -145,6 +145,16 @@ jobs:
...
+      # https://github.com/docker/build-push-action
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Build Kyuubi Docker Image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: docker/Dockerfile
+          load: true
+          tags: apache/kyuubi:latest
```
yaooqinn, line 157: add `-pl :kyuubi-common_2.12,:kubernetes-integration-tests_2.12` to 
L174 `./build/mvn clean test -Pkubernetes -Dtest=none -DwildcardSuites=org.apache.kyuubi.kubernetes.test`?

cfmcgrady: Done.

yaooqinn (approved):
LGTM

ulysses-you (approved):
LGTM, thank you @cfmcgrady , merging to master !

cfmcgrady:
Thanks all.

Issue #918, opened 2021-08-11
[Umbrella] Enhance the engine service ha framework


timothy65535:
## Description

From the ha discovery design of engine service in codebase, there're some area that can be improved.

### 1. EngineRef is exposed to the KyuubiSessionImpl

When client connect to kyuubi server, the backendservice will open KyuubiSession. KyuubiSessionImpl use `EngineRef` and `ServiceDiscovery` at the same time in the codebase. 

The responsibility of `EngineRef` is to maintain the startup of the engine, it's better to move `EngineRef` to `EngineServiceDiscovery`. 
![image](https://user-images.githubusercontent.com/86483005/128994238-f1f6856b-9c41-4f6c-b844-829ad7052ee8.png)

### 2. The mechanism for determining whether engine started successfully is flawed

EngineRef use exists znode to determine whether engine started successfully or not(use the last znode), there are inconsistencies in this.
So, take an idea to improve the detection mechanism by adding the started sessionid to the contents of the znone when engine started, then check the sessionid.
![image](https://user-images.githubusercontent.com/86483005/128994276-5c9c7839-0a21-4253-8989-110b88ac0e74.png)

### 3. Missing support multiple engines under `USER` share level

Kyuubi supports `CONNECTION`, `USER`, `SERVER` share levels, we want to use kyuubi as hiveserver which can support multiple servers under same zookeeper namespace to share load pressure, client can pick hiveserver randomly. However, due to the influence of the second point above, kyuubi can not create other new engine under same user. Thes is also a little redundant to `subDomain`.
![image](https://user-images.githubusercontent.com/86483005/128994314-c978bf61-6a8f-4d95-9ea6-9fb96ca786cc.png)

### 4. Additional context

The first and second points can be optimized step by step, the third point requires more thought, welcome to discuss.

ulysses-you:
thank you @timothy65535 for working on this.

About the second point. What if one user create two sessions, shall we reuse the same engine ? It sounds like the idea about engine pool that user can decide how many engine can be used in parallel.

timothy65535:
Welcome @ulysses-you, the second point focus on improve the detection mechanism for engine startup. 

> What if one user create two sessions, shall we reuse the same engine

In my view, user still cannot create a second one unless the third point is completed, or meet concurrent bug.

The engine pool may needs to be implemented based on it. The third point a bit sames to https://github.com/apache/incubator-kyuubi/issues/786

yaooqinn:
[...]

PR #958, opened 2021-08-18
[TEST] Ensure two connections in user mode share the same engine

yaooqinn:

Ensure changes like #935 do not break the rule for the semantics of engine sharing

If and only if we want engines to be pooling,  this test can be changed

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
@@ -17,6 +17,8 @@
 
 package org.apache.kyuubi.operation
 
+import org.scalatest.time.SpanSugar._
+
 import org.apache.kyuubi.WithKyuubiServer
 import org.apache.kyuubi.config.KyuubiConf
 
@@ -26,5 +28,31 @@ class KyuubiOperationPerUserSuite extends WithKyuubiServer with JDBCTests {
 
   override protected val conf: KyuubiConf = {
     KyuubiConf().set(KyuubiConf.ENGINE_SHARE_LEVEL, "user")
+  }
+
+  test("ensure two connections in user mode share the same engine") {
+    var r1: String = null
+    var r2: String = null
+    new Thread {
+      override def run(): Unit = withJdbcStatement() { statement =>
+        val res = statement.executeQuery("set spark.app.name")
+        assert(res.next())
+        r1 = res.getString("value")
+      }
+    }.start()
+
+    new Thread {
+      override def run(): Unit = withJdbcStatement() { statement =>
+        val res = statement.executeQuery("set spark.app.name")
+        assert(res.next())
+        r2 = res.getString("value")
+      }
+    }.start()
+
+    eventually(timeout(120.seconds), interval(100.milliseconds)) {
+      assert(r1 != null && r2 != null)
+    }
+
+    assert(r1 === r2)
   }
 }


yaooqinn:
cc @timothy65535 @pan3793

timothy65535:
Hi @yaooqinn, got your point.

> ensure two connections in user mode share the same engine

I think it is not the time to add this test case, it is the third point mentioned under https://github.com/apache/incubator-kyuubi/issues/918 .

After we support the engine pool feature, we can add this test case, like
```
test("ensure two connections in user mode share the same engine") {
     
    val conf = KyuubiConf(false)
    conf.set("kyuubi.engine.pool.size", 1)

    var r1: String = null
    var r2: String = null
    new Thread {
      override def run(): Unit = withJdbcStatement() { statement =>
        val res = statement.executeQuery("set spark.app.name")
        assert(res.next())
        r1 = res.getString("value")
      }
    }.start()

    new Thread {
      override def run(): Unit = withJdbcStatement() { statement =>
        val res = statement.executeQuery("set spark.app.name")
        assert(res.next())
        r2 = res.getString("value")
      }
    }.start()

    eventually(timeout(120.seconds), interval(100.milliseconds)) {
      assert(r1 != null && r2 != null)
    }

    assert(r1 === r2)
  }
```

yaooqinn:
> Hi @yaooqinn, got your point.
> 
> > ensure two connections in user mode share the same engine
> 
> I think it is not the time to add this test case, it is the third point mentioned on #918 .
> 
> After we support the engine pool feature, we can add this test case, like
> 
> ```
> test("ensure two connections in user mode share the same engine") {
>      
>     val conf = KyuubiConf(false)
>     conf.set("kyuubi.engine.pool.size", 1)
> 
>     var r1: String = null
>     var r2: String = null
>     new Thread {
>       override def run(): Unit = withJdbcStatement() { statement =>
>         val res = statement.executeQuery("set spark.app.name")
>         assert(res.next())
>         r1 = res.getString("value")
>       }
>     }.start()
> 
>     new Thread {
>       override def run(): Unit = withJdbcStatement() { statement =>
>         val res = statement.executeQuery("set spark.app.name")
>         assert(res.next())
>         r2 = res.getString("value")
>       }
>     }.start()
> 
>     eventually(timeout(120.seconds), interval(100.milliseconds)) {
>       assert(r1 != null && r2 != null)
>     }
> 
>     assert(r1 === r2)
>   }
> ```

why can not this test have to wait engine pool？

timothy65535:
> why does this test have to wait engine pool？

In current codebase, kyuubi can not support multiple engines under USER share level.

Kyuubi supports `CONNECTION`, `USER`, `SERVER` share levels, we want to use kyuubi as hiveserver which can support multiple servers under same zookeeper namespace to share load pressure, client can pick hiveserver randomly. However, due to the influence of the second point above, kyuubi can not create other new engine under same user. 

image

yaooqinn:
> client can pick hiveserver randomly. 

this is  equivalent to `client can pick Kyuubi server randomly`

timothy65535:
Hi @yaooqinn 

It's also ok we add this test case first, we need to come back and modify it after we finish supporting the engine pool.

timothy65535:
> > client can pick hiveserver randomly.
> 
> this is equivalent to `client can pick Kyuubi server randomly`

Right, it's not a concurrent issue, it's a pick strategy issue.

If we use `getServerHost` like following, kyuubi will can not create a new more engine under the same USER share level as before.

```
EngineRef.java

  private def create(zkClient: CuratorFramework): (String, Int) = tryWithLock(zkClient) {
    // TODO: improve this after support engine pool
    var engineRef = getServerHost(zkClient, engineSpace)
    // Get the engine address ahead if another process has succeeded
    if (engineRef.nonEmpty) return engineRef.get

    ...
 }

```

timothy65535 (approved):

yaooqinn:
I don't why your case has to rely on a feature called engine pool.

For high client concurrency, `client can pick hiveserver randomly` you said, this is a feature that Kyuubi does have provided, called `High Availability`, you can exposing all Kyuubi server instances to a single namespace for all clients to get. This is the hive thing you mentioned above.

And then we provide `connection` mode, which a user can use to create as many engines as they want according to `this semantic`, let's call it `connection per engine(CPE)`. Underline here, this mode is against JDBC client connection pool (CP), as CP mode should share everything for each connection, but certainly, CPE will create different Spark applications for each connection, and all applications are isolated. Generally speaking, we can also name it `an engine pool`, which does not cache engine instances but bare computing resources, them dynamically creates-uses-terminates engine with the resources.

Relatively speaking, the `user` mode is a perfect thing for CP, as it shares the same engine for all connections, and we do have a `subDomain` for this mode if users want the engine to be only used by connections within a particular CP, not other connections.

For the feature of  `engine pool`, this is a feature definitely NOT for  `connection` mode but `user` or `server`. It does break the CP use case, but it may be good for some particular use cases that do need **a group of long-running engines** within a `USER/subDomain` for extremely high performance and not care about the things for share just like `connection` mode does.

A `subDomain` is also a kind of MANUAL engine pool. It helps a user create more engines in `user` mode but does not have an auto-mechanism, e.g. sequentially / randomly / load-based picking for `subDomain or engines`. Or, maybe we can apply such auto-mechanism to creating and picking `subDomain` to implement the feature of the engine pool.

yaooqinn:
Another option is that we use `subDomain` for registering engines, and on the retrieving side, we can recursively pick all engines

yaooqinn:
> > > client can pick hiveserver randomly.
> > 
> > 
> > this is equivalent to `client can pick Kyuubi server randomly`
> 
> Right, it's not a concurrent issue, it's a pick strategy issue.
> 
> If we use `getServerHost` like following, kyuubi will can not create a new more engine under the same USER share level as before.
> 
> ```
> EngineRef.java
> 
>   private def create(zkClient: CuratorFramework): (String, Int) = tryWithLock(zkClient) {
>     // TODO: improve this after support engine pool
>     var engineRef = getServerHost(zkClient, engineSpace)
>     // Get the engine address ahead if another process has succeeded
>     if (engineRef.nonEmpty) return engineRef.get
> 
>     .....
>  }
> ```

Another option is that we use `subDomain` for registering engines, and on the retrieving side, we can recursively pick all engines

timothy65535:
@yaooqinn Thanks for detail mechanism about kyuubi engine, I understand what you mean very much.

And agree following points 
- For the feature of engine pool, this is a feature is design for `USER` mode.
- A subDomain is also a kind of MANUAL engine pool. It helps a user create more engines in user mode but does not have an auto-mechanism, e.g. sequentially / randomly / load-based picking for subDomain or engines. Or, maybe we can apply such auto-mechanism to creating and picking subDomain to implement the feature of the engine pool.

yaooqinn:
Hi @timothy65535  have you addressed #925 ? It would be great if you can share your story there:)

PR #959, opened 2021-08-19
Reduce EventLoggingService.onEvent time to one with compiled state

ulysses-you:

`setState`  has already call `EventLoggingService.onEvent`, so we don't need to call `EventLoggingService.onEvent` twice with compiled state .

```
    super.setState(newState)
    statementEvent.state = newState.toString
    statementEvent.stateTime = lastAccessTime
    EventLoggingService.onEvent(statementEvent)
  }
```


yaooqinn (approved):

ulysses-you:
thanks, merging to master/branch-1.3.0

Issue #956, opened 2021-08-18
Fail to get table list if there is "_" in schema's name


Baoqi:
### 1. Describe the bug

I used "kyuubi-tpcds" tool to generate test data.  And the result schema is: tpcds_sf1

Then, I use datagrip to connect to kyuubi. Datagrip will not find any tables under tpcds_sf1.

This is because a bug in src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala , function toJavaRegex().

The input is:     tpcds\_sf1

and in function

```
 def toJavaRegex(input: String): String = {
    val res = if (StringUtils.isEmpty(input) || input == "*") {
      "%"
    } else {
      input
    }
    val in = res.toIterator
    val out = new StringBuilder()

    while (in.hasNext) {
      in.next match {
        case c if c == '\\' && in.hasNext => Pattern.quote(Character.toString(in.next()))
        case c if c == '\\' && !in.hasNext => Pattern.quote(Character.toString(c))
        case '_' => out ++= "."
        case '%' => out ++= ".*"
        case c => out ++= Character.toString(c)
      }
    }
    out.result()
  }
```

the " case c if c == '\\' && in.hasNext => Pattern.quote(Character.toString(in.next()))" line,    the Pattern.quote result is just ignored.   So, later, we are querying "tpcdssf1"  as the schema, which definitely don't have any tables (which don't exist)

#### 2.1 Versions

- Kyuubi
  - [x] 1.1.0
  - [ ] 1.0.3
  - [ ] 0.8.x and earlier 

- Spark
  - [x] 3.1.x
  - [ ] 3.0.x
  - [ ] 2.4.x and earlier 

#### 4.2 Log of the corresponding engine

yaooqinn:
@Baoqi Hi, can u send a PR for this bug?

PR #964, opened 2021-08-19
[KYUUBI #956] Fail to get table list if there is _ in schema's name

Baoqi:
Just write some basic tests out in other projects. 

Since this toJavaRegex is not a static function, which is hard to write unit test if don't move it out as  as static function in object.  (But in that case, the change may be too large for this).

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
@@ -18,7 +18,6 @@
 package org.apache.kyuubi.engine.spark.operation
 
 import java.time.ZoneId
-import java.util.regex.Pattern
 
 import org.apache.commons.lang3.StringUtils
 import org.apache.hive.service.rpc.thrift.{TRowSet, TTableSchema}
@@ -62,6 +61,8 @@ abstract class SparkOperation(spark: SparkSession, opType: OperationType, sessio
    *
    * Underscores (_) are converted to '.' and percent signs (%) are converted to '.*'.
    *
+   * (referred to Spark's implementation: convertPattern function in file MetadataOperation.java)
+   *
    * @param input the SQL pattern to convert
    * @return the equivalent Java regular expression of the pattern
    */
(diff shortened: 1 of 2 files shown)


yaooqinn:
can we add a test then for `if there is _ in schema's name` ?

yaooqinn:
we can add an end to end test for this case，it’s very easy to use the current test framework

yaooqinn:
also cc @pan3793 as the release manager of 1.3.0

Baoqi:
> can we add a test then for `if there is _ in schema's name` ?

Sure, after some check, it seems the best place to add this test is in "get schemas" of BasicJDBCTests.scala.

it will be included in the following 5 final test suites:
  - KyuubiOperationHiveCatalogSuite
  - KyuubiOperationPerServerSuite
  - KyuubiOperationPerUserSuite
  - KyuubiStatementMonitorSuite
  - SparkOperationSuite

Without my fix, but with my unit test change,   the first 3 suite still success, but the last 2 will fail.  (haven't got time to learn what is different between first 3 suites with the last 2 suites)

After my fix, for the new unit test change,  all 5 suites will success.

pan3793:
The changes LGTM, but may I know where the original implementations come from?

Baoqi:
> The changes LGTM, but may I know where the original implementations come from?

Hi, as in the comments of source code:

      * (referred to Spark's implementation: convertPattern function in file MetadataOperation.java)

pan3793:
> > can we add a test then for `if there is _ in schema's name` ?
> 
> Sure, after some check, it seems the best place to add this test is in "get schemas" of BasicJDBCTests.scala.
> 
> it will be included in the following 5 final test suites:
> 
> * KyuubiOperationHiveCatalogSuite
> * KyuubiOperationPerServerSuite
> * KyuubiOperationPerUserSuite
> * KyuubiStatementMonitorSuite
> * SparkOperationSuite
> 
> Without my fix, but with my unit test change, the first 3 suite still success, but the last 2 will fail. (haven't got time to learn what is different between first 3 suites with the last 2 suites)
> 
> After my fix, for the new unit test change, all 5 suites will success.

The first 3 suites are `jdbc-client` ==> `kyuubi-spark-engine`.
The later 2 suites are `jdbc-client` ==> `kyuubi-server` ==> `kyuubi-spark-engine`.

Seems somethings happens in `kyuubi-server` when passing through the request, like remove escape chars.

pan3793:
> > The changes LGTM, but may I know where the original implementations come from?
> 
> Hi, as in the comments of source code:
> 
> ```
>   * (referred to Spark's implementation: convertPattern function in file MetadataOperation.java)
> ```
> 
> see https://github.com/apache/spark/blob/master/sql/hive-thriftserver/src/main/java/org/apache/hive/service/cli/operation/MetadataOperation.java#L103
> 
> only the 3 lines are likely same with Spark's implementation.
> 
> Since Kyuubi is based on Spark, and both projects belongs to Apache Foundation, so, I assume this is fine. But please let me know if we need a modified version. Thanks.

Yea, I see, what I mean **original implementations** is the existing code before this PR. Have a quick look on Hive codebase, didn't find the clue. @yaooqinn Do you remember that?

pan3793 (approved):
LGTM

yaooqinn (approved):

yaooqinn:
thanks, merged to master and 1.3

Issue #922, opened 2021-08-12
Add OperationEvent for server to track operation state


hddong:
### 1. Describe the feature
Track operation state change and support output to custom system.

zhang1002:
c #903

hddong:
Thanks, all events is for spark engine now, I think we also need to track server event changes.

yaooqinn:
yea, we can add the events at the server side too

PR #968, opened 2021-08-20
[KYUUBI #922] Add OperationEvent for server to track operation state

hddong:

Track operation state change and support output to custom system.


yaooqinn:
can you detail the changes in your PR to help us review?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/StatementEvent.scala
```diff
@@ -42,7 +42,7 @@ case class StatementEvent(
     var state: String,
     var stateTime: Long,
     var queryExecution: String = "",
-    var exeception: String = "") extends KyuubiEvent {
+    var exception: String = "") extends KyuubiSparkEvent {
```
yaooqinn, line 45: nice catch

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/events/EventLoggingServiceSuite.scala
```diff
@@ -26,10 +26,11 @@ import org.scalatest.time.SpanSugar._
 import org.apache.kyuubi.Utils
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.spark.{KyuubiSparkUtil, WithSparkSQLEngine}
+import org.apache.kyuubi.events.JsonProtocol
 import org.apache.kyuubi.operation.{JDBCTestUtils, OperationHandle}
 
 class EventLoggingServiceSuite extends WithSparkSQLEngine with JDBCTestUtils {
-  import EventLoggerType._
+  import org.apache.kyuubi.events.EventLoggerType._
```
yaooqinn, line 33: move up

hddong:
Yeah, the mainly changes:
1. Move `KyuubiEvent` to `kyuubi-common` and add two sub trait (`KyuubiServerEvent`, `KyuubiSparkEvent`) . 
  `SPARK` event logger is only supported in `KyuubiSparkEvent`.
2. Move `EventLogger` and `JsonEventLogger` to `kyuubi-common`
3. Create `AbstractEventLoggingService`, `kyuubi-server` and `kyuubi-engine` each has a subclass `EventLoggingService` to get their own event loggers and catch events.
4. Add `OperationEvent` for server

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -682,13 +682,36 @@ object KyuubiConf {
...
+        s" <li>CUSTOM: to be done.</li></ul>")
+      .version("1.4.0")
+      .stringConf
+      .transform(_.toUpperCase(Locale.ROOT))
+      .toSequence()
+      .checkValue(_.toSet.subsetOf(Set("JSON", "JDBC", "CUSTOM")),
+        "Unsupported event loggers")
+      .createWithDefault(Nil)
+
   val ENGINE_EVENT_LOGGERS: ConfigEntry[Seq[String]] =
```
yaooqinn, line 715: Do we have time to rename those configs to `kyuubi.event.[server|engine].xxx` before 1.3.0? cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/events/StatementEvent.scala
```diff
@@ -0,0 +1,66 @@
...
+ * @param statementId: the identifier of operationHandler
+ * @param statement: the sql that you execute
+ * @param remoteIp: the ip of user
+ * @param sessionId: the identifier of a session
+ * @param createTime: the create time of this statement
+ * @param state: store each state that the sql has
+ * @param stateTime: the time that the sql's state change
+ * @param exception: caught exception if have
+ */
+case class StatementEvent(
```
ulysses-you, line 36: This enigne type is conflict with engine's `StatementEvent` if we write into same path.

kyuubi-common/src/main/scala/org/apache/kyuubi/events/EventLoggingService.scala
```diff
@@ -0,0 +1,53 @@
...
+   * Get configured event log type.
+   */
+  protected def getLoggers(conf: KyuubiConf): Seq[EventLoggerType]
+
+  def onEvent(event: T): Unit = {
+    eventLoggers.foreach(_.logEvent(event))
+  }
+
+  override def initialize(conf: KyuubiConf): Unit = {
+    getLoggers(conf).foreach(analyseEventLoggerType)
```
ulysses-you, line 41: It seems overkill to add two methods here, how about override `initialize` in implementation ?

kyuubi-server/src/main/scala/org/apache/kyuubi/events/StatementEvent.scala
hddong: > This event type is conflict with engine's `StatementEvent` if we write into same path.

`StatementEvent` is a path too, they have differnet file name under `statement`. 
And add a check for this case. 
```
ll /var/folders/xl/jv1jvw6s6hv42ht1ty6gnk100000gn/T/kyuubi-bb4a5f34-451c-4783-9f50-1906889386cf/statement/day=20210824/
-rw-rw----  1 *  staff  9472 Aug 24 16:27 local-1629793620908.json
-rw-rw----  1 *  staff  3588 Aug 24 16:27 server-172.20.144.37.json
```

kyuubi-common/src/main/scala/org/apache/kyuubi/events/EventLoggingService.scala
hddong: yeah, change to override `initialize`.

kyuubi-server/src/main/scala/org/apache/kyuubi/events/StatementEvent.scala
ulysses-you: Since they have different schema, if we want to create table on that path, how can we read them together ? I think we should use different event type for this case so that we can create tables for them.

hddong: > Since ... I think we should use different event type for this case so that we can create tables for them.

Added prifix for StatementEvents.

ulysses-you (approved):

ulysses-you:
thank you @hddong , master to master for 1.4.0.

Can you send a new PR for branch-1.3.0 only to fix the typo issue ?
`var exeception: String = "") extends KyuubiEvent {` -> `var exception: String = "") extends KyuubiEvent {`

hddong:
@ulysses-you : Sure, I will.

kyuubi-common/src/main/scala/org/apache/kyuubi/events/KyuubiEvent.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.apache.kyuubi.events
+
+import java.util.Locale
+
+trait KyuubiEvent extends Product {
+  final lazy val eventType: String = {
+    this.getClass.getSimpleName.stripSuffix("Event")
+      .replaceAll("(.)([A-Z])", "$1-$2")
```
yaooqinn, line 25: can we use `"$1_$2"`? `-` is unfriendly to sql

hddong: > can we use `"$1_$2"`? `-` is unfriendly to sql

Yep.

Issue #962, opened 2021-08-19
Support multiple engines under USER share level


timothy65535:
### 1. Description

Kyuubi already supports `CONNECTION`, `USER`, `SERVER` share levels. A `subDomain` is a kind of MANUAL engine pool. It helps a user create more engines in user mode but require user assigns a unique character sequence. Based on #933, we can start to implement the feature of the engine pool. 

![image](https://user-images.githubusercontent.com/86483005/128994314-c978bf61-6a8f-4d95-9ea6-9fb96ca786cc.png)

### 2. Things to consider

- Which share level to set the pool size?
- How to control the concurrent create the engine?
- How to share the pool size over two more kyuubi server? 
- Selection strategy(sequentially / randomly / load-based)?

### 3. Additional context

This needs a full design, any thoughs are welcome.

@wForget this issue is a bit overlapped with #786, we can split this to several tasks if possible 😃.

wForget:
> How to control the concurrent create the engine?

Is it possible to use `InterProcessSemaphoreV2` to control concurrency.

wForget:
> How to share the pool size over two more kyuubi server?

The pool size should be configured through the user default configuration, and the user default configuration of multiple Kyuubi Servers should be consistent.

yaooqinn:
Can we create a proposal like what we did for the KYUUBI REST APIs https://docs.google.com/document/d/1yc-fk28oWbCNZjdFr1UXFbMtwSDQTpcwcNObn9ODnhQ/edit

You can make a `draft` on the google doc first, and we can polish it together before coding. Once it's been done, we can move it to https://github.com/apache/incubator-kyuubi/discussions/categories/ideas as a KPIP.

timothy65535:
thanks @wForget, will open a github discussion first.

@yaooqinn, can we use [github discussions](https://github.com/apache/incubator-kyuubi/discussions/categories/ideas) directly? Google is restricted, can not open the google doc.

yaooqinn:
you use any public doc which can be viewed and modified by others. gh discussion is only modifiable for committers and author

timothy65535:
hi @yaooqinn, does kyuubi has apache wiki space https://wiki.apache.org/? 

> and modified by others.

This maybe is not necessary. ok, will try to find online public doc, or put design on this issue first.

timothy65535:
If github issue assigns to other users, can others modify it?

[...]

PR #974, opened 2021-08-22
[KYUUBI #962] Support multiple engines under USER share level

timothy65535:
# Motivation

Apache Kyuubi provides three engine share levels:

- CONNECTION: engine will not be shared but only used by the current client connection
- SERVER: the App will be shared by Kyuubi servers
- USER: engine will be shared by all sessions created by a unique username

In the current design, kyuubi can provides engine pool through `kyuubi.engine.share.level.sub.domain` option for same user under `USER` share level. It does not have an auto-mechanisn, e.g. create two more engines under same `subDomain`, random picking engines.  The goal of this proposal is to provide such auto-mechanism to creating and picking subDomain to implement the feature of the engine pool.

# Implementation

In this section, describe the implementation of Kyuubi Engine Pool. We need to consider several things

- How to control the concurrent create the engine?
- How to share the pool size over two more kyuubi server?
- Engine pick strategy(sequentially / randomly / load-based)?

The key point is to set engine pool size as the content of engine space znode when create the engine space firstlly, so that we can share the pool size over two more kyuubi server, and avoid the inconsistent configuration between different servers effectively.

### 1. High level 

image

### 2. Detail work flow

image

# Tests

- KyuubiOperationPerConnectionSuite
- KyuubiOperationPerServerSuite
- KyuubiOperationPerUserSuite


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -126,11 +130,33 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
-    // Get the engine address ahead if another process has succeeded
-    if (engineRef.nonEmpty) return engineRef.get
 
+    // USER share level support engine pool
+    if (shareLevel.equals(USER)) {
+
+      // create engine space
+      createEngineSpaceIfNotExists(zkClient, engineSpace, conf.get(HA_ZK_ENGINE_POOL_SIZE))
+
+      // create engine if not reach pool size
```
yaooqinn, line 140: How can we reach to pool size？

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
```diff
@@ -108,4 +109,19 @@ object HighAvailabilityConf {
     .version("1.4.0")
     .stringConf
     .createOptional
+
+  val HA_ZK_ENGINE_POOL_SIZE: ConfigEntry[Int] =
+    buildConf("ha.engine.pool.size")
+      .doc("Maximum number of engines when using USER share level.")
+      .version("1.4.0")
+      .intConf
+      .createWithDefault(1)
```
zwangsheng, line 118: How about use `checkValues` to make sure this configuration remains positive?

timothy65535: Make sense, thanks.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: When creating engine, will check `pool.size` and `engineNum`.

_`pool size`_: read from engine space.
_`engine num`_: children of engine space.


image

yaooqinn: can you add a test to verify this？

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ProvidePolicy.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package org.apache.kyuubi.engine
+
+object ProvidePolicy extends Enumeration {
+
+  type ProvidePolicy = Value
+
+  val
+  /** A random policy that pick engines randomly. */
+  RANDOM = Value
```
cfmcgrady, line 27: Only `RANDOM` policy support now?

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
```diff
@@ -21,13 +21,15 @@ import org.scalatest.time.SpanSugar._
...
 import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.ha.HighAvailabilityConf
 
 class KyuubiOperationPerUserSuite extends WithKyuubiServer with JDBCTests {
 
   override protected def jdbcUrl: String = getJdbcUrl
 
   override protected val conf: KyuubiConf = {
     KyuubiConf().set(KyuubiConf.ENGINE_SHARE_LEVEL, "user")
+    KyuubiConf().set(HighAvailabilityConf.HA_ZK_ENGINE_POOL_SIZE, 1)
```
cfmcgrady, line 32: It's better to add more test cases for `HA_ZK_ENGINE_POOL_SIZE > 1`?

timothy65535: Make sense, thanks.

Had updated the test case.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
```diff
@@ -52,3 +61,59 @@ class EngineServiceDiscovery private(
...
+
+  val mapper: ObjectMapper = new ObjectMapper().registerModule(DefaultScalaModule)
+
+  def getEngineByPolicy(
+      zkClient: CuratorFramework,
+      namespace: String,
+      providePolicy: ProvidePolicy): Option[(String, Int)] = {
+    providePolicy match {
+      case RANDOM =>
+        Random.shuffle(getServiceNodesInfo(zkClient, namespace, silent = true)) match {
```
zwangsheng, line 75: I have a question, please correct me if there is any problem.Here All engines under namespce are obtained. If there is one engine, the create phase is not entered, and the engine state is always one.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ProvidePolicy.scala
timothy65535: Yeah, we can add other policy later.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: Updated `KyuubiOperationPerUserSuite`, I am not sure if the coding is elegant.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
timothy65535: Just wait a moment, try to understand your comment.

timothy65535: @zwangsheng, you're right, had improved the logic.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
```diff
@@ -30,29 +31,68 @@ class KyuubiOperationPerUserSuite extends WithKyuubiServer with JDBCTests {
...
+          assert(res.next())
+          r2 = res.getString("value")
+        }
+      }.start()
+
+      eventually(timeout(120.seconds), interval(100.milliseconds)) {
+        assert(r1 != null && r2 != null)
+      }
+
+      assert(r1 != r2)
```
yaooqinn, line 94: can we add a test for connect 3？

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
yaooqinn: So, the pool actually makes connect 2 a very worse case
1)  it always needs to create a new engine even there is an available engine there. the time cost is much higher than just reusing.
2)  it will fail if it fails to create a new engine. The failure risk is much higher to create more engines.
3) if we can fall back to reusing after failures in 2), the time cost is inevitable.

If the failure happens, connection 3 seems very like to repeat what connection 2 does.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
```diff
@@ -52,3 +61,59 @@ class EngineServiceDiscovery private(
     super.stop()
   }
 }
+
+object EngineServiceDiscovery extends Logging {
```
yaooqinn, line 65: EngineServiceDiscovery is only visited at the engine side before

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -125,12 +128,34 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
       }
   }
 
-  private def create(zkClient: CuratorFramework): (String, Int) = tryWithLock(zkClient) {
-    // TODO: improve this after support engine pool. (KYUUBI #918)
-    var engineRef = getServerHost(zkClient, engineSpace)
-    // Get the engine address ahead if another process has succeeded
-    if (engineRef.nonEmpty) return engineRef.get
+  def getOrCreate(zkClient: CuratorFramework): (String, Int) = tryWithLock(zkClient) {
```
yaooqinn, line 131: Hmm, we need to wait for the lock every time..

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
timothy65535: EngineServiceDiscovery still visited at the engine side in this patch.

yaooqinn: image

timothy65535: ? Isn't EngineServiceDiscovery still visited at the engine side in this patch?

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
timothy65535: make sense, had updated.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: Thanks for raising concerns:

> it always needs to create a new engine even there is an available engine there

This is the characteristic of the auto-mechanism, we can provide option to skip create a new engine.

> if we can fall back to reusing after failures in 2), the time cost is inevitable.

This work as `subDomain`. When using `subDomain` and encountering failure, time-consuming is also unavoidable.

**Improve**

- provide option to skip create a new engine
- fall back to reusing after creating failure

zwangsheng:
I'd like to ask if we have the configuration of engine expiration recycling?

timothy65535:
> I'd like to ask if we have the configuration of engine expiration recycling?

Hi, my understanding is that there is no such configuration currently, we can clear it through kyuubi-ctl, hope answered your question.

zwangsheng:
> Hi, my understanding is that there is no such configuration currently, we can clear it through kyuubi-ctl, hope answered your question.

I agree that the idle engine can be cleared through Kyuubi cli, but considering that Kyuubi is a resident service, maybe the PR can mention relevant parameters later?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
yaooqinn: > This is the characteristic of the auto-mechanism, we can provide option to skip create a new engine.

It makes the feature itself even more unfriendly.

> This work as subDomain. When using subDomain and encountering failure, time-consuming is also unavoidable.

Not really the same. the original intention of a subdomain is to provide isolation but creating multiple engines under a subdomain is for speed and power. But now, it does bring extra creation overhead for a lot of cases

timothy65535:
> > Hi, my understanding is that there is no such configuration currently, we can clear it through kyuubi-ctl, hope answered your question.
> 
> I agree that the idle engine can be cleared through Kyuubi cli, but considering that Kyuubi is a resident service, maybe the PR can mention relevant parameters later?

ok, will find a suitable location to mention relevant parameters.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: For most scene, we use sparksql as offline analysis, `engine pool` is useful for us. Ok, if user very case about the time cost, keep `kyuubi.ha.engine.pool.size=1`, the kyuubi server will work like before.

yaooqinn: Not exactly too, you add a zookeeper distribute lock for all connections, this is much heavier than before

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
```diff
@@ -30,29 +33,118 @@ class KyuubiOperationPerUserSuite extends WithKyuubiServer with JDBCTests {
...
-    }.start()
 
-    eventually(timeout(120.seconds), interval(100.milliseconds)) {
-      assert(r1 != null && r2 != null)
+      assert(r1 != r2)
     }
+  }
 
-    assert(r1 === r2)
+  test("ensure two of three connections share the same engine when engine pool size is 2.") {
```
yaooqinn, line 101: Can we add another case that connection 2, 3 concurrently created while connection 1's engine creation is finished

timothy65535: Let me have a try.

timothy65535: > Can we add another case that connection 2, 3 concurrently created while connection 1's engine creation is finished

hi @yaooqinn, what's the aim of the test case? what will we assert?

yaooqinn: we will exactly create 2 engines and one for 2 connections and the other for 1 connection

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
yaooqinn: we are talking about the engine pool itself, for what I am trying to improve based on your current implementation. I am not sure if my concerns can be ignored or persuade like that.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
yaooqinn: I don't get it

yaooqinn: the logging is not used anywhere too

yaooqinn: Now the class EngineServiceDiscovery is for the engine module to visit AS-IS, but the new object EngineServiceDiscovery is used by the server module

timothy65535: The `EngineRef` exists in the server module, so `engine module` == `server module` from current codebase.

Please correct me if wrong, thanks.

image

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: Yeah, I am thinking how to improve this.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
timothy65535: Had updated the test case, correct me if wrong.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
yaooqinn: I mean module  like
image not a package

dev/dependencyList
```diff
@@ -29,6 +29,8 @@ htrace-core4/4.1.0-incubating//htrace-core4-4.1.0-incubating.jar
 jackson-annotations/2.11.4//jackson-annotations-2.11.4.jar
 jackson-core/2.11.4//jackson-core-2.11.4.jar
 jackson-databind/2.11.4//jackson-databind-2.11.4.jar
+jackson-module-paranamer/2.11.4//jackson-module-paranamer-2.11.4.jar
```
pan3793, line 32: We need to update LICENSE-binary once `dependencyList` changed

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
timothy65535: :sweat_smile: Maybe I am not clear-headed now, will back here tomorrow.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
```diff
@@ -108,4 +109,20 @@ object HighAvailabilityConf {
     .version("1.4.0")
     .stringConf
     .createOptional
+
+  val HA_ZK_ENGINE_POOL_SIZE: ConfigEntry[Int] =
+    buildConf("ha.engine.pool.size")
+      .doc("Maximum number of engines when using USER share level.")
+      .version("1.4.0")
+      .intConf
+      .checkValue(s => s > 0 && s < 50, "Invalid Engine Pool Size.")
```
pan3793, line 118: `"Invalid Engine Pool Size."` => `"Invalid Engine Pool Size, should be between 0 and 50"`

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
```diff
@@ -108,4 +109,20 @@ object HighAvailabilityConf {
...
+      .createWithDefault(1)
+
+  val HA_ZK_ENGINE_PROVIDE_POLICY: ConfigEntry[String] =
+    buildConf("ha.engine.provide.policy")
+      .doc("Multiple engines can be exposed under same space when using USER share level, " +
+        "choose the appropriate strategy according to different scenarios.")
+      .version("1.4.0")
+      .stringConf
+      .createWithDefault(ProvidePolicy.RANDOM.toString)
```
pan3793, line 127:
```
    .stringConf
    .transform(_.toUpperCase(Locale.ROOT))
    .checkValues(ProvidePolicy.values.map(_.toString))
```

dev/dependencyList
timothy65535: Thanks for reviewing, done.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
timothy65535: Good !!!

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/HighAvailabilityConf.scala
timothy65535: 👍

pan3793: I think you missed `.checkValues(ProvidePolicy.values.map(_.toString))`

dev/dependencyList
pan3793: Please open the file `LICENSE-binary` and scroll down to the bottom, we also need to list the deps there.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
```diff
@@ -52,3 +61,59 @@ class EngineServiceDiscovery private(
...
+
+  val mapper: ObjectMapper = new ObjectMapper().registerModule(DefaultScalaModule)
+
+  def getEngineByPolicy(
+      zkClient: CuratorFramework,
+      namespace: String,
+      providePolicy: ProvidePolicy): Option[(String, Int)] = {
+    providePolicy match {
+      case RANDOM =>
+        Random.shuffle(getServiceNodesInfo(zkClient, namespace, silent = true)).take(1) match {
```
pan3793, line 75: It can be simplified by `Seq.headOption`

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
```diff
@@ -52,3 +61,59 @@ class EngineServiceDiscovery private(
...
+          case Seq(sn) => Some((sn.host, sn.port))
+          case _ => None
+        }
+      case _ => throw new IllegalArgumentException(s"Not support provide policy $providePolicy")
```
pan3793, line 79: Seems not required to add a default branch for enum, pls remove it and see if the compiler print any warnings

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -125,12 +128,34 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
+      // create engine space
+      createEngineSpaceIfNotExists(zkClient, engineSpace, conf.get(HA_ZK_ENGINE_POOL_SIZE))
+
+      // create engine if not reach pool size
+      val notFull = checkEnginePoolCapacity(zkClient, engineSpace)
+      if (notFull) {
+        createInternal(zkClient)
+      } else {
+        val engineRef = getEngineByPolicy(zkClient, engineSpace, providePolicy)
+        if (engineRef.nonEmpty) return engineRef.get
```
pan3793, line 145: replace by pattern
```
engineRefOpt match {
  case Some(engineRef) => ...
  case None => ...
}
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -125,12 +128,34 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
+        else throw new Exception("Unexpected exception!!!")
```
pan3793, line 146: Avoid using raw `Exception`, pls pick a specific `KyuubiException` or create a new subclass if there is no suitable one.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: 👍

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
timothy65535: > EngineServiceDiscovery is only visited at the engine side before

Get your point. just like class `ServiceDiscovery` is visited in engine side and object `ServiceDiscovery` is visited in servier side before. Will follow you if you have a better suggestion.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: When support two more engines under same space, we‘ll face concurrency issues. Had tried serveral ways to improve the lock, but failed. Clients will wait for the lock only when connect to the kyuubi server, and these connections are long-time connections.

yaooqinn:
#### 一些想法

##### 什么是引擎池？

一个引擎池是面向一个租户的，不同的用户使用不同的引擎池。一个引擎池是在一个目录下或者根据相应的目录过滤条件可发现的引擎集合。

假设我们为tom和jerry两位用户开启了引擎池。前者的大小为2，后者的大小为3。

所以，tom的引擎池看起来是这样的

```
kyuubi_USER/tom/engine-pool-0/serviceUri=ip:port;version=1.3.0;sequence=..../data
kyuubi_USER/tom/engine-pool-1/serviceUri=ip:port;version=1.3.0;sequence=..../data
```

而jerry的引擎池则看起来是这样的

```
kyuubi_USER/jerry/engine-pool-0/serviceUri=ip:port;version=1.3.0;sequence=..../data
kyuubi_USER/jerry/engine-pool-1/serviceUri=ip:port;version=1.3.0;sequence=..../data
kyuubi_USER/jerry/engine-pool-2/serviceUri=ip:port;version=1.3.0;sequence=..../data
```

注意，这里的整个“目录”设计规则和Kyuubi目前的规则没有任何的区别。其中，

 - 第一层目录为server space加下划线再加share level，这里的share_level会是USER/Server。
 - 第二层目录为用户名
 - 第三层目录为subDomain。这里使我们需要关注的重点，因为引擎池化的逻辑主要围绕这一层目录的查找、创建暂开。我们可以复用和拓展现有的并发控制逻辑。在我们给的例子中’engine-pool-‘没有特别的含义，我们后面可以将它设置为参数或者hardcode的字符串前缀。而后缀则是0到pool size -1的自然数序列。如上面的例子所示，tom的引擎池会产生2个subDomain的目录。
- 剩余的目录就和现在的设计一样了，需强调的是和以前一样一个目录下依然只维护一个engine实例


##### 引擎池大小控制

- `kyuubi.engine.pool.size.threshold`, 引入该参数为服务端参数，由服务端控制引擎池的绝对上限。如果用户的设置超过这个设置值，则报错或者取该值
- `kyuubi.engine.pool.size`，引入该参数为用户和服务端都可设置的参数，参数取值范围在`kyuubi.engine.pool.size.threshold`以下。当服务端统一或者用用户默认参数设置该值时，服务端会在统计当前存在subDomain目录数，即当前池大小，如果还未达预期，则递增一个subDomain目录并在该目录下创建一个引擎。

引擎的生命周期和现在的实现不变，可依据现在的缓存方式进行自我缓存，当生命周期到达上限，则自己退出并删除自己所属的subDomain


##### 引擎的创建和复用 (RANDOM)

在 RANDOM 模式下，一切都是随机的，引擎的复用和引擎的创建在这个 POLICY 下是等价的。

举例，当jerry 在随机算法下使用连接池，假设 engine pool 的中间值为如下所示的状态，槽0和2已经有 engine 实例，而槽1 还未被初始化

```
kyuubi_USER/jerry/engine-pool-0/serviceUri=ip:port;version=1.3.0;sequence=..../data
kyuubi_USER/jerry/engine-pool-1/
kyuubi_USER/jerry/engine-pool-2/serviceUri=ip:port;version=1.3.0;sequence=..../data
```
此时，当 jerry 尝试连接时，

- jerry 显示配置sub.domain，几个特例
  - sub.domain=engine-pool-0, 复用连接池内槽0的实例
  - sub.domain=engine-pool-1，**只需对 engine-pool-1 加锁**，直接创建槽1的实例，“手动”完成初始化，并连接
  - sub.domain=someothers，绕过引擎池的使用
- jerry 未配置sub.domain， 则随机算法生成 engine-pool-x 
  - 当x=0时，复用连接池内槽0的实例
  - 当x=1时，只需对 engine-pool-1 加锁，创建槽1的实例，完成初始化
  - 当x=2时，复用连接池内槽1的实例
- jerry 使用其他模式，照旧


##### 两个 Kyuubi Server 对某用户配置了不同的 kyuubi.engine.pool.size 值怎么办？

假如 Server1 为4，Server2 为 5，则Server1，在[0, 3]区间内随机，而Server2可在[0, 4]之间取值。两者不存在耦合关系，当然最好的实践还是配置相同的参数值

##### kyuubi.engine.pool.size 是否对用户暴露，暴露后用户连接每次 kyuubi.engine.pool.size 不一致怎么办？不合理设置怎么办？

理论上可以对用户直接暴露，
当用户设置不合理时，比如过大，则由kyuubi.engine.pool.size.threshold 控制
当用户设置不一致时，比如增加，则 pool size 可随机增加。抑或减小，则新区间外的引擎将因随机不到而自我回收。

timothy65535:
@yaooqinn 优秀，思路清奇 👍，尝试实现一把。

同时，多谢各位的REVIEW、点评，收获很多的建议，包括逻辑、SCALA语法等。

Issue #979, opened 2021-08-24
[SUB-TASK][KPIP-1] Introduce AbstractFrontendService and KyuubiFrontendService


yanghua:
See the design document here #917 .

PR #980, opened 2021-08-24
[KYUUBI #979] Introduce AbstractFrontendService and KyuubiFrontendService

yanghua:

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -33,15 +33,17 @@ import org.apache.kyuubi.engine.spark.SparkSQLEngine.countDownLatch
 import org.apache.kyuubi.engine.spark.events.{EngineEvent, EventLoggingService}
 import org.apache.kyuubi.ha.HighAvailabilityConf._
 import org.apache.kyuubi.ha.client.{EngineServiceDiscovery, RetryPolicies, ServiceDiscovery}
-import org.apache.kyuubi.service.{Serverable, Service, ServiceState}
+import org.apache.kyuubi.service.{Serverable, Service, ServiceState, ThriftFrontendService}
 import org.apache.kyuubi.util.SignalRegister
 
 case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngine") {
 
   lazy val engineStatus: EngineEvent = EngineEvent(this)
 
+  private val OOMHook = new Runnable { override def run(): Unit = stop() }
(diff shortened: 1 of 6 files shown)


yanghua:
@yaooqinn

kyuubi-common/src/main/scala/org/apache/kyuubi/service/AbstractFrontendService.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package org.apache.kyuubi.service
+
+/**
+ * A basic abstraction for frontend services.
+ */
+abstract class AbstractFrontendService(name: String, be: BackendService)
+  extends CompositeService(name) {
+
+  def connectionUrl(server: Boolean = false): String
```
yaooqinn, line 26: shouldn't it be a map?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -33,15 +33,17 @@ import org.apache.kyuubi.engine.spark.SparkSQLEngine.countDownLatch
...
 import org.apache.kyuubi.util.SignalRegister
 
 case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngine") {
 
   lazy val engineStatus: EngineEvent = EngineEvent(this)
 
+  private val OOMHook = new Runnable { override def run(): Unit = stop() }
   private val eventLogging = new EventLoggingService(this)
   override val backendService = new SparkSQLBackendService(spark)
+  override val frontendService = new ThriftFrontendService(backendService, OOMHook)
```
yaooqinn, line 46: we can create a `SparkSQLFrontendService` like `SparkSQLBackendService`, and we can remove the `server` flag which is a bit hacky

kyuubi-common/src/main/scala/org/apache/kyuubi/service/AbstractFrontendService.scala
yanghua: It would be a map, of cause. Currently, we have not introduced the `RestFrontendService` and relevant config options.
This PR only introduced a framework for the abstraction, e.g. we only let `KyuubiFrontendService` delegate thrift front service.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
yanghua: sounds good.

yanghua: Will open a separated pr to remove the `server` flag.

Issue #983, opened 2021-08-25
[SUB-TASK][KPIP-1] Refactor the basic abstraction of frontend service


yanghua:
See detailed design document: #917

PR #984, opened 2021-08-25
[KYUUBI #983] Refactor the basic abstraction of frontend service

yanghua:
…

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiFrontendService.scala
@@ -19,14 +19,14 @@ package org.apache.kyuubi.server
 
 import org.apache.kyuubi.Logging
 import org.apache.kyuubi.config.KyuubiConf
-import org.apache.kyuubi.service.{AbstractFrontendService, BackendService, ServiceState, ThriftFrontendService}
+import org.apache.kyuubi.service.{AbstractFrontendService, BackendService, CompositeService, ServiceState, ThriftFrontendService}
 
 /**
  * A kyuubi frontend service is a kind of composite service, which used to
  * composite multiple frontend services for kyuubi server.
  */
 class KyuubiFrontendService private(name: String, be: BackendService)
-  extends AbstractFrontendService(name, be) with Logging {
+  extends CompositeService(name) with Logging {
 
   private val OOMHook = new Runnable { override def run(): Unit = stop() }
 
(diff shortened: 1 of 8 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/service/ThriftFrontendService.scala
```diff
@@ -105,11 +105,11 @@ class ThriftFrontendService private(name: String, be: BackendService, oomHook: R
     super.initialize(conf)
   }
 
-  override def connectionUrl(server: Boolean = false): String = {
+  override def connectionUrl: String = {
     getServiceState match {
       case s @ ServiceState.LATENT => throw new IllegalStateException(s"Illegal Service State: $s")
       case _ =>
-        if (server || conf.get(ENGINE_CONNECTION_URL_USE_HOSTNAME)) {
+        if (conf.get(ENGINE_CONNECTION_URL_USE_HOSTNAME)) {
```
yaooqinn, line 112: this is an engine config, we need to create `SparkSQLFrontendService` to remove  the `server` flag

yanghua: Did you mean this PR introduce the side-effort? When diving into the code base, I found the server flag only be used at [here](https://github.com/apache/incubator-kyuubi/blob/master/kyuubi-common/src/main/scala/org/apache/kyuubi/service/ThriftFrontendService.scala#L131), right? Before this PR, the same logic is also been judged in `KyuubiServer`(KyuubiServer used ThriftFrontendService).

In fact, I haven't thought about the real role of SparkSQLFrontendService. Will it be a CompositeService? ThriftFrontendService will still be used by KyuubiServer. You mean we need to put the judgment logic of the engine configuration in CompositeService (SparkSQLFrontendService)?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiFrontendService.scala
```diff
@@ -19,30 +19,30 @@ package org.apache.kyuubi.server
...
 import org.apache.kyuubi.Logging
 import org.apache.kyuubi.config.KyuubiConf
-import org.apache.kyuubi.service.{AbstractFrontendService, BackendService, ServiceState, ThriftFrontendService}
+import org.apache.kyuubi.service.{AbstractFrontendService, BackendService, CompositeService, ServiceState, ThriftFrontendService}
 
 /**
  * A kyuubi frontend service is a kind of composite service, which used to
  * composite multiple frontend services for kyuubi server.
  */
 class KyuubiFrontendService private(name: String, be: BackendService)
```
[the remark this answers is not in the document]
yanghua: addressed

yaooqinn (approved):

yanghua:
@yaooqinn Can we merge this PR so that I can apply it in my local?

pan3793:
Thanks, merging to master

PR #985, opened 2021-08-25
Add logCaptureThreadReleased flag

ulysses-you:

`isInterrupted` is always false if some `InterruptedException` are throwed. So we should use a new flag to check if log capture thread is released.


ulysses-you:
all test passed with three times, it's not flaky now cc @yaooqinn @pan3793

pan3793 (approved):
LGTM, and modified the PR title to reflect more details of the change.

pan3793:
Thanks, merging to master/v1.3

PR #986, opened 2021-08-25
[KYUUBI #962] Support engine pool feature

timothy65535:

For detail, please go to

_Originally posted by @yaooqinn in https://github.com/apache/incubator-kyuubi/issues/962#issuecomment-905116261_


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -606,6 +606,36 @@ object KyuubiConf {
...
+    .checkValue(s => s > 0 && s <= 32, "Invalid Engine Pool Threshold Size, " +
+      "should be between 0 and 33.")
+    .createWithDefault(16)
+
+  val ENGINE_POOL_SIZE: ConfigEntry[Int] = buildConf("engine.pool.size")
+    .doc("Max number of engines that can be provided to clients.")
+    .version("1.4.0")
+    .intConf
+    .checkValue(s => s > 0, "Invalid Engine Pool Size, must be positive number.")
+    .createWithDefault(1)
```
zwangsheng, line 637: I have a question about why split it into two parameters here?

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -257,6 +240,30 @@ object ServiceDiscovery extends Logging {
     }
     serviceNode
   }
+
+  def getEngineByPolicy(
```
yaooqinn, line 244: we don't need this, this has already been done at `Some("engine-pool-" + Random.nextInt(poolSize))`, we can use the old logic to get or create engine directly

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
timothy65535: I think the option `kyuubi.engine.pool.size.threshold` cat give users the opportunity to set the threshold.

if we use `checkValue`, the `50` is hardcode.
`.checkValue(s => s > 0 && s < 50, "Invalid Engine Pool Size, should be between 0 and 50")`

zwangsheng: I can accept this, but why the threshold of ENGINE_POOL_SIZE_THRESHOLD is 32?

timothy65535: For the threshold`32`, willing to talk with community.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -57,6 +59,16 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
   private val shareLevel: ShareLevel = ShareLevel.withName(conf.get(ENGINE_SHARE_LEVEL))
 
   private val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN)
+    .orElse {
+      if (conf.get(ENGINE_POOL_ENABLED)) {
```
yaooqinn, line 63: how about we save a config here?  if the final poolSize > 1

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
timothy65535: Make sense.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
yaooqinn: How about?

for server-side config `engine.pool.size.threshold`, we can leave it to the system administrators to set, I vote positive numbers as valid values and maybe 9(which also means `kyuu`LOL) as default

 for user-faced config `engine.pool.size` maybe we need not check it, and use -1 by default to turn off the pool featuer

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: hi, you mean remove `ENGINE_POOL_ENABLED` and use `poolSize > 1` to determine whether to enable engine pool?

yaooqinn: yes

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -57,6 +59,16 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
+        // engine.pool.size.threshold is a server-side parameter
+        val poolThreshold = KyuubiConf().loadFileDefaults().get(ENGINE_POOL_SIZE_THRESHOLD)
+        val poolSize = conf.get(ENGINE_POOL_SIZE).min(poolThreshold)
+        Some("engine-pool-" + Random.nextInt(poolSize))
```
yaooqinn, line 67: We can add a TODO here for pooling policies. Currently, it is random and later we can add a sequentially policy such as `AtomicInteger % poolSize`, and son on.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -57,6 +59,16 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
   private val shareLevel: ShareLevel = ShareLevel.withName(conf.get(ENGINE_SHARE_LEVEL))
 
   private val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN)
+    .orElse {
+      if (conf.get(ENGINE_POOL_ENABLED)) {
+        // engine.pool.size.threshold is a server-side parameter
+        val poolThreshold = KyuubiConf().loadFileDefaults().get(ENGINE_POOL_SIZE_THRESHOLD)
```
yaooqinn, line 65: this is a bit tricky and heavy, please refine this

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: make sense.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
timothy65535: agree.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -56,7 +58,20 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
   // Share level of the engine
   private val shareLevel: ShareLevel = ShareLevel.withName(conf.get(ENGINE_SHARE_LEVEL))
 
+  // This a is a server-side option
+  val poolThreshold: Int = KyuubiConf().loadFileDefaults().get(ENGINE_POOL_SIZE_THRESHOLD)
```
yaooqinn, line 62: this is not what I mean...

timothy65535: :sweat_smile:

yaooqinn: the right way is to get it from `sessionConf` of KyuubiSessionImpl before merging the client configs.. then the system defaults and user defaults are respected(we can set different threshold for different users)

timothy65535: Got.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -56,7 +58,21 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
+      Option(conf.get(ENGINE_POOL_SIZE))
+        .filter(size => size > 0)
+        .map(size => {
+          val poolSize = size.min(poolThreshold)
+
+          // TODO: Currently, we use random policy,
+          //  and later we can add a sequential policy, such as AtomicInteger % poolSize.
+          "engine-pool-" + Random.nextInt(poolSize)
+        })
+    }
```
yaooqinn, line 75:
```suggestion
    private lazy val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN).orElse {
    val clientPoolSize: Int = conf.get(ENGINE_POOL_SIZE)
    if (conf.get(ENGINE_POOL_SIZE) <= 1) {
      None
    } else {
      val poolSize = if (clientPoolSize <= poolThreshold) {
        clientPoolSize
      } else {
        warn(s"Request engine pool size($clientPoolSize) exceeds, fallback to system threshold " +
          s"$poolThreshold")
        poolThreshold
      }
      // TODO: Currently, we use random policy,
      //  and later we can add a sequential policy, such as AtomicInteger % poolSize.
      Some("engine-pool-" + Random.nextInt(poolSize))
    }
  }
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -56,7 +58,21 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
   private val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN)
+    .orElse {
+      Option(conf.get(ENGINE_POOL_SIZE))
+        .filter(size => size > 0)
+        .map(size => {
+          val poolSize = size.min(poolThreshold)
+
+          // TODO: Currently, we use random policy,
+          //  and later we can add a sequential policy, such as AtomicInteger % poolSize.
+          "engine-pool-" + Random.nextInt(poolSize)
```
yaooqinn, line 73: ENGINE_SHARE_LEVEL_SUB_DOMAIN has restrictions for the values,`^[a-zA-Z_]{1,10}$`, we need improve that as the prefix `engine-pool-` breaks this rule

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: Nice.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
timothy65535: Got

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -572,7 +572,7 @@ object KyuubiConf {
     .checkValues(ShareLevel.values.map(_.toString))
     .createWithDefault(ShareLevel.USER.toString)
 
-  private val validEngineSubDomain: Pattern = "^[a-zA-Z_]{1,10}$".r.pattern
+  private val validEngineSubDomain: Pattern = "^[a-zA-Z_-]{1,14}$".r.pattern
```
yaooqinn, line 575: we also need to update its doc

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -606,6 +606,21 @@ object KyuubiConf {
...
+    .doc("This parameter is introduced as a server-side parameter, " +
+      "and controls the upper limit of the engine pool.")
+    .version("1.4.0")
+    .intConf
+    .checkValue(s => s > 0 && s < 32, "Invalid Engine Pool Threshold Size, " +
+      "should be between 0 and 33.")
+    .createWithDefault(9)
+
+  val ENGINE_POOL_SIZE: ConfigEntry[Int] = buildConf("engine.pool.size")
+    .doc("Max number of engines that can be provided to clients.")
```
yaooqinn, line 619: we can also describe the relationship between the threshold and itself, and also how it works, e.g.  if size <= 1 balalah.. if balalalah

timothy65535: ok

yaooqinn:
please also 
1) add some specific tests in `EngineRefSuite`for the changes in `EngineRef`
2) create `class KyuubiOperationEnginePoolSuite extends WithKyuubiServer with JDBCTestUtils` like `KyuubiOperationPerConnectionSuite` with some e2e tests for verification

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -606,6 +606,21 @@ object KyuubiConf {
     .version("1.2.0")
     .fallbackConf(LEGACY_ENGINE_SHARE_LEVEL)
 
+  val ENGINE_POOL_SIZE_THRESHOLD: ConfigEntry[Int] = buildConf("engine.pool.size.threshold")
+    .doc("This parameter is introduced as a server-side parameter, " +
+      "and controls the upper limit of the engine pool.")
+    .version("1.4.0")
+    .intConf
+    .checkValue(s => s > 0 && s < 32, "Invalid Engine Pool Threshold Size, " +
```
yaooqinn, line 614: Invalid engine pool threshold size

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -606,6 +606,21 @@ object KyuubiConf {
...
+      "should be between 0 and 33.")
```
yaooqinn, line 615: 31?

timothy65535: will improve the doc

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -572,12 +572,12 @@ object KyuubiConf {
...
     .createWithDefault(ShareLevel.USER.toString)
 
-  private val validEngineSubDomain: Pattern = "^[a-zA-Z_]{1,10}$".r.pattern
+  private val validEngineSubDomain: Pattern = "^[a-zA-Z_-]{1,14}$".r.pattern
 
   val ENGINE_SHARE_LEVEL_SUB_DOMAIN: OptionalConfigEntry[String] =
     buildConf("engine.share.level.sub.domain")
       .doc("Allow end-users to create a sub-domain for the share level of an engine. A" +
-        " sub-domain is a case-insensitive string values in `^[a-zA-Z_]{1,10}$` form." +
+        " sub-domain is a case-insensitive string values in `^[a-zA-Z_-]{1,14}$` form." +
```
cfmcgrady, line 580: We also need to update the error message of `checkValue`.

```
"must be [1, 14] length alphabet string, e.g. 'abc', 'apache'"
```

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationEnginePoolSuite.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.kyuubi.WithKyuubiServer
+import org.apache.kyuubi.config.KyuubiConf
+
+class KyuubiOperationEnginePoolSuite extends WithKyuubiServer with JDBCTests {
+
+  override protected def jdbcUrl: String = getJdbcUrl
+
+  override protected val conf: KyuubiConf = {
+    KyuubiConf().set(KyuubiConf.ENGINE_SHARE_LEVEL, "user")
```
cfmcgrady, line 30: The sub-domain doesn't work with the `CONNECTION` share level, do we need a unit test for this case?

timothy65535: Sounds resonable.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationEnginePoolSuite.scala
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  test("ensure app name contains engine-pool when engine pool is enabled.") {
+    withSessionConf()(
+      Map(
+        KyuubiConf.ENGINE_POOL_SIZE.key -> "2"
+      ))(Map.empty) {
+
+      var r1: String = null
```
yaooqinn, line 40: we do not need a new thead here

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -56,7 +58,27 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
...
   private val shareLevel: ShareLevel = ShareLevel.withName(conf.get(ENGINE_SHARE_LEVEL))
 
-  private val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN)
+  // This a is a server-side option
+  private val poolThreshold: Int = KyuubiConf().loadFileDefaults().get(ENGINE_POOL_SIZE_THRESHOLD)
+
+  @VisibleForTesting
+  private[kyuubi] val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN).orElse {
+    val clientPoolSize: Int = conf.get(ENGINE_POOL_SIZE)
+    if (conf.get(ENGINE_POOL_SIZE) <= 1) {
```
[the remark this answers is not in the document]
yaooqinn: I do not have strong opinion on this，both are OK to me

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -56,7 +58,27 @@ private[kyuubi] class EngineRef private(conf: KyuubiConf, user: String, sessionI
   // Share level of the engine
   private val shareLevel: ShareLevel = ShareLevel.withName(conf.get(ENGINE_SHARE_LEVEL))
 
-  private val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN)
+  // This a is a server-side option
+  private val poolThreshold: Int = KyuubiConf().loadFileDefaults().get(ENGINE_POOL_SIZE_THRESHOLD)
+
+  @VisibleForTesting
+  private[kyuubi] val subDomain: Option[String] = conf.get(ENGINE_SHARE_LEVEL_SUB_DOMAIN).orElse {
```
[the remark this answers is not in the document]
timothy65535: Thanks, seems a little hard to understand. +1 to https://github.com/apache/incubator-kyuubi/pull/986#discussion_r695687989

Issue #939, opened 2021-08-17
[Umbrella] Add Z-Order extensions to optimize table with zorder


qphien:
# Describe the proposal
To perform highly selective file pruning and make queries more efficient, files must be mutually exclusive in their data. Z-order  is a popular data skipping technique using in a variety of RDBMS systems as well as cloud based databases to enhance query efficiency with complex filters.

databricks docs:
https://docs.databricks.com/spark/2.x/spark-sql/language-manual/optimize.html

## zorder work
There is some work to do with zorder:

- Support SQL: `OPTIMIZE <table> [WHERE <predicate>] ZORDER BY (col1, col2, ...)`
   Optimize data matching the given partition predicate. `WHERE` clause only supports partition filters.
- Support SQL: `CREATE <table> ... ZORDER BY (col1, col2, ...)`
   Data will be sorted by zorder automatically before writing to zorder table.

## about test 
See the picture, the SQL extension test coverage is less than other module.
![image](https://user-images.githubusercontent.com/12025282/132435246-51404158-820b-4e34-9a4e-1ec778a07950.png)

# Task list

- #990
- #1045 
- #1055
- #1070
- #1157

yaooqinn:
Awsome!

ulysses-you:
To developers:

We merged zorder https://github.com/apache/incubator-kyuubi/pull/990 into master branch (for 1.4.0) recently. There are still many things can be done for the zorder and the test.

# zorder work

- support more relation (only `HiveTableRelation` for now)
- support more table format, since Spark doesn't support insert overwrite and read the same datasource table
- support optimzie zorder through tbl properteis; I think @qphien is working on this
- zorder expression with codegen
- zorder docs
- beanchmark of zorder for write side
- beanchmark of zorder for read side
- etc ...

# test coverage
we only have the end-end test of zorder, . as far as I know, we need more test include:

- unit test of analyzed zorder plan
- test different column data type
- test different partition data type
- test different table format
- etc ...

I wonder If you are interested in this feature. Look forward to your idea. @qphien  @cfmcgrady @hzxiongyinke @timothy65535 @hddong @zhaomin1423

PengleiShi:
[...]

PR #990, opened 2021-08-26
[KYUUBI #939] Add Z-Order extensions to support optimize SQL

qphien:
This PR adds Z-Order extension to support `OPTIMIZE` SQL in https://github.com/apache/incubator-kyuubi/issues/939.

Currently this PR only supports optimize `HiveTableRelation`.

We need to set `hive.exec.dynamic.partition` to `true` and `hive.exec.dynamic.partition.mode` to `nonstrict` when a partitioned table is optimized.

### DIFF
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/OptimizeZorderCommand.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/OptimizeZorderCommand.scala
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.sql.zorder
+
+import org.apache.spark.sql.catalyst.analysis.{UnresolvedAttribute, UnresolvedRelation}
+import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, Expression, NullsLast, SortOrder}
+import org.apache.spark.sql.catalyst.plans.logical.{Filter, LogicalPlan, Sort, UnaryNode}
+
+case class OptimizeZorderCommand(child: LogicalPlan) extends UnaryNode {
+  override def output: Seq[Attribute] = child.output
+}
+
+object OptimizeZorderCommand {
+
+  def apply(tableIdent: Seq[String],
+            whereExp: Option[Expression],
+            sortArr: Seq[UnresolvedAttribute]): OptimizeZorderCommand = {
+    val table = UnresolvedRelation(tableIdent)
+    val child = whereExp match {
+      case Some(x) => Filter(x, table)
+      case None => table
+    }
+
+    val sortOrder = SortOrder(Zorder(sortArr), Ascending, NullsLast, Seq.empty)
+    val zorderSort = Sort(Seq(sortOrder), true, child)
+    OptimizeZorderCommand(zorderSort)
+  }
+}
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/Zorder.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/Zorder.scala
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.sql.zorder
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{BoundReference, Expression}
+import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
+import org.apache.spark.sql.types.{BinaryType, BooleanType, ByteType, DataType, DateType, DecimalType, DoubleType, FloatType, IntegerType, LongType, ShortType, StringType, TimestampType}
+
+case class Zorder(children: Seq[Expression]) extends Expression with CodegenFallback {
+  private lazy val defaultNullValues = {
+    children.map {
+      case bf: BoundReference =>
+        bf.dataType match {
+          case BooleanType =>
+            false
+          case ByteType =>
+            Byte.MaxValue
+          case ShortType =>
+            Short.MaxValue
+          case IntegerType =>
+            Int.MaxValue
+          case LongType =>
+            Long.MaxValue
+          case FloatType =>
+            Float.MaxValue
+          case DoubleType =>
+            Double.MaxValue
+          case StringType =>
+            ""
+          case TimestampType =>
+            Long.MaxValue
+          case DateType =>
+            Int.MaxValue
+          case d: DecimalType =>
+            Long.MaxValue
+          case other: Any =>
+            throw new ZorderException("Unsupported z-order type: " + other.getClass)
+        }
+      case other: Any =>
+        throw new ZorderException("Unknown z-order column: " + other)
+    }
+  }
+
+  override def nullable: Boolean = false
+
+  override def eval(input: InternalRow): Any = {
+    val evaluated = children.zipWithIndex.map { case (child: Expression, index) =>
+      val v = child.eval(input)
+      if (v == null) {
+        defaultNullValues(index)
+      } else {
+        v
+      }
+    }
+
+    val binaryArr = evaluated.map(ZorderBytesUtils.toByte).toArray
+    ZorderBytesUtils.interleaveMultiByteArray(binaryArr)
+  }
+
+  override def dataType: DataType = BinaryType
+}
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/ZorderSparkSqlExtensionsParser.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/ZorderSparkSqlExtensionsParser.scala
@@ -1,0 +1,134 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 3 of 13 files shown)


dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/spark/sql/execution/ZorderBytesUtils.scala
```diff
@@ -0,0 +1,143 @@
...
+    intToByte(fi)
+  }
+
+  def doubleToByte(a: Double): Array[Byte] = {
+    val dl = jDouble.doubleToRawLongBits(a)
+    longToByte(dl)
+  }
+
+  def stringToByte(str: String): Array[Byte] = {
+    // truncate or padding str to 8 byte
```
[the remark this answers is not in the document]
qphien: @cfmcgrady Yep, current implementation can not resolve this problem. I think a way to solve this problem is to map string to rank and then generate z-value with rank value. However, this solution needs to sort all string values globally. In consideration of globally sort cost and z-order data skipping effectiveness, a comprise solution is to take sample of rows and get some ordered bounds, then map each row to bound index. For example:
```
Rows:
["http://www.example.com/a",
"http://www.example.com/d",
"http://www.example.com/g",
"http://www.example.com/h",
"http://www.example.com/i",
"http://www.example.com/n",
"http://www.example.com/o",
"http://www.example.com/q",
"http://www.example.com/t"]

Sample Bounds:
["http://www.example.com/a", "http://www.example.com/g"]
["http://www.example.com/h", "http://www.example.com/n"]
["http://www.example.com/o", "http://www.example.com/t"]

Map row to bound index:
{ ("http://www.example.com/a" -> 0),
("http://www.example.com/d" -> 0),
("http://www.example.com/g" -> 0),
("http://www.example.com/h" -> 1),
("http://www.example.com/i" -> 1),
("http://www.example.com/n" -> 1),
("http://www.example.com/o" ->2),
("http://www.example.com/q" -> 2),
("http://www.example.com/t" -> 2)]
```

Bound index is used to generate z-value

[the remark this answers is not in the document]
cfmcgrady: This is a more general solution as I know. Since this is an approximate solution, do we have a benchmark?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/spark/sql/execution/ZorderOrdering.scala
```diff
@@ -0,0 +1,100 @@
...
+              if (row.isNullAt(index)) Double.MaxValue else row.getDouble(index)
+            )
+          case StringType =>
+            ZorderBytesUtils.stringToByte(
+              if (row.isNullAt(index)) "" else row.getString(index)
+            )
+          case _ => null
+//          case DateType
+//          case TimestampType
+//          case DecimalType
```
ulysses-you, line 78: if we don't support them now, we can add exception during analysis

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/spark/sql/execution/ZorderOrdering.scala
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+package org.apache.spark.sql.execution
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{AttributeReference, BaseOrdering, SortOrder}
+import org.apache.spark.sql.types.{BinaryType, BooleanType, ByteType, DoubleType, FloatType, IntegerType, LongType, ShortType, StringType, StructType}
+
+class ZorderOrdering(ordering: Seq[SortOrder],
+                      schema: StructType) extends BaseOrdering {
```
ulysses-you, line 25: Can we make a new `Expression` which return the binary type value and produced by Z-order ?

IMO we can apply Z-order with Spark `Ordering` simply, like `Sort(zorder(c1, c2, c3), ...)` instead. We can still name it as `zorder by c1, c2, c3` in `.g4`. The benifits of this approach is:

* more cleaner since less code copy
* support global z-order, otherwise we need to add a new partitioner

ulysses-you (commented):
Thank you @qphien for the great work !

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/ZorderSparkSqlExtensionsParser.scala
```diff
@@ -0,0 +1,134 @@
...
+      case e: ParseException =>
+        throw e.withCommand(command)
+      case e: AnalysisException =>
+        val position = Origin(e.line, e.startPosition)
+        throw new ParseException(Option(command), e.message, position, position)
+    }
+  }
+
+  private def isZorderCommand(sqlText: String): Boolean = {
+    val normalized = sqlText.toLowerCase(Locale.ROOT).trim().replaceAll("\\s+", " ")
```
cfmcgrady, line 86: Hm, Do you means `replaceAll("\\s+", "")`?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/spark/sql/execution/ZorderBytesUtils.scala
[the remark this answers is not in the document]
qphien: @cfmcgrady Not yet, benchmark will be made when this PR is completed.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/ZorderSparkSqlExtensionsParser.scala
qphien: @cfmcgrady In this case, just `OPTIMIZE` SQL is supported, I think both are okay.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/spark/sql/execution/ZorderOrdering.scala
qphien: OK, It is more concise to use zorder UDF compared with current solution.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/ZorderSparkSqlExtensionsParser.scala
cfmcgrady: I'm saying that you need to replace `"\\s+"` with `""`, not `" "`, we need `""` here, right?

ulysses-you:
@qphien can you fix the conflict ? thanks !

qphien:
> @qphien can you fix the conflict ? thanks !

OK, I have rebased all commit based on master branch.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/OptimizeBeforeWrite.scala
```diff
@@ -0,0 +1,97 @@
...
+        val newChild = child.resolveOperatorsUp {
+          case f@Filter(condition, _) if f.resolved =>
+            if (relation.isDefined) {
+              val tableRelation = relation.get
+              if (!tableRelation.isPartitioned) {
+                throw new ZorderException("Filters are only supported for partitioned table")
+              }
+
+              val partitionKeyIds = AttributeSet(tableRelation.partitionCols)
+              val predicates = splitConjunctivePredicates(condition)
```
ulysses-you, line 44: why not check reference directly instead of split + check ?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/OptimizeBeforeWrite.scala
```diff
@@ -0,0 +1,97 @@
...
+
+case class OptimizeBeforeWrite(session: SparkSession) extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    plan match {
+      case o@OptimizeZorderCommand(child) =>
+        var relation: Option[HiveTableRelation] = None
+        var partitionSpec: Map[String, Option[String]] = Map.empty
+        val hadoopConf = session.sessionState.newHadoopConf()
+        val newChild = child.resolveOperatorsUp {
+          case f@Filter(condition, _) if f.resolved =>
```
ulysses-you, line 36: We can make this code more clearer since we only support two pattern here:
```
case f @ Filter(condition, r: HiveTableRelation) =>
  checkRelation()

case r: HiveTableRelation =>
  checkRelation()

```

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/OptimizeBeforeWrite.scala
```diff
@@ -0,0 +1,97 @@
...
+                !predicate.references.isEmpty &&
+                  predicate.references.subsetOf(partitionKeyIds)
+              }
+
+              if (otherPredicates.nonEmpty) {
+                throw new ZorderException("Only partition column filters are allowed")
+              }
+
+              if (!hadoopConf.get("hive.exec.dynamic.partition", "true").toBoolean) {
+                throw new SparkException(ErrorMsg.DYNAMIC_PARTITION_DISABLED.getMsg)
```
ulysses-you, line 55: do we need the check ? Spark has already checked in `InsertIntoHiveTable`.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/OptimizeZorderCommand.scala
```diff
@@ -0,0 +1,43 @@
...
+            whereExp: Option[Expression],
+            sortArr: Seq[UnresolvedAttribute]): OptimizeZorderCommand = {
+    val table = UnresolvedRelation(tableIdent)
+    val child = whereExp match {
+      case Some(x) => Filter(x, table)
+      case None => table
+    }
+
+    val sortOrder = SortOrder(Zorder(sortArr), Ascending, NullsLast, Seq.empty)
+    val zorderSort = Sort(Seq(sortOrder), false, child)
```
ulysses-you, line 40: I think we support the global zorder now

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/Zorder.scala
```diff
@@ -0,0 +1,104 @@
...
+            Long.MaxValue
+          case other: Any =>
+            throw new ZorderException("Unsupported z-order type: " + other.getClass)
+        }
+      case other: Any =>
+        throw new ZorderException("Unknown z-order column: " + other)
+    }
+  }
+
+  override def nullable: Boolean = true
```
ulysses-you, line 62: does zorder will produce null ?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/Zorder.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  override def eval(input: InternalRow): Any = {
+    var i = 0
+    val evaluated = ArrayBuffer[Any]()
```
ulysses-you, line 66:
```scala
val evaluated = children.zipWithIndex { case (child, index) =>
  val v = child.eval(input)
  if (v == null) {
    defaultNullValues(index)
  } else {
    v
  }
}
```

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/Zorder.scala
```diff
@@ -0,0 +1,104 @@
...
+      val ev = children(i).eval(input)
+      if (ev == null) {
+        evaluated += defaultNullValues(i)
+      } else {
+        evaluated += ev
+      }
+      i += 1
+    }
+
+    val binaryArr = evaluated.toArray.map {
```
ulysses-you, line 77: can we move this code into `ZorderBytesUtils` ?

qphien:
@ulysses-you Thanks for reviewing. All problems mentioned above are resolved.

ulysses-you (approved):

ulysses-you:
thanks @qphien , merging to master for 1.4.0

ulysses-you:
@cfmcgrady @zhouyifan279 @timothy65535 @turboFei feel free to send PR if you have any thought about this

Issue #915, opened 2021-08-10
Renew DelegationTokens for user after user connected to Kyuubi Server


zhouyifan279:
### 1. Describe the feature
Renew DelegationTokens of HDFS, HiveMetaStore for each user.

### 2. Motivation
This is a sub-task under umbrella issue #913.

### 3. Describe the solution
Steps:
1. Run a DelegationTokenRenew task immediatelly when user first logged in.
2. Save obtained DelegationTokens in a <user, credentials> map.
3. Run the following tasks periodically (configured by property kyuubi.security.credentials.renewalInterval).

Introduced properties:

Key | Default | Meaning | Type
---- | -------- | ---------- | -----
kyuubi.credentials.renewal.interval | PT1H | How often Kyuubi renews one user's DelegationTokens | duration
kyuubi.credentials.renewal.retryWait | PT1M| How long to wait before retrying to fetch new credentials after a failure | duration
kyuubi.credentials.hive.enabled | true |  Whether to renew HiveMetaStore DelegationToken | boolean
kyuubi.credentials.hadoopfs.enabled | true | Whether to renew HadoopFS DelegationToken | boolean
kyuubi.credentials.hadoopfs.uris | \<undefined> | Extra Hadoop filesystem URIs for which to request delegation tokens. The filesystem that hosts fs.defaultFS does not need to be listed here. | string

zhouyifan279:
Close as all sub-tasks are completed.

Issue #913, opened 2021-08-10
[Umbrella] Support long running SQL engine without principal and keytab provided


zhouyifan279:
# Describe the proposal

Due to Spark's limitation, a long running SQL engine must be submitted wtih principal and keytab in order to access secured Hadoop cluster. Turn to Kyuubi, this means before using principal and keytab in JDBC url, keytab files need to be deploy on each host where Kyuubi Servers are running.

This umbrellla issue propsed a new way to enable Kyuubi to launch long running SQL engine without principal and keytab provided:
1. Hadoop Cluster manager configures Kyuubi Server's principal to impersonate all Kyuubi users by adding [hadoop proxyuser conf](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/Superusers.html).
2.  Kyuubi Server will obtain DelegationTokens for each use and periodically renew and push DelegationTokens to SQL engine.
 
# Task list

- [x] #915 
  - [x] #1007
  - [x] #1008
  - [x] #1009
- [x] #916
- [x] #1090

link3280:
My 2 cents: Is this feature scoped to YARN setups only? And what if the application needs to access services that do not support Hadoop delegation token (e.g. Kafka)?

link3280:
@ulysses-you Thanks a lot for your input! Now I have a better sense of the PR.

pan3793:
Moved to v1.4.0.

turboFei:
Moved to v1.5.0.

zhouyifan279:
This issue is in v1.4.0 and is finished. Milestone should not be v1.5.0.

turboFei:
thanks @zhouyifan279  have corrected it.

PR #996, opened 2021-08-29
[WIP][KYUUBI #915] Renew DelegationTokens for user after user connected to Kyuubi Server

zhouyifan279:

Due to Spark's limitation, a long running SQL engine must be submitted wtih principal and keytab in order to access secured Hadoop cluster. Turn to Kyuubi, this means before using principal and keytab in JDBC url, keytab files need to be deploy on each host where Kyuubi Servers are running.

The umbrellla issue #913 propsed a new way to enable Kyuubi to launch long running SQL engine without principal and keytab provided and this PR corresponds to the first subtask #915.


zhouyifan279:
This PR has been separated into several sub-tasks.

PR #997, opened 2021-08-30
Fix build error when project path contains `kyuubi-`

turboFei:
The building will fail if the project name contains `kyuubi-`.
For example: for our jenkins job, its name is `kyuubi-binary-build`, and the TMP dir won't be created because all the jar files are filtered because their absolute path contains `kyuubi-`.

### DIFF
--- a/build/release/collect-licenses.sh
+++ b/build/release/collect-licenses.sh
@@ -43,12 +43,14 @@ if [ "${SRC}" = "-h" ]; then
 	exit 0
 fi
 
-for jar_file in $(find -L "${SRC}" -name "*.jar" | grep -v "kyuubi-")
+for jar_file in $(find -L "${SRC}" -name "*.jar")
 do
-	DIR="${TMP}/$(basename -- "${jar_file}" .jar)"
-	mkdir -p "${DIR}"
-	JAR=$(realpath "${jar_file}")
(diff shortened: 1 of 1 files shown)


build/release/collect-licenses.sh
```diff
@@ -43,12 +43,14 @@ if [ "${SRC}" = "-h" ]; then
...
 fi
 
-for jar_file in $(find -L "${SRC}" -name "*.jar" | grep -v "kyuubi-")
+for jar_file in $(find -L "${SRC}" -name "*.jar")
 do
-	DIR="${TMP}/$(basename -- "${jar_file}" .jar)"
-	mkdir -p "${DIR}"
-	JAR=$(realpath "${jar_file}")
-	(cd "${DIR}" && jar xf ${JAR} META-INF/NOTICE META-INF/licenses)
+  if [[ $(basename $jar_file) != kyuubi-* ]]; then
```
pan3793, line 48: Would you please also replace all `tab` in this script with `2 spaces`?
BTW, please keep the same code style with original content, e.g. `${jar_file}`, and quote string by `"`

pan3793 (approved):
LGTM

pan3793:
Thanks, merging to master/v1.3

PR #1002, opened 2021-08-30
Add classification of SQL_Type: this PR is just for DDL

zhang1002:
Add classification of SQL_Type: this PR is just for DDL

Now we use the simple name to judge this operation belong to which classification of SQL_Type.
The matching rule store in #{sql-classification-default.json}. 
The key is the simpleClassName of this analyzed logical plan, the value is the classification of this sql.
It will store in the #{sparksession.conf}, the configuration item is #{spark.sql.classification}.

Why we use the file to store the whole rules?
If we need to change the matching rule, we only need to revise this file.

When the kyuubi started, it will load this json file only once.
When user execute a sql, it will generate a logicalPlan and get its simple name.
By the rules, the function will return the classification that belong to this sql, then it will set this value into #{sparksession.conf}.

For this function, I use SparkSessionExtension to develop it.

By these SQL_Type, we can do something below:
1. Reduce some log print:
	For Auxiliary Statements(the sql type is others), we don't need to print operation log.
2. Optimizing some configuration item.
	For example, in final stage, the conf is different between DML and DQL.
	Through this configuration item, we can use different conf for them.


dev/kyuubi-extension-spark-3-1/src/test/scala/org/apache/spark/sql/KyuubiExtensionSuite.scala
```diff
@@ -418,4 +418,183 @@ class KyuubiExtensionSuite extends QueryTest with SQLTestUtils with AdaptiveSpar
       }
     }
   }
+
+  test("get simple name for DDL") {
+
+    import scala.collection.mutable.Set
```
yaooqinn, line 424: move up?

zhang1002: this test is used for getting simple name for myself...
i use those data set into #{sql-classification-default.json}

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
```diff
@@ -0,0 +1,57 @@
...
+ * we will load the rule from sql-classification-default.json.
+ */
+object KyuubiGetSqlClassification {
+
+  private val enabled: Boolean = SQLConf.get.getConf(SQL_CLASSIFICATION_ENABLED)
+  private var jsonNode: JsonNode = null
+  if (enabled) {
+    try {
+      val defaultSqlClassificationFile =
+        this.getClass.getClassLoader.getResource("sql-classification-default.json").getPath
```
yaooqinn, line 39: Thread.currentThread().getContextClassLoader？

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -67,4 +67,13 @@ object KyuubiSQLConf {
       .version("1.2.0")
       .booleanConf
       .createWithDefault(false)
+
+  val SQL_CLASSIFICATION_ENABLED =
+    buildConf("kyuubi.spark.sql.classification.enabled")
+      .doc("When true, allows Kyuubi engine to judge this SQL's classification " +
+        "and set it into sessionConf." +
+        "Through this configuration item, Spark can optimizing configuration dynamic")
+      .version("1.3.0")
```
yaooqinn, line 76: 1.4

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -67,4 +67,13 @@ object KyuubiSQLConf {
       .version("1.2.0")
       .booleanConf
       .createWithDefault(false)
+
+  val SQL_CLASSIFICATION_ENABLED =
+    buildConf("kyuubi.spark.sql.classification.enabled")
+      .doc("When true, allows Kyuubi engine to judge this SQL's classification " +
+        "and set it into sessionConf." +
```
ulysses-you, line 74: nit: miss a blank

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
zhang1002: done

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
zhang1002: done

yaooqinn:
thanks, merged to master

PR #1004, opened 2021-08-31
Creates the default zk address instead of falling back to frontend bind address

cfmcgrady:

As the Spark driver may launch from another node, when we start the Kyuubi server with an embedded zookeeper and bind the frontend address to `0.0.0.0`, the backend engine can't register correctly.

This pr changes the default behavior of `kyuubi.zookeeper.embedded.client.port.address`, creates the default zookeeper address with local hostname instead of falling back to `kyuubi.frontend.bind.host`.


cfmcgrady:
cc @ulysses-you @yaooqinn @pan3793

yaooqinn:
can you take a look at the test failure？

cfmcgrady:
> can you take a look at the test failure？

The `Check all kyuubi configs` failed. Due to the current documents check mechanism, we can't generate dynamic config at runtime (e.g. create a config entry with default value local hostname).
I will update later.

yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #1013, opened 2021-09-01
Enhance bootstrap shell script

pan3793:

To make the error message more clear when `SPARK_HOME` is invalid.

### DIFF
--- a/bin/load-kyuubi-env.sh
+++ b/bin/load-kyuubi-env.sh
@@ -37,14 +37,14 @@ done
 
 KYUUBI_ENV_SH="${KYUUBI_CONF_DIR}"/kyuubi-env.sh
 if [[ -f ${KYUUBI_ENV_SH} ]]; then
-   set -a
-   if [ $silent -eq 0 ]; then
+  set -a
+  if [ $silent -eq 0 ]; then
     echo "Using kyuubi environment file ${KYUUBI_ENV_SH} to initialize..."
-   fi
-   . "${KYUUBI_ENV_SH}"
-   set +a
+  fi
+  . "${KYUUBI_ENV_SH}"
+  set +a
 else
-   echo "Warn: Not find kyuubi environment file ${KYUUBI_ENV_SH}, using default ones..."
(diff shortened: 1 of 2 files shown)


bin/load-kyuubi-env.sh
```diff
@@ -76,7 +76,7 @@ if [[ -f ${KYUUBI_HOME}/RELEASE ]]; then
   SPARK_BUILTIN="${KYUUBI_HOME}/externals/spark-$SPARK_VERSION_BUILD-bin-hadoop${HADOOP_VERSION_BUILD:0:3}"
 fi
 
-if [[ ! -d ${SPARK_BUILTIN} ]]; then
+if [[ ! -f ${KYUUBI_HOME}/RELEASE && ! -d ${SPARK_BUILTIN} ]]; then
```
yaooqinn, line 79: just `else` here is enough

yaooqinn:
```
-fi
-
-if [[ ! -d ${SPARK_BUILTIN} ]]; then
+else
```

yaooqinn:
```shell
if [[ -z ${SPARK_HOME} ]]; then
  echo "Error: SPARK_HOME IS NOT SET! CANNOT PROCEED." >&2
  exit 1
elif [ ! -d ${SPARK_HOME} ]; then
  echo "Error: SPARK_HOME[${SPARK_HOME}] DOES NOT EXIST! CANNOT PROCEED." >&2
  exit 1
else
  if [[ ! -x "${SPARK_HOME}/bin/spark-submit" ]]; then
    echo "Error: INVALID SPARK DISTRIBUTION! CANNOT PROCEED." >&2
    exit 1
  fi
fi
```

add a new branch for `bin/kyuubi`

bin/load-kyuubi-env.sh
pan3793: Updated.

yaooqinn (approved):

yaooqinn:
thanks ! merged to master and 1.3

Issue #1007, opened 2021-09-01
Implement  delegation token renewal framework


zhouyifan279:
### 1. Describe the feature
This is the first sub-task under umbrella issue #913.
In this task, the basic delegation token renewal framework is implemented in a service named `HadoopCredentialsManager`.

Delegation tokens are managed in user granularity and are periodically renewed inside `HadoopCredentialsManager`.

Note: Sending delegation token from server to  engine will be done in another task.

### 2. Motivation
See #913 

#### Renew and send delegation tokens
HadoopCredentialsManager will be a member of class KyuubiSessionManager.
After opening a KyuubiSession, `HadoopCredentialsManager#sendCredentialsIfNeeded(engineId: String, appUser: String, send: String => Unit)` is called to send delegation tokens  to the SQL engine.   
Method `sendCredentialsIfNeeded` executes the following steps:

1. Get or create a cached CredentialsRef(contains delegation tokens) object by key appUser. If CredentialsRef is newly created, spawn a scheduled task to renew the delegation tokens.
2. Get or create a cached session credentials epoch object by key sessionId.
3. Compare CredentialsRef epoch with session credentials epoch. (Both epochs are set to -1 when created. CredentialsRef epoch is increased when delegation tokens are renewed.)
4. If epochs are equal, return. Else, send delegation tokens to the SQL engine.
5. If sending succeeds, set session credentials epoch to CredentialsRef epoch. Else, record the exception and return.
6. Cached session credentials epochs are removed using removeSessionCredentialsEpoch when session closes.

#### HadoopDelegationTokenProvider
A interface HadoopDelegationTokenProvider is also introduced in this feature.
HadoopDelegationTokenManager loads HadoopDelegationTokenProvider implementations through Java ServiceLoader to obtains delegation tokens from different services.
 HadoopDelegationTokenProvider implementations of Hadoop FileSystem and Hive metastore server will be added in subsequent features.

PR #1015, opened 2021-09-02
[KYUUBI #1007] Implement delegation token renewal framework

zhouyifan279:

This PR finishes sub-task #1007 under umbrella issue #913.


kyuubi-common/src/test/scala/org/apache/kyuubi/util/KyuubiHadoopUtilsSuite.scala
```diff
@@ -35,4 +37,12 @@ class KyuubiHadoopUtilsSuite extends KyuubiFunSuite {
     assert(hadoopConf.get(xyz) === "abc")
     assert(hadoopConf.get(test) === "t")
   }
+
+  test("encode/decode Writable object") {
+    val writable = new Text("Hello, World!")
+    val decoded = new Text()
+
+    KyuubiHadoopUtils.decodeWritable(decoded, KyuubiHadoopUtils.encodeWritable(writable))
```
yaooqinn, line 45: why doesn't `decodeWritable` named decodeString and return `Text`?

scalastyle-config.xml
```diff
@@ -206,7 +206,7 @@ This file is divided into 3 sections:
       <parameter name="groups">java,scala,3rdParty,kyuubi</parameter>
       <parameter name="group.java">javax?\..*</parameter>
       <parameter name="group.scala">scala\..*</parameter>
-      <parameter name="group.3rdParty">(?!org\.apache\.kyuubi\.).*</parameter>
+<!--      <parameter name="group.3rdParty">(?!org\.apache\.kyuubi\.).*</parameter>-->
```
yaooqinn, line 209: why disable this?

kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiHadoopUtils.scala
```diff
@@ -33,4 +35,22 @@ object KyuubiHadoopUtils {
   def getServerPrincipal(principal: String): String = {
     SecurityUtil.getServerPrincipal(principal, "0.0.0.0")
   }
+
+  def encodeWritable(obj: Writable): String = {
```
yaooqinn, line 39: is `Credentials` more clear?

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/EngineCredentialsRef.scala
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.credentials
+
+class EngineCredentialsRef() {
```
yaooqinn, line 20: What are the differences between EngineCredentialsRef and UserCredentialsRef? can you add some comment here

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,225 @@
...
+  private var renewalRetryWait: Long = _
+  private var renewalExecutor: ScheduledExecutorService = _
+  private var hadoopConf: Configuration = _
+
+  override def initialize(conf: KyuubiConf): Unit = {
+    val cacheLoader = new CacheLoader[String, EngineCredentialsRef] {
+      override def load(key: String): EngineCredentialsRef = new EngineCredentialsRef
+    }
+    engineRefCache = CacheBuilder.newBuilder()
+      .concurrencyLevel(conf.get(KyuubiConf.SERVER_EXEC_POOL_SIZE))
```
yaooqinn, line 59: why we need SERVER_EXEC_POOL_SIZE  here?

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,225 @@
...
+      .expireAfterAccess(conf.get(KyuubiConf.ENGINE_IDLE_TIMEOUT), TimeUnit.MILLISECONDS)
```
yaooqinn, line 60: ENGINE_IDLE_TIMEOUT is per engine, is it right to use here

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
```diff
@@ -0,0 +1,205 @@
...
+private class ExceptionThrowingDelegationTokenProvider extends HadoopDelegationTokenProvider {
+  ExceptionThrowingDelegationTokenProvider.constructed = true
+  throw new IllegalArgumentException
+
+  override def serviceName: String = "throw"
+
+  override def delegationTokensRequired(
+      hadoopConf: Configuration,
+      kyuubiConf: KyuubiConf): Boolean =
+    throw new IllegalArgumentException
```
yaooqinn, line 146: how about we add some message here?

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
```diff
@@ -0,0 +1,205 @@
...
+
+  override def obtainDelegationTokens(
+      hadoopConf: Configuration,
+      kyuubiConf: KyuubiConf,
+      owner: String,
+      creds: Credentials): Unit = throw new IllegalArgumentException
```
yaooqinn, line 152: ditto

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,225 @@
...
+    with Logging {
+
+  def this() = this(classOf[HadoopCredentialsManager].getSimpleName)
+
+  private val userRefMap = new ConcurrentHashMap[String, UserCredentialsRef]()
+
+  private var engineRefCache: LoadingCache[String, EngineCredentialsRef] = _
+
+  private var providers: Map[String, HadoopDelegationTokenProvider] = _
+  private var requiredProviders: Map[String, HadoopDelegationTokenProvider] = _
```
yaooqinn, line 47: why do we need two maps

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,225 @@
...
+  def sendCredentialsIfNeeded(
+      engineId: String,
+      appUser: String,
+      send: String => Unit): Boolean = {
+    require(renewalExecutor != null, "renewalExecutor should be initialized")
+
+    val userRef = getUserRef(appUser)
+    val engineRef = getEngineRef(engineId)
+
+    var sent = false
```
yaooqinn, line 118: where is it used?

kyuubi-common/src/test/scala/org/apache/kyuubi/util/KyuubiHadoopUtilsSuite.scala
zhouyifan279: The methods naming and implements are from `org.apache.hadoop.security.token.Token#encodeWritable` and `org.apache.hadoop.security.token.Token#decodeWritable` and  should be able to convert any `Writable` object. 
Here, `encodeWritable` is used to convert Credentials to String at Kyuubi Server side and `decodeWritable` is used to convert String to Credentials at Kyuuib SQL engine side.
Change `decodeWritable` to `decodeString` and return `Text` maybe is not proper  for its usage.

scalastyle-config.xml
zhouyifan279: Oops~ IntelliJ IDEA can not sort imports properly when this line is enabled. So I commented it locally.
This should not be committed. I will revert it.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/EngineCredentialsRef.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  @volatile
+  private var epoch = -1L
```
yaooqinn, line 23: can't we just add another epoch to UserCredentialsRef?

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,225 @@
...
+          requiredProviders.values
+            .foreach(_.obtainDelegationTokens(hadoopConf, conf, userRef.getAppUser, creds))
+          userRef.updateCredentials(creds)
+          scheduleRenewal(userRef, renewalInterval)
+        } catch {
+          case _: InterruptedException =>
+          // Server is shutting down
+          case e: Exception =>
+            val duration = Duration.ofMillis(renewalRetryWait)
+            warn(s"Failed to update tokens, will try again in ${duration}", e)
```
yaooqinn, line 181: duration's string representation may hard to read

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/EngineCredentialsRef.scala
yaooqinn: I see

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,225 @@
...
+   * `appUser`'s credentials are created if not exist.
+   * [[HadoopCredentialsManager]] takes care created credentials' renewal.
+   * @param engineId SQL engine identifier
+   * @param appUser  User identity that the SQL engine uses.
+   * @param send     Function to send encoded credentials to SQL engine
+   * @return `true` if credentials are sent successfully. Else, false.
+   * @throws KyuubiException if sending credentials fails.
+   */
+  def sendCredentialsIfNeeded(
+      engineId: String,
```
yaooqinn, line 110: maybe sessionId is more easy to implement

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Thought to set  a proper concurrency level for `engineRefCache`. 
But since loading and getting from `engineRefCache` are almost an instant action and won't become a bottleneck, maybe it is better to remove it.

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
zhouyifan279: Test class copied from Spark. Remove it since it is confusing and has no influence in test case.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Moved method `loadProviders` to companion object. Using `loadProviders` returned result to test provider loading mechanism.

Field `providers` is removed.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Used by test code to verify sending result only. 

Removed since sending result already verified by comparing `EngineCredentialsRef#getEpoch` with `UserCredentialsRef#getEpoch`.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Changed to represent milliseconds directly.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Good idea!

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Adjust engine credentials epoch tracking to session scope. Expire time is removed.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,245 @@
...
+ * @note [[RemoteCredentialsRef]]s are created in session scope and should be removed using
+ *       [[removeRemoteCredentialsRef]] when session closes.
+ */
+class HadoopCredentialsManager private (name: String) extends AbstractService(name)
+    with Logging {
+
+  def this() = this(classOf[HadoopCredentialsManager].getSimpleName)
+
+  private val localRefMap = new ConcurrentHashMap[String, LocalCredentialsRef]()
+  private val remoteRefMap = new ConcurrentHashMap[String, RemoteCredentialsRef]()
```
yaooqinn, line 73: Seems it can be a string->long map?

yaooqinn, line 73: string -> long is enough

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * @note [[RemoteCredentialsRef]]s are created in session scope and should be removed using
+ *       [[removeRemoteCredentialsRef]] when session closes.
+ */
+class HadoopCredentialsManager private (name: String) extends AbstractService(name)
+    with Logging {
+
+  def this() = this(classOf[HadoopCredentialsManager].getSimpleName)
+
+  private val localRefMap = new ConcurrentHashMap[String, LocalCredentialsRef]()
```
yaooqinn, line 72: make the name here more clear

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,245 @@
...
+          val currentEpoch = userRef.getEpoch
+          val currentCreds = userRef.getEncodedCredentials
+          info(s"Send new credentials with epoch $currentEpoch to SQL engine through session " +
+            s"$sessionId")
+          Try(send(currentCreds)) match {
+            case Success(_) =>
+              info(s"Update session scope epoch from ${engineRef.getEpoch} to $currentEpoch")
+              engineRef.setEpoch(currentEpoch)
+            case Failure(exception) =>
+              throw new KyuubiException(
```
yaooqinn, line 146: How do we handle this exception later?

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,245 @@
...
+      sessionId: String,
+      appUser: String,
+      send: String => Unit): Unit = {
+    require(renewalExecutor != null, "renewalExecutor should be initialized")
+
+    val userRef = getOrCreateUserRef(appUser)
+    val engineRef = getOrCreateEngineRef(sessionId)
+    if (userRef.getEpoch != engineRef.getEpoch) {
+      engineRef synchronized {
+        if (userRef.getEpoch != engineRef.getEpoch) {
```
yaooqinn, line 136: can you explain a bit? the code here is hard to read

zhouyifan279: Verified that session scope credentials epoch won't be modified concurrently and removed synchronization code.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Since method caller can do nothing but record the exception, changed to record exception in current method.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Renamed to `userCredentialsRefMap`

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: Done

zhouyifan279:
@yaooqinn modified according to the latest advice.  Ready to be reviewed again. CC @ulysses-you @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,242 @@
...
+  def sendCredentialsIfNeeded(
+      sessionId: String,
+      appUser: String,
+      send: String => Unit): Unit = {
+    require(renewalExecutor != null, "renewalExecutor should be initialized")
+
+    val userRef = getOrCreateUserCredentialsRef(appUser)
+    val sessionEpoch = getSessionCredentialsEpoch(sessionId)
+
+    if (userRef.getEpoch != sessionEpoch) {
```
ulysses-you, line 135: so there are two cases the user epoch can be lager than session epoch
* the new created user cred
* we failed to send created before

How about use `if (userRef.getEpoch > sessionEpoch)` ?

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopDelegationProvider.scala
```diff
@@ -0,0 +1,53 @@
...
+   * Name of the service to provide delegation tokens. This name should be unique. Kyuubi will
+   * internally use this name to differentiate delegation token providers.
+   */
+  def serviceName: String
+
+  /**
+   * Returns true if delegation tokens are required for this service. By default, it is based on
+   * whether Hadoop security is enabled.
+   */
+  def delegationTokensRequired(hadoopConf: Configuration, kyuubiConf: KyuubiConf): Boolean
```
ulysses-you, line 38: in general, one configuration is enough. Is there a chance that we can remove one ?

ulysses-you (commented):
thank you @zhouyifan279 for the work !

kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiHadoopUtils.scala
```diff
@@ -33,4 +35,24 @@ object KyuubiHadoopUtils {
   def getServerPrincipal(principal: String): String = {
     SecurityUtil.getServerPrincipal(principal, "0.0.0.0")
   }
+
+  def encodeCredentials(creds: Credentials): String = {
+    val buf = new DataOutputBuffer
+    creds.write(buf)
+    val encoder = new Base64(0, null, false)
+    val raw = new Array[Byte](buf.getLength)
+    System.arraycopy(buf.getData, 0, raw, 0, buf.getLength)
```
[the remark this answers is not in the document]
zhouyifan279: DataOutputBuffer's underlying byte array grows 1 times bigger when it is full. `DataOutputBuffer#getData` returns the underlying byte array and may contains extra bytes.

kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiHadoopUtils.scala
```diff
@@ -33,4 +35,24 @@ object KyuubiHadoopUtils {
   def getServerPrincipal(principal: String): String = {
     SecurityUtil.getServerPrincipal(principal, "0.0.0.0")
   }
+
+  def encodeCredentials(creds: Credentials): String = {
```
yaooqinn, line 39:
```scala
  def serialize(creds: Credentials): Array[Byte] = {
    val byteStream = new ByteArrayOutputStream
    val dataStream = new DataOutputStream(byteStream)
    creds.writeTokenStorageToStream(dataStream)
    byteStream.toByteArray
  }

  def deserialize(tokenBytes: Array[Byte]): Credentials = {
    val tokensBuf = new ByteArrayInputStream(tokenBytes)

    val creds = new Credentials()
    creds.readTokenStorageStream(new DataInputStream(tokensBuf))
    creds
  }

```

the implementation in Spark side looks more elegant to me

yaooqinn: we can store bytes in memory directly?

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopDelegationProvider.scala
zhouyifan279: `kyuubiConf` is needed because some providers (if we need to implement in the future), eg kafka, can read configs(server address, or client config file path ) from KyuubiConf;
`hadoopConf`can be created using `KyuubiHadoopUtils#newHadoopConf`in each providers. But maybe it is better to keep this logic outside providers to make it a little simpler to write the provider.
Besides, it will be easier if we want to change `hadoopConf`'s loading logic.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/CredentialsRef.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def updateCredentials(creds: Credentials): Unit = {
+    encodedCredentials = KyuubiHadoopUtils.encodeCredentials(creds)
+    epoch += 1
+  }
+
+}
+
+object CredentialsRef {
+  val UNSET_EPOCH: Long = -1L
```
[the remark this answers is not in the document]
zhouyifan279: Prefer to keep using `-1` to represent the state of no credentials.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,242 @@
...
+  private var renewalExecutor: ScheduledExecutorService = _
+  private var hadoopConf: Configuration = _
+
+  override def initialize(conf: KyuubiConf): Unit = {
+    hadoopConf = KyuubiHadoopUtils.newHadoopConf(conf)
+    providers = HadoopCredentialsManager.loadProviders(conf)
+      .filter { case (_, provider) =>
+        val required = provider.delegationTokensRequired(hadoopConf, conf)
+        if (!required) {
+          info(s"Service ${provider.serviceName} does not require a token." +
```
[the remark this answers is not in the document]
zhouyifan279: done

kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiHadoopUtils.scala
zhouyifan279: Use Spark's implementation since it is simpler. 
**Note:** `ByteArrayOutputStream` also involves  an array copy action.

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -0,0 +1,252 @@
...
+   *
+   * @param sessionId Specify the session which is talking with SQL engine
+   * @param appUser  User identity that the SQL engine uses.
+   * @param send     Function to send encoded credentials to SQL engine
+   */
+  def sendCredentialsIfNeeded(
+      sessionId: String,
+      appUser: String,
+      send: String => Unit): Unit = {
+    val userRef = getOrCreateUserCredentialsRef(appUser)
```
yaooqinn, line 138: what can skip this if providers is empty?

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
zhouyifan279: done

yaooqinn (approved):

Issue #1022, opened 2021-09-03
Add basic EngineStatusStore for events


timothy65535:
### 1. Describe the feature

For more detail, please go to https://github.com/apache/incubator-kyuubi/issues/981

### 2. Additional context

`EventManager` helps to push events to listener bus
`EngineStatusStore` is a memory store that tracking the number of statements and sessions

PR #1023, opened 2021-09-03
[KYUUBI #1022] Add basic EngineStatusStore for events

timothy65535:

For more detail, please go to https://github.com/apache/incubator-kyuubi/issues/981

`EngineStatusStore` helps to push events to listener bus

`EngineStatusStore` is a memory store that tracking the number of statements and sessions, it provides:
- stores all elements, and sorted by startTimestamp.
- cleanup the last elements when reach a certain threshold.
 


timothy65535:
cc @yaooqinn, realized a single memory store called `EngineStatusStore`.

It's ok to rework with `ElementTrackingStore`. https://github.com/apache/incubator-kyuubi/issues/981#issuecomment-912440645

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EventManager.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  def onStatementFinish(id: String): Unit = {
+    postListenerEvent(KyuubiEngineOperationFinishEvent(id, System.currentTimeMillis()))
+  }
+
+  def onOperationClosed(id: String): Unit = {
+    postListenerEvent(KyuubiEngineOperationClosedEvent(id, System.currentTimeMillis()))
+  }
+}
+
```
cfmcgrady, line 74: Does `EventLoggingService` need to store these events? @yaooqinn @ulysses-you

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EventManager.scala
```diff
@@ -0,0 +1,114 @@
...
+case class KyuubiEngineSessionCreatedEvent(
```
cfmcgrady, line 75: It's duplicate with `org.apache.kyuubi.engine.spark.events.SparkStatementEvent`?

yaooqinn: yes，we can use the current events directly

timothy65535:
Thanks @cfmcgrady @yaooqinn, make the current patch to draft, starting rework this patch by using `ElementTrackingStore`.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EventManager.scala
timothy65535: Maybe we make `EventLoggingService` as a sparklistener is better.

image

timothy65535: cc @yaooqinn @ulysses-you @cfmcgrady, looking forward to your views.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EventManager.scala
timothy65535: > we can use the current events directly

`ExecutionInfo` has more fields than `SparkStatementEvent`, can not use current events directly.

We can make `EventLoggingService` as a sparklistener, then reuse `SessionInfo` and `ExecutionInfo`.

yaooqinn:
How about we reuse the current `event logging service` and only introduce the `ElementTrackingStore` or something else as an in-memory implementation?

timothy65535:
> How about we reuse the current event logging service 

It's not a right way to reuse event logging service.

- We should push events to `SparkListenerBus`, if not, the spark history server will not receive these custom events.
- It's better to let different listeners do their own independent things

> and only introduce the ElementTrackingStore or something else as an in-memory implementation?

In fact, using `ElementTrackingStore` is the least expensive, if we create something as an in-memory implementation will spend more time. I had tried these two ways.

yaooqinn:
> We should push events to SparkListenerBus, if not, the spark history server will not receive these custom events.

can we check SparkHistoryEventLogger?

cfmcgrady:
> How about we reuse the current `event logging service` and only introduce the `ElementTrackingStore` or something else as an in-memory implementation?

+1

timothy65535:
hi @yaooqinn, I can not understand why kyuubi need `SparkHistoryEventLogger`? it seems duplicated to `spark.eventLog.enabled=true`.

timothy65535:
Looked into the logger service, the loggers are executed sequentially, and the loggers are called synchronously. 

image

<br />

And checked the `LiveListenerBus`, it is an asynchronous listener bus for Spark events. [SPARK-20863]( https://github.com/apache/spark/pull/18083)

yaooqinn:
> hi @yaooqinn, I can not understand why kyuubi need SparkHistoryEventLogger?

SparkHistoryEventLogger is used to log all Kyuubi's events to the same file of spark's event log. We also support other destinations, like Kyuubi's own JSON log store, JDBC(planned), etc..

> it seems duplicated to spark.eventLog.enabled=true.

It's not. it follows spark.eventLog.enabled when logging. We just don't push the events to the listener bus but call the event log listener directly.

> And, the current design of logger service is not clear to me. As we know a service means that it should listen to some port and can start / stop.

Hmm... do not go too far with the spark's listeners and the listener bus if you are not understanding what you are doing here.

IIUC, you want everything to work with spark's listeners and the listener bus, this is not what we want.  We are a Spark application **BUT** not only a spark application that is limited to spark's functionalities.

What we want is only a size-limit buffer (using `ElementTrackingStore` is just an option) that holds all the current Kyuubi events in memory. Then render one or more Spark UI pages/tables based on this buffer.

yaooqinn:
> Looked into the logger service, the loggers are executed sequentially, and the loggers are called synchronously.
> 
> image
> 
> 
> And checked the `LiveListenerBus`, it is an asynchronous listener bus for Spark events. [SPARK-20863](https://github.com/apache/spark/pull/18083)

Yes, this is a spot that we can improve too

timothy65535:
> Yes, this is a spot that we can improve too

hi @yaooqinn, there seems to be no good way to solve the synchronization problem. For `EventLoggerType.SPARK` and `EventLoggerType.JSON`, we need to think out a asynchronous writer for local path or hdfs path.

It seems we should improve it first, because if we listener these events in this pr, it may affects the performance.

yaooqinn:
Can we start with session Events only in another PR, which can minimize the review burdern

yaooqinn:
We can combine the history logging and the live data logging  with the listener bus but we can still use the event logging service

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkContextHelper.scala
```diff
@@ -30,6 +32,11 @@ object SparkContextHelper {
   def createSparkHistoryLogger(sc: SparkContext): EventLogger[KyuubiSparkEvent] = {
     new SparkHistoryEventLogger(sc)
   }
+
+  def postListenerEvent(event: SparkListenerEvent): Unit = {
+    KyuubiSparkUtil.globalSparkContext.listenerBus.post(event)
+  }
+
```
yaooqinn, line 39: we can move this to L48, `sc.listenerBus.post(event)`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetFunctions.scala
```diff
@@ -54,6 +55,17 @@ class GetFunctions(
         " function within its schema")
   }
 
+  override protected def beforeRun(): Unit = {
```
yaooqinn, line 58: can we keep the meta operations unchanged for now

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkContextHelper.scala
yaooqinn: Then EventManager seems t not need anymore

yaooqinn:
> Hmm... do not go too far with the spark's listeners and the listener bus if you are not understanding what you are doing here.

I take my early judgment back, and the listener bus looks the right way to go. https://github.com/apache/incubator-kyuubi/pull/1023#discussion_r702801909

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/ui/EngineAppStatusStore.scala
```diff
@@ -0,0 +1,141 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import org.apache.spark.status.KVUtils.KVIndexParam
+import org.apache.spark.util.kvstore.{KVIndex, KVStore}
+
+/**
+ * Provides a view of a KVStore with methods that make it easy to query SQL-specific state. There's
+ * no state kept in this class, so it's ok to have multiple instances of it in an application.
+ */
+class EngineAppStatusStore(store: KVStore) {
```
yaooqinn, line 31: Where do we use this?

timothy65535:
> Can we start with session Events only in another PR, which can minimize the review burdern

Sure, will do.

yaooqinn:
> there seems to be no good way to solve the synchronization problem. 
For `EventLoggerType.SPARK`

This looks simple, Please send a separate PR to solve this issue

```git
-    sc.eventLogger.foreach(_.onOtherEvent(kyuubiEvent))
+    sc.listenerBus.post(kyuubiEvent)
```

> and `EventLoggerType.JSON`, we need to think out a asynchronous writer for local path or hdfs 

we can implement this latter too

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/ui/EngineAppStatusStore.scala
timothy65535: The `EngineAppStatusStore` is only used in tests in this patch, we will use it in two places in the future: 
- UI pages
- history server

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkContextHelper.scala
timothy65535: Nice, got.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/GetFunctions.scala
timothy65535: ok

timothy65535:
> We can combine the history logging and the live data logging with the listener bus but we can still use the event logging service

Had tried serveal times, but failed.

For `EventLoggerType.JSON` logger, it requires all fields are set.

But for `EventLoggerType.SPARK` logger, it don't needs all fields.

### spark origin event log
```
{"Event":"SparkListenerJobEnd","Job ID":2,"Completion Time":1630997161352,"Job Result":{"Result":"JobSucceeded"}}
{"Event":"org.apache.spark.sql.execution.ui.SparkListenerSQLExecutionEnd","executionId":3,"time":1630997161355}
{"Event":"org.apache.spark.sql.hive.thriftserver.ui.SparkListenerThriftServerOperationFinish","id":"8f2c5e68-c9fe-4777-bffe-75d171fa18cb","finishTime":1630997161355}
{"Event":"org.apache.spark.sql.hive.thriftserver.ui.SparkListenerThriftServerOperationClosed","id":"8f2c5e68-c9fe-4777-bffe-75d171fa18cb","closeTime":1630997161374}
{"Event":"org.apache.spark.sql.hive.thriftserver.ui.SparkListenerThriftServerSessionClosed","sessionId":"7e3f25d8-5688-4170-8532-5ddee4112991","finishTime":1630997192176}
{"Event":"SparkListenerApplicationEnd","Timestamp":1630997194335}
```

timothy65535:
hi @yaooqinn 

Had read the whole ui design of sparkthriftserver 
https://github.com/apache/spark/tree/master/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/ui

And already tried toiImplement a memory status store
https://github.com/timothy65535/kyuubi/blob/ky-981-t1/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/ui/EngineStatusStore.scala

At last, in terms of design compatibility and simplicity, I think it's better to reuse the thriftserver's ui design. We need to consider not only the pages on engine server, but also the pages on spark history server.

timothy65535:
cc @yaooqinn the patch updated with custom store.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -124,6 +126,10 @@ class ExecuteStatement(
             rejected)
           setOperationException(ke)
           throw ke
+        case NonFatal(e) =>
```
yaooqinn, line 129: Is this a bug you try to fix here？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EventLoggingService.scala
```diff
@@ -68,5 +68,13 @@ object EventLoggingService {
   def onEvent(event: KyuubiSparkEvent): Unit = {
     _service.foreach(_.onEvent(event))
   }
+
+  def onSessionCreated(event: SessionEvent): Unit = {
```
yaooqinn, line 72: we don‘t need this

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EventLoggingService.scala
```diff
@@ -68,5 +68,13 @@ object EventLoggingService {
...
+    onEvent(event)
+  }
+
+  def onSessionClosed(event: SessionEvent): Unit = {
```
yaooqinn, line 76: ditto

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/EngineStatusListener.scala
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.spark.kyuubi.ui._
+import org.apache.spark.scheduler._
+
+import org.apache.kyuubi.Logging
+import org.apache.kyuubi.engine.spark.events.SessionEvent
+
+/**
+ * A [[SparkListener]] for engine status events handling.
+ */
+class EngineStatusListener(store: EngineStatusStore) extends SparkListener with Logging {
```
yaooqinn, line 29: we can move this to SparkSQLEngineListener

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
timothy65535: refered to

https://github.com/apache/spark/blob/cb30683b65714bd1c1ae15d31d1411aeb66cef12/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/SparkExecuteStatementOperation.scala#L260

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/EngineStatusListener.scala
timothy65535: > move this to SparkSQLEngineListener

maybe SparkSQLEngineListener will be heavy.

timothy65535:
@yaooqinn already to go, if use customized store, we will convert `ElementTrackingStore` when support sparkhistoryserver.

yaooqinn:
can you update the PR description and maybe also some comments in the code changes to help us review?

timothy65535:
> can you update the PR description and maybe also some comments in the code changes to help us review?

Updated. The store based on `TreeBasedStore`, it supports row keys and column keys are ordered by natural ordering by default.

timothy65535:
cc @cfmcgrady when you are free, thanks.

yaooqinn:
cc @zhang1002

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -868,6 +868,13 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val ENGINE_UI_SESSION_LIMIT: ConfigEntry[Int] =
+    buildConf("engine.ui.retainedSessions")
+      .doc("The number of SQL client sessions kept in the Kyuubi Query Engine web UI.")
+      .version("1.4.0")
+      .intConf
```
cfmcgrady, line 875: this conf must be a positive integer, we can add `checkValue`.

timothy65535: make sense.

timothy65535:
cc @pan3793 @ulysses-you, help to review when free, thanks.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * A memory store that tracking the number of statements and sessions, it provides:
+ *
+ * - stores all events, and sorted by startTimestamp.
+ * - cleanup the last events when reach a certain threshold.
+ */
+class EngineEventsStore(conf: KyuubiConf) {
+
+  private type Long = java.lang.Long
```
yaooqinn, line 35: import java.lang.{Long => JLong}

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
```diff
@@ -0,0 +1,74 @@
...
+    sessions.put(sessionEvent.startTime, sessionEvent.sessionId, sessionEvent)
+    checkSessionCapacity()
+  }
+
+  /**
+   * cleanup the last session events if reach the threshold
+   */
+  private def checkSessionCapacity(): Unit = {
+    val countToDelete = sessions.size - retainedSessions
+    if (countToDelete > 0) {
```
yaooqinn, line 68: is there always 1 or 0 to remove?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  def getSession(sessionId: String): Option[SessionEvent] = {
+    sessions.values().asScala.find(s => s.sessionId == sessionId)
+  }
+
+  /**
+   * save session events and check the capacity threshold
+   */
+  def saveSession(sessionEvent: SessionEvent): Unit = synchronized {
+    sessions.put(sessionEvent.startTime, sessionEvent.sessionId, sessionEvent)
```
yaooqinn, line 59: shall we use endTime?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.config.KyuubiConf.ENGINE_UI_SESSION_LIMIT
+
+/**
+ * A memory store that tracking the number of statements and sessions, it provides:
+ *
+ * - stores all events, and sorted by startTimestamp.
+ * - cleanup the last events when reach a certain threshold.
+ */
+class EngineEventsStore(conf: KyuubiConf) {
```
yaooqinn, line 33: where do we use this store for UI?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
timothy65535: use startTime is more resonable.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
timothy65535: seems right

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
timothy65535: ok

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
timothy65535: this store will be used in page.

yaooqinn: would you please add a TODO with a issue ticket then

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
yaooqinn: can you tell me the reason then？ it seem to me that we very likely to keep late started but closed sessions，in the meanwhile，drop those still alive？

timothy65535: the origin design about the retain may like a fixed size queue. if submit 300 queries at the same time, and the retain size is 200, no matter whether starttime or endtime is used, there will also be 100 active queries are excluded.

timothy65535: if sort the events order by starttime, seems more easier to understand.

yaooqinn: can we always keep the live sessions?

timothy65535: ok, will updated.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
timothy65535: updated.

timothy65535:
cc @yaooqinn ready to review, looking for your new advice.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
```diff
@@ -0,0 +1,89 @@
...
+   */
+  def saveSession(sessionEvent: SessionEvent): Unit = {
+    sessions.put(sessionEvent.sessionId, sessionEvent)
+    checkSessionCapacity()
+  }
+
+  /**
+   * cleanup the session events if reach the threshold
+   */
+  private def checkSessionCapacity(): Unit = {
```
yaooqinn, line 71: how efficient is this？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
```diff
@@ -0,0 +1,89 @@
...
+    var countToDelete = sessions.size - retainedSessions
+    
+    val reverseSeq = sessions.values().asScala.toSeq.sortBy(_.startTime).reverse
```
yaooqinn, line 74: we sort the value set for event single event？...

timothy65535: if we use treemap, and make startTime or endTime as key, it will remove events if key repeat.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
timothy65535: Can't find a suitable opensource library to supports following characteristic:
- Concurrency: concurrent read、update、remove
- Update
- Order
- Complexity

If we implement a new Map which extends AbstractMap, it seems will be complex.

If we keep 200 events in memory, efficiency may not be affected.

timothy65535:
cc @ulysses-you @pan3793

timothy65535:
had already tried serval implements:
V1: custom store based on treemap
V2: use spark ElementTrackingStore
V3: custom store based on TreebasedTable, ordered by startTime
V4: custom store based on ConcurrentHashMap, ordered by finishTime

looking for good advice, thanks

timothy65535:
BTW, at this stage, seems we don't need to spend a lot of time on details, especially on how to implement the store, it already cost two week.

timothy65535:
hi @ulysses-you @pan3793 @cfmcgrady @yaooqinn, any more thought?

timothy65535:
cc @ulysses-you @pan3793 @cfmcgrady @yaooqinn, any more thought? thanks

yaooqinn:
I am +0 on this, but since no one has an opposite option on this implementation, I get this merged.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
zhang1002: here you should use asyn, by this you can sort the value only once when countToDelete reached the set value.
`if (retainedSessions/sessions.size >= threshold) {
    new thread {
        sort()
        delete()
}
}
`
also you can see Guava Cache, you can set the expire strategy by youself but it will waste some Mem...

timothy65535:
OK, new idea to the implement of the store is welcome.

Issue #703, opened 2021-06-18
Hudi support


zxding:
Spark thrift server support hudi sql feature, such as insert into, update,delete,merge into grammer;will kyuubi plan to support feature like this?

yaooqinn:
yes. hudi shall work fine with Kyuubi, but we've tested that. are you interested in this?

zxding:
Yes，will have product application using kyuubi ,and I try to add some new grammer feature using kyuubi 0.7 version to support hudi spark sql. If you have already work on this, would you please tell the deadline , so that I can using your code and change my plan. 3q.

yaooqinn:
0.7 is pretty old, have you tried the laster 1.2.0-rc5?

zxding:
Ok，1.2.0 have already support hudi spark sql update and merge into grammer??

yaooqinn:
you are talking about a feature that should be supported by Spark and Hudi's side. If they both do not have any plan to make it work, we will consider adding some extensions

zxding:
If possible ,I wanna to talk about this issue with wechat.I work at hik.

yaooqinn:
yaooqinn

lxk000000:
We have achieved kyuubi support hudi with hudi datasource mode on the spark version 2.4.5. 
Follow-up consideration in version 3.0. Hope to merge these codes

yaooqinn:
that sounds nice. But can you detail what we are going to have in kyuubi? Do we have a design doc for it?

PR #1028, opened 2021-09-04
[KYUUBI #703][FOLLOWUP][DEPS][TEST] Reduce Hudi dependencies

pan3793:

HBase deps is optional for creating hudi table and query, but required for writing data to hudi table, as currently we only test create table and query for data lake formats, it's reasonable to remove the hbase deps to reduce the deps pulled from hudi.

This PR also change all data lake test suites in spark engine from `hive` catalog to `in-memory` catalog.

### DIFF
--- a/kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
+++ b/kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
@@ -42,6 +42,6 @@ trait HudiSuiteMixin extends DataLakeSuiteMixin {
     "spark.sql.defaultCatalog" -> catalog,
     "spark.sql.extensions" -> "org.apache.spark.sql.hudi.HoodieSparkSessionExtension",
     "spark.serializer" -> "org.apache.spark.serializer.KryoSerializer",
-    "hoodie.bootstrap.index.class" -> "org.apache.hudi.common.bootstrap.index.HFileBootstrapIndex",
+    "hoodie.bootstrap.index.class" -> "org.apache.hudi.common.bootstrap.index.NoOpBootstrapIndex",
     "spark.jars" -> extraJars)
 }
--- a/kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
+++ b/kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
@@ -42,6 +42,6 @@ trait HudiSuiteMixin extends DataLakeSuiteMixin {
     "spark.sql.defaultCatalog" -> catalog,
     "spark.sql.extensions" -> "org.apache.spark.sql.hudi.HoodieSparkSessionExtension",
     "spark.serializer" -> "org.apache.spark.serializer.KryoSerializer",
-    "hoodie.bootstrap.index.class" -> "org.apache.hudi.common.bootstrap.index.NoOpBootstrapIndex",
+    "hoodie.bootstrap.index.class" -> "org.apache.hudi.common.bootstrap.index.HFileBootstrapIndex",
     "spark.jars" -> extraJars)
 }
--- a/kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
+++ b/kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
@@ -38,6 +38,7 @@ trait HudiSuiteMixin extends DataLakeSuiteMixin {
   }
 
   override protected def extraConfigs = Map(
+    "spark.sql.catalogImplementation" -> "in-memory",
     "spark.sql.defaultCatalog" -> catalog,
     "spark.sql.extensions" -> "org.apache.spark.sql.hudi.HoodieSparkSessionExtension",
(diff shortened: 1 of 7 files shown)


pan3793:
@simon824 seems that HUDI test suites passed without any HBase depenedencies

kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
```diff
@@ -42,5 +42,6 @@ trait HudiSuiteMixin extends DataLakeSuiteMixin {
     "spark.sql.defaultCatalog" -> catalog,
     "spark.sql.extensions" -> "org.apache.spark.sql.hudi.HoodieSparkSessionExtension",
     "spark.serializer" -> "org.apache.spark.serializer.KryoSerializer",
+    "hoodie.bootstrap.index.class" -> "org.apache.hudi.common.bootstrap.index.HFileBootstrapIndex",
```
pan3793, line 45: Actually, `HFileBootstrapIndex` is the default value, it's not matter with/without this conf I think.

simon824:
> @simon824 seems that HUDI test suites passed without any HBase depenedencies

Create operation is working fine,but insert /update operation will cause exception, does ut need add  insert /update operation ?

kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
simon824: Yes, HFileBootstrapIndex is the default value, if you want reduce hbase dep, this config should be 'org.apache.hudi.common.bootstrap.index.NoOpBootstrapIndex'.

pan3793:
The current ut for data lake formats only contain create and query, I think it's OK to remove the hbase deps now, and maybe hudi will have improved the deps management when we add the write test cases in the future.

simon824:
> The current ut for data lake formats only contain create and query, I think it's OK to remove the hbase deps now, and maybe hudi will have improved the deps management when we add the write test cases in the future.

OK

kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
pan3793: Yeah, just change it back to default for test. BTW, is it the right approach to overwrite hudi conf?

simon824: Overwriting the config when creating the table

https://hudi.apache.org/docs/quick-start-guide
image

kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
```diff
@@ -38,7 +38,6 @@ trait HudiSuiteMixin extends DataLakeSuiteMixin {
   }
 
   override protected def extraConfigs = Map(
-    "spark.sql.catalogImplementation" -> "in-memory",
```
simon824, line 41: this config is necessary

pan3793: Yeah, I see, just move it to `WithSparkSQLEngine`. 
BTW, why hudi doesn't work with `hive` catalog? `in-memory` should only used for test, I think most users would use hive for production.

kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
pan3793: Thanks, we need refactor `DataLakeSuiteMixin` and expose create table options to subclass in future.

pan3793:
cc @yanghua @yaooqinn

kyuubi-common/src/test/scala/org/apache/kyuubi/HudiSuiteMixin.scala
simon824: It work well if hive metastore is base on mysql  and will cause Concurrent Exception  if base on derby. 
Derby is the default config value, and ut is base on it .

pan3793: How does Hudi community do such a Hive integration test?

simon824:
```
spark.sql("...")
```
It works well with derby.

yaooqinn (approved):

pan3793:
Thanks, merging to master for v1.4.0 and unblock #1029

pom.xml
```diff
@@ -1034,37 +1030,81 @@
 
             <dependency>
                 <groupId>org.apache.hudi</groupId>
-                <artifactId>hudi-spark3_${scala.binary.version}</artifactId>
+                <artifactId>hudi-spark-common_${scala.binary.version}</artifactId>
```
yanghua, line 1033: +1 for introducing the common artifact

yanghua:
thought: shall we introduce the integration test for the third-party's basic function, e.g. write path & read path?

Issue #1031, opened 2021-09-06
[SUB-TASK][KPIP-1] Implement api: /sessions/count


yanghua:
It's a subtask of umbrella issue #950

PR #1038, opened 2021-09-06
[KYUUBI #1031] Implement api: /sessions/count

yanghua:

### DIFF
--- a/dev/dependencyList
+++ b/dev/dependencyList
@@ -47,8 +47,10 @@ jcl-over-slf4j/1.7.30//jcl-over-slf4j-1.7.30.jar
 jersey-client/2.30//jersey-client-2.30.jar
 jersey-common/2.30//jersey-common-2.30.jar
 jersey-container-servlet-core/2.30//jersey-container-servlet-core-2.30.jar
+jersey-entity-filtering/2.30//jersey-entity-filtering-2.30.jar
 jersey-hk2/2.30//jersey-hk2-2.30.jar
 jersey-media-jaxb/2.30//jersey-media-jaxb-2.30.jar
+jersey-media-json-jackson/2.30//jersey-media-json-jackson-2.30.jar
 jersey-server/2.30//jersey-server-2.30.jar
 jetty-http/9.4.41.v20210516//jetty-http-9.4.41.v20210516.jar
 jetty-io/9.4.41.v20210516//jetty-io-9.4.41.v20210516.jar
--- a/dev/dependencyList
+++ b/dev/dependencyList
@@ -33,6 +33,7 @@ htrace-core4/4.1.0-incubating//htrace-core4-4.1.0-incubating.jar
 jackson-annotations/2.11.4//jackson-annotations-2.11.4.jar
 jackson-core/2.11.4//jackson-core-2.11.4.jar
 jackson-databind/2.11.4//jackson-databind-2.11.4.jar
+jackson-module-jaxb-annotations/2.9.9//jackson-module-jaxb-annotations-2.9.9.jar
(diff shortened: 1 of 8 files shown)


kyuubi-server/src/test/scala/org/apache/kyuubi/server/RestFrontendServiceSuite.scala
```diff
@@ -80,10 +83,41 @@ class RestFrontendServiceSuite extends KyuubiFunSuite{
     server.stop()
   }
 
-  class RestNoopServer extends NoopServer {
+  test("kyuubi rest frontend service sessions resource") {
+    val server = new RestNoopServer()
```
ulysses-you, line 87: shall we add a closure and simplify to `withKyuubiRestServer {}`

kyuubi-server/src/test/scala/org/apache/kyuubi/server/RestFrontendServiceSuite.scala
```diff
@@ -61,29 +64,62 @@ class RestFrontendServiceSuite extends KyuubiFunSuite{
...
+          assert(html.toLowerCase(Locale.ROOT).equalsIgnoreCase(expectedStr))
+        }
+    }
+  }
+
+  def withKyuubiRestServer(f: (RestNoopServer, RestFrontendService) => Unit): Unit = {
     val server = new RestNoopServer()
     server.stop()
     val conf = KyuubiConf()
     conf.set(KyuubiConf.FRONTEND_REST_BIND_HOST, "localhost")
```
ulysses-you, line 94: we can also specify the port here

kyuubi-server/src/main/scala/org/apache/kyuubi/server/RestFrontendService.scala
```diff
@@ -114,11 +114,13 @@ private[server] class RestFrontendService private(name: String, be: BackendServi
   private def stopHttpServer(): Unit = {
     if (jettyServer != null) {
       try {
+        connector.stop()
```
ulysses-you, line 117: shall we do this in another try-catch ?

kyuubi-server/src/test/scala/org/apache/kyuubi/server/RestFrontendServiceSuite.scala
yanghua: Here for testing, we can use the default port?

ulysses-you: We can return host:port in closure so we don't need hard code them. And the `server` seems not useful, seems `frontendService`  is enough. what do you thank ?

yanghua: reasonable, +1

dev/dependencyList
```diff
@@ -33,6 +33,7 @@ htrace-core4/4.1.0-incubating//htrace-core4-4.1.0-incubating.jar
 jackson-annotations/2.11.4//jackson-annotations-2.11.4.jar
 jackson-core/2.11.4//jackson-core-2.11.4.jar
 jackson-databind/2.11.4//jackson-databind-2.11.4.jar
+jackson-module-jaxb-annotations/2.9.9//jackson-module-jaxb-annotations-2.9.9.jar
```
pan3793, line 36: we should pin `jackson-module-jaxb-annotations:${jackson.version}` in root `pom.xml`

yaooqinn (approved):

ulysses-you (approved):

ulysses-you:
thank you @yanghua , merging to master

PR #1040, opened 2021-09-06
Add NoOpBootstrapIndex config for hudi table

simon824:
Add create table options for DataLakeSuiteMixin

### DIFF
--- a/kyuubi-common/src/test/scala/org/apache/kyuubi/operation/BasicHudiJDBCTests.scala
+++ b/kyuubi-common/src/test/scala/org/apache/kyuubi/operation/BasicHudiJDBCTests.scala
@@ -67,13 +67,8 @@ trait BasicHudiJDBCTests extends JDBCTestUtils with HudiSuiteMixin {
     val schema = "default"
     val tableType = "TABLE"
 
-    val opts = getTableOptions(
-      "primaryKey" -> "id",
-      "preCombineField" -> "ts"
-    )
-
     withJdbcStatement(table) { statement =>
-      val sql =
+      statement.execute(
         s"""
            | create table $table (
            |  id int,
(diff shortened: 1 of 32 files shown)


yaooqinn:
cc @pan3793 @yanghua

pan3793:
Have a second thought on it, I change my mind now, because it seems that `tableOptions` is not a common field for DDL, as we create several tables in different methods using different `options`.

But as we disscue before, it's valuable to set `hoodie.bootstrap.index.class=org.apache.hudi.common.bootstrap.index.NoOpBootstrapIndex` for hudi tables.

@simon824 Sorry for making you back and forth.

simon824:
> Have a second thought on it, I change my mind now, because it seems that `tableOptions` is not a common field for DDL, as we create several tables in different methods using different `options`.
> 
> But as we disscue before, it's valuable to set `hoodie.bootstrap.index.class=org.apache.hudi.common.bootstrap.index.NoOpBootstrapIndex` for hudi tables.
> 
> @simon824 Sorry for making you back and forth.

what about separate common config like `hoodie.bootstrap.index.class`  and custom config like `primaryKey`?

https://github.com/apache/incubator-kyuubi/pull/1040/files#diff-98f5ef031c8c38b4be598220eaa8c0f8c23522b32fa8c63fdcf71e46f44a9dacR42-R52

pan3793 (approved):

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/BasicHudiJDBCTests.scala
```diff
@@ -75,13 +75,14 @@ trait BasicHudiJDBCTests extends JDBCTestUtils with HudiSuiteMixin {
            | ) using $format
            | options (
            |   primaryKey = 'id',
-           |   preCombineField = 'ts'
+           |   preCombineField = 'ts',
+           |   hoodie.bootstrap.index.class =
```
pan3793, line 79: remove the space at end of line

pan3793:
Thanks, merging to master

PR #1054, opened 2021-09-08
[KYUUBI #1018] Set spark.sql.execution.topKSortFallbackThreshold to 10000

### LINKED ISSUE #1018: Set spark.sql.execution.topKSortFallbackThreshold to a reasonable value
### 1. Describe the feature

In https://github.com/apache/spark/pull/33904, I faced a performance issue with the topK scenario where K is very large which can be avoided by setting spark.sql.execution.topKSortFallbackThreshold less than the K here.

For Kyuubi users who run SQL only, improper topK SQLs will hold engines for quite a long time and they don't know why. 

With a proper value for spark.sql.execution.topKSortFallbackThreshold  is more suitable for Kyuubi's cases

I'd suggest this value be set to 10000

byyue:

To avoid performance issues in the topK scenario and close #1018 .


yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #1055, opened 2021-09-08
Refine zorder plan

ulysses-you:

- Change origin `OptimizeZorderCommand` to `OptimizeZorderStatement` which holds the parsed plan
- Add `OptimizeZorderCommand` which delegate an another command(e.g. `InsertIntoHive`) to execute; the reason we should add a new command (not reuse `InserIntoStatement`) is we may need to implement read and overwrite same table in future. And also it's tricky to add `InserIntoStatement` during analyze.
- Change `ZorderBeforeWrite` to `ResolveZorder`, so we can resolve `OptimizeZorderStatement` to `OptimizeZorderCommand`

So the code should look cleaner


dev/kyuubi-extension-spark-3-1/pom.xml
```diff
@@ -32,6 +32,12 @@
     <name>Kyuubi Project Dev Spark Extensions (for Spark 3.1)</name>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
ulysses-you, line 39: this is for `KyuubiSQLException`

ulysses-you:
cc @qphien @hzxiongyinke @cfmcgrady @yaooqinn @pan3793

dev/kyuubi-extension-spark-3-1/pom.xml
ulysses-you: ah, seems we need to define a new exception in extension.

pan3793: Yeah, decouple it from kyuubi-common will make life simple :)

cfmcgrady (approved):
LGTM

yaooqinn (approved):

ulysses-you:
thanks for review, merged to master

PR #1057, opened 2021-09-08
[KYUUBI#1056] Output error log when currentEngine is None

hddong:

When createSpark in SparkSqlEngine get sparksession or execute initialize.sql failed, we cannot get any error info in logs. At this time, currentEngine is None.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -152,6 +152,9 @@ object SparkSQLEngine extends Logging {
       countDownLatch.await()
     } catch {
       case t: Throwable =>
+        if (!currentEngine.isDefined) {
+          error("Create SparkSQL Engine Failed", t)
+        }
         currentEngine.foreach { engine =>
           val status =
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -152,6 +152,9 @@ object SparkSQLEngine extends Logging {
       countDownLatch.await()
     } catch {
       case t: Throwable =>
+        if (!currentEngine.isDefined) {
```
cfmcgrady, line 155: How about 
```scala
      case t: Throwable if currentEngine.isDefined =>
        currentEngine.foreach { engine =>
          val status =
            engine.engineStatus.copy(diagnostic = s"Error State SparkSQL Engine ${t.getMessage}")
          EventLoggingService.onEvent(status)
          error(status, t)
          engine.stop()
        }
      case t =>
        error("Create SparkSQL Engine Failed", t)
```

hddong: Thanks, done.

cfmcgrady (approved):

ulysses-you:
@hddong  the change LGTM, is it possible to add a test ?

yaooqinn (approved):

yaooqinn:
thanks, merged to master and v1.3.1

Issue #1059, opened 2021-09-08
Add Plan Only Operations


yaooqinn:
### 1. Describe the feature

Introduce a session-level configuration to  change ExecuteStatement to PlanOnlyStatement,

PlanOnlyStatement will only parse, analyze or optimize the statements instead of executing them.

This feature could run all Spark SQLs to verify the semantic compatibility without having side effects, which is very useful for testing, migrating without query rewrite.

PR #1060, opened 2021-09-08
[KYUUBI #1059] Add Plan Only Operations

yaooqinn:

Introduce a session-level configuration `operation.plan.only.mode` to change ExecuteStatement to PlanOnlyStatement,

PlanOnlyStatement will only parse, analyze or optimize the statements instead of executing them.

This feature could run all Spark SQLs to verify the semantic compatibility without having side effects, which is very useful for testing, migrating without query rewrite.

This is similar to the `explain` command


yaooqinn:
cc @pan3793 @ulysses-you @yanghua @hddong @turboFei

ulysses-you (approved):

cfmcgrady:
Should we also log statement events when the `operation.plan.only.mode` is not `None`?

ulysses-you:
> Should we also log statement events when the operation.plan.only.mode is not None?

Not see the requirements, this operation is more similar with`GetSchemas` etc ..

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
```diff
@@ -58,7 +61,15 @@ class SparkSQLOperationManager private (name: String) extends OperationManager(n
       runAsync: Boolean,
       queryTimeout: Long): Operation = {
     val spark = getSparkSession(session.handle)
-    val operation = new ExecuteStatement(spark, session, statement, runAsync, queryTimeout)
+
+    val operationMode = {
+      val modeStr = spark.conf.get(OPERATION_PLAN_ONLY.key, NONE.toString).toUpperCase(Locale.ROOT)
```
pan3793, line 66: `.toUpperCase(Locale.ROOT)` seems not necessary

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/JDBCTests.scala
```diff
@@ -379,4 +379,31 @@ trait JDBCTests extends BasicJDBCTests {
...
+
+  test("KYUUBI #1059: Plan only operations") {
+    val ddl = "create table t(a int) using parquet"
+    val dql = "select * from t"
+    withJdbcStatement("t") { statement =>
+      try {
+        statement.execute("SET kyuubi.operation.plan.only.mode=optimize")
+        val set = statement.executeQuery(ddl)
+        assert(set.next())
+        assert(set.getString("plan") startsWith "Create")
```
pan3793, line 391: Is the result stable? If not stable, can we paste a demo result here?

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/JDBCTests.scala
```diff
@@ -379,4 +379,31 @@ trait JDBCTests extends BasicJDBCTests {
...
+        val e1 = intercept[SQLException](statement.executeQuery(dql))
+        assert(e1.getMessage.contains("Table or view not found"))
+        statement.execute("SET kyuubi.operation.plan.only.mode=analyze")
+        val e2 = intercept[SQLException](statement.executeQuery(dql))
+        assert(e2.getMessage.contains("Table or view not found"))
+        statement.execute("SET kyuubi.operation.plan.only.mode=parse")
+        val set1 = statement.executeQuery(dql)
+        assert(set1.next())
+        assert(set1.getString("plan") contains "Unresolved")
```
pan3793, line 400: ditto

yanghua:
LGTM

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
yaooqinn: we are getting it from spark runtime config

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/JDBCTests.scala
yaooqinn: If assert fails, we can see the original plan

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
hddong, line 66: If we set a default value other than `None` in kyuubi server, seems the setting will not work here.

pan3793 (approved):
LGTM

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
yaooqinn: updated

hddong:
LGTM

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/PlanOnlyOperationSuite.scala
```diff
@@ -0,0 +1,65 @@
+/*
```
timothy65535, line 1: `PlanOnlyOperationSuite.scala` to `KyuubiPlanOnlyOperationSuite.scala` will be better.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
```diff
@@ -52,13 +55,21 @@ class SparkSQLOperationManager private (name: String) extends OperationManager(n
...
       statement: String,
       runAsync: Boolean,
       queryTimeout: Long): Operation = {
     val spark = getSparkSession(session.handle)
-    val operation = new ExecuteStatement(spark, session, statement, runAsync, queryTimeout)
+
+    val operationModeStr =
+      spark.conf.get(OPERATION_PLAN_ONLY.key, operationModeDefault).toUpperCase(Locale.ROOT)
+    val operation = OperationModes.withName(operationModeStr) match {
+      case NONE => new ExecuteStatement(spark, session, statement, runAsync, queryTimeout)
```
timothy65535, line 70: nit: do we need to consider the following case?
```
ser kyuubi.operation.plan.only.mode=abc;
select 1;
```

yaooqinn: would you like to make a follow up？

timothy65535: Sorry, don't want to do.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/PlanOnlyOperationSuite.scala
timothy65535: ?

Issue #1034, opened 2021-09-06
Engine may deadlock when close operationLog


qiuliang988:
### 1. Describe the bug
The lock order for close operationLog is OperationLog -> RootLogger
```
"FrontendServiceHandler-Pool: Thread-588":
    at org.apache.log4j.Category.callAppenders(Category.java:204)
    - waiting to lock <0x000000015fc4b7e8> (a org.apache.log4j.spi.RootLogger)
    at org.apache.log4j.Category.forcedLog(Category.java:391)
    at org.apache.log4j.Category.log(Category.java:856)
    at org.slf4j.impl.Log4jLoggerAdapter.error(Log4jLoggerAdapter.java:576)
    at org.apache.kyuubi.Logging.error(Logging.scala:78)
    at org.apache.kyuubi.Logging.error$(Logging.scala:76)
    at org.apache.kyuubi.operation.log.OperationLog.error(OperationLog.scala:86)
    at org.apache.kyuubi.operation.log.OperationLog.close(OperationLog.scala:137)
    - locked <0x00000002c3818180> (a org.apache.kyuubi.operation.log.OperationLog)
    at org.apache.kyuubi.engine.spark.operation.SparkOperation.$anonfun$close$1(SparkOperation.scala:142)
    at org.apache.kyuubi.engine.spark.operation.SparkOperation.$anonfun$close$1$adapted(SparkOperation.scala:142)
    at org.apache.kyuubi.engine.spark.operation.SparkOperation$$Lambda$2491/880482099.apply(Unknown Source)
    at scala.Option.foreach(Option.scala:407)
    at org.apache.kyuubi.engine.spark.operation.SparkOperation.close(SparkOperation.scala:142)
    at org.apache.kyuubi.operation.OperationManager.closeOperation(OperationManager.scala:103)
    at org.apache.kyuubi.session.AbstractSession.$anonfun$closeOperation$1(AbstractSession.scala:197)
    at org.apache.kyuubi.session.AbstractSession$$Lambda$2490/481951832.apply$mcZ$sp(Unknown Source)
    at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.java:23)
    at org.apache.kyuubi.session.AbstractSession.withAcquireRelease(AbstractSession.scala:78)
    at org.apache.kyuubi.session.AbstractSession.closeOperation(AbstractSession.scala:196)
[...]

PR #1061, opened 2021-09-08
[KYUUBI #1034]Engine may deadlock when close operationLog

qiuliang988:


yaooqinn:
hi @qiuliang988 can we restore this PR？

PR #1071, opened 2021-09-09
[KYUUBI #1070] Support insert zorder by table properties 

### LINKED ISSUE #1070: [FEATURE] Support insert zorder by table properties
### Code of Conduct

### Search before asking

### Describe the feature

Add zorder before writing dynamically.

This feature support:
- insert into datasource table
- create datasource table as select
- insert into hive table
- create hive table as select

### Motivation

Zorder is a popular approach of data skipping. We have supported `OPTIMIZE <table> [WHERE <predicate>] ZORDER BY col1, col2, ...` but it's not enough for user.

Let's say, If we have a  hot table that adds a partition every day so we decide to apply zorder to optimize it. The table properties become to a good place to store the zorder info. When you realize that this is marked as a zorder table, we can optimize it.

### Describe the solution

Support parse two keys in table properties:
- kyuubi.zorder.enabled
  decide if we should add zorder before writing this table
- kyuubi.zorder.cols
  which columns should be included with zorder

And we also add a global config `spark.sql.optimizer.insertZorderBeforeWriting.enabled`.

### Additional context

_No response_

ulysses-you:

Closes #1070 

### DIFF
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
@@ -20,8 +20,7 @@ package org.apache.kyuubi.sql
 import org.apache.spark.sql.SparkSessionExtensions
 
 import org.apache.kyuubi.sql.sqlclassification.KyuubiSqlClassification
-import org.apache.kyuubi.sql.zorder.ResolveZorder
-import org.apache.kyuubi.sql.zorder.ZorderSparkSqlExtensionsParser
+import org.apache.kyuubi.sql.zorder.{InsertZorderBeforeWritingDatasource, InsertZorderBeforeWritingHive, ResolveZorder, ZorderSparkSqlExtensionsParser}
 
 // scalastyle:off line.size.limit
 /**
@@ -35,6 +34,13 @@ class KyuubiSparkSQLExtension extends (SparkSessionExtensions => Unit) {
     // inject zorder parser and related rules
     extensions.injectParser{ case (_, parser) => new ZorderSparkSqlExtensionsParser(parser) }
     extensions.injectResolutionRule(ResolveZorder)
+
+    // Note that:
+    // InsertZorderBeforeWritingDatasource and InsertZorderBeforeWritingHive
+    // should be applied before
+    // RepartitionBeforeWrite and RepartitionBeforeWriteHive
+    extensions.injectPostHocResolutionRule(InsertZorderBeforeWritingDatasource)
+    extensions.injectPostHocResolutionRule(InsertZorderBeforeWritingHive)
 
     extensions.injectPostHocResolutionRule(KyuubiSqlClassification)
     extensions.injectPostHocResolutionRule(RepartitionBeforeWrite)
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
@@ -39,6 +39,7 @@ class KyuubiSparkSQLExtension extends (SparkSessionExtensions => Unit) {
     // InsertZorderBeforeWritingDatasource and InsertZorderBeforeWritingHive
     // should be applied before
     // RepartitionBeforeWrite and RepartitionBeforeWriteHive
+    // because we can only apply one of them (i.e. Global Sort or Repartition)
     extensions.injectPostHocResolutionRule(InsertZorderBeforeWritingDatasource)
     extensions.injectPostHocResolutionRule(InsertZorderBeforeWritingHive)
 
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/InsertZorderBeforeWriting.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/InsertZorderBeforeWriting.scala
@@ -31,6 +31,7 @@ import org.apache.spark.sql.hive.execution.{CreateHiveTableAsSelectCommand, Inse
 import org.apache.kyuubi.sql.KyuubiSQLConf
 
 /**
+ * TODO: shall we forbid zorder if it's dynamic partition inserts ?
  * Insert zorder before writing datasource if the target table properties has zorder properties
  */
 case class InsertZorderBeforeWritingDatasource(session: SparkSession)
@@ -60,6 +61,7 @@ case class InsertZorderBeforeWritingDatasource(session: SparkSession)
 }
 
 /**
+ * TODO: shall we forbid zorder if it's dynamic partition inserts ?
  * Insert zorder before writing hive if the target table properties has zorder properties
  */
 case class InsertZorderBeforeWritingHive(session: SparkSession)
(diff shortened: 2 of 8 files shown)


dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/ZorderBytesUtils.scala
```diff
@@ -71,23 +72,23 @@ object ZorderBytesUtils {
...
       case l: Long =>
-        ZorderBytesUtils.longToByte(l)
+        longToByte(l)
       case f: Float =>
-        ZorderBytesUtils.floatToByte(f)
+        floatToByte(f)
       case d: Double =>
-        ZorderBytesUtils.doubleToByte(d)
-      case str: String =>
-        ZorderBytesUtils.stringToByte(str)
```
ulysses-you, line 88: this is a bug, we should use `UTF8String` instead of `String`..

ulysses-you:
cc @qphien @cfmcgrady @hddong @timothy65535 @yaooqinn @pan3793 @yanghua

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/InsertZorderBeforeWritingDatasource.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  def isZorderEnabled(props: Map[String, String]): Boolean = {
+    props.contains(KYUUBI_ZORDER_ENABLED) &&
+      "true".equalsIgnoreCase(props(KYUUBI_ZORDER_ENABLED)) &&
+      props.contains(KYUUBI_ZORDER_COLS)
+  }
+
+  def getZorderColumns(props: Map[String, String]): Seq[String] = {
+    val cols = props.get(KYUUBI_ZORDER_COLS)
+    assert(cols.isDefined)
```
cfmcgrady, line 111: Zorder required at least one column?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
```diff
@@ -36,6 +35,13 @@ class KyuubiSparkSQLExtension extends (SparkSessionExtensions => Unit) {
     extensions.injectParser{ case (_, parser) => new ZorderSparkSqlExtensionsParser(parser) }
     extensions.injectResolutionRule(ResolveZorder)
 
+    // Note that:
```
cfmcgrady, line 38: Seems we can add a unit test for this check.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/InsertZorderBeforeWritingDatasource.scala
ulysses-you: yea, it has already checked with
```
    val cols = getZorderColumns(catalogTable.properties)
    val attrs = plan.output.map(attr => (attr.name, attr)).toMap
    if (cols.exists(!attrs.contains(_))) {
      logWarning(s"target table does not contain all zorder cols: ${cols.mkString(",")}, " +
        s"please check your table properties ${KYUUBI_ZORDER_COLS}.")
      plan
```

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
ulysses-you: The order is hard code in extension so I think it's safe, and also checked plan if insert zorder or not:
```
  def canInsertZorder(query: LogicalPlan): Boolean = query match {
    case Project(_, child) => canInsertZorder(child)
    // TODO: actually, we can force zorder even if existed some shuffle
    case _: Sort => false
    case _: RepartitionByExpression => false
    case _: Repartition => false
    case _ => true
  }
```

We can add some unit test for this method (e.g. the plan top level node is `RepartitionByExpression`).

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
```diff
@@ -20,8 +20,7 @@ package org.apache.kyuubi.sql
 import org.apache.spark.sql.SparkSessionExtensions
 
 import org.apache.kyuubi.sql.sqlclassification.KyuubiSqlClassification
-import org.apache.kyuubi.sql.zorder.ResolveZorder
-import org.apache.kyuubi.sql.zorder.ZorderSparkSqlExtensionsParser
+import org.apache.kyuubi.sql.zorder.{InsertZorderBeforeWritingDatasource, InsertZorderBeforeWritingHive, ResolveZorder, ZorderSparkSqlExtensionsParser}
```
timothy65535, line 23: this line length exceeds 100 characters.

yaooqinn (approved):

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
ulysses-you: it's ok in `import`, see the scalastyle configuration
```
  <check level="error" class="org.scalastyle.file.FileLineLengthChecker" enabled="true">
    <parameters>
      <parameter name="maxLineLength"><![CDATA[100]]></parameter>
      <parameter name="tabSize"><![CDATA[2]]></parameter>
      <parameter name="ignoreImports">true</parameter>
    </parameters>
  </check>
```

ulysses-you:

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
timothy65535: btw, my Intellij IDEA always can't work, how to solve it? thanks.

image

ulysses-you: I think it's your JDK version issue,  the following is my JDK

image

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/zorder/InsertZorderBeforeWriting.scala
```diff
@@ -0,0 +1,155 @@
...
+      } else {
+        ctas.copy(query = newQuery)
+      }
+
+    case _ => plan
+  }
+}
+
+/**
+ * TODO: shall we forbid zorder if it's dynamic partition inserts ?
```
qphien, line 64: I think we can throw an exception when table partition columns are specified as zorder columns since it makes no sense.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala
timothy65535: ok, will try it

Issue #1052, opened 2021-09-08
[SUB-TASK][KPIP-1] Implement api: /sessions (POST)


yanghua:
It's a subtask of umbrella issue #950

PR #1073, opened 2021-09-10
[KYUUBI #1052] Implement api: /sessions (POST)

yanghua:

### DIFF
--- a/kyuubi-server/pom.xml
+++ b/kyuubi-server/pom.xml
@@ -87,6 +87,28 @@
         </dependency>
 
         <dependency>
+            <groupId>org.glassfish.jersey.test-framework</groupId>
+            <artifactId>jersey-test-framework-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jersey.test-framework.providers</groupId>
+            <artifactId>jersey-test-framework-provider-jetty</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.eclipse.jetty</groupId>
+                    <artifactId>jetty-util</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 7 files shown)


kyuubi-server/pom.xml
```diff
@@ -86,6 +86,28 @@
...
+            <groupId>org.glassfish.jersey.test-framework</groupId>
+            <artifactId>jersey-test-framework-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish.jersey.test-framework.providers</groupId>
+            <artifactId>jersey-test-framework-provider-jetty</artifactId>
+            <scope>test</scope>
+            <exclusions>
```
turboFei, line 99: these exclusions has been declared in parent pom.
Should we remove them here?

kyuubi-server/src/main/java/org/apache/kyuubi/server/api/v1/dto/SessionOpenedCount.java
```diff
@@ -33,4 +34,21 @@ public int getOpenSessionCount() {
...
+    if (!(o instanceof SessionOpenedCount)) {
+      return false;
+    }
+    SessionOpenedCount that = (SessionOpenedCount) o;
+    return getOpenSessionCount() == that.getOpenSessionCount();
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(getOpenSessionCount());
```
turboFei, line 52: Invoke openSessionCount directly? 
return Objects.hash(openSessionCount);

turboFei (approved):

kyuubi-server/src/main/java/org/apache/kyuubi/server/api/v1/dto/SessionOpenedCount.java
yanghua: OK

yanghua:
@turboFei Thanks for your review. Have addressed your concerns.

turboFei:
thanks, merged to master

Issue #916, opened 2021-08-10
Send DelegationTokens to SQL engines


zhouyifan279:
### 1. Describe the feature
This is another sub-task under umbrella issue #913 and depends on task #1007.

This task covers the work of sending Kyuubi Server maintained delegation tokens to SQL engines.

### 2. Motivation
See #913 

#### Server side
Create a dynamic proxy `KyuubiSyncThriftClientProxy` wtih field `engineId` read from `ServiceNodeInfo#sessionId`, field `appUser` from `EngineRef#appUser` and a  function `send` implemented with `TCLIService.Client#RenewDelegationToken`.

When method is invoked, `KyuubiSyncThriftClientProxy` will invoke `HadoopCredentialsManager#sendCredentialsIfNeeded` before invoking `KyuubiSyncThriftClient`'s corresponding method. 

Exceptions thrown by `sendCredentialsIfNeeded` are suppressed.

#### Engine side
SQL engine receives delegation tokens through `ThriftFrontendService#RenewDelegationToken`.
Delegation tokens that not exist in current credentials(`UserGroupInformation#getCurrentUser()#getCredentials()`) and delegation tokens whose issueDate is greater than the current one's issueDate are collected in a new Credentials object.

Then, according to different Spark `SchedulerBackend` implements, different delegation token update ways are selected:

- `LocalSchedulerBackend` -> Just add Credentials object into currentUser's credentials.
- `CoarseGrainedSchedulerBackend`(Deploy mode Standalone, YARN, Kubernetes and Mesos are covered here) -> Send serialized Credentials object to `CoarseGrainedSchedulerBackend#driverEndpoint`
- `MesosFineGrainedSchedulerBackend` -> throw exception since `MesosFineGrainedSchedulerBackend` does not have a 
delegation token update mechanism.

PR #1074, opened 2021-09-10
[KYUUBI #916] Send DelegationTokens to SQL engines

zhouyifan279:
This PR finishes issue #916 


zhouyifan279:
@yaooqinn @ulysses-you @pan3793 Ready to be reviewed.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLBackendService.scala
```diff
@@ -33,9 +38,40 @@ import org.apache.kyuubi.session.SessionManager
...
+        } else {
+          if (KyuubiHadoopUtils.getTokenIssueDate(newToken) >
+              KyuubiHadoopUtils.getTokenIssueDate(oldToken)) {
+            updateCreds.addToken(alias, newToken)
+          } else {
+            warn(s"Ignore token with earlier issue date: $newToken")
+          }
+        }
+      }
+      SparkContextHelper.updateDelegationTokens(
```
ulysses-you, line 72: shall we update delegation token if:
`KyuubiHadoopUtils.getTokenIssueDate(newToken) <= KyuubiHadoopUtils.getTokenIssueDate(oldToken)` ?

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClientHandler.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.kyuubi.session.SessionHandle
+
+class KyuubiSyncThriftClientHandler(
+    credentialsManager: HadoopCredentialsManager,
+    client: KyuubiSyncThriftClient,
+    sessionId: String,
+    appUser: String) extends InvocationHandler {
+
+  override def invoke(proxy: Any, method: Method, args: Array[AnyRef]): AnyRef = {
+    credentialsManager.sendCredentialsIfNeeded(
```
ulysses-you, line 32: what happen if user has given engine keytab ?

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -30,7 +30,67 @@ import org.apache.kyuubi.operation.FetchOrientation.FetchOrientation
 import org.apache.kyuubi.session.SessionHandle
 import org.apache.kyuubi.util.ThriftUtils
 
-class KyuubiSyncThriftClient(protocol: TProtocol) extends TCLIService.Client(protocol) {
+trait KyuubiSyncThriftClient {
```
yaooqinn, line 33: Is this necessary to create this？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLBackendService.scala
```diff
@@ -33,9 +38,40 @@ import org.apache.kyuubi.session.SessionManager
...
-  extends AbstractBackendService(name) {
+  extends AbstractBackendService(name) with Logging {
 
   def this(spark: SparkSession) = this(classOf[SparkSQLBackendService].getSimpleName, spark)
 
   override val sessionManager: SessionManager = new SparkSQLSessionManager(spark)
+
+  override def renewDelegationToken(
+      sessionHandle: SessionHandle,
+      authFactory: KyuubiAuthenticationFactory,
```
yaooqinn, line 49: where do we use it？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLBackendService.scala
zhouyifan279: How about `updateCreds.numberOfTokens() > 1` ?

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClientHandler.scala
zhouyifan279: Credentials will also be sent to engine. 

Actually, no matter engine has keytab or not is not, `DFSClient` and `HiveMetaStore` always use Tokens to talk with server. So there should be not difference.

zhouyifan279: Just think of that there is a problem in another case:  
User designates a HDFS service or Hive metastore server different from the one in kyuubi server's configuration.

It should be fine to `DFSClient` because we update tokens by alias and HDFS delegation token's alias is the NameNode address. We will just add some tokens not needed.

But it is a problem to `HiveMetaStoreClient` because Hive delegation token's alias is fixed to "hive.server2.delegation.token" in spark. The initial correct token will be updated to a token for another server.

To fix this problem, I think we should add the following procedure:

1.  At kyuubi server side, set Hive delegation token's service name to Hive metastore uris.
2. At engine side, before update the token, select the first token whose metastore uris is compatible with the uris defined in HiveConf. Metastore uris are compatible if they have at least one same uri.
3. Set the token's service to "" and add it to engine's credentials. (Hive metastore delegation token's service in engine has to be "" as `HiveMetaStoreClient` selects the first token whose service is "" and kind is "HIVE_DELEGATION_TOKEN" to talk to server.)

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -272,6 +272,19 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val CREDENTIALS_HIVE_EXTRA_METASTORE_URIS: ConfigEntry[Seq[String]] =
```
yaooqinn, line 275: why do we need this?

zhouyifan279: Kyuubi administrator can set this conf to let kyuubi renew token for extra Hive metastore servers, which are not in hive-site.xml.

Without this, if user specified any other Hive metastore server in JDBC url,  kyuubi engine can not access it after the initial token expires.

yaooqinn: > Kyuubi administrator can set this conf to let kyuubi renew token for extra Hive metastore servers, which are not in hive-site.xml.

why didn't he modify the hive-site.xml directly?

zhouyifan279: User specified Hive metastore server may not belong to the Hive metastore cluster defined in hive-site.xml.

As this is a patch of `HiveDelegationTokenProvider`, I'll remove it from this PR.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLBackendService.scala
zhouyifan279: Moved method to a new class `SparkThriftFrontendService`.

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
zhouyifan279: Removed dynamic proxy and send delegation tokens only during `ExecuteStatement#waitStatementComplete`.
No longer need to change `KyuubiSyncThriftClient` to a trait.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
```diff
@@ -120,6 +118,13 @@ class ExecuteStatement(
           val ke = KyuubiSQLException(s"UNKNOWN STATE for $statement")
           setOperationException(ke)
       }
+
+      val appUser = session.asInstanceOf[KyuubiSessionImpl].engine.appUser
+      val sessionManager = session.sessionManager.asInstanceOf[KyuubiSessionManager]
+      sessionManager.credentialsManager.sendCredentialsIfNeeded(
```
ulysses-you, line 124: do we need send token if remote state is terminated ?

zhouyifan279: No matter where in `ExecuteStatement#waitStatementComplete` we put the send method, there is possibility that Statement has been closed terminated when we send credentials.  

Let's make the possibility more obvious.

ulysses-you (approved):

yaooqinn (approved):

ulysses-you:
thanks, merging to master !

Issue #1068, opened 2021-09-09
[Bug] event log cann't upload to hdfs


zhang1002:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

when you set the hdfs's path and expect the event log upload to this path, the server will throw exception and stop.
the code will only generate the localPath.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1075, opened 2021-09-10
[KYUUBI #1068] event log can't upload to hdfs

zhang1002:
bug:
#1068 

This version, event log don't support upload file to hdfs


zhang1002:
cc @yaooqinn @ulysses-you @zhouyifan279

timothy65535:
hi @zhang1002, following steps will helpful for new patch
```
git checkout master
git pull --rebase origin master
git checkout -b <new branch>
```

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/events/EventLoggingServiceSuite.scala
```diff
@@ -32,7 +32,7 @@ import org.apache.kyuubi.operation.{JDBCTestUtils, OperationHandle}
 
 class EventLoggingServiceSuite extends WithSparkSQLEngine with JDBCTestUtils {
 
-  private val logRoot = Utils.createTempDir()
+  private val logRoot = "file:" + Utils.createTempDir()
```
timothy65535, line 35: `logRoot` from `Path` type to `String` type, so we don't need to call `toString` again.

```
  private val logRoot = "file:" + Utils.createTempDir()
  private val currentDate = Utils.getDateFromTimestamp(System.currentTimeMillis())


  override def withKyuubiConf: Map[String, String] = Map(
    KyuubiConf.ENGINE_EVENT_LOGGERS.key -> s"$JSON,$SPARK",
    KyuubiConf.ENGINE_EVENT_JSON_LOG_PATH.key -> logRoot.toString,
    "spark.eventLog.enabled" -> "true",
    "spark.eventLog.dir" -> logRoot.toString
  )
```

zhang1002: it's good.
i don't find this problem.

yaooqinn (approved):

PR #1085, opened 2021-09-12
Add forcedMaxOutputRows rule for limitation to avoid huge output unexpectly

i7xh:
Add MaxOutputRows rule for output rows limitation to avoid huge output unexpectedly

We support the PR feature with limitation that avoid huge output rows in user ad-hoc query unexpected，Generally ad-hoc query seems needle in a Haystack, user pick few computed result data in huge data from warehouse,  we mainly used in below cases:

- CASE 1:
```
SELECT [c1, c2, ...] 
```
- CASE 2:
```
WITH CTE AS (...)
SELECT [c1, c2, ...] FROM Express(CTE) ...
```

截屏2021-09-12 下午12 19 28

### DIFF
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
@@ -97,9 +97,8 @@ object KyuubiSQLConf {
 
   val WATCHDOG_FORCED_MAXOUTPUTROWS =
     buildConf("spark.sql.watchdog.forcedMaxOutputRows")
-      .doc("Add MaxOutputRows rule for output rows limitation " +
-        "to avoid huge output rows of non-limit query unexpectedly, " +
-        "it's a optional, it's optional that works with defined")
+      .doc("Add ForcedMaxOutputRows rule to avoid huge output rows of non-limit query " +
+        "unexpectedly, it's optional that works with defined")
       .version("1.4.0")
       .intConf
       .createOptional
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
@@ -94,4 +94,13 @@ object KyuubiSQLConf {
       .version("1.4.0")
       .intConf
       .createOptional
+
+  val WATCHDOG_FORCED_MAXOUTPUTROWS =
+    buildConf("spark.sql.watchdog.forcedMaxOutputRows")
+      .doc("Add MaxOutputRows rule for output rows limitation " +
+        "to avoid huge output rows of non-limit query unexpectedly, " +
+        "it's a optional, it's optional that works with defined")
+      .version("1.4.0")
+      .intConf
+      .createOptional
 }
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
@@ -19,7 +19,7 @@ package org.apache.kyuubi.sql.watchdog
 
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.dsl.expressions._
-import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, GlobalLimit, LogicalPlan, Project}
+import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, Limit, LogicalPlan, Project}
 import org.apache.spark.sql.catalyst.rules.Rule
 
 import org.apache.kyuubi.sql.KyuubiSQLConf
@@ -41,42 +41,31 @@ import org.apache.kyuubi.sql.KyuubiSQLConf
 * SELECT [c1, c2, ...] FROM CTE ...
 * }}}
 *
-* case 3 (customize yourself):
-*  {{{
-*   INSERT OVERWRITE DIRECTORY (path=STRING)?
-*   USING format OPTIONS ([option1_name "option1_value", option2_name "option2_value", ...])
-*   SELECT ...
-* }}}
 * The Logical Rule add a GlobalLimit node before root project
 * */
 case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPlan] {
-  override def apply(plan: LogicalPlan): LogicalPlan = {
-    if (!plan.resolved) {
-      plan
-    } else {
-      conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
-        case Some(forcedMaxOutputRows) => plan.maxRows match {
-          case Some(maxRows) => if (maxRows > forcedMaxOutputRows) {
-            GlobalLimit(forcedMaxOutputRows, plan)
-          } else {
-            plan
-          }
-          case None => plan match {
-            case project: Project => GlobalLimit(forcedMaxOutputRows, project)
-            case agg: Aggregate => agg.maxRows match {
-              case Some(maxRows) => if (maxRows > forcedMaxOutputRows) {
-                GlobalLimit(forcedMaxOutputRows, agg)
-              } else {
-                agg
-              }
-              case None => GlobalLimit(forcedMaxOutputRows, agg)
-            }
-            // TODO: Customize your required process node
-            case _ => plan
-          }
+
+  private def canInsertLimit(p: LogicalPlan): Boolean = {
+
+    conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
+      case Some(forcedMaxOutputRows) => val supported = p match {
+          case _: Project => true
+          case _: Aggregate => true
+          case Limit(_, _) => true
+          case _ => false
         }
-        case None => plan
-      }
+        supported && !p.maxRows.exists(_ <= forcedMaxOutputRows)
+      case None => false
+    }
+
+  }
+
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    plan match {
+      case p if p.resolved && canInsertLimit(p) => Limit(
+        conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS).get, plan)
+      case _ => plan
     }
   }
+
 }
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
@@ -45,9 +45,9 @@ import org.apache.kyuubi.sql.KyuubiSQLConf
 * */
 case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPlan] {
 
-  private def canInsertLimit(p: LogicalPlan): Boolean = {
+  private def canInsertLimit(p: LogicalPlan, maxOutputRowsOpt: Option[Int]): Boolean = {
 
-    conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
+    maxOutputRowsOpt match {
       case Some(forcedMaxOutputRows) => val supported = p match {
           case _: Project => true
           case _: Aggregate => true
@@ -61,9 +61,10 @@ case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPl
   }
 
   override def apply(plan: LogicalPlan): LogicalPlan = {
+    val maxOutputRowsOpt = conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS)
     plan match {
-      case p if p.resolved && canInsertLimit(p) => Limit(
-        conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS).get, plan)
+      case p if p.resolved && canInsertLimit(p, maxOutputRowsOpt) => Limit(
+        maxOutputRowsOpt.get, plan)
       case _ => plan
     }
   }
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
@@ -55,16 +55,27 @@ case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPl
       plan
     } else {
       conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
-        case Some(forcedMaxOutputRows) => plan match {
-          case project: Project => GlobalLimit(forcedMaxOutputRows, project)
-          case agg: Aggregate => agg.maxRows match {
-            case Some(_) => agg
-            case None => GlobalLimit(forcedMaxOutputRows, agg)
+        case Some(forcedMaxOutputRows) => plan.maxRows match {
+          case Some(maxRows) => if (maxRows > forcedMaxOutputRows) {
+            GlobalLimit(forcedMaxOutputRows, plan)
+          } else {
+            plan
           }
-          // TODO: Customize your required process node
-          case _ => plan
+          case None => plan match {
+            case project: Project => GlobalLimit(forcedMaxOutputRows, project)
+            case agg: Aggregate => agg.maxRows match {
+              case Some(maxRows) => if (maxRows > forcedMaxOutputRows) {
+                GlobalLimit(forcedMaxOutputRows, agg)
+              } else {
+                agg
+              }
+              case None => GlobalLimit(forcedMaxOutputRows, agg)
+            }
+            // TODO: Customize your required process node
+            case _ => plan
+          }
         }
-        case _ => plan
+        case None => plan
       }
     }
   }
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
@@ -21,7 +21,6 @@ import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.dsl.expressions._
 import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, GlobalLimit, LogicalPlan, Project}
 import org.apache.spark.sql.catalyst.rules.Rule
-import org.apache.spark.sql.catalyst.trees.Origin
 
 import org.apache.kyuubi.sql.KyuubiSQLConf
 
@@ -57,20 +56,12 @@ case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPl
     } else {
       conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
         case Some(forcedMaxOutputRows) => plan match {
-          case project: Project => project.origin match {
-            case Origin(Some(_), Some(0)) => GlobalLimit(forcedMaxOutputRows, project)
-            case _ => project
+          case project: Project => GlobalLimit(forcedMaxOutputRows, project)
+          case agg: Aggregate => agg.maxRows match {
+            case Some(_) => agg
+            case None => GlobalLimit(forcedMaxOutputRows, agg)
           }
-          case agg: Aggregate => agg.origin match {
-            case Origin(None, None) => agg.maxRows match {
-              case Some(_) => agg
-              case None => GlobalLimit(forcedMaxOutputRows, agg)
-            }
-            case _ => agg
-          }
-
           // TODO: Customize your required process node
-
           case _ => plan
         }
         case _ => plan
--- a/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
+++ b/dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
@@ -19,10 +19,9 @@ package org.apache.kyuubi.sql.watchdog
 
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.dsl.expressions._
-import org.apache.spark.sql.catalyst.plans.logical.{GlobalLimit, LogicalPlan, Project}
+import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, GlobalLimit, LogicalPlan, Project}
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.catalyst.trees.Origin
-import org.apache.spark.sql.execution.command.InsertIntoDataSourceDirCommand
 
 import org.apache.kyuubi.sql.KyuubiSQLConf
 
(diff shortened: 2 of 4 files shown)


dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala
```diff
@@ -0,0 +1,73 @@
...
+*   WITH CTE AS (
+*   ...)
+* SELECT [c1, c2, ...] FROM CTE ...
+* }}}
+*
+* The Logical Rule add a GlobalLimit node before root project
+* */
+case class MaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    if (!plan.resolved) {
```
ulysses-you, line 55: it's better to inject this rule into optimizer: `injectOptimizerRule` then we don't need this check

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala
```diff
@@ -0,0 +1,73 @@
...
+      plan
+    } else {
+      conf.getConf(KyuubiSQLConf.WATCHDOG_MAX_OUTPUTROWS) match {
+        case Some(maxOutputRows) => plan match {
```
ulysses-you, line 59: how about match `plan.maxRows > maxOutputRows` ?

ulysses-you (commented):
thank you @i7xh for the work , the idea looks good

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala
i7xh: It's origin limit the non-limit query, but It cann't cover with existed limit statement, It's easy to misunderstanding the option name，I rename `spark.sql.watchdog.maxOutputRows` to `spark.sql.watchdog.forcedMaxOutputRows`

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala
i7xh: the rule will insert GlobalLimit node in tree nodes, it changed original logicalPlan target, it make difference target between Analyzer and Optimizer, yet It should be idempotent

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala
i7xh: > how about match `plan.maxRows > maxOutputRows` ?

if non-limit statement, plan.maxRows is None type，the rule will match non-limit statement and skip statement with limit

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
```diff
@@ -0,0 +1,73 @@
...
+case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    if (!plan.resolved) {
+      plan
+    } else {
+      conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
+        case Some(forcedMaxOutputRows) => plan match {
+          case insert@InsertIntoDataSourceDirCommand(_, _, Project(_, _), _) =>
+            insert.copy(query = GlobalLimit(forcedMaxOutputRows, insert.query))
+          case project@Project(_, _) => project.origin match {
```
ulysses-you, line 62: we can match more nodes e.g. (Filter, Aggregate, Join ...), how about:
```
    case _: ExecutedCommandExec => plan
    case _: CommandResultExec => plan
    case _: V2CommandExec => plan
    case c: DataWritingCommandExec => c.copy(child = apply(c.child))
    case p => apply(p)
```

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
```diff
@@ -0,0 +1,73 @@
...
+            case Origin(None, None) => GlobalLimit(forcedMaxOutputRows, project)
+            case Origin(Some(_), Some(0)) => GlobalLimit(forcedMaxOutputRows, project)
```
ulysses-you, line 64: why we need match `Origin` if the plan is resolved ?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala
ulysses-you: we can add limit if:
-  plan.maxRows is none
- plan.maxRows.get > maxOutputRows

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
```diff
@@ -0,0 +1,73 @@
...
+* */
+case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    if (!plan.resolved) {
+      plan
+    } else {
+      conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
+        case Some(forcedMaxOutputRows) => plan match {
+          case insert@InsertIntoDataSourceDirCommand(_, _, Project(_, _), _) =>
+            insert.copy(query = GlobalLimit(forcedMaxOutputRows, insert.query))
```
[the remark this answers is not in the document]
i7xh: > why we need handle `InsertIntoDataSourceDirCommand` specially ? Seems this PR is only for query and not include the writing command (e.g. `InsertIntoHiveTable` etc)
In our scenarios, ad-hoc query result save as  hdfs files with `INSERT OVERWRITE DIRECTORY`(case 1) statement，`SELECT ...` or `WITH CTE ...` will lead to save as redis，but it's conscious InsertIntoHiveTable for user to produce the data as they want

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
i7xh: > why we need match `Origin` if the plan is resolved ?

1. It can actually remove `case Origin(None, None) => GlobalLimit(forcedMaxOutputRows, project)` before plan is resolved
2. Match `Origin` mainly applied to `SELECT ...` or `WITH CTE(...) SELECT...`, output nodes such as `Project` always starts  0 position on Analyzer，so it judge the output nodes depend on `Origin(_, Some(0))`

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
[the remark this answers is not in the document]
ulysses-you: It's quite customized to match `InsertIntoDataSourceDirCommand, let's match query only.

And for your use case, we can apply this limit rule first then you can apply some other cutom rule (e.g. add `InsertIntoDataSourceDirCommand`).

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
ulysses-you: what's difference if we only match `Project` ? do you mean we will get a wrong node to add limit ?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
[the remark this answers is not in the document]
i7xh: > It's quite customized to match `InsertIntoDataSourceDirCommand, let's match query only.
> 
> And for your use case, we can apply this limit rule first then you can apply some other cutom rule (e.g. add `InsertIntoDataSourceDirCommand`).

OK，So I remove `InsertIntoDataSourceDirCommand` node process from the code，reserve pure query only

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
i7xh: SQL trees may contain other node of same type, especially there may have `Project` or `Aggregate`  in sub-query
It's critical to find the root output node such as `Project` or `Aggregate`, so root `Project match Origin(Some(_), Some(0)` and `Aggregate match Origin(None, None)`

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
i7xh: > we can match more nodes e.g. (Filter, Aggregate, Join ...), how about:
> 
> ```
>     case _: ExecutedCommandExec => plan
>     case _: CommandResultExec => plan
>     case _: V2CommandExec => plan
>     case c: DataWritingCommandExec => c.copy(child = apply(c.child))
>     case p => apply(p)
> ```

Generally root output node is `Project` or `Aggregate` in Analyzer，tree node level or order will be changed in Optimizer

i7xh:
I have already support `Aggregate` process

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala
i7xh: About this issue

1. process `plan.maxRows is none` that node without limit statement
2. `plan.maxRows.get > maxOutputRows` illustrate exists limit statement in query, so don't cover user choice, or else user query result exceed expectation as a bug, so skip the situation

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
ulysses-you: Seems a little hack, the `Origin` is used to distinguish if the plan is subquery ?

i7xh: indeed, this implement is obscure, Shall we have other solution to find tree output root node, I want to know better way to instead this obscure match

ulysses-you: is there any issue if we also add limit for suquery ? I think we don't need care subquery as following:

- scalar subquery; it return at most one row or exception
- in subquery; it's ok to add limit so we can protect engine's driver (OOM)
- exist subquery; it will be rewrite to scalar subquery, so it's also safe.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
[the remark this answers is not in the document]
ulysses-you: also cc @turboFei  for the use case:

> In our scenarios, ad-hoc query result save as hdfs files with INSERT OVERWRITE DIRECTORY(case 1) statement，SELECT ... or WITH CTE ... will lead to save as redis，but it's conscious InsertIntoHiveTable for user to produce the data as they want, so it's no need to limit InsertIntoHiveTable output

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -94,4 +94,13 @@ object KyuubiSQLConf {
       .version("1.4.0")
       .intConf
       .createOptional
+
+  val WATCHDOG_FORCED_MAXOUTPUTROWS =
+    buildConf("spark.sql.watchdog.forcedMaxOutputRows")
+      .doc("Add MaxOutputRows rule for output rows limitation " +
+        "to avoid huge output rows of non-limit query unexpectedly, " +
+        "it's a optional, it's optional that works with defined")
```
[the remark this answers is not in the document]
i7xh: Could you give me some suggestion?😂

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
i7xh: > Seems a little hack, the `Origin` is used to distinguish if the plan is subquery ?

The `Origin` provides a location for TreeNodes in the context of their origin, it define as below(reference: https://github.com/apache/spark/blob/branch-3.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala#L49
):
```scala 
case class Origin(
  line: Option[Int] = None,
  startPosition: Option[Int] = None)
The chief output node as root with init position(None or Some(0)), it's not used to distinguish subquery, we can use subquery class to identify



> is there any issue if we also add limit for suquery ? I think we don't need care subquery as following:
> 
> * scalar subquery; it return at most one row or exception
> * in subquery; it's ok to add limit so we can protect engine's driver (OOM)
> * exist subquery; it will be rewrite to scalar subquery, so it's also safe.

Add limit for subquery may make difference with origin query，my implement will not limit node to subquery
Do you tend to add subquery with above three reasons or not？

Comment (i7xh) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala:61]: > is there any issue if we also add limit for suquery ? I think we don't need care subquery as following:
> 
> * scalar subquery; it return at most one row or exception
> * in subquery; it's ok to add limit so we can protect engine's driver (OOM)
> * exist subquery; it will be rewrite to scalar subquery, so it's also safe.

Add limit for subquery may make difference with origin query，my implement will not limit node to subquery
Do you tend to add subquery with above three reasons or not？

File: dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala (hunk 6 above)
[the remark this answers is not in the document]
Comment (i7xh) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala]: > Words here seem to need a little improvement.

Fixed：https://github.com/apache/incubator-kyuubi/pull/1085/commits/c1a88e54cdb83f77dde8b820a4d3ee6688e86cff

File: dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala (hunk 4 above)
Comment (ulysses-you) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala:61]: > The Origin provides a location for TreeNodes in the context of their origin
> it's not used to distinguish subquery, we can use subquery class to identify

So I'm still not see why you need match origin here, the origin is used to help track error sql string. Is there any use case that Spark match the origin to do the simialr thing ?

Comment (i7xh) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala:61]: > > The Origin provides a location for TreeNodes in the context of their origin
> > it's not used to distinguish subquery, we can use subquery class to identify
> 
> So I'm still not see why you need match origin here, the origin is used to help track error sql string. Is there any use case that Spark match the origin to do the simialr thing ?

Why I use origin to judge the output root node?
According SQL parse principle, it parse with top-down, the output root node is the top node, so I depend on `startPosition` to find root node, It works on sql parse with `antlr` , now the implement mainly use the `startPosition ` in `Token` parsed by antlr, `startPosition` notes that The index of the first character of this token relative to the beginning of the line at which it occurs, 0..n-1

As you said `Origin` is used to track sql string, it almost not match `Origin` in spark

Comment (ulysses-you) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala:61]: does the project is not root node as following ?
```
plan match {
  case _: Project
}
```

can you give a case that we must match origin ?

Comment (i7xh) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala]: I think out this issue, the plan run `match` operator on this rule , it's just the root node, so it doesn't travel matched plan children, thank your heuristic!

File: dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala (hunk 2 above)
Comment (ulysses-you) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala]: Do you mean if user's query limit 100 and the config is 10 then we can't add the limit ? I think we should add limit as long as `plan.maxRows is none` or `plan.maxRows.get > maxOutputRows`. Otherwise it's hard to say which limit is added by us.

Comment (i7xh) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala]: > Do you mean if user's query limit 100 and the config is 10 then we can't add the limit ? 
Yes.
> I think we should add limit as long as `plan.maxRows is none` or `plan.maxRows.get > maxOutputRows`. Otherwise it's hard to say which limit is added by us.
Do you mean using maxOutputRows cover user limit setting ？

Comment (ulysses-you) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala]: yes, otherwise this config has no meaning

Comment (i7xh) [reply dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxOutputRowsRule.scala]: > yes, otherwise this config has no meaning

OK, Final limit according to `spark.sql.watchdog.forcedMaxOutputRows`, commit:  https://github.com/apache/incubator-kyuubi/pull/1085/commits/5b0383dd5d46dbb752ecd1bd37fea5f8fbbb60c8

File: dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala [hunk 7]
```diff
@@ -0,0 +1,82 @@
...
+* }}}
+* The Logical Rule add a GlobalLimit node before root project
+* */
+case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    if (!plan.resolved) {
+      plan
+    } else {
+      conf.getConf(KyuubiSQLConf.WATCHDOG_FORCED_MAXOUTPUTROWS) match {
+        case Some(forcedMaxOutputRows) => plan.maxRows match {
```
ulysses-you, line 58: can we format the code like this:
```
  private def canInsertLimit(p: LogicalPlan): Boolean = {
    val supported = p match {
      case _: Project => true
      case _: Aggregate => true
      case _: Filter => true
      case _: Sort => true
      case _: Union => true
      case _: Join => true
      case _: Window => true
      case _: Generate => true
      case Limit(_, _) => true
      case _: Repartition => true
      case _: RepartitionByExpression => true
      case _ => false
    }
    supported && !p.maxRows.exists(_ <= forcedMaxOutputRows)
  }

  plan match {
    case p if canInsertLimit(p) => Limit(Literal(forcedMaxOutputRows), p)
    case _ => plan
  }
```

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
```diff
@@ -0,0 +1,82 @@
...
+* }}}
+*
+* case 2:
+* {{{
+*   WITH CTE AS (
+*   ...)
+* SELECT [c1, c2, ...] FROM CTE ...
+* }}}
+*
+* case 3 (customize yourself):
```
ulysses-you, line 44: this docs is outdate

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
i7xh: OK, It's concise

dev/kyuubi-extension-spark-3-1/src/test/scala/org/apache/spark/sql/KyuubiExtensionSuite.scala
```diff
@@ -1303,4 +1303,96 @@ class KyuubiExtensionSuite extends KyuubiSparkSQLExtensionTest {
...
+
+      assert(!sql("SELECT count(*) FROM VALUES(1, 'a'),(2, 'b') AS t(c1, c2)")
+        .queryExecution.analyzed.isInstanceOf[GlobalLimit])
+
+      assert(sql(
+        """
+          |SELECT c1, COUNT(*)
+          |FROM VALUES(1, 'a'), (2, 'b') AS t(c1, c2)
+          |GROUP BY c1
+          |""".stripMargin).queryExecution.analyzed.isInstanceOf[GlobalLimit])
```
ulysses-you, line 1328: It's flaky with Spark3.2.0 that the issue of some logical plan has not overridden maxRows is fixed. see https://github.com/apache/spark/pull/30443.

We can use the test table, see `KyuubiExtensionSuite.setupData`.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/watchdog/ForcedMaxOutputRowsRule.scala
```diff
@@ -0,0 +1,71 @@
...
+*   WITH CTE AS (
+*   ...)
+* SELECT [c1, c2, ...] FROM CTE ...
+* }}}
+*
+* The Logical Rule add a GlobalLimit node before root project
+* */
+case class ForcedMaxOutputRowsRule(session: SparkSession) extends Rule[LogicalPlan] {
+
+  private def canInsertLimit(p: LogicalPlan): Boolean = {
```
ulysses-you, line 48: can we put it as parameter `maxOutputRows: Option[Int]` ?

i7xh: > can we put it as parameter `maxOutputRows: Option[Int]` ?
OK, I unified the maxOutputRows quote

dev/kyuubi-extension-spark-3-1/src/test/scala/org/apache/spark/sql/KyuubiExtensionSuite.scala
i7xh: > It's flaky with Spark3.2.0 that the issue of some logical plan has not overridden maxRows is fixed. see [apache/spark#30443](https://github.com/apache/spark/pull/30443).
> 
> We can use the test table, see `KyuubiExtensionSuite.setupData`.

_见贤思齐_
I read the [apache/spark#30443](https://github.com/apache/spark/pull/30443). It seems to get more exactly maxRows in Optimizer? I don't understand how affect the test case changed, as `VALUES(1, 'a'), (2, 'b') AS t(c1, c2)`

i7xh: I have replaced it with setupData

i7xh: > > It's flaky with Spark3.2.0 that the issue of some logical plan has not overridden maxRows is fixed. see [apache/spark#30443](https://github.com/apache/spark/pull/30443).
> > We can use the test table, see `KyuubiExtensionSuite.setupData`.
> 
> _见贤思齐_
> I read the [apache/spark#30443](https://github.com/apache/spark/pull/30443). It seems to get more exactly maxRows in Optimizer? I don't understand how affect the test case changed yet, as `VALUES(1, 'a'), (2, 'b') AS t(c1, c2)`?

ulysses-you: This query used the plan `LocalRelation` and after 3.2.0 the `LocalRelation`'s maxRows will be the data size. That said, if you have a query `SELECT * FROM VALUES(1, 'a'), (2, 'b') AS t(c1, c2)`, the root Project node maxRows will be `Some(2)` instead of `None`. Then the test will fail if we switch the Spark version from 3.1 to 3.2.

ulysses-you:
thanks, merged to master

dev/kyuubi-extension-spark-3-1/src/test/scala/org/apache/spark/sql/KyuubiExtensionSuite.scala
i7xh: > This query used the plan `LocalRelation` and after 3.2.0 the `LocalRelation`'s maxRows will be the data size. That said, if you have a query `SELECT * FROM VALUES(1, 'a'), (2, 'b') AS t(c1, c2)`, the root Project node maxRows will be `Some(2)` instead of `None`. Then the test will fail if we switch the Spark version from 3.1 to 3.2.
Clear! Thanks for your advice!

Issue #937, opened 2021-08-17
Push Kyuubi docker image


ulysses-you:
### 1. Describe the feature
Push Kyuubi docker image to dockerhub

Some test tailed due to the image is outdate, we need a official docker image repo.

https://hub.docker.com/repository/docker/apache/kyuubi/general

yaooqinn:
LGTM

ulysses-you:
created a ticket https://issues.apache.org/jira/browse/INFRA-22212

ulysses-you:
Now the docker repo is ready, but we need [secrets](https://docs.github.com/en/actions/guides/publishing-docker-images) for github action.

Created a new ticket https://issues.apache.org/jira/browse/INFRA-22341

PR #1117, opened 2021-09-17
[KYUUBI #937] Push Kyuubi docker image

ulysses-you:

Push docker image if we commit to master

test in forked repo.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,0 +1,28 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  push_to_registry:
+    name: Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    concurrency:
+      # this group should be global unique
+      group: push-docker-image
+      cancel-in-progress: true
+    steps:
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v2
+        with:
+          file: docker/Dockerfile
+          push: true
+          tags: apache/kyuubi:master-snapshot


timothy65535:
Nice Idea.

.github/workflows/image.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  push_to_registry:
+    name: Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
timothy65535, line 20: `secrets.DOCKERHUB_TOKEN` to `secrets.DOCKERHUB_PASSWORD`

https://docs.github.com/en/actions/reference/encrypted-secrets

.github/workflows/image.yml
```diff
@@ -0,0 +1,37 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
kezhenxu94, line 7: Pushing docker images for every pull request is not a good practice, it’s better to push the images only when the commit is pushed to master branch.

yaooqinn: +1

ulysses-you: agree, the workflow is only for test now

ulysses-you:
The workflow it now works in my forked repo and docker hub. We can see how the github action works after merged this PR.

ulysses-you:
let me it merge to master first and see what will happen

kezhenxu94:
> let me merge it to master first and see what will happen

You don't need to merge it into master just for testing. You can create a new branch on this repo (not your fork repo) and make the workflow run in that repo.

Also, the docker repo must be the same as the GitHub repo name `apache/incubator-kyuubi`, not `apache/kyuubi`

ulysses-you:
thank you @kezhenxu94 , seems the GA failed with 
```
2021-09-23T02:15:35.6185062Z error: denied: requested access to the resource is denied
2021-09-23T02:15:36.2512609Z ##[error]buildx failed with: error: denied: requested access to the resource is denied
```

Not sure why it happened, I will revert the commit and use new branch to test

Issue #1039, opened 2021-09-06
Kyuubi can't cancle the app that the running queue has no resource until user kill manually


zhang1002:
### 1. Describe the bug
If the queue that you submmit the app through kyuubi that has no resource, the app will always stay in a state of ACCEPTED until you kill manually.

The yarnApplicationState: ACCEPTED: waiting for AM container to be allocated, launched and register with RM.

### 2. Environments
yarn: 2.9.2
kyuubi: 1.1.0

#### 2.1 Versions

- Kyuubi
  - [ ] 1.1.0
  - [ ] 1.0.3
  - [ ] 0.8.x and earlier 

- Spark
  - [ ] 3.1.x
  - [ ] 3.0.x
  - [ ] 2.4.x and earlier 

#### 4.2 Log of the corresponding engine

akizminet:
So if ACCEPTED stages take longer than `kyuubi.session.engine.initialize.timeout` and kyuubi doesn't kill those apps. They will take resources when the queue is free and live for a long time on YARN cluster. I think this is a bug.

yaooqinn:
@zhang1002 Are you still working on this?

simon824:
If no one else following can assign it to me.

yaooqinn:
@simon824 are you willing to work on this？

simon824:
> @simon824 are you willing to work on this？

yep

PR #1119, opened 2021-09-17
[KYUUBI #1039] kill yarn job when engine initialize timeout and yarnApplicationState is ACCEPTED

simon824:
kill yarn job when engine initialize timeout and yarnApplicationState is ACCEPTED
#1039


kyuubi-server/src/test/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilderSuite.scala
```diff
@@ -225,6 +226,17 @@ class SparkProcessBuilderSuite extends KerberizedTestHelper {
     val b2 = new SparkProcessBuilder("test", conf)
     assert(b2.mainResource.getOrElse("") != jarPath.toString)
   }
+
+  test("test kill application") {
+    val config = KyuubiConf()
+    val pb = new FakeSparkProcessBuilder(config)
+    val regex = Pattern.compile("application_\\d+_\\d+")
```
pan3793, line 233: Use `APPLICATION_REGEX`

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +150,38 @@ trait ProcBuilder {
     proc
   }
 
+  def execKillYarnJob(appId: String): Int = {
+    val hadoopHomeOpt = env.get("HADOOP_HOME").orElse {
+      throw KyuubiSQLException(s"Failed to kill yarn job. HADOOP_HOME is not set! " +
```
pan3793, line 155: Should we also find `/usr/bin/yarn`?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +150,38 @@ trait ProcBuilder {
...
+        "For more detail information on installing and configuring Yarn, please visit " +
+        "https://kyuubi.apache.org/docs/stable/deployment/on_yarn.html")
+    }.map { dir =>
```
pan3793, line 158: `dir => hadoopHome`

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +150,38 @@ trait ProcBuilder {
...
+      exitValue = Some(process.exitValue())
+    } catch {
+      case e: Exception =>
+        throw KyuubiSQLException(s"Failed to kill yarn job. Command[$command]", e.getCause)
+    }
+    info(s"Successfully kill yarn job $appId")
+    exitValue.get
+  }
+
+  private val APPLICATION_REGEX = Pattern.compile("application_\\d+_\\d+")
```
pan3793, line 176: `APPLICATION_REGEX` => `YARN_APP_NAME_REGEX`, prefer to use Scala regex, https://docs.scala-lang.org/tour/regular-expression-patterns.html

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
simon824: I think get the absolute path is enough, how do you concider it ?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +149,38 @@ trait ProcBuilder {
...
+      throw KyuubiSQLException(s"Failed to kill yarn job. HADOOP_HOME is not set! " +
+        "For more detail information on installing and configuring Yarn, please visit " +
+        "https://kyuubi.apache.org/docs/stable/deployment/on_yarn.html")
+    }.map { hadoopHome =>
+      Paths.get(hadoopHome, "bin", "yarn").toAbsolutePath.toFile.getCanonicalPath
+    }.get
+    val command = hadoopHomeOpt + " application -kill " + appId
+    val pb = new ProcessBuilder("/bin/sh", "-c", command)
+    var exitValue: Option[Int] = None
+
```
zhang1002, line 163: we have hadoop-client dependency, keyTab and principle. why we don't use yarn-client to kill this application?

simon824: Thank you for your suggestion, this is indeed a better solution. I have not implemented it in this way before, I will refine it later or next pr.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +149,38 @@ trait ProcBuilder {
...
 
+  def execKillYarnJob(appId: String): Int = {
+    val hadoopHomeOpt = env.get("HADOOP_HOME").orElse {
+      throw KyuubiSQLException(s"Failed to kill yarn job. HADOOP_HOME is not set! " +
+        "For more detail information on installing and configuring Yarn, please visit " +
+        "https://kyuubi.apache.org/docs/stable/deployment/on_yarn.html")
+    }.map { hadoopHome =>
+      Paths.get(hadoopHome, "bin", "yarn").toAbsolutePath.toFile.getCanonicalPath
+    }.get
+    val command = hadoopHomeOpt + " application -kill " + appId
```
timothy65535, line 160: we can use yarn rest api to kill

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
pan3793: Yeah, I'm fun with that.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -182,6 +182,7 @@ private[kyuubi] class EngineRef(
           }
         }
         if (started + timeout <= System.currentTimeMillis()) {
+          builder.closeYarnJob()
```
zhouyifan279, line 185: Maybe `killApplication` is a more suitable as engine may be submitted to other Cluster Managers such as kubernetes.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +149,38 @@ trait ProcBuilder {
...
+
+  val YARN_APP_NAME_REGEX: Regex = "application_\\d+_\\d+".r
+
+  def killApplication(row: String = lastRowOfLog): Unit = {
+    val matcher = YARN_APP_NAME_REGEX.findAllMatchIn(row)
+
+    if (matcher.nonEmpty && row.contains("state: ACCEPTED")) {
+      execKillYarnJob(matcher.next().group(0))
+    }
+  }
```
pan3793, line 182: change to 
```
  def closeYarnJob(line: String = lastRowOfLog): Unit = {
    YARN_APP_NAME_REGEX.findFirstIn(line) match {
      case Some(appName) if line.contains("state: ACCEPTED") =>
        execKillYarnJob(appName)
      case None =>
    }
  }
```

pan3793: Please also rename variable `lastRowOfLog` to `lastLineOfLog`

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilderSuite.scala
```diff
@@ -225,6 +224,21 @@ class SparkProcessBuilderSuite extends KerberizedTestHelper {
...
+  test("test kill application") {
+    val config = KyuubiConf()
+    val pb: FakeSparkProcessBuilder = new FakeSparkProcessBuilder(config)
+    val lastRowOfLog = "Application report for application_1593587619692_19713 (state: ACCEPTED)"
+
+    val e = intercept[KyuubiSQLException] {
+      pb.killApplication(lastRowOfLog)
+    }
+
+    assert(e.getMessage === "Failed to kill yarn job. " +
```
pan3793, line 237: It's not a robust test. The test will fail if the user set `HADOOP_HOME` in their machine.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
pan3793: This way makes more sense to me, because it's doesn't require setting up hadoop client in Kyuubi machine, and specific versions of hadoop client jars.

pan3793:
So far, there are 3 approaches to kill yarn applications

1. Execute command `yarn kill $app_name`
2. Use Yarn RestFul API
3. Use Yarn client

Personally, I prefer option 2, then option 1.

Option 1 requires the user to set up Hadoop client on the Kyuubi server node(s), and also set `HADOOP_HOME`, suppose that the user usually should deploy Kyuubi on the Hadoop node, it's acceptabe.

If we choose option 3, we need to test with different versions of Yarn cluster to ensure that the compatibility between Kyuubi server bundled Hadoop client jars and Yarn cluster.

Seems that option 2 has less dependencies and good compatibility.

@timothy65535 @simon824 @zhouyifan279 @yaooqinn WDYT?

zhang1002:
> Use Yarn RestFul API

we often use option-2 to get some data from hadoop...
this method i think it's better. we only need core-site.xml and some files about kerberos.
this method is independent of hadoop_version.

zhouyifan279:
YARN REST API supports killing apps since 2.5.0. That should be enough to cover most YARN versions used in production.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +149,39 @@ trait ProcBuilder {
...
+      Paths.get(hadoopHome, "bin", "yarn").toAbsolutePath.toFile.getCanonicalPath
+    }.get
+    hadoopHomeOpt + " application -kill " + appId
+  }
+
+  val YARN_APP_NAME_REGEX: Regex = "application_\\d+_\\d+".r
+
+  def killApplication(line: String = lastRowOfLog): Unit =
+    YARN_APP_NAME_REGEX.findFirstIn(line) match {
+      case Some(appId) if line.contains("state: ACCEPTED") => execCommand(getCommand(appId))
```
zhouyifan279, line 181: It is possible, although rare, that the last row of log reports state to be NEW/NEW_SAVING/SUBMITTED when engine is submitted in YARN Cluster mode and `spark.yarn.submit.waitAppCompletion` is `false`.

timothy65535:
In fact, the yarn client already support kill the application by `LauncherProtocol` mechanism. We will not kill application by shell script or yarn api.

**REFER TO**:
https://github.com/apache/spark/blob/9d8ac7c8e90c1c2a6060e8f1e8f2f21e19622567/resource-managers/yarn/src/main/scala/org/apache/spark/deploy/yarn/Client.scala#L137-L145

cc @simon824 @pan3793 @yaooqinn

yaooqinn:
The issue seems to be caused by the improper usage of `process.destory` which will not invoke the shutdown hooks of the spark/engine app.

Instead of calling `process.destory`, it looks like we'd better notify the engine to shutdown itself

timothy65535:
> it looks like we'd better notify the engine to shutdown itself

When app in `ACCEPTED` status, the engine has not initialized, so it can not shutdown itself. `LauncherProtocol` can work.

yaooqinn:
> > it looks like we'd better notify the engine to shutdown itself
> 
> When app in `ACCEPTED` status, the engine has not initialized, so it can not shutdown itself. `LauncherProtocol` can work.

I don't know what you mean

timothy65535:
hi @yaooqinn, when in ACCEPTED state, means that the spark driver hasn't been inited. We can only use yarn api to kill the app.

yaooqinn:
I know the final step is using yarnClient to kill the application if the cluster manager is yarn in this case. But do we have any higher-level API or actions that we can take to make this happen?  e.g. `by calling SparkContext.stop`

we are not a yarn-specific platform, right?

timothy65535:
> I know the final step is using yarnClient to kill the application if the cluster manager is yarn in this case. But do we have any higher-level API or actions that we can take to make this happen? e.g. `by calling SparkContext.stop`
> 
> we are not a yarn-specific platform, right?

Right.

timothy65535:
The spark internal launcher already expose the interface to comminicate with resource manager client.

- YARN: implemented.
- Meson: implemented.
- K8s: seems has not implemented.

https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/launcher/LauncherBackend.scala

simon824:
> The spark internal launcher already expose the interface to comminicate with resource manager client.
> 
> * YARN: implemented.
> * Meson: implemented.
> * K8s: seems has not implemented.
> 
> [_**spark core**_]
> https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/launcher/LauncherBackend.scala
> 
> ```
>   def connect(): Unit = {
>     val port = conf.getOption(LauncherProtocol.CONF_LAUNCHER_PORT)
>       .orElse(sys.env.get(LauncherProtocol.ENV_LAUNCHER_PORT))
>       .map(_.toInt)
>     val secret = conf.getOption(LauncherProtocol.CONF_LAUNCHER_SECRET)
>       .orElse(sys.env.get(LauncherProtocol.ENV_LAUNCHER_SECRET))
>     if (port.isDefined && secret.isDefined) {
>       val s = new Socket(InetAddress.getLoopbackAddress(), port.get)
>       connection = new BackendConnection(s)
>       connection.send(new Hello(secret.get, SPARK_VERSION))
>       clientThread = LauncherBackend.threadFactory.newThread(connection)
>       clientThread.start()
>       _isConnected = true
>     }
>   }
> ```

Is it common use in different hadoop versions？
Mesos and k8s also support restful api.

timothy65535:
> Is it common use in different hadoop versions？

Right. 

> Restful API

If use restful api, and contains some auth properties(e.g kerberos), we can not call the restful api directly.

simon824:
> So far, there are 3 approaches to kill yarn applications
> 
> 1. Execute command `yarn kill $app_name`
> 2. Use Yarn RestFul API
> 3. Use Yarn client
> 
> Personally, I prefer option 2, then option 1.
> 
> Option 1 requires the user to set up Hadoop client on the Kyuubi server node(s), and also set `HADOOP_HOME`, suppose that the user usually should deploy Kyuubi on the Hadoop node, it's acceptabe.
> 
> If we choose option 3, we need to test with different versions of Yarn cluster to ensure that the compatibility between Kyuubi server bundled Hadoop client jars and Yarn cluster.
> 
> Seems that option 2 has less dependencies and good compatibility.
> 
> @timothy65535 @simon824 @zhouyifan279 @yaooqinn WDYT?

Now we have a new option: 
4. Use SparkLauncher

It requires initializing SparkLauncher Object instead of building  spark-submit command ,  then  get appId and killApp by SparkAppHandle. 
And it is independent of hadoop_version.

@zhang1002 @zhouyifan279 @pan3793 WDYT?

timothy65535:
+1, use SparkLauncher may be the best way to solve this problem, it contains `LauncherProtocol`

simon824:
> +1, use SparkLauncher may be the best way to solve this problem, it contains `LauncherProtocol`

Have you achieved it by SparkLauncher before? I just did a test and found that if job state is RUNNING, SparkAppHandle can stop/kill application. But if SparkAppHandle.State is SUBMITTED  (yarn state is ACCEPTED) , it was failed.

timothy65535:
if sparklaunch can not stop it, yarn rest api will also can not kill it. Need to check it why it can not kill the app.

timothy65535:
关于及时停止超时应用:

层级一: 应用一直处于ACCEPT，但是提交程序已经超时
- 可以通过REST API或者SHELL方式停止，这两种方式存在兼容性或者是认证方面的问题，尤其是使用了kerberos认证，这种调用一定失败
- 通过上层进程与子进程交互来停止，也就是一直提到的`LauncherProtocol`，这个接口已经被用在Spark Core模块里面了，每个资源框架各自实现如何通过api方式来停止任务。比如yarn，它就是利用yarn clint去停止任务。不管是第一点提到的SHELL还是REST，底层肯定是yarn client来停止。所以说如果这种方式都停不了，那就是yarn本身就停不掉

层级二: 每次提交spark任务都会提前创建好zk上的space，如果层级一没有做到，在engine启动的时候可以检测一下space是否存在，不存在及时退出即可

层级三: 每个engine自己会有默认30分钟的检测时间，即没有session，engine也会自己退出

关于层级一，yarn client api是可以做到停掉处于accwpt的任务。如果spark launch没有做到，可以debug下。

kyuubi-server/pom.xml
```diff
@@ -231,6 +231,11 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-core_${scala.binary.version}</artifactId>
```
pan3793, line 236: Do we need `spark-core`? Is `spark-launcher` sufficient?

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilderSuite.scala
```diff
@@ -45,8 +45,8 @@ class SparkProcessBuilderSuite extends KerberizedTestHelper {
     assert(Files.exists(Paths.get(commands.last)))
 
     val process = builder.start
-    assert(process.isAlive)
-    process.destroyForcibly()
+    assert(SparkAppHandle.State.values().contains(process.getState))
```
timothy65535, line 48: seems don't need this `assert`

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -49,18 +48,9 @@ trait ProcBuilder {
 
   protected def env: Map[String, String] = conf.getEnvs
 
-  protected val workingDir: Path
-
-  final lazy val processBuilder: ProcessBuilder = {
-    val pb = new ProcessBuilder(commands: _*)
+  protected def launcher: SparkLauncher
```
yaooqinn, line 51: can SparkLauncher be able to launch the coming Flink SQL eninge?

kyuubi-server/pom.xml
```diff
@@ -55,6 +55,11 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
```
yaooqinn, line 58: When I create the architecture of Kyuubi, I was trying to avoid using dependencies from engines(spark/flink etc..)

timothy65535: 追求极致，通过相同的框架来做到提交FLINK和SPARK是挺好的，不过FLINK中不存在spark launch模块。FLINK内置的提交程序没有做与SPARK一样的`LauncherProtocol`。换言之，SPARK在这方面更好，预留接口来对接各个资源框架（YARN、K8s）等。

其次，针对这个ISSUE的案例来说，生产经常遇到资源不足的情况。如果开启动态资源，初始资源可以设置小一点，基本也难遇到ACCEPT的情况。而且生产一般应该会做离线集群和实时集群分开，再结合队列预留资源，这种情况几率也不会太大。

所以，这个看各位同仁意见吧，比较难做到同一套程序既满足SPARK和FLINK任务的提交，也满足适配不同的资源框架，同时支持及时停止任务的提交。

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
timothy65535: Probably not.

kyuubi-server/pom.xml
timothy65535: > 层级二: 每次提交spark任务都会提前创建好zk上的space，如果层级一没有做到，在engine启动的时候可以检测一下space是否存在，不存在及时退出即可

这个我们可以尝试

simon824: > > 层级二: 每次提交spark任务都会提前创建好zk上的space，如果层级一没有做到，在engine启动的时候可以检测一下space是否存在，不存在及时退出即可
> 
> 这个可以尝试~

Server module will not create the `namespace` on zk in advance, it only passed the `namespace`  parameter to the engine . And engine will create the `namespace` on zk when it started.

simon824:
Now we have differences. I think we need to fully discuss the implement of this problem before coding.

1. Execute command yarn kill $app_name
2. Use Yarn RestFul API
3. Use Yarn client
4. Use SparkLauncher

@pan3793 @zhouyifan279 @zhang1002 @yaooqinn @timothy65535 @turboFei @ulysses-you
What do you think?

yaooqinn:
> Now we have differences. I think we need to fully discuss the implement of this problem before coding.
> 
> 1. Execute command yarn kill $app_name
> 2. Use Yarn RestFul API
> 3. Use Yarn client
> 4. Use SparkLauncher
> 
> @pan3793 @zhouyifan279 @zhang1002 @yaooqinn @timothy65535 @turboFei @ulysses-you
> What do you think?

Or we provide a develop-oriented post executed script if timeout. we pass some application details, such as user/applicaiontId to it and the script can contain something like `yarn kill ...` or ```spark-submit --kill ...```

zhang1002:
ok.
at first, we only consider app running in yarn. This is problematic because we also can run them in k8s or others...

kyuubi-server/pom.xml
timothy65535: yeah, know it.

pan3793: I think it's not a big deal, `spark-launcher` has only a few deps.

simon824:
> > Now we have differences. I think we need to fully discuss the implement of this problem before coding.
> > 
> > 1. Execute command yarn kill $app_name
> > 2. Use Yarn RestFul API
> > 3. Use Yarn client
> > 4. Use SparkLauncher
> > 
> > @pan3793 @zhouyifan279 @zhang1002 @yaooqinn @timothy65535 @turboFei @ulysses-you
> > What do you think?
> 
> Or we provide a develop-oriented post executed script if timeout. we pass some application details, such as user/applicaiontId to it and the script can contain something like `yarn kill ...` or `spark-submit --kill ...`

It means that kyuubi doesn't actually execute the script, just generates it?

simon824:
Maybe we considered too much, we don’t have to find a unified method to stop application in different resource managers and different compute engines, we can implement stop() in different child class and then make some conditional judgments.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -127,7 +127,7 @@ class SparkProcessBuilder(
       buffer += CONF
       buffer += s"$k=$v"
     }
-    // iff the keytab is specified, PROXY_USER is not supported
+    // if the keytab is specified, PROXY_USER is not supported
```
ulysses-you, line 130: the old comment is right, iff means only if

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +149,23 @@ trait ProcBuilder {
     proc
   }
 
+  val YARN_APP_NAME_REGEX: Regex = "application_\\d+_\\d+".r
+
+  def killApplication(line: String = lastRowOfLog): Int =
```
ulysses-you, line 154: can we return msg here directly and move following code into this method ?
```
 val killMessage = builder.killApplication() match {
            case id if id != 0 => "And failed to kill Application, please kill it manually. "
            case _ => "And killed Application successfully. "
          }
```

simon824: Done

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
simon824: Get it

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -148,6 +149,26 @@ trait ProcBuilder {
...
+          System.getProperty("user.dir") + "/bin/stop-application.sh", appId)
+
+        pb.environment()
+          .putAll(env.asJava)
+        pb.redirectError(engineLog)
+        pb.redirectOutput(engineLog)
+        val process = pb.start()
+        process.waitFor() match {
+          case id if id != 0 => "Failed to kill Application, please kill it manually. "
+          case _ => "Killed Application successfully. "
```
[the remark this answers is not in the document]
simon824: Done

ulysses-you (approved):

bin/stop-application.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "USAGE: $0 <application_id>"
+  exit 1
+fi
+
+if [[ -z ${HADOOP_HOME} ]]; then
+  echo "Error: HADOOP_HOME IS NOT SET! CANNOT PROCEED."
+  exit 1
+fi
+
+$HADOOP_HOME/bin/yarn application -kill $1
```
timothy65535, line 29: 如果是K8s，也就需要再定义一个相同的脚本？

Issue #950, opened 2021-08-18
[Umbrella][KPIP-1] Support RESTful API for Kyuubi


yanghua:
# Describe the proposal

See the proposal document: #917 

# Task list

- [x] #952 @yanghua 
- [x] #979 @yanghua 
- [x] #983 @yanghua 
- [x] #999 @yanghua 
- [x] #1011 @yanghua 
- [x] #1031 @yanghua 
- [x] #1052 @yanghua 
- [x] #1088 @yanghua 
- [x] #1092 @yanghua 
- [x] #1102 @yanghua  
- [x] #1106  @simon824  
- [x] #1135 @cfmcgrady 
- [x] #1258 @simon824  
- [x] #1275 @simon824 
- [x] #1294 @simon824  
- [x] #1351 @simon824 
- [ ] #1444 @simon824  
- [x] #1612
- [x] #1611
- [x] #1614
- [x] #1620 @simon824 
- [x] #1575 @simon824 
- [x] #1631
- [x] #1639
- [ ] #1672 @yaooqinn

mahengyang:
这个功能好

turboFei:
seems all of them are completed.

should we mark it as finished? @yanghua

simon824:
> seems all of them are completed.
> 
> should we mark it as finished? @yanghua

It's not completed yet, still having many things to do.

yaooqinn:
ping @yanghua @simon824, branch-1.5's been cut. Does anything more need to be fixed in this ticket? Otherwise, we shall mark this work as completed.

BTW, I notice that we have nothing in doc about this feature, right? It's not good for our users to use this feature.

simon824:
> ping @yanghua @simon824, branch-1.5's been cut. Does anything more need to be fixed in this ticket? Otherwise, we shall mark this work as completed.
> 
> BTW, I notice that we have nothing in doc about this feature, right? It's not good for our users to use this feature.

Yes, it can be marked as completed, if there is any bug in the future, can ping me to fix , and I will add the document later.

yaooqinn:
@simon824 thanks for the confirmation. actually, they're a lot of APIs through Rest, we can summarize and improve them in the future.

PR #1135, opened 2021-09-22
[SUB-TASK][KPIP-1] Register default ObjectMapper instead of manually serialize/deserialize JSON input argument.

cfmcgrady:

Register default `ObjectMapper` instead of manually serialize/deserialize JSON input argument.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


cfmcgrady:
cc @yanghua @yaooqinn

yanghua:
@cfmcgrady Thanks, LGTM. BTW, can you refactor it thoroughly? IMO, after changing the return value of the request method, we may also remove the code like `OBJECT_MAPPER.readValue` . WDYT?

cfmcgrady:
> @cfmcgrady Thanks, LGTM. BTW, can you refactor it thoroughly? IMO, after changing the return value of the request method, we may also remove the code like `OBJECT_MAPPER.readValue` . WDYT?

SGTM, updated.

ulysses-you:
@cfmcgrady the change LGTM, but I think the change is not minor. Can we also collect the PR to https://github.com/apache/incubator-kyuubi/issues/950 ?

cfmcgrady:
> @cfmcgrady the change LGTM, but I think the change is not minor. Can we also collect the PR to #950 ?

It seems to I don't have edit issue permission, need @yanghua's help to add to the task list.

ulysses-you (approved):

ulysses-you:
thanks, merging to master

yanghua:
> > @cfmcgrady the change LGTM, but I think the change is not minor. Can we also collect the PR to #950 ?
> 
> It seems to I don't have edit issue permission, need @yanghua's help to add to the task list.

sure.

cfmcgrady:
Thanks for the review.

Issue #1067, opened 2021-09-09
[Bug] Exception is not matching the real issue


zhang1002:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The exception is not matching the real issue.
When you open the configuration item of : kyuubi.engine.event.json.log.path that you don't create this path, the server will stop and throw exception.
The exception is : 
rg.apache.kyuubi.KyuubiSQLException: Error opening session SessionHandle [ce33ab32-7ded-409a-89a5-cdb0dafd4d40] for hive due to Exception in thread "main" java.lang.IllegalStateException: Illegal Service State: LATENT

It does not match the real issue.
The real exception should : java.io.FileNotFoundException

### Kyuubi Server Log Output

```logtalk
2021-09-08 14:51:50.519 INFO zookeeper.ClientCnxn: EventThread shut down for session: 0x78fe21e68f0325
2021-09-08 14:51:50.524 WARN service.ThriftFrontendService: Error opening session: 
org.apache.kyuubi.KyuubiSQLException: Error opening session SessionHandle [ce33ab32-7ded-409a-89a5-cdb0dafd4d40] for hive due to Exception in thread "main" java.lang.IllegalStateException: Illegal Service State: LATENT
	at org.apache.kyuubi.service.ThriftFrontendService.connectionUrl(ThriftFrontendService.scala:110)
	at org.apache.kyuubi.service.Serverable.connectionUrl(Serverable.scala:34)
	at org.apache.kyuubi.engine.spark.events.EngineEvent$.apply(EngineEvent.scala:107)
	at org.apache.kyuubi.engine.spark.SparkSQLEngine.engineStatus$lzycompute(SparkSQLEngine.scala:41)
	at org.apache.kyuubi.engine.spark.SparkSQLEngine.engineStatus(SparkSQLEngine.scala:41)
	at org.apache.kyuubi.engine.spark.SparkSQLEngine$.$anonfun$main$1(SparkSQLEngine.scala:152)
	at org.apache.kyuubi.engine.spark.SparkSQLEngine$.$anonfun$main$1$adapted(SparkSQLEngine.scala:150)
	at scala.Option.foreach(Option.scala:407)
[...]

PR #1136, opened 2021-09-22
[KYUUBI #1067] Catch the real exception when the log directory is not exist

zhang1002:
BUGFIX:
#1067 

fix:
2021-09-23 16:03:43.248 INFO HiveMetaStore.audit: ugi=zhangyu	ip=unknown-ip-addr	cmd=get_databases: *	
2021-09-23 16:03:43.319 ERROR events.JsonEventLogger: Get log directory /Users/zhangyuxiang/Downloads/kyuubi_test failed
java.io.FileNotFoundException: File /Users/zhangyuxiang/Downloads/kyuubi_test does not exist
	at org.apache.hadoop.fs.RawLocalFileSystem.deprecatedGetFileStatus(RawLocalFileSystem.java:666)
	at org.apache.hadoop.fs.RawLocalFileSystem.getFileLinkStatusInternal(RawLocalFileSystem.java:987)
	at org.apache.hadoop.fs.RawLocalFileSystem.getFileStatus(RawLocalFileSystem.java:656)
	at org.apache.hadoop.fs.FilterFileSystem.getFileStatus(FilterFileSystem.java:454)
	at org.apache.kyuubi.events.JsonEventLogger.requireLogRootWritable(JsonEventLogger.scala:79)
	at org.apache.kyuubi.events.JsonEventLogger.initialize(JsonEventLogger.scala:93)
	at org.apache.kyuubi.service.CompositeService.$anonfun$initialize$1(CompositeService.scala:40)
	at org.apache.kyuubi.service.CompositeService.$anonfun$initialize$1$adapted(CompositeService.scala:40)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)


zhang1002:
cc @pan3793 @zhouyifan279

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/events/EventLoggingWithoutDirectoryServiceSuite.scala
```diff
@@ -0,0 +1,55 @@
...
+  override def startSparkEngine(): Unit = {
+    try {
+      super.startSparkEngine()
+    } catch {
+      case e: IOException =>
+        exception = e
+    }
+  }
+
+  test("get IOException when this file is not exist") {
```
pan3793, line 51: => `"throw IOException if event log directory does not exist"`

pan3793:
Could you please also add a case for the following code?

```
if (!fileStatus.isDirectory) {
  throw new IllegalArgumentException(s"Log directory $logRoot is not a directory.")
}
```

kyuubi-common/src/main/scala/org/apache/kyuubi/events/JsonEventLogger.scala
```diff
@@ -75,7 +75,13 @@ class JsonEventLogger[T <: KyuubiEvent](logName: String,
   }
 
   private def requireLogRootWritable(): Unit = {
-    val fileStatus = fs.getFileStatus(new Path(logRoot))
+    val fileStatus = try {
+      fs.getFileStatus(new Path(logRoot))
+    } catch {
+      case e: IOException =>
+        error(s"Get log directory $logRoot failed", e)
```
ulysses-you, line 82: what's the behavior before, do you mean we cannot log the exception ?

zhang1002: Yes, we cannot log this exception.

Before fix this bug, when the dir is not exist, the engineLog only can get: java.lang.IllegalStateException: Illegal Service State: LATENT.
The cause of this exception is:
when we stop the engine before, we need to get this engine's engineStatus(SparkSQLEngine line 158).
this engineStatus will store the connectionUrl from frontendService. But if this serviceState is LATENT, it will throw this exception.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/events/EventLoggingWithoutDirectoryServiceSuite.scala
zhang1002: done

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/events/EventLoggingWithoutDirectoryServiceSuite.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.kyuubi.Utils
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.engine.spark.WithSparkSQLEngine
+import org.apache.kyuubi.events.EventLoggerType._
+import org.apache.kyuubi.operation.JDBCTestUtils
+
+class EventLoggingWithoutDirectoryServiceSuite extends WithSparkSQLEngine with JDBCTestUtils {
+
+  private val logRoot = "file:///tmp/aaa/bbb"
```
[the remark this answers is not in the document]
zhang1002: done

yaooqinn:
thanks for reporting this bug, but the current implementation is not correct.

One good thing about this PR is that we can log and capture the real error in the engine log, but we still need to `suppress` the ambiguous error message reported by the event logger.

zhang1002:
> One good thing about this PR is that we can log and capture the real error in the engine log, but we still need to suppress the ambiguous error message reported by the event logger.

this exception will log in engine log, not event log.

yaooqinn:
> this exception will log in engine log, not event log.

yes, that's what I mean as I said above. You can try to understand what I said in the latter part of my comment

zhang1002:
> One good thing about this PR is that we can log and capture the real error in the engine log, but we still need to suppress the ambiguous error message reported by the event logger.

done.
When the service's state is LATENT, this engine will not create engineEvent.

PR #1037, opened 2021-09-06
Add classification of SQL_Type: for dql and auxiliary statement

zhang1002:
This PR, we have two different of sql classification:
- other:  all auxiliary statement fail into this classification
- dql

We support the user use the self-defined matching rule: sql-classification.json.
If there have no this named jsonFile, the service will upload the default matching rule: sql-classification-default.json.

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.readthedocs.io/en/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #1035, opened 2021-09-06
Add classification of SQL_Type: this PR is just for DML

zhang1002:
Add classification of SQL_Type: this PR is just for DML

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.readthedocs.io/en/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #1140, opened 2021-09-23
[KYUUBI #1002][FOLLOWUP] Refine sql classification rule

ulysses-you:

Follow up [KYUUBI #1002] [KYUUBI #1035] [KYUUBI #1037]

- correct the config key; we should use `spark.` as the prefix
- change the config to false by default
- correct the class loader; thread context is not safe, spark and hive often change it
- update the log and exception

following is the failed error msg:
```
Caused by: java.io.FileNotFoundException: file:/Users/cathy/Desktop/tmp/spark-3.1.2-bin-hadoop2.7/jars/kyuubi-extension-spark-3-1_2.12-1.4.0-SNAPSHOT.jar!/sql-classification-default.json (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at com.fasterxml.jackson.core.JsonFactory.createParser(JsonFactory.java:916)
	at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:2745)
	at org.apache.kyuubi.sql.sqlclassification.KyuubiGetSqlClassification$.<init>(KyuubiGetSqlClassification.scala:51)
	at org.apache.kyuubi.sql.sqlclassification.KyuubiGetSqlClassification$.<clinit>(KyuubiGetSqlClassification.scala)
	... 59 more

```


dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
```diff
@@ -35,28 +35,25 @@ import org.apache.kyuubi.sql.KyuubiSQLConf._
...
+        var url: URL = getClass.getClassLoader.getResource("sql-classification.json")
         if (url == null) {
-          url = Thread.currentThread().getContextClassLoader
-            .getResource("sql-classification-default.json")
+          logInfo("sql-classification.json is not found, use default config instead")
+          url = getClass.getClassLoader.getResource("sql-classification-default.json")
         }
-        val defaultSqlClassificationFile = url.getPath
-        Some(objectMapper.readTree(new File(defaultSqlClassificationFile)))
+        Some(objectMapper.readTree(url))
```
cfmcgrady, line 48: 1. Shall we use `Option` instead?
2. Shall we also handle a malformed JSON case? `objectMapper.readTree()` may throws `IOException/JsonProcessingException`.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
```diff
@@ -35,28 +35,25 @@ import org.apache.kyuubi.sql.KyuubiSQLConf._
...
       case false =>
         None
     }
   }
 
   /**
    * Notice:
-   *    You need to make sure that the configuration item: kyuubi.spark.sql.classification.enabled
+   *    You need to make sure that the configuration item: SQL_CLASSIFICATION_ENABLED
     *   is true
```
cfmcgrady, line 57: nit: indent

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
ulysses-you: I think it's ok to throw exception if the error config from user. And by default, we will never get malformed data.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
```diff
@@ -35,28 +35,25 @@ import org.apache.kyuubi.sql.KyuubiSQLConf._
...
-object KyuubiGetSqlClassification {
-
+object KyuubiGetSqlClassification extends Logging {
   private val jsonNode: Option[JsonNode] = {
     SQLConf.get.getConf(SQL_CLASSIFICATION_ENABLED) match {
       case true =>
         val objectMapper = new ObjectMapper
-        var url: URL = Thread.currentThread().getContextClassLoader
-          .getResource("sql-classification.json")
+        var url: URL = getClass.getClassLoader.getResource("sql-classification.json")
```
zhang1002, line 43: why you use getClass.getClassLoader to get resource?

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
zhang1002: here should check the jsonFile's format that the user provided... 
i plan to do it after this PR...

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
ulysses-you: It's better to keep these consistent, user can still specify the custom config.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
ulysses-you: if we really want to check the file whether is malformed or not, we can do it at a new PR in further.

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
```diff
@@ -35,29 +35,26 @@ import org.apache.kyuubi.sql.KyuubiSQLConf._
  *  If there have no this named jsonFile,
  *  the service will upload the default matching rule: sql-classification-default.json.
  */
-object KyuubiGetSqlClassification {
-
+object KyuubiGetSqlClassification extends Logging {
   private val jsonNode: Option[JsonNode] = {
     SQLConf.get.getConf(SQL_CLASSIFICATION_ENABLED) match {
       case true =>
```
yaooqinn, line 41: can we also change case true and false to if-else, the code here looks not elegant

yaooqinn (approved):

dev/kyuubi-extension-spark-3-1/src/main/scala/org/apache/kyuubi/sql/sqlclassification/KyuubiGetSqlClassification.scala
```diff
@@ -35,29 +35,25 @@ import org.apache.kyuubi.sql.KyuubiSQLConf._
...
-        val defaultSqlClassificationFile = url.getPath
-        Some(objectMapper.readTree(new File(defaultSqlClassificationFile)))
-      case false =>
-        None
+    if (SQLConf.get.getConf(SQL_CLASSIFICATION_ENABLED)) {
+      val objectMapper = new ObjectMapper
+      var url: URL = getClass.getClassLoader.getResource("sql-classification.json")
+      if (url == null) {
+        logInfo("sql-classification.json is not found, use default config instead")
+        url = getClass.getClassLoader.getResource("sql-classification-default.json")
```
pan3793, line 45: You provide an example `file:/Users/cathy/Desktop/tmp/spark-3.1.2-bin-hadoop2.7/jars/kyuubi-extension-spark-3-1_2.12-1.4.0-SNAPSHOT.jar!/sql-classification-default.json` if `sql-classification-default.json` is outside of the jar, but if `sql-classification-default.json` is inside a jar, what would the `url` look like? Does it also work?

pan3793 (approved):
The change LGTM, updated title and descriptions to link the related PRs.

pan3793:
Thanks, merging to master.

Issue #1145, opened 2021-09-24
[Bug] Suppress MiniDFS/YARNCluster/ Kerberos output in tests


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

```
HadoopFsDelegationTokenProviderSuite:
>>>DEBUG: TCPClient reading 173 bytes
>>> KrbKdcReq send: #bytes read=173
>>>Pre-Authentication Data:
	 PA-DATA type = 19
	 PA-ETYPE-INFO2 etype = 17, salt = null, s2kparams = null

>>> KdcAccessibility: remove localhost:50936
>>> KDCRep: init() encoding tag is 126 req type is 11
>>>KRBError:
	 sTime is Fri Sep 24 10:40:47 CST 2021 1632451247000
	 suSec is 100
	 error code is 25
	 error Message is Additional pre-authentication required
	 sname is client/localhost@KERBERIZEDTESTHELPER.COM
	 eData provided.
	 msgType is 30
>>>Pre-Authentication Data:
	 PA-DATA type = 19
	 PA-ETYPE-INFO2 etype = 17, salt = null, s2kparams = null

KRBError received: Additional pre-authentication required
KrbAsReqBuilder: PREAUTH FAILED/REQ, re-send AS-REQ
default etypes for default_tkt_enctypes: 17.
Looking for keys for: client/localhost@KERBERIZEDTESTHELPER.COM
Added key: 16version: 1
Added key: 17version: 1
Looking for keys for: client/localhost@KERBERIZEDTESTHELPER.COM
Added key: 16version: 1
Added key: 17version: 1
default etypes for default_tkt_enctypes: 17.
>>> EType: sun.security.krb5.internal.crypto.Aes128CtsHmacSha1EType
>>> KrbAsReq creating message
>>> KrbKdcReq send: kdc=localhost TCP:50936, timeout=30000, number of retries =3, #bytes=260
>>> KDCCommunication: kdc=localhost TCP:50936, timeout=30000,Attempt =1, #bytes=260
>>>DEBUG: TCPClient reading 620 bytes
>>> KrbKdcReq send: #bytes read=620
>>> KdcAccessibility: remove localhost:50936
[...]

PR #1150, opened 2021-09-24
[KYUUBI #1145] Suppress MiniDFSCluster/MiniYARNCluster/Kerberos output in tests

zhouyifan279:

- Kerberos login module outputs lots of debug level log records to stdout, which is annoying at most of the time.
- MiniYARNCluster already outputs most of the log records to unit-tests.log. But some referred classes, such as `GuiceComponentProviderFactory` use java.util.logging.Logging and output log records to stdout.

In this PR, we turn off kerberos logging by default and allow users to turn on it by setting java system property `sun.security.krb5.debug`. java.util.logging.Logging is redirected to slf4j using `SLF4JBridgeHandler`.

**Note:** MiniDFSCluster outputs a few log records using `System.stdout`/`System.stderr`.
```
Formatting using clusterid: testClusterID
isSaslEnabled:false
Opened streaming server at /0.0.0.0:1025
Successfully obtained privileged resources (streaming port = ServerSocket[addr=/0:0:0:0:0:0:0:0,localport=1025] ) (http listener port = 9864)
Opened info server at /0.0.0.0:9864
```
It should be acceptable.

### DIFF
--- a/kyuubi-common/pom.xml
+++ b/kyuubi-common/pom.xml
@@ -53,11 +53,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>jul-to-slf4j</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client-api</artifactId>
         </dependency>
--- a/kyuubi-common/pom.xml
+++ b/kyuubi-common/pom.xml
@@ -53,6 +53,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client-api</artifactId>
         </dependency>
--- a/kyuubi-common/src/test/scala/org/apache/kyuubi/KyuubiFunSuite.scala
+++ b/kyuubi-common/src/test/scala/org/apache/kyuubi/KyuubiFunSuite.scala
@@ -36,7 +36,7 @@ trait KyuubiFunSuite extends AnyFunSuite
   with ThreadAudit
   with Logging {
 
-  // Redirect jcl to sl4j
+  // Redirect jul to sl4j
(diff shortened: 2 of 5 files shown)


kyuubi-common/src/test/scala/org/apache/kyuubi/KyuubiFunSuite.scala
```diff
@@ -34,6 +35,11 @@ trait KyuubiFunSuite extends AnyFunSuite
   with Eventually
   with ThreadAudit
   with Logging {
+
+  // Redirect jcl to sl4j
```
pan3793, line 39: jul?

pan3793:
The change LGTM, please update the PR title, `Kyuubi #1145` => `[Kyuubi #1145]`

kyuubi-common/src/test/scala/org/apache/kyuubi/KyuubiFunSuite.scala
zhouyifan279: Yes, updated.

zhouyifan279:
@yaooqinn ready for review.

kyuubi-common/pom.xml
```diff
@@ -52,6 +52,11 @@
             <artifactId>slf4j-log4j12</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+        </dependency>
```
yaooqinn, line 58: test only

zhouyifan279: kyuubi-server's test suites can not see this dependency if scope is set to test.

zhouyifan279: kyuubi-server's test suites can not see this dependency if scope is set to test.

pan3793: Add it in `kyuubi-server/pom.xml` again

zhouyifan279: > Add it in kyuubi-server/pom.xml again
This is quite inconvenient because all modules depends `kyuubi-common` in test scope have to add `jul-to-slf4j`.

pan3793 (approved):

yaooqinn (approved):

pan3793:
Thanks, merging to master

Issue #1131, opened 2021-09-22
[Umbrella] Rework kyuubi-hive-jdbc module


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the proposal

Rework `kyuubi-hive-jdbc` module, remove hive dependencies to make `kyuubi-hive-jdbc` light and maintainable.

### Task list

- #1147
- #1191
- #1203
- #1149
- #1151
- #1152
- #1231

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yaooqinn:
you need to create actual sub-task issues instead of use the id of this ticket everytime

PR #1152, opened 2021-09-24
[KYUUBI #1131][TEST] Replace hive-jdbc by kyuubi-hive-jdbc in UT

pan3793:


pan3793:
cc @zhouyifan279, since this PR got merged, we sealed all hive deps used by `kyuubi-hive-jdbc` module to its own pom. Could you please clean up the hive deps used by `kyuubi-server` module? e.g. avoid using `hive-exec` fat jar, manage all exclusion in the root pom `<dependencyManagement>` block.

PR #1153, opened 2021-09-24
[KYUUBI #1066] Automatically create LogRoot when the directory does not exist 

zhang1002:
bugfix:
#1066 


zhang1002:
CC @ulysses-you @yaooqinn

kyuubi-common/src/main/scala/org/apache/kyuubi/events/JsonEventLogger.scala
```diff
@@ -75,6 +75,15 @@ class JsonEventLogger[T <: KyuubiEvent](logName: String,
   }
 
   private def requireLogRootWritable(): Unit = {
+    // Create logRoot
+    // Here should notice, if the user has no permission, it will throw exception
+    try {
+      FileSystem.mkdirs(fs, new Path(logRoot), JSON_LOG_DIR_PERM)
```
yaooqinn, line 81: LogRoot probably shares across users, is it good practice that making it be created by one single user?

zhang1002: got it.
thanks.

PR #1154, opened 2021-09-24
[KYUUBI #1067]catch exception if logRoot is not exist

zhang1002:
BUGFIX:
#1067 
This PR is just for version_1.3.0

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -65,8 +65,10 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
   }
 
   override def stop(): Unit = {
-    eventLogging.onEvent(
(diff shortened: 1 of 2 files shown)


yaooqinn (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -65,8 +65,10 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
   }
 
   override def stop(): Unit = {
-    eventLogging.onEvent(
-      engineStatus.copy(state = ServiceState.STOPPED.id, endTime = System.currentTimeMillis()))
+    if (!state.equals(ServiceState.LATENT)) {
```
ulysses-you, line 68: should it be `if (state == ServiceState.STARTED)` ?

zhang1002: i think state judgement should in the method of JsonEventLogger.getOrUpdate.
only "STARTED" state can provided this service.

this need @yaooqinn to confirm.

PR #1155, opened 2021-09-25
[BUILD] Support passthrough env var CI into build container

pan3793:

Continuous integration(aka. CI) services like GitHub Actions, Travis always provide an environment variable `CI` in runners, and we detect this variable to run some specific actions.

In [KYUUBI #742], we add a `CI` env var detection in `build/mvn` to suppress noisy maven logs in CI scenes.

This PR introduces a new build arg `CI` in `docker/Dockerfile`, which is intent to passthrough env var CI into the build container to suppress noisy maven logs in GitHub Action and Travis.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -154,7 +154,8 @@ jobs:
       - name: Build Kyuubi Docker Image
         uses: docker/build-push-action@v2
         with:
-          build-args: CI="GTIHUB_ACTION_DOCKER"
+          # passthrough CI into build container
+          build-args: CI=${CI}
           context: .
           file: docker/Dockerfile
           load: true
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -141,6 +141,7 @@ jobs:
         run: |
           ./build/mvn clean install -Pspark-3.1 -DskipTests -pl :kyuubi-spark-sql-engine_2.12,:kyuubi-common_2.12,:kyuubi-ha_2.12,:kyuubi-zookeeper_2.12,:kyuubi-spark-monitor_2.12
           ./build/mvn test -Pspark-3.1 -Dtest=none -DwildcardSuites=org.apache.kyuubi.operation.tpcds -Dmaven.plugin.scalatest.exclude.tags=''
+
   minikube-it:
     name: Minikube Integration Test
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


pan3793:
@ulysses-you Just curiosity, why `ARG kyuubi_uid=10009` is lowercase?

.github/workflows/master.yml
```diff
@@ -153,6 +154,7 @@ jobs:
       - name: Build Kyuubi Docker Image
         uses: docker/build-push-action@v2
         with:
+          build-args: CI="GTIHUB_ACTION_DOCKER"
```
pan3793, line 157: We can set everything except empty string to `CI` to suppress maven logs.

docker/Dockerfile
```diff
@@ -28,8 +32,11 @@ ARG BASE_IMAGE=openjdk:8-jdk
 
 FROM maven:3.6-jdk-8 as builder
 
+ARG CI
 ARG MVN_ARG
 
+ENV CI ${CI}
```
yaooqinn, line 38: what’s this？

pan3793: Both GitHub Action and Travis will add `CI` into env.

https://github.com/apache/incubator-kyuubi/blob/master/build/mvn#L27-L29

ulysses-you: I think the name is not readable, can we change it to a more meaning name ? or can we use `MAVEN_CLI_OPTS` directly ?

pan3793: How about change to
```
ARG LESS_MAVEN_LOG
ENV CI ${LESS_MAVEN_LOG}
```

yaooqinn:
the Dockerfile I made at the very beginning is release-oriented, now is dev-oriented, and after this PR is test-oriented?

BTW, as maintainers, can we put some content to describe what the PR does? If we are not following this, how can we force others to follow？

pan3793:
> the Dockerfile I made at the very beginning is release-oriented, now is dev-oriented, and after this PR is test-oriented?

If we don't set the new build arg `CI`, it will keep the current behavior. Set build arg `CI` in GA and Travis are straightforward I think.

> BTW, as maintainers, can we put some content to describe what the PR does? If we are not following this, how can we force others to follow？

Will update soon.

docker/Dockerfile
pan3793: The previous PR title and description may be misleading. Updated and please review again.

cfmcgrady (approved):

cfmcgrady:
I think it's already test-oriented after [PR-955](https://github.com/apache/incubator-kyuubi/pull/955).

yaooqinn (approved):

ulysses-you (approved):

ulysses-you:
thanks, merging to master

ulysses-you:
I think we should consider make a new dockerfile that use to release.

PR #1160, opened 2021-09-26
Support to config operation log root dir both for kyuubi server and engine sides

turboFei:

Support to config operation log root dir with kyuubi configuration both for server and engine sides.


kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -25,16 +25,26 @@ import java.nio.file.{Files, Path, Paths}
...
+import org.apache.kyuubi.config.KyuubiConf.{ENGINE_OPERATION_LOG_DIR_ROOT, SERVER_OPERATION_LOG_DIR_ROOT}
 import org.apache.kyuubi.operation.OperationHandle
 import org.apache.kyuubi.session.SessionHandle
 
 object OperationLog extends Logging {
+  val conf = KyuubiConf()
 
-  def LOG_ROOT: String = if (Utils.isTesting) {
-    "target/operation_logs"
+  def SERVER_LOG_ROOT: String = if (Utils.isTesting) {
```
yaooqinn, line 36: move this to session managers.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -31,6 +32,12 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
 
   def this() = this(classOf[KyuubiSessionManager].getSimpleName)
 
+  override def LOG_ROOT: String = if (Utils.isTesting) {
```
yaooqinn, line 35: we don't need this anymore, we can set SERVER_OPERATION_LOG_DIR_ROOT in test or maven pom file

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -31,6 +32,12 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
...
+    "target/server_operation_logs"
+  } else {
+    conf.get(SERVER_OPERATION_LOG_DIR_ROOT).getOrElse("server_operation_logs")
```
yaooqinn, line 38: it seems that we need to pre-touch or validate the root dir

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
yaooqinn, line 35: operationLogRoot

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -41,6 +41,8 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
 
   @volatile private var shutdown = false
 
+  def LOG_ROOT: String
```
yaooqinn, line 44: // initialize it in `def initialize(conf: KyuubiConf)`
protected var _operationLogRoot 
def operationLogRoot = _operationLogRoot

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -41,6 +43,20 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
 
   @volatile private var shutdown = false
 
+  protected var _operationLogRoot: String = null
```
turboFei, line 46: have to set it to null, otherwise, we need override it in sub class

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -41,6 +43,20 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
...
+
+  def operationLogRoot: String = _operationLogRoot
+
+  private def initOperationLogRootDir(): Unit = {
```
turboFei, line 50: We need init it in session manager, but I am not sure whether throw the exception if it failes

yaooqinn: maybe we can make operationLogRoot and Option[String], if we fail to initialize the _operationLogRoot, then we don't initialize the OperationLog for operations and just return the result

yaooqinn: or maybe we just remove this step for now

zhang1002: i think it should throw the exception or logger it in the log.

when user set this configuration item by themself, this dir perhaps has no permission for KYUUBI to create. It should let user to know.

yaooqinn: > i think it should throw the exception or logger it in the log.
> 
> when user set this configuration item by themself, this dir perhaps has no permission for KYUUBI to create. It should let user to know.

It seems that you misunderstand something here, the config is from sessionManager not from users

zhang1002: user set this configuration item.
KYUUBI create this dir.

yaooqinn (approved):

turboFei:
thanks for your review, merged to master

Issue #1164, opened 2021-09-27
[Bug] Grep WARN in dependency.sh


hddong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When `MVN dependency` has WARN log, `./build/dependency.sh --replace` will exit with execption:
```
awk: trying to access out of range field -1
 input record number 1, file 
 source line number 2
```

### Affects Version(s)

master/1.3.0

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1165, opened 2021-09-27
[KYUUBI #1164] Grep WARN in dependency.sh

hddong:

When MVN dependency has WARN log, ./build/dependency.sh --replace will exit with execption:
```
awk: trying to access out of range field -1
 input record number 1, file 
 source line number 2
```


yaooqinn (approved):

pan3793:
Thanks, merging to master.

PR #1167, opened 2021-09-27
Add zorder core benchmark

ulysses-you:

Add zorder core benchmark, so we can get the perf number of zorder.

This PR adds two column type `int` and `long` benchmark:
- 2 int cols
- 3 int cols
- 4 int cols
- 2 long cols
- 3 long cols
- 4 long cols


yaooqinn:
for read or write？

yaooqinn (approved):

Issue #1169, opened 2021-09-27
[FEATURE] Decouple EngineLoggingService From SparkSQLEngine


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

EngineLoggingService initializes and starts separately and ahead of SparkSQLEngine so that we can trace all status changes for engine life cycle 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1171, opened 2021-09-27
[KYUUBI #1169] Decouple EngineLoggingService From SparkSQLEngine

yaooqinn:

1. Make EngineLoggingService tolerant for creating if any error occurs, and then the main code goes.

2. Separate EngineLoggingService From SparkSQLEngine, and we will try to start EngineLoggingService before SparkSQLEngine to capture all the lifecycle statuses. On the other hand, add a shutdown hook that executed after SparkSQLEngine.stop for  EngineLoggingService.stop

3. try-catch for every step of SparkSQLEngine bootstrapping for  creating better diagnosis messages.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -149,26 +149,23 @@ object SparkSQLEngine extends Logging {
     SignalRegister.registerLogger(logger)
     var spark: SparkSession = null
     try {
+      spark = createSpark()
       try {
-        spark = createSpark()
-        try {
-          startEngine(spark)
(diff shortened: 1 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -127,38 +111,64 @@ object SparkSQLEngine extends Logging {
...
     }
   }
 
   def main(args: Array[String]): Unit = {
     SignalRegister.registerLogger(logger)
     var spark: SparkSession = null
     try {
-      spark = createSpark()
-      startEngine(spark)
+      try {
```
ulysses-you, line 152: shall we combine this line `try` and line 151 `try` ?

ulysses-you (approved):

yaooqinn:
thanks, merged to master

Issue #1008, opened 2021-09-01
Implement Hadoop File System Delegation Token Provider


zhouyifan279:
### 1. Describe the feature
This is another sub-task under umbrella issue #913 and depends on task #1007.

In this task, a HadoopFsDelegationTokenProvider is implemented to obtain delegation tokens from Hadoop File System server.

By default, HadoopFsDelegationTokenProvider only obtains delegation token of the file system that hosts fs.defaultFS.
Extra Hadoop filesystem URLs can be specified in config `kyuubi.credentials.hadoopfs.urls` in a comma separated format.

### 2. Motivation
See #913 

### 3. Describe the solution
Requirement:

- Hadoop client configurations should be placed under Kyuubi Server's classpath.
- `kyuuib.kinit.principal` should be enabled to impersonate users in Hadoop File System server.

Implement:

- Use `UserGroupInformation#isSecurityEnabled` to decide whether delegation tokens are required.
- Use `FileSystem#getDelegationToken` to obtain delegation tokens.

PR #1173, opened 2021-09-28
[KYUUBI #1008][FOLLOWUP] No need to send private token from Kyuubi server to engine

zhouyifan279:
When communicating with NameNode HA, private tokens are created from original token:
image
In above image, "ha-hdfs:mycluster" is the original token. "ha-hdfs://192.168.56.101:8020" and "ha-hdfs://192.168.56.102:8020" are private tokens.

`KyuubiHadoopUtils.getCredentialsInternal` was supposed to extract these private tokens at Kyuubi server side and send them to SQL engine.

But in fact, SQL engine side private tokens are automatically created when adding original token, 
 inside `org.apache.hadoop.security.Credentials#addToken`.

SequenceNumber of private tokens always equal to original token:
image
image

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiHadoopUtils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiHadoopUtils.scala
@@ -19,7 +19,6 @@ package org.apache.kyuubi.util
 
 import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream, DataOutputStream}
 import java.util.{Map => JMap}
-import javax.security.auth.Subject
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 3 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiHadoopUtils.scala
```diff
@@ -76,19 +69,6 @@ object KyuubiHadoopUtils {
...
 
-  /**
-   * Get all tokens in [[UserGroupInformation#subject]] including
-   * [[org.apache.hadoop.security.token.Token.PrivateToken]] as
-   * [[UserGroupInformation#getCredentials]] returned Credentials do not contain
-   * [[org.apache.hadoop.security.token.Token.PrivateToken]].
-   */
-  def getCredentialsInternal(ugi: UserGroupInformation): Credentials = {
-    // Synchronize to avoid credentials being written while cloning credentials
-    subjectField.get(ugi).asInstanceOf[Subject] synchronized {
```
yaooqinn, line 87: rm subjectField too

Issue #1174, opened 2021-09-28
[FEATURE] Service discovery supports kerberos


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

The client can connect to the kyuubi servcie with kerberos authentication through zookeeper, and it is also convenient for users to migrate to kyuubi.

### Motivation

Now kyuubi servcie uses kerberos authentication and registers with the zookeeper to the node information, but the client cannot connect to the service.

### Describe the solution

Write kerberos authentication information to zookeeper znode.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1175, opened 2021-09-28
[KYUUBI #1174] Service discovery supports kerberos

cxzl25:
https://github.com/apache/incubator-kyuubi/issues/1174
Now kyuubi servcie uses kerberos authentication and registers with the zookeeper to the node information, but the client cannot connect to the service.


cxzl25:
## Current 
image
## Enable
image

yaooqinn:
Kyuubi supports Kerberos authentication already, what will these magic configurations help?

cxzl25:
> Kyuubi supports Kerberos authentication already, what will these magic configurations help?

Use the command
```
bin/beeline -u "jdbc:hive2://Zookeeper/default;serviceDiscoveryMode=zooKeeper;zooKeeperNamespace=kyuubi"
```
Get error
```
21/09/28 19:03:39 ERROR Utils: Unable to read HiveServer2 configs from ZooKeeper
Unknown HS2 problem when communicating with Thrift server.
Error: Could not open client transport for any of the Server URI's in ZooKeeper: Peer indicated failure: Unsupported mechanism type PLAIN (state=08S01,code=0)
java.sql.SQLException: Could not open client transport for any of the Server URI's in ZooKeeper: Peer indicated failure: Unsupported mechanism type PLAIN
	at org.apache.hive.jdbc.HiveConnection.<init>(HiveConnection.java:224)
```

Similar to this configuration and function introduced by Hive    
https://issues.apache.org/jira/browse/HIVE-11581  
HIVE-11581: HiveServer2 should store connection params in ZK when using dynamic service discovery for simpler client connection string

yaooqinn:
thanks, merged to master

pan3793:
@cxzl25 we should change the default value `kyuubi.ha.zookeeper.publish.configs` to `false`, otherwise will cause the low version hive client failed to parse service info when kerberos enabled.

cxzl25:
> @cxzl25 we should change the default value of `kyuubi.ha.zookeeper.publish.configs` to `false`, otherwise will cause the low version hive client failed to parse service info when kerberos enabled.

OK, I will submit a pr.

PR #1192, opened 2021-10-08
Refine zorder utils toByte

ulysses-you:

Fix some issue:
1. int/long type miss update highest byte of byte array
2. short/int/long same value may have different byte array
3. the mask value of long is error. (1L << 31 instead of 1 << 31)
4. the default value of string should be Long.Max since we pad it to 8 bytes so we can keep the ordering of null value (null last)
5. boolean to byte doesn't need 4 bytes, 1 byte is enough


ulysses-you:
thanks for review, merging to master !

PR #1218, opened 2021-10-12
Z-order int and long should respect negative number

ulysses-you:

Made a mistake at https://github.com/apache/incubator-kyuubi/pull/1192 that we can't reverse the short highest bit for int and int highest bit for long. Otherwise the negative number will get the wrong order.


ulysses-you:
cc @cfmcgrady @yaooqinn

Issue #1225, opened 2021-10-13
[FEATURE] Add kyuubi-extension-spark-3-2 module


ulysses-you:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add kyuubi-extension-spark-3-2 module so we can use some new nodes of Spark-3.2

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1226, opened 2021-10-13
[KYUUBI #1225] Add kyuubi-extension-spark-3-2 module

ulysses-you:

see https://github.com/apache/incubator-kyuubi/issues/1225

Note that:
- this PR only supports `Zorder` and some common rules with Spark 3.2.0. The others `watchdog` and `sqlclassification` should be supported in future. 
- this PR dones't supprpt shade jar and we also need to update build/dist script in a new PR.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -49,7 +49,6 @@ jobs:
             spark: '3.1'
             spark-hadoop: '3.2'
             profiles: '-Pkyuubi-extension-spark-3-2'
-            codecov: 'true'
           - java: 11
             spark: '3.1'
             spark-hadoop: '3.2'


ulysses-you:
cc @cfmcgrady @i7xh @zhang1002 @yaooqinn @pan3793 @turboFei , since it's a big change for spark extension module.

pan3793:
I propose to simplify maven profiles in this way.

1. keep spark-3.0, spark-3.1, spark-3.1
2. when spark-3.1 is active, enable `kyuubi-extension-spark-common` module and `kyuubi-extension-spark-3-1` module
3. do same thing when spark-3.2 is active

Then we can remove profiles `kyuubi-extension-spark-3-1` and `kyuubi-extension-spark-3-2`, and not required to overwrite `spark.version` in sub modules. @ulysses-you @yaooqinn WDYT?

Issue #1176, opened 2021-09-28
[Bug] InvalidACL appears in the engine when zookeeper acl is turned on


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When `kyuubi.ha.zookeeper.acl.enabled=true`, both service and engine will use zookeeper acl to create znode, but engine has no keytab information and cannot write information to zookeeper, throwing an exception.

```java
Caused by: org.apache.zookeeper.KeeperException$InvalidACLException: KeeperErrorCode = InvalidACL for /kyuubi_USER/XXXX
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:124)
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:54)
	at org.apache.zookeeper.ZooKeeper.create(ZooKeeper.java:792)
	at org.apache.kyuubi.shade.org.apache.curator.framework.imps.CreateBuilderImpl$11.call(CreateBuilderImpl.java:740)
```

### Affects Version(s)

master

### Kyuubi Server Configurations

```yaml
kyuubi.ha.zookeeper.acl.enabled true
```

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1235, opened 2021-10-14
[KYUUBI #1176][FOLLOWUP] Fix NoAuth when engine disable acl

cxzl25:
The `EngineRef` of the server uses `InterProcessSemaphoreMutex`, which will create a znode with acl, causing the engine to be unable to write to the znode (`KeeperErrorCode = NoAuth`)


zhouyifan279:
As far as I know, this lock znode is only accessed by Kyuubi Server. Can you provide the error stack trace to tell us when engine writes to this znode ?

PR #1240, opened 2021-10-15
Update release docs

ulysses-you:

Make the docs easy to use.

docs only


ulysses-you:
cc @pan3793

PR #1247, opened 2021-10-18
[KYUUBI #1085] Fix-Enforce maxOutputRows for Resolve View, Multi Insert and Distribute By 

lsm1:

1. Should not add limit when multi insert tables.

```sql
FROM tmp_table
insert into tmp_table1 select * limit 2
insert into tmp_table2 select * 
```

We should not add any limit when insert table 

2. Should add limit when distribute by 

```sql
select * from tmp_table distribute by a
```

3. Should not add limit when resolve view

```sql
SELECT * FROM
(select * from tmp_view1 union select * from tmp_view2)
ORDER BY a DESC
```
image

Like this,  when the limit is added in resolve view, the complete data cannot be obtained, so the final result is incorrect

image


lsm1:
> thank you @lsm1 for the fix, this is based on Spark3.1 right ? also cc @i7xh

yes, this is based on Spark3.1

ulysses-you:
thank you @lsm1 , merging to master

i7xh:
@lsm1 nice job!
🤝

Issue #1258, opened 2021-10-19
[SUB-TASK][KPIP-1] Implement /sessions/execpoolsize and /sessions/execpoolactivecount 


simon824:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

This is a subtask of umbrella issue #KPIP-1

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yanghua:
Hi @simon824 Can you please also archive this issue to #950 ?

PR #1259, opened 2021-10-19
[KYUUBI #1258] Implement sessions/execpool/statistic

simon824:
This is a subtask of umbrella issue #KPIP-1


simon824:
> @simon824 thanks for your contribution.
> 
> After serious thinking, I think these two APIs are a bit ugly. It's my problem, No intention to offend you.
> 
> I am considering two candidates:
> 
> * `/sessions/execpool/size` & `/sessions/execpool/activecount`
> * `/sessions/execpool/statistic` which returns a packaged object contains size & activecount and so on
> 
> Which one do you prefer?

Thanks a lot for your review , I think option 2 is better.

PR #1301, opened 2021-10-27
Support kyuubi procedures

turboFei:

Support kyuubi defined procedures, refer Iceberg-Spark procedures(https://iceberg.apache.org/spark-procedures/)


pan3793:
cc @ulysses-you since this PR rename lots of package name of classes.

turboFei:
> cc @ulysses-you since this PR renames lots of package name of classes.

Some class is only accessible in package `org.apache.spark.sql`, so I rename the package for convenience.

And I think the extension is only for spark-sql, so, it makes senses.

turboFei:
But it seems that it also rename class  `KyuubiSparkSQLExtension`.

So, let me make it compatible with before.

turboFei:
> cc @ulysses-you since this PR renames lots of package name of classes.

revert back and just add a util class for some classes that only can be accessed in package `org.apache.spark.sql`.

ulysses-you:
I think we can combine this grammar (`EXEC ...` ?) and zorder, adding a new .g4 and something else seems not friendly for maintain.

turboFei:
> I think we can combine this grammar (`EXEC ...` ?) and zorder, adding a new .g4 and something else seems not friendly for maintain.
sure

turboFei:
if it is needed, will reopen it

Issue #1300, opened 2021-10-27
[Bug] Prompt read timeout when executing select * query in beeline


jiaoqingbo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

show test demo as below

CREATE DATABASE IF NOT EXISTS aqe_demo_db;
USE `aqe_demo_db;

CREATE TABLE items
USING parquet
AS
SELECT id AS i_item_id,
CAST(rand() * 1000 AS INT) AS i_price
FROM RANGE(30000000);

then use beeline to connect

./bin/beeline -u 'jdbc:hive2://ocdp39:2181,vm01.asia.com:2181,vm02.asia.com:2181/default;serviceDiscoveryMode=zooKeeper;zooKeeperNamespace=kyuubi;principal=spark/_HOST@ocdp;'

and then execute

use aqe_demo_db;

select * from items ;

after that，I saw an exception message

<img width="834" alt="微信截图_20211027161048" src="https://user-images.githubusercontent.com/14961757/139026504-5c864ac4-d412-4acc-9efe-4c22edacad11.png">

### Affects Version(s)

1.3.0

### Kyuubi Server Log Output

```logtalk
2021-10-27 16:08:56.235 INFO operation.ExecuteStatement: Processing ambari-qa's query[d45d9f09-ea2a-4367-9983-9ddec491fa0c]: RUNNING_STATE -> ERROR_STATE, statement: select * from items, time taken: 25.232 seconds
2021-10-27 16:08:56.236 DEBUG server.KyuubiBackendService: OperationHandle [type=EXECUTE_STATEMENT, identifier: d45d9f09-ea2a-4367-9983-9ddec491fa0c]: The background operation was aborted, org.apache.kyuubi.KyuubiSQLException: Error operating EXECUTE_STATEMENT: java.net.SocketTimeoutException: Read timed out
2021-10-27 16:09:00.826 WARN operation.ExecuteStatement: Ignore exception in terminal state with d45d9f09-ea2a-4367-9983-9ddec491fa0c: org.apache.thrift.TApplicationException: CloseOperation failed: out of sequence response
```

### Kyuubi Engine Log Output

```logtalk
No abnormal information
```

### Kyuubi Server Configurations

```yaml
[...]

PR #1312, opened 2021-10-29
[KYUUBI #1300] Detecting critical errors

yaooqinn:

For critical errors at engine side, it is not handled properly. For example,when engine oom
- server may not be able to get operation statuses because of no response of engine side. In this case, client only get a ambiguous `read timeout` as a final cause.
- the oom hook of engine side might directly crash the engine, when the server is still trying to get operation status

In this PR,
- a config for retry to make the operation status updating process more robust
- make the engine oom hook only de register itself to make it able to recover for some transient errors 


yaooqinn:
cc @turboFei @pan3793 @zhouyifan279

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
```diff
@@ -83,18 +83,39 @@ class ExecuteStatement(
...
+      } catch {
+        case e: TException if currentAttempts >= maxStatusPollOnFailure =>
+          error(s"Failed to get ${session.user}'s query[$getHandle] status after" +
+            s" $maxStatusPollOnFailure times, aborting", e)
+          throw e
+        case e: TException =>
+          currentAttempts += 1
+          warn(s"Failed to get ${session.user}'s query[$getHandle] status" +
+            s" ($currentAttempts / $maxStatusPollOnFailure)", e)
+          Thread.sleep(500)
```
pan3793, line 102: Configurable?

pan3793 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
yaooqinn: Look fine to me to be hardcoded. As users already have operation.status.polling.timeout

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkThriftBinaryFrontendService.scala
```diff
@@ -147,6 +147,15 @@ class SparkThriftBinaryFrontendService(
       s"${serverAddr.getHostAddress}:$portNum"
     }
   }
+
+  // When a OOM occurs, here we de-register the engine by stop its discoveryService.
+  // Then the current engine will not be connected by new client anymore but keep the existing ones
+  // alive. In this case we can reduce the engine's overhead and make it possible recover from that.
+  // We shall not tear down the whole engine by serverable.stop to make the engine unreachable for
+  // the exising clients which are still get statuses and report the end-users.
```
zhouyifan279, line 155: 1. `exising` -> `existing`
2. `report the end-users` -> `report to the end-users`

pan3793:
Thanks, merging to master

PR #1317, opened 2021-11-01
[KYUUBI #1106] Implement /sessions/${identifier}/info/${infotype}

simon824:
This is a subtask of umbrella issue #KPIP-1


yaooqinn:
cc @yanghua @pan3793 PTAL

yanghua:
> cc @yanghua @pan3793 PTAL

Got it! Will review soon.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/dto.scala
```diff
@@ -32,6 +32,11 @@ case class SessionOverview(
   sessionHandle: SessionHandle
 )
 
+case class InfoValue(
```
yanghua, line 35: `InfoOverview` or `InfoDetail` sounds better? Because you have a field named `infoValue `, so it is a bit confusing. WDYT?

simon824: Make sense, updated to InfoDetail.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -61,6 +57,16 @@ private[v1] class SessionsResource extends ApiRequestContext {
     }
   }
 
+  @GET
+  @Path("{sessionHandle}/info/{infotype}")
```
yanghua, line 61: Keep the same code style about `sessionHandle ` and `infotype `? camelCase or not?

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/SessionsResourceSuite.scala
```diff
@@ -171,4 +171,26 @@ class SessionsResourceSuite extends KyuubiFunSuite {
...
+      (_, _, _, webTarget) =>
+        var response: Response = webTarget.path("api/v1/sessions")
+          .request(MediaType.APPLICATION_JSON_TYPE)
+          .post(Entity.entity(requestObj, MediaType.APPLICATION_JSON_TYPE))
+
+        val sessionHandle = response.readEntity(classOf[SessionHandle])
+        val serializedSessionHandle = s"${sessionHandle.identifier.publicId}|" +
+          s"${sessionHandle.identifier.secretId}|${sessionHandle.protocol.getValue}"
+
+        response = webTarget.path(s"api/v1/sessions/$serializedSessionHandle/info/13")
```
yanghua, line 189: We may need an API (in the following PR) to get all the enums of `TGetInfoType`, otherwise, no one knows what does the `13` mean and how many numbers like this (from the client). WDYT?

simon824: OK

yanghua (approved):
@simon824 Thanks for your contribution. LGTM cc @yaooqinn

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/SessionsResourceSuite.scala
yanghua: sounds good

yanghua: FYI: added this API into the design doc.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -61,14 +57,24 @@ private[v1] class SessionsResource extends ApiRequestContext {
     }
   }
 
+  @GET
+  @Path("{sessionHandle}/info/{infoType}")
+  def getInfo(@PathParam("sessionHandle") sessionHandleStr: String,
+              @PathParam("infoType") infoType: Int): InfoDetail = {
+    val sessionHandle = getSessionHandle(sessionHandleStr)
+    val info = TGetInfoType.findByValue(infoType.toInt)
```
yaooqinn, line 65: can we test a negative case

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -61,14 +57,30 @@ private[v1] class SessionsResource extends ApiRequestContext {
...
```
yaooqinn, line 65: What if infoType is invalid int？

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -61,14 +57,30 @@ private[v1] class SessionsResource extends ApiRequestContext {
...
+
+    try {
+      val infoValue = backendService.getInfo(sessionHandle, info)
+      InfoDetail(info.toString, infoValue.getStringValue)
+    } catch {
+      case _: Throwable =>
```
yaooqinn, line 71: Shall we use NonFatal？

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -61,14 +57,30 @@ private[v1] class SessionsResource extends ApiRequestContext {
...
+        throw new NotFoundException()
```
yaooqinn, line 72: shall we add a error message？

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -88,12 +100,16 @@ private[v1] class SessionsResource extends ApiRequestContext {
...
+
+  def getSessionHandle(sessionHandleStr: String): SessionHandle = {
     val splitSessionHandle = sessionHandleStr.split("\\|")
     val handleIdentifier = new HandleIdentifier(
       UUID.fromString(splitSessionHandle(0)), UUID.fromString(splitSessionHandle(1)))
     val protocolVersion = TProtocolVersion.findByValue(splitSessionHandle(2).toInt)
-    val sessionHandle = new SessionHandle(handleIdentifier, protocolVersion)
-    backendService.closeSession(sessionHandle)
-    Response.ok().build()
+    new SessionHandle(handleIdentifier, protocolVersion)
```
yaooqinn, line 113: What if the sessionHandleStr is invalid

splitSessionHandle‘s size is 1?

or the generated sessionHandle is absent

simon824:
cc @yaooqinn , thanks.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -45,19 +47,33 @@ private[v1] class SessionsResource extends ApiRequestContext {
...
+        throw new NotFoundException(s"Invalid $sessionHandle")
+    }
+  }
+
+  @GET
+  @Path("{sessionHandle}/info/{infoType}")
+  def getInfo(@PathParam("sessionHandle") sessionHandleStr: String,
+              @PathParam("infoType") infoType: Int): InfoDetail = {
+    val sessionHandle = getSessionHandle(sessionHandleStr)
+    val info = TGetInfoType.findByValue(infoType)
```
yaooqinn, line 68: what happens if info results `null`

simon824: backendService.getInfo will throw KyuubiSQLException and be catched.

yaooqinn (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -88,12 +104,27 @@ private[v1] class SessionsResource extends ApiRequestContext {
...
     backendService.closeSession(sessionHandle)
     Response.ok().build()
   }
+
+  def getSessionHandle(sessionHandleStr: String): SessionHandle = {
+    try {
+      val splitSessionHandle = sessionHandleStr.split("\\|")
+      val handleIdentifier = new HandleIdentifier(
+        UUID.fromString(splitSessionHandle(0)), UUID.fromString(splitSessionHandle(1)))
+      val protocolVersion = TProtocolVersion.findByValue(splitSessionHandle(2).toInt)
```
yaooqinn, line 117: @simon824 can you send a PR to remove TProtocolVersion for REST API. the client protocol is decided by client libraries that users use, not by users explicitly. So it's odd here to use both as a user-specified part of SessionHandle and also it is not related to the REST API at all. Also cc @yanghua

yanghua: The session handle is composed of `TProtocolVersion `. And it is the index of a session for the rest client. Here, `TProtocolVersion` is used to parse from a string presentation and recover an instance of a `SessionHandle`. Do you mean encapsulate the marshaled logic between string and `SessionHandle` object into `SessionHandle` itself, so that we can hide it from the rest component?

yaooqinn: For a session handle instance, TProtocolVersion is not necessary, as you can see the overridden equal/hashCode funcs. It's mainly used to decide the compatibility for kyuubi server and its thrift client, e.g. the result layout. I am not very familiar with REST API, does the `v1` also indicate the compatibility thing between rest clients and servers?

yanghua: restful API tries to interact with Kyuubi Server AFAP. It may provide a full capacity like the thrift protocol has provided (create sessions then fetch them). On the other hand, it also may provide management for Kyuubi server at any time, which means it would not create them but can fetch and interact with them. So it should support the session handles which contain  `TProtocolVersion `?

Issue #1318, opened 2021-11-02
[Bug] The CONN_OPEN count should be increased when the openSession succeeds, otherwise, when the open session fails, CONN_FAIL+CONN_OPEN! =CONN_TOTAL


jiaoqingbo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

in KyuubiSessionImpl.scala

```
override def open(): Unit = {
    MetricsSystem.tracing { ms =>
      ms.incCount(CONN_TOTAL)
      ms.incCount(MetricRegistry.name(CONN_OPEN, user))
    }
    withZkClient(sessionConf) { zkClient =>
      val (host, port) = engine.getOrCreate(zkClient)
      openSession(host, port)
    }
    // we should call super.open after kyuubi session is already opened
    super.open()
  }
```

The CONN_OPEN count should be increased when the openSession succeeds, otherwise, when the open session fails, CONN_FAIL+CONN_OPEN! =CONN_TOTAL

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1319, opened 2021-11-02
[KYUUBI #1318] Increase CONN_OPEN iff openSession is successful

jiaoqingbo:

The CONN_OPEN count should be increased when the openSession succeeds, otherwise, when the open session fails, CONN_FAIL+CONN_OPEN! =CONN_TOTAL

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -67,11 +67,13 @@ class KyuubiSessionImpl(
   override def open(): Unit = {
     MetricsSystem.tracing { ms =>
       ms.incCount(CONN_TOTAL)
-      ms.incCount(MetricRegistry.name(CONN_OPEN, user))
     }
     withZkClient(sessionConf) { zkClient =>
       val (host, port) = engine.getOrCreate(zkClient)
       openSession(host, port)
+      MetricsSystem.tracing { ms =>
+        ms.incCount(MetricRegistry.name(CONN_OPEN, user))
+      }
     }
     // we should call super.open after kyuubi session is already opened
     super.open()


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -67,11 +67,13 @@ class KyuubiSessionImpl(
...
     MetricsSystem.tracing { ms =>
       ms.incCount(CONN_TOTAL)
-      ms.incCount(MetricRegistry.name(CONN_OPEN, user))
     }
     withZkClient(sessionConf) { zkClient =>
       val (host, port) = engine.getOrCreate(zkClient)
       openSession(host, port)
+      MetricsSystem.tracing { ms =>
+        ms.incCount(MetricRegistry.name(CONN_OPEN, user))
+      }
```
yaooqinn, line 76: MetricsSystem.tracing(_.incCount(MetricRegistry.name(CONN_OPEN, user)))

pan3793 (approved):

pan3793:
Thanks, merging to master/1.3

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
pan3793: Sorry I have't seen this comment before merging, seems not a big deal

zhouyifan279:
A little late to comment. But I found some issue with this PR:
If `KyuubiSessionImpl#openSession` failed in `KyuubiSessionImpl#open`, metric <CONN_OPEN, user> is not increased before we decrease it in `KyuubiSessionImpl#close`. And we will get a negative number if failures occurs more often than successes.

Besides, I think this PR misunderstood the usage of metric <CONN_OPEN, user>. 
To my understanding, <CONN_OPEN, user> represents opening sessions. This convetion is also applied to engine scope CONN_OPEN and STATEMENT_OPEN.
If we want a metric to represent succeeded sessions, I think we should add a new metric such as <CONN_SUCCEED, user>.

cc @jiaoqingbo @pan3793 @yaooqinn

pan3793:
@zhouyifan279 is right, I'm going to revert this PR.

BTW, we also find other issues when reviewing the code, the method `KyuubiSessionImpl#close` has a chance to be invoked more than one time, e.g. client can send a `TCloseSessionReq` to Kyuubi server, meanwhile `TimeoutChecker` may invoke `closeSession` in the background task thread, and there may some other clean up logic will invoke `KyuubiSessionImpl#close`, there will cause `CONN_FAIL` incorrect, we'd better add a `_closed` flag and put the close logic into `synchronized { }` block.

@jiaoqingbo, would you like to take this task?

jiaoqingbo:
@zhouyifan279 I think there still has a problem ,in `ms.incCount(MetricRegistry.name(CONN_OPEN, user))` 
the `CONN_OPEN` definition is ` final val CONN_OPEN: String = CONN + "**opened**"` that means the session opened successful but in real it does not 


@pan3793  I would like to do this

zhouyifan279:
> @zhouyifan279 I think there still has a problem ,in `ms.incCount(MetricRegistry.name(CONN_OPEN, user))` the `CONN_OPEN` definition is ` final val CONN_OPEN: String = CONN + "**opened**"` that means the session opened successful but in real it does not
> 
> @pan3793 I would like to do this

Yes, you are right. `opened` should be equivalent to  "opened successfully".

PR #1325, opened 2021-11-03
[KYUUBI#1324] Add resource limit for kyuubi server on kubernetes

hddong:

Add resource limit config for kyuubi server on kubernetes.


yaooqinn (approved):

docker/helm/values.yaml
```diff
@@ -41,4 +41,16 @@ service:
...
+  # Used to specify resource, default unlimited.
+  # If you do want to specify resources:
+  #   1. remove the curly braces after 'resources:'
+  #   2. uncomment the following lines
+  # limits:
+  #   cpu: 4
+  #   memory: 10Gi
+  # requests:
+  #   cpu: 2
+  #   memory: 4Gi
```
pan3793, line 56: nit: add newline eof

hddong: Done

pan3793 (approved):

docker/helm/templates/kyuubi-deployment.yaml
```diff
@@ -46,6 +46,8 @@ spec:
             - name: frontend-port
               containerPort: {{ .Values.server.bind.port }}
               protocol: TCP
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
```
ulysses-you, line 50: shall we use `indent` rather than `nindent` here ? otherwise we will get a new empty line. I just read some [related docs](https://helm.sh/docs/chart_template_guide/function_list/#nindent).

hddong: > shall we use `indent` rather than `nindent` here ? otherwise we will get a new empty line. I just read some [related docs](https://helm.sh/docs/chart_template_guide/function_list/#nindent).

It seems that it must be `nindent` here, if use `indent`, we will get error ` mapping values are not allowed in this context`.

ulysses-you:
thanks, merged to master

Issue #1323, opened 2021-11-02
[SUB-TASK][KPIP-2] Introduce EngineType to distinguish multiple engines


yanghua:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

This is a sub-task of KPIP-2 #KPIP-2

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yaooqinn:
I can’t wait to see this happen.

PR #1326, opened 2021-11-03
[KYUUBI #1323] Introduce EngineType to distinguish multiple engines

yanghua:


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -758,6 +758,21 @@ object KyuubiConf {
     .version("1.2.0")
     .fallbackConf(LEGACY_ENGINE_SHARE_LEVEL)
 
+  val ENGINE_TYPE: ConfigEntry[String] = buildConf("engine.type")
+    .doc("Specify the detailed engine that supported by the Kyuubi. The engine type bindings to" +
+      " SESSION scope. Currently, available configs are: <ul>" +
```
yaooqinn, line 763: let's add `This configuration is experimental.`

yanghua: OK~

pan3793 (approved):
LGTM

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/EngineType.scala
```diff
@@ -0,0 +1,28 @@
...
+package org.apache.kyuubi.engine
+
+/**
+ * Defines different engine types supported by Kyuubi.
+ */
+object EngineType extends Enumeration {
+  type EngineType = Value
+  val
+  SPARK_SQL,
+  FLINK_SQL = Value
```
yaooqinn, line 27: Let's follow this scala style

* An enumeration class or object which extends the Enumeration class shall follow the convention for classes and objects, i.e. its name should be in PascalCase style. Enumeration values shall be in the upper case with words separated by the underscore character _. For example: 

```scala
  private object ParseState extends Enumeration {
  type ParseState = Value

  val PREFIX,
      TRIM_BEFORE_SIGN,
      SIGN,
      TRIM_BEFORE_VALUE,
      VALUE,
      VALUE_FRACTIONAL_PART,
      TRIM_BEFORE_UNIT,
      UNIT_BEGIN,
      UNIT_SUFFIX,
      UNIT_END = Value
}
```

yaooqinn (approved):

ulysses-you (approved):

PR #1330, opened 2021-11-04
fix tool cleaner process bug

zwangsheng:

When using the tool to help clean up spark on K8s residual cache files, I encountered unreported sleep conditions.
After analysis, it was found that the SSD mounted when `needToDeepClean` was run encountered an incorrect catch exception.
Now, `Try Catch` is added to ensure the normal operation of the function, and deep cleaning is performed by default if a problem occurs. In this case, the disk space still overruns after the cleaning opportunity is missed.

### DIFF
--- a/tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
+++ b/tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
@@ -154,13 +154,20 @@ object KubernetesSparkBlockCleaner extends Logging {
   import scala.sys.process._
 
   private def needToDeepClean(dir: String): Boolean = {
-    val used = (s"df $dir" #| s"grep $dir").!!
-      .split(" ").filter(_.endsWith("%")) {
-      0
-    }.replace("%", "")
-    info(s"$dir now used $used% space")
+    try {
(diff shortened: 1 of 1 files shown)


zwangsheng:
cc @ulysses-you

tools/spark-block-cleaner/src/main/scala/org/apache/kyuubi/tools/KubernetesSparkBlockCleaner.scala
```diff
@@ -154,13 +154,20 @@ object KubernetesSparkBlockCleaner extends Logging {
...
+    try {
+      val used = (s"df $dir" #| s"grep $dir").!!
+        .split(" ").filter(_.endsWith("%")) {
+        0
+      }.replace("%", "")
+      info(s"$dir now used $used% space")
+
+      used.toInt > (100 - freeSpaceThreshold)
+    } catch {
+      case e: RuntimeException =>
```
ulysses-you, line 166: how about catch `NonFatal` ?

zwangsheng: Ok, i will do it this

ulysses-you:
thanks, merging to master/branch-1.3

ulysses-you (approved):

PR #1331, opened 2021-11-04
[TEST] Refactor test classes to match multi frontend protocol

pan3793:

As Kyuubi supports REST frontend protocol now, and MySQL frontend protocol is coming soon, Thrift is not the only one, this PR refacotor the test classes to match the changes.


kyuubi-server/src/test/scala/org/apache/kyuubi/WithSecuredDFSService.scala
```diff
@@ -59,5 +59,5 @@ trait WithSecuredDFSService extends KerberizedTestHelper {
     super.afterAll()
   }
 
-  def getHadoopConf(): Configuration = miniDFSService.getHadoopConf
+  def getHadoopConf: Configuration = miniDFSService.getHadoopConf
```
pan3793, line 62: Remove parentheses because the function has no side effects.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/RestFrontendService.scala
```diff
@@ -33,7 +33,7 @@ import org.apache.kyuubi.service.{AbstractFrontendService, Serverable, Service}
  * A frontend service based on RESTful api via HTTP protocol.
  * Note: Currently, it only be used in the Kyuubi Server side.
  */
-private[server] class RestFrontendService(override val serverable: Serverable)
+class RestFrontendService(override val serverable: Serverable)
```
pan3793, line 36: Maybe `RESTfulFrontendService` is a better name? cc @yanghua

yanghua: sounds good. +1

pan3793:
any thoughts? @yaooqinn

yaooqinn:
https://docs.github.com/en/rest

yaooqinn:
https://developer.twitter.com/en/docs

yaooqinn:
https://developer.paypal.com/docs/api/overview/#make-rest-api-calls

pan3793:
Sound like we should change `rest` to `REST`?

yaooqinn:
https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm

yaooqinn:
> Sound like we should change `rest` to `REST`?

the `rest` is ok to me from the given URLs, they are mostly contains lower case `rest`

yaooqinn:
I can not find any supporting evidence from your PR description so I did some research on my own. Some from the best practices of big companies and some from the original author.

pan3793:
WDYT other changes of this PR?

kyuubi-common/src/test/scala/org/apache/kyuubi/service/NoopBackendService.scala
```diff
@@ -21,6 +21,7 @@ import org.apache.kyuubi.KyuubiException
 import org.apache.kyuubi.session.{NoopSessionManager, SessionManager}
 
 class NoopBackendService extends AbstractBackendService("noop") {
+
```
yaooqinn, line 24: nit

yaooqinn (approved):

pan3793:
Thanks, merging to master

PR #1334, opened 2021-11-04
[DRAFT] Kyuubi MySQL Frontend protocol implementation draft

pan3793:
### _Why are the changes needed?_

This is a draft implementation of #1219, I'm going to break it into small PRs for review.

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.readthedocs.io/en/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #1348, opened 2021-11-08
[DEPS] Add netty and mysql jdbc driver to kyuubi-server module

pan3793:

This PR is one part of #1334.

Add netty to `kyuubi-server` module compile scope, and mysql jdbc driver to test scope, it's required for implementing MySQL frontend protocol

### DIFF
--- a/kyuubi-server/pom.xml
+++ b/kyuubi-server/pom.xml
@@ -203,6 +203,11 @@
         </dependency>
 
         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-all</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.glassfish.jersey.core</groupId>
             <artifactId>jersey-server</artifactId>
         </dependency>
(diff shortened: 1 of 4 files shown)


kyuubi-server/pom.xml
```diff
@@ -427,6 +432,12 @@
             <artifactId>bcpkix-jdk15on</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <scope>test</scope>
+        </dependency>
```
ulysses-you, line 440: do we plan to add a new module for mysql protocol ? or just mix in kyuubi-server ?

pan3793: I think things are quite complex than a new module. we also need to refactor thrift-related classes to make them more abstracted but prefer to do that refactor after adding functionality.

pan3793: The current draft implantation is mixed in `kyuubi-server`, you can get it at #1334.

PR #1356, opened 2021-11-10
Refine KyuubiStatementEvent

yaooqinn:

1. When a statement error happened, the event will be logged twice, in `setState` and `setException`, we fix it
2. We break down KyuubiStatementEvent into three parts - statement basis, status, and sessionId/user, see the improved comments
3. log the remote operation handle too


ulysses-you (approved):

ulysses-you:
thanks, merging to master

pan3793:
It seems a breaking change, will it affect users?

yaooqinn:
> It seems a breaking change, will it affect users?
yes, but this feature is still experimental so far.

pan3793:
Marked `release highlight`

PR #1359, opened 2021-11-10
Support setting zk chroot path when initialize cluster

murong00:
Users may encounter the following exception when trying to set `kyuubi.ha.zookeeper.quorum` to `localhost:2181/lakehouse` with chroot path `/lakehouse` nonexisted:
```
2021-11-10 10:56:34.510 ERROR server.KyuubiThriftBinaryFrontendService: Error starting service KyuubiServiceDiscovery
org.apache.kyuubi.KyuubiException: Failed to create namespace '/kyuubi'
        at org.apache.kyuubi.ha.client.ServiceDiscovery$.createServiceNode(ServiceDiscovery.scala:225)
        at org.apache.kyuubi.ha.client.ServiceDiscovery.start(ServiceDiscovery.scala:101)
......
Caused by: org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /kyuubi
        at org.apache.zookeeper.KeeperException.create(KeeperException.java:114)
......
```
It is wonderful to support this since zookeeper connection with chroot path is generally recommended in production environments. With this feture the znodes in zookeeper likes below:
```
[zk: localhost:2181(CONNECTED) 28] ls /lakehouse
[kyuubi, kyuubi_USER, kyuubi_USER_SPARK_SQL]
```

### DIFF
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
@@ -22,10 +22,14 @@ import javax.security.auth.login.Configuration
 
 import org.apache.curator.framework.{CuratorFramework, CuratorFrameworkFactory}
 import org.apache.curator.retry._
+import org.apache.curator.utils.ZKPaths
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.hadoop.security.token.delegation.ZKDelegationTokenSecretManager.JaasConfiguration
+import org.apache.zookeeper.CreateMode.PERSISTENT
+import org.apache.zookeeper.KeeperException
+import org.apache.zookeeper.KeeperException.NodeExistsException
 
-import org.apache.kyuubi.Logging
+import org.apache.kyuubi.{KyuubiException, Logging}
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.ha.HighAvailabilityConf._
 import org.apache.kyuubi.util.KyuubiHadoopUtils
@@ -54,6 +58,35 @@ object ZooKeeperClientProvider extends Logging {
         new BoundedExponentialBackoffRetry(baseSleepTime, maxSleepTime, maxRetries)
       case UNTIL_ELAPSED => new RetryUntilElapsed(maxSleepTime, baseSleepTime)
       case _ => new ExponentialBackoffRetry(baseSleepTime, maxRetries)
+    }
+    val chrootIndex = connectionStr.indexOf("/")
+    val chrootOption = {
+      if (chrootIndex > 0) Some(connectionStr.substring(chrootIndex))
+      else None
+    }
+    // make sure zookeeper chroot path exists
+    chrootOption.foreach { chroot =>
+      val zkConnectionForChrootCreation = connectionStr.substring(0, chrootIndex)
+      val clonedConf = KyuubiConf(false)
+      clonedConf.set(HA_ZK_QUORUM, zkConnectionForChrootCreation)
+      val zkClient = buildZookeeperClient(clonedConf)
+      zkClient.start()
+      if (zkClient.checkExists().forPath(chroot) == null) {
+        val chrootPath = ZKPaths.makePath(null, chroot)
+        try {
+          zkClient
+            .create()
+            .creatingParentsIfNeeded()
+            .withMode(PERSISTENT)
+            .forPath(chrootPath)
+        } catch {
+          case _: NodeExistsException => // do nothing
+          case e: KeeperException =>
+            throw new KyuubiException(s"Failed to create chroot path '$chrootPath'", e)
+        }
+        info(s"Created zookeeper chroot path $chrootPath")
+      }
+      zkClient.close()
     }
     val builder = CuratorFrameworkFactory.builder()
       .connectString(connectionStr)
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
@@ -22,14 +22,10 @@ import javax.security.auth.login.Configuration
 
 import org.apache.curator.framework.{CuratorFramework, CuratorFrameworkFactory}
 import org.apache.curator.retry._
-import org.apache.curator.utils.ZKPaths
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.hadoop.security.token.delegation.ZKDelegationTokenSecretManager.JaasConfiguration
-import org.apache.zookeeper.CreateMode.PERSISTENT
-import org.apache.zookeeper.KeeperException
-import org.apache.zookeeper.KeeperException.NodeExistsException
 
-import org.apache.kyuubi.{KyuubiException, Logging}
+import org.apache.kyuubi.Logging
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.ha.HighAvailabilityConf._
 import org.apache.kyuubi.util.KyuubiHadoopUtils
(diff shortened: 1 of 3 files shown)


yaooqinn:
> localhost:2181/lakehouse

According to the syntax of JDBC URL for service discovery `jdbc:hive2://<zookeeper quorum>/<dbName>;ServiceDiscoveryMode=zookeeper;zooKeeperNameSpace=kyuubi`

The `lakehouse` is a `dbName` or part of `<zookeeper quorum>`?

murong00:
> > localhost:2181/lakehouse
> 
> According to the syntax of JDBC URL for service discovery `jdbc:hive2://<zookeeper quorum>/<dbName>;ServiceDiscoveryMode=zookeeper;zooKeeperNameSpace=kyuubi`
> 
> The `lakehouse` is a `dbName` or part of `<zookeeper quorum>`?

The latter, users can use this znode to avoid creating znodes(such as: kyuubi_USER, kyuubi_USER_SPARK_SQ) in root path (`/`).

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
```diff
@@ -55,6 +59,35 @@ object ZooKeeperClientProvider extends Logging {
...
     }
+    val chrootIndex = connectionStr.indexOf("/")
+    val chrootOption = {
+      if (chrootIndex > 0) Some(connectionStr.substring(chrootIndex))
+      else None
+    }
+    // make sure zookeeper chroot path exists
+    chrootOption.foreach { chroot =>
+      val zkConnectionForChrootCreation = connectionStr.substring(0, chrootIndex)
+      val clonedConf = KyuubiConf(false)
```
pan3793, line 70: Why create a new one instead of clone from `conf`?

pan3793: I see, it's a recursive call

yaooqinn:
Can we make this change directly to ServiceDiscovery.createServiceNode? For example, when `makePath` for the namespace `val ns = ZKPaths.makePath(null, namespace)`, we can mix the chroot part in?

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
```diff
@@ -55,6 +59,35 @@ object ZooKeeperClientProvider extends Logging {
...
+      clonedConf.set(HA_ZK_QUORUM, zkConnectionForChrootCreation)
```
pan3793, line 71: We can remove this logic and add a parameter `createChrootIfNecessary: Boolean = false` to the method to avoid checking chroot everytime.

ulysses-you:
> Can we make this change directly to ServiceDiscovery.createServiceNode?

agree, `buildZookeeperClient` is too heavy to do such things.

murong00:
> Can we make this change directly to ServiceDiscovery.createServiceNode? For example, when `makePath` for the namespace `val ns = ZKPaths.makePath(null, namespace)`, we can mix the chroot part in?

@yaooqinn @ulysses-you I have tried this, however the `_zkClient` is inited by `localhost:2181/chroot`, if we use `val ns = ZKPaths.makePath(chroot, namespace)` it will lead to exception `NoNode for /chroot/kyuubi`, use a  `_zkClient`  inited by `localhost:2181` is correct to create chroot path.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
murong00: Nice advice, I will refine this if the other reviewers accept.

murong00:
> > Can we make this change directly to ServiceDiscovery.createServiceNode?
> 
> agree, `buildZookeeperClient` is too heavy to do such things.

I think this step is needed (such as Kafka/Pulsar do) unless users create the chroot path in advance.

yaooqinn:
How about 

```github
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
@@ -40,6 +40,8 @@ object KyuubiServer extends Logging {
       zkServer.start()
       conf.set(HA_ZK_QUORUM, zkServer.getConnectString)
       conf.set(HA_ZK_AUTH_TYPE, ZooKeeperAuthTypes.NONE.toString)
+    } else {
+      // create the chroot if necessary?
     }
```

murong00:
> // create the chroot if necessary?

This seems okay to me, I will refine this later, thanks for your tips.

yaooqinn:
@murong00 Thanks for your contribution and welcome!

murong00:
@yaooqinn @pan3793 I have addressed above comments, please help to take a look again, thanks.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -42,6 +47,35 @@ object KyuubiServer extends Logging {
       zkServer.start()
       conf.set(HA_ZK_QUORUM, zkServer.getConnectString)
       conf.set(HA_ZK_AUTH_TYPE, ZooKeeperAuthTypes.NONE.toString)
+    } else {
+      // create chroot path if necessary
+      val connectionStr = conf.get(HA_ZK_QUORUM)
+      val chrootIndex = connectionStr.indexOf("/")
```
yaooqinn, line 53: `connectionStr` is a comma-separated list of servers to connect to.

```
localhost:2181/lakehouse,localhost:2181/lakehouse
localhost:2181/lakehouse,localhost:2181
localhost:2181,localhost:2181/lakehouse
```

Are we able to handle all these valid and invalid cases?

murong00: > `connectionStr` is a comma-separated list of servers to connect to.
> 
> ```
> localhost:2181/lakehouse,localhost:2181/lakehouse
> localhost:2181/lakehouse,localhost:2181
> localhost:2181,localhost:2181/lakehouse
> ```
> 
> Are we able to handle all these valid and invalid cases?

In general, the third one is recommended and used, so I think we only need to support this and we can provide an example to cover this case in parameter document.

yaooqinn: Shall we fail the invalid cases directly？

murong00: Sounds good, I will refine later.

yaooqinn: thanks~

murong00: Done.

murong00:
@yaooqinn @pan3793 PTAL, again.

ulysses-you:
thank you @murong00 ! do you have a chance to add a test ? I hope we can have a better coverage since the code about zk is important for Kyuubi.

yaooqinn (approved):

murong00:
@ulysses-you @pan3793 I have added a test to cover these changes, please take a look again, thanks.

pan3793 (approved):

ulysses-you (approved):

ulysses-you:
thanks, merging to master

pan3793:
@murong00 Seems the client does not support chroot?
```
jdbc:hive2://<zookeeper quorum>/<dbName>;ServiceDiscoveryMode=zookeeper;zooKeeperNamespace=kyuubi
```

dragon-distributed:
use:  jdbc:hive2://<zookeeper quorum>/<dbName>;ServiceDiscoveryMode=zookeeper;zooKeeperNamespace=/<chroot>/kyuubi

pan3793:
I see,  the client should use `zooKeeperNamespace=<chroot>/kyuubi`

PR #1366, opened 2021-11-12
Add a todo in zorder that we can do rebalance partitions before local sort of zorder

ulysses-you:

Rebalance partitions can resolve data skew issue before writing, so it should be good to do before we do the local sort of zorder, and also we can support dynamic partition insertion with zorder. However Spark only support this after 3.3.0, see [PR](https://github.com/apache/spark/pull/34542).

So add a todo first.

not needed


yaooqinn (approved):

ulysses-you:
thanks for review, merged to master

PR #1377, opened 2021-11-14
Support to get launch engine log asynchronously with KyuubiConnection

turboFei:

Support to get launch engine log in client side.

When open session with async mode, return the launch engine operation handle by open session response configuration.

For the KyuubiConnection, it will launch a thread to fetch engine log.

image

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
@@ -34,7 +34,7 @@ abstract class AbstractOperation(opType: OperationType, session: Session)
 
   private final val createTime = System.currentTimeMillis()
   private final val handle = OperationHandle(opType, session.protocol)
-  private final val operationTimeout: Long = {
+  protected val operationTimeout: Long = {
     session.sessionManager.getConf.get(OPERATION_IDLE_TIMEOUT)
   }
 
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
@@ -34,7 +34,7 @@ abstract class AbstractOperation(opType: OperationType, session: Session)
 
   private final val createTime = System.currentTimeMillis()
   private final val handle = OperationHandle(opType, session.protocol)
-  protected val operationTimeout: Long = {
+  private final val operationTimeout: Long = {
     session.sessionManager.getConf.get(OPERATION_IDLE_TIMEOUT)
   }
 
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
@@ -17,10 +17,13 @@
 
 package org.apache.kyuubi.operation.log
 
-import java.io.IOException
+import java.io.{BufferedReader, IOException}
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Path, Paths}
+import java.util.{ArrayList => JArrayList}
+
+import scala.collection.mutable.ListBuffer
 
 import org.apache.hive.service.rpc.thrift.{TColumn, TRow, TRowSet, TStringColumn}
 
@@ -83,6 +86,16 @@ class OperationLog(path: Path) {
   private val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
   private val reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)
 
+  private val extraReaders: ListBuffer[BufferedReader] = ListBuffer()
+
+  def addExtraLog(path: Path): Unit = synchronized {
+    try {
+      extraReaders += Files.newBufferedReader(path, StandardCharsets.UTF_8)
+    } catch {
+      case _: IOException =>
+    }
+  }
+
   /**
    * write log to the operation log file
    */
@@ -95,12 +108,25 @@ class OperationLog(path: Path) {
     }
   }
 
-  /**
-   * write a line to the operation log file
-   */
-  def writeLine(line: String): Unit = {
-    if (line != null) {
-      write(line + "\n")
+  private def readLogs(
+      reader: BufferedReader,
+      lastRows: Int,
+      maxRows: Int): (JArrayList[String], Int) = {
+    val logs = new java.util.ArrayList[String]
+    var i = 0
+    try {
+      var line: String = reader.readLine()
+      while ((i < lastRows || maxRows <= 0) && line != null) {
+        logs.add(line)
+        line = reader.readLine()
+        i += 1
+      }
+      (logs, i)
+    } catch {
+      case e: IOException =>
+        val absPath = path.toAbsolutePath
+        val opHandle = absPath.getFileName
+        throw KyuubiSQLException(s"Operation[$opHandle] log file $absPath is not found", e)
     }
   }
 
@@ -110,21 +136,14 @@ class OperationLog(path: Path) {
    * @param maxRows maximum result number can reach
    */
   def read(maxRows: Int): TRowSet = synchronized {
-    val logs = new java.util.ArrayList[String]
-    var i = 0
-    try {
-      var line: String = reader.readLine()
-      while ((i < maxRows || maxRows <= 0) && line != null) {
-        logs.add(line)
-        line = reader.readLine()
-        i += 1
-      }
-    } catch {
-      case e: IOException =>
-        val absPath = path.toAbsolutePath
-        val opHandle = absPath.getFileName
-        throw KyuubiSQLException(s"Operation[$opHandle] log file $absPath is not found", e)
+    val (logs, lines) = readLogs(reader, maxRows, maxRows)
+    var lastRows = maxRows - lines
+    for (extraReader <- extraReaders if lastRows > 0 || maxRows <= 0) {
+      val (extraLogs, extraRows) = readLogs(extraReader, lastRows, maxRows)
+      lastRows = lastRows - extraRows
+      logs.addAll(extraLogs)
     }
+
     val tColumn = TColumn.stringVal(new TStringColumn(logs, ByteBuffer.allocate(0)))
     val tRow = new TRowSet(0, new java.util.ArrayList[TRow](logs.size()))
     tRow.addToColumns(tColumn)
@@ -136,6 +155,7 @@ class OperationLog(path: Path) {
       reader.close()
       writer.close()
       Files.delete(path)
+      closeExtraReaders()
     } catch {
       case e: IOException =>
         // Printing log here may cause a deadlock. The lock order of OperationLog.write
@@ -144,6 +164,16 @@ class OperationLog(path: Path) {
         // processing at the invocations
         throw new IOException(
           s"Failed to remove corresponding log file of operation: ${path.toAbsolutePath}", e)
+    }
+  }
+
+  private def closeExtraReaders(): Unit = {
+    extraReaders.foreach { extraReader =>
+      try {
+        extraReader.close()
+      } catch {
+        case _: IOException => // for the outside log file reader, ignore it
+      }
     }
   }
 }
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
@@ -152,9 +152,9 @@ class OperationLog(path: Path) {
 
   def close(): Unit = synchronized {
     try {
+      closeExtraReaders()
       reader.close()
       writer.close()
-      closeExtraReaders()
       Files.delete(path)
     } catch {
       case e: IOException =>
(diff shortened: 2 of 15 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/service/ThriftBinaryFrontendService.scala
```diff
@@ -500,6 +501,27 @@ abstract class ThriftBinaryFrontendService(name: String)
...
+    try {
+      OperationHandle(tOperationHandle)
+    } catch {
+      case ue: UnsupportedOperationException =>
+        val handleIdentifier = HandleIdentifier(tOperationHandle.getOperationId)
+        be.sessionManager.operationManager.findOperation(handleIdentifier)
+          .filter(opHandle => OperationType.isKyuubiDefinedOperationType(opHandle.typ))
+          .getOrElse(throw ue)
+    }
+  }
```
yaooqinn, line 523: If we can map it to `UNKNOWN`, or change `OperationHandle.equals` that when typ=UNKNOWN then only compare the `id`, we can make it more simple?.

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -95,6 +95,15 @@ class OperationLog(path: Path) {
     }
   }
 
+  /**
+   * write a line to the operation log file
+   */
+  def writeLine(line: String): Unit = {
```
yaooqinn, line 101: exposing this will cause deadlock

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -95,27 +108,42 @@ class OperationLog(path: Path) {
...
-   * Read to log file line by line
-   *
-   * @param maxRows maximum result number can reach
-   */
-  def read(maxRows: Int): TRowSet = synchronized {
+  private def readLogs(
+      reader: BufferedReader,
+      lastRows: Int,
+      maxRows: Int): (JArrayList[String], Int) = {
     val logs = new java.util.ArrayList[String]
```
pan3793, line 115: `java.util.ArrayList` => `JArrayList`

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -178,6 +164,7 @@ public KyuubiConnection(String uri, Properties info) throws SQLException {
 
       // open client session
       openSession();
+      getLaunchEngineLog();
```
pan3793, line 167: Seems not required for `EmbeddedMode`, but no side effect.

pan3793 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -127,6 +155,7 @@ class OperationLog(path: Path) {
       reader.close()
       writer.close()
       Files.delete(path)
+      closeExtraReaders()
```
yaooqinn, line 158: to L157?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -686,12 +739,31 @@ private void openSession() throws SQLException {
...
-      String serverFetchSize =
-        openResp.getConfiguration().get("hive.server2.thrift.resultset.default.fetch.size");
+      String serverFetchSize = openRespConf.get("hive.server2.thrift.resultset.default.fetch.size");
       if (serverFetchSize != null) {
         fetchSize = Integer.parseInt(serverFetchSize);
       }
+
+      // Get launch engine operation handle
+      String launchEngineOpHandleGuid =
+        openRespConf.get("kyuubi.session.launch.engine.operation.handle.identifier.guid");
```
yaooqinn, line 751: kyuubi.session.launch.engine.handle.guid？

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -174,7 +175,8 @@ private[kyuubi] class EngineRef(
       }
   }
 
-  private def create(zkClient: CuratorFramework): (String, Int) = tryWithLock(zkClient) {
+  private def create(zkClient: CuratorFramework, extraEngineLog: Option[OperationLog]):
```
yaooqinn, line 178:
```scala
private def create(
  zkClient: CuratorFramework,
  extraEngineLog: Option[OperationLog]): (String, Int) 
```

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
```diff
@@ -61,4 +73,13 @@ class LaunchEngine(session: KyuubiSessionImpl, override val shouldRunAsync: Bool
 
     if (!shouldRunAsync) getBackgroundHandle.get()
   }
+
+  override def close(): Unit = {
+    OperationLog.removeCurrentOperationLog()
```
yaooqinn, line 78: can it remove the variable safely?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
```diff
@@ -37,21 +42,28 @@ class LaunchEngine(session: KyuubiSessionImpl, override val shouldRunAsync: Bool
     setState(OperationState.PENDING)
   }
 
-  override protected def afterRun(): Unit = {
-    OperationLog.removeCurrentOperationLog()
-  }
```
yaooqinn, line 42: what if we keep this?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiThriftBinaryFrontendService.scala
```diff
@@ -32,6 +36,23 @@ class KyuubiThriftBinaryFrontendService(
...
+      if (launchEngineOp.shouldRunAsync) {
+        val opHandleIdentifier = launchEngineOp.getHandle.identifier.toTHandleIdentifier
+        resp.getConfiguration.put("kyuubi.session.launch.engine.operation.handle.identifier.guid",
+          Base64.encodeBase64String(opHandleIdentifier.getGuid))
+        resp.getConfiguration.put("kyuubi.session.launch.engine.operation.handle.identifier.secret",
+          Base64.encodeBase64String(opHandleIdentifier.getSecret))
+      }
+    }
+    resp
+  }
```
yaooqinn, line 54: how about rewriting the `super.OpenSession(req)` directly here？

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
turboFei: it will close the operation log after runInternal.


Maybe the launch engine operation is still running

yaooqinn: OperationLog.removeCurrentOperationLog() does not close the log?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
```diff
@@ -37,21 +42,28 @@ class LaunchEngine(session: KyuubiSessionImpl, override val shouldRunAsync: Bool
...
+          override def run(): Unit = {
+            if (shouldRunAsync) {
+              // delay to close it for async mode to enable client to get more launch engine log
+              Thread.sleep(DEFAULT_LAUNCH_ENGINE_OPERATION_CLOSE_DELAY)
+            }
+            if (!isClosedOrCanceled) {
+              session.closeOperation(getHandle)
+            }
+          }
+        }.start()
```
yaooqinn, line 66: how about we delay deleting the launch log when the related session being closed?

yaooqinn: It gives more opportunity to for admins to check what happens

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
turboFei: seems can not, it is a threadlocal

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -126,6 +154,7 @@ class OperationLog(path: Path) {
     try {
       reader.close()
       writer.close()
+      closeExtraReaders()
```
yaooqinn, line 157: maybe line 155 is better

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -229,10 +232,12 @@ private[kyuubi] class EngineRef(
   /**
    * Get the engine ref from engine space first first or create a new one
    */
-  def getOrCreate(zkClient: CuratorFramework): (String, Int) = {
+  def getOrCreate(
+      zkClient: CuratorFramework,
+      extraEngineLog: Option[OperationLog] = None): (String, Int) = {
```
yaooqinn, line 237: can we add some comments?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
turboFei: make senses

turboFei:
thanks for your review, has addressed all the comments

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
```diff
@@ -38,21 +42,19 @@ class LaunchEngine(session: KyuubiSessionImpl, override val shouldRunAsync: Bool
   }
 
   override protected def afterRun(): Unit = {
+    if (shouldRunAsync) {
+      waitLaunchEngineComplete()
+    }
     OperationLog.removeCurrentOperationLog()
```
yaooqinn, line 48: It's safe to remove it directly from thread local to me

yaooqinn: De we rely on the thread local log to write anything else inside this operation after run?

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
```diff
@@ -34,7 +34,7 @@ abstract class AbstractOperation(opType: OperationType, session: Session)
 
   private final val createTime = System.currentTimeMillis()
   private final val handle = OperationHandle(opType, session.protocol)
-  private final val operationTimeout: Long = {
+  protected val operationTimeout: Long = {
```
yaooqinn, line 37: OPERATION_IDLE_TIMEOUT is for operation idle check, it seems to me we don't need to special handle for Laungh engine operation

yaooqinn:
just last two minor comments, otherwise LGTM

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/LaunchEngine.scala
turboFei: I think there is still remind operation log to write after this thread local is removed.
is it safe?

turboFei: for example, for async mode, the launch engine is still running in backend.

But it might run `afterRun` before the launch engine operation finished.

Then if the launch engine failed eventually, it may do not recored the error info into launch engine operation log.

yaooqinn: register and remove to thread local is just for a code path for writing. I think it's safe. If we didn't handle them in pairs, then I guess it's unsafe for potential memory leak

turboFei: but I did test for the case I mentioned.

It seems that, remove thread local directly does not matter.

yaooqinn: > But it might run afterRun before the launch engine operation finished.

Yes, in this case, we actually need to register the log in that async exec thread and remove it after. But this operation seems to have no log output in the async exec thread directly, so register/remove seem to be able to omitted as we do now.

register and remove in beforeRun/afterRun take affect in the FE's handler thread, it logs nothing in the async exec thread

turboFei: thanks

yaooqinn (approved):

turboFei:
thanks for your review, merging to master

yaooqinn:
great work

PR #1379, opened 2021-11-15
Make SESSION_ENGINE_LAUNCH_ASYNC true by default

turboFei:


yaooqinn:
this is a good choice after ASYNC is finally stable, maybe we shall do this in v1.5.0

turboFei:
sure

yaooqinn (approved):

turboFei:
thanks, merging to master

zinking:
@turboFei  can you provide a little context on the significance of `launchAsync` here ?  is it just to unblock the initialization process of session creation? 

or in other way,  do we still need async = false ?  why

PR #1389, opened 2021-11-15
Refine helm service template to support LoadBalancer type

murong00:
The helm service template is currently hard-coded with `NodePort`, we should make it configurable to support other service type such as `LoadBalancer`.

### DIFF
--- a/docker/helm/templates/kyuubi-service.yaml
+++ b/docker/helm/templates/kyuubi-service.yaml
@@ -18,13 +18,18 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: kyuubi-nodeport
+  name: kyuubi-svc
+  labels:
+    app: {{ template "kyuubi.name" . }}
+  annotations:
+  {{ toYaml .Values.service.annotations | indent 4 }}
 spec:
   ports:
-    - nodePort: {{ .Values.service.port }}
+    - name: http
+      nodePort: {{ .Values.service.port }}
       port: {{ .Values.server.bind.port }}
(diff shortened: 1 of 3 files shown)


yaooqinn:
cc @hddong

lamberken (approved):
+1

hddong:
LGTM

docker/helm/templates/kyuubi-service.yaml
```diff
@@ -18,13 +18,18 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: kyuubi-nodeport
+  name: kyuubi-svc
+  labels:
+    app: {{ template "kyuubi.name" . }}
+  annotations:
+  {{ toYaml .Values.service.annotations | indent 4 }}
```
ulysses-you, line 25: nit: two more indention, just for readable

ulysses-you (approved):

ulysses-you:
not related to this PR, but is it possible to add integration test with `minikube` and `helm` ? if so we can create a new umbrella.

docker/helm/templates/kyuubi-service.yaml
murong00: Done.

murong00:
> not related to this PR, but is it possible to add integration test with `minikube` and `helm` ? if so we can create a new umbrella.

It seems that a third-party plugin or ci can achieve this, not so sure.

ulysses-you:
thanks, merging to master

PR #1413, opened 2021-11-18
Improve usage of zookeeper cli

murong00:
This change enabled the JLine support so that the `kyuubi-zookeeper-shell` command can be easier to use with history, prompt and completion, etc. Note that the `jline` version `0.9.94` is dependent on `zookeeper` version `3.4.14` which is a lower version. BTW, users can consider upgrading the `zookeeper` version for the sake of performance improvements and new features.


pan3793:
sounds that jline is optional for zookeeper-shell?

murong00:
> sounds that jline is optional for zookeeper-shell?

The `zkCli.sh` enabled this as default for flexibility and efficiency of use.

pan3793:
The change make sense to me.
BTW, we keep using zk 3.4 client because of https://github.com/apache/incubator-kyuubi/issues/838#issuecomment-883241209

pan3793 (approved):

pan3793:
CI will fill, you need to update dependencyList, and a BSD license copy should be add into `license-binary/`

pan3793:
@yaooqinn Does `kyuubi-zookeeper-shell` too long? Maybe we can short the name

yaooqinn:
> @yaooqinn Does `kyuubi-zookeeper-shell` too long? Maybe we can short the name

I don't have a strong opinion here, what's your advice?

pan3793:
The original script names `zkCli.sh`, I suggest we rename it to `zk-cli` to match the style of other kyuubi commands

murong00:
> The original script names `zkCli.sh`, I suggest we rename it to `zk-cli` to match the style of other kyuubi commands

Seems ok to me, renamed to `kyuubi-zk-cli`. Thanks for the above tips.

pan3793:
Thanks, merging to master

kyuubi-assembly/pom.xml
```diff
@@ -94,6 +94,18 @@
...
+            <groupId>jline</groupId>
+            <artifactId>jline</artifactId>
+            <version>${jline.version}</version>
+            <exclusions>
+                <exclusion>
+                    <artifactId>junit</artifactId>
+                    <groupId>junit</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
```
pan3793, line 107: Sorry, I just notice you pin `<version>` and `<exclusions>` directly in child pom `<dependencies>` block, could you please move them to root pom `<dependencyManagement>` block? @murong00

murong00: @pan3793 Sure, refined in #1416, please help to take a look, thanks.

Issue #1417, opened 2021-11-19
[Bug] An exception occurred while executing the 'bin/kyuubi-ctl list server' command


wForget:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

An exception occurred while executing the bin/kyuubi-ctl list server command.

Error message:
```
21/11/18 17:43:54 ERROR client.ServiceDiscovery: Failed to get service node info
java.lang.ArrayIndexOutOfBoundsException: 1
    at org.apache.kyuubi.ha.client.ServiceDiscovery$.$anonfun$getServiceNodesInfo$2(ServiceDiscovery.scala:194)
    at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
    at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
    at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
    at scala.collection.TraversableLike.map(TraversableLike.scala:286)
    at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
    at scala.collection.AbstractTraversable.map(Traversable.scala:108)
    at org.apache.kyuubi.ha.client.ServiceDiscovery$.getServiceNodesInfo(ServiceDiscovery.scala:189)
    at org.apache.kyuubi.ctl.ServiceControlCli.getServiceNodes(ServiceControlCli.scala:138)
    at org.apache.kyuubi.ctl.ServiceControlCli.$anonfun$list$1(ServiceControlCli.scala:127)
    at org.apache.kyuubi.ctl.ServiceControlCli.$anonfun$list$1$adapted(ServiceControlCli.scala:122)
    at org.apache.kyuubi.ha.client.ZooKeeperClientProvider$.withZkClient(ZooKeeperClientProvider.scala:81)
    at org.apache.kyuubi.ctl.ServiceControlCli.list(ServiceControlCli.scala:122)
    at org.apache.kyuubi.ctl.ServiceControlCli.doAction(ServiceControlCli.scala:68)
[...]

PR #1418, opened 2021-11-19
[KYUUBI #1417] Parse instance information in a compatible way

wForget:

Fix bug #1417.

After fix, run the `bin/kyuubi-ctl list server` command successfully.

image

### DIFF
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
@@ -22,6 +22,8 @@ import java.nio.charset.StandardCharsets
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 
+import com.google.common.annotations.VisibleForTesting
+
 import scala.collection.JavaConverters._
 
 import org.apache.curator.framework.CuratorFramework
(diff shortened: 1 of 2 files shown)


yaooqinn (approved):

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
```diff
@@ -205,6 +204,23 @@ object ServiceDiscovery extends Logging {
...
+    if (maybeInfos.size > 0) {
+      (maybeInfos.get("hive.server2.thrift.bind.host").get,
+        maybeInfos.get("hive.server2.thrift.port").get.toInt)
+    } else {
+      val strings = instance.split(":")
+      (strings(0), strings(1).toInt)
+    }
+  }
+
+
```
ulysses-you, line 223: nit: more one line

ulysses-you (approved):

pan3793:
Also cc @cxzl25

pan3793:
Thanks, merging to master

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ServiceDiscovery.scala
cxzl25: LGTM

Hive uses regular expressions to be compatible.

https://issues.apache.org/jira/browse/HIVE-13326

org.apache.hive.jdbc.ZooKeeperHiveClientHelper#configureConnParams
```java
  private static final Pattern kvPattern = Pattern.compile("([^=;]*)=([^;]*)[;]?");


      Matcher matcher = kvPattern.matcher(dataStr);
      // If dataStr is not null and dataStr is not a KV pattern,
      // it must be the server uri added by an older version HS2
      if ((dataStr != null) && (!matcher.find())) {
        String[] split = dataStr.split(":");
        if (split.length != 2) {
          throw new ZooKeeperHiveClientException("Unable to read HiveServer2 uri from ZooKeeper: "
              + dataStr);
        }
        connParams.setHost(split[0]);
        connParams.setPort(Integer.parseInt(split[1]));
```

PR #1425, opened 2021-11-20
Remove hardcode skipTests for maven-surefire-plugin

turboFei:

So that we can transfer skipTests variable when build kyuubi binary.


ulysses-you (approved):

ulysses-you:
thanks , merging to master

pan3793:
We disable surefire in default because most modules writen in scala, surefire is for Java

PR #1428, opened 2021-11-20
Add metrics reporters conf value check

zhenjiaguo:

Configuration metrics reporter need to check that match kyuubi provided reporter.

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConf.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConf.scala
@@ -46,6 +46,8 @@ object MetricsConf {
     .stringConf
     .transform(_.toUpperCase())
     .toSequence()
+    .checkValue(_.forall(ReporterType.values.map(_.toString).contains),
+      s"the reporter type should be one or more of ${ReporterType.values.mkString(",")}")
     .createWithDefault(Seq(JSON.toString))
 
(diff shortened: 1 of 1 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConf.scala
```diff
@@ -46,6 +46,8 @@ object MetricsConf {
     .stringConf
     .transform(_.toUpperCase())
     .toSequence()
+    .checkValue(_.forall(ReporterType.values.map(_.toString).contains),
```
yaooqinn, line 49: use checkValues？

zhenjiaguo: `checkValues` is used one of pre-defined set, not a sequence, like EngineType, Either `SPARK_SQL` or `FLINK_SQL`. ReporterType is conf as sequence, like `CONSOLE,JMX`. So I don't use checkValue method.

By the way, I see a lot of sequence use `checkValue` method, maybe we should add new check method called `checkSeqValues`.

zhenjiaguo: I found it hard to implement the method `checkSeqValues`. Leave it.

yaooqinn (approved):

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #1438, opened 2021-11-23
Respect kyuubi.authentication

ulysses-you:

To respect the [docs](https://kyuubi.apache.org/docs/latest/security/authentication.html#using-kerberos). We should do kinit if user specify the `kyuubi.anthentication`. Otherwise the user of ugi will be the linux user.

And if we start kyuubi with no anthentication, some warn logging will be print:
```
2021-11-22 13:20:56.028 WARN credentials.HadoopCredentialsManager: Service hadoopfs does not require a token. Check your configuration to see if security is disabled or not.
2021-11-22 13:20:56.241 WARN credentials.HadoopCredentialsManager: Service hive does not require a token. Check your configuration to see if security is disabled or not.
2021-11-22 13:20:56.245 WARN credentials.HadoopCredentialsManager: No delegation token is required by services.
```

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KinitAuxiliaryService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KinitAuxiliaryService.scala
@@ -23,6 +23,7 @@ import org.apache.hadoop.security.UserGroupInformation
 
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.service.AbstractService
+import org.apache.kyuubi.service.authentication.AuthTypes
 import org.apache.kyuubi.util.{KyuubiHadoopUtils, ThreadUtils}
 
 class KinitAuxiliaryService() extends AbstractService("KinitAuxiliaryService") {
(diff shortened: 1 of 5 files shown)


ulysses-you:
cc @yaooqinn @pan3793 @zhouyifan279

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KinitAuxiliaryService.scala
```diff
@@ -36,6 +37,12 @@ class KinitAuxiliaryService() extends AbstractService("KinitAuxiliaryService") {
   private var kinitTask: Runnable = _
 
   override def initialize(conf: KyuubiConf): Unit = {
+    val authTypes = conf.get(KyuubiConf.AUTHENTICATION_METHOD).map(AuthTypes.withName)
+    if (UserGroupInformation.isSecurityEnabled != authTypes.contains(AuthTypes.KERBEROS)) {
```
yaooqinn, line 41: How about
if (authTypes.contains(AuthTypes.KERBEROS)) {
    require(UserGroupInformation.isSecurityEnabled, "msg")
}

yaooqinn, line 41: How about
if (authTypes.contains(AuthTypes.KERBEROS)) {
    require(UserGroupInformation.isSecurityEnabled, "msg")
}

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -36,7 +37,9 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
   val credentialsManager = new HadoopCredentialsManager()
 
   override def initialize(conf: KyuubiConf): Unit = {
-    addService(credentialsManager)
+    if (UserGroupInformation.isSecurityEnabled) {
+      addService(credentialsManager)
+    }
```
ulysses-you, line 42: cc @zhouyifan279

zhouyifan279: This `if` statement should be placed in `HadoopCredentialsManager#initialize` like how we do in `KinitAuxiliaryService#initialize`. 
Otherwise, we will use an uninitialized `HadoopCredentialsManager`.

PR #1442, opened 2021-11-24
Fix encode MySQLHandshakePacket UT

pan3793:

Previous the UT depends on `KYUUBI_VERSION`, so bump version casue it to fail.

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/server/mysql/authentication/MySQLAuthPacketSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/server/mysql/authentication/MySQLAuthPacketSuite.scala
@@ -31,15 +31,17 @@ class MySQLAuthPacketSuite extends KyuubiFunSuite with MySQLCodecHelper {
   }
 
   test("encode MySQLHandshakePacket") {
-    val packet = MySQLHandshakePacket(2, authPluginData)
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/test/scala/org/apache/kyuubi/server/mysql/authentication/MySQLAuthPacketSuite.scala
```diff
@@ -31,15 +31,17 @@ class MySQLAuthPacketSuite extends KyuubiFunSuite with MySQLCodecHelper {
   }
 
   test("encode MySQLHandshakePacket") {
-    val packet = MySQLHandshakePacket(2, authPluginData)
+    val packet = new MySQLHandshakePacket(2, authPluginData) {
+      override def serverVersion: String = "5.7.22-Kyuubi-Server v1.4.0-incubating"
```
turboFei, line 35: should we use `dummy` here?

```
      override def serverVersion: String = "dummy"
```

pan3793: Make sense

turboFei (approved):

yaooqinn (approved):

turboFei:
thanks, merged to master

yaooqinn:
we need  in branch-1.4 as it will also have new versions

Issue #1396, opened 2021-11-16
[FEATURE] Need Scripts Help Binary Packages Build Images


zwangsheng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When users use official binary packages, there is no script to assist in building docker images. Manual docker build is required. We can provide a `docker-image-tool.sh` similar to Spark.

### Motivation

Add `KYUUBI_HOME/bin/docker-image-tool.sh`

### Describe the solution

We can provide a `docker-image-tool.sh` similar to Spark.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1447, opened 2021-11-24
[KYUUBI #1396][K8S] for test

ulysses-you:


zwangsheng:
Still fail about using new Dockerfile

PR #1459, opened 2021-11-28
Kyuubi Server Event Test change

zhenjiaguo:

Use regular matching to replace jackson Json parsing in kyuubi server event info test. Because the Json string is queried by SparkSQL, the special characters will be converted. This causes the configuration of environment variables to be read again and cannot be parsed into JSON.
For example, the environment variable contains the variable `ENV_VAR="\"string\""`, which becomes `ENV_VAR=""string""` after being read by SparkSQL, which cannot be converted into JSON.


ulysses-you:
@zhenjiaguo  please resolve the conflict

ulysses-you:
thanks, merging to master

PR #1462, opened 2021-11-29
[DOC] Use relative paths instead of absolute paths

zwangsheng:

1. Use relative paths instead of absolute paths, avoid jumping to the latest document page.
2. fix `setting` to `settings`

### DIFF
--- a/docs/quick_start/quick_start_with_datagrip.md
+++ b/docs/quick_start/quick_start_with_datagrip.md
@@ -23,7 +23,7 @@
 ### Get DataGrip And Install
 Please go to [Download DataGrip](https://www.jetbrains.com/datagrip/download) to get and install an appropriate version for yourself.
 ### Get Kyuubi Started
-[Get kyuubi server started](https://kyuubi.apache.org/docs/latest/quick_start/quick_start.html) before you try DataGrip with kyuubi.
+[Get kyuubi server started](../quick_start/quick_start.md) before you try DataGrip with kyuubi.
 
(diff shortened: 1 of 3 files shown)


docs/quick_start/quick_start_with_datagrip.md
```diff
@@ -23,7 +23,7 @@
 ### Get DataGrip And Install
 Please go to [Download DataGrip](https://www.jetbrains.com/datagrip/download) to get and install an appropriate version for yourself.
 ### Get Kyuubi Started
-[Get kyuubi server started](https://kyuubi.apache.org/docs/latest/quick_start/quick_start.html) before you try DataGrip with kyuubi.
+[Get kyuubi server started](../quick_start/quick_start.md) before you try DataGrip with kyuubi.
```
ulysses-you, line 26: `./quick_start.html` ?

zwangsheng: We currently do not have a uniform suffix, using `. Md` works without `#` behind.

ulysses-you: oh sorry, I mean `./quick_start.md`

zwangsheng: Now Kyuubi Doc is having a problem with 404 when it needs to specify a title to jump to if the suffix is `.md`.
So, do we need to unify the suffix, and if we do, it's just `.html`.

ulysses-you:
thanks, merging to master

PR #1463, opened 2021-11-29
Enabled to preserve symbolic links when copying dist dir

turboFei:

Before, the soft symbolic links are not reserved when making copy.

`man cp`
```
 -R    If source_file designates a directory, cp copies the directory and the entire subtree connected at that point.  If the source_file ends in
           a /, the contents of the directory are copied rather than the directory itself.  This option also causes symbolic links to be copied,
           rather than indirected through, and for cp to create special files rather than copying them as normal files.  Created directories have the
           same mode as the corresponding source directory, unmodified by the process' umask.

           In -R mode, cp will continue copying even if errors are detected.

           Note that cp copies hard-linked files as separate files.  If you need to preserve hard links, consider using tar(1), cpio(1), or pax(1)
           instead.

```

### DIFF
--- a/build/dist
+++ b/build/dist
@@ -244,7 +244,7 @@ if [[ "$MAKE_TGZ" == "true" ]]; then
   TARDIR_NAME="apache-kyuubi-$VERSION-bin$SUFFIX"
   TARDIR="$KYUUBI_HOME/$TARDIR_NAME"
   rm -rf "$TARDIR"
-  cp -r "$DISTDIR" "$TARDIR"
+  cp -R "$DISTDIR" "$TARDIR"
   tar czf "$TARDIR_NAME.tgz" -C "$KYUUBI_HOME" "$TARDIR_NAME"
   rm -rf "$TARDIR"
   echo "The Kyuubi tarball $TARDIR_NAME.tgz is successfully generated in $KYUUBI_HOME."


build/dist
```diff
@@ -245,7 +245,7 @@ if [[ "$MAKE_TGZ" == "true" ]]; then
   TARDIR="$KYUUBI_HOME/$TARDIR_NAME"
   rm -rf "$TARDIR"
   cp -r "$DISTDIR" "$TARDIR"
-  tar czf "$TARDIR_NAME.tgz" -C "$KYUUBI_HOME" "$TARDIR_NAME"
+  tar zcf "$TARDIR_NAME.tgz" -C "$KYUUBI_HOME" "$TARDIR_NAME"
```
pan3793, line 248: What's the difference?

turboFei: I am looking into it, and will test it locally.

pan3793: Does the different behavior also affected by BSD and GNU? Because macOS use BSD tar but Linux use GNU tar.

turboFei: maybe, I tried the tar command manually and it looks fine.

pan3793 (approved):

turboFei:
merged to master, thanks

PR #1466, opened 2021-11-29
Unify the guava version for kyuubi server and kyuubi beeline

turboFei:

Unify the guava version between kyuubi server and kyuubi beeline.

Make local testing.
- open kyuubi connection
- show tables
- select from table


yaooqinn (approved):

turboFei:
merged to master, thanks

PR #1471, opened 2021-11-30
Improve log if user specify the spark principal and keytab but not same with proxy user

ulysses-you:

Add the warning log instead of hiding the behavior implicitly.


yaooqinn (approved):

ulysses-you:
thanks, merging to master

Issue #1473, opened 2021-11-30
[Bug] Not graceful exit when engine idle


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

It is not graceful to exit when engine is idle, which may cause am to try again.
The current workaround is to configure `spark.yarn.maxAppAttempts`.
```
21/11/30 18:35:00 INFO yarn.ApplicationMaster: Final app status: FAILED, exitCode: 16, (reason: Shutdown hook called before final status was reported.)
```

### Affects Version(s)

master

### Kyuubi Engine Log Output

```logtalk
21/11/30 18:34:50 INFO session.SparkSQLSessionManager: Idled for more than 1800000 ms, terminating
21/11/30 18:34:50 INFO handler.ContextHandler: Stopped o.s.j.s.ServletContextHandler@4a134605{/kyuubi,null,STOPPED,@Spark}
21/11/30 18:34:50 INFO handler.ContextHandler: Stopped o.s.j.s.ServletContextHandler@6d44ce3f{/kyuubi/json,null,STOPPED,@Spark}
21/11/30 18:34:50 INFO handler.ContextHandler: Stopped o.s.j.s.ServletContextHandler@671a6a5{/kyuubi/session,null,STOPPED,@Spark}
21/11/30 18:34:50 INFO handler.ContextHandler: Stopped o.s.j.s.ServletContextHandler@39db6d9f{/kyuubi/session/json,null,STOPPED,@Spark}
21/11/30 18:34:50 INFO spark.SparkSQLEngine: Service: [SparkThriftBinaryFrontendService] is stopping.
21/11/30 18:34:50 INFO spark.SparkThriftBinaryFrontendService: SparkThriftBinaryFrontendService has stopped
21/11/30 18:34:50 INFO spark.SparkThriftBinaryFrontendService: Service: [EngineServiceDiscovery] is stopping.

21/11/30 18:34:50 INFO client.EngineServiceDiscovery: Service[EngineServiceDiscovery] is stopped.
21/11/30 18:34:50 INFO spark.SparkThriftBinaryFrontendService: Service[SparkThriftBinaryFrontendService] is stopped.
[...]

PR #1474, opened 2021-11-30
[KYUUBI #1473] Exit gracefully when engine idle

cxzl25:
https://github.com/apache/incubator-kyuubi/issues/1473

It is not graceful to exit when engine is idle, which may cause AM to try again.
The current workaround is to configure `spark.yarn.maxAppAttempts`.

```
yarn.ApplicationMaster: Final app status: FAILED, exitCode: 16, (reason: Shutdown hook called before final status was reported.)
```

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -55,10 +55,8 @@ case class SparkSQLEngine(
     // Start engine self-terminating checker after all services are ready and it can be reached by
     // all servers in engine spaces.
     backendService.sessionManager.startTerminatingChecker(() => {
-      currentEngine match {
-        case Some(engine) =>
-          engine.stop()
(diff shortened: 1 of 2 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -54,7 +54,12 @@ case class SparkSQLEngine(
     super.start()
     // Start engine self-terminating checker after all services are ready and it can be reached by
     // all servers in engine spaces.
-    backendService.sessionManager.startTerminatingChecker()
+    backendService.sessionManager.startTerminatingChecker(() => {
+      currentEngine match {
```
ulysses-you, line 58: if we go into here that means the `currentEngine` has already initialized. I think we can do:
```java
assert(currentEngine.isDefined)
currentEngine.get.stop()
```

cxzl25:
# current
image
# fix
image

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #1458, opened 2021-11-26
[Bug] Delta lake table columns won't show up in DBeaver


mmuru:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Please, see the details in the below discussion
````
https://github.com/apache/incubator-kyuubi/discussions/1189
````

### Affects Version(s)

1.3.0

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yaooqinn:
@mmuru thanks for reporting this

nousot-cloud-guy:
@mmuru @yaooqinn This doesn't appear to be working for me. I had to make [this](https://github.com/nousot/kyuubi/commit/f7e4675a77a0d220e4b73dd5e54e4dba751c781e) change in order for table schemas to correctly show up in DBeaver. Without it, I get a "table not found" error when listing columns. Do I have a misconfiguration somewhere or is this a bug?

mmuru:
It works only for "default" schema. For a custom schema, your fix should
work. I've made a fix but did not raise PR.

PR #1476, opened 2021-11-30
[KYUUBI #1458] Delta lake table columns won't show up in DBeaver.

mmuru:

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v2_4.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v2_4.scala
@@ -21,6 +21,7 @@ import java.util.regex.Pattern
 
 import org.apache.spark.sql.{Row, SparkSession}
 import org.apache.spark.sql.catalyst.TableIdentifier
+import org.apache.spark.sql.types.StructType
 
 class CatalogShim_v2_4 extends SparkCatalogShim {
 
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v2_4.scala
```diff
@@ -125,7 +126,12 @@ class CatalogShim_v2_4 extends SparkCatalogShim {
     databases.flatMap { db =>
       val identifiers = catalog.listTables(db, tablePattern, includeLocalTempViews = true)
       catalog.getTablesByName(identifiers).flatMap { t =>
-        t.schema.zipWithIndex.filter(f => columnPattern.matcher(f._1.name).matches())
+        var tableSchema: StructType = null;
+        tableSchema = t.schema
+        if (t.provider.getOrElse("") == "delta") {
+          tableSchema = spark.table(t.identifier.table).schema
+        }
+        tableSchema.zipWithIndex.filter(f => columnPattern.matcher(f._1.name).matches())
```
yaooqinn, line 134: How about

``` scala
val tableSchema = if (t.provider.getOrElse("") == "delta") {
  spark.table(t.identifier.table).schema
} else {
  t.schema
}
```

BTW, is 'delta' here case-sensitive?

mmuru: Yes, I changed it.  No, 'delta' is always lower case so I added equalsIgnoreCase.

yaooqinn (approved):

yaooqinn:
please run `dev/reformat` https://github.com/apache/incubator-kyuubi/blob/master/dev/reformat to fix style issue

mmuru:
@yaooqinn: Thanks for your review feedback.

yaooqinn (approved):

pan3793 (approved):

pan3793:
Thanks, merging to master

Issue #1393, opened 2021-11-16
[Umbrella] Add more metrics similar to hive server2


ulysses-you:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the proposal

[Cloudera hive server2 metrics](https://docs.cloudera.com/documentation/enterprise/latest/topics/cm_metrics_hiveserver2.html).

It should be helpful if we can add more metrics based on `kyuubi-metrics` module:
- avg rpc time for each operation on Kyuubi server
- current connection and opreation number
- etc ...

### Task list

TODO

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zhenjiaguo:
LGTM. This is very helpful for us comprehensive evaluation of kyuubi.

yaooqinn:
@zhenjiaguo would you like to take some of the subtask or maybe start with which ones we can add

zhenjiaguo:
Yes, I'm very happy to take some subtask. Just assign the task to me if I can help.

ulysses-you:
thank you @zhenjiaguo ! also cc @zhouyifan279

PR #1485, opened 2021-12-02
[KYUUBI #1393][TASK-2] Add operation number metric

zhenjiaguo:

Current connection number is already had. I add current operation number.

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
@@ -39,4 +39,10 @@ object MetricsConstants {
   final val ENGINE_FAIL: String = ENGINE + "failed"
   final val ENGINE_TIMEOUT: String = ENGINE + "timeout"
   final val ENGINE_TOTAL: String = ENGINE + "total"
+
+  final private val OPERATION = KYUUBI + "operation."
+  final val OPERATION_OPEN: String = OPERATION + "opened"
+  final val OPERATION_FAIL: String = OPERATION + "failed"
+  final val OPERATION_TOTAL: String = OPERATION + "total"
+
 }
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
@@ -30,11 +30,6 @@ object MetricsConstants {
   final val CONN_FAIL: String = CONN + "failed"
   final val CONN_TOTAL: String = CONN + "total"
 
-  final private val STATEMENT = KYUUBI + "statement."
-  final val STATEMENT_OPEN: String = STATEMENT + "opened"
-  final val STATEMENT_FAIL: String = STATEMENT + "failed"
-  final val STATEMENT_TOTAL: String = STATEMENT + "total"
-
   final private val ENGINE = KYUUBI + "engine."
   final val ENGINE_FAIL: String = ENGINE + "failed"
   final val ENGINE_TIMEOUT: String = ENGINE + "timeout"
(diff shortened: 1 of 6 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
```diff
@@ -39,4 +39,10 @@ object MetricsConstants {
   final val ENGINE_FAIL: String = ENGINE + "failed"
   final val ENGINE_TIMEOUT: String = ENGINE + "timeout"
   final val ENGINE_TOTAL: String = ENGINE + "total"
+
+  final private val OPERATION = KYUUBI + "operation."
```
yaooqinn, line 43: The current statement metrics is a subset of operation ones after this PR, will this be redundant?

yaooqinn: maybe we shall unify them

zhenjiaguo: Yes, operation metrics contains statement metrics. I will try to unify them.

kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
```diff
@@ -30,13 +30,14 @@ object MetricsConstants {
   final val CONN_FAIL: String = CONN + "failed"
   final val CONN_TOTAL: String = CONN + "total"
 
-  final private val STATEMENT = KYUUBI + "statement."
```
yaooqinn, line 33: is it safe to remove existing Metrics items directly？ cc @pan3793

zhenjiaguo: Original metric item `kyuubi.statement.total`, `kyuubi.statement.opened`, `kyuubi.statement.failed.${errorType}`, will be gone. It will be replaced with `kyuubi.operation.total.ExecuteStatement`, `kyuubi.operation.opened.ExecuteStatement`, `kyuubi.operation.failed.ExecuteStatement.${errorType}`.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
```diff
@@ -36,6 +36,14 @@ import org.apache.kyuubi.util.ThriftUtils
 abstract class KyuubiOperation(opType: OperationType, session: Session)
   extends AbstractOperation(opType, session) {
 
+  private val operationClassName = getClass.getSimpleName
```
yaooqinn, line 39: use opType?

yaooqinn (approved):
LGTM, only one comment

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
zhenjiaguo: OK, I will use `private val opTypeName = statement.toLowerCase` to replace it.

zhenjiaguo: I also find [LAUNCH_ENGINE](https://github.com/apache/incubator-kyuubi/blob/master/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationType.scala#L24-L26) maybe redundant. I see no anywhere else use it.

yaooqinn: The class name could be inconsistent during development, opType can be more traceable 
we can remove LAUNCH_ENGINE

yaooqinn:
thanks, merged to master

PR #1487, opened 2021-12-03
Correct the KEYS link and add script to generate vote and announcement

turboFei:

Correct the KEYS link in release template.


yaooqinn (approved):

build/release/script/announce.sh
```diff
@@ -15,17 +17,31 @@
...
-Title: [ANNOUNCE] Apache Kyuubi (Incubating) released {release_version}
+set -o pipefail
+set -e
 
-Content:
-Hi all,
+RELEASE_DIR="$(cd "$(dirname "$0")"/..; pwd)"
+
+######### Please modify the variables ##########
+release_version=v1.4.0-incubating
```
pan3793, line 26: How to change it to use env var?

turboFei: I will leave them be empty by defaults.

turboFei: it is not convenient to use env var, because they are specified.

build/release/script/dev_kyuubi_vote.sh
```diff
@@ -0,0 +1,102 @@
...
+The commit list since the ${prev_release_rc_no}:
+https://github.com/apache/incubator-kyuubi/compare/${prev_release_rc_tag}...${release_rc_tag}
+"
+fi
+
+RELEASE_TEMP_DIR=${RELEASE_DIR}/tmp
+mkdir -p ${RELEASE_TEMP_DIR}
+DEV_VOTE=${RELEASE_TEMP_DIR}/${release_rc_tag}_dev_vote.temp
+
+cat <<EOF >${DEV_VOTE}
```
pan3793, line 63: Does the space after `EOF` matter? If not, change to `cat <<EOF > ${DEV_VOTE}`.

BTW, I only saw `cat > ${DEV_VOTE} <<EOF` before, not sure which style is popular

build/release/script/announce.sh
turboFei: I prefer that the release manager modify the variable value in this script

build/release/script/dev_kyuubi_vote.sh
turboFei: thanks, it looks better, addreesed

build/release/script/announce.sh
turboFei: have refactored it

build/release/script/dev_kyuubi_vote.sh
```diff
@@ -0,0 +1,110 @@
...
+#
+
+set -o pipefail
+set -e
+
+RELEASE_DIR="$(cd "$(dirname "$0")"/..; pwd)"
+
+######### Please modify the variables ##########
+# release version, e.g. 1.4.0-incubating
+release_version=${release_version-""}
```
pan3793, line 27: `RELEASE_VERSION=${RELEASE_VERSION:?"RELEASE_VERSION is required"}`

turboFei: thanks, fixed

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #1488, opened 2021-12-03
Add JDBC mode to store engine events.

zhang1002:
Here have some problems should be noted:
1. we need to use third-party dependency:mysql-connection-#{version}.jar and should copy to external/engines/spark/ by manual
2. the jdbc url that you set in conf should contains username and password


externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -167,6 +167,7 @@
                     <shadedArtifactAttached>false</shadedArtifactAttached>
                     <artifactSet>
                         <includes>
+                            <include>com.alibaba:druid</include>
```
pan3793, line 170: HiKariCP should the better choice, evaluated from popularity, performance, and has native support of dropwizard and prometheus.  
https://github.com/apache/dolphinscheduler/issues/6881

externals/kyuubi-spark-sql-engine/src/main/resources/engine-event.sql
```diff
@@ -0,0 +1,58 @@
+CREATE TABLE IF NOT EXISTS `engine_event_summary` (
+  `cluster` varchar(50) DEFAULT NULL COMMENT 'kyuubiServe集群名称，kyuubi展示服务可以管理多个集群的kyuubiServer',
```
pan3793, line 2: Please using English in Kyuubi codebase.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JdbcEventLogger.scala
```diff
@@ -0,0 +1,284 @@
...
+ */
+
+package org.apache.kyuubi.engine.spark.events
+
+import java.io._
+import java.sql.{PreparedStatement, SQLException, SQLIntegrityConstraintViolationException, Statement}
+import java.util.concurrent.TimeUnit
+
+import com.alibaba.druid.pool.DruidPooledConnection
+import com.google.common.cache.{Cache, CacheBuilder}
```
pan3793, line 25: Please ensure your are using the stable API of Guava, because the Kyuubi Spark SQL engine will pick the Guava from Spark runtime. Generally, it should be compatible with Guava 11, 14, 19, and not be marked @Beta after Guava 21.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JdbcEventLogger.scala
```diff
@@ -0,0 +1,284 @@
...
+      while ({tmp = reader.readLine(); tmp} != null) {
+        sqlStringBuffer.append(" ")
+          .append(tmp)
+      }
+    } catch {
+      case e: Exception =>
+        error(s"Loading the file named engine-event failed. $e")
+    } finally {
+      in.close()
+      reader.close()
```
pan3793, line 70: close `reader` will close `in`

externals/kyuubi-spark-sql-engine/pom.xml
zhang1002: i think in this situation, Druid can meet our requirements.
even though HiKariCP is springboot‘s default pool.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/JdbcEventLogger.scala
zhang1002: good.
done.

externals/kyuubi-spark-sql-engine/src/main/resources/engine-event.sql
zhang1002: done

externals/kyuubi-spark-sql-engine/pom.xml
pan3793: > even though HiKariCP is springboot‘s default pool.

It's doesn't matter, the metrics is matter, both kyuubi and spring use dropwizard, and HiKariCP already in Spark runtime classpath

ulysses-you: We can list the reason why we choose `Druid`, otherwise the `HiKariCP` seems enough.

zhang1002: ok, wait

zhang1002: done
use Hikari

Issue #1480, opened 2021-12-01
[FEATURE] Provides a base-image mirror configuration method for Spark instead of local copy


zwangsheng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Provides a base-image mirror configuration method for Spark instead of local copy.

### Motivation

Add Spark to Kyuubi by using the Spark image as `BASE_IMAGE`.

### Describe the solution

Provides a base-image mirror configuration method for Spark instead of local copy.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1495, opened 2021-12-06
[KYUUBI #1480] Fix bug in docker-image-tool & Provider image way to config SPARK instead of local copy

zwangsheng:

1. fix when using `-t` may cause build fail, because tag invalid reference format
2. add new option `-S`; Declare SPARK_HOME in Docker Image. When you configured -S, you need to provide an image
  with Spark as BASE_IMAGE.

### DIFF
--- a/bin/docker-image-tool.sh
+++ b/bin/docker-image-tool.sh
@@ -115,14 +115,15 @@ function build {
     KYUUBI_ROOT="$CTX_DIR/base"
   fi
 
-  # cp spark for kyuubi as submit client
-  # if user set -s(spark-provider), use if
-  # else use builtin spark
+  # mkdir spark-binary to cache spark
+  # clean cache if spark-binary exists
   if [[ ! -d "$KYUUBI_ROOT/spark-binary" ]]; then
     mkdir "$KYUUBI_ROOT/spark-binary"
   else
     rm -rf "$KYUUBI_ROOT/spark-binary/*"
   fi
+  # if with spark image means, kyuubi won't copy spark from local.
+  # In this case, we just pass SPARK_HOME without copy.
   if [[ "${WITHSPARKIMAGE}" != "false" ]]; then
     BUILD_ARGS+=(--build-arg spark_home=$SPARK_HOME)
   else
--- a/bin/docker-image-tool.sh
+++ b/bin/docker-image-tool.sh
@@ -124,10 +124,12 @@ function build {
   else
     rm -rf "$KYUUBI_ROOT/spark-binary/*"
   fi
-  # if with spark image means, kyuubi won't copy spark from local.
-  # In this case, we just pass SPARK_HOME without copy.
-  if [[ "${SPARK_PROVIDED}" != "false" ]]; then
-    BUILD_ARGS+=(--build-arg spark_home=$SPARK_HOME)
+
+  # If SPARK_HOME_IN_DOCKER configured,
+  # Kyuubi won't copy local spark into docker image.
+  # Use SPARK_HOME_IN_DOCKER as SPARK_HOME in docker image.
+  if [[ -n "${SPARK_HOME_IN_DOCKER}" ]]; then
+    BUILD_ARGS+=(--build-arg spark_home_in_docker=$SPARK_HOME_IN_DOCKER)
     BUILD_ARGS+=(--build-arg spark_provided="spark_provided")
   else
     if [[ ! -d "$SPARK_HOME" ]]; then
@@ -232,8 +234,8 @@ NOCACHEARG=
 BUILD_PARAMS=
 KYUUBI_UID=
 CROSS_BUILD="false"
-SPARK_PROVIDED="false"
-while getopts f:r:t:Xnb:u:s:p option
+SPARK_HOME_IN_DOCKER=
+while getopts f:r:t:Xnb:u:s:d: option
 do
  case "${option}"
  in
@@ -245,7 +247,7 @@ do
  X) CROSS_BUILD=1;;
  u) KYUUBI_UID=${OPTARG};;
  s) SPARK_HOME=${OPTARG};;
- p) SPARK_PROVIDED=1;;
+ d) SPARK_HOME_IN_DOCKER=${OPTARG};;
  esac
 done
 
--- a/bin/docker-image-tool.sh
+++ b/bin/docker-image-tool.sh
@@ -199,6 +199,8 @@ Options:
                         be used separately for each build arg.
   -s                    Put the specified Spark into the Kyuubi image to be used as the internal SPARK_HOME
                         of the container.
+  -S                    Declare SPARK_HOME in Docker Image. When you configured -S, you need to provide an image
+                        with Spark as BASE_IMAGE.
 
 Examples:
 
@@ -219,6 +221,9 @@ Examples:
   - Build with Spark placed "/path/spark"
     $0 -s /path/spark build
 
+  - Build with Spark Image myrepo/spark:3.1.0
+    $0 -S /opt/spark -b BASE_IMAGE=myrepo/spark:3.1.0 build
+
 EOF
 }
 
@@ -235,7 +240,7 @@ BUILD_PARAMS=
 KYUUBI_UID=
 CROSS_BUILD="false"
 SPARK_HOME_IN_DOCKER=
-while getopts f:r:t:Xnb:u:s:d: option
+while getopts f:r:t:Xnb:u:s:S: option
 do
  case "${option}"
  in
@@ -247,7 +252,7 @@ do
  X) CROSS_BUILD=1;;
  u) KYUUBI_UID=${OPTARG};;
  s) SPARK_HOME=${OPTARG};;
- d) SPARK_HOME_IN_DOCKER=${OPTARG};;
+ S) SPARK_HOME_IN_DOCKER=${OPTARG};;
  esac
 done
 
--- a/bin/docker-image-tool.sh
+++ b/bin/docker-image-tool.sh
@@ -124,7 +124,7 @@ function build {
   fi
   # if with spark image means, kyuubi won't copy spark from local.
   # In this case, we just pass SPARK_HOME without copy.
-  if [[ "${WITHSPARKIMAGE}" != "false" ]]; then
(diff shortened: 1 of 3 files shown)


bin/docker-image-tool.sh
```diff
@@ -118,11 +120,17 @@ function build {
   # else use builtin spark
   if [[ ! -d "$KYUUBI_ROOT/spark-binary" ]]; then
     mkdir "$KYUUBI_ROOT/spark-binary"
+  else
+    rm -rf "$KYUUBI_ROOT/spark-binary/*"
```
ulysses-you, line 124: why we need delete it, if it is not existed

zwangsheng: The logic is if this dir exists, it is considered that the spark was used last time.  The spark version that was used last time is saved. Delete spark first and do copy or nothing according to `-w` parameters.

docker/Dockerfile
```diff
@@ -42,12 +44,12 @@ RUN set -ex && \
...
     apt install -y bash tini libc6 libpam-modules krb5-user libnss3 procps && \
     useradd -u ${kyuubi_uid} -g root kyuubi && \
-    mkdir -p ${KYUUBI_HOME} ${KYUUBI_LOG_DIR} ${KYUUBI_PID_DIR} ${KYUUBI_WORK_DIR_ROOT} ${SPARK_HOME} && \
+    mkdir -p ${KYUUBI_HOME} ${KYUUBI_LOG_DIR} ${KYUUBI_PID_DIR} ${KYUUBI_WORK_DIR_ROOT} ${SPARK_BINARY} && \
     chmod ug+rw -R ${KYUUBI_HOME} && \
     chmod a+rwx -R ${KYUUBI_WORK_DIR_ROOT} && \
     rm -rf /var/cache/apt/*
 
-COPY spark-binary ${SPARK_HOME}
+COPY spark-binary ${SPARK_BINARY}
```
ulysses-you, line 52: if user specify the `${spark_home}`, seems we don't need `COPY` the `spark-binary`. So here are tow options:

* user specify the `${spark_home}`; we just do `ENV SPARK_HOME ${spark_home}` without mkdir and `COPY` the `spark-binary`
* user doesn't  specify the `${spark_home}`; we should mkdir and `COPY` the `spark-binary` and `ENV SPARK_HOME SPARK_BINARY`

zwangsheng: However, Dockerfile seems have not ability to user `if else` to chose copy or not. As I envision it, spark-binary will have no effect on the inherent Spark in the mirror when building with the Spark image as the base image.

bin/docker-image-tool.sh
```diff
@@ -113,16 +115,23 @@ function build {
...
   fi
-  if [[ ! -d "$SPARK_HOME" ]]; then
-    error "Cannot found dir $SPARK_HOME, you must configure SPARK_HOME correct."
+  # if with spark image means, kyuubi won't copy spark from local.
+  # In this case, we just pass SPARK_HOME without copy.
+  if [[ "${WITHSPARKIMAGE}" != "false" ]]; then
+    BUILD_ARGS+=(--build-arg spark_home=$SPARK_HOME)
```
ulysses-you, line 128: can we also do `cp -r "$SPARK_HOME/" "$KYUUBI_ROOT/spark-binary/` here ? so we can make `SPARK_HOME` 
 and `SPARK_BINARY` consisent in `dockerfile`.

bin/docker-image-tool.sh
```diff
@@ -240,6 +244,7 @@ do
  X) CROSS_BUILD=1;;
  u) KYUUBI_UID=${OPTARG};;
  s) SPARK_HOME=${OPTARG};;
+ w) WITHSPARKIMAGE=1;;
```
ulysses-you, line 247: how about `SPARK_PROVIDED`

zwangsheng: Good advice, modify later.

bin/docker-image-tool.sh
zwangsheng: Here, meams kyuubi use spark which from BASE_IMAGE. Spark already in Image, we can not move here.

zwangsheng:
Local test success, please retest.

zwangsheng:
OK to test

bin/docker-image-tool.sh
ulysses-you: so this `spark_home` is the directory in docker rather than in real world ?

zwangsheng: We can think of it as having two `SPARK_HOME` on the outside and inside the container. SPARK_HOME inside the container is `/opt/spark` by default, which is the folder inside the container that you COPY to.

docker/Dockerfile
ulysses-you: cc @cfmcgrady do you have good idea ?

cfmcgrady: How about
```Dockerfile
ARG BASE_IMAGE=openjdk:8-jre-slim

ARG BUILD_ENV=copy

FROM ${BASE_IMAGE} as build_copy
ONBUILD RUN echo "copy user specify spark binary...."
ONBUILD COPY user_specify_spark_path /path_to_spark
ONBUILD ENV SPARK_HOME /path_to_spark

FROM ${BASE_IMAGE} as build_no_copy
ONBUILD RUN echo "I don't copy"
ONBUILD ARG BUILDIN_SPARK_HOME="/opt/spark"
ONBUILD ENV SPARK_HOME ${BUILDIN_SPARK_HOME}

FROM build_${BUILD_ENV}
```

- build w/ copy user specify spark binary package

```shell
docker build --build-arg BUILD_ENV=copy -t ${your_image_tag}  -f docker/Dockerfile .
```

- build w/o copy spark binary package
```shell
docker build --build-arg BUILD_ENV=no_copy --build-arg BUILDIN_SPARK_HOME="/path/to/spark_home" -t ${your_image_tag}  -f docker/Dockerfile .
```

cfmcgrady: FYI: https://stackoverflow.com/questions/31528384/conditional-copy-add-in-dockerfile

ulysses-you: ah, I like the idea using two options of `FROM` + `ONBUILD`, @zwangsheng can you try it ?

zwangsheng: Nice advice, i will try use `as build`.

zwangsheng:
pass local test, cc @cfmcgrady @ulysses-you

docker/Dockerfile
```diff
@@ -24,15 +24,25 @@
...
+ARG spark_provided="spark_builtin"
 
-ARG kyuubi_uid=10009
+FROM ${BASE_IMAGE} as builder_spark_provided
+
+FROM ${BASE_IMAGE} as builder_spark_builtin
+
+ONBUILD ENV SPARK_BINARY /opt/spark
+ONBUILD RUN mkdir -p  ${SPARK_BINARY}
+ONBUILD COPY spark-binary ${SPARK_BINARY}
```
ulysses-you, line 36: do you need `SPARK_BINARY` ? seems we can just `COPY spark-binary ${spark_home}`

zwangsheng: OK, good idea. I will advance the `ARG spark_home`.

bin/docker-image-tool.sh
```diff
@@ -113,16 +115,26 @@ function build {
...
     mkdir "$KYUUBI_ROOT/spark-binary"
+  else
+    rm -rf "$KYUUBI_ROOT/spark-binary/*"
   fi
-  if [[ ! -d "$SPARK_HOME" ]]; then
-    error "Cannot found dir $SPARK_HOME, you must configure SPARK_HOME correct."
+  # if with spark image means, kyuubi won't copy spark from local.
+  # In this case, we just pass SPARK_HOME without copy.
+  if [[ "${SPARK_PROVIDED}" != "false" ]]; then
+    BUILD_ARGS+=(--build-arg spark_home=$SPARK_HOME)
```
ulysses-you, line 130: A little confuse here. the `SPARK_HOME` means the directory inside docker container. But in general, the `SPARK_HOME` means the directory in real world.

So should the name be `SPARK_HOME_IN_DOCKER` ? and we don't need the flag `SPARK_PROVIDED` any more since we can detect `SPARK_HOME_IN_DOCKER` if existed.

Also we should document the `SPARK_HOME_IN_DOCKER`

docker/Dockerfile
```diff
@@ -24,15 +24,24 @@
 #     -t the target repo and tag name
 #     more options can be found with -h
 
-ARG BASE_IMAGE=8-jre-slim
-FROM openjdk:${BASE_IMAGE}
+ARG BASE_IMAGE=openjdk:8-jre-slim
+ARG spark_provided="spark_builtin"
+ARG spark_home="/opt/spark"
```
cfmcgrady, line 29: Does this ARG `spark_home` work? 

From the docker official documents:
> An ARG instruction goes out of scope at the end of the build stage where it was defined. To use an arg in multiple stages, each stage must include the ARG instruction.

FYI: https://docs.docker.com/engine/reference/builder/#scope

bin/docker-image-tool.sh
zwangsheng: I will try.

docker/Dockerfile
zwangsheng: Nice catch, i will check.

bin/docker-image-tool.sh
```diff
@@ -240,6 +247,7 @@ do
  X) CROSS_BUILD=1;;
  u) KYUUBI_UID=${OPTARG};;
  s) SPARK_HOME=${OPTARG};;
+ d) SPARK_HOME_IN_DOCKER=${OPTARG};;
```
ulysses-you, line 250: how about `S` ? `d` looks strange. And please update this option in `usage`

zwangsheng: fixed

ulysses-you (approved):
looks good to me, @cfmcgrady any comment ?

cfmcgrady (approved):

ulysses-you:
thanks, merging to master

zwangsheng:
Thanks for help.

PR #1513, opened 2021-12-06
Added support liveness and readiness for kyuubi pod

murong00:
It is better to support `livenessProbe` and `readinessProbe` for indicating whether the container is running.

### DIFF
--- a/docker/helm/values.yaml
+++ b/docker/helm/values.yaml
@@ -31,14 +31,14 @@ image:
 probe:
   liveness:
     enabled: true
-    failureThreshold: 10
     initialDelaySeconds: 30
     periodSeconds: 10
+    failureThreshold: 10
   readiness:
     enabled: true
-    failureThreshold: 10
     initialDelaySeconds: 30
     periodSeconds: 10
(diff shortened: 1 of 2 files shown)


yaooqinn:
cc @ulysses-you @hddong

docker/helm/values.yaml
```diff
@@ -28,6 +28,18 @@ image:
...
+  liveness:
+    enabled: true
+    failureThreshold: 10
+    initialDelaySeconds: 30
+    periodSeconds: 10
+  readiness:
+    enabled: true
+    failureThreshold: 10
+    initialDelaySeconds: 30
+    periodSeconds: 10
```
ulysses-you, line 41: can we keep the same ordering with `kyuubi-depolyment.yaml` ?
```
initialDelaySeconds
periodSeconds
failureThreshold
```

murong00: @ulysses-you Done, thanks.

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #1523, opened 2021-12-07
Refine usage about option '--namespace' in kyuubi-ctl

murong00:
The current option `-n` of command `kyuubi-ctl create` is confusing for new users. For example, when using the below command without setting `kyuubi.ha.zookeeper.namespace` it throws following exception which seems confused:
```
$ bin/kyuubi-ctl create -s localhost -p 10009 -n test
Only support expose Kyuubi server instance to another domain, but the default
namespace is [None] and specified namespace is [test]

        at org.apache.kyuubi.ctl.ServiceControlCliArguments.fail(ServiceControlCliArguments.scala:294)
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.validateCreateActionArguments(ServiceControlCliArguments.scala:195)
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.validateArguments(ServiceControlCliArguments.scala:176)
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.parse(ServiceControlCliArguments.scala:136)
        at org.apache.kyuubi.ctl.ServiceControlCliArguments.<init>(ServiceControlCliArguments.scala:36)
        ...
```
In addtion, the description of option `-n` in `kyuubi-ctl --help` is `using kyuubi-defaults/conf if absent`, however when using below without option `-n` throws like:
```
$ bin/kyuubi-ctl create -s localhost -p 10009
Zookeeper namespace is not specified
...
```
this because the default value is not copied to `cli.args` for `create` action. Users may confuse about this.

### DIFF
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
@@ -151,9 +151,7 @@ class ServiceControlCliArguments(args: Seq[String], env: Map[String, String] = s
       }
     }
 
-    // for create action, it only expose Kyuubi service instance to another domain,
-    // so we do not use namespace from default conf
-    if (arguments.action != ServiceControlAction.CREATE && arguments.namespace == null) {
+    if (arguments.namespace == null) {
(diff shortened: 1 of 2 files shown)


yaooqinn (approved):

yaooqinn:

Issue #1522, opened 2021-12-07
[Bug] The EngineTab in Spark UI does not display client IP correctly.


wForget:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The EngineTab in Spark UI does not display client IP correctly. it displays the ip of KyuubiServer.

![image](https://user-images.githubusercontent.com/17894939/145004003-ae3fc5b0-860d-46c2-8589-141977563449.png)

### Affects Version(s)

1.4.0

### Additional context

The Client IP is not passed into the engine. I will try to pass it in through SessionConf.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1524, opened 2021-12-07
[KYUUBI #1522] Pass the client ip to engine

wForget:

Pass the client ip to engine. For details: #1522

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
@@ -57,7 +57,8 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
       password: String,
       ipAddress: String,
       conf: Map[String, String]): SessionHandle = {
-    info(s"Opening session for $user@$ipAddress")
+    val clientIp = conf.getOrElse(CLIENT_IP_KEY, ipAddress)
+    info(s"Opening session for $user@$clientIp")
     val sparkSession =
       try {
         if (singleSparkSession) {
@@ -78,7 +79,7 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
       protocol,
       user,
       password,
-      ipAddress,
+      clientIp,
       conf,
       this,
       sparkSession)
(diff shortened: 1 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -78,7 +79,7 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
       protocol,
       user,
       password,
-      ipAddress,
```
ulysses-you, line 81: Seems the ipAddress of the kyuubi server is also useful if we deploy in ha. We can get the information which Kyuubi server opened the engine session.

kyuubi-common/src/main/scala/org/apache/kyuubi/session/package.scala
```diff
@@ -27,4 +27,6 @@ package object session {
   val METACONF_PREFIX = "metaconf:"
 
   val SPARK_PREFIX = "spark."
+
+  val CLIENT_IP_KEY = "kyuubi.clent.ip"
```
yaooqinn, line 31: client

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
wForget: > Seems the ipAddress of the kyuubi server is also useful if we deploy in ha. We can get the information which Kyuubi server opened the engine session.

Yes, but the current usage is more inclined to clientip, do we need to add an extra ip field?

yaooqinn (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
ulysses-you: We can do this in a new PR, it's better to show client ip at UI side . And just record the kyuubi server ipAddress in event log.

wForget: OK, I'll record an issue first.

yaooqinn:
thanks, merged to master

PR #1545, opened 2021-12-11
[TASK-1] Add avg rpc time for each operation

zhenjiaguo:

Subtask of #1393. Add avg rpc time for each operation on Kyuubi server.

Use MetricRegistry default histogram to record.

Screenshot of the json file generated by the json reporter:

Screenshot from 2021-12-11 09-20-35

### DIFF
--- a/dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/DataGenerator.scala
+++ b/dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/DataGenerator.scala
@@ -570,31 +570,31 @@ object DataGenerator {
       $"web_tax_percentage"       .decimal(5, 2))
     // format: on
 
-   Seq(
-     store_sales,
-     store_returns,
-     catalog_sales,
-     catalog_returns,
-     web_sales,
-     web_returns,
-     inventory,
-     catalog_page,
-     call_center,
-     customer,
-     customer_address,
-     customer_demographics,
-     household_demographics,
-     store,
-     warehouse,
-     item,
-     income_band,
-     web_site,
-     web_page,
-     date_dim,
-     time_dim,
-     ship_mode,
-     reason,
-     promotion)
+    Seq(
+      store_sales,
+      store_returns,
+      catalog_sales,
+      catalog_returns,
+      web_sales,
+      web_returns,
+      inventory,
+      catalog_page,
+      call_center,
+      customer,
+      customer_address,
+      customer_demographics,
+      household_demographics,
+      store,
+      warehouse,
+      item,
+      income_band,
+      web_site,
+      web_page,
+      date_dim,
+      time_dim,
+      ship_mode,
+      reason,
+      promotion)
   }
 
   def run(config: RunConfig): Unit = {
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
@@ -40,4 +40,5 @@ object MetricsConstants {
   final val OPERATION_FAIL: String = OPERATION + "failed"
   final val OPERATION_TOTAL: String = OPERATION + "total"
 
+  final val OPERATION_RPC_TIME: String = OPERATION + "rpc_time"
 }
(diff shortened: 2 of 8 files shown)


dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/DataGenerator.scala
```diff
@@ -570,31 +570,31 @@ object DataGenerator {
...
-     item,
-     income_band,
-     web_site,
-     web_page,
-     date_dim,
-     time_dim,
-     ship_mode,
-     reason,
-     promotion)
+    Seq(
```
zhenjiaguo, line 573: I use `dev/reformat` to reformat the code. This file format is no correct. Ignore this file when review.

kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
```diff
@@ -40,4 +40,5 @@ object MetricsConstants {
   final val OPERATION_FAIL: String = OPERATION + "failed"
   final val OPERATION_TOTAL: String = OPERATION + "total"
 
+  final val OPERATION_RPC_TIME: String = OPERATION + "rpc_time"
```
yaooqinn, line 43: how about rpc_ms or rpc_millis?

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiThriftClientCallTimeStatics.scala
```diff
@@ -0,0 +1,133 @@
...
+      timeCost += System.currentTimeMillis() - startTime
+    }
+  }
+
+  override def openSession(
+      protocol: TProtocolVersion,
+      user: String,
+      password: String,
+      configs: Map[String, String]): SessionHandle = {
+    time(client.openSession(protocol, user, password, configs))
```
yaooqinn, line 53: this is for session RPC, why do we wrap them to track as operation metrics. IIUC, they are not used acturally.

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiThriftClientCallTimeStatics.scala
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  override def closeSession(): Unit = {
```
yaooqinn, line 56: ditto

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiThriftClientCallTimeStatics.scala
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Get the time cost millis on each call of the function in the [[KyuubiThriftClient]]
+   */
+  def getTimeCostMillis: Long = {
+    timeCost
+  }
+
+  @throws[Exception]
+  private def time[T](f: => T): T = {
+    val startTime = System.currentTimeMillis()
```
ulysses-you, line 40: shall we use `System.nanoTime` ?

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiThriftClientCallTimeStatics.scala
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  override def closeSession(): Unit = {
+    client.closeSession()
+  }
+
+  override def executeStatement(
+      statement: String,
+      shouldRunAsync: Boolean,
+      queryTimeout: Long): TOperationHandle = {
```
ulysses-you, line 63: time {
  xxx
}

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiThriftClientCallTimeStatics.scala
```diff
@@ -0,0 +1,133 @@
...
+    time(client.executeStatement(statement, shouldRunAsync, queryTimeout))
```
ulysses-you, line 64:
```scala
time {
  xxx
}
```

ulysses-you: all methods should use this style

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiThriftClientCallTimeStatics.scala
zhenjiaguo: I think it is no need to use such precision.

Issue #1550, opened 2021-12-13
[FEATURE] Provide a specific user guide about connecting to kerberized kyuubi


zhouyifan279:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Provide a more specific user guide about connecting to kerberized Kyuubi

### Motivation

Many kyuubi users are not familiar with kerberos authentication mechanism and encountered troubles connecting to kerberized Kyuubi.

### Describe the solution

Add a new page under [Kyuubi usage guide | Client Documentation](https://kyuubi.apache.org/docs/latest/client/index.html#client-documentation)

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zhouyifan279:
cc @yaooqinn @pan3793 . Also cc @mmuru to verify if I missed some important steps.

PR #1552, opened 2021-12-13
[KYUUBI #1550] Provide a specific user guide about connecting to kerberized kyuubi

zhouyifan279:
Many kyuubi users are not familiar with kerberos authentication mechanism and encountered troubles connecting to kerberized Kyuubi.

Configurations are verified in #1432 


zhouyifan279:
cc @yaooqinn @pan3793 . Also cc @mmuru to verify if I missed some important steps.

docs/security/kinit.md
```diff
@@ -78,8 +79,20 @@ kyuubi\.kinit\.keytab|<div style='width: 80pt;word-wrap: break-word;white-space:
...
-it indicates that we are targeting a secured cluster, then we need to specify `kyuubi.kinit.principal` and `kyuubi.kinit.keytab` for authentication.
+When working with a kerberos-enabled Hadoop cluster, we should ensure that `hadoop.security.authentication` 
+is set to `KERBEROS` in `$HADOOP_CONF_DIR/core-site` or `$KYUUBI_HOME/conf/kyuubi-defaults.conf`. 
+Then we need to specify `kyuubi.kinit.principal` and `kyuubi.kinit.keytab` for authentication. 
+
+
+For example,
+
+```bash
+kyuubi.kinit.principal=spark/kyuubi.apache.org@KYUUBI.APACHE.ORG
```
zwangsheng, line 90: Maybe we need to add `kyuubi.authentication=KERBEROS` here.

docs/client/kerberized_kyuubi.md
```diff
@@ -0,0 +1,120 @@
...
+ - limitations under the License.
+ -->
+
+<div align=center>
+
+![](../imgs/kyuubi_logo.png)
+
+</div>
+
+# Access kerberized Kyuubi with beeline & BI tools
```
yaooqinn, line 24: Capitalize all, and do the same for all titles

docs/security/kinit.md
zhouyifan279: I think it is better to clarify this in the doc:
> Kinit auxiliary service is critical to Kyuubi kerberos authentication, but not vice versa. 
Kinit auxiliary service can also work with other authentication mode.

docs/client/kerberized_kyuubi.md
zhouyifan279: Updated

docs/client/kerberized_kyuubi.md
```diff
@@ -0,0 +1,120 @@
...
+## Instructions
+When Kyuubi is secured by Kerberos, we can not connect to Kyuubi simply by providing a JDBC url 
+with username and password.
+Instead, following steps should be taken.
+
+## Installing and Configuring the Kerberos Clients
+Usually, Kerberos client is installed as default. You can validate it using klist tool.
+
+```bash
+$ klist -V
```
pan3793, line 35: It does not work on macOS, `klist -v` works

pan3793: Consider lots of users use BI tools on Windows, should we also give them some instructions?

docs/client/kerberized_kyuubi.md
```diff
@@ -0,0 +1,120 @@
...
+Kerberos 5 version 1.15.1
+```
+
+If the client is not installed, you should install it ahead based on the OS platform.
+
+`krb5.conf` is a configuration file for tuning up the creation of Kerberos ticket cache.
+The default location is `/etc` on Linux,
```
pan3793, line 42: Does it has a high priority user specific default location?

zhouyifan279: No, `krb5.conf` does not have a user level default location.

docs/client/kerberized_kyuubi.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+**Note**: 
+If Kyuubi is running on the same host, take care not to overwrite the ticket cache file used by Kyuubi.
+As the default ticket cache file is `/tmp/krb5cc_$(id -u)` on Linux, we should either set environment 
+variable `KRB5CCNAME` or switch to another OS user before executing `kinit` command.
+
+## Ensure core-site.xml Exists in Classpath
+Like hadoop clients, `hadoop.security.authentication` should be set to `KERBEROS` in `core-site.xml` 
+to let Hive JDBC driver use Kerberos authentication. `core-site.xml` should be placed under beeline's
```
turboFei, line 77: for beeline, maybe just need append `-Dhadoop.security.authentication=KERBEROS` into KYUUBI_BEELINE_OPTS

docs/client/kerberized_kyuubi.md
zhouyifan279: > Consider lots of users use BI tools on Windows, should we also give them some instructions?

Added configuration details for different OS platforms.

mmuru:
@zhouyifan279: Please, include Kyuubi/DBeaver/Kerberos Authentication flow diagram. On MacOS, the cache location is not /tmp directory instead it uses KCM. Also, the user needs to install krb5 via brew because Kerberos pre installed and configured with MacOS by default and it does not work. The user needs to create a /etc/krb5.conf file. Otherwise, the instructions look good to me.

ulysses-you:
can you provide a screen snapshot of `docs/client/kerberized_kyuubi.md` ?

docs/security/kinit.md
```diff
@@ -23,9 +23,13 @@
...
 
-In order to work with a kerberos-enabled cluster, Kyuubi provides this kinit auxiliary service.
-It will periodically re-kinit with to keep the Ticket Cache fresh.
+Kinit auxiliary service is a critical service both for authentication between Kyuubi client/server 
+and for authentication between Kyuubi server/Hadoop cluster in a Kerberos environment.
+It will get a Kerberos Ticket Cache from KDC and periodically re-kinit to keep the Ticket Cache fresh.
 
+**Note**:
+Kinit auxiliary service is critical to Kyuubi Kerberos authentication, but not vice versa.
+Kinit auxiliary service can also work with other authentication mode.
```
ulysses-you, line 32: It seems more readable like this ?
```
- xxx
- xxx
```

otherwise the tow lines in md will be inlined.

zhouyifan279: Thanks for the advice. Updated.

docs/security/kinit.md
```diff
@@ -78,8 +82,21 @@ kyuubi\.kinit\.keytab|<div style='width: 80pt;word-wrap: break-word;white-space:
 kyuubi\.kinit\.interval|<div style='width: 80pt;word-wrap: break-word;white-space: normal'>PT1H</div>|<div style='width: 200pt;word-wrap: break-word;white-space: normal'>How often will Kyuubi server run `kinit -kt [keytab] [principal]` to renew the local Kerberos credentials cache</div>|<div style='width: 20pt'>1.0.0</div>
 kyuubi\.kinit\.max<br>\.attempts|<div style='width: 80pt;word-wrap: break-word;white-space: normal'>10</div>|<div style='width: 200pt;word-wrap: break-word;white-space: normal'>How many times will `kinit` process retry</div>|<div style='width: 20pt'>1.0.0</div>
 
-When `hadoop.security.authentication` is set to `KERBEROS`, in `$HADOOP_CONF_DIR/core-site` or `$KYUUBI_HOME/conf/kyuubi-defaults.conf`,
-it indicates that we are targeting a secured cluster, then we need to specify `kyuubi.kinit.principal` and `kyuubi.kinit.keytab` for authentication.
+When working with a Kerberos-enabled Hadoop cluster, we should ensure that `hadoop.security.authentication` 
+is set to `KERBEROS` in `$HADOOP_CONF_DIR/core-site` or `$KYUUBI_HOME/conf/kyuubi-defaults.conf`.
```
yaooqinn, line 86: core-site.xml

zhouyifan279: Updated

docs/client/kerberized_kyuubi.md
zhouyifan279: Tried but it does not work.

zhouyifan279:
> @zhouyifan279: Please, include Kyuubi/DBeaver/Kerberos Authentication flow diagram. On MacOS, the cache location is not /tmp directory instead it uses KCM. Also, the user needs to install krb5 via brew because Kerberos pre installed and configured with MacOS by default and it does not work. The user needs to create a /etc/krb5.conf file. Otherwise, the instructions look good to me.

@mmuru Thanks for the advice and rectification. 
Added a diagram about the authentication and emphasized ticket cache differences on different OS.
As to the installation of kerberos client, there are already many instructions on the Internet. I'd like to not add it in this doc.

yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #1560, opened 2021-12-14
Align all Jackson libs version

pan3793:

Fix missing version of `jackson-module-jaxb-annotations` in `dependencyManagement`

### DIFF
--- a/dev/dependencyList
+++ b/dev/dependencyList
@@ -48,13 +48,12 @@ jackson-module-jaxb-annotations/2.11.4//jackson-module-jaxb-annotations-2.11.4.j
 jackson-module-paranamer/2.11.4//jackson-module-paranamer-2.11.4.jar
 jackson-module-scala_2.12/2.11.4//jackson-module-scala_2.12-2.11.4.jar
 jakarta.activation-api/1.2.1//jakarta.activation-api-1.2.1.jar
-jakarta.activation/2.0.0//jakarta.activation-2.0.0.jar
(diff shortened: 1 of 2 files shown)


dev/dependencyList
```diff
@@ -44,9 +44,10 @@ jackson-dataformat-yaml/2.11.4//jackson-dataformat-yaml-2.11.4.jar
 jackson-datatype-jsr310/2.11.4//jackson-datatype-jsr310-2.11.4.jar
 jackson-jaxrs-base/2.11.4//jackson-jaxrs-base-2.11.4.jar
 jackson-jaxrs-json-provider/2.11.4//jackson-jaxrs-json-provider-2.11.4.jar
-jackson-module-jaxb-annotations/2.9.9//jackson-module-jaxb-annotations-2.9.9.jar
+jackson-module-jaxb-annotations/2.11.4//jackson-module-jaxb-annotations-2.11.4.jar
 jackson-module-paranamer/2.11.4//jackson-module-paranamer-2.11.4.jar
 jackson-module-scala_2.12/2.11.4//jackson-module-scala_2.12-2.11.4.jar
+jakarta.activation-api/1.2.1//jakarta.activation-api-1.2.1.jar
```
ulysses-you, line 50: shall we keep the same version between `jakarta.activation-api` and `jakarta.activation` ?

pan3793: Seems there have different package group, let me investigate

pan3793:
The `jakarta` and `javax` dependencies need more time to evaluate, anyway, it works now

PR #1592, opened 2021-12-20
[TEST][ICEBERG][DELTA] Introduce row level operation test for data lake format

pan3793:

Introduce row level operation test for data lake format and remove redundant tests


kyuubi-common/src/test/scala/org/apache/kyuubi/operation/DeltaMetadataTests.scala
```diff
@@ -87,162 +87,4 @@ trait DeltaMetadataTests extends HiveJDBCTestHelper with DeltaSuiteMixin {
       assert(!rs3.next())
     }
   }
-
-  test("get type info") {
```
pan3793, line 91: It's independent with table format, but engine.

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/IcebergRowLevelOperationTests.scala
```diff
@@ -0,0 +1,103 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+import org.apache.kyuubi.IcebergSuiteMixin
+
+trait IcebergRowLevelOperationTests extends HiveJDBCTestHelper with IcebergSuiteMixin {
+
+  private def createAndInitTable(tableName: String)(records: Seq[(Int, String)]): Unit = {
```
yaooqinn, line 24: we can have a statement as parameter here

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkMetadataTests.scala
```diff
@@ -287,4 +292,188 @@ trait SparkMetadataTests extends HiveJDBCTestHelper {
...
+        () => metaData.autoCommitFailureClosesAllResultSets,
+        () => metaData.getClientInfoProperties,
+        () => metaData.getFunctionColumns("", "%", "%", "%"),
+        () => metaData.getPseudoColumns("", "%", "%", "%"),
+        () => metaData.generatedKeyAlwaysReturned).foreach { func =>
+        val e = intercept[SQLFeatureNotSupportedException](func())
+        assert(e.getMessage === "Method not supported")
+      }
+
+      import org.apache.kyuubi.KYUUBI_VERSION
```
yaooqinn, line 407: let's fix this too

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/DeltaRowLevelOperationTests.scala
```diff
@@ -0,0 +1,103 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+import org.apache.kyuubi.DeltaSuiteMixin
+
+trait DeltaRowLevelOperationTests extends HiveJDBCTestHelper with DeltaSuiteMixin {
```
yaooqinn, line 22: how about we rename it to DML and bring more cases in?

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/IcebergRowLevelOperationTests.scala
pan3793: Do you mean
```
def createAndInitTable(tableName: String)
                      (records: Seq[(Int, String)])
                      (initSql: Option[String] = None): Unit
```

yaooqinn: private def createAndInitTable(statement: Statement, tableName: String)

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/DeltaRowLevelOperationTests.scala
pan3793: Row level operation is the spark offical term introduced in https://issues.apache.org/jira/browse/SPARK-35801, I'm planning to add other cases later in `TimeTravelTests` `SecondaryIndexTests` etc to reflect different capbilities in different table format

yaooqinn: The complexity of mixins brings more overhead of maintenance. I have never experienced using traits for classification, not abstraction

ulysses-you: what's the behavior of iceberg or delta ?

pan3793: Any suggestion for classification expect traits in scalatest? @yaooqinn

pan3793: > what's the behavior of iceberg or delta ?

I don't get your point.

pan3793: BTW, I explained the idea in #1287, if we reject to use traits for classification, we need to reconsider the test classes hierarchical

yaooqinn:
thanks, merged to master

Issue #1594, opened 2021-12-21
[Bug] Kyuubi On Kubernetes Run Spark Client Mode, Executor Pod can unknow driver host


zwangsheng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When using Kyuubi On Kubernetes submit Spark SQL Engine by Client mode, spark driver will running in kyuubi pod. And when driver pull up executor will pass driver url like `spark://CoarseGrainedScheduler@kyuubi-example:42845`. Executor pod will catch UnKnowHostException.

### Affects Version(s)

1.4.0

### Additional context

Executor Log
```
Caused by: java.io.IOException: Failed to connect to kyuubi-example:42845
	at org.apache.spark.network.client.TransportClientFactory.createClient(TransportClientFactory.java:287)
	at org.apache.spark.network.client.TransportClientFactory.createClient(TransportClientFactory.java:218)
	at org.apache.spark.network.client.TransportClientFactory.createClient(TransportClientFactory.java:230)
	at org.apache.spark.rpc.netty.NettyRpcEnv.createClient(NettyRpcEnv.scala:204)
	at org.apache.spark.rpc.netty.Outbox$$anon$1.call(Outbox.scala:202)
	at org.apache.spark.rpc.netty.Outbox$$anon$1.call(Outbox.scala:198)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.net.UnknownHostException: kyuubi-example
	at java.net.InetAddress.getAllByName0(InetAddress.java:1281)
	at java.net.InetAddress.getAllByName(InetAddress.java:1193)
	at java.net.InetAddress.getAllByName(InetAddress.java:1127)
	at java.net.InetAddress.getByName(InetAddress.java:1077)
	at io.netty.util.internal.SocketUtils$8.run(SocketUtils.java:156)
	at io.netty.util.internal.SocketUtils$8.run(SocketUtils.java:153)
[...]

PR #1596, opened 2021-12-21
[KYUUBI #1594] SparkProcessBuild submit add --conf spark.driver.host=

zwangsheng:

detail #1594 
When Kyuubi On Kubernetes submit spark engine by client mode, executor pod can not connect to driver by kyuubi pod name, which is default set.


zwangsheng:
cc @yaooqinn @ulysses-you @pan3793 , please help check.

yaooqinn:
why not just configure `spark.driver.host` directly?

zwangsheng:
> why not just configure `spark.driver.host` directly?

When using kyuubi HA, we do not know driver will up on which node. So we can not point `spark.driver.host`  directly.

zwangsheng:
And, local test build success. please retest.

yaooqinn:
> > why not just configure `spark.driver.host` directly?
> 
> When using kyuubi HA, we do not know driver will up on which node. So we can not point `spark.driver.host` directly.

why, I saw you just point to a local host here

zwangsheng:
> > > why not just configure `spark.driver.host` directly?
> > 
> > 
> > When using kyuubi HA, we do not know driver will up on which node. So we can not point `spark.driver.host` directly.
> 
> why, I saw you just point to a local host here

Kyuubi On Kubernetes use client deploy-mode, Spark Sql Engine will run in Kyuubi Pod, so use `InetAddress.getLocalHost.getHostAddress` get Kyuubi Pod IP.

SimonYang-CS:
@zwangsheng, if we deploy Kyuubi as a statefulset app, then Kyuubi /Spark driver host can be reach from Spark executor.

[discussions 1606](https://github.com/apache/incubator-kyuubi/discussions/1606)

kyuubi-statefulset-example-2.kyuubi-server-headless.namespace.svc.cluster.local

which one do you prefer ? pod ip address or statefulset  host name ?

zwangsheng:
Thanks for your idea, @SimonYang-CS ! I'll give it a try using `statefulset` kyuubi.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -157,6 +158,10 @@ class SparkProcessBuilder(
       buffer += s"$newKey=$v"
     }
 
+    buffer += CONF
```
yaooqinn, line 161: conf.get(FRONTEND_THRIFT_BINARY_BIND_HOST).foreach

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -157,6 +158,10 @@ class SparkProcessBuilder(
...
+    buffer += s"spark.driver.host=${conf.get(FRONTEND_THRIFT_BINARY_BIND_HOST)
+      .getOrElse(findLocalInetAddress.getCanonicalHostName)}"
```
yaooqinn, line 163: getHostName?

zwangsheng:
In test, `StatefulSet Kyuubi` can solve driver url problem. But we also need help Kyuubi in `Pod` or `Deployment` Run Spark Client Mode Success.

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilderSuite.scala
```diff
@@ -271,6 +271,26 @@ class SparkProcessBuilderSuite extends KerberizedTestHelper {
     assert(b1.toString.contains(s"--conf spark.files=$testKeytab"))
 
   }
+
+  test("Engine Bind on hostName or IP with Kyuubi") {
```
yaooqinn: bind        
 host name

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -157,6 +157,18 @@ class SparkProcessBuilder(
       buffer += s"$newKey=$v"
     }
 
+    /**
+     * Kyuubi respect user setting config, if user set `spark.driver.host`, will pass it on.
+     * If user don't set this, will use thrift binary bind host to set.
+     * Kyuubi wants the Engine to bind hostName or IP with Kyuubi.
```
yaooqinn, line 163: Let's also explain the spark behavior of handling this conf

zwangsheng: OK, will add.

yaooqinn:

turboFei:
@zwangsheng 
if the deploy-mode is cluster, does it work?

turboFei:
I see, it does not impact, please ignore it.

```
    // Ignore invalid spark.driver.host in cluster modes.
    if (deployMode == CLUSTER) {
      sparkConf.remove(DRIVER_HOST_ADDRESS)
    }
```

bydeath:
If the configuration item `spark.driver.host` only works when there is only one instance pod? Because there is  a exact kyuubi pod   which references the executors.  But here are two kyuubi server pods default. And I am using the kyuubi of version 1.6.0.
Also there is only `kyuubi-deployment.yaml`, no statefulset yaml file.

PR #1598, opened 2021-12-21
[KYUUBI #1582] Use ClientTypeSignature to further analysis of trino column type

hddong:

Use ClientTypeSignature to further analysis of trino column type

### DIFF
--- a/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/schema/RowSet.scala
+++ b/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/schema/RowSet.scala
@@ -23,6 +23,7 @@ import java.nio.charset.StandardCharsets
 import scala.collection.JavaConverters._
 
 import io.trino.client.ClientStandardTypes._
+import io.trino.client.ClientTypeSignature
 import io.trino.client.Column
 import io.trino.client.Row
 import org.apache.hive.service.rpc.thrift.TBinaryColumn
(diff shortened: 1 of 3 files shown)


externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/schema/RowSet.scala
```diff
@@ -237,46 +238,32 @@ object RowSet {
         // Only match string in nested type values
         "\"" + s + "\""
 
-      // for Array Map and Row, temporarily convert to string
-      // TODO further analysis of type
-      case (list: java.util.List[_], _) =>
-        formatValue(list)
+      case (list: java.util.List[_], ARRAY) if typ.getArgumentsAsTypeSignatures.size() > 0 =>
```
yaooqinn, line 241: why the guard is needed here?

hddong: > why the guard is needed here?

Have call `getArgumentsAsTypeSignatures .get(0)` below, used to avoid out of index. However, it may not be necessary, because `getArgumentsAsTypeSignatures.get(0)` always has a value.
Have removed it.

ulysses-you: so shall we `assert(getArgumentsAsTypeSignatures.nonEmpty)` first ?

hddong: sure, done.

yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #1593, opened 2021-12-21
[Bug] Embedded zkServer.getConnectString shall respect user's setting ahead


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Currently, we force it bind to the HostName. It's better to respect user's settings ahead

### Affects Version(s)

1.4.0,1.3.1,1.2.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yaooqinn:
cc @rvacaru

zwangsheng:
I'm willing to fix this.

PR #1600, opened 2021-12-21
[KYUUBI #1593] add config help use default local ip instead of default local hostname

zwangsheng:

User can select IP or hostname for binding.


zwangsheng:
Using wrong path, will open another.

PR #1602, opened 2021-12-21
[KYUUBI #1593] use user set host or ip instead of read hostname from user set

zwangsheng:

User can select IP or hostname for binding.
Detail #1593 


yaooqinn:
we dont need a extra conf，we can respect user directly

zwangsheng:
> we dont need a extra conf，we can respect user directly

Do you mean just use the string that the user set?

yaooqinn:
> > we dont need a extra conf，we can respect user directly
> 
> Do you mean just use the string that the user set?

yes

yaooqinn (approved):

ulysses-you (approved):

ulysses-you:
thanks, merging to master/branch-1.4

PR #1603, opened 2021-12-21
[KYUUBI #1579] Implement basic ability of executing statement in Flink engine

yanghua:


externals/kyuubi-flink-sql-engine/pom.xml
```diff
@@ -139,6 +139,12 @@
       <artifactId>jul-to-slf4j</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.flink</groupId>
+      <artifactId>flink-test-utils_2.12</artifactId>
```
pan3793, line 145: Kyuubi uses log4j12, but this has transitive deps log4j2. We should exclude other log4j2 deps except to `log4j2-core`, and bridge it to sfl4j

pan3793: nit: `2.12` => `${scala_binary_version}`

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -0,0 +1,172 @@
...
+
+package org.apache.kyuubi.engine.flink.operation
+
+import java.util
+import java.util.List
+import java.util.concurrent.{RejectedExecutionException, ScheduledExecutorService, TimeUnit}
+import java.util.stream.IntStream
+
+import scala.collection.JavaConversions._
+import scala.collection.JavaConverters.collectionAsScalaIterableConverter
```
pan3793, line 26: `JavaConverters` is deprecated, use `import scala.collection.JavaConversions._` and `[util.List object].asScala`

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -0,0 +1,172 @@
...
+      val resultID = resultDescriptor.getResultId
+
+      var rows: util.List[Row] = new util.ArrayList[Row]()
+
+      var loop = true
+      while (loop) {
+        Thread.sleep(50) // slow the processing down
+
+        val result = executor.snapshotResult(sessionId, resultID, 2)
+        if (result.getType eq TypedResult.ResultType.PAYLOAD) {
```
pan3793, line 129: can we use 
```
result.getType match {
  case TypedResult.ResultType.PAYLOAD =>
  case TypedResult.ResultType.EOS =>
}
```
here?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -0,0 +1,172 @@
...
+    override val shouldRunAsync: Boolean,
+    queryTimeout: Long)
+  extends FlinkOperation(OperationType.EXECUTE_STATEMENT, session) with Logging {
+
+  private val operationLog: OperationLog =
+    OperationLog.createOperationLog(session, getHandle)
+
+  private var resultDescriptor: ResultDescriptor = _
+
+  private var columnInfos: List[ColumnInfo] = _
```
pan3793, line 53: `List[ColumnInfo]` => `util.List[ColumnInfo]`

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -0,0 +1,172 @@
...
+          IntStream.rangeClosed(1, result.getPayload).forEach((page: Int) => {
+            def foo(page: Int) = {
+              for (row <- executor.retrieveResultPage(resultID, page)) {
+                rows.add(row)
+              }
+            }
+
+            foo(page)
+          })
+        } else if (result.getType eq TypedResult.ResultType.EOS) loop = false
```
pan3793, line 140: Ideally, we should return a iterator and lazy fetch the next snapshot (or periodically update the snapshot), but it's ok for now, as we just tend to support the simplest query `select 1` in this initial PR.

pan3793 (approved):

pan3793:
Thanks, merging to master for v1.5

Issue #1564, opened 2021-12-15
[Bug] Server connectionUrl shall respect user's setting first


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Currently, we force it to hostname, this is not right.

### Affects Version(s)

1.4.0,1.3.1,1.2.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yaooqinn:
cc @JeremyTon

JeremyTon:
I'm working on this

yaooqinn:
thank you @JeremyTon

JeremyTon:
I have been pull request on #1604 
please check it ,thank you @yaooqinn

PR #1604, opened 2021-12-22
[KYUUBI #1564] fix Server connectionUrl force to hostname 

JeremyTon:

Currently, we force it to hostname, this is not right.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
@@ -40,7 +40,7 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
   private var portNum: Int = _
   private var jettyServer: Server = _
   private var connector: ServerConnector = _
-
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -76,7 +76,11 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
 
   override def connectionUrl: String = {
     checkInitialized()
-    s"${serverAddr.getCanonicalHostName}:$portNum"
+    conf.get(FRONTEND_REST_BIND_HOST) match {
```
ulysses-you, line 79: shall we also update line 47 in `initialize` ?

ulysses-you:
thank you @JeremyTon for the contribution ! also cc @zwangsheng

zwangsheng:
thanks for the contribution! How about adding a few test cases to this function?

yaooqinn:
https://github.com/apache/incubator-kyuubi/pull/1602 maybe we shall simply follow this PR

yaooqinn:
this is fixed, i will close this

PR #1636, opened 2021-12-28
Correct the java version in maven-compiler-plugin

ulysses-you:

The default java version of `maven-compiler-plugin` is 1.6, it will fail if we use the higher version sugar.

We should define the version explicitly.

Pass CI

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1243,6 +1243,8 @@
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>${maven.plugin.compiler.version}</version>
                     <configuration>
+                        <source>${java.version}</source>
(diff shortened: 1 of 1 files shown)


pan3793:
IIRC, the plugin will auto pick the maven property `java.version`

pan3793:
```
      <configuration>
        ...
        <source implementation="java.lang.String" default-value="1.6">${maven.compiler.source}</source>
        ...
        <target implementation="java.lang.String" default-value="1.6">${maven.compiler.target}</target>
        ...
      </configuration>
```

ulysses-you:
ah, I see.  It lost the `maven.compiler.source` and `maven.compiler.target`. Closes this.

PR #1642, opened 2021-12-29
[KYUUBI #1641] Add Trino client

hddong:

Add trino client to communicate with trino cluster.

### DIFF
--- a/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
+++ b/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
@@ -19,10 +19,11 @@ package org.apache.kyuubi.engine.trino
 
 import java.util.ArrayList
 import java.util.concurrent.ArrayBlockingQueue
+import java.util.concurrent.Executors
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration
 import scala.concurrent.duration.Duration
@@ -35,15 +36,22 @@ import io.trino.client.StatementClient
 import io.trino.client.StatementClientFactory
 
 import org.apache.kyuubi.KyuubiSQLException
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.engine.trino.TrinoConf.DATA_PROCESSING_POOL_SIZE
 import org.apache.kyuubi.engine.trino.TrinoStatement._
 
 /**
  * Trino client communicate with trino cluster.
  */
-class TrinoStatement(trinoContext: TrinoContext, sql: String) {
+class TrinoStatement(trinoContext: TrinoContext, kyuubiConf: KyuubiConf, sql: String) {
 
   private lazy val trino = StatementClientFactory
     .newStatementClient(trinoContext.httpClient, trinoContext.getClientSession, sql)
+
+  private lazy val dataProcessingPoolSize = kyuubiConf.get(DATA_PROCESSING_POOL_SIZE)
+
+  implicit val ec: ExecutionContext =
+    ExecutionContext.fromExecutor(Executors.newFixedThreadPool(dataProcessingPoolSize))
 
   def getTrinoClient: StatementClient = trino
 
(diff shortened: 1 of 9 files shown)


externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
```diff
@@ -0,0 +1,195 @@
...
+import io.trino.client.StatementClient
+import io.trino.client.StatementClientFactory
+
+import org.apache.kyuubi.KyuubiSQLException
+import org.apache.kyuubi.engine.trino.TrinoStatement._
+
+/**
+ * Trino client communicate with trino cluster.
+ */
+class TrinoStatement(trinoContext: TrinoContext, sql: String) {
```
yaooqinn, line 43: Are we going to fully implement trino-jdbc?

hddong: No. Trino client is Restful API, and there is no function to get the resultSet directly. We need to collect result to local, as `execute` does, similar with `spark.sql("sql")`.

hddong: BTW, for jdbc,  it seems that we can support connecting other JDBC services with `kyuubi-jdbc-engine` in the future.
For trino, use trino client directly, we can better manage logs, exceptions and progressing. And omit the process of intermediate conversion to `Java.sql`

kubernetes/integration-tests/test-trino.yaml
```diff
@@ -0,0 +1,143 @@
...
+    component: coordinator
+data:
+  node.properties: |
+    node.environment=production
+    node.data-dir=/data/trino
+    plugin.dir=/usr/lib/trino/plugin
+
+  jvm.config: |
+    -server
+    -Xmx8G
```
pan3793, line 56: There is 7G limitation on GitHub Action runner

kubernetes/integration-tests/test-trino.yaml
```diff
@@ -0,0 +1,143 @@
...
+          configMap:
+            name: tcb-trino-coordinator
+        - name: catalog-volume
+          configMap:
+            name: tcb-trino-catalog
+      imagePullSecrets:
+        - name: registry-credentials
+      containers:
+        - name: trino-coordinator
+          image: "trinodb/trino:latest"
```
pan3793, line 122: how about use a specific version rather than latest?

externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
```diff
@@ -0,0 +1,195 @@
...
+    val future = Future[Unit] {
+      while (trino.isRunning) {
+        val data = trino.currentData().getData()
+        if (data != null) {
+          data.asScala.map(_.asScala.toList)
+            .foreach(e => putOrThrow(rowQueue, e))
+        }
+        trino.advance()
+      }
+    }
```
pan3793, line 86: Any blocking operation in this block? IMO, `scala.concurrent.ExecutionContext.Implicits.global` is not a good choise for that. BTW, please give the variable a meaningful name rather than just `future`.

kubernetes/integration-tests/test-trino.yaml
hddong: Have used `testcontainers` instead. ITtest will add when the trino engine complete and will address those comments.

externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
hddong: > Any blocking operation in this block?

The resultSet have multiple batches, read and flush to buffer need blocking.

pan3793: Then let's use a dedicated `ExecutionContext` and make the thread size configurable rather than `scala.concurrent.ExecutionContext.Implicits.global`, the global one is something like global fork join pool, usually for noblocking calculation.

hddong: Have addressed

Issue #1597, opened 2021-12-21
[Bug] StatementStatsPagedTable maps exception to query execution wrongly


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

in table head, we have a column named `Query Execution`, but in the data field we put the query exception in

### Affects Version(s)

1.4.0,1.3.1

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yaooqinn:
cc @cody1993

PR #1657, opened 2021-12-30
[KYUUBI #1597] fix row value in Query Execution should be queryExecution

xifeng:

row value in Query Execution should be queryExecution

image


yaooqinn (approved):

xifeng:
@yaooqinn I have updated the code. When I test locally, I found that queryExecution should be arouded by `td` label. I also update the test evidence.

yaooqinn:
thanks for the update @xifeng

yaooqinn:
thanks, merged to master

PR #1662, opened 2021-12-30
Turn up maven options in mvn script

yaooqinn:

### DIFF
--- a/build/mvn
+++ b/build/mvn
@@ -22,7 +22,7 @@ _DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # Preserve the calling directory
 _CALLING_DIR="$(pwd)"
 # Options used during compilation
-_COMPILE_JVM_OPTS="-Xms2g -Xmx2g -Xss8m -XX:ReservedCodeCacheSize=1g -Xss128m"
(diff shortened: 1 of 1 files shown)


build/mvn
```diff
@@ -22,7 +22,7 @@ _DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # Preserve the calling directory
 _CALLING_DIR="$(pwd)"
 # Options used during compilation
-_COMPILE_JVM_OPTS="-Xms2g -Xmx2g -Xss8m -XX:ReservedCodeCacheSize=512m"
+_COMPILE_JVM_OPTS="-Xms2g -Xmx2g -Xss8m -XX:ReservedCodeCacheSize=1g -Xss128m"
```
pan3793, line 25: Do we meet stack overflow? The default stack size is 1m, 128m is too large in my view

yaooqinn: No evidence for s.o. 128 is from spark

yaooqinn: but I suspect that the flink tests cause this as the builds mostly end up with killed there

pan3793: I guess it's caused by heap OOM, let wait for CI results

yaooqinn:
thanks, merged to master

Issue #1656, opened 2021-12-30
[Bug] dist failed with  --spark-provided


shenbinglife:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

exec dist cmd, but failed to create dist file with error log 
```
./build/dist --tgz --spark-provided  

... 
[INFO] Kyuubi Project Download Externals .................. SUCCESS [  2.678 s]
[INFO] Kyuubi Project Hive JDBC Client .................... SUCCESS [  5.889 s]
[INFO] Kyuubi Project Hive JDBC Shaded Client ............. SUCCESS [ 22.951 s]
[INFO] Kyuubi Project Engine Flink SQL .................... SUCCESS [ 40.334 s]
[INFO] Kyuubi Project Engine Spark SQL .................... SUCCESS [ 42.998 s]
[INFO] Kyuubi Project Server .............................. SUCCESS [ 49.469 s]
[INFO] Kyuubi Project Dev Code Coverage ................... SUCCESS [  4.162 s]
[INFO] Kyuubi Project Engine Trino ........................ SUCCESS [ 19.811 s]
[INFO] Kyuubi Project Assembly ............................ SUCCESS [  1.333 s]
[INFO] Kyuubi Project Hive Beeline ........................ SUCCESS [  2.725 s]
[INFO] kyuubi-rest-client ................................. SUCCESS [  5.348 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  04:56 min
[INFO] Finished at: 2021-12-30T11:45:33+08:00
[INFO] ------------------------------------------------------------------------
/d/workspace/github/incubator-kyuubi-shenbinglife
cp: cannot stat '/d/workspace/github/incubator-kyuubi-shenbinglife/externals/kyuubi-download/target/flink-1.14.2/': No such file or directory

shenb@T3-GW468X2 MINGW64 /d/workspace/github/incubator-kyuubi-shenbinglife (master)
```

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1664, opened 2021-12-31
[KYUUBI #1656] Fix dist failed with --spark-provided

shenbinglife:

image
image

### DIFF
--- a/build/dist
+++ b/build/dist
@@ -171,7 +171,7 @@ fi
 
 MVN_DIST_OPT="-DskipTests"
 if [[ "$SPARK_PROVIDED" == "true" ]]; then
-  MVN_DIST_OPT="$MVN_DIST_OPT -Pflink-provided,spark-provided"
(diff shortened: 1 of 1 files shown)


build/dist
```diff
@@ -171,7 +171,7 @@ fi
 
 MVN_DIST_OPT="-DskipTests"
 if [[ "$SPARK_PROVIDED" == "true" ]]; then
-  MVN_DIST_OPT="$MVN_DIST_OPT -Pflink-provided,spark-provided"
+  MVN_DIST_OPT="$MVN_DIST_OPT -Pspark-provided"
```
pan3793, line 174: We should also handle `FLINK_PROVIDED` here, we also need add somethings like
```
if [[ "$FLINK_PROVIDED" == "true" ]]; then
  MVN_DIST_OPT="$MVN_DIST_OPT -Pflink-provided"
fi
```

pan3793 (approved):

pan3793:
Thanks, merging to master

Issue #1675, opened 2022-01-04
[FEATURE] Redact secret information from SparkSQLEngine's Log


minyk:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

We redact secret informations on UI with #1634. Do the same thing for SparkSQLEngine's log.

### Motivation

Secret informations on log is dangerous on the production environment. Secret info should be redacted in some way.

### Describe the solution

As Discussed in #1634's comments, move `redactedStatement` to `AbstractOperation` from `SparkOperation`. then logging with `redactedStatement` in `setState()`. 
In this way, other engine's log is redacted too if it only override `redactedStatement` in `AbstractOperation`'s subclass.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1677, opened 2022-01-04
[KYUUBI #1675][FEATURE] Redact secret information from SparkSQLEngine log

minyk:

Secret informations on log is dangerous on the production environment. Secret info should be redacted in some way.

capture from Spark Driver's STDERR
Screen Shot 2022-01-04 at 11 56 43 AM

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
@@ -60,6 +60,8 @@ abstract class AbstractOperation(opType: OperationType, session: Session)
 
   def statement: String = opType.toString
 
+  def redactedStatement: String = statement
+
   protected def setHasResultSet(hasResultSet: Boolean): Unit = {
     this.hasResultSet = hasResultSet
     handle.setHasResultSet(hasResultSet)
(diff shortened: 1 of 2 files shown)


ulysses-you (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
```diff
@@ -80,7 +82,7 @@ abstract class AbstractOperation(opType: OperationType, session: Session)
       case _ =>
     }
     info(s"Processing ${session.user}'s query[$statementId]: ${state.name} -> ${newState.name}," +
-      s" statement: $statement$timeCost")
+      s" statement: $redactedStatement$timeCost")
```
ulysses-you, line 85: not related this PR, shall we and `, ` between `$redactedStatement` and `$timeCost`  ?

yaooqinn (approved):
LGTM

yaooqinn:
thanks, merged to master

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
minyk: @ulysses-you comma is already added from here: https://github.com/apache/incubator-kyuubi/blob/e4f5ba7010dd497f40d51f5677b16d3cf7b882d4/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala#L81

ulysses-you: ah I see, thank you !

Issue #1679, opened 2022-01-04
[Bug] Remove invalid column and field number check


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

ResultSet checks if the column number of data matches the one in metadata by calling `iterator.next()` to get a sample row, which would make the first row missing in the final result.

The bug didn't appear in the tests, because the condition is wrong. Now it checks the column number only if data is empty, which should be another bug.

### Affects Version(s)

1.5.0

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yaooqinn:
👍

PR #1680, opened 2022-01-04
[KYUUBI #1679] Remove invalid column and field number check

link3280:

ResultSet checks if the column number of data matches the one in metadata by calling iterator.next() to get a sample row, which would make the first row missing in the final result.

The bug didn't appear in the tests, because the condition is wrong. Fixed by c0e9072a4c3383562b0fe1938e3bfa1947376754.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/java/org/apache/kyuubi/engine/flink/result/ResultSet.java
+++ b/externals/kyuubi-flink-sql-engine/src/main/java/org/apache/kyuubi/engine/flink/result/ResultSet.java
@@ -53,11 +53,6 @@ public class ResultSet {
     this.resultKind = Preconditions.checkNotNull(resultKind, "resultKind must not be null");
     this.columns = Preconditions.checkNotNull(columns, "columns must not be null");
     this.data = Preconditions.checkNotNull(data, "data must not be null");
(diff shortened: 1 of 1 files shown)


yaooqinn (approved):

externals/kyuubi-flink-sql-engine/src/main/java/org/apache/kyuubi/engine/flink/result/ResultSet.java
```diff
@@ -53,11 +53,6 @@ private ResultSet(
     this.resultKind = Preconditions.checkNotNull(resultKind, "resultKind must not be null");
     this.columns = Preconditions.checkNotNull(columns, "columns must not be null");
     this.data = Preconditions.checkNotNull(data, "data must not be null");
-    if (!data.hasNext()) {
-      Preconditions.checkArgument(
-          columns.size() == data.next().getArity(),
```
yanghua, line 58: good catch, if we want to check, we would expect the type of data is a `List`.

Then we can do:

```
if (!data.isEmpty()) {
  Preconditions.checkArgument(columns.size() == data.get(0).getArity()
}
```

But for Iterator, we can ignore the check here.

yanghua (approved):
LGTM

PR #1682, opened 2022-01-05
[KYUUBI #1678] Fixed: The value of hasResultSet is always set to true in Kyuubi Server

### LINKED ISSUE #1678: [Bug] The value of hasResultSet is always set to true in Kyuubi Server
### Code of Conduct

### Search before asking

### Describe the bug

https://github.com/apache/incubator-kyuubi/blob/47533100f6f69b02a90985637a8fcc3f2cebb28c/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala#L56-L73

`hasResultSet` is set to true by default in the `beforeRun` method. `hasResultSet` value can be obtained from `remoteOpHandle`, whether to consider setting the value in the `executeStatement` ？

Like this: 
```
 private def executeStatement(): Unit = {
    try {
      // We need to avoid executing query in sync mode, because there is no heartbeat mechanism
      // in thrift protocol, in sync mode, we cannot distinguish between long-run query and
      // engine crash without response before socket read timeout.
      _remoteOpHandle = client.executeStatement(statement, true, queryTimeout)
      setHasResultSet(_remoteOpHandle.isHasResultSet) 
    } catch onError()
  }
```

### Affects Version(s)

master/1.4.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

iodone:
Fixes #1678

Updating `hasResultSet` state which is determined by the remote service

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
@@ -69,6 +69,7 @@ class ExecuteStatement(
       // in thrift protocol, in sync mode, we cannot distinguish between long-run query and
       // engine crash without response before socket read timeout.
       _remoteOpHandle = client.executeStatement(statement, true, queryTimeout)
+      setHasResultSet(_remoteOpHandle.isHasResultSet)
     } catch onError()
   }
 


ulysses-you:
Seems test failed due to the flink memory issue, can you retrigger the GA @iodone thanks?

```java
FlinkOperationSuite:
Starting cluster.
Starting standalonesession daemon on host fv-az212-351.
Starting taskexecutor daemon on host fv-az212-351.
./build/mvn: line 118:  1840 Killed                  ${MVN_BIN} $MAVEN_CLI_OPTS "$@"
```

also cc @yanghua

pan3793:
I'm finding the solution to reduce flink cluster memory usage.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
```diff
@@ -69,6 +69,7 @@ class ExecuteStatement(
       // in thrift protocol, in sync mode, we cannot distinguish between long-run query and
       // engine crash without response before socket read timeout.
       _remoteOpHandle = client.executeStatement(statement, true, queryTimeout)
+      setHasResultSet(_remoteOpHandle.isHasResultSet)
```
yaooqinn, line 72: Shall we also update it according to TGetOperationStatusResp in waitStatementComplete

iodone: https://github.com/apache/incubator-kyuubi/blob/e0d9458144e68411f1dd62afb91efd1e1ef92cc5/kyuubi-common/src/main/scala/org/apache/kyuubi/service/ThriftBinaryFrontendService.scala#L415-L437

There is a unified configuration inside `GetOperationStatus` method, here we only need to change the operation's state

yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #1683, opened 2022-01-05
[BUILD] Reduce Flink local cluster memory usage

pan3793:

Make CI stable

### DIFF
--- a/kyuubi-server/src/test/resources/flink_conf/flink-conf.yaml
+++ b/kyuubi-server/src/test/resources/flink_conf/flink-conf.yaml
@@ -1,0 +1,272 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+
+#==============================================================================
+# Common
+#==============================================================================
+
+# The external address of the host on which the JobManager runs and can be
+# reached by the TaskManagers and any clients which want to connect. This setting
+# is only used in Standalone mode and may be overwritten on the JobManager side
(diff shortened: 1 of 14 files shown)


kyuubi-server/src/test/resources/flink_conf/zoo.cfg
```diff
@@ -0,0 +1,36 @@
+################################################################################
```
yaooqinn, line 1: Can zookeeper peer be skipped?

pan3793: Copied from `$FLINK_HOME/conf`

yaooqinn: why?

kyuubi-server/src/test/resources/flink_conf/flink-conf.yaml
```diff
@@ -0,0 +1,272 @@
...
+# The RPC port where the JobManager is reachable.
+
+jobmanager.rpc.port: 6123
+
+
+# The total process memory size for the JobManager.
+#
+# Note this accounts for all memory usage within the JobManager process, including JVM metaspace and other overhead.
+
+jobmanager.memory.process.size: 768m
```
pan3793, line 44: The only change is L44 and L51, and tune it to use the minimal valid value. (may we can reduce more, but need to tune other memory confs)

kyuubi-server/src/test/resources/flink_conf/flink-conf.yaml
```diff
@@ -0,0 +1,272 @@
...
+
+
+# The total process memory size for the TaskManager.
+#
+# Note this accounts for all memory usage within the TaskManager process, including JVM metaspace and other overhead.
+
+taskmanager.memory.process.size: 1024m
```
pan3793, line 51: mark

kyuubi-server/src/test/resources/flink_conf/zoo.cfg
pan3793: I can not find the way to overwrite the configure defined in `$FLINK_HOME/conf/flink-conf.yaml`, thus copied the whole `$FLINK_HOME/conf` to `flink_conf`, and then modify the `flink_conf/flink-conf.yaml`, `zoo.cfg` is not required for our test, because we don't turn on HA.

The current approach is copy ``$FLINK_HOME/conf` with minimal modification and keep others as-is, or we just pick the required conf files?

yaooqinn: I don't follow your thoughts, 'when you need to modify one config, you cp the whole dir.`

pan3793: 1. we can only use `FLINK_CONF_DIR`, there is no option like `--conf-file=xxx.yaml`
2. I can not found a way to overwrite a property when `bin/start-cluster`

any suggestion?

yanghua:
IMO, we can close this PR. WDYT? @yaooqinn @pan3793

Issue #1681, opened 2022-01-05
[SUB-TASK][KPIP-2] Extend Flink ExecuteStatement to support more operations


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

`ExecuteStatement` of Flink engine now supports QueryOperation only. We can extend it to support more operations that Flink Executor already supports. e.g. `ShowTableOperation`,`CreateTableOperation`, `ExplainOperation`.

### Motivation

This is a sub-task of KPIP-2 #1322 .

### Describe the solution

Add a case matching before executing operations. Execute QueryOperation (which returns ResultDescriptor) via `executor.executeQuery`, while execute other operations (which returns TableResult) via `executor.executeOperation`. 

There're several operations that `executor.executeOperation` doesn't support, such as `SetOperation`. We may address that afterward.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yanghua:
@link3280 Thanks for opening this issue, it would be better to follow this issue prefix: `[SUB-TASK][KPIP-2]` for archiving the Flink SQL engine.

link3280:
> @link3280 Thanks for opening this issue, it would be better to follow this issue prefix: `[SUB-TASK][KPIP-2]` for archiving the Flink SQL engine.

OK, I'll follow the pattern. Thanks for the reminder.

Issue #1322, opened 2021-11-02
[Umbrella][KPIP-2] Support FlinkSQL Engine


yanghua:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the proposal

- [X] The detailed proposal is [here](https://github.com/apache/incubator-kyuubi/discussions/1132). #1132 

### Task list

- [x] #1323 @yanghua 
- [x] #1329 @yanghua  
- [x] #1340 @yanghua 
- [x] #1386 @yanghua 
- [x] #1492 @yanghua  
- [x] #1503 @yanghua 
- [x] #1525 @yanghua 
- [x] #1542 @yanghua 
- [x] #1555 @yanghua  
- [x] #1578 @yanghua 
- [x] #1579 @yanghua 
- [x] #1587 @pan3793 
- [x] #1629 @pan3793  
- [x] #1643 @SteNicholas
- [x] #1645 @SteNicholas
- [x] #1646 @xifeng 
- [x] #1647 @yanghua
- [x] #1648 @SteNicholas
- [x] #1663
- [x] #1665 @yanghua 
- [x] #1679 @link3280 
- [x] #1681 @link3280 
- [x] #1685 @SteNicholas
- [x] #1687 @yanghua 
- [x] #1705 @yanghua 
- [x] #1704 @link3280 
- [x] #1709 @yanghua 
- [x] #1774 @yanghua 
- [x] #1787 @yanghua
- [x] #1778 @link3280 
- [x] #1814 @SteNicholas
- [x] #1819 @SteNicholas
- [x] #1825 @link3280 
- [x] #1829 @SteNicholas
- [x] #1838 @link3280 
- [x] #1839 @link3280 
- [x] #1867 @SteNicholas
- [x] #1883 @link3280 
- [x] #1919 @SteNicholas 
- [x] #1926 @yanghua 
- [x] #1866 @deadwind4 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

ulysses-you:
great work

SteNicholas:
@yanghua @yaooqinn , any other subtasks? In above task list, the tasks are completed.

pan3793:
> In above task list, the tasks are completed.

[...]

PR #1693, opened 2022-01-07
[KYUUBI #1681] Extend Flink ExecuteStatement to support more operations

link3280:

ExecuteStatement of Flink engine now supports QueryOperation only. We can extend it to support more operations that Flink Executor already supports. e.g. ShowTableOperation,CreateTableOperation, ExplainOperation.

This is a sub-task of KPIP-2 #1322 .


link3280:
There are two things worth noting about this PR: 

-  I removed some classes in the `result` package to better align with Flink's TableResult. Please take a look if it's okay. cc @yanghua 
- There's a bug in executing DML like `insert` statements. Looks like the session was closed prematurely. It may need more time to investigate. I've temporarily ignored the test for inserts and will make a follow-up PR to fix the problem. But I'm OK if we require to fix that before getting the PR merged.

yaooqinn (approved):
LGTM

ulysses-you:
the style can be formatted using `./dev/reformat`

link3280:
@ulysses-you Thanks for your input. I'll use this tool to format codes in the coming PRs.

link3280:
I've found that the insert query failed due to this issue https://github.com/apache/incubator-kyuubi/issues/1704.

yanghua:
> There are two things worth noting in this PR:
> 
> * I removed some classes in the `result` package to better align with Flink's TableResult. Please take a look if it's okay. cc @yanghua 

It's ok to me.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -47,7 +49,7 @@ class ExecuteStatement(
 
   private var resultDescriptor: ResultDescriptor = _
 
-  private var columnInfos: util.List[ColumnInfo] = _
+  private var columns: util.List[Column] = _
```
yanghua, line 52: unnecessary class level variable?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -101,46 +103,58 @@ class ExecuteStatement(
...
+      .columns(resultDescriptor.getResultSchema.getColumns)
+      .data(rows.toArray[Row])
+      .build
+    setState(OperationState.FINISHED)
+  }
+
+  private def runOperation(operation: Operation): Unit = {
+    val result = executor.executeOperation(sessionId, operation)
+    result.await()
+    columns = result.getResolvedSchema.getColumns
```
yanghua, line 153: unnecessary?

yanghua (commented):
@link3280 Thanks for your contribution. Left two comments you can consider.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -47,7 +49,7 @@ class ExecuteStatement(
 
   private var resultDescriptor: ResultDescriptor = _
```
yanghua, line 50: useless?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -101,46 +103,58 @@ class ExecuteStatement(
     try {
       setState(OperationState.RUNNING)
 
-      columnInfos = new util.ArrayList[ColumnInfo]
+      columns = new util.ArrayList[Column]
```
yanghua, line 106: ditto

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -101,46 +103,58 @@ class ExecuteStatement(
...
-      setState(OperationState.FINISHED)
     } catch {
       onError(cancel = true)
     } finally {
       statementTimeoutCleaner.foreach(_.shutdown())
     }
   }
 
+  private def runQueryOperation(operation: QueryOperation): Unit = {
+    resultDescriptor = executor.executeQuery(sessionId, operation)
```
yanghua, line 121: `val resultDescriptor`?

link3280:
@yanghua Thanks a lot for your review! I've addressed the comments and rebased the PR on master.

yanghua (approved):
LGTM. @link3280 Thanks

yanghua:
@link3280 There are many CI fails, caused by Flink cluster startup. Will try to use `MiniCluster` to fix it recently.

link3280:
> @link3280 There are many CI fails, caused by Flink cluster startup. Will try to use `MiniCluster` to fix it recently.

@yanghua Very much appreciate that! Thanks!

PR #1695, opened 2022-01-07
[KYUUBI #1510] Add TRINO engine type and TrinoProcessBuilder

hddong:

Add TRINO engine type and TrinoProcessBuilder


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -585,6 +585,28 @@ object KyuubiConf {
       .stringConf
       .createOptional
 
+  val ENGINE_TRINO_MAIN_RESOURCE: OptionalConfigEntry[String] =
+    buildConf("session.engine.trino.main.resource")
```
yaooqinn, line 589: seems that we have hardcoded in the shell script

hddong: Passed through the env variable (`TRINO_ENGINE_JAR`)  of the ProcessBuilder.

yaooqinn: I see

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/trino/TrinoProcessBuilder.scala
```diff
@@ -0,0 +1,175 @@
...
+
+  override protected def module: String = "kyuubi-trino-engine"
+
+  override protected def mainClass: String = "org.apache.kyuubi.engine.trino.TrinoSqlEngine"
+
+  override protected def childProcEnv: Map[String, String] = conf.getEnvs +
+    ("TRINO_ENGINE_JAR" -> mainResource.get)
+
+  override protected def commands: Array[String] = {
+    val buffer = new ArrayBuffer[String]()
```
yaooqinn, line 100: How about we pass all the configurations through normal java options?

we can simply get all the stuff inside the trino engine by single `new KyuubiConf()` only. besides, we can same a scopt dependency at engine side

hddong: Done.

externals/kyuubi-trino-engine/bin/trino-engine.sh
```diff
@@ -0,0 +1,37 @@
...
+  exit 1
+fi
+
+RUNNER="${JAVA_HOME}/bin/java"
+
+if [[ "$TRINO_ENGINE_HOME" == "$KYUUBI_HOME/externals/engines/trino" ]]; then
+  TRINO_CLIENT_JAR="$TRINO_ENGINE_JAR"
+else
+  echo "\nTRINO_ENGINE_HOME $TRINO_ENGINE_HOME doesn't match production directory, assuming in development environment..."
+  TRINO_CLIENT_JAR=$(find "$TRINO_ENGINE_HOME/target" -regex ".*/kyuubi-trino-engine_.*.jar")
```
pan3793, line 29: Should we exclude `*-sources.jar` `*-tests.jar` etc. ?

externals/kyuubi-trino-engine/pom.xml
```diff
@@ -85,4 +84,72 @@
 
     </dependencies>
 
+    <build>
+        <outputDirectory>target/scala-${scala.binary.version}/classes</outputDirectory>
+        <testOutputDirectory>target/scala-${scala.binary.version}/test-classes</testOutputDirectory>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
```
pan3793, line 93: Some thoughts on shade engine.

Shade always should be the latest approach to solve the problems. Shade and class relocation makes it difficult to debug, and when `NoClassFoundException` or `NoSuchMethodException` occurs, it's not easy to dig which it comes from which jars. Class relocation also may breaks the reflection. There are some bugs on `maven-shade-plugin` which may hang the compile.

We shade Spark engine and Flink engine because we need to submit the jar to run with the engine runtime jars, so there is a risk which may cause class conflict, because we don't know which jars user put into their engine runtime jar folder.

For Trino engine, I think we don't need to run it with Trino runtime jars. A simple layout e.g. `bin/`, `conf/`, `lib/` maybe a better choice.

externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoEngineArguments.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.trino
+
+case class TrinoEngineArguments(
```
yaooqinn, line 20: this is not used anymore?

externals/kyuubi-trino-engine/pom.xml
hddong: Agree

externals/kyuubi-trino-engine/bin/trino-engine.sh
hddong: There is no `*-sources.jar`, exclude `*-tests.jar` and `*-javadoc.jar`

externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoEngineArguments.scala
hddong: Deleted

externals/kyuubi-trino-engine/bin/trino-engine.sh
pan3793: `*-sources.jar` `*-javadoc.jar` should be created when `-Papache-release`, and we need create a test jar final for it module, just like flink-sql-engine

externals/kyuubi-trino-engine/bin/trino-engine.sh
```diff
@@ -0,0 +1,43 @@
...
+  echo "[ERROR] JAVA_HOME IS NOT SET! CANNOT PROCEED."
+  exit 1
+fi
+
+RUNNER="${JAVA_HOME}/bin/java"
+
+if [[ "$TRINO_ENGINE_HOME" == "$KYUUBI_HOME/externals/engines/trino" ]]; then
+  TRINO_CLIENT_JAR="$TRINO_ENGINE_JAR"
+  TRINO_CLIENT_JARS_DIR="$TRINO_ENGINE_HOME/jars"
+  TRINO_CLIENT_CLASSPATH=$(find $TRINO_CLIENT_JARS_DIR -regex '.*\.jar' | grep -v '.*/kyuubi-trino-engine_.*.jar' | tr '\n' ':')
```
pan3793, line 28: Any reason not use `$TRINO_ENGINE_HOME/jars/*:`?

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/trino/TrinoProcessBuilderSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.kyuubi.config.KyuubiConf.ENGINE_TRINO_CONNECTION_CATALOG
+import org.apache.kyuubi.config.KyuubiConf.ENGINE_TRINO_CONNECTION_URL
+
+class TrinoProcessBuilderSuite extends KyuubiFunSuite {
+  private def conf = KyuubiConf().set("kyuubi.on", "off")
+
+  test("trino process builder") {
+    val trinoConf = conf
+      .set(ENGINE_TRINO_CONNECTION_URL, "http://0.0.0.0:8080")
+      .set(ENGINE_TRINO_CONNECTION_CATALOG, "tpch")
```
pan3793, line 32: Does this catalog has data or only tables definition?

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/trino/TrinoProcessBuilderSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.config.KyuubiConf.ENGINE_TRINO_CONNECTION_CATALOG
+import org.apache.kyuubi.config.KyuubiConf.ENGINE_TRINO_CONNECTION_URL
+
+class TrinoProcessBuilderSuite extends KyuubiFunSuite {
+  private def conf = KyuubiConf().set("kyuubi.on", "off")
+
+  test("trino process builder") {
+    val trinoConf = conf
+      .set(ENGINE_TRINO_CONNECTION_URL, "http://0.0.0.0:8080")
```
pan3793, line 31: We need get the address and port from the started container

pan3793: e.g. https://github.com/housepower/spark-clickhouse-connector/blob/47430ff7c2c6047575287e6ee4fb7f273bcc33c9/clickhouse-core-integration/src/testFixtures/scala/xenon/clickhouse/base/ClickHouseSingleMixIn.scala#L52

hddong: It is used to test process builder, not real connect, not necessary here. Removed

externals/kyuubi-trino-engine/bin/trino-engine.sh
hddong: Done

externals/kyuubi-trino-engine/pom.xml
```diff
@@ -70,7 +70,6 @@
             <scope>test</scope>
         </dependency>
 
-        <!-- https://mvnrepository.com/artifact/org.testcontainers/trino -->
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>trino</artifactId>
```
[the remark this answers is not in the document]
hddong: Will open a new PR to change to `testcontainers-scala-trino`. This PR is not related to testcontainers, and this change was deleted.

externals/kyuubi-trino-engine/bin/trino-engine.sh
hddong: Got it, and done.

yaooqinn:
thanks, merged to master

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/trino/TrinoProcessBuilder.scala
```diff
@@ -0,0 +1,153 @@
...
+      }
+      if (Files.isDirectory(working)) {
+        working
+      } else {
+        Utils.createTempDir(rootAbs, proxyUser)
+      }
+    }.getOrElse {
+      Utils.createTempDir(namePrefix = proxyUser)
+    }
+  }
```
yanghua, line 141: We can reuse the ProcBuilder#workingDir after merging #1758 . WDYT?

yaooqinn: +1

hddong: +1, I will address it.

Issue #1696, opened 2022-01-08
[SUB-TASK][FEATURE] Fetch results and logs rows rate


zhenjiaguo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add fetch results and logs rows rate metric. Use [Meters](https://metrics.dropwizard.io/4.2.0/manual/core.html#meters) to record the rows rate that fetch results and fetch logs.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1707, opened 2022-01-09
[KYUUBI #1696] Add fetch logs and results rows rate

zhenjiaguo:

Add fetch results and logs rows rate metric.


kyuubi-server/src/main/scala/org/apache/kyuubi/server/BackendServiceFetchRowsRate.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.kyuubi.service.BackendService
+
+trait BackendServiceFetchRowsRate extends BackendService {
+  abstract override def fetchResults(
+      operationHandle: OperationHandle,
+      orientation: FetchOrientation,
+      maxRows: Int,
+      fetchLog: Boolean): TRowSet = {
+    val rowSet = super.fetchResults(operationHandle, orientation, maxRows, fetchLog)
+    if (rowSet == null) return rowSet
```
yaooqinn, line 34: when will this happen?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/BackendServiceFetchRowsRate.scala
```diff
@@ -0,0 +1,58 @@
...
+          case t: TI32Column => t.getValues.size()
+          case t: TI16Column => t.getValues.size()
+          case t: TBoolColumn => t.getValues.size()
+          case t: TByteColumn => t.getValues.size()
+          case t: TBinaryColumn => t.getValues.size()
+          case _ => 0
+        }
+      } else {
+        rowSet.getRowsSize
+      }
```
yaooqinn, line 50: simply getRowsSize?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -136,7 +136,7 @@ class KyuubiServer(name: String) extends Serverable(name) {
   def this() = this(classOf[KyuubiServer].getSimpleName)
 
   override val backendService: AbstractBackendService =
-    new KyuubiBackendService() with BackendServiceTimeMetric
+    new KyuubiBackendService() with BackendServiceTimeMetric with BackendServiceFetchRowsRate
```
yaooqinn, line 139: can we merge these 2 traits

kyuubi-server/src/main/scala/org/apache/kyuubi/server/BackendServiceFetchRowsRate.scala
zhenjiaguo: I will not happen. I will remove it.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
zhenjiaguo: yes

kyuubi-server/src/main/scala/org/apache/kyuubi/server/BackendServiceFetchRowsRate.scala
yaooqinn: we don't have RowsSize for column based set?

yaooqinn:
```scala
  def toColumnBasedSet(rows: Seq[Row], schema: StructType, timeZone: ZoneId): TRowSet = {
    val size = rows.length
    val tRowSet = new TRowSet(0, new java.util.ArrayList[TRow](size))
    schema.zipWithIndex.foreach { case (filed, i) =>
      val tColumn = toTColumn(rows, i, filed.dataType, timeZone)
      tRowSet.addToColumns(tColumn)
    }
    tRowSet
  }
```

zhenjiaguo: Yes, we have `RowsSize` for column based set. And RowSet may be a row base set. So I first `if` try to get column base set rows and `else` to get row base set rows.

yaooqinn: I see
Is it efficient?

Or we can use the used `columnCount` for deliver this value from engine side

zhenjiaguo: If we use `columnCount` to deliver `rowSize`, it may cause some confuse, I don't think it's a good idea.
I think it's efficient, because it just a `match case`. And I do a little test, it shows that my guess is right:
```
             count = 8508349
         mean rate = 12083.92 calls/second
     1-minute rate = 16944.96 calls/second
     5-minute rate = 14413.72 calls/second
    15-minute rate = 7229.02 calls/second
               min = 0.00 milliseconds
               max = 0.00 milliseconds
              mean = 0.00 milliseconds
            stddev = 0.00 milliseconds
            median = 0.00 milliseconds
              75% <= 0.00 milliseconds
              95% <= 0.00 milliseconds
              98% <= 0.00 milliseconds
              99% <= 0.00 milliseconds
            99.9% <= 0.00 milliseconds
```

yaooqinn:
thanks, merged to master

Issue #1697, opened 2022-01-08
[SUB-TASK][FEATURE] BackendService method avg call times and rate.


zhenjiaguo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

 `BackendService` method avg call times and rate. Replace [Histograms](https://metrics.dropwizard.io/4.2.0/manual/core.html#Histograms) with [Timer](https://metrics.dropwizard.io/4.2.0/manual/core.html#Timer) to record it. Timer include both [Meters](https://metrics.dropwizard.io/4.2.0/manual/core.html#meters) and [Histograms](https://metrics.dropwizard.io/4.2.0/manual/core.html#Histograms).

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1708, opened 2022-01-09
[KYUUBI #1697] BackendService method avg call times and rate

zhenjiaguo:

BackendService method avg call times and rate. Replace Histograms with Timer to record it. Timer include both Meters and Histograms.


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
```diff
@@ -89,4 +95,16 @@ object MetricsSystem {
...
     maybeSystem.foreach(func(_))
   }
+
+  @throws[Exception]
+  def timerTracing[T](name: String)(f: => T): T = {
+    val startTime = System.nanoTime()
+    try {
+      f
+    } finally {
+      MetricsSystem.tracing(
```
ulysses-you, line 105: nit: `tracing(`

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #1716, opened 2022-01-10
Unify Hive deps for server and client

yaooqinn:

This PR make the hive deps be consistent between server and clients for a single release, we shall not play client-server compatibility tricky in one release


kyuubi-assembly/pom.xml
```diff
@@ -94,6 +94,12 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
```
yaooqinn, line 97: > Contains com.google.common.util.concurrent.internal.InternalFutureFailureAccess and InternalFutures. Most users will never need to use this artifact. Its classes is conceptually a part of Guava, but they're in this separate artifact so that Android libraries can use them without pulling in all of Guava (just as they can use ListenableFuture by depending on the listenablefuture artifact).

yaooqinn: we shall avoid introduce to server path, but this is needed for test

yaooqinn:
cc @pan3793

pom.xml
```diff
@@ -141,20 +142,13 @@
...
-        <hive.client.commons-lang.version>2.6</hive.client.commons-lang.version>
         <hive.client.commons-cli.version>1.2</hive.client.commons-cli.version>
         <hive.client.commons-codec.version>1.15</hive.client.commons-codec.version>
-        <hive.client.curator.version>2.12.0</hive.client.curator.version>
-        <hive.client.guava.version>14.0.1</hive.client.guava.version>
         <hive.client.httpcomponents-client.version>4.5.13</hive.client.httpcomponents-client.version>
         <hive.client.httpcomponents-core.version>4.4.13</hive.client.httpcomponents-core.version>
         <hive.client.jline.version>2.12</hive.client.jline.version>
-        <hive.client.storage-api.version>2.4.0</hive.client.storage-api.version>
+        <hive.client.storage-api.version>2.7.2</hive.client.storage-api.version>
```
pan3793, line 150: Why does this change required?

pom.xml
```diff
@@ -309,6 +299,12 @@
                 </exclusions>
             </dependency>
 
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>failureaccess</artifactId>
+                <version>1.0.1</version>
```
pan3793, line 305: nit: use property

pom.xml
yaooqinn: align with spark

pan3793:
I'm +0 on this change. The initial idea of using shaded version is to hide the implementation of kyuubi hive jdbc client, and only expose standard JDBC API, then we can refactor the jdbc client step by step to cut the relation between hive deps.

yaooqinn:
why does a unified hive.version block the cutting thing you mentioned?

yaooqinn:
> if we manage the exclusion of hive in root pom

We are still do exclusion from module to module?

yaooqinn:
Or you mean the `hive-service-rpc` and it's deps(thrift s) sepically?

pan3793:
> > if we manage the exclusion of hive in root pom
> 
> We are still do exclusion from module to module?

Specific for Hive stuffs, yes.

yaooqinn:
> > > if we manage the exclusion of hive in root pom
> > 
> > 
> > We are still do exclusion from module to module?
> 
> Specific for Hive stuffs, yes.

NONONO, I mean "We are still do exclusion from module to module"

pan3793:
`hive-service-rpc` are quite light, but idealy we can 1) relocated it or 2)regenete from thrift IDL but change package name to `o.a.kyuubi` then we will never conflict with Spark runtime jars.

yaooqinn:
what's your point? in this pr, we still manage the exclustion as-is except hive-service-rpc

yaooqinn:
also, cc @ulysses-you and @turboFei

ulysses-you:
looks fine. 

I think we needn't specify the hive version in spark extension module and it's clear to make it consistent with client and kyuubi server. We can do further optimization if we meet the real issue about dependency.

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #1722, opened 2022-01-11
[NOTICE] Update NOTICE for 2022

pan3793:

Welcome to 2022

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,6 +1,6 @@
 
 Apache Kyuubi (Incubating)
-Copyright 2021 The Apache Software Foundation.
+Copyright 2021-2022 The Apache Software Foundation.
 
 This product includes software developed at
 The Apache Software Foundation (https://www.apache.org/).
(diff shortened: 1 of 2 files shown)


NOTICE
```diff
@@ -1,6 +1,6 @@
 
 Apache Kyuubi (Incubating)
-Copyright 2021 The Apache Software Foundation.
+Copyright 2021-2022 The Apache Software Foundation.
```
yaooqinn, line 3: using and onwards

NOTICE-binary
```diff
@@ -1,6 +1,6 @@
 
 Apache Kyuubi (Incubating)
-Copyright 2021 The Apache Software Foundation.
+Copyright 2021-2022 The Apache Software Foundation.
```
yaooqinn, line 3: ditto

NOTICE
pan3793: We can not do that.
https://www.mail-archive.com/general@incubator.apache.org/msg75717.html

yaooqinn: thanks for sharing

yaooqinn (approved):
LGTM

yaooqinn:
thanks merged to master, 1.4, 1.3

Issue #1731, opened 2022-01-11
[FEATURE] Kyuubi server should not fail to start when Hadoop FileSystem class not found


zhouyifan279:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Kyuubi server should not fail to start when Hadoop FileSystem class not found

### Motivation

User may set `fs.defaultFS` to other Hadoop-compatible FileSystem, such as Alluxio, in `core-site.xml`.
As alluxio dependencies is not included in kyuubi, Kyuubi fails to start at the initialization of `HadoopCredentialsManager`.

Kyuubi should tolerate such situation.

### Describe the solution

Print warning message about missing Hadoop FileSystem class and continue to boot.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1737, opened 2022-01-12
[WIP][KYUUBI #1731][FEATURE] Kyuubi server should not fail to start when Hadoop FileSystem class not found

zhouyifan279:
Kyuubi server fails to start when Hadoop FileSystem class not found.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopFsDelegationTokenProvider.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopFsDelegationTokenProvider.scala
@@ -21,6 +21,7 @@ import java.lang.reflect.UndeclaredThrowableException
 import java.security.PrivilegedExceptionAction
 
 import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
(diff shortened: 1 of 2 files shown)


zhouyifan279:
cc @pan3793 @ulysses-you

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopFsDelegationTokenProvider.scala
```diff
@@ -93,13 +94,26 @@ object HadoopFsDelegationTokenProvider {
...
+            Some(value)
+          case Failure(e) =>
+            warn(s"Failed to get Hadoop FileSystem instance by URI: $uri", e)
+            None
+        }
+      }
       .toSet
 
-    filesystemsToAccess + defaultFS
+    Try(FileSystem.get(hadoopConf)) match {
```
ulysses-you, line 110: I think it should fail if the FileSytem can not initialize

zhouyifan279: User may set fs.defaultFS to other Hadoop-compatible FileSystem, such as Alluxio, in core-site.xml.
These FileSystems may not require DelegationTokens at all.

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #1749, opened 2022-01-13
[KYUUBI #1643] Implement GetFunctions operation

SteNicholas:

Implement GetFunctions operation.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.flink.operation
+
+import scala.collection.JavaConverters._
+
+import org.apache.commons.lang3.StringUtils
+import org.apache.flink.table.api.{DataTypes, ResultKind, TableEnvironment}
+import org.apache.flink.table.catalog.Column
+import org.apache.flink.table.types.logical.VarCharType
+import org.apache.flink.types.Row
+
+import org.apache.kyuubi.engine.flink.result.ResultSet
+import org.apache.kyuubi.operation.OperationType
+import org.apache.kyuubi.operation.meta.ResultSetSchemaConstant._
+import org.apache.kyuubi.session.Session
+
+class GetFunctions(
+    session: Session,
+    catalogName: String,
+    schemaName: String,
+    functionName: String)
+  extends FlinkOperation(OperationType.GET_FUNCTIONS, session) {
+
+  override protected def runInternal(): Unit = {
+    try {
+      var columns: Map[String, (Int, Boolean)] = Map(
+        FUNCTION_CAT -> Tuple2(VarCharType.DEFAULT_LENGTH, false),
+        FUNCTION_SCHEM -> Tuple2(VarCharType.DEFAULT_LENGTH, false),
+        FUNCTION_NAME -> Tuple2(VarCharType.DEFAULT_LENGTH, false))
+      var rows: List[Row] = List()
+      val tableEnv: TableEnvironment = sessionContext.getExecutionContext.getTableEnvironment
+      tableEnv.listFunctions().diff(tableEnv.listUserDefinedFunctions())
+        .filter(f => StringUtils.isEmpty(functionName) || f == functionName)
+        .foreach { f =>
+          columns += (FUNCTION_NAME -> Tuple2(math.max(f.length, columns(FUNCTION_NAME)._1), true))
+          rows = Row.of(null, null, f) +: rows
+        }
+      tableEnv.listCatalogs()
+        .filter(c => StringUtils.isEmpty(catalogName) || c == catalogName)
+        .foreach { c =>
+          columns += (FUNCTION_CAT -> Tuple2(math.max(c.length, columns(FUNCTION_CAT)._1), true))
+          val catalog = tableEnv.getCatalog(c).get()
+          catalog.listDatabases().asScala
+            .filter(d => StringUtils.isEmpty(schemaName) || d == schemaName)
+            .foreach { d =>
+              columns += (
+                FUNCTION_CAT -> Tuple2(math.max(d.length, columns(FUNCTION_SCHEM)._1), true))
+              catalog.listFunctions(d).asScala
+                .filter(f => StringUtils.isEmpty(functionName) || f == functionName)
+                .foreach { f =>
+                  columns += (
+                    FUNCTION_NAME -> Tuple2(math.max(f.length, columns(FUNCTION_NAME)._1), true))
+                  rows = Row.of(c, d, f) +: rows
+                }
+            }
+        }
+      resultSet = ResultSet.builder.resultKind(ResultKind.SUCCESS_WITH_CONTENT)
+        .columns(
+          Column.physical(
+            columns.head._1,
+            if (columns.head._2._2) DataTypes.VARCHAR(columns.head._2._1).notNull()
+            else DataTypes.VARCHAR(columns.head._2._1)),
+          Column.physical(
+            columns.toList(1)._1,
+            if (columns.toList(1)._2._2) DataTypes.VARCHAR(columns.toList(1)._2._1).notNull()
+            else DataTypes.VARCHAR(columns.toList(1)._2._1)),
+          Column.physical(
+            columns.last._1,
+            if (columns.last._2._2) DataTypes.VARCHAR(columns.last._2._1).notNull()
+            else DataTypes.VARCHAR(columns.last._2._1)))
+        .data(rows.asJava.toArray(new Array[Row](0)))
+        .build
+    } catch {
+      onError()
+    }
+  }
+}
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
@@ -41,24 +41,19 @@ class GetFunctions(
     try {
       val schemaPattern = toJavaRegex(schemaName)
       val functionPattern = toJavaRegex(functionName)
-      var rows: List[Row] = List()
       val tableEnv: TableEnvironment = sessionContext.getExecutionContext.getTableEnvironment
-      filterPattern(
+      val systemFunctions = filterPattern(
         tableEnv.listFunctions().diff(tableEnv.listUserDefinedFunctions()),
         functionPattern)
-        .foreach { f =>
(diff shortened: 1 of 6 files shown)


pan3793:
Thanks for contribution! Could you please rebase you branch on latest master? BTW, please run `dev/reformat` before commit your code, you can also configure your IDEA use scalafmt to formatter your code rather than the default one.

SteNicholas:
@pan3793 , sorry for not rebasing the master branch. I have rebased the latest master branch. Please help to review this pull request.

SteNicholas:
@pan3793 , sorry for not using the `dev/reformat`. I have rebased the latest master branch and reformated the codes. Please help to review this pull request.

yaooqinn:
cc @yanghua @link3280

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,93 @@
...
+      var rows: List[Row] = List()
+      val tableEnv: TableEnvironment = sessionContext.getExecutionContext.getTableEnvironment
+      tableEnv.listFunctions().diff(tableEnv.listUserDefinedFunctions())
+        .filter(f => StringUtils.isEmpty(functionName) || f == functionName)
+        .foreach { f =>
+          columns += (FUNCTION_NAME -> Tuple2(math.max(f.length, columns(FUNCTION_NAME)._1), true))
+          rows = Row.of(null, null, f) +: rows
+        }
+      tableEnv.listCatalogs()
+        .filter(c => StringUtils.isEmpty(catalogName) || c == catalogName)
```
pan3793, line 55: I think it should be pattern match not exactly string match

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,93 @@
...
+        .foreach { c =>
+          columns += (FUNCTION_CAT -> Tuple2(math.max(c.length, columns(FUNCTION_CAT)._1), true))
+          val catalog = tableEnv.getCatalog(c).get()
+          catalog.listDatabases().asScala
+            .filter(d => StringUtils.isEmpty(schemaName) || d == schemaName)
```
pan3793, line 60: ditto

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,93 @@
...
+            .foreach { d =>
+              columns += (
+                FUNCTION_CAT -> Tuple2(math.max(d.length, columns(FUNCTION_SCHEM)._1), true))
+              catalog.listFunctions(d).asScala
+                .filter(f => StringUtils.isEmpty(functionName) || f == functionName)
```
pan3793, line 65: ditto

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+      resultSet = ResultSet.builder.resultKind(ResultKind.SUCCESS_WITH_CONTENT)
+        .columns(
+          Column.physical(
+            columns.head._1,
+            if (columns.head._2._2) DataTypes.VARCHAR(columns.head._2._1).notNull()
+            else DataTypes.VARCHAR(columns.head._2._1)),
+          Column.physical(
+            columns.toList(1)._1,
```
link3280, line 80: Do we create multiple lists by calling toList multiple times?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,93 @@
...
+    schemaName: String,
+    functionName: String)
+  extends FlinkOperation(OperationType.GET_FUNCTIONS, session) {
+
+  override protected def runInternal(): Unit = {
+    try {
+      var columns: Map[String, (Int, Boolean)] = Map(
+        FUNCTION_CAT -> Tuple2(VarCharType.DEFAULT_LENGTH, false),
+        FUNCTION_SCHEM -> Tuple2(VarCharType.DEFAULT_LENGTH, false),
+        FUNCTION_NAME -> Tuple2(VarCharType.DEFAULT_LENGTH, false))
```
link3280, line 45: I think function names must not be empty or null in any case. WDYT?

link3280 (commented):
@SteNicholas Thanks a lot for your work! I left some comments, please kindly take a look. 

Besides that, I'm not very sure if we should determine column types (e.g. its scale or precision) based on the returned data. I think types should reflect all possible values of the query, but now we may get `FUNTION_NAME varchar(16)` at one result set, `FUNTION_NAME varchar(20)` from another. WDYT?

SteNicholas:
> @SteNicholas Thanks a lot for your work! I left some comments, please kindly take a look.
> 
> Besides that, I'm not very sure if we should determine column types (e.g. its scale or precision) based on the returned data. I think types should reflect all possible values of the query, but now we may get `FUNTION_NAME varchar(16)` at one result set, `FUNTION_NAME varchar(20)` from another. WDYT?

@link3280 , I refer to the implementation of the `GetCatalogs` operation worked by @yanghua . In `GetCatalogs` operation, the column types are determined by the returned data. IMO, the colum types determined by data makes sense to me because the data types are actually get by the actual data.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
SteNicholas: Yes, I agree with the point that the `FUNCTION_NAME` column couldn't be nullable. I have set this non nullable after getting the functions.

yaooqinn:
I am +1 with @link3280, the get catalog API shall have a static result schema including column size, name, type, and restraint. Otherwise, we will result in supersize for clients like HUE/tableu/dbeaver

link3280:
@SteNicholas +1 to `STRING`.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,75 @@
...
+      val schemaPattern = toJavaRegex(schemaName)
+      val functionPattern = toJavaRegex(functionName)
+      var rows: List[Row] = List()
+      val tableEnv: TableEnvironment = sessionContext.getExecutionContext.getTableEnvironment
+      filterPattern(
+        tableEnv.listFunctions().diff(tableEnv.listUserDefinedFunctions()),
+        functionPattern)
+        .foreach { f =>
+          rows = Row.of(null, null, f) +: rows
+        }
```
pan3793, line 51: We can leverage `map` `flatMap` to simplify the code.
```suggestion
    val globalFunctions = filterPattern(
        tableEnv.listFunctions().diff(tableEnv.listUserDefinedFunctions()),
        functionPattern)
        .map { f => Row.of(null, null, f) }
```

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,75 @@
...
+      tableEnv.listCatalogs()
+        .filter { c => StringUtils.isEmpty(catalogName) || c == catalogName }
+        .foreach { c =>
+          val catalog = tableEnv.getCatalog(c).get()
```
pan3793, line 55: Your filter accepts `null`, will it cause NPE if catalogs contains `null`?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/util/StringUtils.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.flink.util
+
+import java.util.regex.PatternSyntaxException
+
+import org.apache.kyuubi.Logging
+
+object StringUtils extends Logging {
```
pan3793, line 24: Seems `extends Logging` is not required

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/util/StringUtils.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * This utility can be used for filtering pattern in the "Like" of "Show Tables / Functions" DDL
```
pan3793, line 27: Let's add comments where this function come from, and if possible, also add a ut to help user understand how it works.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/util/StringUtils.scala
```diff
@@ -0,0 +1,46 @@
...
+   * @param names the names list to be filtered
+   * @param pattern the filter pattern, only '*' and '|' are allowed as wildcards, others will
+   *                follow regular expression convention, case insensitive match and white spaces
+   *                on both ends will be ignored
+   * @return the filtered names list in order
+   */
+  def filterPattern(names: Seq[String], pattern: String): Seq[String] = {
+    val funcNames = scala.collection.mutable.SortedSet.empty[String]
+    pattern.trim().split("\\|").foreach { subPattern =>
```
link3280, line 36: What about regex patterns that contain '|' or '*'? It seems that they're not properly escaped.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetFunctions.scala
```diff
@@ -0,0 +1,75 @@
...
+  extends FlinkOperation(OperationType.GET_FUNCTIONS, session) {
+
+  override protected def runInternal(): Unit = {
+    try {
+      val schemaPattern = toJavaRegex(schemaName)
+      val functionPattern = toJavaRegex(functionName)
+      var rows: List[Row] = List()
+      val tableEnv: TableEnvironment = sessionContext.getExecutionContext.getTableEnvironment
+      filterPattern(
+        tableEnv.listFunctions().diff(tableEnv.listUserDefinedFunctions()),
```
link3280, line 47: Do we have a consensus that we need built-in system functions only? Docs of [DatabaseMetaData](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getFunctions-java.lang.String-java.lang.String-java.lang.String-) that you refer to states user functions should be included as well.

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
```diff
@@ -119,6 +122,36 @@ class FlinkOperationSuite extends WithFlinkSQLEngine with HiveJDBCTestHelper {
...
+              assert(resultSet.getString(FUNCTION_CAT) == null)
+              assert(resultSet.getString(FUNCTION_SCHEM) === null)
+              assert(resultSet.getString(FUNCTION_NAME) === func)
+            }
+          }
+        }
+      val expected =
+        List("currentTimestamp", "currentDate", "currentTime", "localTimestamp", "localTime")
+      Seq("current", "local")
+        .foreach { func =>
```
link3280, line 142: `func` -> `pattern`?

link3280 (commented):
@SteNicholas Thanks for your contribution. I left some minor comments, PTAL.

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #1762, opened 2022-01-14
[KYUUBI #1700] add wait lock thread count

zhenjiaguo:

Add `KyuubiSyncThriftClient` wait locks thread count


yaooqinn:
can we update a report.json with offline test in the pr desc?

zhenjiaguo:
> can we update a report.json with offline test in the pr desc?

Oh, It's always run as asynchronous mode, this metric is meaningless.

PR #1766, opened 2022-01-14
[TEST] Remove SessionManagerSuite to reduce duplicated tests

yaooqinn:

SessionManagerSuite re-runs all tests in the super class, it's unnecessary.


yaooqinn:
cc @yanghua

ulysses-you (approved):

ulysses-you:
thanks, merging to master

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/TClientTestUtils.scala
```diff
@@ -0,0 +1,75 @@
...
+    transport.open()
+    try {
+      f(client)
+    } finally {
+      socket.close()
+    }
+  }
+
+  /**
+   * s shall be [[TFrontendService]]
```
pan3793, line 48: Then why declare it as a `FrontendService`?

yaooqinn: unnecessary cast at callerside

Issue #1704, opened 2022-01-08
[SUB-TASK][KPIP-2] Complete Flink data type to Hive data type mapping


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, only a subset of Flink data types is supported. Queries that return results with unsupported data types would fail with Thrift TTransportExecption. We should fix this by completing Flink data type to Hive data type mapping.

### Motivation

This is a sub-task of KPIP-2 #1322.

### Describe the solution

Complete mappings in RowSet.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1770, opened 2022-01-16
[KYUUBI #1704] Complete Flink-Hive data type mapping

link3280:

Currently, only a subset of Flink data types is supported. Queries that return results with unsupported data types would fail with Thrift TTransportExecption. We should fix this by completing Flink data type to Hive data type mapping.


yaooqinn:
thanks, merged to master

PR #1779, opened 2022-01-17
Fix flaky test using connection share level by default

ulysses-you:

The CI become flaky after we adding more engines (flink/trino) and more tests(can see in package `org.apache.kyuubi.operation`) 

We should release engine as far as possible.

Pass CI


ulysses-you:
It seems make connection by default will slow down the CI. We should make session lifecycle consistent with test (beforeAll/afterAll).

PR #1785, opened 2022-01-17
Stop Discovery services properly

yaooqinn:

Fix some issues:

1. EngineServiceDiscovery calls deregister twice
2. When LOST zk-server, the delay for gracefully stop shall be calculated via retry policy not retrywait * maxWaitMs
     - engine default using n time retry policy, we can avoid unnecessary cache for engines when the whole test file finished and server stopped
     - otherwise, we are very likey to hit the GA memory limits
```
     KyuubiOperationPerGroupSuite:
./build/mvn: line 99:  1725 Killed                  ${MVN_BIN} $MAVEN_CLI_OPTS "$@"
Error: Process completed with exit code 137.
```
3. When LOST, we shall skip connection retries, closing znode & closingclient

### DIFF
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
@@ -48,6 +48,8 @@ class EngineServiceDiscovery(
         case _ =>
       }
       discoveryClient.closeClient()
+    } else {
+      warn(s"The Zookeeper ensemble is LOST")
     }
     super.stop()
   }
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
@@ -31,21 +31,25 @@ class EngineServiceDiscovery(
     fe: FrontendService) extends ServiceDiscovery("EngineServiceDiscovery", fe) {
 
   override def stop(): Unit = synchronized {
-    discoveryClient.deregisterService()
-    conf.get(ENGINE_SHARE_LEVEL) match {
-      // For connection level, we should clean up the namespace in zk in case the disk stress.
-      case "CONNECTION" =>
-        try {
-          if (discoveryClient.postDeregisterService) {
-            info("Clean up discovery service due to this is connection share level.")
+    if (!isServerLost.get()) {
+      discoveryClient.deregisterService()
+      conf.get(ENGINE_SHARE_LEVEL) match {
+        // For connection level, we should clean up the namespace in zk in case the disk stress.
+        case "CONNECTION" =>
+          try {
+            if (discoveryClient.postDeregisterService) {
+              info("Clean up discovery service due to this is connection share level.")
+            }
+          } catch {
+            case NonFatal(e) =>
+              warn("Failed to clean up Spark engine before stop.", e)
           }
-        } catch {
-          case NonFatal(e) =>
-            warn("Failed to clean up Spark engine before stop.", e)
-        }
 
-      case _ =>
+        case _ =>
+      }
+      discoveryClient.closeClient()
     }
     super.stop()
   }
+
 }
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
@@ -20,6 +20,8 @@ package org.apache.kyuubi.ha.client
 import java.io.{File, IOException}
 import javax.security.auth.login.Configuration
 
+import scala.util.Random
+
 import org.apache.curator.framework.{CuratorFramework, CuratorFrameworkFactory}
 import org.apache.curator.retry._
 import org.apache.hadoop.security.UserGroupInformation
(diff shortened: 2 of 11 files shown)


yaooqinn:
cc @hddong @yanghua @turboFei @ulysses-you thanks

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
```diff
@@ -31,21 +31,27 @@ class EngineServiceDiscovery(
...
-        try {
-          if (discoveryClient.postDeregisterService) {
-            info("Clean up discovery service due to this is connection share level.")
+    if (!isServerLost.get()) {
+      discoveryClient.deregisterService()
+      conf.get(ENGINE_SHARE_LEVEL) match {
+        // For connection level, we should clean up the namespace in zk in case the disk stress.
+        case "CONNECTION" =>
+          try {
+            if (discoveryClient.postDeregisterService) {
```
ulysses-you, line 40: Shall we still clean the zk namespace even the Kyuubi server lost ?

Here we should skip `discoveryClient.deregisterService()` since the server node with its sequential number is not existed, but the namspace is alived.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
```diff
@@ -71,6 +73,27 @@ object ZooKeeperClientProvider extends Logging {
...
+    val retryPolicyName = conf.get(HA_ZK_CONN_RETRY_POLICY)
+    RetryPolicies.withName(retryPolicyName) match {
+      case ONE_TIME => baseSleepTime
+      case N_TIME => maxRetries * baseSleepTime
+      case BOUNDED_EXPONENTIAL_BACKOFF =>
+        (0 until maxRetries).map { retryCount =>
+          val retryWait = baseSleepTime * Math.max(1, Random.nextInt(1 << (retryCount + 1)))
+          Math.min(retryWait, maxSleepTime)
+        }.sum
+      case UNTIL_ELAPSED => maxSleepTime
```
ulysses-you, line 89: it should be `maxSleepTime  * maxRetries` ?

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
```diff
@@ -71,6 +73,27 @@ object ZooKeeperClientProvider extends Logging {
...
+    val maxRetries = conf.get(HA_ZK_CONN_MAX_RETRIES)
+    val retryPolicyName = conf.get(HA_ZK_CONN_RETRY_POLICY)
+    RetryPolicies.withName(retryPolicyName) match {
+      case ONE_TIME => baseSleepTime
+      case N_TIME => maxRetries * baseSleepTime
+      case BOUNDED_EXPONENTIAL_BACKOFF =>
+        (0 until maxRetries).map { retryCount =>
+          val retryWait = baseSleepTime * Math.max(1, Random.nextInt(1 << (retryCount + 1)))
+          Math.min(retryWait, maxSleepTime)
+        }.sum
```
ulysses-you, line 88: shall we compare this value and `.min(maxSleepTime * maxRetries)` ?

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
```diff
@@ -71,6 +73,27 @@ object ZooKeeperClientProvider extends Logging {
...
+      case UNTIL_ELAPSED => maxSleepTime
+      case EXPONENTIAL_BACKOFF =>
+        (0 until maxRetries).map { retryCount =>
+          baseSleepTime * Math.max(1, Random.nextInt(1 << (retryCount + 1)))
+        }.sum
```
ulysses-you, line 93: ditto

ulysses-you:
thank you for the catch, I was also considering the similar thing to make egnine release as far as possible.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
ulysses-you: actually the name of `maxSleepTime` should be `maxSleepTimePerRetry`

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
yaooqinn: we are not handling kyuubi server lost at all

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
yaooqinn: not necessary

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
yaooqinn: seems not

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
yaooqinn: ditto

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
yaooqinn: > actually the name of `maxSleepTime` should be `maxSleepTimePerRetry`

not for this rule，but maybe for the last one only. if we using specific name，we may introduce plenty of configs to meet different meanings under different rules

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/EngineServiceDiscovery.scala
yaooqinn: I have made the pr clear to the zk server not kyuubi. your suggestion will introduce extra connection timeout+session timeout for this unnecessary operation

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
ulysses-you: Just clarify it, so if user set the retry policy to `UNTIL_ELAPSED`, the value of `HA_ZK_CONN_MAX_RETRY_WAIT` is actually the totoal wait time. Looks not a big problem.

ulysses-you (approved):

ulysses-you:
thanks, merging to master

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/ZooKeeperClientProvider.scala
yaooqinn: > Just clarify it, so if user set the retry policy to `UNTIL_ELAPSED`, the value of `HA_ZK_CONN_MAX_RETRY_WAIT` is actually the totoal wait time. Looks not a big problem.

FYI, org.apache.curator.retry.RetryUntilElapsed

PR #1797, opened 2022-01-19
Only exposing jdbc url for server side

turboFei:

Only exposing the jdbc url in server side, it is not needed to expose it in engine side and also not secure.


kyuubi-common/src/main/scala/org/apache/kyuubi/service/TBinaryFrontendService.scala
```diff
@@ -113,4 +115,6 @@ abstract class TBinaryFrontendService(name: String)
     server.foreach(_.stop())
     server = None
   }
+
+  protected def isServer(): Boolean = true
```
yaooqinn, line 119: make it false？

yaooqinn (approved):

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #1698, opened 2022-01-08
[SUB-TASK][FEATURE] Added metrics for all kyuubi server operations state rate


zhenjiaguo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Added metrics for all kyuubi server operations state rate, include `INITIALIZED`, `PENDING`, `RUNNING`, etc. Use [Meters](https://metrics.dropwizard.io/4.2.0/manual/core.html#meters) to record it.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1809, opened 2022-01-19
[KYUUBI #1698] Add kyuubi server operation state rate

zhenjiaguo:

Add kyuubi server operation state rate. It will be helpful for us to look into kyuubi server running states.

Screenshot from 2022-01-19 19-13-21

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
@@ -39,7 +39,7 @@ object MetricsConstants {
   final val OPERATION_OPEN: String = OPERATION + "opened"
   final val OPERATION_FAIL: String = OPERATION + "failed"
   final val OPERATION_TOTAL: String = OPERATION + "total"
-  final val OPERATION_STATE_RATE: String = OPERATION + "state_rate"
+  final val OPERATION_STATE: String = OPERATION + "state"
 
   final private val BACKEND_SERVICE = KYUUBI + "backend_service."
   final val BS_FETCH_LOG_ROWS_RATE = BACKEND_SERVICE + "fetch_log_rows_rate"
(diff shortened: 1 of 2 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
```diff
@@ -39,6 +39,7 @@ object MetricsConstants {
   final val OPERATION_OPEN: String = OPERATION + "opened"
   final val OPERATION_FAIL: String = OPERATION + "failed"
   final val OPERATION_TOTAL: String = OPERATION + "total"
+  final val OPERATION_STATE_RATE: String = OPERATION + "state_rate"
```
yaooqinn, line 42: is rate in a rate redundant？cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
```diff
@@ -158,4 +160,10 @@ abstract class KyuubiOperation(opType: OperationType, session: Session)
   }
 
   override def shouldRunAsync: Boolean = false
+
+  override def setState(newState: OperationState): Unit = {
+    super.setState(newState)
+    MetricsSystem.tracing(
+      _.markMeter(MetricRegistry.name(OPERATION_STATE_RATE, newState.toString.toLowerCase)))
```
yaooqinn, line 167: do we need toLowerCase? seems we are not doing this for OperationTypes

zhenjiaguo: Yes, we need. `OpertationType` is also use `toLowerCase`. Use low case to make them consistent.

kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
zhenjiaguo: Ok, I will remove it.

yaooqinn (approved):
LGTM

yaooqinn:
thanks, merged to master

yaooqinn:

PR #1813, opened 2022-01-20
Enable the internal secure authentication between Kyuubi server and engine

turboFei:

Enable the internal secure access between Kyuubi server and engine.

This pr use secret and cipher to encrypt a dynamic token identifier for authentication.


kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/SecureAccessor.scala
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  def authToken(tokenStr: String): Unit = {
+    if (supportSecureAccess()) {
+      val identifier = KyuubiInternalAccessIdentifier.fromJson(decrypt(tokenStr))
+      if (identifier.issueDate + identifier.maxDate > System.currentTimeMillis()) {
+        throw KyuubiSQLException("The engine access token is expired")
+      }
+    }
```
turboFei, line 78: is it ok for the internal secure access mechanism?

It is still WIP.

cc @yaooqinn

turboFei:
have tested it in our dev environment, it works well.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1070,6 +1070,59 @@ object KyuubiConf {
...
+    buildConf("engine.secure.token.max.lifetime")
+      .doc("The max lifetime of the token used for secure access between Kyuubi server and engine.")
+      .version("1.5.0")
+      .timeConf
+      .createWithDefault(Duration.ofMinutes(10).toMillis)
+
+  val ENGINE_SECURE_SECRET_PROVIDER: ConfigEntry[String] =
+    buildConf("engine.secure.secret.provider")
+      .doc("The class used to manage the engine secure access secret. This class must be a " +
+        "subclass of EngineSecureAccessSecretProvider.")
```
ulysses-you, line 1090: seems the class name `EngineSecureAccessSecretProvider` is wrong.

ulysses-you (commented):
It ssems the flag `isServer` is just a global static value rather than a class method. We can define the flag after server/engine started. cc @yaooqinn @pan3793

kyuubi-ha/src/main/scala/org/apache/kyuubi/service/authentication/ZooKeeperEngineSecuritySecretProviderImpl.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  private var conf: KyuubiConf = _
+
+  override def initialize(conf: KyuubiConf): Unit = {
+    this.conf = conf
+  }
+
+  override def getSecret(): String = {
+    var secret: String = ""
+    conf.get(HA_ZK_ENGINE_SECURE_SECRET_NODE).map { zkNode =>
```
ulysses-you, line 37: `assert (conf.get(HA_ZK_ENGINE_SECURE_SECRET_NODE).isDefined)`, avoid using `var secret` here

ulysses-you (commented):
What happens if the token is outdate ? we don't have a renew process. Seems it will cause user can not use Kyuubi until the session timeout ?

turboFei:
> What happens if the token is outdate ? we don't have a renew process. Seems it will cause user can not use Kyuubi until the session timeout ?


Now we issue a new token for each kyuubiSession.

ulysses-you:
thanks, merging to master

Issue #1796, opened 2022-01-19
[Bug] When the Spark application fails to submit, the session still waits for the timeout to exit


wForget:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Because `spark.yarn.submit.waitAppCompletion=false` is configured, the spark submit process exits normally before the spark application completes, so when the spark application fails, the session still waits for a timeout to exit.

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk
2022-01-19 08:50:01,593 ERROR [KyuubiThriftBinaryFrontendServiceHandler-Pool: Thread-13751]: org.apache.kyuubi.server.KyuubiThriftBinaryFrontendService(72) - Error opening session: 
org.apache.kyuubi.KyuubiSQLException: Error opening session for cupid_dp client ip *.*.*.*, due to org.apache.kyuubi.KyuubiSQLException: Timeout(43200000 ms) to launched SPARK_SQL engine with /*/bin/spark-submit \
        --class org.apache.kyuubi.engine.spark.SparkSQLEngine \
        --conf spark.kyuubi.session.engine.initialize.timeout=43200000 \
......
```

### Kyuubi Engine Log Output

```logtalk
22/01/18 20:51:59 ERROR ApplicationMaster: Uncaught exception: 
java.util.concurrent.TimeoutException: Futures timed out after [100000 milliseconds]
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:259)
	at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:263)
	at org.apache.spark.util.ThreadUtils$.awaitResult(ThreadUtils.scala:293)
	at org.apache.spark.deploy.yarn.ApplicationMaster.runDriver(ApplicationMaster.scala:504)
	at org.apache.spark.deploy.yarn.ApplicationMaster.run(ApplicationMaster.scala:268)
	at org.apache.spark.deploy.yarn.ApplicationMaster$$anon$3.run(ApplicationMaster.scala:899)
	at org.apache.spark.deploy.yarn.ApplicationMaster$$anon$3.run(ApplicationMaster.scala:898)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #1815, opened 2022-01-21
[KYUUBI #1796] Destroy the process after the engine starts

wForget:

Destroy the process after the engine starts. #1796

add conf: `kyuubi.session.engine.startup.waitCompletion  false`
image


turboFei (approved):
LGTM

It seems that the kyuubi server admin need set some limitation if this parameter is set to false.
Such as the `spark.master` and `deployMode` need be add into ignore.list in server side.
Otherwise, for yarn-client/local mode, the engine will be killed if the process is destroyed.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -694,6 +694,14 @@ object KyuubiConf {
       .checkValue(_ > 0, "the maximum must be positive integer.")
       .createWithDefault(10)
 
+  val SESSION_ENGINE_STARTUP_WAIT_COMPLETION: ConfigEntry[Boolean] =
+    buildConf("session.engine.startup.waitCompletion")
+      .doc("Whether to wait for completion after engine starts." +
```
yaooqinn, line 699: shall we make this config sounds more dangerous to adjust?

wForget: How about this description?

> Whether to wait for completion after engine starts. If false, the startup process will be destroyed after the engine is started. Notice, it is only used in yarn-cluster mode, for yarn-client/local mode, the engine will be killed.

turboFei: How about
```
Note that only use it when the driver is not running locally, such as yarn-cluster mode; otherwise, the engine will be killed.
```

wForget: > How about
> 
> > Note that only use it when the driver is not running locally, such as yarn-cluster mode; Otherwise, the engine will be killed.

+1, I wil add it

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -703,6 +703,16 @@ object KyuubiConf {
...
+        " If false, the startup process will be destroyed after the engine is started." +
+        " Note that only use it when the driver is not running locally," +
+        " such as yarn-cluster mode; Otherwise, the engine will be killed.")
```
SteNicholas, line 711: Does this note need to be shown in the log?

yaooqinn: yea, we shall add log when process.destroyForcibly()

wForget: Yes, I will add log

SteNicholas (approved):
LGTM.

yaooqinn (approved):

SteNicholas:
@pan3793, please help to merge.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -190,6 +191,10 @@ trait ProcBuilder {
     if (logCaptureThread != null) {
       logCaptureThread.interrupt()
     }
+    if (!waitCompletion && process != null) {
+      info("Destroy the process, since waitCompletion is false.")
+      process.destroyForcibly()
+    }
   }
```
pan3793, line 198: Let's synchronized the method `close()`

wForget: > Let's synchronized the method `close()` and reset `process` to `null`

Do we also need to reset `logCaptureThread` to `null`?

SteNicholas: > > Let's synchronized the method `close()` and reset `process` to `null`
> 
> Do we also need to reset `logCaptureThread` to `null`?

IMO, it's better to set `logCaptureThread` to null.

SteNicholas (approved):
LGTM. cc @pan3793

pan3793 (approved):

pan3793:
Thanks, merging to master

turboFei:
I had merged this pr and tested it on our test env.

It seems that it does not work.


I have to found another temporary workaround.

```
  protected val KYUUBI_PROC_UNIQUE_ID: String = "kyuubi.proc.uniqueId"

  protected lazy val procUniqueId = UUID.randomUUID().toString

  protected def procConf(): Map[String, String] = Map(KYUUBI_PROC_UNIQUE_ID -> procUniqueId)
```

```
    var allConf = conf.getAll ++ procConf()
```


```
  protected def killProcessByUniqueId(): Unit = {
    val psEf = Runtime.getRuntime.exec("ps -ef")
    val input = new BufferedReader(new InputStreamReader(psEf.getInputStream))
    val regex = s"$KYUUBI_PROC_UNIQUE_ID=$procUniqueId".r
    var psInfo = input.readLine()
    var found = false
    while (!found && psInfo != null) {
      regex findFirstIn (psInfo) match {
        case Some(_) =>
          found = true
          psInfo.trim.split("\\s+").drop(1).headOption.foreach { psId =>
            Runtime.getRuntime.exec(s"kill -9 $psId")
          }
        case None =>
      }
      psInfo = input.readLine()
    }
  }

```

Issue #1685, opened 2022-01-05
[SUB-TASK][KPIP-2] Support jar and lib start option


yanghua:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

This is a subtask of KPIP-2 #1322 .

### Motivation

Currently, for quick go through the main path of flink SQL engine. We did not support the `jar` and `lib` config options when starting. After we have time, it would be better to have these options. See: https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sqlclient/#sql-client-startup-options

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1818, opened 2022-01-21
[KYUUBI #1685] Support jar and lib start option

SteNicholas:

Support jar and lib start option.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
@@ -17,17 +17,24 @@
 
 package org.apache.kyuubi.engine.flink
 
+import java.io.File
+import java.net.URL
 import java.time.Instant
 import java.util.concurrent.CountDownLatch
 
 import scala.collection.JavaConverters._
+import scala.collection.mutable.ListBuffer
 
 import org.apache.flink.client.cli.{CliFrontend, CustomCommandLine, DefaultCLI}
-import org.apache.flink.configuration.{DeploymentOptions, GlobalConfiguration}
+import org.apache.flink.configuration.DeploymentOptions
+import org.apache.flink.configuration.GlobalConfiguration
+import org.apache.flink.table.client.SqlClientException
 import org.apache.flink.table.client.gateway.context.DefaultContext
+import org.apache.flink.util.JarUtils
 
-import org.apache.kyuubi.Logging
-import org.apache.kyuubi.Utils.{addShutdownHook, currentUser, FLINK_ENGINE_SHUTDOWN_PRIORITY}
+import org.apache.kyuubi.{KyuubiSQLException, Logging}
+import org.apache.kyuubi.Utils.{addShutdownHook, FLINK_ENGINE_SHUTDOWN_PRIORITY}
+import org.apache.kyuubi.Utils.currentUser
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.flink.FlinkSQLEngine.{countDownLatch, currentEngine}
 import org.apache.kyuubi.service.Serverable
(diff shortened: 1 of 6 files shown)


externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -103,4 +112,37 @@ object FlinkSQLEngine extends Logging {
...
+        JarUtils.checkJarFile(url)
+        dependencies = url :: dependencies
+      }
+      // find jar files in library directories
+      libraries.foreach { libUrl =>
+        val dir: File = new File(libUrl.toURI)
+        if (!dir.isDirectory) throw new SqlClientException("Directory expected: " + dir)
+        else if (!dir.canRead) throw new SqlClientException("Directory cannot be read: " + dir)
+        val files: Array[File] = dir.listFiles
+        if (files == null) throw new SqlClientException("Directory cannot be read: " + dir)
```
link3280, line 132: Should be "File cannot be read"?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -103,4 +112,37 @@ object FlinkSQLEngine extends Logging {
       addShutdownHook(() => engine.stop(), FLINK_ENGINE_SHUTDOWN_PRIORITY + 1)
     }
   }
+
+  private def discoverDependencies(
+      jars: Seq[URL],
+      libraries: Seq[URL]): List[URL] = {
+    try {
+      var dependencies: List[URL] = List()
```
link3280, line 120: I think a mutable list like `ListBuffer` might be better.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -103,4 +112,37 @@ object FlinkSQLEngine extends Logging {
...
+        if (!dir.isDirectory) throw new SqlClientException("Directory expected: " + dir)
+        else if (!dir.canRead) throw new SqlClientException("Directory cannot be read: " + dir)
+        val files: Array[File] = dir.listFiles
+        if (files == null) throw new SqlClientException("Directory cannot be read: " + dir)
+        files.foreach { f =>
+          // only consider jars
+          if (f.isFile && f.getAbsolutePath.toLowerCase.endsWith(".jar")) {
+            val url: URL = f.toURI.toURL
+            JarUtils.checkJarFile(url)
+            dependencies = url :: dependencies
```
link3280, line 138: Do we reverse the order of the files here by concatenating the elements at the head? I think the order matters, especially when there are potential class conflicts (maybe we should sort the files alphabetically as well).

link3280 (commented):
@SteNicholas Thanks a lot for your work! I left some comments, PTAL.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
SteNicholas: @link3280, this should be `Directory cannot be read`. Because the directory has no files.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
SteNicholas: @link3280, I don't thank the order matters because the potential class conflicts couldn't be avoided by the order.

SteNicholas:
@yaooqinn, @pan3793, could you please review this pull request?

yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #1816, opened 2022-01-21
[FEATURE] Implement KyuubiHistoryServerPlugin


wForget:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Implement KyuubiHistoryServerPlugin like HiveThriftServer2HistoryServerPlugin. #1812

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

wForget:
The implementation steps:
1. Add the @KVIndex property to the event, which will be used for KvStore
2. Refactor SparkSQLEngineListener, no longer use EngineEventsStore to store Events, but store them through ElementTrackingStore.
3. Add Trigger for cleanup events in SparkSQLEngineListener.
4. Refactor EngineEventsStore to get Events through ElementTrackingStore.
5. Implement KyuubiHistoryServerPlugin

cc @yaooqinn

yaooqinn:
After we log our events to spark's event logger, they become raw JSON data, why do we need `ElementTrackingStore` here?

wForget:
> After we log our events to spark's event logger, they become raw JSON data, why do we need `ElementTrackingStore` here?

The `ElementTrackingStore` is provided in the `org.apache.spark.sql.hive.thriftserver.ui.HiveThriftServer2HistoryServerPlugin#createListeners` method to initialize the SparkListener.

Previously, EngineEvents were stored in `org.apache.kyuubi.engine.spark.events.EngineEventsStore`, after using `ElementTrackingStore`, it will be stored in `org.apache.spark.util.kvstore.InMemoryStore`.

yaooqinn:
Oh, I get your point.  You will make a fat jar based on the engine module and then put it into the history server classpath

yaooqinn:
You can give it a try. But I don't know whether the private package changes so far will break something unexpectedly

yaooqinn:
the currently spark engine module is 'once compiled and works for all spark 3.x', it's a long effort we have made so far and good luck with not breaking the it

wForget:
> the currently spark engine module is 'once compiled and works for all spark 3.x', it's a long effort we have made so far and good luck with not breaking the it

[...]

PR #1820, opened 2022-01-21
[KYUUBI #1816] Implement KyuubiHistoryServerPlugin

wForget:

Implement KyuubiHistoryServerPlugin. #1816

**SparkUI:**
sparkui001
sparkui002

**SparkHistoryServer:**
history001
history002


wForget:
Hi @yaooqinn , this implementation seems feasible. Can you help me review it?

yaooqinn:
thanks, @wForget for the great job, will take a look tomorrow

yaooqinn:
also cc @turboFei @ulysses-you

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineTab.scala
```diff
@@ -24,25 +24,42 @@ import scala.util.control.NonFatal
...
-  engine.spark.sparkContext.ui.foreach { ui =>
+  val startTime = engine.map(_.getStartTime).getOrElse {
+    sparkUI
+      .map(ui => ui.store.applicationInfo().attempts.head.startTime.getTime)
+      .getOrElse(0L)
+  }
+
+  def endTime(): Long = engine.map(_ => System.currentTimeMillis()).getOrElse {
+    sparkUI
+      .map(ui => ui.store.applicationInfo().attempts.head.endTime.getTime)
```
yaooqinn, line 51: why the start/end Time always use the first attempt？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkSQLEngineEventListener.scala
```diff
@@ -0,0 +1,111 @@
...
+class SparkSQLEngineEventListener(
+    kvstore: ElementTrackingStore,
+    sparkConf: SparkConf,
+    kyuubiConf: KyuubiConf) extends SparkListener {
+
+  /**
+   * The number of SQL client sessions kept in the Kyuubi Query Engine web UI.
+   */
+  private val retainedSessions: Int = {
+    sparkConf.getOption(SQLConf.THRIFTSERVER_UI_SESSION_LIMIT.key) match {
```
yaooqinn, line 38: shall we simply use kyuubiConf？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/KyuubiHistoryServerPlugin.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.spark.kyuubi.SparkSQLEngineEventListener
+import org.apache.spark.scheduler.SparkListener
+import org.apache.spark.status.{AppHistoryServerPlugin, ElementTrackingStore}
+
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.engine.spark.events.EngineEventsStore
+
+class KyuubiHistoryServerPlugin extends AppHistoryServerPlugin {
+
+  def kyuubiConf: KyuubiConf = KyuubiConf()
```
yaooqinn, line 30: private?

yaooqinn, line 30: shall we also load kyuubi configurations from sparkConf like SparkSQLEngine?

yaooqinn:
> Hi @yaooqinn , this implementation seems feasible. Can you help me review it?

sorry for the late reply, plans were disrupted by omicron in HZ.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineTab.scala
wForget: I see that the spark source code also uses `attempts.head` to get attempt.
like: https://github.com/apache/spark/blob/2a9416abd5c8d83901c09a08433bf91e649dc43f/core/src/main/scala/org/apache/spark/ui/jobs/AllJobsPage.scala#L277

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/KyuubiHistoryServerPlugin.scala
wForget: Sounds good, I will load the kyuubi configurations from sparkConf and only use kyuubiConf.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/KyuubiHistoryServerPlugin.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import org.apache.spark.SparkConf
+import org.apache.spark.kyuubi.SparkSQLEngineEventListener
+import org.apache.spark.scheduler.SparkListener
+import org.apache.spark.status.{AppHistoryServerPlugin, ElementTrackingStore}
+
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.engine.spark.events.EngineEventsStore
+
+class KyuubiHistoryServerPlugin extends AppHistoryServerPlugin {
```
ulysses-you, line 28: Can we add some comments about how to use this class ? It should work with history server together.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/KyuubiHistoryServerPlugin.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  private val kyuubiConf: KyuubiConf = KyuubiConf()
+
+  override def createListeners(conf: SparkConf, store: ElementTrackingStore): Seq[SparkListener] = {
+    addKyuubiConfFromSpark(conf)
+    Seq(new SparkSQLEngineEventListener(store, kyuubiConf))
+  }
+
+  override def setupUI(ui: SparkUI): Unit = {
+    addKyuubiConfFromSpark(ui.conf)
```
ulysses-you, line 38: it looks like we should build kyuubi conf for every Spark app UI instead of use a global kyuubi conf.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineSessionPage.scala
```diff
@@ -55,23 +55,30 @@ case class EngineSessionPage(parent: EngineTab)
...
-        {new Date(parent.engine.backendService.sessionManager.latestLogoutTime)}
-      </li>
-      <li>
-        <strong>Time since start: </strong>
-        {formatDurationVerbose(timeSinceStart)}
-      </li>
-    </ul>
-  }
+  private def generateBasicStats(): Seq[Node] =
+    if (parent.engine != null) {
```
ulysses-you, line 59: the type of `parent.engine` is Option ?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/KyuubiHistoryServerPlugin.scala
wForget: > it looks like we should build kyuubi conf for every Spark app UI instead of use a global kyuubi conf.

These sparkConf seem to be the same object, should we create multiple kyuubiConf?

wForget: If we need to get the configuration of different spark app, we can get it using `org.apache.spark.status.AppStatusStore#environmentInfo`, do we need to do that?

ulysses-you: > These sparkConf seem to be the same object, should we create multiple kyuubiConf?

I just check the Spark build-in plugin, do not see a global value. The code is confused if the input conf is always same. It's better to craete multiple kyuubiConf.

> If we need to get the configuration of different spark app, we can get it using org.apache.spark.status.AppStatusStore#environmentInfo, do we need to do that?

It seems trivial, we dont need to update.

yaooqinn (approved):

ulysses-you:
thanks, merging to master

theoctober19th:
Hi, do we have a documentation for how to use this extension yet?

wForget:
> Hi, do we have a documentation for how to use this extension yet?

There is usage in comment of `KyuubiHistoryServerPlugin.scala`

https://github.com/apache/kyuubi/blob/c8e645734b0cee93627741c24d3e9d8ff16a4e11/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/KyuubiHistoryServerPlugin.scala#L29-L34

wForget:
Just copy `kyuubi-spark-sql-engine.jar` to `$SPARK_HOME/jars` of `SparkHistoryServer`.

Issue #1778, opened 2022-01-17
[SUB-TASK][KPIP-2] Support Flink Set/Reset operation


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Set/Reset operations are crucial for tunning Flink SQL jobs and enabling/disabling features, but they're not executed as SQL statements in Flink, thus can't be supported by the current `ExecuteStatement` implementation. We should extend `ExecuteStatement` to support these operations.

### Motivation

This is a sub-task of KPIP-2 #1322.

### Describe the solution

Add matching cases for SetOperation and ResetOperation, and execute them via other table APIs.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

link3280:
cc @yanghua

yanghua:
> cc @yanghua

+1

PR #1822, opened 2022-01-21
[KYUUBI #1778] Support Flink Set/Reset Operations

link3280:

Set/Reset operations are crucial for tunning Flink SQL jobs and enabling/disabling features, but they're not executed as SQL statements in Flink, thus can't be supported by the current ExecuteStatement implementation. We should extend ExecuteStatement to support these operations.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -28,11 +28,11 @@ import org.apache.flink.table.api.{DataTypes, ResultKind}
 import org.apache.flink.table.catalog.Column
 import org.apache.flink.table.client.gateway.{Executor, TypedResult}
 import org.apache.flink.table.operations.{Operation, QueryOperation}
-import org.apache.flink.table.operations.command.SetOperation
+import org.apache.flink.table.operations.command.{ResetOperation, SetOperation}
 import org.apache.flink.types.Row
 
 import org.apache.kyuubi.{KyuubiSQLException, Logging}
-import org.apache.kyuubi.engine.flink.result.ResultSet
+import org.apache.kyuubi.engine.flink.result.{OperationUtil, ResultSet}
 import org.apache.kyuubi.operation.{OperationState, OperationType}
 import org.apache.kyuubi.operation.log.OperationLog
 import org.apache.kyuubi.session.Session
@@ -106,6 +106,7 @@ class ExecuteStatement(
       operation match {
         case queryOperation: QueryOperation => runQueryOperation(queryOperation)
         case setOperation: SetOperation => runSetOperation(setOperation)
+        case resetOperation: ResetOperation => runResetOperation(resetOperation)
         case operation: Operation => runOperation(operation)
       }
     } catch {
@@ -167,7 +168,7 @@ class ExecuteStatement(
       val entries = ArrayBuffer.empty[Row]
       properties.forEach((key, value) => entries.append(Row.of(PROPERTY_FORMAT.format(key, value))))
 
-      if (entries.nonEmpty ) {
+      if (entries.nonEmpty) {
         val prettyEntries = entries.sortBy(_.getField(0).asInstanceOf[String])
         resultSet = ResultSet.builder
           .resultKind(ResultKind.SUCCESS_WITH_CONTENT)
@@ -182,6 +183,18 @@ class ExecuteStatement(
           .build
       }
     }
+    setState(OperationState.FINISHED)
+  }
+
+  private def runResetOperation(resetOperation: ResetOperation): Unit = {
+    if (resetOperation.getKey.isPresent) {
+      // reset the given property
+      executor.resetSessionProperty(sessionId, resetOperation.getKey.get())
+    } else {
+      // reset all properties
+      executor.resetSessionProperties(sessionId)
+    }
+    resultSet = OperationUtil.successResultSet()
     setState(OperationState.FINISHED)
   }
 
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -50,8 +50,6 @@ class ExecuteStatement(
 
   private var statementTimeoutCleaner: Option[ScheduledExecutorService] = None
 
-  private val PROPERTY_FORMAT = "%s = %s"
-
   override def getOperationLog: Option[OperationLog] = Option(operationLog)
 
   @VisibleForTesting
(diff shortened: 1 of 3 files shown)


link3280:
cc @yaooqinn @yanghua @pan3793

link3280:
ping @yanghua @pan3793 @yaooqinn

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -139,6 +146,58 @@ class ExecuteStatement(
...
+  private def runSetOperation(setOperation: SetOperation): Unit = {
+    if (setOperation.getKey.isPresent) {
+      val key: String = setOperation.getKey.get.trim
+
+      if (setOperation.getValue.isPresent) {
+        val newValue: String = setOperation.getValue.get.trim
+        executor.setSessionProperty(sessionId, key, newValue)
+      }
+
+      val value = executor.getSessionConfigMap(sessionId).getOrDefault(key, "")
```
yaooqinn, line 158: So looks that we support `set`, `set k=v` and `set k`, and the last one is for retrieving a single value from a matched key? shall we also test it?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -139,6 +146,58 @@ class ExecuteStatement(
...
+      val properties: util.Map[String, String] = executor.getSessionConfigMap(sessionId)
+
+      val entries = ArrayBuffer.empty[Row]
+      properties.forEach((key, value) => entries.append(Row.of(PROPERTY_FORMAT.format(key, value))))
+
+      if (entries.nonEmpty) {
+        val prettyEntries = entries.sortBy(_.getField(0).asInstanceOf[String])
+        resultSet = ResultSet.builder
+          .resultKind(ResultKind.SUCCESS_WITH_CONTENT)
+          .columns(Column.physical("set", DataTypes.STRING()))
```
yaooqinn, line 175: Is it more convenient to return 2 columns, one for the key and the other for the value?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -139,6 +146,58 @@ class ExecuteStatement(
...
+
+  private def runResetOperation(resetOperation: ResetOperation): Unit = {
+    if (resetOperation.getKey.isPresent) {
+      // reset the given property
+      executor.resetSessionProperty(sessionId, resetOperation.getKey.get())
+    } else {
+      // reset all properties
+      executor.resetSessionProperties(sessionId)
+    }
+    resultSet = OperationUtil.successResultSet()
```
yaooqinn, line 197: Does it make more sense to return the original value to which we reset the key?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: > Is it more convenient to return 2 columns, one for the key and the other for the value?

Yes, I thought about this as well. But I found spark/hive/flink are returning key-value pairs, so I adopt the common format.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: Yes, but Flink doesn't support `set k` at the moment, so I skip the test for it.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: @yaooqinn I think it's a valid consideration, but now hive returns nothing for `reset` while flink returns `OK` (which is the same with some queries like DDL), so I think it's okay to be aligned with them. WDYT?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -139,6 +146,58 @@ class ExecuteStatement(
...
+        .resultKind(ResultKind.SUCCESS_WITH_CONTENT)
+        .columns(Column.physical("set", DataTypes.STRING()))
+        .data(Array(Row.of(PROPERTY_FORMAT.format(key, value))))
+        .build
+    } else {
+      // show all properties if set without key
+      val properties: util.Map[String, String] = executor.getSessionConfigMap(sessionId)
+
+      val entries = ArrayBuffer.empty[Row]
+      properties.forEach((key, value) => entries.append(Row.of(PROPERTY_FORMAT.format(key, value))))
```
yaooqinn, line 169:
```java
private void callSet(SetOperation setOperation) {
        // set a property
        if (setOperation.getKey().isPresent() && setOperation.getValue().isPresent()) {
            String key = setOperation.getKey().get().trim();
            String value = setOperation.getValue().get().trim();
            executor.setSessionProperty(sessionId, key, value);
            printInfo(MESSAGE_SET_KEY);
        }
        // show all properties
        else {
            final Map<String, String> properties = executor.getSessionConfigMap(sessionId);
            if (properties.isEmpty()) {
                terminal.writer()
                        .println(CliStrings.messageInfo(CliStrings.MESSAGE_EMPTY).toAnsi());
            } else {
                List<String> prettyEntries = new ArrayList<>();
                for (String key : properties.keySet()) {
                    prettyEntries.add(
                            String.format(
                                    "'%s' = '%s'",
                                    EncodingUtils.escapeSingleQuotes(key),
                                    EncodingUtils.escapeSingleQuotes(properties.get(key))));
                }
                prettyEntries.sort(String::compareTo);
                prettyEntries.forEach(entry -> terminal.writer().println(entry));
            }
            terminal.flush();
        }
    }
```


the string formater is not exactly the same as the one of flink, any special reason?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
yaooqinn: will `set k` passes the parsing phase or fail early before entering this func?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
yaooqinn: I am OK with this if flink takes `OK` as its side effect for all DDLs

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: Sorry, I was executing the set query via our in-house platform, not native spark-sql. Maybe I get the result because the platform is using beeline. I'm okay with two separate columns.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: @yaooqinn I just thought spark/hive doesn't wrap the key-values with quotes. I'm okay to make it align with Flink.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: Yes, it will cause a SQL parse error.

yaooqinn: then I am OK with this

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
yaooqinn: Let's return 2 columns then. Actually, for CLIs like beeline, users may not have any opportunity to handle results col-by-col, but for callers of JDBC-API, 2 columns make a big improvement on UX.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
yaooqinn: OK, Let's keep it in your way

yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #1825, opened 2022-01-22
[SUB-TASK][KPIP-2] Generate appName for Flink applications


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Generate Flink application names according to the pattern in Spark engine.

### Motivation

This is a sub-task of KPIP-2 #1322.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

link3280:
cc @yanghua

PR #1827, opened 2022-01-22
[KYUUBI #1825] Generate appName for Flink applications

link3280:

Generate Flink application names for per-job clusters and application-mode clusters with respect to the pattern in Spark engine.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
@@ -89,8 +89,6 @@ object FlinkSQLEngine extends Logging {
             val appName = s"kyuubi-${user}-flink-${Instant.now}"
             flinkConf.setString("kubernetes.cluster-id", appName)
           }
-        case null =>
-          error("No execution.target specified for Flink engine")
         case other =>
           debug(s"Skip generating app name for execution target $other")
       }
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
@@ -17,16 +17,17 @@
 
 package org.apache.kyuubi.engine.flink
 
+import java.time.Instant
 import java.util.concurrent.CountDownLatch
 
 import scala.collection.JavaConverters._
 
 import org.apache.flink.client.cli.{CliFrontend, CustomCommandLine, DefaultCLI}
-import org.apache.flink.configuration.GlobalConfiguration
+import org.apache.flink.configuration.{DeploymentOptions, GlobalConfiguration}
 import org.apache.flink.table.client.gateway.context.DefaultContext
 
 import org.apache.kyuubi.Logging
-import org.apache.kyuubi.Utils.{addShutdownHook, FLINK_ENGINE_SHUTDOWN_PRIORITY}
+import org.apache.kyuubi.Utils.{addShutdownHook, currentUser, FLINK_ENGINE_SHUTDOWN_PRIORITY}
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.flink.FlinkSQLEngine.{countDownLatch, currentEngine}
 import org.apache.kyuubi.service.Serverable
@@ -62,6 +63,8 @@ object FlinkSQLEngine extends Logging {
   val kyuubiConf: KyuubiConf = KyuubiConf()
   var currentEngine: Option[FlinkSQLEngine] = None
 
+  private val user = currentUser
+
   private val countDownLatch = new CountDownLatch(1)
 
   def main(args: Array[String]): Unit = {
@@ -72,6 +75,20 @@ object FlinkSQLEngine extends Logging {
     try {
       val flinkConfDir = CliFrontend.getConfigurationDirectoryFromEnv
       val flinkConf = GlobalConfiguration.loadConfiguration(flinkConfDir)
+
+      val executionTarget = flinkConf.getString(DeploymentOptions.TARGET)
+      // set cluster name for per-job and application mode
+      executionTarget match {
+        case "yarn-per-job" | "yarn-application" =>
+          val appName = s"kyuubi_${user}_flink_${Instant.now}"
+          flinkConf.setString("yarn.application.name", appName)
+        case "kubernetes-application" =>
+          val appName = s"kyuubi-${user}-flink-${Instant.now}"
+          flinkConf.setString("kubernetes.cluster-id", appName)
+        case null =>
+          error("No execution.target specified for Flink engine")
+      }
+
       val engineContext = new DefaultContext(
         List.empty.asJava,
         flinkConf,
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
@@ -80,11 +80,15 @@ object FlinkSQLEngine extends Logging {
       // set cluster name for per-job and application mode
       executionTarget match {
         case "yarn-per-job" | "yarn-application" =>
-          val appName = s"kyuubi_${user}_flink_${Instant.now}"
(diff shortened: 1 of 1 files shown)


externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -72,6 +75,20 @@ object FlinkSQLEngine extends Logging {
     try {
       val flinkConfDir = CliFrontend.getConfigurationDirectoryFromEnv
       val flinkConf = GlobalConfiguration.loadConfiguration(flinkConfDir)
+
+      val executionTarget = flinkConf.getString(DeploymentOptions.TARGET)
+      // set cluster name for per-job and application mode
+      executionTarget match {
+        case "yarn-per-job" | "yarn-application" =>
```
yanghua, line 82: Will we consider `yarn-session`?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -72,6 +75,20 @@ object FlinkSQLEngine extends Logging {
...
+          val appName = s"kyuubi_${user}_flink_${Instant.now}"
+          flinkConf.setString("yarn.application.name", appName)
+        case "kubernetes-application" =>
+          val appName = s"kyuubi-${user}-flink-${Instant.now}"
+          flinkConf.setString("kubernetes.cluster-id", appName)
+        case null =>
```
yanghua, line 88: Currently, we only support flink local(standalone) mode. Can we only print the target for now?

yanghua (commented):
@link3280 Thanks, left some comments.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
link3280: IIUC, in session mode we need to start a yarn session beforehand, so the cluster name/id is already determined, so here only includes application-mode cluster and per-job cluster, which is initialized lazily.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
link3280: Yes, we don't need to set app names for standalone clusters. 

Do you mean for now we should only validate and print the value of the execution target to ensure we are in standalone mode? But previously I see discussions saying that we are aiming for the application mode. Am I missing something?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
yanghua: yes, both for session and yarn-session are pre-started.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
yanghua: > Do you mean for now we should only validate and print the value of the execution target to ensure we are in standalone mode? 

Yes

> But previously I see discussions saying that we are aiming for the application mode. 
also yes

> Am I missing something?
No, nothing you have missed. I just suggest we say that it's an error after we would support application mode？ WDYT？

yanghua: After rethink about this. It's also OK for me about the current behavior.

yanghua: After rethink about this. It's also OK for me about the current behavior.

yanghua (approved):
LGTM. @link3280 thanks for your contribution.

link3280:
@yanghua Thanks a lot for your review!

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -72,6 +75,20 @@ object FlinkSQLEngine extends Logging {
...
+          error("No execution.target specified for Flink engine")
+      }
```
pan3793, line 90: we need a `case _ =>` branch to handle other cases. BTW, can `executionTarget` be `null`?

link3280: Oh, I thought we could bypass session clusters and standalone clusters because they already have cluster names, so what a `case _ =>` branch can do is simply validate the values. Is this what you expect? 

> BTW, can executionTarget be `null`?

The default value is `null`, which would cause a validation error on job submission.

SteNicholas (changes requested):

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -72,6 +75,20 @@ object FlinkSQLEngine extends Logging {
     try {
       val flinkConfDir = CliFrontend.getConfigurationDirectoryFromEnv
       val flinkConf = GlobalConfiguration.loadConfiguration(flinkConfDir)
+
+      val executionTarget = flinkConf.getString(DeploymentOptions.TARGET)
+      // set cluster name for per-job and application mode
+      executionTarget match {
```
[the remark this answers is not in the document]
yanghua: @SteNicholas Do you still have concerns?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -72,6 +75,20 @@ object FlinkSQLEngine extends Logging {
...
+        case "yarn-per-job" | "yarn-application" =>
+          val appName = s"kyuubi_${user}_flink_${Instant.now}"
+          flinkConf.setString("yarn.application.name", appName)
```
yaooqinn, line 84: Does it need to be `set if unset`?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
pan3793: if no `case _ =>` branch, the program will crash when receive the unmatched value
image

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
link3280: Yes, it'd be better not to override the existing value. Updating the PR.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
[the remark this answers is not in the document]
SteNicholas: @yanghua , my concern is that if the `remote` and `local` execution target have no any behavior, the null execution target is same as the `remote` and `local`.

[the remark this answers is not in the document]
link3280: @SteNicholas How about removing the `null ` case?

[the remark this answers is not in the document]
SteNicholas: > @SteNicholas How about removing the `null` case?

@link3280, I agree with this point.

SteNicholas (approved):
LGTM.

yaooqinn:
thanks, merged to master

PR #1828, opened 2022-01-22
[KYUUBI #1786] Fix the Spark sql engine logger level changed to WARN issue

### LINKED ISSUE #1786: [Bug] spark.repl.class.outputDir using Main.outputDir.getAbsolutePath will change engine log to WARN
### Code of Conduct

### Search before asking

### Describe the bug

Main object will init logger with isInterpreter=true, which changes logger to WARN level

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

turboFei:

To close #1786 
Main object will init logger with isInterpreter=true, which changes logger to WARN level.
In this pr, we do not use Main.output as `spark.repl.class.outputDir`
Passed existing tests.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -75,7 +75,7 @@ object SparkSQLEngine extends Logging {
 
   private val countDownLatch = new CountDownLatch(1)
 
-  val replOutputDir = Utils.createTempDir(namePrefix = "repl").toFile
+  val replOutputDir = Utils.createTempDir(namePrefix = "repl").toFile.getAbsolutePath
 
   def createSpark(): SparkSession = {
     val sparkConf = new SparkConf()
@@ -85,7 +85,7 @@ object SparkSQLEngine extends Logging {
     sparkConf.setIfMissing("spark.ui.port", "0")
     // register the repl's output dir with the file server.
     // see also `spark.repl.classdir`
-    sparkConf.set("spark.repl.class.outputDir", replOutputDir.getAbsolutePath)
+    sparkConf.set("spark.repl.class.outputDir", replOutputDir)
     sparkConf.setIfMissing(
       "spark.hadoop.mapreduce.input.fileinputformat.list-status.num-threads",
       "20")
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -24,10 +24,9 @@ import scala.util.control.NonFatal
 
 import org.apache.spark.{ui, SparkConf}
 import org.apache.spark.kyuubi.SparkSQLEngineListener
-import org.apache.spark.repl.Main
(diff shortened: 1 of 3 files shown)


yaooqinn:
this is a direct fix that I have to admit. But any chance that we fix it by not using Main？

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -76,6 +75,8 @@ object SparkSQLEngine extends Logging {
 
   private val countDownLatch = new CountDownLatch(1)
 
+  val replOutputDir = Utils.createTempDir(namePrefix = "repl").toFile
```
SteNicholas, line 78: Why not directly `val replOutputDir = Utils.createTempDir(namePrefix = "repl").toFile. getAbsolutePath`?

SteNicholas (changes requested):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
turboFei: thanks

SteNicholas (approved):
LGTM. @yaooqinn, could you please take a look and merge this pull request?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -76,6 +75,8 @@ object SparkSQLEngine extends Logging {
 
   private val countDownLatch = new CountDownLatch(1)
 
+  val replOutputDir = Utils.createTempDir(namePrefix = "repl").toFile.getAbsolutePath
```
yaooqinn, line 78: It will always create a temp dir in `java.io.tmpDir`, shall we do as same as what spark does?

```
val rootDir = conf.getOption("spark.repl.classdir").getOrElse(Utils.getLocalDir(conf))
  val outputDir = Utils.createTempDir(root = rootDir, namePrefix = "repl")
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala
```diff
@@ -40,7 +42,7 @@ private[spark] case class KyuubiSparkILoop private (
     val interpArguments = List(
       "-Yrepl-class-based",
       "-Yrepl-outdir",
-      s"${Main.outputDir.getAbsolutePath}")
+      s"${SparkSQLEngine.replOutputDir}")
```
yaooqinn, line 45: we can retrieve it from sc.conf

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
turboFei: org.apache.spark.util.Utils is only accessible in package `org.apache.spark`

yaooqinn (approved):

turboFei:
thanks @SteNicholas @yaooqinn, merging to master

PR #1831, opened 2022-01-24
[DOCS] Fix the jdbc connection url format doc

turboFei:

Fix the jdbc connection url format doc.

In fact, the kyuubiConfs part and kyuubiVars part should have the same effect.
But it is better to split the engine initialization conf and runtime vars to two parts, so here I do not modify these two parts.


yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #1834, opened 2022-01-25
[BUILD] Add support of external mvn with default mvn jvm opts when making distribution

goldenbean:

most of time, we use the mvn pre-installed in the OS 

After running the script `build/dist`, it should echo the 'MVN' location before building the release package

```bash
build/dist --tgz --name release-demo \
        --spark-provided --flink-provided \
        --mvn $(which mvn) \
        -DskipTests \
        -Pjava-8 -Pspark-hadoop-3.2 -Pspark-3.0
```


build/dist
```diff
@@ -68,6 +68,10 @@ while (( "$#" )); do
     --spark-provided)
       SPARK_PROVIDED=true
       ;;
+    --mvn)
```
pan3793, line 71: let's also update usage

pan3793:
> most of time, we use the mvn pre-installed in the OS

FYI, the `build/mvn` will fallback to `mvn` command if your pre-installed mvn version equals or greater than `maven.version` defined in `pom.xml` rather than download new one. We don't recommand use the lower maven version.

yaooqinn (approved):

build/dist
goldenbean: sure

build/dist
```diff
@@ -41,12 +41,13 @@ function usage {
   echo ""
   echo "Usage:"
   echo "+--------------------------------------------------------------------------------------+"
-  echo "| ./build/dist [--name <custom_name>] [--tgz] [--flink-provided] [--spark-provided] <maven build options> |"
+  echo "| ./build/dist [--name <custom_name>] [--tgz] [--flink-provided] [--spark-provided] [--mvn <maven_executable>] <maven build options> |"
```
pan3793, line 44: Let's break it into multi-lines, and keep it a good format

goldenbean:
> > most of time, we use the mvn pre-installed in the OS
> 
> FYI, the `build/mvn` will fallback to `mvn` command if your pre-installed mvn version equals or greater than `maven.version` defined in `pom.xml` rather than download new one. We don't recommand to use the lower maven version.

Hi @pan3793 ,  thank you for the comments. We currently use only the maven 3.6.3 to build Kyuubi, and deploy in one of the staging environments. It seems fine for now. :) 

If there were some issues, I will follow the recommendation.

pan3793 (approved):

pan3793:
Thanks @goldenbean, merging to master. And for maven versions, we encounter some Scala compile issues and maven shade plugin issue in past, and sometimes bump maven version can solve.

PR #1769, opened 2022-01-16
[BUILD] Migrate from log4j1 to log4j2

turboFei:
### _Why are the changes needed?_

This patch proposes to migrate from log4j1 to log4j2 in Kyuubi.
Refer the spark patch: https://github.com/apache/spark/pull/34895

### Does this PR introduce any user-facing change?
Yes. Users may need to rewrite their log4j properties file for log4j2. As of version 2.4, log4j now supports configuration via properties files. Note that the property syntax is not the same as the syntax used in log4j 1, but during the migration I found the syntax is pretty close so the migration should be straightforward.

### _How was this patch tested?_
Passed all existing tests.

PR #1842, opened 2022-01-26
[KYUUBI #1769][FOLLOWUP] Some cleanups for log4j properties

turboFei:

This is a followup of #1769. Cleanup the log4j properties files and using log4j2.properties for UT.

Passed UT.


yaooqinn:
and the licenses?

turboFei:
> and the licenses?

Sorry, which file should I modify?

yaooqinn:
LICENSE-binary

yaooqinn (approved):

pan3793 (approved):

LICENSE-binary
```diff
@@ -247,7 +247,6 @@ org.apache.thrift:libfb303
 org.apache.thrift:libthrift
 org.webjars:swagger-ui
 org.yaml:snakeyaml
-log4j:log4j
```
yaooqinn, line 250: it still on our `depenencyList` can we also remove it there?

turboFei:
```
    <dependencyManagement>
        <dependencies>
            <dependency>
                <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
                <exclusions>
                    <exclusion>
                        <groupId>log4j</groupId>
                        <artifactId>log4j</artifactId>
                    </exclusion>
                </exclusions>
            </dependency>

...
    </dependencyManagement>
```

pom.xml
```diff
@@ -240,6 +240,18 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
+                <groupId>org.apache.kyuubi</groupId>
```
yaooqinn, line 244: emm?

turboFei: Without this, the log4j-1.2.17 dependency will be transferred.

yaooqinn: to be clear, Do you mean that the common module needs log4j-1.2.17 but the server module not? so you need to exclustion?

turboFei: The kyuubi-common module is depended by other modules, such as kyuubi_assembly.

For command, `mvn dependency:tree -pl :kyuubi-common_2.12|grep 1.2.17`, the result is empty.


But if you execute below command:
```
mvn dependency:tree |grep  -C 20 1.2.17
```

You can find that the log4j-1.2.17 is from kyuubi-common module.

turboFei: So, I think there is some dependency transfer issue.


And declare the kyuubi-common in dependencyManagement can help fix this issue.

yaooqinn: why not exclude it in common？

turboFei: let me have a try.

turboFei:
thanks, merged to master

PR #1845, opened 2022-01-27
[KYUUBI #1829] Support executing statement async for Flink SQL engine

SteNicholas:

Support executing statement async for Flink SQL engine.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -24,6 +24,7 @@ import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 
 import com.google.common.annotations.VisibleForTesting
+import org.apache.calcite.rel.metadata.{DefaultRelMetadataProvider, JaninoRelMetadataProvider, RelMetadataQueryBase}
 import org.apache.flink.table.api.{DataTypes, ResultKind}
 import org.apache.flink.table.catalog.Column
 import org.apache.flink.table.client.gateway.{Executor, TypedResult}
@@ -75,14 +76,7 @@ class ExecuteStatement(
       val asyncOperation = new Runnable {
         override def run(): Unit = {
           OperationLog.setCurrentOperationLog(operationLog)
-          val sessionHandle = session.sessionManager.openSession(
-            session.protocol,
-            session.user,
-            session.password,
-            session.ipAddress,
-            session.conf)
           executeStatement()
-          session.sessionManager.closeSession(sessionHandle)
         }
       }
 
@@ -107,6 +101,9 @@ class ExecuteStatement(
     try {
       setState(OperationState.RUNNING)
 
+      // set the thread variable THREAD_PROVIDERS
+      RelMetadataQueryBase.THREAD_PROVIDERS.set(
+        JaninoRelMetadataProvider.of(DefaultRelMetadataProvider.INSTANCE))
       val operation = executor.parseStatement(sessionId, statement)
       operation match {
         case queryOperation: QueryOperation => runQueryOperation(queryOperation)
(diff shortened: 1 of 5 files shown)


SteNicholas:
@pan3793 @yanghua @yaooqinn , could you please help to review this pull request?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -62,7 +62,7 @@ abstract class SparkOperation(opType: OperationType, session: Session)
   protected def cleanup(targetState: OperationState): Unit = state.synchronized {
     if (!isTerminalState(state)) {
       setState(targetState)
-      Option(getBackgroundHandle).foreach(_.cancel(true))
+      if (shouldRunAsync) Option(getBackgroundHandle).foreach(_.cancel(true))
```
hddong, line 65: Wrong file? This is `SparkOperation`

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
```diff
@@ -323,9 +323,6 @@ class SparkOperationSuite extends WithSparkSQLEngine with HiveMetadataTests with
 
   test("basic open | execute | close") {
     withThriftClient { client =>
-      val operationManager =
```
hddong, line 326: ditto

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
SteNicholas: The `cleanup` of the `SparkOperation` doesn't check the `shouldRunAsync` to cancel the background handle. Therefore I fixed the logic in this pull request.

hddong: Seems the check is dispensable, `Option(getBackgroundHandle)` is None when sync execute.

SteNicholas: @hddong, as you mentioned, the `cleanup` of the `FlinkOperation` doesn't also check the `shouldRunAsync`, right?

hddong: @SteNicholas : I'm not clear for `FlinkOperation`, just mean the change here is unnecessary.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -75,11 +75,18 @@ class ExecuteStatement(
       val asyncOperation = new Runnable {
         override def run(): Unit = {
           OperationLog.setCurrentOperationLog(operationLog)
+          val sessionHandle = session.sessionManager.openSession(
```
yaooqinn, line 78: we only need executeStatement here

SteNicholas: @yaooqinn, I don't think this only need executeStatement because the `THREAD_PROVIDERS` thread local get nothing in Flink.

yaooqinn: I do not follow you. However，there will not be any chance that we call openSession in an operation

yaooqinn (approved):
lgtm

hddong (approved):

Issue #1798, opened 2022-01-19
[FEATURE] Add EventBus module to unify the distribution and subscription of Kyuubi's events 


iodone:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] #1799

### Describe the feature

1. Add EventBus module to unify the distribution and subscription of Kyuubi's events.
2. The addition of the EventBus module will not have task-intrusive changes to the existing business logic.

### Motivation

The EventBus concept was introduced to solve the following problems.
1. Kyuubi Event does not implement a unified event distribution logic and needs to rely on the event's subscribers to implement the receiving logic (`EventLoggerService.onEvent`). If you need to access many subscribers like EventLoggerService, it will result in calling the subscriber's methods for receiving events one by one. 
2. `EventLoggerService` should only focus on processing events and register to receive events of the corresponding type in a unified place, without coupling in each business logic to call onEvent to receive events. 
3. If we follow EventLoggerSevice's way to implement #908  to add AuditLog, we will intrusively modify the `EventLoggerService` code because we need to add the `onAuditEvent` method to receive the corresponding new events.
4. Provides a unified asynchronous event handling model.

### Describe the solution

EventBus Implemented in the following ways:
1. Decouple the distribution and subscription logic of events through EventBus. The subscribed events can be similar to EventLogService pattern or any other event processing, not necessarily event logger related. 
2. Adding EventBus will not change the existing EventLoggerService logic, just register `EventLoggerService` to`EventBus` to finish listening to events.
3. All events are triggered by `EventBus.post` instead of `EventLoggerService.onEvent`. 
4. If you want to implement AudiLog, you only need to add the corresponding `AuditEvent` and the corresponding subscription logic implementation, the implementation of the subscription logic can be reused `EventLoggerService` pattern, add an AuditLogger can be; or a separate implementation of a similar AuditLoggerService, will be registered to the EventBus.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yaooqinn:
thanks for driving this

PR #1853, opened 2022-01-28
[KYUUBI #1798] Add EventBus module to unify the distribution and subscription of Kyuubi's events

iodone:

Use case is simple:
```
    // scribe sync
    EventBus.register[Test0KyuubiEvent](new Test0Handler with EventHandler)
   
    // scribe async
     EventBus.registerAsync[Test0KyuubiEvent](new Test0Handler with EventHandler)

    // send event
    EventBus.post(KyuubiEvent("test0"))
    EventBus.post(KyuubiEvent("test1"))
```


externals/kyuubi-download/pom.xml
```diff
@@ -45,6 +45,7 @@
                         </goals>
                         <configuration>
                             <skip>${spark.archive.download.skip}</skip>
+                            <skipCache>true</skipCache>
```
yaooqinn, line 48: this seems irrelevant

yaooqinn:
cc @wForget, can you help check this PR?

kyuubi-server/src/main/resources/kyuubi-defaults.conf
```diff
@@ -0,0 +1,26 @@
+#
```
yaooqinn, line 1: this file is unnecessary?

kyuubi-common/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def registerAsync[T <: KyuubiEvent: ClassTag](et: EventHandler[T]): EventBus =
+    defaultEventBus.registerAsync[T](et)
+
+  private case class EventBusLive() extends EventBus {
+    private[this] val eventHandlerRegistry = new Registry
+    private[this] val asyncEventHandlerRegistry = new Registry
+    implicit private[this] val ec: ExecutionContext =
+      ExecutionContext.fromExecutor(
+        ThreadUtils.newDaemonQueuedThreadPool(10, 10, 60000L, "event-bus-exec-pool"))
```
wForget, line 55: The queue size seems to be too small, and how do we handle subsequent events when the queue is full?

wForget:
> cc @wForget, can you help check this PR?

Sorry, I'm confused about this, this is overcorrect for my needs.
For me, I just need to store the event asynchronously and don't need to add some handlers to handle it when producing events. I will read the event in storage(like es), then process and audit it.

In addition, I feel this is similar to before:
1. EventBus is similar to EventLoggingService
2. EventHandler is similar to EventLogger

iodone:
> > cc @wForget, can you help check this PR?
> 
> Sorry, I'm confused about this, this is overcorrect for my needs. For me, I just need to store the event asynchronously and don't need to add some handlers to handle it when producing events. I will read the event in storage(like es), then process and audit it.
> 
> In addition, I feel this is similar to before:
> 
> 1. EventBus is similar to EventLoggingService
> 2. EventHandler is similar to EventLogger

@wForget The motivation for the implementation of eventBus can be found in #1798 

EventBus is just a Handler for triggering and adding events, it does not contain any business logic, EventLoggerService can be seen as a concrete business implementation of one of these Handlers. Because the previous EventLoggerService implementation was slightly complex and coupled, it was adapted using EventBus in order to be compatible with the previous implementation logic, and at the same time, it is very clear and simple to use EventBus.

externals/kyuubi-download/pom.xml
iodone: A small problem found during testing, need to submit a new issue to fix?

kyuubi-common/src/main/scala/org/apache/kyuubi/events/EventBus.scala
iodone: For asynchronous processing is relatively simple to implement, the size of the thread pool is configured by kyuubi Config or changed to a ForkJoin thread pool?

yaooqinn:
Now, we have EventHandler/EventLogger/EventBus/EventBusLive/EventLoggingService, it's confusing. Maybe we can make the new design to keep it simple.

yaooqinn:
The `EventLogger` s, such as `JsonEventLogger` extends `AbstractService` now which makes it hard to developers to implement.

And the EventLoggingService is now sync and we shall make it async simply.

WDYT? @wForget @iodone

wForget:
> The `EventLogger` s, such as `JsonEventLogger` extends `AbstractService` now which makes it hard to developers to implement. we shall make it AbstractService-free.
> 
> And the EventLoggingService is now sync and we shall make it async simply.
> 
> WDYT? @wForget @iodone

+1, I prefer this.

iodone:
> The `EventLogger` s, such as `JsonEventLogger` extends `AbstractService` now which makes it hard to developers to implement. we shall make it AbstractService-free.
> 
> And the EventLoggingService is now sync and we shall make it async simply.
> 
> WDYT? @wForget @iodone

This `AbstractService` is actually the `EventHandler` implemention in `EventBus`

iodone:
image
The above is the overall design of the `EventBus` implementation, there is no relationship with the `EventLoggerService`, `EventLoggerService` is seen as a specific implementation of the `EventHandler`. `EventBus` main functions: 
1. Register a `KyuubiEvent` eventHandler, where the `EventHandler` can be `EventLoggerService`, `AuditService`, or any other logic for handling events
2. Trigger an event at some place

By adding `EventBus` is mainly to decouple `EventLoggerService`, I think `EventLoggerService` is only suitable to handle events, it is just an `EventHandler`, it should not trigger event, this is the responsibility of `EventBus`.

iodone:
The definition of `EventHandler` is a function with few constraints, and the input is `KyuubiEvent`, which is a fully AbstractService-free:
```scala
package object events {
  type EventHandler[T <: KyuubiEvent] = T => Unit
}
```
We can implement `EventHandler`, either following the `AbstractService` pattern, or freely writing any implementation pattern.
In this commit it is easy to convert the previously implemented `EventLoggerService` into an `EventHandler` without any intrusiveness.
```scala
 private val eventLoggingService =
    new EventLoggingService with EventHandler[KyuubiServerEvent] {
      override def apply(event: KyuubiServerEvent): Unit = onEvent(event)
    }
```
@yaooqinn @wForget Does this explanation for `EventBus` meet your expectations?

yaooqinn:
Sorry for the late reply.

the idea is great, but can we make it a bit simple. As `EventLogger` can be a typical `EventHandler`, we can remove one layer(EventLoggerService) directly.

iodone:
> Sorry for the late reply.
> 
> the idea is great, but can we make it a bit simple. As `EventLogger` can be a typical `EventHandler`, we can remove one layer(EventLoggerService) directly.

OK，i will improve it

iodone:
@yaooqinn , Could you help me review the code?

yaooqinn:
@iodone thanks for the update. I will review this ASAP. also cc @wForget @ulysses-you

wForget:
LGTM, in addition, I have some comments:
1. Why don't we register handlers with `EventBus.registerAsync`?
2. `SparkJsonLoggingEventHandler` and `ServerJsonLoggingEventHandler` don't seem to do anything, they seem unnecessary.
3. Do we need to use a separate `ExecutionContext` for EventBus instead of `Implicits.global`?

iodone:
> LGTM, in addition, I have some comments:
> 
> 1. Why don't we register handlers with `EventBus.registerAsync`?
> 2. `SparkJsonLoggingEventHandler` and `ServerJsonLoggingEventHandler` don't seem to do anything, they seem unnecessary.
> 3. Do we need to use a separate `ExecutionContext` for EventBus instead of `Implicits.global`?
 
@wForget, thank you for your advice. 
For points 1, 3:
Currently, since the Async pattern has not been validated in real scenarios: 
1. Just the same implementation of the previous LoggerService is done, not upgraded to asynchronous.
2. Scala global `executioncontext` uses the `ForkJoin` settings related to the system, which can ensure optimal performance. We can tuning it when using async EventHandler in the future like the implement of `AsyncElasticSearchLoggingHandler`

For points 2:
For decoupling and extensibility, while also facilitating independent unit testing

wForget:
> > LGTM, in addition, I have some comments:
> > 
> > 1. Why don't we register handlers with `EventBus.registerAsync`?
> > 2. `SparkJsonLoggingEventHandler` and `ServerJsonLoggingEventHandler` don't seem to do anything, they seem unnecessary.
> > 3. Do we need to use a separate `ExecutionContext` for EventBus instead of `Implicits.global`?
> 
> @wForget, thank you for your advice. For points 1, 3: Currently, since the Async pattern has not been validated in real scenarios:
> 
> 1. Just the same implementation of the previous LoggerService is done, not upgraded to asynchronous.
> 2. Scala global `executioncontext` uses the `ForkJoin` settings related to the system, which can ensure optimal performance. We can tuning it when using async EventHandler in the future like the implement of `AsyncElasticSearchLoggingHandler`
> 
> For points 2: For decoupling and extensibility, while also facilitating independent unit testing

yea, these are some small doubts. The overall implementation looks good. @yaooqinn @ulysses-you  WDYT?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -170,15 +169,37 @@ object SparkSQLEngine extends Logging {
...
+
+    def initLoggerEventHandler(conf: KyuubiConf): Unit = {
+      conf.get(ENGINE_EVENT_LOGGERS)
+        .map(EventLoggerType.withName)
+        .foreach {
+          case EventLoggerType.SPARK =>
+            EventBus.register[KyuubiEvent](new SparkHistoryLoggingEventHandler(spark.sparkContext))
+          case EventLoggerType.JSON =>
+            val handler = SparkJsonLoggingEventHandler(
+              spark.sparkContext.applicationAttemptId.getOrElse(spark.sparkContext.applicationId),
```
ulysses-you, line 189: `applicationAttemptId` is a number of attempt count, e.g. `1`

The right name maybe:
```
spark.sparkContext.applicationAttemptId
  .map(id => s"${spark.sparkContext.applicationId}_$id")
  .getOrElse(spark.sparkContext.applicationId)
```

ulysses-you: I see this code is coied, just fixes it together

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -170,15 +169,37 @@ object SparkSQLEngine extends Logging {
...
+              spark.sparkContext.hadoopConfiguration,
+              conf)
+
+            // register JsonLogger as a event handler for default event bus
+            EventBus.register[KyuubiEvent](handler)
+          case logger =>
+            // TODO: Add more implementations
+            throw new IllegalArgumentException(s"Unrecognized event logger: $logger")
+        }
+
```
ulysses-you, line 200: unnecessary line

kyuubi-common/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -0,0 +1,106 @@
...
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.reflect.{classTag, ClassTag}
+import scala.util.Try
+
+import org.apache.kyuubi.Logging
+import org.apache.kyuubi.events.handler.EventHandler
+
+sealed trait EventBus {
```
ulysses-you, line 29: can we enrich the class description ?

kyuubi-common/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -0,0 +1,106 @@
...
+  def post[T <: KyuubiEvent](event: T): Unit
+
+  def register[T <: KyuubiEvent: ClassTag](et: EventHandler[T]): EventBus
```
ulysses-you, line 32: why this and some other place name it `et`...  how about `eventHander` ?

kyuubi-common/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  def register[T <: KyuubiEvent: ClassTag](et: EventHandler[T]): EventBus =
+    defaultEventBus.register[T](et)
+
+  def registerAsync[T <: KyuubiEvent: ClassTag](et: EventHandler[T]): EventBus =
+    defaultEventBus.registerAsync[T](et)
+
+  private case class EventBusLive() extends EventBus {
+    private[this] val eventHandlerRegistry = new Registry
+    private[this] val asyncEventHandlerRegistry = new Registry
```
ulysses-you, line 53: add `lazy`

ulysses-you (commented):
sorry for the late, overall looks good. thank you @iodone

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
iodone: Thanks, i will resolve it ASAP

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #1854, opened 2022-01-29
[KYUUBI #1769][FOLLOWUP] Revert the dependency change for kyuubi-hive-jdbc(shaded) modules

turboFei:

Revert the dependency changes for kyuubi-hive-jdbc and kyuubi-hive-jdbc-shaded modules from #1842

### DIFF
--- a/kyuubi-hive-beeline/pom.xml
+++ b/kyuubi-hive-beeline/pom.xml
@@ -102,8 +102,23 @@
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-log4j12</artifactId>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j-impl</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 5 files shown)


pan3793:
I don't think we need touch `jcl-over-slf4j` and `slf4j-api` in both `kyuubi-hive-jdbc` and `kyuubi-hive-jdbc-shaded` modules, why delete them in the previous PR?

turboFei:
> I think we should not touch `jcl-over-slf4j` and `slf4j-api` in both `kyuubi-hive-jdbc` and `kyuubi-hive-jdbc-shaded` modules, why delete them in the previous PR?

I thought that they all involve common dependencies, it is a mistake.

kyuubi-hive-beeline/pom.xml
```diff
@@ -102,23 +102,8 @@
...
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-core</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-1.2-api</artifactId>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
```
pan3793, line 106: we should not revert this part, it's fine use log4j2 in beeline moudle

pan3793 (approved):

PR #1859, opened 2022-01-30
[KYUUBI #1838] Clean up query results after query operations finish

link3280:
Query results would be cached in memory, and we should clean it up when all rows are fetched.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -222,8 +222,8 @@ class ExecuteStatement(
     try {
       executor.cancelQuery(sessionId, resultId)
     } catch {
-      case _: Throwable =>
-      // ignore further exceptions for cleanup
+      case t: Throwable =>
+        warn("Failed to clean result set " + resultId + " in session " + sessionId, t)
     }
   }
 
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -115,33 +115,42 @@ class ExecuteStatement(
   }
 
   private def runQueryOperation(operation: QueryOperation): Unit = {
-    val resultDescriptor = executor.executeQuery(sessionId, operation)
+    var resultId: String = null
+    try {
+      val resultDescriptor = executor.executeQuery(sessionId, operation)
 
-    val resultID = resultDescriptor.getResultId
+      resultId = resultDescriptor.getResultId
 
-    val rows = new ArrayBuffer[Row]()
-    var loop = true
-    while (loop) {
-      Thread.sleep(50) // slow the processing down
+      val rows = new ArrayBuffer[Row]()
+      var loop = true
 
-      val result = executor.snapshotResult(sessionId, resultID, 2)
-      result.getType match {
-        case TypedResult.ResultType.PAYLOAD =>
-          rows.clear()
-          (1 to result.getPayload).foreach { page =>
-            rows ++= executor.retrieveResultPage(resultID, page).asScala
-          }
-        case TypedResult.ResultType.EOS => loop = false
-        case TypedResult.ResultType.EMPTY =>
+      while (loop) {
+        Thread.sleep(50) // slow the processing down
+
+        val result = executor.snapshotResult(sessionId, resultId, 2)
+        result.getType match {
+          case TypedResult.ResultType.PAYLOAD =>
+            rows.clear()
+            (1 to result.getPayload).foreach { page =>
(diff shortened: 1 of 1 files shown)


link3280:
Now Flink backend does not have very good testabilities, so I manually verify the modification via operation logs. 

截屏2022-01-30 15 07 40

I'd like to add a test suite for Flink engine backend in another issue and add some tests about result cleanups after that.

yaooqinn (approved):

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -115,33 +115,42 @@ class ExecuteStatement(
...
+    try {
+      val resultDescriptor = executor.executeQuery(sessionId, operation)
+
+      resultId = resultDescriptor.getResultId
+
+      val rows = new ArrayBuffer[Row]()
+      var loop = true
+
+      while (loop) {
+        Thread.sleep(50) // slow the processing down
```
SteNicholas, line 128: Could the sleep time be configurable?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -209,6 +218,15 @@ class ExecuteStatement(
     setState(OperationState.FINISHED)
   }
 
+  private def cleanupQueryResult(resultId: String): Unit = {
+    try {
+      executor.cancelQuery(sessionId, resultId)
+    } catch {
+      case _: Throwable =>
+      // ignore further exceptions for cleanup
```
SteNicholas, line 226: Any exception could be logged?

SteNicholas (commented):
@link3280, thanks for your contribution. I left my minor comments for thess changes. Please take a look.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: I think so, but it's out of the scope of the PR (maybe another issue to discuss?). Actually, this snippet is the same as the original one. I'm wrapping it with try-finally.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: I thought of logging the exception, but Flink SQL Client ignores it for some reason (maybe to reduce the noises), so I keep aligned with Flink SQL Client. Either way is okay to me.

SteNicholas: @link3280, I don't think that this should keep aligned with Flink SQL Client. We could log here for the cancel exception so that the users could know canceling query cause the exception which may not cancel successfully.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
SteNicholas: @link3280, I think you could push another commit for the configurable sleep time here because the changes in this pull request cover this logic.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
yanghua: +1 for sharing more information makes sense

link3280: @SteNicholas @yanghua Sorry for the late reply. Fixed.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -209,6 +218,15 @@ class ExecuteStatement(
     setState(OperationState.FINISHED)
   }
 
+  private def cleanupQueryResult(resultId: String): Unit = {
+    try {
+      executor.cancelQuery(sessionId, resultId)
+    } catch {
+      case t: Throwable =>
+        warn("Failed to clean result set " + resultId + " in session " + sessionId, t)
```
pan3793, line 226: `s"xxx $var"`

link3280: @pan3793 thx. Updated as suggested.

pan3793 (approved):

yanghua (approved):
@link3280 LGTM, thanks.

Issue #1868, opened 2022-02-08
[Improvement] Upgrate netty from 4.1.68 to 4.1.73


ulysses-you:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Upgrate netty from 4.1.68 to 4.1.73

### How should we improve?

include log4j from 2.17.0 to 2.17.1

more details see netty [release note](https://netty.io/news/2022/01/12/4-1-73-Final.html) and [change list](https://github.com/netty/netty/issues?page=1&q=is%3Aclosed+milestone%3A4.1.73.Final)

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1869, opened 2022-02-08
[KYUUBI #1868] Upgrate netty from 4.1.68 to 4.1.73

ulysses-you:

close https://github.com/apache/incubator-kyuubi/issues/1868

Pass CI


ulysses-you:
the dependency list changed since https://github.com/netty/netty/pull/11732

ulysses-you:
cc @pan3793 @yaooqinn

externals/kyuubi-flink-sql-engine/pom.xml
```diff
@@ -165,7 +165,7 @@
               <include>com.google.guava:failureaccess</include>
               <include>com.google.guava:guava</include>
               <include>commons-codec:commons-codec</include>
-              <include>io.netty:netty-all</include>
+              <include>io.netty:*</include>
```
yaooqinn, line 168: why do we have netty here？

ulysses-you: the artifact id of netty changed since netty does not provide a fat package

yaooqinn: where do we use netty，especially between server and engine

ulysses-you: The obvious place is `KyuubiConf`, it uses `NettyUtils.MAX_NETTY_THREADS`.

The netty dependency is introduced for mysql frontend.  So I think we can remove the netty dependency at engine side ? @pan3793

yaooqinn: we can remove it from common

yaooqinn: in this pr，let's do it your way

pan3793: Move `NettyUtils` to `kyuubi-server` module makes sense to me.

yaooqinn (approved):

ulysses-you:
thanks, merging to master

Issue #1536, opened 2021-12-09
[FEATURE] Add session conf provider


ulysses-you:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add session conf provider to decouple the engine builder and session conf.

Introduce a new method in `SessionConfAdvisor`

```java
/** The returned conf will overwrite the session conf */
default Map<String, String> getConfOverlay(String user, Map<String, String> sessionConf) {
  return new HashMap<>();
}
```

So we can build the different conf by session.

### Motivation

We can get the session config from remote service.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yaroslavambrozyak:
Hi guys, this is so useful feature and I'd like to use it on our Kyuubi server. I am just curious do you know approximately when the PR will be merged? Thanks

yaooqinn:
sure. And you can help review the PR. I think it can be ready in v1.5.0

PR #1892, opened 2022-02-11
[KYUUBI #1536] Add session conf advisor

ulysses-you:

closes https://github.com/apache/incubator-kyuubi/issues/1536


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1172,6 +1172,13 @@ object KyuubiConf {
       .checkValues(OperationLanguages.values.map(_.toString))
       .createWithDefault(OperationLanguages.SQL.toString)
 
+  val SESSION_CONF_ADVISOR: ConfigEntry[String] =
+    buildConf("session.conf.advisor")
+      .doc("Choose a custom config advisor for session, the default implementation return empty.")
```
SteNicholas, line 1177: Is it better that the default implementation return null?

SteNicholas (commented):
@ulysses-you, thanks for your contribution. I have left minor comments for this pull request.

kyuubi-server/pom.xml
```diff
@@ -31,6 +31,12 @@
     <packaging>jar</packaging>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-plugin</artifactId>
```
yaooqinn, line 36: kyuubi-server-plugin

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -45,11 +50,33 @@ class KyuubiSessionImpl(
...
+      case ("kyuubi.engine.pool.size.threshold", _) =>
+      case (key, value) =>
+        sessionConf.set(key, value)
+        userSpecifiedConf.put(key, value)
+    }
+
+    val sessionConfMap = ImmutableMap.copyOf[String, String](sessionConf.getAll.asJava)
+    sessionManager.sessionConfAdvisor.getConfSuggestion(
+      user,
+      sessionConfMap).asScala.foreach {
```
yaooqinn, line 67: NPE?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -27,13 +27,15 @@ import org.apache.kyuubi.credentials.HadoopCredentialsManager
...
 import org.apache.kyuubi.operation.KyuubiOperationManager
+import org.apache.kyuubi.plugin.{PluginLoader, SessionConfAdvisor}
 
 class KyuubiSessionManager private (name: String) extends SessionManager(name) {
 
   def this() = this(classOf[KyuubiSessionManager].getSimpleName)
 
   val operationManager = new KyuubiOperationManager()
   val credentialsManager = new HadoopCredentialsManager()
+  lazy val sessionConfAdvisor: SessionConfAdvisor = PluginLoader.loadSessionConfAdvisor(conf)
```
yaooqinn, line 38: don't be lazy?

kyuubi-plugin/src/main/java/org/apache/kyuubi/plugin/SessionConfAdvisor.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Provide the session configuration according to the user and session configuration. */
+public interface SessionConfAdvisor {
+  /** The returned conf is used if it is not existed in session conf */
+  default Map<String, String> getConfSuggestion(String user, Map<String, String> sessionConf) {
+    return new HashMap<>();
+  }
```
yaooqinn, line 28: on second thought, we'd better remove this API to keep it unambiguous

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
ulysses-you: changed to optional

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -45,11 +50,30 @@ class KyuubiSessionImpl(
...
+      sessionConfMap)
+    if (confOverlay != null) {
+      confOverlay.asScala.foreach {
+        case (key, value) =>
+          sessionConf.set(key, value)
+          userSpecifiedConf.put(key, value)
+      }
+    }
+
+    userSpecifiedConf.asScala.toMap
```
yaooqinn, line 76: simply add ?

```scala
private[kyuubi] val optimizedConf: Map[String, String] = {
    val confOverlay = sessionManager.sessionConfAdvisor.getConfSuggestion(
      user,
      normalizedConf.asJava)
    normalizedConf ++ confOverlay.asScala
  }
```

.github/workflows/style.yml
```diff
@@ -41,6 +41,12 @@ jobs:
         with:
           distribution: zulu
           java-version: 8
+      - name: install kyuubi with the latest commit
```
yaooqinn, line 44: RELATED?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1172,6 +1172,13 @@ object KyuubiConf {
       .checkValues(OperationLanguages.values.map(_.toString))
       .createWithDefault(OperationLanguages.SQL.toString)
 
+  val SESSION_CONF_ADVISOR: OptionalConfigEntry[String] =
+    buildConf("session.conf.advisor")
+      .doc("Choose a custom config advisor for session, the default implementation return empty.")
```
yaooqinn, line 1177: let's detail the doc of the config, e.g. what's happening while using this plugin, not just duplicates the config name with some words reordered, besides `the default implementation return empty` is redundant too as we will have a `<td>` for default value.

yaooqinn: Additionally, we need a separate doc for the plugin. there is nowhere to find the relation between SessionConfAdvisor and this config

.github/workflows/style.yml
ulysses-you: the new added failed will fail in style check workflow if we do not install first. So I changed the ga in this PR

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
yaooqinn: i see,

.github/workflows/style.yml
yaooqinn, line 44:
```suggestion
      - name: Install
```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1172,6 +1172,16 @@ object KyuubiConf {
       .checkValues(OperationLanguages.values.map(_.toString))
       .createWithDefault(OperationLanguages.SQL.toString)
 
+  val SESSION_CONF_ADVISOR: OptionalConfigEntry[String] =
+    buildConf("session.conf.advisor")
+      .doc("A config advisor plugin for Kyuubi Server. This plugin can provide some custom " +
+        "configs for different user or session configs and overwrite the session config before " +
+        "open a new session. This config value should be a class which is a child of " +
+        "org.apache.kyuubi.plugin.SessionConfAdvisor.")
```
yaooqinn, line 1180: zero-arg constructor?

kyuubi-server-plugin/src/main/java/org/apache/kyuubi/plugin/SessionConfAdvisor.java
```diff
@@ -0,0 +1,29 @@
...
+package org.apache.kyuubi.plugin;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Provide the session configuration according to the user and session configuration. */
+public interface SessionConfAdvisor {
+  /** The returned conf will overwrite the session conf */
+  default Map<String, String> getConfOverlay(String user, Map<String, String> sessionConf) {
+    return new HashMap<>();
```
yaooqinn, line 27: Nit:Collections.emptyMap();?

kyuubi-server/pom.xml
```diff
@@ -31,6 +31,12 @@
     <packaging>jar</packaging>
 
     <dependencies>
+        <dependency>
```
yaooqinn, line 34: to L40?

kyuubi-server/src/main/scala/org/apache/kyuubi/plugin/PluginLoader.scala
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.kyuubi.plugin
+
+import scala.util.control.NonFatal
+
+import org.apache.kyuubi.KyuubiException
+import org.apache.kyuubi.config.KyuubiConf
+
+object PluginLoader {
```
yaooqinn, line 25: nit:private[kyuubi]

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
yaooqinn: can we init immediately when `initialize`

.github/workflows/style.yml
```diff
@@ -41,6 +41,12 @@ jobs:
         with:
           distribution: zulu
           java-version: 8
+      - name: install
```
yaooqinn, line 44: nit: Install？

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -45,8 +47,20 @@ class KyuubiSessionImpl(
...
+      normalizedConf ++ confOverlay.asScala
+    } else {
+      warn(s"the server plugin return null value for user: $user, ignore it")
+      normalizedConf
+    }
+  }
+
   // TODO: needs improve the hardcode
-  normalizedConf.foreach {
+  optimizedConf.foreach {
```
yaooqinn, line 63: we shall also update it at L107

ulysses-you: ah, good catch !

kyuubi-server/pom.xml
```diff
@@ -37,6 +37,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
```
yaooqinn, line 40: cp this to codecov

yaooqinn (approved):

ulysses-you:
thanks, merging to master

PR #1907, opened 2022-02-14
[KYUUBI #1905] log4j2-defaults.properties is not working since log4j 2 is always intialized by default

cfmcgrady:

As per [SPARK-37746](https://github.com/apache/spark/pull/35026), Kyuubi also has the same problem with initializing Log4j2.

[log4j2](https://github.com/apache/logging-log4j2/blob/rel/2.17.1/log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java#L717) provides a default configuration so there is always an appender (ConsoleAppender) with ERROR level

This pr changes the behavior of checking the initialization of log4j2.

```
LogManager.getRootLogger
  .asInstanceOf[org.apache.logging.log4j.core.Logger]
  .getAppenders
  .asScala
  .foreach {
    case (name, appender) =>
      println(s"default appender: $name, class: ${appender.getClass.getName}")
  }
```

```
default appender: DefaultConsole-2, class: org.apache.logging.log4j.core.appender.ConsoleAppender
```

After this pr, the root level is `INFO`.

截屏2022-02-14 下午5 32 26

### DIFF
--- a/kyuubi-common/src/main/resources/log4j2-defaults.properties
+++ b/kyuubi-common/src/main/resources/log4j2-defaults.properties
@@ -19,20 +19,8 @@
 rootLogger.level = info
 rootLogger.appenderRef.stdout.ref = STDOUT
 
-# Console Appender
 appender.console.type = Console
 appender.console.name = STDOUT
 appender.console.target = SYSTEM_OUT
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d{HH:mm:ss.SSS} %p %c: %m%n
-
-appender.console.filter.1.type = Filters
-
-appender.console.filter.1.a.type = ThresholdFilter
-appender.console.filter.1.a.level = fatal
-
-# SPARK-34128: Suppress undesirable TTransportException warnings, due to THRIFT-4805
-appender.console.filter.1.b.type = RegexFilter
-appender.console.filter.1.b.regex = .*Thrift error occurred during processing of message.*
-appender.console.filter.1.b.onMatch = deny
-appender.console.filter.1.b.onMismatch = neutral
+appender.console.layout.pattern = %d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n
--- a/kyuubi-common/src/main/resources/log4j2-defaults.properties
+++ b/kyuubi-common/src/main/resources/log4j2-defaults.properties
@@ -19,8 +19,20 @@
 rootLogger.level = info
 rootLogger.appenderRef.stdout.ref = STDOUT
 
+# Console Appender
 appender.console.type = Console
 appender.console.name = STDOUT
 appender.console.target = SYSTEM_OUT
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n
+appender.console.layout.pattern = %d{HH:mm:ss.SSS} %p %c: %m%n
+
+appender.console.filter.1.type = Filters
+
+appender.console.filter.1.a.type = ThresholdFilter
+appender.console.filter.1.a.level = INFO
+
+# SPARK-34128: Suppress undesirable TTransportException warnings, due to THRIFT-4805
+appender.console.filter.1.b.type = RegexFilter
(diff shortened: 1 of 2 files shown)


cfmcgrady:
cc @pan3793 @yaooqinn @turboFei @link3280

kyuubi-common/src/main/scala/org/apache/kyuubi/Logging.scala
```diff
@@ -179,4 +179,20 @@ object Logging {
     val binderClass = StaticLoggerBinder.getSingleton.getLoggerFactoryClassStr
     "org.apache.logging.slf4j.Log4jLoggerFactory".equals(binderClass)
   }
+
+  /**
+   * Return true if log4j2 is initialized by default configuration which has one
+   * appender with error level. See `org.apache.logging.log4j.core.config.DefaultConfiguration`.
+   */
+  private[kyuubi] def islog4j2DefaultConfigured(): Boolean = {
```
yaooqinn, line 187: private?

kyuubi-common/src/main/scala/org/apache/kyuubi/Logging.scala
```diff
@@ -92,16 +93,15 @@ trait Logging {
...
       // scalastyle:off println
-      if (!log4j2Initialized) {
+      if (Logging.islog4j2DefaultConfigured()) {
         Logging.useDefault = true
         val defaultLogProps = "log4j2-defaults.properties"
         Option(Thread.currentThread().getContextClassLoader.getResource(defaultLogProps)) match {
           case Some(url) =>
             val context = LogManager.getContext(false).asInstanceOf[LoggerContext]
             context.setConfigLocation(url.toURI)
+            System.err.println(s"Using Kyuubi's default log4j profile: $url")
```
yaooqinn, line 104: looks unnecessary

yaooqinn:
I have tested it locally, LGTM overall.

yaooqinn (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/Logging.scala
cfmcgrady: Only for debug, it's ok for me to revert.

yaooqinn (approved):

kyuubi-common/src/main/resources/log4j2-defaults.properties
```diff
@@ -29,7 +29,7 @@ appender.console.layout.pattern = %d{HH:mm:ss.SSS} %p %c: %m%n
 appender.console.filter.1.type = Filters
 
 appender.console.filter.1.a.type = ThresholdFilter
-appender.console.filter.1.a.level = fatal
+appender.console.filter.1.a.level = INFO
```
link3280, line 32: Minor: should we keep in lower case to align with others?

link3280 (commented):
@cfmcgrady Thanks for the PR. LGTM in general, with a minor concern of mine.

kyuubi-common/src/main/resources/log4j2-defaults.properties
jiwq: Agree to @link3280 . If need change, should better modify global.

kyuubi-common/src/main/scala/org/apache/kyuubi/Logging.scala
```diff
@@ -179,4 +178,20 @@ object Logging {
     val binderClass = StaticLoggerBinder.getSingleton.getLoggerFactoryClassStr
     "org.apache.logging.slf4j.Log4jLoggerFactory".equals(binderClass)
   }
+
+  /**
+   * Return true if log4j2 is initialized by default configuration which has one
+   * appender with error level. See `org.apache.logging.log4j.core.config.DefaultConfiguration`.
+   */
+  private def islog4j2DefaultConfigured(): Boolean = {
```
jiwq, line 186:
```suggestion
  private def isLog4j2DefaultConfigured(): Boolean = {
```

link3280 (approved):

yaooqinn:
thanks @cfmcgrady for the fix and @jiwq @link3280 for the review, merged to master

Issue #1876, opened 2022-02-08
[Improvement] Set Spark 3.2.1 as default


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

dep

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

SteNicholas:
@yaooqinn, could you please assign this ticket to me?

PR #1913, opened 2022-02-15
[KYUUBI #1876] Set Spark 3.2.1 as default

SteNicholas:

Set Spark 3.2.1 as default.


docs/integrations/delta_lake_with_azure_blob.md
```diff
@@ -106,7 +106,7 @@ Welcome to
       ____              __
      / __/__  ___ _____/ /__
     _\ \/ _ \/ _ `/ __/  '_/
-   /___/ .__/\_,_/_/ /_/\_\   version 3.1.2
+   /___/ .__/\_,_/_/ /_/\_\   version 3.2.1
```
pan3793, line 109: please revert this one, it may makes the context conflict

pom.xml
```diff
@@ -132,7 +132,7 @@
         <scopt.version>4.0.1</scopt.version>
         <slf4j.version>1.7.35</slf4j.version>
         <log4j.version>2.17.1</log4j.version>
-        <spark.version>3.1.2</spark.version>
+        <spark.version>3.2.1</spark.version>
```
pan3793, line 135: we should also change `delta.version`, because the minor version of delta and spark is one-one mapping

pan3793: we can left some comments here, e.g. 'if change minor version of spark, don't forget to change the deafult `delta.version`'

SteNicholas:
@pan3793, thanks for your review. I have addressed your comments and please help to review again.

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #1937, opened 2022-02-19
Build: Use Spark 3.2 for cross version verification

pan3793:

Follow up #1876

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -23,21 +23,21 @@ jobs:
             spark-hadoop: '2.7'
           - java: 8
             spark: '3.1'
+            spark-hadoop: '3.2'
+          - java: 8
+            spark: '3.2'
             spark-hadoop: '2.7'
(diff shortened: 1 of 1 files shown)


SteNicholas (approved):
LGTM. cc @yaooqinn

pan3793:
Thanks, merging to master

.github/workflows/master.yml
```diff
@@ -23,21 +23,21 @@ jobs:
             spark-hadoop: '2.7'
           - java: 8
             spark: '3.1'
+            spark-hadoop: '3.2'
+          - java: 8
+            spark: '3.2'
```
yaooqinn, line 28: `spark-hadoop: '2.7'` means breaking hadoop3.2?

pan3793: I don't get your ponit.

Previously, we use `kyuubi-spark-sql-engine.jar` build against Spark 3.1 do corss test on spark-3.0 & hadoop-2.7 and spark 3.2 & hadoop 3.2.

Now, we use `kyuubi-spark-sql-engine.jar` build against Spark 3.2 do corss test on spark-3.0 & hadoop-2.7 and spark 3.1 & hadoop 3.2.

yaooqinn: I mean cross check shall use default profiles

pan3793: The default hadoop version of Spark 3.1 is 3.2. https://mvnrepository.com/artifact/org.apache.spark/spark-core_2.12/3.1.3

yaooqinn: shall we use 3.2 instead of 2.7

PR #1946, opened 2022-02-20
Close the kyuubi connection on launch engine operation failure

turboFei:

For kyuubi connection, if the launch engine operation fails, we need close the kyuubi connection.

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -78,6 +78,7 @@ import org.slf4j.LoggerFactory;
 public class KyuubiConnection implements java.sql.Connection, KyuubiLoggable {
   public static final Logger LOG = LoggerFactory.getLogger(KyuubiConnection.class.getName());
   public static final String BEELINE_MODE_PROPERTY = "BEELINE_MODE";
+  public static final int DEFAULT_ENGINE_LOG_THREAD_TIMEOUT = 10 * 1000;
 
   private String jdbcUriString;
   private String host;
@@ -100,6 +101,7 @@ public class KyuubiConnection implements java.sql.Connection, KyuubiLoggable {
   private boolean initFileCompleted = false;
 
   private TOperationHandle launchEngineOpHandle = null;
+  private Thread engineLogThread;
   private boolean engineLogInflight = true;
   private volatile boolean launchEngineOpCompleted = false;
 
@@ -263,7 +265,7 @@ public class KyuubiConnection implements java.sql.Connection, KyuubiLoggable {
   private void showLaunchEngineLog() {
     if (launchEngineOpHandle != null) {
       LOG.info("Starting to get launch engine log.");
-      Thread logThread =
+      engineLogThread =
           new Thread("engine-launch-log") {
 
             @Override
@@ -282,7 +284,7 @@ public class KyuubiConnection implements java.sql.Connection, KyuubiLoggable {
               LOG.info("Finished to get launch engine log.");
             }
           };
-      logThread.start();
+      engineLogThread.start();
     }
   }
 
@@ -935,6 +937,17 @@ public class KyuubiConnection implements java.sql.Connection, KyuubiLoggable {
         }
       }
     }
+  }
+
+  public void closeOnLaunchEngineFailed(Thread logThread) throws SQLException {
+    if (logThread != null && logThread.isAlive()) {
+      logThread.interrupt();
+      try {
+        logThread.join(DEFAULT_ENGINE_LOG_THREAD_TIMEOUT);
+      } catch (Exception e) {
+      }
+    }
+    close();
   }
 
   /*
@@ -1669,12 +1682,16 @@ public class KyuubiConnection implements java.sql.Connection, KyuubiLoggable {
               break;
           }
         }
-      } catch (SQLException e) {
-        engineLogInflight = false;
-        throw e;
       } catch (Exception e) {
         engineLogInflight = false;
-        throw new SQLException(e.toString(), "08S01", e);
+        if (!isBeeLineMode) {
+          closeOnLaunchEngineFailed(engineLogThread);
+        }
+        if (e instanceof SQLException) {
+          throw e;
+        } else {
+          throw new SQLException(e.toString(), "08S01", e);
+        }
       }
     }
   }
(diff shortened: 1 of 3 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -1669,12 +1682,16 @@ public void waitLaunchEngineToComplete() throws SQLException {
               break;
           }
         }
-      } catch (SQLException e) {
-        engineLogInflight = false;
-        throw e;
       } catch (Exception e) {
         engineLogInflight = false;
-        throw new SQLException(e.toString(), "08S01", e);
+        if (!isBeeLineMode) {
```
pan3793, line 1687: why skip in beeline mode?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -1669,12 +1682,16 @@ public void waitLaunchEngineToComplete() throws SQLException {
...
+          closeOnLaunchEngineFailed(engineLogThread);
+        }
+        if (e instanceof SQLException) {
+          throw e;
+        } else {
+          throw new SQLException(e.toString(), "08S01", e);
```
pan3793, line 1693: `e.toString()` => `e.getMessage()`

pan3793 (approved):

SteNicholas:
@turboFei, could you please add the test case for this kyubbi connection close?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -78,6 +78,7 @@
 public class KyuubiConnection implements java.sql.Connection, KyuubiLoggable {
   public static final Logger LOG = LoggerFactory.getLogger(KyuubiConnection.class.getName());
   public static final String BEELINE_MODE_PROPERTY = "BEELINE_MODE";
+  public static final int DEFAULT_ENGINE_LOG_THREAD_TIMEOUT = 10 * 1000;
```
SteNicholas, line 81: Could this add an option for log thread timeout?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -937,6 +943,17 @@ public void close() throws SQLException {
     }
   }
 
+  public void closeOnLaunchEngineFailure(Thread logThread) throws SQLException {
+    if (logThread != null && logThread.isAlive()) {
+      logThread.interrupt();
+      try {
+        logThread.join(DEFAULT_ENGINE_LOG_THREAD_TIMEOUT);
+      } catch (Exception e) {
+      }
```
SteNicholas, line 952: Does the `logThread` need to set to null?

SteNicholas (changes requested):
@turboFei, thanks for your contribution. I have left the comments for the changes. What'more, could you please add the test case for this kyubbi connection close?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -937,6 +943,17 @@ public void close() throws SQLException {
     }
   }
 
+  public void closeOnLaunchEngineFailure(Thread logThread) throws SQLException {
```
ulysses-you, line 946: since `engineLogThread` is a global variable, why we need pass it here.

turboFei: thanks

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
turboFei: addressed, thanks

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
turboFei: it is not needed, just a thread timeout.

turboFei: Just refer: https://github.com/apache/hive/blob/2d05298f5b4a40b5eb339efcd0701e31b694ccc7/beeline/src/java/org/apache/hive/beeline/Commands.java#L75

turboFei: the thread.join should return very fast, the timeout is for the worst case.

SteNicholas: @turboFei, IMO, the timeout value is configurable, whatever any reason, which could be 5*1000 etc.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -937,6 +943,18 @@ public void close() throws SQLException {
...
   }
 
+  private void closeOnLaunchEngineFailure() throws SQLException {
+    if (engineLogThread != null && engineLogThread.isAlive()) {
+      engineLogThread.interrupt();
+      try {
+        engineLogThread.join(DEFAULT_ENGINE_LOG_THREAD_TIMEOUT);
+      } catch (Exception e) {
+      }
+      engineLogThread = null;
```
SteNicholas, line 953: Does this need to consider the condition that the `engineLogThread` is not null and not alive?

SteNicholas (commented):
@turboFei , thanks for your update. I left the minor comments for the `engineLogThread` close.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
pan3793: It should not a big deal, we can remove the `final`, if users really want to change it, they can re-assign the value of the static variable.

SteNicholas: @pan3793, I agree with you for the point that removes the `final`.

turboFei: thanks, resolved.

SteNicholas (approved):
LGTM. @yaooqinn @pan3793. Could you please help to review and merge?

pan3793:
Thanks, merging to master/1.4

PR #1947, opened 2022-02-21
Revamp Github Action settings for CI

yaooqinn:

1. Add missing asf header
2. Disable fail-fast to let jobs run individually
3. Group defaults maven builds with java & spark version matrix
    - so this PR brings fully tests with JDK 11 back
4. Group others(now cross-version check only) in `include`
5. replace `actions/cache@v2` with `actions/setup-java@v2` - `cache: 'maven'`

Need to pass all CI jobs


.github/workflows/master.yml
```diff
@@ -53,36 +63,21 @@ jobs:
...
-          restore-keys: |
-            ${{ runner.os }}-maven-net-
-      - uses: actions/cache@v2
-        with:
-          path: ~/.m2/repository/io
-          key: ${{ runner.os }}-maven-io-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-io-
-      - name: Build with Maven
+          cache: 'maven'
```
ulysses-you, line 66: this is a good code improvement ! more details [action docs](https://github.com/actions/setup-java#caching-packages-dependencies)

ulysses-you (approved):

.github/workflows/master.yml
pan3793: Awosome, do we need to un-pin `actions/checkout@v2.3.5`?

pan3793 (approved):

.github/workflows/master.yml
```diff
@@ -13,76 +29,55 @@ on:
...
     strategy:
+      fail-fast: false
       matrix:
+        java:
+          - 8
+          - 11
+        spark:
+          - '3.0'
+          - '3.1'
+          - '3.2'
```
pan3793, line 43: minor style point: since it's too short, `['3.0', '3.1', '3.2']` looks better

.github/workflows/docker-image.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       cancel-in-progress: true
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.3.5
```
pan3793, line 19: Please also add license to this file

.github/workflows/master.yml
```diff
@@ -1,4 +1,20 @@
-name: Kyuubi
+name: CI
+#
```
pan3793, line 2: The workflows yamls has different style of empty line numbers before/after license header, could you please make them consistent?

.github/workflows/master.yml
yaooqinn: the current style looks better to me, many other `short` places use this style too, and it's easy to maintainance for changing

.github/workflows/docker-image.yml
```diff
@@ -1,4 +1,20 @@
 name: Publish Docker image
+#
```
cfmcgrady, line 2: Shall we put the license in the file header?

yaooqinn: do you mean to move to L1? that's invalid for GA

cfmcgrady: It works fine for spark, Is anything wrong?

https://github.com/apache/spark/blob/1112240ffc5456778124e43ec75aae5d8cb7121f/.github/workflows/build_and_test.yml#L1-L18

yaooqinn: Ok, thanks for the inputs. I might make some mistake before

yaooqinn:
thanks @pan3793 @ulysses-you @cfmcgrady for the help, merged to master

PR #1950, opened 2022-02-21
Remove ambiguous SPARK_HADOOP_VERSION

yaooqinn:

The original idea of SPARK_HADOOP_VERSION is used to concat spark release names only, now we need to remove it as
- SPARK_HADOOP_VERSION is misunderstood by developers and misused somewhere like the one of kyuubi compiled
- multi-engine support now
- the release names  of spark(or something else) are very easy to get through code with different environments, prod/test/dev
- A `mvn` job is bundled with `bin/load-kyuubi-env.sh` which is truly worrisome
- SPARK_HADOOP_VERSION on spark side hass broken already for spark 3.2 which actually bundled with hadoop 3.3, see https://github.com/apache/spark-website/pull/361#discussion_r730716668

### DIFF
--- a/bin/load-kyuubi-env.sh
+++ b/bin/load-kyuubi-env.sh
@@ -71,32 +71,11 @@ fi
 export KYUUBI_SCALA_VERSION="${KYUUBI_SCALA_VERSION:-"2.12"}"
 
 if [[ -f ${KYUUBI_HOME}/RELEASE ]]; then
-  KYUUBI_VERSION="$(grep "Kyuubi " "$KYUUBI_HOME/RELEASE" | awk -F ' ' '{print $2}')"
-  FLINK_VERSION_BUILD="$(grep "Flink " "$KYUUBI_HOME/RELEASE" | awk -F ' ' '{print $2}')"
-  SPARK_VERSION_BUILD="$(grep "Spark " "$KYUUBI_HOME/RELEASE" | awk -F ' ' '{print $2}' | grep -v 'Hadoop')"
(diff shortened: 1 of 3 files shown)


bin/load-kyuubi-env.sh
```diff
@@ -109,7 +88,6 @@ if [ $silent -eq 0 ]; then
   echo "JAVA_HOME: ${JAVA_HOME}"
 
   echo "KYUUBI_HOME: ${KYUUBI_HOME}"
-  echo "KYUUBI_VERSION: ${KYUUBI_VERSION}"
```
ulysses-you, line 112: shall we keep the kyuubi version here ?

yaooqinn: it will be print in log ASAP

yaooqinn: it will be printed in stdout asap

pom.xml
```diff
@@ -1901,20 +1900,6 @@
             </properties>
         </profile>
 
-        <profile>
```
ulysses-you, line 1904: I'm not sure if it's good to remove the Spark Hadoop version profile. The Spark master branch has already moved to Hadoop 3.3.1. So if we upgrate Spark to 3.3 we still need the profile.

yaooqinn: when we update spark

before: we update 2 parameters `spark.version` and `hadoop.binary.version`, and we need to ensure in our mind that the `spark.archive.name` is also need to update.

now: we update 1 parameter `spark.version` only or sometimes 2 if `spark.archive.name` changes its static pattern

ulysses-you: I see, then I'm fine with this change. We have already use `spark.archive.name` in ga workflow, so the Spark Hadoop version can be removed.

yaooqinn:
thanks, merged to master

Issue #1948, opened 2022-02-21
[Improvement] Upgrade thrift version to 0.16.0


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Due to spark side will upgrade thrift version to 0.16, see https://github.com/apache/spark/commit/4789e1f234a92b3c17d1f962c8f374ef9478612a which contains breaking API and it will be backported to 3.1 and 3.2.

So we need to sync with spark and shade the thrift libs for engines

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

SteNicholas:
@yaooqinn, could you please assign this ticket to me?

PR #1953, opened 2022-02-22
[KYUUBI #1948] Upgrade thrift version to 0.16.0

SteNicholas:

Upgrade libthrift to 0.16.0 due to [CVE-2020-13949](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-13949) and the coming upstream change of Spark https://github.com/apache/spark/pull/34362

### _What changed in this PR?_

- Upgrade libthrift to 0.16.0

- Shade and relocate `thrift` and `hive-service-rpc` classes in `kyuubi-spark-engine`, it's necessary to avoide conflicting with old thrift libs bundled in Spark binary releases.

- Due to thrift change the method signature, the subclasses those interfaces in Kyuubi also need to modify to pass compile.
We rely on Hive 2.3.9 jars in some components, e.g. `kyuubi-hive-jdbc`, `LocalMetaServer` in `kyuubi-server` test classes.

Some classes in Hive jars compiled against old thrift interfaces which are not compatible with thrift 0.16.0, it causes runtime link error, we found the following classes which breaks the test and copied them with necessary modification to make it work with thrift 0.16.0.

    - `TFramedTransport`
    - `TFilterTransport`
    - `TUGIAssumingTransport`
    - `TUGIContainingTransport`

- Next Steps, I think it's worth to do in separated PRs.

    - Recover the `HiveDelegationTokenProviderSuite`, one approach is use an isolate classloader to load HMS classes and thrift 0.9.3 classes from Maven, this approach can also be used for the planed Zoopkeeper upgrading to help us verficating the compatibility of Zookeeper Server 3.4.x.
    - Rewrite `kyuubi-hive-jdbc` to make it decouple with Hive jars, because there maybe other places which may not work with thrift 0.16.0 but the UTs does not cover.


pan3793:
We need to shaded thrift libs into kyuubi-spark-sql-engine because Spark bundles old thrift libs in binary release.

pom.xml
```diff
@@ -133,6 +133,7 @@
         <scopt.version>4.0.1</scopt.version>
         <slf4j.version>1.7.35</slf4j.version>
         <log4j.version>2.17.1</log4j.version>
+        <libthrift.version>0.16.0</libthrift.version>
```
pan3793, line 136: Let's define `fb303.version` and `thrift.version`

pom.xml
```diff
@@ -519,11 +520,6 @@
                 <version>${kubernetes-client.version}</version>
             </dependency>
 
-            <!--
-              because of THRIFT-4805, we don't upgrade to libthrift:0.12.0,
-              because of THRIFT-5274, we don't upgrade to libthrift:0.13.0,
-              so just keep hive-service-rpc:2.3.9 transitive dependency libthrift:0.9.3
-            -->
```
pan3793, line 526: Then we can exclude *:* from `hive-service-rpc`, and add fb303 and thrift explicitly

pom.xml
```diff
@@ -1307,6 +1303,22 @@
             <artifactId>scalatest_${scala.binary.version}</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
```
pan3793, line 1307: move to L555

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -201,6 +201,13 @@
                                 <include>org.apache.curator.**</include>
                             </includes>
                         </relocation>
+                        <relocation>
+                            <pattern>org.apache.thrift</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.org.apache.thrift</shadedPattern>
+                            <includes>
+                                <include>org.apache.thrift.**</include>
+                            </includes>
+                        </relocation>
```
pan3793, line 210: Should also be declared at L193

dev/dependencyList
```diff
@@ -69,8 +69,6 @@ jetty-servlet/9.4.41.v20210516//jetty-servlet-9.4.41.v20210516.jar
 jetty-util-ajax/9.4.41.v20210516//jetty-util-ajax-9.4.41.v20210516.jar
 jetty-util/9.4.41.v20210516//jetty-util-9.4.41.v20210516.jar
 jline/0.9.94//jline-0.9.94.jar
-libfb303/0.9.3//libfb303-0.9.3.jar
-libthrift/0.9.3//libthrift-0.9.3.jar
```
pan3793, line 73: Why them disappeared?

dev/dependencyList
```diff
@@ -20,6 +20,7 @@ classgraph/4.8.95//classgraph-4.8.95.jar
 commons-codec/1.15//commons-codec-1.15.jar
 commons-lang/2.6//commons-lang-2.6.jar
 commons-lang3/3.10//commons-lang3-3.10.jar
+commons-logging/1.2//commons-logging-1.2.jar
```
pan3793, line 23: `commons-logging` should be excluded.

dev/dependencyList
```diff
@@ -36,6 +37,8 @@ hk2-api/2.6.1//hk2-api-2.6.1.jar
 hk2-locator/2.6.1//hk2-locator-2.6.1.jar
 hk2-utils/2.6.1//hk2-utils-2.6.1.jar
 htrace-core4/4.1.0-incubating//htrace-core4-4.1.0-incubating.jar
+httpclient/4.5.13//httpclient-4.5.13.jar
+httpcore/4.4.15//httpcore-4.4.15.jar
```
pan3793, line 41: Add them LGTM, prepare for Thrift-HTTP mode.

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HiveDelegationTokenProviderSuite.scala
```diff
@@ -93,7 +93,8 @@ class HiveDelegationTokenProviderSuite extends KerberizedTestHelper {
     FileUtils.deleteDirectory(hadoopConfDir)
   }
 
-  test("obtain hive delegation token") {
+  // Ignore the test because LocalMetaServer can not work with Thrift 0.16.0.
+  ignore("obtain hive delegation token") {
```
pan3793, line 97: cc @zhouyifan279 
We can not start `HiveMetastore` using thrift 0.16.0, because the Interface method signature changed, the class is not compatible.

kyuubi-hive-jdbc/src/main/java/org/apache/hadoop/hive/thrift/TFilterTransport.java
```diff
@@ -0,0 +1,111 @@
...
+    wrapped.consumeBuffer(len);
+  }
+
+  @Override
+  public TConfiguration getConfiguration() {
+    return null;
+  }
+
+  @Override
+  public void updateKnownMessageSize(long size) throws TTransportException {}
```
pan3793, line 107: cc @yaooqinn 
We need to copy this and some related classes from Hive Jars and rebuild with thrift 0.16.0. Because `TTransport` changed.

pan3793:
Summary of this PR,
1. Upgrade `libthrift` to 0.16.0 due to CVE issues and the coming upstream change of Spark
2. Shade and relocate `thrift` and `hive-service-rpc` classes in `kyuubi-spark-engine`, it's necessary to avoding conflict with old thrift libs bundled in Spark binary releases.
3. Due to thrift change the method signature, the subclasses of Kyuubi also need to change to make compile passed.
4. We rely on Hive 2.3.9 jars in some components, e.g. `kyuubi-hive-jdbc`, `LocalMetaServer` in `kyuubi-server` test classes, some classes in Hive jars compiled against old thrift classes which are not compatible with thrift 0.16.0, which cause runtime link error, we find all broken tests and copy the broken classes with necessary modification to make it work with thrift 0.16.0, but there may still risk somewhere the UT not covered.

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -191,6 +191,8 @@
                             <include>org.apache.curator:curator-client</include>
                             <include>org.apache.curator:curator-framework</include>
                             <include>org.apache.curator:curator-recipes</include>
+                            <include>org.apache.thrift:*</include>
```
yaooqinn, line 194: shall we not use *？

yaooqinn:
Thanks for your summary, @pan3793
> Some classes in Hive jars compiled against old thrift interfaces which are not compatible with thrift 0.16.0, it causes runtime link error, we found all broken tests and copy the broken classes with necessary modification to make it work with thrift 0.16.0, but there may still risk somewhere the UT not covered.

BTW, please change `Some classes ` exactly what they are, and move the summary to the PR desc.

yaooqinn:
the test failure here seems relevant, can we check on that? @SteNicholas

pan3793:
Updated PR description, also added **Next Steps** @yaooqinn

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HiveDelegationTokenProviderSuite.scala
zhouyifan279: I will create a PR later to fix it.

pan3793 (approved):

yaooqinn (approved):
Excellent work @SteNicholas
+1, LGTM

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HiveDelegationTokenProviderSuite.scala
yanghua: > Shall we govern these ignored tests with flaky tests together? cc @yanghua

I suggest we split them into two categories. The ignored tests may have other reasons not only flaky tests?

yaooqinn:
thanks, merged to master

Issue #1963, opened 2022-02-22
[Improvement] Make plan only mode skippable for configuable plans


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the plan only mode skip for certern plans, like use/set/reset, etc, we can add a conf to make it configurable

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

SteNicholas:
@yaooqinn, could you please assign this ticket to me?

PR #1967, opened 2022-02-23
[KYUUBI #1963] Make plan only mode skippable for configuable plans

SteNicholas:

Make plan only mode skippable for configuable plans.


SteNicholas:
@yaooqinn @pan3793 @iodone, could you please help to review this pull request?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1148,6 +1148,15 @@ object KyuubiConf {
       .checkValues(OperationModes.values.map(_.toString))
       .createWithDefault(OperationModes.NONE.toString)
 
+  val OPERATION_PLAN_ONLY_SKIP: ConfigEntry[Seq[String]] =
+    buildConf("operation.plan.only.skip")
```
yaooqinn, line 1152: operation.plan.only.excludes

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1148,6 +1148,15 @@ object KyuubiConf {
...
+      .doc("Skipped statements or commands in a PARSE, ANALYZE, OPTIMIZE, PHYSICAL, EXECUTION " +
+        "only way without executing the query.")
+      .version("1.5.0")
+      .stringConf
+      .toSequence()
+      .createWithDefault(Nil)
```
yaooqinn, line 1158: Let add the Plans that match set/reset/use here to keep the backward compatibility and better UX.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -32,7 +31,8 @@ import org.apache.kyuubi.session.Session
 class PlanOnlyStatement(
     session: Session,
     override val statement: String,
-    mode: OperationMode)
+    mode: OperationMode,
+    skip: Seq[String])
```
yaooqinn, line 35: we can move it out of the class parameters to make it simple，like

private val exculdedPlans = spark.conf.getOrElse(xxx, session.getConf.get)

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -36,6 +36,11 @@ class PlanOnlyStatement(
   extends SparkOperation(OperationType.EXECUTE_STATEMENT, session) {
 
   private val operationLog: OperationLog = OperationLog.createOperationLog(session, getHandle)
+  private val configExcludes: Option[String] =
+    spark.conf.getOption(OPERATION_PLAN_ONLY_EXCLUDES.key)
+  private val planExcludes: Seq[String] =
+    if (configExcludes.isEmpty) session.sessionManager.getConf.get(OPERATION_PLAN_ONLY_EXCLUDES)
+    else configExcludes.get.split(",").toSeq
```
yaooqinn, line 43:
```suggestion
  private val planExcludes: Seq[String] = {
    spark.conf.getOption(OPERATION_PLAN_ONLY_EXCLUDES.key).map( _.split(",").map(_.trim).toSeq)
      .getOrElse(session.sessionManager.getConf.get(OPERATION_PLAN_ONLY_EXCLUDES))
  }
```

ulysses-you, line 43: the priority of session conf should be higher than spark conf

ulysses-you (commented):
just one comment, otherwise LGTM

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
cfmcgrady: we need to split the session conf `session.sessionManager.getConf.get(OPERATION_PLAN_ONLY_EXCLUDES)` by comma as well.

SteNicholas: @cfmcgrady, the return value of the `session.sessionManager.getConf.get(OPERATION_PLAN_ONLY_EXCLUDES)` is string sequence. Thus, the above suggestion has something wrong because you map the `Option`.

cfmcgrady: Yes, you are right, I misread before :)

SteNicholas: @ulysses-you, IMO, the priority of the spark conf is higher than session conf. cc @yaooqinn

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -46,24 +51,11 @@ class PlanOnlyStatement(
...
-    className == "CreateViewStatement" ||
-    className == "CreateViewCommand"
-  }
-
   override protected def runInternal(): Unit = {
     try {
       val parsed = spark.sessionState.sqlParser.parsePlan(statement)
       parsed match {
-        case cmd if shouldDirectRun(cmd) =>
+        case cmd if planExcludes.contains(cmd.getClass.getSimpleName) =>
```
iodone, line 58: IMO, the configuration here is to provide the exact className, if you want to skip cacheTable,
`CacheTableStatement` in spark 3.0 and `CacheTableCommand` in Spark 3.2, it's too hard for users to set the right options. Could the options here be more generic, like set/reset/cacheTable/CreateTempView, rather than requiring the actual class name?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
cfmcgrady: +1 for @ulysses-you 's suggestion.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1148,6 +1148,19 @@ object KyuubiConf {
...
+      .doc("Comma-separated list of query plan names, in the form of simple class names, i.e, " +
+        "for `set abc=xyz`, the value will be `SetCommand`. For those auxiliary plans, such as " +
+        "`switch databases`, `set properties`, or `create temporary view` e.t.c, " +
+        "which are used for setup evaluating environments for analyzing actual queries, " +
+        "we can use this config to exclude them and let them take effect. " +
+        s"See also ${OPERATION_PLAN_ONLY_MODE.key}.")
+      .version("1.5.0")
+      .stringConf
+      .toSequence()
+      .createWithDefault(Seq("ResetCommand", "SetCommand", "SetNamespaceCommand", "UseStatement"))
```
iodone, line 1162: IMO, the configuration here is to provide the exact className, if you want to skip cacheTable,
`CacheTableStatement` in spark 3.0 and `CacheTableCommand` in Spark 3.2, it's too hard for users to set the right options. Could the options here be more generic, like set/reset/cacheTable/CreateTempView, rather than requiring the actual class name?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
SteNicholas: @yaooqinn, does this suggestion is wrong? As below mentioned, the priority of session conf should be higher than spark conf.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
SteNicholas: @iodone, the point above doesn't make sense. The engine side couldn't do any mapping for the user-defined options.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
yaooqinn: seesion conf has already been set to spark confs

yaooqinn: SessionManger's getConf is not session conf, its engine conf

SteNicholas: @ulysses-you, as you mentioned, the [SparkOperation.scala#L73](https://github.com/apache/incubator-kyuubi/blob/master/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala#L73) has the same problem, right?

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/PlanOnlyOperationSuite.scala
```diff
@@ -79,31 +81,25 @@ class PlanOnlyOperationSuite extends WithKyuubiServer with HiveJDBCTestHelper {
...
-  }
-
-  test("KYUUBI #1920: Plan only operations with CreateViewStatement or CreateViewCommand skiped") {
-    withSessionConf()(Map(KyuubiConf.OPERATION_PLAN_ONLY.key -> OPTIMIZE.toString))(Map.empty) {
+  test("KYUUBI #1920: Plan only operations with CreateViewStatement or CreateViewCommand") {
+    withSessionConf()(
+      Map(KyuubiConf.OPERATION_PLAN_ONLY_EXCLUDES.key -> "CreateViewStatement,CreateViewCommand"))(
+      Map.empty) {
       withJdbcStatement() { statement =>
         val result = statement.executeQuery("create temp view temp_view as select 1")
```
iodone, line 100: https://github.com/apache/incubator-kyuubi/pull/1921/files#r811719564 @cfmcgrady‘s Tips

yaooqinn: we can play the trick with `withJdbcStatement("view_temp") { ... "create temp view view_temp"}`

SteNicholas: @iodone,  using `statement.execute` could drop the `temp_view`, no need to use `statement.executeQuery` to drop the view.

yaooqinn (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
ulysses-you: I see, we have merged session conf into spark session conf. LGTM

ulysses-you (approved):

SteNicholas:
@yaooqinn, please help to merge.

yaooqinn:
thanks, merged to master

Issue #1814, opened 2022-01-21
[SUB-TASK][KPIP-2] Make flink engine local quickstart demo hadoop lib free


yanghua:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

This is a subtask of KPIP-2 #1322 .

### Motivation

Currently, the Flink engine can not start without Hadoop classpath. We should fix it.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

SteNicholas:
@yanghua, IMO, the Flink engine could start without hadoop classpath because the value of `FLINK_HADOOP_CLASSPATH` could be empty string in `flink-sql-engine.sh`.

SteNicholas:
@yanghua @yaooqinn, could you please assign this ticket to me? I have pushed a pull request for this issue.

SteNicholas:
@yanghua, as #1973 comment mentions, the Kyuubi module dependencies of the Flink engine need the hadoop client jars in the classpath. Therefore, the Flink engine couldn't be maken hadoop library free and this issue could be closed.

yaooqinn:
thanks

PR #1973, opened 2022-02-24
[KYUUBI #1814] Make flink engine local quickstart demo hadoop lib free

SteNicholas:

Make flink engine local quickstart demo hadoop lib free.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/bin/flink-sql-engine.sh
+++ b/externals/kyuubi-flink-sql-engine/bin/flink-sql-engine.sh
@@ -45,8 +45,6 @@ else
   echo -e "\nFLINK_SQL_ENGINE_HOME $FLINK_SQL_ENGINE_HOME doesn't match production directory, assuming in development environment..."
   FLINK_SQL_ENGINE_LIB_DIR="$FLINK_SQL_ENGINE_HOME/target"
   FLINK_SQL_ENGINE_JAR=$(find "$FLINK_SQL_ENGINE_LIB_DIR" -regex '.*/kyuubi-flink-sql-engine_.*\.jar$' | grep -v '\-javadoc.jar$' | grep -v '\-tests.jar$')
-  _FLINK_SQL_ENGINE_HADOOP_CLIENT_JARS=$(find $FLINK_SQL_ENGINE_LIB_DIR -regex '.*/hadoop-client-.*\.jar$' | tr '\n' ':')
-  FLINK_HADOOP_CLASSPATH="${_FLINK_SQL_ENGINE_HADOOP_CLIENT_JARS%:}"
   log_file="unused.log"
   log4j2_conf_file="file:$FLINK_CONF_DIR/log4j-session.properties" # which send all logs to console
   logback_conf_file="unused.xml"
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -140,7 +140,54 @@ object Utils extends Logging {
     dir
   }
 
-  def currentUser: String = UserGroupInformation.getCurrentUser.getShortUserName
+  /**
+   * Get the ClassLoader which loaded Kyuubi.
+   */
+  def getKyuubiClassLoader: ClassLoader = getClass.getClassLoader
+
+  /**
+   * Get the Context ClassLoader on this thread or, if not present, the ClassLoader that
+   * loaded Kyuubi.
+   *
+   * This should be used whenever passing a ClassLoader to Class.ForName or finding the currently
+   * active loader when setting up ClassLoader delegation chains.
+   */
+  def getContextOrKyuubiClassLoader: ClassLoader =
+    Option(Thread.currentThread().getContextClassLoader).getOrElse(getKyuubiClassLoader)
+
+  /**
+   * Preferred alternative to Class.forName(className), as well as
+   * Class.forName(className, initialize, loader) with current thread's ContextClassLoader.
+   */
+  def classForName[C](
+      className: String,
+      initialize: Boolean = true,
+      fallbackKyuubiClassLoader: Boolean = true): Class[C] = {
+    if (fallbackKyuubiClassLoader) {
+      Class.forName(className, initialize, getContextOrKyuubiClassLoader).asInstanceOf[Class[C]]
+    } else {
+      Class.forName(
+        className,
+        initialize,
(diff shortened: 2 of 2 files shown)


SteNicholas:
@yanghua, could you please help to review this pull request?

yaooqinn:
do we need to bundle the hadoop client jars instead?

yanghua:
> do we need to bundle the hadoop client jars instead?

+1 for this idea

pan3793:
Flink has its own abstractions instead of resuing Hadoop classes, e.g. `FileSystem`, `JobContext`, for K8s mode, user would not intend to add Hadoop jars, I'm +1 for the direction to make Flink engine runable without Hadoop jars

yanghua:
> Flink has its own abstractions instead of resuing Hadoop classes, e.g. `FileSystem`, `JobContext`, for K8s mode, user would not intend to add Hadoop jars, I'm +1 for the direction to make Flink engine runable without Hadoop jars

s.g. after discussing with @SteNicholas offline I changed my mind to support this idea. Flink has decoupled with hadoop.  Let me verify the quickstart demo e2e to find if we missed something.

SteNicholas:
> > Flink has its own abstractions instead of resuing Hadoop classes, e.g. `FileSystem`, `JobContext`, for K8s mode, user would not intend to add Hadoop jars, I'm +1 for the direction to make Flink engine runable without Hadoop jars
> 
> s.g. after discussing with @SteNicholas offline I changed my mind to support this idea. Flink has decoupled with hadoop. Let me verify the quickstart demo e2e to find if we missed something.

@yaooqinn @pan3793, WDYT?

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -140,7 +140,9 @@ object Utils extends Logging {
     dir
   }
 
-  def currentUser: String = UserGroupInformation.getCurrentUser.getShortUserName
+  def currentUser: String =
+    try UserGroupInformation.getCurrentUser.getShortUserName
+    catch { case _: Exception => sys.props.get("user.name").get }
```
pan3793, line 145: Can we introduce a util method `hadoopClassPresent`? We may need this in other places.

```
private[spark] def hadoopClassPresent: Boolean = {
    try {
      Utils.classForName("org.apache.hadoop.security.UserGroupInformation")
      true
    } catch {
      case _: ClassNotFoundException | _: NoClassDefFoundError => false
    }
  }
```

`sys.props.get("user.name").get` => `sys.props("user.name")`

pan3793:
Please also update `externals/kyuubi-flink-sql-engine/bin/flink-sql-engine.sh` and remove `_FLINK_SQL_ENGINE_HADOOP_CLIENT_JARS` from Flink classpath, let's see if anythings broken.

SteNicholas:
@pan3793, thanks for your detailed review. I have addressed your comments above mentioned. Please help to review again.

externals/kyuubi-flink-sql-engine/bin/flink-sql-engine.sh
```diff
@@ -37,31 +37,26 @@ FLINK_SQL_ENGINE_HOME="$(cd `dirname $0`/..; pwd)"
 if [[ "$FLINK_SQL_ENGINE_HOME" == "$KYUUBI_HOME/externals/engines/flink" ]]; then
   FLINK_SQL_ENGINE_LIB_DIR="$FLINK_SQL_ENGINE_HOME/lib"
   FLINK_SQL_ENGINE_JAR=$(find "$FLINK_SQL_ENGINE_LIB_DIR" -regex ".*/kyuubi-flink-sql-engine_.*\.jar")
-  FLINK_HADOOP_CLASSPATH="$INTERNAL_HADOOP_CLASSPATHS"
```
pan3793, line 40: Do we need this one? What if user provides the hadoop jar?

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -140,7 +140,54 @@ object Utils extends Logging {
...
+    Option(Thread.currentThread().getContextClassLoader).getOrElse(getKyuubiClassLoader)
+
+  /**
+   * Preferred alternative to Class.forName(className), as well as
+   * Class.forName(className, initialize, loader) with current thread's ContextClassLoader.
+   */
+  def classForName[C](
+      className: String,
+      initialize: Boolean = true,
+      noKyuubiClassLoader: Boolean = false): Class[C] = {
```
pan3793, line 165: Could you please change this variable to something positive?

pan3793: e.g. `fallbackKyuubiClassLoader`

externals/kyuubi-flink-sql-engine/bin/flink-sql-engine.sh
SteNicholas: @pan3793, the user should configure the `HADOOP_CLASSPATH` or `HADOOP_CONF_DIR` or `YARN_CONF_DIR`
because INTERNAL_HADOOP_CLASSPATHS=${HADOOP_CLASSPATH}:${HADOOP_CONF_DIR}:${YARN_CONF_DIR}

pan3793: Then after removing this, Kyuubi can not see Hadoop classes even user provides them?

yanghua:
@SteNicholas check CI.

SteNicholas:
@yanghua, the result of the CI is successful. Please help to review again.

externals/kyuubi-flink-sql-engine/bin/flink-sql-engine.sh
```diff
@@ -52,16 +52,14 @@ else
   logback_conf_file="unused.xml"
 fi
 
-FULL_CLASSPATH="$FLINK_SQL_ENGINE_JAR:$FLINK_SQL_CLIENT_JAR:$CC_CLASSPATH:$FLINK_HADOOP_CLASSPATH"
```
pan3793, line 55: This change is unnecessary, if you want to change the order of jars, just modify itself rather than inline it

pan3793: And also, I don't think the change of this script proves the flink sql engine can run without hadoop.

We should pick the Hadoop jars in production if user provides it, but we should not include `hadoop-client-*` jars in UT, right?

pan3793:
The change does not achieve the goal mentioned in PR title, the UT still depends on Hadoop jars

PR #1981, opened 2022-02-28
Remove unused artifacts for common and parent module

yaooqinn:

### DIFF
--- a/kyuubi-metrics/pom.xml
+++ b/kyuubi-metrics/pom.xml
@@ -89,11 +89,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.kyuubi</groupId>
             <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
             <version>${project.version}</version>
--- a/kyuubi-metrics/pom.xml
+++ b/kyuubi-metrics/pom.xml
@@ -89,6 +89,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.kyuubi</groupId>
             <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
             <version>${project.version}</version>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -532,6 +532,10 @@
                 <artifactId>libthrift</artifactId>
                 <version>${thrift.version}</version>
                 <exclusions>
+                    <exclusion>
+                        <groupId>org.apache.commons</groupId>
+                        <artifactId>commons-lang3</artifactId>
```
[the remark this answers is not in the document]
pan3793: Then please also exclude apache http clients jars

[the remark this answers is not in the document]
yaooqinn: nice suggestion

pan3793 (approved):

kyuubi-metrics/pom.xml
```diff
@@ -88,6 +88,11 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
+        <dependency>
```
SteNicholas, line 91: Does the guava package need to be shaded?

SteNicholas (commented):
@yaooqinn, thanks for your contribution. I left the minor comments for the guava package which need to be shaded.

kyuubi-metrics/pom.xml
ulysses-you: it is only used at Kyuubi Server, so there is no other version of guava

ulysses-you (approved):

kyuubi-metrics/pom.xml
SteNicholas: > it is only used at Kyuubi Server, so there is no other version of guava

@ulysses-you, no matter whether the guava package is only used at Kyuubi Server, the guava package is better to be shaded for the users.

pan3793: I'm afraid we not do that, because curator need it https://cwiki.apache.org/confluence/display/CURATOR/TN13

yaooqinn: Thanks, @SteNicholas  I checked the guava usages in this module, I think we can remove guava here

SteNicholas (approved):

SteNicholas:
LGTM.

yaooqinn:
thanks @pan3793 @SteNicholas @ulysses-you for the help, merged to master

PR #1995, opened 2022-03-02
[KYUUBI #1988] Tune initialization of vars depend on system level confs

pan3793:

The `new SparkConf()` and `KyuubiConf()` executes early than UT, and can not re-initialize after changing system properties, which makes conf overwriting does not work.

After the change, the `sparkConf.setIfMissing("spark.sql.catalogImplementation", defaultCat)` work as expected when we overwriting `spark.sql.catalogImplementation` in UT, and Hudi uses `in-memory` catalog to create table.

Xnip2022-03-02_19-10-06

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -71,13 +71,13 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
 
 object SparkSQLEngine extends Logging {
 
-  val sparkConf: SparkConf = new SparkConf()
+  lazy val sparkConf: SparkConf = new SparkConf()
 
-  val kyuubiConf: KyuubiConf = KyuubiConf()
+  lazy val kyuubiConf: KyuubiConf = KyuubiConf()
 
   var currentEngine: Option[SparkSQLEngine] = None
 
-  private val user = currentUser
+  private lazy val user = currentUser
 
   private val countDownLatch = new CountDownLatch(1)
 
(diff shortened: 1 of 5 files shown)


pan3793:
cc @wForget @yaooqinn

ulysses-you:
seems the failed test is relevant

SteNicholas (approved):
LGTM.

ulysses-you:
does it works ? it seems to me the problem is order of extra config and initialization config. We should setup conf first then overwrite using `withKyuubiConf`. 

See the code in `WithSparkSQLEngine#startSparkEngine`
```scala
......
withKyuubiConf.foreach { case (k, v) =>
  System.setProperty(k, v)
  kyuubiConf.set(k, v)
}

SparkSession.clearActiveSession()
SparkSession.clearDefaultSession()
SparkSQLEngine.setupConf()
spark = SparkSQLEngine.createSpark()
SparkSQLEngine.startEngine(spark)
......
```

pan3793:
I don't think the order is a problem, we use setIfMissing in `SparkSQLEngine#setupConf`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -71,41 +71,48 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
...
 
   var currentEngine: Option[SparkSQLEngine] = None
 
-  private val user = currentUser
+  private lazy val user = currentUser
 
   private val countDownLatch = new CountDownLatch(1)
 
+  SignalRegister.registerLogger(logger)
+  setupConf()
```
wForget, line 87: Should the `SparkSQLEngine.setupConf()` call in `WithSparkSQLEngine#startSparkEngine` be removed?

pan3793: No, after changing system properties, we need re-init the SparkConf and KyuubiConf to let them pick up the system properties

ulysses-you (approved):

ulysses-you:
thank, merging to master

PR #2045, opened 2022-03-06
Preparing v1.6.0-SNAPSHOT

yaooqinn:

Preparing v1.6.0-SNAPSHOT with branch-1.5 cut


yaooqinn:
the style ci is not related to this change, just because the snapshot is not in the repo, I'd merge this to fix this instead of bringing a heavy install to fix this.

Issue #2047, opened 2022-03-06
[FEATURE] Support more MySQL JDBC driver versions


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Support more MySQL JDBC driver versions.

### Motivation

The user may use the 5.x mysql driver to connect to the kyuubi server.

### Describe the solution

Match more versions of initialization SQL.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2048, opened 2022-03-06
[KYUUBI #2047] Support more MySQL JDBC driver versions

cxzl25:
Support more MySQL JDBC driver versions.
The user may use the 5.x mysql driver to connect to the kyuubi server.

https://github.com/apache/incubator-kyuubi/issues/2047

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/mysql/MySQLDialectHelper.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/mysql/MySQLDialectHelper.scala
@@ -41,33 +41,14 @@ object MySQLDialectHelper {
         |  'utf8mb4' as `@@character_set_server`,
         |  'utf8mb4' as `@@character_set_database`
         |""".stripMargin
-    // mysql-connector-java:8 initialized query
-    case sql
-        if sql.contains("select  @@session.auto_increment_increment as auto_increment_increment, @@character_set_client as character_set_client, @@character_set_connection as character_set_connection, @@character_set_results as character_set_results, @@character_set_server as character_set_server, @@collation_server as collation_server, @@collation_connection as collation_connection, @@init_connect as init_connect, @@interactive_timeout as interactive_timeout, @@license as license, @@lower_case_table_names as lower_case_table_names, @@max_allowed_packet as max_allowed_packet, @@net_write_timeout as net_write_timeout, @@performance_schema as performance_schema, @@query_cache_size as query_cache_size, @@query_cache_type as query_cache_type, @@sql_mode as sql_mode, @@system_time_zone as system_time_zone, @@time_zone as time_zone, @@transaction_isolation as transaction_isolation, @@wait_timeout as wait_timeout") =>
-      """SELECT
-        |  1                    AS auto_increment_increment,
-        |  'utf8mb4'            AS character_set_client,
-        |  'utf8mb4'            AS character_set_connection,
-        |  'utf8mb4'            AS character_set_results,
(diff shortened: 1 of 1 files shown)


pan3793 (approved):
The current approach is fragile, it should be replaced by antlr based SQL parser eventually.

cxzl25:
[mysql_init_select.sql.txt](https://github.com/apache/incubator-kyuubi/files/8192825/mysql_init_select.sql.txt)


Grab the initialization sql of mysql-connector-java version from 5.1.20 to 8.0.28, and get the parameter list after deduplication.

```
@@auto_increment_increment
@@character_set_client
@@character_set_connection
@@character_set_results
@@character_set_server
@@collation_connection
@@collation_server
@@have_query_cache
@@init_connect
@@interactive_timeout
@@license
@@lower_case_table_names
@@max_allowed_packet
@@net_buffer_length
@@net_write_timeout
@@performance_schema
@@query_cache_size
@@query_cache_type
@@sql_mode
@@system_time_zone
@@time_zone
@@transaction_isolation
@@tx_isolation
@@wait_timeout
```

cxzl25:
> The current approach is fragile, it should be replaced by antlr based SQL parser eventually.

Yes, I tried to use the mysql protocol on Friday and encountered this problem, and I also considered whether to consider introducing antlr.

pan3793:
Kyuubi will introduce Antlr eventually, we have a prototype discussion previously, we'd like to introduce some Kyuubi defined SQL-like grammer, e.g. `SHOW ENGINES [LIKE ...]`, `ALTER SESSION SET ...`, it's something like [ShardingSphere DistSQL](https://github.com/apache/shardingsphere/releases)

cxzl25:
> Kyuubi will introduce Antlr eventually, we have a prototype discussion previously, we'd like to introduce some Kyuubi defined SQL-like grammer, e.g. `SHOW ENGINES [LIKE ...]`, `ALTER SESSION SET ...`, it's something like [ShardingSphere DistSQL](https://github.com/apache/shardingsphere/releases)

Make sense. 
May be I missed the previous discussion.

yaooqinn (approved):

pan3793:
> May be I missed the previous discussion.

Sorry, it happened in offline discussion.

Don't hesitate if you think we need to introduce antlr deps, even now we don't have detail design of the Kyuubi dialect  SQL grammar, but I need some intuitive commands like `SHOW ENGINES`, `SHOW SESSIONS`, and we also need to implement a basic single table `SELECT ... FROM ... WHERE ... LIMIT ... OFFSET ...` for MySQL.

cxzl25:
> Could not find artifact org.apache.kyuubi:kyuubi-XXX_2.12:jar:1.6.0-SNAPSHOT in apache.snapshots

apache snapshot does not have deploy 1.6 version, causing UT to fail.

https://repository.apache.org/content/groups/snapshots/org/apache/kyuubi/

cxzl25:
> Don't hesitate if you think we need to introduce antlr deps, even now we don't have detail design of the Kyuubi dialect SQL grammar, but I'm sure we need some intuitive commands like `SHOW ENGINES`, `SHOW SESSIONS`, and we also need to implement a basic single table `SELECT ... FROM ... WHERE ... LIMIT ... OFFSET ...` for MySQL.

ok, I will continue to pay attention to the development of the kyuubi project and see what I can do, thanks.

ulysses-you (approved):

SteNicholas (approved):
LGTM.

pan3793:
Thanks, merging to master/1.5

Issue #2072, opened 2022-03-09
[Improvement]  Improve rest server behavior 


ulysses-you:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

1. use host address rather than host name, since we also use host address in thrift server
2. remove alert in web ui, it's noisy..

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2073, opened 2022-03-09
[KYUUBI #2072] Improve rest server behavior

ulysses-you:

1. use host address rather than host name, since we also use host address in thrift server
2. remove alert in web ui, it's noisy..

### DIFF
--- a/kyuubi-server/src/main/resources/org/apache/kyuubi/ui/static/kyuubi.js
+++ b/kyuubi-server/src/main/resources/org/apache/kyuubi/ui/static/kyuubi.js
@@ -43,5 +43,4 @@ $(function(){
 		});
 	}
 	$('.item.server.version.right.padded').val(kyuubiVersion)
-	alert(kyuubiVersion);
 });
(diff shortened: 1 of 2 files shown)


ulysses-you:
cc @cfmcgrady @yaooqinn

yaooqinn:
please fix the PR title to be more specific and triage PR by labels

yaooqinn (approved):

cfmcgrady (approved):

kyuubi-server/src/main/resources/org/apache/kyuubi/ui/static/kyuubi.js
```diff
@@ -43,5 +43,4 @@ $(function(){
 		});
 	}
 	$('.item.server.version.right.padded').val(kyuubiVersion)
-	alert(kyuubiVersion);
```
SteNicholas, line 46: It's better to replace this with `console.log(kyuubiVersion)`.

ulysses-you: it's a common function and will be invoked in many places..

yaooqinn:
thanks, merged to master

Issue #2081, opened 2022-03-09
[Improvement] YARN_CONF_DIR shall be added to kyuubi server classpath as HADOOP_CONF_DIR


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

YARN_CONF_DIR aslo contains configurations for kyuubi server to respect

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

jiaoqingbo:
please assign to me

ulysses-you:
> please assign to me

@jiaoqingbo done, thank you !

jiaoqingbo:
> > please assign to me
> 
> @jiaoqingbo done, thank you !

please assign to @SteNicholas，he has submitted a more reasonable pr ，thx

ulysses-you:
thank you @jiaoqingbo and @SteNicholas !

PR #2082, opened 2022-03-10
[KYUUBI #2081] YARN_CONF_DIR shall be added to kyuubi server classpath as HADOOP_CONF_DIR

SteNicholas:

YARN_CONF_DIR shall be added to kyuubi server classpath as HADOOP_CONF_DIR.


conf/kyuubi-env.sh.template
```diff
@@ -38,6 +38,7 @@
 # - KYUUBI_WORK_DIR_ROOT    Root directory for launching sql engine applications.
 #                           (Default: $KYUUBI_HOME/work)
 # - HADOOP_CONF_DIR         Directory containing the Hadoop / YARN configuration to use.
+# - YARN_CONF_DIR         Directory containing the YARN configuration to use.
```
yaooqinn, line 41: intention

yaooqinn, line 41: what are the precedences between these, will they fail some engines if we propagate them together to the engine side?

SteNicholas: The classpath is `${HADOOP_CLASSPATH}:${HADOOP_CONF_DIR}:${YARN_CONF_DIR}` is Flink. Therefore this will not fails.

yaooqinn (approved):
LGTM

yaooqinn:
thanks, merged to master

PR #2095, opened 2022-03-10
Remove useless logic about add conf when create a new engine

zwangsheng:

Remove useless logic about add conf when create a new engine
This code worked before commit [575e9c95999d70ac88e8ee0bdd3b58bc016bc934] which unset `FRONTEND_THRIFT_BINARY_BIND_HOST` when copy conf.

The code which removed in this PR, only help kyuubi on k8s with spark client mode. Can be replace by set `spark.driver.host`.
Considering this approach is not recommended, so remove this code in this PR.


yaooqinn (approved):

pan3793:
Thanks, merging to master

Issue #2112, opened 2022-03-12
[Improvement] Improve the compatibility of queryTimeout in more version clients


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Clients less than version 2.1 have no [HIVE-4924](https://issues.apache.org/jira/browse/HIVE-4924) Patch, no `queryTimeout `parameter and no `TIMEOUT` status. When the server enables `kyuubi.operation.query.timeout`, this will cause the client of the lower version to get stuck.

### How should we improve?

Check thrift protocol version <= HIVE_CLI_SERVICE_PROTOCOL_V8, convert TIMEDOUT_STATE to CANCELED.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2113, opened 2022-03-12
[KYUUBI #2112]  Improve the compatibility of queryTimeout in more version clients

cxzl25:
https://github.com/apache/incubator-kyuubi/issues/2112

Clients less than version 2.1 have no [HIVE-4924](https://issues.apache.org/jira/browse/HIVE-4924) Patch, no queryTimeout parameter and no TIMEOUT status. When the server enables kyuubi.operation.query.timeout, this will cause the client of the lower version to get stuck.

Introduce a parameter, when it is turned on, no matter what the version of the client is, the queryTimeout parameter is not set, and the TIMEOUT state is converted to CANCELED. If queryTimeout is set, the TIMEOUT state is still maintained.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
@@ -17,6 +17,8 @@
 
 package org.apache.kyuubi.operation
 
+import java.util.concurrent.TimeUnit
+
 import scala.collection.JavaConverters._
 
 import org.apache.hive.service.rpc.thrift.TGetOperationStatusResp
@@ -25,6 +27,7 @@ import org.apache.thrift.TException
 
 import org.apache.kyuubi.KyuubiSQLException
 import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.config.KyuubiConf.OPERATION_QUERY_TIMEOUT
 import org.apache.kyuubi.events.{EventLogging, KyuubiOperationEvent}
 import org.apache.kyuubi.operation.FetchOrientation.FETCH_NEXT
 import org.apache.kyuubi.operation.OperationState.OperationState
@@ -51,6 +54,22 @@ class ExecuteStatement(
     session.sessionManager.getConf.get(KyuubiConf.OPERATION_STATUS_POLLING_MAX_ATTEMPTS)
   }
 
+  private val operationQueryTimeoutCompatibleState = {
+    session.sessionManager.getConf.get(KyuubiConf.OPERATION_QUERY_TIMEOUT_COMPATIBLE_STATE)
+  }
+
+  private val timeout = {
+    val systemQueryTimeout = session.sessionManager.getConf.get(OPERATION_QUERY_TIMEOUT)
+      .map(TimeUnit.MILLISECONDS.toSeconds)
+
+    systemQueryTimeout match {
+      case Some(systemQueryTimeout) if queryTimeout > 0 =>
+        math.min(systemQueryTimeout, queryTimeout)
+      case Some(systemQueryTimeout) => systemQueryTimeout
+      case None => queryTimeout
+    }
+  }
+
   override def getOperationLog: Option[OperationLog] = Option(_operationLog)
 
   override def beforeRun(): Unit = {
@@ -68,7 +87,7 @@ class ExecuteStatement(
       // We need to avoid executing query in sync mode, because there is no heartbeat mechanism
       // in thrift protocol, in sync mode, we cannot distinguish between long-run query and
       // engine crash without response before socket read timeout.
-      _remoteOpHandle = client.executeStatement(statement, confOverlay, true, queryTimeout)
+      _remoteOpHandle = client.executeStatement(statement, confOverlay, true, timeout)
       setHasResultSet(_remoteOpHandle.isHasResultSet)
     } catch onError()
   }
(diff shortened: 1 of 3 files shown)


cxzl25:
When the client has no `TIMEOUT` enumeration value, isSetOperationState is always false, and the loop will not exit.

org.apache.hive.jdbc.HiveStatement#execute

```java
    while (!operationComplete) {
        if (statusResp.isSetOperationState()) {
          switch (statusResp.getOperationState()) {
          case XXX:
```


## Fix

image

pan3793:
Is there possible to use protocol version to dynamic change the state?

cxzl25:
> Is there possible to use protocol version to dynamic change the state?

In the statement, we can get the protocol version through `getSession.protocol` .

`HIVE_CLI_SERVICE_PROTOCOL_V8` was introduced by [HIVE-14191](https://issues.apache.org/jira/browse/HIVE-14191) and took effect in Hive 2.1.1 and 2.2.0, but querytimeout was introduced by [HIVE-4924](https://issues.apache.org/jira/browse/HIVE-4924) and was available in Hive 2.1.0.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
```diff
@@ -125,7 +144,17 @@ class ExecuteStatement(
             setState(OperationState.CANCELED)
 
           case TIMEDOUT_STATE =>
-            setState(OperationState.TIMEOUT)
+            val state =
+              if (operationQueryTimeoutCompatibleState) {
+                if (queryTimeout > 0) {
```
yaooqinn, line 149: do we need this？TIMEDOUT_STATE only happens when `queryTimeout>0` is satisfied on the engine side.
this may be much simpler

case TIMEDOUT_STATE if operationQueryTimeoutCompatibleState => OperationState.CANCELED
case TIMEDOUT_STATE => OperationState.TIMEOUT

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationManagerTimeoutCompatibleSuite.scala
```diff
@@ -0,0 +1,59 @@
...
+
+class KyuubiOperationManagerTimeoutCompatibleSuite extends WithKyuubiServer
+  with HiveJDBCTestHelper {
+  override protected val conf: KyuubiConf = {
+    KyuubiConf().set(OPERATION_QUERY_TIMEOUT.key, "PT1S").set(
+      OPERATION_QUERY_TIMEOUT_COMPATIBLE_STATE.key,
+      "true")
+  }
+
+  test("clientTimeout is not set, timeout state switches to cancel state") {
```
yaooqinn, line 34: shall we move this test to KyuubiOperationManagerSuite

we can use org.apache.kyuubi.operation.HiveJDBCTestHelper#withSessionConf to set session-level configurations for a single test

ulysses-you:
is it possible change operation state dynamically through thrift protocol version ?

It seems we can check if the thrift protocol version <= [HIVE_CLI_SERVICE_PROTOCOL_V8](https://github.com/apache/hive/blob/branch-2.0/service/src/gen/thrift/gen-javabean/org/apache/hive/service/cli/thrift/TProtocolVersion.java) and disable timeout, since Hive-2.1 introduce [HIVE_CLI_SERVICE_PROTOCOL_V9](https://github.com/apache/hive/commit/cfb11f8039c8d9176a77c2c1d7f3624f8d378474)

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
cxzl25: It was considered before that the server has enabled `queryTimeout`, but although the client supports this feature, `queryTimeout` is not set.
Now change it a little easier, turn it on and convert it.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationManagerTimeoutCompatibleSuite.scala
cxzl25: After using `withSessionConf` to set parameters, the session of `ExecuteStatement` needs to be converted into `KyuubiSessionImpl` to get `sessionConf`.
So i did not modify.

yaooqinn (approved):

yaooqinn:
> is it possible change operation state dynamically through thrift protocol version ?
> 
> It seems we can check if the thrift protocol version <= [HIVE_CLI_SERVICE_PROTOCOL_V8](https://github.com/apache/hive/blob/branch-2.0/service/src/gen/thrift/gen-javabean/org/apache/hive/service/cli/thrift/TProtocolVersion.java) and disable timeout, since Hive-2.1 introduce [HIVE_CLI_SERVICE_PROTOCOL_V9](https://github.com/apache/hive/commit/cfb11f8039c8d9176a77c2c1d7f3624f8d378474)

Oh I missed this comment, seems reasonable to me, We can return corresponding state by hive client protocol

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
```diff
@@ -124,6 +124,11 @@ class ExecuteStatement(
           case CANCELED_STATE =>
             setState(OperationState.CANCELED)
 
+          case TIMEDOUT_STATE
+              if getSession.protocol.getValue <=
+                TProtocolVersion.HIVE_CLI_SERVICE_PROTOCOL_V8.getValue =>
```
ulysses-you, line 129: can we also add some comments to describe the reason why we do this

pan3793 (approved):

ulysses-you (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/ExecuteStatement.scala
cxzl25: done

ulysses-you:
thanks, merging to master

PR #2122, opened 2022-03-13
[KYUUBI #2102] Support to retry the internal thrift request call and add engine liveness probe to enable fast fail before retry

### LINKED ISSUE #2102: [FEATURE] Enhance the retry logic for thrift rpc
### Code of Conduct

### Search before asking

### Describe the feature

For the ExecuteStatement operation, if the internal thrift connection(between kyuubi server and kyuubi engine) has broken, it still retry and costs long time.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

- [x] Yes I am willing to submit a PR!

turboFei:

To close #2102 

Support to retry all the internal thrift request calls(except RenewDelegationToken now), and fast fail if the remote engine is not stable or not alive.

In this PR, it supports engine liveness probe.

If it is enabled, a companion thrift client will be created and open a liveness probe session when opening remote engine session.
It will send some simple thrift request(GetInfo) to check whether the remote engine is alive, and fast fail before retry if remote engine is not connectable.

#### Why not use the same thrift client to check engine liveness before retry?
I tried that, but met `out of resp sequence` error.
For example:
1. send getOperationStatus request
2.  read time out
3. send GetInfoType request
4. receive getOperationStatus response (out of resp sequence)


turboFei:
For method:
```
  /**
   * Lock every rpc call to send them sequentially
   */
  private def withLockAcquired[T](block: => T): T = {
    try {
      lock.lock()
      if (!protocol.getTransport.isOpen) {
        throw KyuubiSQLException.connectionDoesNotExist()
      }
      block
    } finally lock.unlock()
  }
```
It has checked whether the underlying socket is connected, so I wonder whether this pr will take affect.

turboFei:
According to KyuubiConnection::isValid:
```
  @Override
  public boolean isValid(int timeout) throws SQLException {
    if (timeout < 0) {
      throw new SQLException("timeout value was negative");
    }
    boolean rc = false;
    try {
      String productName =
          new KyuubiDatabaseMetaData(this, client, sessHandle).getDatabaseProductName();
      rc = true;
    } catch (SQLException e) {
      // IGNORE
    }
    return rc;
  }
```
It seems that we need send a simple rpc to check whether the connection is valid.

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -55,6 +57,47 @@ class KyuubiSyncThriftClient private (protocol: TProtocol)
...
+      } catch {
+        case e: Throwable =>
+          error(s"The internal connection is not valid, mark the thrift client as closed", e)
+          setClosed()
+      }
+      isValid
+    }
+
+    try {
+      withLockAcquired(block)
```
ulysses-you, line 84: It seems the lock should be high level. It shoud be atomical even if we retry one operation with several rpcs. That said  the lock is working between operation and operation, not rpc and rpc.

turboFei: thanks

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -17,31 +17,75 @@
...
+          try {
+            client.GetInfo(tGetInfoReq).getInfoValue.getStringValue
+            aliveProbeConnectionBroken = false
+            remoteEngineBroken = false
+          } catch {
+            case e: Throwable =>
+              if (!aliveProbeConnectionBroken) {
+                error(s"The alive probe connection is broken, assume the engine is not alive", e)
+                aliveProbeConnectionBroken = true
+                remoteEngineBroken = true
```
ulysses-you, line 76: It will be flaky if the probe change to broke as long as fail. shall we also should try with some attempts ?

ulysses-you: We can respect the timeout inside Spark that driver think a executor is lose if executor does not send heartbeat after 120s.

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -69,27 +140,49 @@ class KyuubiSyncThriftClient private (protocol: TProtocol)
...
-      if (protocol.getTransport.isOpen) protocol.getTransport.close()
+      Seq(protocol).union(engineAliveProbeProtocol.toSeq).foreach { tProtocol =>
+        if (tProtocol.getTransport.isOpen) tProtocol.getTransport.close()
+      }
+      Option(engineAliveThreadPool).foreach(_.shutdown())
+      if (_aliveProbeSessionHandle != null && !remoteEngineBroken) {
+        engineAliveProbeClient.foreach { client =>
+          Utils.tryLogNonFatalError {
+            val req = new TCloseSessionReq(_aliveProbeSessionHandle)
+            val resp = client.CloseSession(req)
```
ulysses-you, line 181: the odering seems wrong, you has closed the transport.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala
```diff
@@ -216,6 +216,26 @@ class KyuubiOperationPerConnectionSuite extends WithKyuubiServer with HiveJDBCTe
...
+      KyuubiConf.ENGINE_REQUEST_TIMEOUT.key -> "100"))(Map.empty)(Map.empty) {
+      withSessionHandle { (client, handle) =>
+        val executeStmtReq = new TExecuteStatementReq()
+        executeStmtReq.setStatement("select java_method('java.lang.System', 'exit', 1)")
+        executeStmtReq.setSessionHandle(handle)
+        executeStmtReq.setRunAsync(false)
+        val startTime = System.currentTimeMillis()
+        val executeStmtResp = client.ExecuteStatement(executeStmtReq)
+        assert(executeStmtResp.getStatus.getStatusCode === TStatusCode.ERROR_STATUS)
+        assert(System.currentTimeMillis() - startTime < 20 * 1000)
```
ulysses-you, line 235: should be assert `>` here ? we use more time due to the attempts

ulysses-you (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
turboFei: thanks

turboFei:
thanks, merging to master

yaooqinn:
https://github.com/apache/incubator-kyuubi/issues/2299 there are test failures caused by this pr

Issue #2123, opened 2022-03-13
[Improvement] Output engine information after openEngineSession call fails


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Now after the openEngineSession call fails, there is no engine information output, and it is more difficult to know which engine has the problem in pool mode.

### How should we improve?

Output engine information after catch exception.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2124, opened 2022-03-13
[KYUUBI #2123] Output engine information after openEngineSession call fails

cxzl25:
https://github.com/apache/incubator-kyuubi/issues/2123

Now after the openEngineSession call fails, there is no engine information output, and it is more difficult to know which engine has the problem in pool mode.


pan3793 (approved):

cxzl25:
## Fix
image

yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #2024, opened 2022-03-04
[Subtask] Hive Backend Engine - ProcBuilder for HiveEngine


yaooqinn:

zhaomin1423:
I am willing to do it.

PR #2146, opened 2022-03-15
[KYUUBI #2024] Hive Backend Engine - ProcBuilder for HiveEngine

zhaomin1423:

Add ProcBuilder for HiveEngine.


kyuubi-common/src/main/scala/org/apache/kyuubi/engine/EngineType.scala
```diff
@@ -23,5 +23,5 @@ package org.apache.kyuubi.engine
 object EngineType extends Enumeration {
   type EngineType = Value
 
-  val SPARK_SQL, FLINK_SQL, TRINO = Value
+  val SPARK_SQL, FLINK_SQL, TRINO, HIVE_SQL = Value
```
pan3793, line 26: I think we should use `HIVE` here, for Flink and Spark, they provide user interfaces not only SQL, for Trino and Hive, only SQL, `HIVE_SQL` is quite redundant.

pan3793: Oh, I see we already named the module `kyuubi-hive-sql-engine`, `kyuubi-trino-sql-engine` looks abnormal then...

externals/kyuubi-hive-sql-engine/bin/hive-sql-engine.sh
```diff
@@ -0,0 +1,42 @@
...
+  HIVE_CLIENT_JAR="$HIVE_ENGINE_JAR"
+  HIVE_CLIENT_JARS_DIR="$HIVE_ENGINE_HOME/jars"
+else
+  echo "\nHIVE_ENGINE_HOME $HIVE_ENGINE_HOME doesn't match production directory, assuming in development environment..."
+  HIVE_CLIENT_JAR=$(find $HIVE_ENGINE_HOME/target -regex '.*/kyuubi-hive-sql-engine_.*.jar$' | grep -v '\-sources.jar$' | grep -v '\-javadoc.jar$' | grep -v '\-tests.jar$')
+  HIVE_CLIENT_JARS_DIR=$(find $HIVE_ENGINE_HOME/target -regex '.*/jars')
+fi
+
+HIVE_CLIENT_CLASSPATH="$HIVE_CLIENT_JARS_DIR/*"
+FULL_CLASSPATH="$HIVE_CLIENT_CLASSPATH:$HIVE_CLIENT_JAR:$HADOOP_CONF_DIR:$HIVE_CONF_DIR"
```
pan3793, line 35: Do we need consider `YARN_CONF_DIR`?

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/EngineType.scala
zhaomin1423: So, whether to rename them anywhere or keep them.

externals/kyuubi-hive-sql-engine/bin/hive-sql-engine.sh
zhaomin1423: Thanks, added.

pan3793 (approved):

pan3793:
Thanks, merging to master

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/EngineType.scala
pan3793: cc @yaooqinn

yaooqinn:
> Thanks, merging to master

Issue #2159, opened 2022-03-17
[SUB-TASK][KPIP-2] Keep Flink exeception newline no lose.


deadwind4:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Like this comment https://github.com/apache/incubator-kyuubi/issues/2154#issuecomment-1070179714 , the `newline` separator of execution message has been lost.

Like the following picture.
![image](https://user-images.githubusercontent.com/12015546/158770047-5806262b-4ccc-468d-8ec0-0775bda78add.png)

The `flink-sql-engine.sh` is the following.
```
+ exec //bin/java -Dline.separator= -Dhttp.proxyPort=7890  .... org.apache.kyuubi.engine.flink.FlinkSQLEngine
```

When I removed `-Dline.separator=` in scala code.
![image](https://user-images.githubusercontent.com/12015546/158770881-8ac389e3-16a1-4ca8-9f65-4b86db3fb270.png)

### How should we improve?

The `flink-sql-engine.sh` will disable '\n', so I want delete `-Dline.separator=` from `%FLINK_SQL_ENGINE_DYNAMIC_ARGS` enviroments. 
Add a scala patter-matching filter in `FlinkProcessBuilder` class to delete `line.separator`.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

deadwind4:
cc @SteNicholas @pan3793  What do you think?

pan3793:
Thanks for investigating this issue, I agree with you, wishing your PR.

PR #2161, opened 2022-03-17
[KYUUBI #2159][FLINK] Prevent loss of exception message line separator

deadwind4:

Flink SQL Exception message loses line separator.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
@@ -99,9 +99,11 @@ class FlinkProcessBuilder(
     ("FLINK_SQL_ENGINE_JAR" -> mainResource.get) +
     ("FLINK_SQL_ENGINE_DYNAMIC_ARGS" ->
       conf.getAll.map { case (k, v) =>
-        k match {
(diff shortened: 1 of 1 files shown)


pan3793:
Except for this one, are there other properties look unreasonable?

yaooqinn (approved):

deadwind4:
I will introduce the white list in this PR.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
```diff
@@ -99,9 +99,11 @@ class FlinkProcessBuilder(
     ("FLINK_SQL_ENGINE_JAR" -> mainResource.get) +
     ("FLINK_SQL_ENGINE_DYNAMIC_ARGS" ->
       conf.getAll.map { case (k, v) =>
-        k match {
-          case "line.separator" => ""
-          case _ => s"-D$k=$v"
+        if (k.startsWith("kyuubi.") || k.startsWith("flink.")
```
pan3793, line 102: We can use `conf.getAll.filter{ ... } .map { ... }.mkString(" ")` to make it more readable

pan3793 (approved):

yaooqinn:
thanks, merged to master

Issue #2119, opened 2022-03-13
[Improvement] Support output progress bar in Spark engine


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In the `ExecuteStatement#waitStatementComplete` method, only output Query is Running to the log, the user does not know the progress of the query.

### How should we improve?

Add a listener to the Spark engine to output the stage progress of the current statement.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

cxzl25:
## Current

![image](https://user-images.githubusercontent.com/3898450/158053283-f0c5f37a-ccc9-4100-a70d-4583f2424c41.png)

## Fix

![image](https://user-images.githubusercontent.com/3898450/158053320-f0c96ae6-106a-4fb8-92cd-da3934792c4f.png)

yaooqinn:
looks great

PR #2162, opened 2022-03-17
[KYUUBI #2119] Support output progress bar in Spark engine

cxzl25:
In the `ExecuteStatement#waitStatementComplete` method, only output Query is Running to the log, the user does not know the progress of the query.

close https://github.com/apache/incubator-kyuubi/issues/2119

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SQLOperationListener.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SQLOperationListener.scala
@@ -18,6 +18,7 @@
 package org.apache.spark.kyuubi
 
 import java.util.Properties
+import java.util.concurrent.ConcurrentHashMap
 
 import org.apache.spark.scheduler._
 import org.apache.spark.sql.SparkSession
@@ -43,6 +44,11 @@ class SQLOperationListener(
   private val activeJobs = new java.util.HashSet[Int]()
   private val activeStages = new java.util.HashSet[Int]()
   private var executionId: Option[Long] = None
+  private val liveStages = new ConcurrentHashMap[StageAttempt, StageInfo]()
+
+  private val consoleProgressBar = new SparkConsoleProgressBar(
+    operation,
+    liveStages)
 
   def getExecutionId: Option[Long] = executionId
 
@@ -98,6 +104,9 @@ class SQLOperationListener(
         val stageInfo = stageSubmitted.stageInfo
         val stageId = stageInfo.stageId
         activeStages.add(stageId)
+        liveStages.put(
+          StageAttempt(stageId, stageInfo.attemptNumber()),
+          new StageInfo(stageId, stageInfo.numTasks))
         withOperationLog {
           info(s"Query [$operationId]: Stage $stageId started with ${stageInfo.numTasks} tasks," +
             s" ${activeStages.size()} active stages running")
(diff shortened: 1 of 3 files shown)


cxzl25:
image

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkConsoleProgressBar.scala
```diff
@@ -0,0 +1,129 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.collection.JavaConverters._
+
+import org.apache.kyuubi.Logging
+import org.apache.kyuubi.operation.Operation
+
+class SparkConsoleProgressBar(
+    operation: Operation,
+    liveStages: ConcurrentHashMap[StageAttempt, StageInfo])
```
ulysses-you, line 32: is it possible to reuse `AppStatusStore` just like Spark's implementation `ConsoleProgressBar`.

ulysses-you:
thank you @cxzl25 for the work !

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkConsoleProgressBar.scala
cxzl25: Yes, the previously implemented v1 version is obtained through `SparkContext.statusStore.activeStages()`, but each `Operation` registers a `SQLOperationListener`, so I avoid each
Operation calls `SparkContext.statusStore.activeStages()`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SQLOperationListener.scala
```diff
@@ -43,6 +44,11 @@ class SQLOperationListener(
   private val activeJobs = new java.util.HashSet[Int]()
   private val activeStages = new java.util.HashSet[Int]()
   private var executionId: Option[Long] = None
+  private val liveStages = new ConcurrentHashMap[StageAttempt, StageInfo]()
+
+  private val consoleProgressBar = new SparkConsoleProgressBar(
```
cxzl25, line 49: TODO add a configuration
When true, show the progress bar in the session log.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkConsoleProgressBar.scala
```diff
@@ -0,0 +1,129 @@
...
+  extends Logging {
+  // Carriage return
+  private val CR = '\r'
+  // Update period of progress bar, in milliseconds
+  private val updatePeriodMSec = 200L
```
cxzl25, line 37: TODO add a configuration.
Configure the update interval.

ulysses-you:
@cxzl25 I think this todo is useful if we merge this feature, the log may become nosiy. We can address it as follow-up.

```
// TODO: Make it configurable
```

cxzl25:
> @cxzl25 I think this todo is useful if we merge this feature, the log may become nosiy. We can address it as follow-up.

I should be able to add the configuration today, wait for me.

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #2029, opened 2022-03-04
[Subtask] Hive Backend Engine - Operation Logs


yaooqinn:

zhaomin1423:
I want to do it.

yaooqinn:
assigned

zhaomin1423:
> assigned

Thanks.

PR #2174, opened 2022-03-18
[KYUUBI #2029] Hive Backend Engine - Operation Logs

KenjiFujima:

Hive Backend Engine - Operation Logs.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
@@ -20,8 +20,7 @@ package org.apache.kyuubi.engine.spark.session
 import org.apache.hive.service.rpc.thrift.TProtocolVersion
 import org.apache.spark.sql.SparkSession
 
-import org.apache.kyuubi.{KyuubiSQLException, Utils}
-import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.KyuubiSQLException
 import org.apache.kyuubi.config.KyuubiConf._
 import org.apache.kyuubi.engine.ShareLevel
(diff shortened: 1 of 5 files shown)


externals/kyuubi-hive-sql-engine/src/main/scala/org/apache/kyuubi/engine/hive/operation/ExecuteStatement.scala
```diff
@@ -31,6 +32,11 @@ class ExecuteStatement(
     override val shouldRunAsync: Boolean,
     queryTimeout: Long)
   extends HiveOperation(OperationType.EXECUTE_STATEMENT, session) {
+
+  private val operationLog: OperationLog = OperationLog.createOperationLog(session, getHandle)
```
yaooqinn, line 36: this is not necessary. this operation file will never be visit at read side.

I have already reimplemented the interface, we shall add a unit test for this to verify it's ok or not(which we shall find out why)

https://github.com/apache/incubator-kyuubi/blob/f22a14f1e50af6faad2a64902367af3f63a4ffc9/externals/kyuubi-hive-sql-engine/src/main/scala/org/apache/kyuubi/engine/hive/operation/HiveOperationManager.scala#L96-L128

KenjiFujima: @yaooqinn, I'm confusing about whether the `Log4j12DivertAppender` needs the operation. The operation log is used for setting the current operation log.

KenjiFujima:
@yaooqinn @pan3793, could you please help to review this again?

yaooqinn:
This PR seems wrong,

1. The PR is about to fix/support hive operation log according to the PR title, but the implementation does not tough any hive engine side code.
2. the server log root use a different base dir to logs of engines, this PR seems to remove this part

yaooqinn:
closed by https://github.com/apache/incubator-kyuubi/pull/2586

PR #2175, opened 2022-03-18
Improve CI with cancel & concurrency & paths filter

yaooqinn:

Better ASF GHA resource usage, see more https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status

- cancel-in-progress enabled. w/ this, a previous build can be canceled if new commit(s) have arrived
- style check w/o install. currently, the style check takes 8~10min to finish because an install step before it. now, it shall return within a minute. note that, this might 'fail' itself when we introduce a new module and others depend on it, but it's ok to verify by the log whether a PR ok to be merged.
- dependency check runs only when pom.xml updates. this flow now runs only pom changes as it is the only way that may cause dependency changes.

**Canceled**: https://github.com/apache/incubator-kyuubi/runs/5604809742?check_suite_focus=true

image

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -52,8 +52,6 @@ jobs:
         with:
           distribution: zulu
           java-version: 8
-          cache: 'maven'
-          check-latest: false
       - name: Scalastyle with maven
         run: build/mvn -T 4 scalastyle:check ${{ matrix.profiles }} -pl '!:kyuubi-codecov_2.12,!:kyuubi-assembly_2.12'
       - name: Upload scalastyle report
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -31,7 +31,7 @@ concurrency:
 # Well, sometimes when we introduce a new module, it may 'fail' the style check for missing
 # dependency we just create for other module to inherit.
 # We can ignore this failure and merge the PR, if we see the style check pass on the new module
- # or via a local quick verification.
+# or via a local quick verification.
 jobs:
   linter:
     name: Style check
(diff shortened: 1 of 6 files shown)


yaooqinn:
cc @pan3793 @ulysses-you @turboFei @SteNicholas @hddong thanks.

pan3793 (approved):

.github/workflows/style.yml
```diff
@@ -15,19 +15,26 @@
...
 
+concurrency:
+  group: linter-${{ github.ref }}
+  cancel-in-progress: true
+
+# no install runs before style check because it is expensive, we can skip build if style fails
+# Well, sometimes when we introduce a new module, it may 'fail' the style check for missing
+# dependency we just create for other module to inherit.
+# We can ignore this failure and merge the PR, if we see the style check pass on the new module
+  # or via a local quick verification.
```
pan3793, line 34: nit: indent

yaooqinn, line 34:
```suggestion
 # or via a local quick verification.
```

.github/workflows/style.yml
```diff
@@ -15,19 +15,26 @@
...
 
+concurrency:
+  group: linter-${{ github.ref }}
+  cancel-in-progress: true
+
+# no install runs before style check because it is expensive, we can skip build if style fails
+# Well, sometimes when we introduce a new module, it may 'fail' the style check for missing
+# dependency we just create for other module to inherit.
+# We can ignore this failure and merge the PR, if we see the style check pass on the new module
+ # or via a local quick verification.
```
yaooqinn, line 34:
```suggestion
# or via a local quick verification.
```

yaooqinn:
thanks, merged to master

Issue #2178, opened 2022-03-19
[Improvement] Add abstract session store to store session


zhaomin1423:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Add abstract session store to store session, It is convenient for session storage.
Releated issue: https://github.com/apache/incubator-kyuubi/issues/1754#issue-1101298742

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

Issue #1754, opened 2022-01-13
[FEATURE] Make external KV store as an option to store session and operation handles


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Currently, we use in-memory maps to store session and operation handles, is it possible to store them via leveldb, zookeeper, redis? Then we can make kyuubi server free to rebooting as the actual executions are on the engine side

### Motivation

make kyuubi server stateless

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

link3280:
Big +1 for this feature! Just to put my two cents:

- Embedded database like leveldb doesn't provide high availability alone. -1 to leveldb/rocksdb.
- Redis doesn't provide perfect data integrity, thus I think it might not be a good option for metadata, but in the case of Kyuubi I'm not sure. +0 to Redis.
- Zookeeper is suitable for metadata. +1 to Zookeeper.
- Why do we scope the alternatives to KV stores? I think a conventional database like MySQL/Postgres may do the job. It's easy to set up and maintain, plus metadata wouldn't put much pressure on it.

yaooqinn:
thanks @link3280 for the inputs.

> Embedded database like leveldb doesn't provide high availability alone. -1 to leveldb/rocksdb.

For leveldb, I am considering it as an alternative to the current in-memory hash map only, which is ease of use for users and add failure recovery on kyuubi server crashes.

For the rest of your questions, I am also waiting for suitable proposals which do not It doesn't have to be limited to kv stores

turboFei:
## KVStore
```
def get[T](key: String): T
def set(key: String, value: Object): Unit
def  keySet(size: Int): Seq[String]
def count(): Int
def  remove(key: String): Unit
```

<img width="937" alt="image" src="https://user-images.githubusercontent.com/6757692/161211397-3102783e-9e10-4dc4-a85f-15b9e6473426.png">

## Session/OperationManager recovery

```
def init(conf: KyuubiConf): Unit = {
  // …
  store = …
}

store.set(...)
store.get(...)
store.remove(...)

// recover from the state store
def  recover(): Unit = {
   store.keySet(0).foreach { … }
}

```

## Engine connection recovery
[...]

PR #2179, opened 2022-03-19
[KYUUBI #2178] Add abstract session store to store session

zhaomin1423:

Add abstract session store to store session.


cfmcgrady:
Does this is a subtask of https://github.com/apache/incubator-kyuubi/issues/1754?

zhaomin1423:
> Is this a subtask of #1754?

yes

wForget:
The abstraction of SessionStore does not need to store Session objects, but only needs to store sessionHandle and Engine instance.
We can add Cache layer to cache Session objects, MemorySessionStore or others that only store sessionHandle and Engine instance, what do you think?

PR #2181, opened 2022-03-19
[KYUUBI #2180] Add abstract operation store to store operation

zhaomin1423:

Add abstract operation store to store operation.


yanghua:
@zhaomin1423 IMO, it would be better to have a design doc.

turboFei:
> @zhaomin1423 IMO, it would be better to have a design doc.

+1.

BTW, the OperationStore is specific for operation.

Maybe we need a more abstract store interface, like spark-kvstore.

FYI: https://github.com/apache/spark/tree/master/common/kvstore

Issue #2195, opened 2022-03-22
[Improvement] Using while-loop or for-loop instead of map/range to improve performance in RowSet


zhaomin1423:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Using while-loop or for-loop instead of map/range to improve performance in RowSet

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2196, opened 2022-03-22
[KYUUBI #2195] Using while-loop or for-loop instead of map/range to improve performance in RowSet

zhaomin1423:

Using while-loop or for-loop instead of map/range to improve performance in RowSet.


ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #2201, opened 2022-03-23
[Improvement] Show ExecutionId when running status on query engine page


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Now we can display the ExecutionId on the query engine page, but generally the sql has been finished.
We can update it when SQL is running and there is an ExecutionId.

### How should we improve?

Update `ExecutionId` in `SQLOperationListener`.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2202, opened 2022-03-23
[KYUUBI #2201] Show ExecutionId when running status on query engine page

cxzl25:

Now we can display the ExecutionId on the query engine page, but generally the sql has been finished.
We can update it when SQL is running and there is an ExecutionId.

close https://github.com/apache/incubator-kyuubi/issues/2201, https://github.com/apache/incubator-kyuubi/issues/921

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -23,7 +23,6 @@ import scala.collection.JavaConverters._
 
 import org.apache.spark.kyuubi.SQLOperationListener
 import org.apache.spark.sql.Row
-import org.apache.spark.sql.catalyst.QueryPlanningTracker
 import org.apache.spark.sql.types._
 
 import org.apache.kyuubi.{KyuubiSQLException, Logging}
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -159,13 +159,21 @@ class ExecuteStatement(
 
   def setCompiledStateIfNeeded(): Unit = synchronized {
     if (getStatus.state == OperationState.RUNNING) {
-      super.setState(OperationState.COMPILED)
-      if (result != null) {
-        val phase = result.queryExecution.tracker.phases
-        if (phase.contains("parsing") && phase.contains("planning")) {
-          val compiledTime = phase("planning").endTimeMs - phase("parsing").startTimeMs
-          lastAccessTime = lastAccessTime + compiledTime
+      val lastAccessCompiledTime =
+        if (result != null) {
+          val phase = result.queryExecution.tracker.phases
+          if (phase.contains("parsing") && phase.contains("planning")) {
+            val compiledTime = phase("planning").endTimeMs - phase("parsing").startTimeMs
+            lastAccessTime + compiledTime
+          } else {
+            0L
+          }
+        } else {
+          0L
         }
+      super.setState(OperationState.COMPILED)
+      if (lastAccessCompiledTime > 0L) {
+        lastAccessTime = lastAccessCompiledTime
       }
       EventBus.post(
         SparkOperationEvent(this, operationListener.getExecutionId))
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -79,8 +79,6 @@ class ExecuteStatement(
       // TODO: Make it configurable
       spark.sparkContext.addSparkListener(operationListener)
       result = spark.sql(statement)
-      // TODO #921: COMPILED need consider eagerly executed commands
-      setState(OperationState.COMPILED)
       debug(result.queryExecution)
       iter =
         if (incrementalCollect) {
@@ -97,6 +95,9 @@ class ExecuteStatement(
             new ArrayFetchIterator(result.take(resultMaxRows))
           }
         }
+      if (getStatus.state != OperationState.COMPILED) {
+        setCompiledState()
+      }
       setState(OperationState.FINISHED)
     } catch {
       onError(cancel = true)
@@ -157,5 +158,9 @@ class ExecuteStatement(
     super.setState(newState)
     EventBus.post(
       SparkOperationEvent(this, operationListener.getExecutionId))
+  }
+
+  def setCompiledState(): Unit = {
+    setState(OperationState.COMPILED)
   }
 }
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -95,9 +95,6 @@ class ExecuteStatement(
             new ArrayFetchIterator(result.take(resultMaxRows))
           }
         }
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/test/scala/org/apache/kyuubi/events/KyuubiEventLoggingServiceSuite.scala
```diff
@@ -82,7 +82,7 @@ class KyuubiEventLoggingServiceSuite extends WithKyuubiServer with HiveJDBCTestH
       val engineTable = engineStatementEventPath.getParent
       val resultSet2 = statement.executeQuery(s"SELECT * FROM `json`.`${engineTable}`" +
         "where statement = \"" + sql + "\"")
-      val engineStates = Array(INITIALIZED, PENDING, RUNNING, COMPILED, FINISHED)
+      val engineStates = Array(INITIALIZED, PENDING, RUNNING, COMPILED, COMPILED, FINISHED)
```
ulysses-you, line 85: I'm afraid the extra event. There is many redundant event info now ..

cxzl25: There are indeed a lot of event logs. I read the spark engine. There are two events for starting the engine, two events for session opening and closing, and a new event for each statement execution state change, which is about 5 indivual.

My changes here will add an event to each statement.

Because the user's sql may not be executed quickly sometimes, the session page can better track the execution of a specific user's entire session.

I also planned to do one thing, like the spark thrift server, to list a list of jobids that are continuously generated by the statement, so that there are more events.

image

ulysses-you: after some thought, I think we can remove the existed `COMPILE`:
```SCALA
// TODO #921: COMPILED need consider eagerly executed commands
setState(OperationState.COMPILED)
```

then catch the event `SparkListenerSQLExecutionStart` in `SQLOperationListener`. At this time, we can post the `COMPILED` state event which contains the execution id.

This approach also fix the `TODO` about the eagerly executed commands.

cxzl25: However, some DDLs do not generate jobs, so there is no executionId
e.g. use db / drop table.

But this TODO, I guess it is because the command such as insert select or ctas will be executed directly, so TODO wants to set this state in advance.

We can do this if we allow some DDL to have no `COMPILED` status.

cxzl25: Now I think of a way, if there is no `COMPILED` state before setting the `FINISHED` state, then set the `COMPILED` state first.

ulysses-you: > However, some DDLs do not generate jobs, so there is no executionId

It's not right, query must have the executionId whether it generates jobs or not. So I said, we can catch  `SparkListenerSQLExecutionStart` rather than `SparkListenerJobStart`

cxzl25: Sorry, after reading the Spark code, each SQL does have an executionId.
The DDL I watched directly from the kyuubi session page yesterday did not show the executionId.
That would also fix two issues by the way.
1. ddl shows executionId.
2. We can also remove listener when no job is generated to prevent memory leaks.

cxzl25: After offline discussion, using `SparkListenerSQLExecutionStart` to get the `executionId`, it is difficult to map the current statement.
So ddl still can't display the `executionId`.
If there is no `executionId`, `SparkOperation#close` will remove the `SQLOperationListener` at the end, so there is no memory leak.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -154,8 +153,27 @@ class ExecuteStatement(
   }
 
   override def setState(newState: OperationState): Unit = {
+    setStateInt(newState)
+  }
+
+  private def setStateInt(newState: OperationState, startTime: Option[Long] = None): Unit = {
     super.setState(newState)
+    if (startTime.isDefined) {
+      this.startTime = startTime.get
```
cxzl25, line 162: When the `newState` of `AbstractOperation#setState` is `running`, `startTime` will be assigned the millisecond of the current time, so I can only reassign `startTime` here.

cxzl25:
@ulysses-you any suggestion?

ulysses-you:
can we simplify code to:

```scala
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -23,6 +23,7 @@ import scala.collection.JavaConverters._

 import org.apache.spark.kyuubi.SQLOperationListener
 import org.apache.spark.sql.Row
+import org.apache.spark.sql.catalyst.QueryPlanningTracker

 import org.apache.kyuubi.{KyuubiSQLException, Logging}
@@ -79,8 +80,7 @@ class ExecuteStatement(
       // TODO: Make it configurable
       result = spark.sql(statement)
-      // TODO #921: COMPILED need consider eagerly executed commands
-      setState(OperationState.COMPILED)
+      setCompiledState(result.queryExecution.tracker)
       debug(result.queryExecution)
       iter =
         if (incrementalCollect) {
@@ -158,4 +158,17 @@ class ExecuteStatement(
     EventBus.post(
       SparkOperationEvent(this, operationListener.getExecutionId))
   }
+
+  private def setCompiledState(tracker: QueryPlanningTracker): Unit = {
+    super.setState(OperationState.COMPILED)
+
+    val phase = tracker.phases
+    if (phase.contains("parsing") && phase.contains("planning")) {
+      val compiledTime = phase("planning").endTimeMs - phase("parsing").startTimeMs
+      lastAccessTime = lastAccessTime + compiledTime
+    }
+
+    EventBus.post(
+      SparkOperationEvent(this, operationListener.getExecutionId))
+  }
 }
```

cxzl25:
Thanks for your advice. @ulysses-you 

I added the `ExecutionId` as much as possible in the `running` phase on the basis of what you provided.
If it is not needed, I will remove the logic related to `SQLOperationListener`.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -158,4 +156,19 @@ class ExecuteStatement(
     EventBus.post(
       SparkOperationEvent(this, operationListener.getExecutionId))
   }
+
+  def setCompiledStateIfNeeded(): Unit = synchronized {
+    if (getStatus.state == OperationState.RUNNING) {
+      super.setState(OperationState.COMPILED)
```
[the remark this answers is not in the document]
cxzl25: `super.setState` will also update `lastAccessTime` directly, so what is the meaning of updating `lastAccessTime` here?

`lastAccessTime = System.currentTimeMillis()`

```scala
    if (getStatus.state == OperationState.RUNNING) {
      if (result != null) {
        val phase = result.queryExecution.tracker.phases
        if (phase.contains("parsing") && phase.contains("planning")) {
          val compiledTime = phase("planning").endTimeMs - phase("parsing").startTimeMs
          lastAccessTime = lastAccessTime + compiledTime
        }
      }
      super.setState(OperationState.COMPILED)
      EventBus.post(
        SparkOperationEvent(this, operationListener.getExecutionId))
    }
```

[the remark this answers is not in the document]
ulysses-you: ah, good point, we should materialized it.
1. val lastAccessCompiledTime = lastAccessTime + compiledTime
2. super.setState(OperationState.COMPILED)
3. lastAccessTime = lastAccessCompiledTime

ulysses-you (approved):

ulysses-you:
thanks, merging to master!

Issue #2203, opened 2022-03-23
[SUB-TASK][KPIP-2] Support setting Flink configuration by setting kyuubi-default.conf


deadwind4:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

As this comment https://github.com/apache/incubator-kyuubi/issues/2153#issuecomment-1073016029, we set Flink options only use JDBC URL and set flink-conf.yaml. We should support` kyuubi-default.conf`.
All prefix `flink.*` in `kyuubi-default.conf` should be set to Flink configuration.

`kyuubi-default.conf`  example:
```
flink.parallelism.default: 12
```

### How should we improve?

I will filter all options with `flink.*` prefix to set into `flinkConf` in the `FlinkSQLEngine` class

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2206, opened 2022-03-24
[KYUUBI #2203][FLINK] Support flink conf set by kyuubi conf file

deadwind4:

Support setting Flink configuration by setting `kyuubi-default.conf`.

### DIFF
--- a/integration-tests/kyuubi-flink-it/src/test/scala/org/apache/kyuubi/it/flink/FlinkSQLEngineSuite.scala
+++ b/integration-tests/kyuubi-flink-it/src/test/scala/org/apache/kyuubi/it/flink/FlinkSQLEngineSuite.scala
@@ -32,13 +32,17 @@ class FlinkSQLEngineSuite extends WithKyuubiServerAndFlinkMiniCluster with HiveJ
 
   test("set kyuubi conf into flink conf") {
     withJdbcStatement() { statement =>
-      statement.execute("CREATE TABLE ods ( a_name STRING ) WITH ('connector' = 'datagen' )")
-      statement.execute("CREATE TABLE ads ( b_name STRING ) WITH ( 'connector' = 'print' )")
-      val resultSet = statement.executeQuery("EXPLAIN JSON_EXECUTION_PLAN " +
-        "INSERT INTO ads SELECT a_name as b_name FROM ods where ods.a_name='\\\\\\\\'")
-      assert(resultSet.next())
-      val operationPlan = resultSet.getString(1)
-      assert(operationPlan.contains("\"parallelism\" : 6"))
+      val resultSet = statement.executeQuery("SET")
+      // Flink does not support set key without value currently,
+      // thus read all rows to find the desired one
+      var success = false
+      while (resultSet.next() && success == false) {
+        if (resultSet.getString(1) == "parallelism.default" &&
+          resultSet.getString(2) == "6") {
+          success = true
+        }
+      }
+      assert(success)
     }
   }
 }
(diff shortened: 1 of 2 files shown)


pan3793:
Can we have a test for it?

pan3793:
If I understand correctly, the behavior of engine-specific properties defined in `kyuubi-default.conf` is different in Spark and Flink. For Spark, we can define the properties as-is, e.g. define `spark.driver.memory` in `kyuubi-default.conf` has a same effect with defining `spark.driver.memory` in `spark-defaults.conf`, but for Flink, user need to add a prefix `flink` in `kyuubi-defaults.conf`, e.g. define `flink.jobmanager.memory.heap.size` in `kyuubi-default.conf` has a same effect with defining `jobmanager.memory.heap.size` in `flink-conf.yaml`. 

I'm fine with prefix design, but we need to document such behavior clearly otherwise users may confusing when they use both Spark and Flink engines.

jiaoqingbo:
> 


> If I understand correctly, the behavior of engine-specific properties defined in `kyuubi-default.conf` is different in Spark and Flink. For Spark, we can define the properties as-is, e.g. define `spark.driver.memory` in `kyuubi-default.conf` has a same effect with defining `spark.driver.memory` in `spark-defaults.conf`, but for Flink, user need to add a prefix `flink` in `kyuubi-defaults.conf`, e.g. define `flink.jobmanager.memory.heap.size` in `kyuubi-default.conf` has a same effect with defining `jobmanager.memory.heap.size` in `flink-conf.yaml`.
> 
> I'm fine with prefix design, but we need to document such behavior clearly otherwise users may confusing when they use both Spark and Flink engines.

It's better to define a specific prefix for each engine and add it to the DOC, and take into account that we may add new engine support

deadwind4:
> If I understand correctly, the behavior of engine-specific properties defined in `kyuubi-default.conf` is different in Spark and Flink. For Spark, we can define the properties as-is, e.g. define `spark.driver.memory` in `kyuubi-default.conf` has a same effect with defining `spark.driver.memory` in `spark-defaults.conf`, but for Flink, user need to add a prefix `flink` in `kyuubi-defaults.conf`, e.g. define `flink.jobmanager.memory.heap.size` in `kyuubi-default.conf` has a same effect with defining `jobmanager.memory.heap.size` in `flink-conf.yaml`.
> 
> I'm fine with prefix design, but we need to document such behavior clearly otherwise users may confusing when they use both Spark and Flink engines.

I will split a new issue to achieve Flink doc including this below.

deadwind4:
> Can we have a test for it?

@pan3793 I have added an integration test case, thanks for your reminder.

integration-tests/kyuubi-flink-it/src/test/scala/org/apache/kyuubi/it/flink/FlinkSQLEngineSuite.scala
```diff
@@ -0,0 +1,44 @@
...
+  override val conf: KyuubiConf = KyuubiConf()
+    .set(ENGINE_TYPE, "FLINK_SQL")
+    .set(FRONTEND_THRIFT_BINARY_BIND_PORT, 10029)
+    .set("flink.parallelism.default", "6")
+
+  override protected def jdbcUrl: String = getJdbcUrl
+
+  test("set kyuubi conf into flink conf") {
+    withJdbcStatement() { statement =>
+      statement.execute("CREATE TABLE ods ( a_name STRING ) WITH ('connector' = 'datagen' )")
```
pan3793, line 35: Does Flink support use `SET` command to echo the key value? If yes, we can simplify the test case

deadwind4: You are right, I have used the `SET` statement to get echo conf.

pan3793 (approved):

yanghua (approved):
LGTM

pan3793:
Thanks all, merging to master.

PR #2208, opened 2022-03-24
Fixed session close operator log session dir not deleted

winfys:
Operator log Session dir not deleted when the session is closed，
image
As time goes on, more and more directories are created, which is a big risk

1、open session
image
session operator dir
image
2、Closing the session directory is deleted
image

image


SteNicholas:
@winfys, could you please add the unit tests for the changes?

winfys:
@SteNicholas ok,I'm going to add

ulysses-you (approved):
looks reasonable

cfmcgrady (approved):

ulysses-you:
thanks, merging to master

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -89,10 +89,23 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
     if (session == null) {
       throw KyuubiSQLException(s"Invalid $sessionHandle")
     }
+    deleteOperationLogSessionDir(sessionHandle)
```
cxzl25, line 92: When the `closeSession` method is called, the operation is not necessarily close. At this time, there are log files, and the log may also increase.

When `session.close()`, it will close the operation first.

So it is suggested here that `deleteOperationLogSessionDir` should be placed after `session.close()`
Also because `Files.deleteIfExists` can only delete empty folders, consider using `org.apache.kyuubi.Utils#deleteDirectoryRecursively`.

ulysses-you: my bad, previous I only considered the empty session case which a session is created without operators

Issue #1021, opened 2021-09-03
Expire CredentialsRef in a proper time to reduce memory usage


zhouyifan279:
### 1. Describe the feature
This is the a sub-task under umbrella issue #913. 

In sub-task #1007,  <user, CredentialsRef> pairs are cached in a Map. Credentials are periodically renewed  and will never be removed.
If a lot of inactive users' credentials are cached, both memory usage and credentials renewal load may become unacceptable.
So we should implement a logic to expire cached <user, CredentialsRef> pairs to erase this risk.

lightning-L:
Hi, I'd like to try to implement this.

ulysses-you:
@lightning-L assigned, thank you

PR #2210, opened 2022-03-24
[KYUUBI #1021] Expire CredentialsRef in a proper time to reduce memor…

lightning-L:
…y usage


ulysses-you:
cc @zhouyifan279 @turboFei

ulysses-you (approved):

ulysses-you:
thanks, merging to master

kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HadoopCredentialsManager.scala
```diff
@@ -239,7 +246,9 @@ class HadoopCredentialsManager private (name: String) extends AbstractService(na
         try {
           promise.trySuccess(updateCredentials(userRef))
 
-          scheduleRenewal(userRef, renewalInterval)
+          if (userCredentialsRefMap.containsKey(userRef.getAppUser)) {
```
zhouyifan279, line 249: Sorry for late reply. But considering the following scenario, I think we should also check whether `userRef` in `userCredentialsRefMap` has been replaced by a new one .

1. `userRef` was waiting for next renewal. 
2. `userRef` expired and was removed from `userCredentialsRefMap`
3.  A new `userRef` was created because of user visiting.
4. Renewal of old `userRef` started again.

If we do not check, both old and new `userRef` will renew periodically.

lightning-L: @zhouyifan279
Thanks. I will do a follow-up commit.

Issue #2207, opened 2022-03-24
[FEATURE] Support newly added spark data types: DayTimeIntervalType/YearMonthIntervalType/TimestampNTZType


Nick-0723:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Spark 3.2.0 introduces new DataTypes, `DayTimeIntervalType` and `YearMonthIntervalType`,  and Spark 3.3.0 introduces `TimestampNTZType` we need to support these feature.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2212, opened 2022-03-24
[KYUUBI #2207] Support newly added spark data types: DayTimeIntervalType/YearMonthIntervalType

Nick-0723:
#2207
Support newly added datatypes
spark3.2 : DayTimeIntervalType/YearMonthIntervalType 


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -39,10 +39,13 @@ object SchemaHelper {
     case _: DecimalType => TTypeId.DECIMAL_TYPE
     case DateType => TTypeId.DATE_TYPE
     case TimestampType => TTypeId.TIMESTAMP_TYPE
+    case tntz if tntz.simpleString.equals("timestamp_ntz") => TTypeId.TIMESTAMP_TYPE
```
yaooqinn, line 42: Do we need also handle this in `toJavaSQLType`

yaooqinn: getDecimalDigits/getDecimalDigits etc

yaooqinn:
Hi @Nick-0723 I check PR and it looks OK to me in general.  For TimestampNTZType, I guess there is still something incomplete, so maybe we shall fix it in another PR to minimize the scope of this one? WDYT？

Nick-0723:
> Hi @Nick-0723 I check PR and it looks OK to me in general. For TimestampNTZType, I guess there is still something incomplete, so maybe we shall fix it in another PR to minimize the scope of this one? WDYT？

I agree ，so do we need revert the TimestampNTZType's code for this PR ?

yaooqinn:
Yes, better revert this part for now

ulysses-you (approved):

yaooqinn (approved):
LGTM

yaooqinn:
thanks, merged to master

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
Nick-0723: Is TimestampNTZType's hive thrift type TTypeId.TIMESTAMP_TYPE and is jdbc type TIMESTAMP ？

yaooqinn: Ideally，you are right. but there are some historical reasons. at the very beginning，hive does not support timestamp with local timezone but spark support，and they both have only one and be treated as same. now，they both have 2 types supported and the problem caused by old mapping stands out. I guess we shall learn how spark deal with this mess.

Nick-0723: The implication is that we need to change the previous Timestamp type to TIMESTAMPLOCALTZ_TYPE ？
```scala
case TimestampType => TTypeId.TIMESTAMPLOCALTZ_TYPE
```

Nick-0723: > The implication is that we need to change the previous Timestamp type to TIMESTAMPLOCALTZ_TYPE ？
> 
> ```scala
> case TimestampType => TTypeId.TIMESTAMPLOCALTZ_TYPE
> ```

But it seems that our current hive version (2.3.9) does not support it yet

PR #2225, opened 2022-03-27
Support to set result max rows for spark engine

turboFei:

Support to set result max rows for spark engine.
And it will only take affect if incremental mode is not enabled.


pan3793:
Can `ForcedMaxOutputRowsRule` be replaced by this one? cc @cfmcgrady @ulysses-you @wForget

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
```diff
@@ -64,7 +65,15 @@ class SparkSQLOperationManager private (name: String) extends OperationManager(n
             case NONE =>
               val incrementalCollect = spark.conf.getOption(OPERATION_INCREMENTAL_COLLECT.key)
                 .map(_.toBoolean).getOrElse(operationIncrementalCollectDefault)
-              new ExecuteStatement(session, statement, runAsync, queryTimeout, incrementalCollect)
+              val resultMaxRows = spark.conf.getOption(ENGINE_SPARK_MAX_ROWS.key)
```
yaooqinn, line 68: move to ExecuteStatement

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -626,6 +626,14 @@ object KyuubiConf {
       .intConf
       .createWithDefault(1000000)
 
+  val ENGINE_SPARK_MAX_ROWS: ConfigEntry[Int] =
+    buildConf("session.engine.spark.max.rows")
```
yaooqinn, line 630: kyuubi.operation.max.result.rows?

turboFei: I saw that there is a parameter named `kyuubi.session.engine.flink.max.rows`.

turboFei: How about fall back to kyuubi.operation.max.result.rows if kyuubi.session.engine.*.max.rows is not specified?

ulysses-you:
> Can ForcedMaxOutputRowsRule be replaced by this one

It's different in physical side. `dataset.take` just like the incremental mode which run job partition by partition if previous partitions data has not satisfied the limit number. `dataset.limit.collect` will run all partitions to do local limit then do global limit in one partition. So in general `dataset.limit.collect` is more effective.

And this PR is useful. Since the limitation is controlled by client side using Incremental mode, but we can control it at engine side using `dataset.take`.

pan3793:
> `dataset.take` just like the incremental mode which run job partition by partition if previous partitions data has not satisfied the limit number.

I would argue that `dataset.take` is same as `dataset.limit.collect`

```
def take(n: Int): Array[T] = head(n)
def head(n: Int): Array[T] = withAction("head", limit(n).queryExecution)(collectFromPlan)
def collect(): Array[T] = withAction("collect", queryExecution)(collectFromPlan)
```

pan3793:
IIRC, the difference you described above is for increment mode.

w/ incremental mode, `CollectLimitExec#executeToIterator` will be called, which calls `getByteArrayRdd` and triggers to compute all partitions of previous stages firstly.

w/o incremental mode, `CollectLimitExec#executeCollect` will be called, which submit partitions increase exponentially utils satisfy the numbers.

ulysses-you:
yea, if so I agree we can use this feature to replace the rule in extension.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: it is to align with `kyuubi.session.engine.flink.max.rows`

Nick-0723:
Does the user know that we truncate the data?

turboFei:
> Does the user know that we truncate the data?

By default, it is not truncated.
It is determined by the service provider.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
pan3793: `kyuubi.operation.max.result.rows` makes sense to me, and we can deprecate `kyuubi.session.engine.flink.max.rows`

turboFei: we can deprecate kyuubi.session.engine.flink.max.rows in another pr.

turboFei: resolved

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
```diff
@@ -136,4 +136,21 @@ class KyuubiOperationPerUserSuite extends WithKyuubiServer with SparkQueryTests
 
     assert(r1 !== r2)
   }
+
+  test("test engine spark result max rows") {
+    withSessionConf()(Map.empty)(Map(KyuubiConf.OPERATION_RESULT_MAX_ROWS.key -> "1")) {
+      withJdbcStatement() { statement =>
```
cfmcgrady, line 142: nit: `withJdbcStatement("va")`

turboFei: thanks

cfmcgrady:
> yea, if so I agree we can use this feature to replace the rule in extension.

+1

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #1498, opened 2021-12-06
[FEATURE][SUBTASK][#1489] Support operation log for ExecuteScala


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Deliver the operation log at engine side to clients

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #2234, opened 2022-03-27
[KYUUBI #1498] Support operation log for ExecuteScala

cxzl25:
close https://github.com/apache/incubator-kyuubi/issues/1498

Pass scala repl output to operation log.


cxzl25:
## Test script
```sql
set kyuubi.operation.language=scala;
val df = spark.range(0, 10, 2, 1).toDF;
df.show();
result.set(df);
```

## Current
image

image

image


## Fix

image

image

image

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteScala.scala
```diff
@@ -71,9 +75,12 @@ class ExecuteScala(
             if (result != null) {
               new ArrayFetchIterator[Row](result.collect())
             } else {
-              // TODO (#1498): Maybe we shall pass the output through operation log
-              // but some clients may not support operation log
-              new ArrayFetchIterator[Row](Array(Row(repl.getOutput)))
+              if (scalaReplToOperationLog) {
+                info("\n" + repl.getOutput)
+                new ArrayFetchIterator[Row](Array.empty[Row])
```
ulysses-you, line 80: shall we also pass repl.output to client ?

yaooqinn:
Maybe we do not need this config and print it to the operation log all the time, the output seems useless in result

cxzl25:
> Maybe we do not need this config and print it to the operation log all the time, the output seems useless in result

I read the previous comment considering that some low-level clients do not support operation log, and query log is only supported in Hive 0.14 because of [HIVE-4629](https://issues.apache.org/jira/browse/HIVE-4629).

Or remove this configuration, as suggested by ulysses-you, output a copy of both operation log and output?

ulysses-you:
if there is no affect for lower version client, I think we can just pass output to operation log and client

yaooqinn (approved):

cxzl25:
I added the scala repl output prefix to make it easier to find.

##  Fix

image

image

image

ulysses-you (approved):

ulysses-you:
thanks, merging to master

PR #2246, opened 2022-03-29
[BUILD] Pick Commons-Logging dependence out of Hudi-Common

zwangsheng:

```
        <dependency>
            <groupId>org.apache.hadoop</groupId>
            <artifactId>hadoop-client-minicluster</artifactId>
            <scope>test</scope>
        </dependency>
```
`hadoop-client-minicluster` need `commons-logging` dependence. Now it works because `hudi-spark-common` include this dependency.

We extract it to better manage dependencies.

### DIFF
--- a/kyuubi-server/pom.xml
+++ b/kyuubi-server/pom.xml
@@ -362,6 +362,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.hudi</groupId>
             <artifactId>hudi-spark_${scala.binary.version}</artifactId>
             <scope>test</scope>
--- a/kyuubi-server/pom.xml
+++ b/kyuubi-server/pom.xml
@@ -356,12 +356,6 @@
         </dependency>
 
         <dependency>
-            <groupId>commons-logging</groupId>
-            <artifactId>commons-logging</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>jcl-over-slf4j</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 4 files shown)


pan3793:
`org.slf4j:jcl-over-slf4j` is a drop-in replacement of `commons-logging`, please use the former.

ulysses-you:
+1 with @pan3793

zwangsheng:
@pan3793 I check https://www.slf4j.org/faq.html#excludingJCL found that provides two methods to exclude commons - logging, I prefer to the second, what do you think?

pan3793:
> @pan3793 I check https://www.slf4j.org/faq.html#excludingJCL found that provides two methods to exclude commons - logging, I prefer to the second, what do you think?

Your change does not implement option 2, it still in test scope.
Personally I prefer option 1, because we don't need to declare versions for un-used deps.

zwangsheng:
@pan3793 I do as you said and found that this class  was missing org.apache.commons.logging.impl.Log4JLogger

pan3793:
I see what happens, it access the `Log4JLogger` directly, let's remove the exclusion from the `hadoop-client-runtime` in root pom

kyuubi-server/pom.xml
```diff
@@ -349,6 +349,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
```
pan3793, line 354: Then we don't need it.

pom.xml
```diff
@@ -180,7 +180,8 @@
         <maven.plugin.scala.version>4.3.0</maven.plugin.scala.version>
         <maven.plugin.surefire.version>2.22.0</maven.plugin.surefire.version>
         <maven.plugin.scalatest.version>2.0.2</maven.plugin.scalatest.version>
-        <maven.plugin.scalatest.exclude.tags>org.apache.kyuubi.tags.ExtendedSQLTest,org.apache.kyuubi.tags.HudiTest</maven.plugin.scalatest.exclude.tags>
+        <maven.plugin.scalatest.exclude.tags>org.apache.kyuubi.tags.ExtendedSQLTest,org.apache.kyuubi.tags.HudiTest
+        </maven.plugin.scalatest.exclude.tags>
```
pan3793, line 184: Unnecessary chagnes?

zwangsheng: `./build/dependency.sh --replace` changed those.

pan3793 (approved):

pan3793:
Thanks, merging.

PR #2251, opened 2022-03-30
[KYUUBI #2207]  Fix DayTimeIntervalType/YearMonthIntervalType Column Size

Nick-0723:

Data type DayTimeIntervalType and YearMonthIntervalType have default column size ,but now is null.


externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
```diff
@@ -57,7 +57,7 @@ class SparkOperationSuite extends WithSparkSQLEngine with HiveMetadataTests with
 
   test("get columns operation") {
     val tableName = "spark_get_col_operation"
-    val schema = new StructType()
+    var schema = new StructType()
```
turboFei, line 60: nit: unneeded change

turboFei (approved):
LGTM

yaooqinn:
thanks, merged to master

Issue #2027, opened 2022-03-04
[Subtask] Hive Backend Engine - Support Multiple Hive releases


yaooqinn:
default 2.3, up to 3.1

pan3793:
It has been reported it engine does not work out-of-box w/ HDP Hive 3.1.3.1.0.6-1

pan3793:
Postpone to 1.8, because this feature is not under rapid development, and it's not supposed to be accomplished in a short time.

pan3793:
Apache Kyuubi 1.9.0 has verified on CDH Hive 2.1.1-cdh6.y.z, Apache Hive 2.3.9, Apache Hive 3.1.3

PR #2254, opened 2022-03-30
[KYUUBI #2027] [Subtask] Hive Backend Engine - Support Multiple Hive releases

zhaomin1423:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1897,6 +1897,13 @@
         </profile>
 
         <profile>
+            <id>hive-3.1</id>
+            <properties>
+                <hive.version>3.1.2</hive.version>
+            </properties>
+        </profile>
+
+        <profile>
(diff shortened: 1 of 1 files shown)


yaooqinn:
I made a follow up for hive proc builder at #2290 

Support multiple hive releases means `complile once, ran on different hive versions` not only download a distribution, you shall actually make the hive engine run all the tests through the version of hive you add

zhaomin1423:
> I made a follow up for hive proc builder at #2290
> 
> Support multiple hive releases means `complile once, ran on different hive versions` not only download a distribution, you shall actually make the hive engine run all the tests through the version of hive you add

Thanks, I will update it later.

pan3793:
Close as we already support both Hive 2.3 and 3.1 now

PR #2271, opened 2022-04-02
[KYUUBI #2270] Implement tag based `SessionConfAdvisor`

### LINKED ISSUE #2270: [FEATURE] Implement a tag based `SessionConfAdvisor`
### Code of Conduct

### Search before asking

### Describe the feature

As discussed in https://github.com/apache/incubator-kyuubi/issues/966#issuecomment-901945984, I want to implement a tag based `SessionConfAdvisor` that gets some session configuration for given tags.

Tag level:
+ SYSTEM
+ SERVER
+ USER
+ TAG

Priority of configuration: TAG > USER > SERVER > SYSTEM

### Motivation

I want to optimize the configuration based on some tags specified by the user.

### Describe the solution

We can implement `FileTagConfAdvisor` first and store the tag configuration in a file.

### Additional context

_No response_

wForget:

close #2270 


wForget:
@yaooqinn @ulysses-you Could you please help me review it and give some suggestions?

yaooqinn:
1. kyuubi-server-plugin is trying to make itself dependency free, especially `scala` and `kyuubi-common`.
2. I don't quite understand the meaning of tag-level?

ulysses-you:
can you provide some some details about your idea ? and it would be good to give some use case.

wForget:
> 1. kyuubi-server-plugin is trying to make itself dependency free, especially `scala` and `kyuubi-common`.

I will change to java implementation and remove scala and kyuubi-common dependencies

> 3. why do we need a new config file called kyuubi-tags.conf? it seems that kyuubi-defaults.conf can do the same thing

Because I want to read it periodically, so that we don't have to restart the service after modifying it.

> 2. I don't quite understand the meaning of tag-level?

The tag-level is used to distinguish tags of different types.

1. SYSTEM:

    The SYSTEM-level tag is used for global configuration. (This is the same as kyuubi-conf, I can remove it.)

2. SERVER: 

    The SERVER-level tag, which uses the server name as a tag, is used for different server configurations.
    We can define different server names for different clusters, so we can use the server tag to get the configuration of different clusters.

3. USER:

    The USER-level tag, which uses the user name as a tag, is used for different user configurations.
    We can use user tag to get user-specific configuration.

4. TAG:

    The TAG-level tag, used for custom tag types. We can customize some tags and give default configuration.
    For example, we can set the user engine-share-level for `adhoc` tag and the connection engine-share-level for `batch` tag.

wForget:
> can you provide some some details about your idea ? and it would be good to give some use case.


There are some use cases in https://github.com/apache/incubator-kyuubi/pull/2271#issuecomment-1094484686, are these reasonable?

Narcasserun:
@wForget  I have a question, can this pr be able to submit ad-hoc and batch tasks at the same time? Whether to start 2 kyuubi server processes or 1 process

Narcasserun:
@wForget  I have a question, can this pr be able to submit ad-hoc and batch tasks at the same time? Whether to start 2 kyuubi server processes or 1 process

wForget:
> @wForget I have a question, can this pr be able to submit ad-hoc and batch tasks at the same time? Whether to start 2 kyuubi server processes or 1 process

@Narcasserun  This feature is only for injecting some configuration into the session through the configured tags, and does not implement the function of request routing, we can also consider it in the future.

wForget:
The current implementation is outdated, I will reimplement it, closing it first.

Issue #2084, opened 2022-03-10
[Improvement] Support arbitrary parameters for KyuubiConf or kyuubi prefixed only


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

  private def loadFromMap(props: Map[String, String] = Utils.getSystemProperties): KyuubiConf = {
    for ((key, value) <- props if key.startsWith("kyuubi.") || key.startsWith("spark.") ||
        // for user specific defaults
        key.startsWith("___")) {
      set(key, value)
    }
    this
  }

In KyuubiConf, we now load kyuubi.* and spark.* from the kyuubi-defaults.conf config file or sys.props, but we now support other engine types, we shall support all engine conf or remove spark.* here

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

SteNicholas:
@jiaoqingbo, do you have interest to take this ticket? Sorry for previous work for the issue assigned to you.

jiaoqingbo:
i would like to do this，thx

PR #2283, opened 2022-04-06
[KYUUBI #2084][FOLLOWUP] Support arbitrary parameters for KyuubiConf

yaooqinn:

In this PR, we narrow the scope of 8f15622d9beef8d3b15602f3180a21ec86c75623 to support arbitrary parameters which configured in kyuubi-defaults.conf only.

We shall avoid propagating all `sys.props` of Kyuubi server to engine side, which is a dangerous behavior

### DIFF
--- a/docs/deployment/settings.md
+++ b/docs/deployment/settings.md
@@ -313,7 +313,6 @@ Key | Default | Meaning | Type | Since
 
 Key | Default | Meaning | Type | Since
 --- | --- | --- | --- | ---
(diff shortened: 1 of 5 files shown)


yaooqinn:
cc @jiaoqingbo, I modify your original PR a bit to only support arbitrary configs in `kyuubi-defaults.conf`, then we don't need to handle hundreds of jvm defaults, WDYT?

also cc @SteNicholas @ulysses-you @pan3793 @turboFei

ulysses-you:
the idea looks reasonable

yaooqinn:
thanks, merged to master

jiaoqingbo:
> cc @jiaoqingbo, I modify your original PR a bit to only support arbitrary configs in `kyuubi-defaults.conf`, then we don't need to handle hundreds of jvm defaults, WDYT?
> 
> also cc @SteNicholas @ulysses-you @pan3793 @turboFei

Look good to me ，sorry for the late reply

PR #2296, opened 2022-04-08
Fix operation log file handler leak

ulysses-you:

If a session is empty which means that it does not contain any operation, then the OperationLog will never be closed.

This bug will happen if we enable `LaunchEngine`

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
@@ -83,10 +83,10 @@ object OperationLog extends Logging {
 
 class OperationLog(path: Path) {
 
-  private val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
-  private val reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)
+  private lazy val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
+  private lazy val reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)
 
-  private val extraReaders: ListBuffer[BufferedReader] = ListBuffer()
+  private lazy val extraReaders: ListBuffer[BufferedReader] = ListBuffer()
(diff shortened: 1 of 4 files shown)


ulysses-you:
cc @turboFei @pan3793 @yaooqinn

pan3793 (approved):

turboFei (approved):
Thanks

turboFei:
thanks, merged to master

ulysses-you:
also backport into branch-1.5

ulysses-you:
@iodone thank you for the info. This pr can also fix your case right ?

iodone:
> @iodone thank you for the info. This pr can also fix your case right ?

Yes, this pr solves the problem

ulysses-you:
thank you for the confirming. FYI, this patch will land into comming 1.5.1

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -83,10 +83,10 @@ object OperationLog extends Logging {
 
 class OperationLog(path: Path) {
 
-  private val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
-  private val reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)
+  private lazy val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
```
pan3793, line 86: The lazy creating saves open files resource, by may cause NPE if we set a high threshold for loggers.
Simply reverting the lazy change should resolve the issue, right?

Another solution I think we can introduce a flag `initialized` to indicate where we start writing logs to the file, `initialized` starts w/ false and should be updated to true after writing the first line to the log file. The reader should get an empty array instead of an exception if `initialized` is false.

thought? @ulysses-you @Yikf

ulysses-you: +1 for the initialized flag

Issue #2021, opened 2022-03-04
[Subtask] Command OptionParser for launching Hive Backend Engine


yaooqinn:

zhaomin1423:
I am willing to do it.

PR #2340, opened 2022-04-12
[WIP][KYUUBI #2021] Command OptionParser for launching Hive Backend Engine

zhaomin1423:


zhaomin1423:
Could you help me review the underway work？If it is feasible, I will continue to do it. @yaooqinn

yaooqinn:
> these args are all in the strict form of --conf key=val

currently, we pass them in procbuilder as java options in -Dkey=val form

zhaomin1423:
> > these args are all in the strict form of --conf key=val
> 
> currently, we pass them in procbuilder as java options in -Dkey=val form

Thanks, I see.

PR #2356, opened 2022-04-13
[KYUUBI #2021] Command OptionParser for launching Hive Backend Engine

zhaomin1423:


yaooqinn:
how about simply

```scala
def fromCommandLineArgs(args: Seq[String], conf: KyuubiConf): Unit = {
    for (i <- args.indices by 2) {
      require(args(i) == "--conf", s"Unrecognized main arguments prefix ${args(i)}")
      require(i + 1 <= args.size , "Command arguments ended unexpectedly")

      args(i + 1).split("=", 2) match {
        case seq if seq.size == 2 => conf.set(seq.head, seq.last)
        case _ => throw new IllegalArgumentException(s" $args(i + 1)")
      }
    }
  }
```

zhaomin1423:
> how about simply
> 
> ```scala
> def fromCommandLineArgs(args: Seq[String], conf: KyuubiConf): Unit = {
>     for (i <- args.indices by 2) {
>       require(args(i) == "--conf", s"Unrecognized main arguments prefix ${args(i)}")
>       require(i + 1 <= args.size , "Command arguments ended unexpectedly")
> 
>       args(i + 1).split("=", 2) match {
>         case seq if seq.size == 2 => conf.set(seq.head, seq.last)
>         case _ => throw new IllegalArgumentException(s" $args(i + 1)")
>       }
>     }
>   }
> ```

but, as if it's inconvenient if need some other arguments.

yaooqinn:
> > how about simply
> > ```scala
> > def fromCommandLineArgs(args: Seq[String], conf: KyuubiConf): Unit = {
> >     for (i <- args.indices by 2) {
> >       require(args(i) == "--conf", s"Unrecognized main arguments prefix ${args(i)}")
> >       require(i + 1 <= args.size , "Command arguments ended unexpectedly")
> > 
> >       args(i + 1).split("=", 2) match {
> >         case seq if seq.size == 2 => conf.set(seq.head, seq.last)
> >         case _ => throw new IllegalArgumentException(s" $args(i + 1)")
> >       }
> >     }
> >   }
> > ```
> 
> but, as if it's inconvenient if need some other arguments.

any cases

zhaomin1423:
> > > how about simply
> > > ```scala
> > > def fromCommandLineArgs(args: Seq[String], conf: KyuubiConf): Unit = {
> > >     for (i <- args.indices by 2) {
> > >       require(args(i) == "--conf", s"Unrecognized main arguments prefix ${args(i)}")
> > >       require(i + 1 <= args.size , "Command arguments ended unexpectedly")
> > > 
> > >       args(i + 1).split("=", 2) match {
> > >         case seq if seq.size == 2 => conf.set(seq.head, seq.last)
> > >         case _ => throw new IllegalArgumentException(s" $args(i + 1)")
> > >       }
> > >     }
> > >   }
> > > ```
> > 
> > 
> > but, as if it's inconvenient if need some other arguments.
> 
> any cases

None, all can be replaced by conf.

kyuubi-common/src/main/scala/org/apache/kyuubi/util/ArgsParser.scala
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.kyuubi.util
+
+import org.apache.kyuubi.config.KyuubiConf
+
+object ArgsParser {
+
+  def fromCommandLineArgs(args: Array[String], conf: KyuubiConf): Unit = {
+    for (i <- args.indices by 2) {
+      require(args(i) == "--conf", s"Unrecognized main arguments prefix ${args(i)}")
+      require(i + 1 <= args.length, "Command arguments ended unexpectedly")
```
yaooqinn, line 26: maybe we can check this once

kyuubi-common/src/main/scala/org/apache/kyuubi/util/ArgsParser.scala
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kyuubi.util
+
+import org.apache.kyuubi.config.KyuubiConf
+
+object ArgsParser {
+
+  def fromCommandLineArgs(args: Array[String], conf: KyuubiConf): Unit = {
```
yaooqinn, line 23: put it in Utils or rename the class
 to Command Line Utils?

kyuubi-common/src/main/scala/org/apache/kyuubi/util/ArgsParser.scala
zhaomin1423: done

kyuubi-common/src/main/scala/org/apache/kyuubi/util/ArgsParser.scala
zhaomin1423: move it to Utils

kyuubi-common/src/main/scala/org/apache/kyuubi/util/ArgsParser.scala
yaooqinn: No in this way, I mean checking out of the loop

zhaomin1423: done

yaooqinn:
thanks, merged to master

PR #2364, opened 2022-04-14
[KYUUBI #2301] Limit the maximum number of concurrent connections per user and ipaddress

### LINKED ISSUE #2301: [FEATURE] Limit the maximum number of concurrent connections per user
### Code of Conduct

### Search before asking

### Describe the feature

As discussed in https://github.com/apache/incubator-kyuubi/discussions/2288#discussioncomment-2521003, we need a configuration like `hive.server2.limit.connections.per.user` to limit the maximum number of concurrent connections per user.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

wForget:

close #2301


kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -69,6 +69,12 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
   private val timeoutChecker =
     ThreadUtils.newDaemonSingleThreadScheduledExecutor(s"$name-timeout-checker")
 
+  @volatile private var _limiter: Option[SessionLimiter] = None
+
+  protected def setSessionLimiter(limiter: SessionLimiter): Unit = {
```
yaooqinn, line 74: shall we add this only in the server？

wForget: > shall we add this only in the server？

I'm also not sure if we need to add it in Engine too? Since the engine can be shared maybe we can also add limiter to the engine. What do you think?

yaooqinn: Yes, maybe. It won't be late when we have plans on such a thing for engines. So let's focus on the PR's scope itself

wForget: I just set the limiter in KyuubiSessionManager, so do I need to make some changes here?

yaooqinn: move all the logic from common to server would be more clear

wForget: > move all the logic from common to server would be more clear

OK, thanks. I'll do that.

wForget: @yaooqinn I need to decrement limit after close session, in KyuubiSessionManager's closeSession method we can't seem to implement it well. Because the session will be removed from handleToSession after closeSession.

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -69,6 +69,12 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
   private val timeoutChecker =
     ThreadUtils.newDaemonSingleThreadScheduledExecutor(s"$name-timeout-checker")
 
+  @volatile private var _limiter: Option[SessionLimiter] = None
```
ulysses-you, line 72: we do not expose this value, so `limiter` is enough

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
wForget: > @yaooqinn I need to decrement limit after close session, in KyuubiSessionManager's closeSession method we can't seem to implement it well. Because the session will be removed from handleToSession after closeSession.

Sorry, please ignore this comment, I have moved it.

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
wForget: > we do not expose this value, so `limiter` is enough

thanks, i have fixed it

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -0,0 +1,108 @@
...
+  private val _counters: java.util.Map[String, AtomicInteger] =
+    new ConcurrentHashMap[String, AtomicInteger]()
+
+  private[session] def counters(): java.util.Map[String, AtomicInteger] = _counters
+
+  override def increment(userIpAddress: UserIpAddress): Unit = {
+    val user = userIpAddress.user
+    val ipAddress = userIpAddress.ipAddress
+    // increment userIpAddress count
+    if (ipAddressLimit > 0 && StringUtils.isNotBlank(user) && StringUtils.isNotBlank(ipAddress)) {
```
ulysses-you, line 46: ipAddressLimit -> userIpAddressLimit

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -0,0 +1,108 @@
...
+      incrLimitCount(
+        s"$user:$ipAddress",
+        userIpAddressLimit,
+        s"Connection limit per user:ipaddress reached" +
```
ulysses-you, line 50: remove `s`

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -0,0 +1,108 @@
...
+    // decrement userIpAddress count
+    if (ipAddressLimit > 0 && StringUtils.isNotBlank(user) && StringUtils.isNotBlank(ipAddress)) {
+      decrLimitCount(s"$user:$ipAddress")
+    }
+  }
+
+  private def incrLimitCount(key: String, limit: Int, errorMsg: String): Unit = {
+    val count = _counters.computeIfAbsent(key, _ => new AtomicInteger())
+    if (count.incrementAndGet() > limit) {
+      count.decrementAndGet()
```
ulysses-you, line 89: shall we syncronized the whole method ? we should make sure the `increment and decrement` atomic

wForget: > shall we syncronized the whole method ? we should make sure the `increment and decrement` atomic

I don't think it's necessary, we just need to make sure that increment and comparison operations are atomic.

yaooqinn:
A general question，does it possible mix with metrics

wForget:
> A general question，does it possible mix with metrics

Sounds good, do we need to ensure atomicity of increment and comparison operations on metric counters?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -43,10 +43,13 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
   // this lazy is must be specified since the conf is null when the class initialization
   lazy val sessionConfAdvisor: SessionConfAdvisor = PluginLoader.loadSessionConfAdvisor(conf)
 
+  @volatile private var limiter: Option[SessionLimiter] = None
```
yaooqinn, line 46: volatile necessary？

yaooqinn (approved):

ulysses-you:
@wForget please resolve the conflicts, thank you

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
wForget: It is unnecessary, I will remove it

yaooqinn:
thanks, merged to master

Issue #1972, opened 2022-02-24
Flaky test: SchedulerPoolSuite: Scheduler pool


yaooqinn:
```
- Scheduler pool *** FAILED ***
  The code passed to eventually never returned normally. Attempted 558 times over 10.005535925 seconds. Last failure message: 1003 was not less than 1000. (SchedulerPoolSuite.scala:99)

```
cc @ulysses-you

wForget:
Can we fix it by increasing the timeout?

ulysses-you:
@wForget I think we can do this

yaooqinn:
- Scheduler pool *** FAILED ***
  The code passed to eventually never returned normally. Attempted 1208 times over 20.014311831 seconds. Last failure message: 1558 was not less than 1000. (SchedulerPoolSuite.scala:99)

Still seeing it

PR #2383, opened 2022-04-15
[KYUUBI #1972][FOLLOWUP] Fix flaky test: SchedulerPoolSuite: Scheduler pool

wForget:

Follow-up to #1972

Make the following changes:
1. change `job2FinishTime - job1FinishTime >= 1000` to `Math.abs(job2FinishTime - job1FinishTime) >= 1000`
2. eventually is only used to wait for all jobs to complete so we can identify the reason for the failure


wForget:
Invalid fix

Issue #2323, opened 2022-04-11
Separate events to a submodule - `kyuubi-event` like `kyuubi-metrics`


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Separate events to a submodule - `kyuubi-event` like `kyuubi-metrics` 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

zhaomin1423:
I am willing to do it.

zhaomin1423:
Event moudule need use common module, but Session in common module also use KyuubiEnvet, there are circular dependencies. Should we move the method to subclass of Session, or separate session to a module?

yaooqinn:
> Should we move the method to subclass of Session

+1

PR #2392, opened 2022-04-16
[KYUUBI #2323] Separate events to a submodule - kyuubi-event

zhaomin1423:

Separate events to a submodule - kyuubi-event


externals/kyuubi-hive-sql-engine/pom.xml
```diff
@@ -46,6 +46,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
```
yaooqinn, line 49: shade

yaooqinn:
lets also add a log4j file in test resource

externals/kyuubi-hive-sql-engine/pom.xml
zhaomin1423: done

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
```diff
@@ -0,0 +1,39 @@
...
+    password: String,
+    ipAddress: String,
+    conf: Map[String, String],
+    sessionManager: KyuubiSessionManager)
+  extends AbstractSession(protocol, user, password, ipAddress, conf, sessionManager) {
+
+  protected val sessionEvent = KyuubiSessionEvent(this)
+  EventBus.post(sessionEvent)
+
+  def getSessionEvent: Option[KyuubiEvent] = {
```
ulysses-you, line 35: shall we just return `Option[KyuubiSessionEvent]` ?

pom.xml
```diff
@@ -84,6 +84,7 @@
         <module>kyuubi-metrics</module>
         <module>kyuubi-server</module>
         <module>kyuubi-zookeeper</module>
+        <module>kyuubi-events</module>
```
ulysses-you, line 87: nit: name ordering

kyuubi-events/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>kyuubi-events_2.12</artifactId>
+    <name>Kyuubi Project Events</name>
+    <packaging>jar</packaging>
+    <url>https://kyuubi.apache.org/</url>
+
+    <dependencies>
+        <dependency>
```
yaooqinn, line 35: we can remove json related deps from common and add them here

zhaomin1423: json also is used in other class, for example, org.apache.kyuubi.service.authentication.KyuubiInternalAccessIdentifier in common module.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
zhaomin1423: do we define only an empty getSession method in KyuubiSession class?

kyuubi-events/pom.xml
yaooqinn: OK, let's leave it then. the original purpose of separating these two is to remove json libs from common.
Let's leave it in this PR, and see if we can remove it from KyuubiInternalAccessIdentifier

pom.xml
zhaomin1423: done

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
zhaomin1423: thanks, done

zhaomin1423:
> lets also add a log4j file in test resource

done

yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #2393, opened 2022-04-16
[KYUUBI #2391] Fix privileges builder return wrong result when there is no project but has filter/join

### LINKED ISSUE #2391: [Bug] Kyuubi Spark AuthZ Module PrivilegesBuilder Build Wrong PrivilegeObjets When Query Without Project But With Filter/Join
### Code of Conduct

### Search before asking

### Describe the bug

Suppose we have two table `tab1 (key int, value string)`, `tab2 (key int, value string)`. When I execute the query `select key, value from tab1 where key < 1`, the logicalplan of this query that used by PrivilegesBuilder would be like:
```shell
Filter (isnotnull(key#0) AND (key#0 < 1))
+- Relation inmemoryprivilegebuildersuite.inmemoryprivilegebuildersuite[key#0,value#1] parquet
```
There is no Projecttion operator, so by the logical in `org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder#buildQuery()`, only the `key` col will be in the returned privilegeObject, but infact both `key` and `value` cols  should be in the  returned privilegeObject.

The same problem will occur in join query without a project operator, here I have a query: `select t1.key, t1.value, t2.key, t2.value from tab1 t1 join tab2 t2 on t1.key=t2.key`, because of the logicalplan used by  PrivilegesBuilder is not contain a project operator, so only `key` col will be in the returened privilegeObject.
```shell
Join Inner, (key#0 = key#2)
:- Filter isnotnull(key#0)
:  +- Relation inmemoryprivilegebuildersuite.inmemoryprivilegebuildersuite[key#0,value#1] parquet
+- Filter isnotnull(key#2)
   +- Relation default.tab1[key#2,value#3] parquet
```
To solve this bug, I think we should consider split the condition cols from the projectionList, and I have a PR.

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

packyan:

To close #2391 
Fixed an issue where PrivilegesBuilder would generate incorrect PrivilegeObjects when there is no Project operator in the LogicalPlan but there has Filter or Join operators.

- [] Add screenshots for manual tests if appropriate

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
@@ -76,7 +76,8 @@ object PrivilegesBuilder {
   private def buildQuery(
       plan: LogicalPlan,
       privilegeObjects: ArrayBuffer[PrivilegeObject],
-      projectionList: Seq[NamedExpression] = Nil): Unit = {
+      projectionList: Seq[NamedExpression] = Nil,
+      conditionList: Seq[NamedExpression] = Nil): Unit = {
 
     def mergeProjection(table: CatalogTable, plan: LogicalPlan): Unit = {
       if (projectionList.isEmpty) {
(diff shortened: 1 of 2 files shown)


yaooqinn:
LGTM if test pass

yaooqinn:
thanks, merged to master

PR #2446, opened 2022-04-22
[KYUUBI #2433] HiveSQLEngine load required jars from HIVE_HADOOP_CLASSPATH

deadwind4:

We submit a hive job to yarn, when we don't set `kyuubi.engine.hive.extra.classpath`, HiveSQLEngine will throw an error.

My improvement is supporting set `HIVE_HADOOP_CLASSPATH`

### DIFF
--- a/conf/kyuubi-env.sh.template
+++ b/conf/kyuubi-env.sh.template
@@ -46,6 +46,10 @@
 # - FLINK_HOME              Flink distribution which you would like to use in Kyuubi.
 # - FLINK_CONF_DIR          Optional directory where the Flink configuration lives.
 #                           (Default: $FLINK_HOME/conf)
+# - HIVE_HOME               Hive distribution which you would like to use in Kyuubi.
+# - HIVE_CONF_DIR           Optional directory where the Hive configuration lives.
+#                           (Default: $HIVE_HOME/conf)
+# - HIVE_HADOOP_CLASSPATH   Required hadoop jars when you use Kyuubi Hive engine on Yarn.
 #
 
 
@@ -54,6 +58,8 @@
 # export JAVA_HOME=/usr/jdk64/jdk1.8.0_152
 # export SPARK_HOME=/opt/spark
 # export FLINK_HOME=/opt/flink
+# export HIVE_HOME=/opt/hive
+# export HIVE_HADOOP_CLASSPATH=${HADOOP_HOME}/share/hadoop/common/lib/commons-collections-3.2.2.jar:${HADOOP_HOME}/share/hadoop/client/hadoop-client-runtime-3.1.0.jar:${HADOOP_HOME}/share/hadoop/client/hadoop-client-api-3.1.0.jar:${HADOOP_HOME}/share/hadoop/common/lib/htrace-core4-4.1.0-incubating.jar
 # export HADOOP_CONF_DIR=/usr/ndp/current/mapreduce_client/conf
 # export YARN_CONF_DIR=/usr/ndp/current/yarn/conf
 # export KYUUBI_JAVA_OPTS="-Xmx10g -XX:+UnlockDiagnosticVMOptions -XX:ParGCCardsPerStrideChunk=4096 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -XX:+CMSConcurrentMTEnabled -XX:CMSInitiatingOccupancyFraction=70 -XX:+UseCMSInitiatingOccupancyOnly -XX:+CMSClassUnloadingEnabled -XX:+CMSParallelRemarkEnabled -XX:+UseCondCardMark -XX:MaxDirectMemorySize=1024m  -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=./logs -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+PrintTenuringDistribution -Xloggc:./logs/kyuubi-server-gc-%t.log -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=10 -XX:GCLogFileSize=5M -XX:NewRatio=3 -XX:MetaspaceSize=512m"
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/hive/HiveProcessBuilder.scala
```diff
@@ -81,16 +81,25 @@ class HiveProcessBuilder(
...
       warn(s"The conf of kyuubi.engine.hive.extra.classpath is empty.")
-      mainResource.foreach { path =>
-        val devHadoopJars = Paths.get(path).getParent
-          .resolve(s"scala-$SCALA_COMPILE_VERSION")
-          .resolve("jars")
-        if (!Files.exists(devHadoopJars)) {
-          throw new KyuubiException(s"The path $devHadoopJars does not exists. Please set " +
-            s"kyuubi.engine.hive.extra.classpath for configuring location of " +
-            s"hadoop client jars, etc")
+      val distJars = new File(s"${env.get(KYUUBI_HOME).orNull}${File.separator}jars")
```
yaooqinn, line 84: Kyuubi Server lib is designed isolated with engine jars, if we mix them it's hard to maintain in the future

deadwind4: Could I package `hadoop-client*jar` and `common-collection-*jar` into `dist/externals/engines/hive/jars`, and load this path into classpath? Or, I enable `HIVE_HADOOP_CLASSPATH` to let users set these jars?

yaooqinn: > Could I package hadoop-client*jar and common-collection-*jar into dist/externals/engines/hive/jars

I have done this before, but this 1) make the binary distribution too large. 2) it may become a blocker for multiple hive version support

kyuubi.engine.hive.extra.classpath will do the same thing as HIVE_HADOOP_CLASSPATH

the thing is the AT-LEAST classpath of a hive engine is made by
- hive-sql-engine.jar
- HIVE_HOME/libs with a hive distribution
- hadoop client jars

And here we mainly focus on the **hadoop client jars** and the deps. 

- Is it possible to make it transparent for users？
   - a fat jar within  hive-sql-engine.jar?
   - make a separate folder and deliver it with the binary release?
- If not, how to make it easier for user to understand and use
   - documentation is required
   - HIVE_HADOOP_CLASSPATH or kyuubi.engine.hive.extra.classpath, or both?
      - what are their priorities?
      - shall we raise an error or not?

yaooqinn: Another thing that we shall not add the Hadoop client jars by default is
- both hive distribution and the hive-sql-engine.jar are not created for specific Hadoop versions, users can set accordingly. If we don't propagate them, class not found exception; if we do, compatibility issue lurks.

deadwind4: > If not, how to make it easier for user to understand and use

I second this. I will achieve loading both HIVE_HADOOP_CLASSPATH and `kyuubi.engine.hive.extra.classpath`.
HIVE_HADOOP_CLASSPATH is required. And I will explain how to set HIVE_HADOOP_CLASSPATH in doc.

deadwind4:
I will complete doc in this PR https://github.com/apache/incubator-kyuubi/pull/2326

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/hive/HiveProcessBuilder.scala
```diff
@@ -39,6 +39,8 @@ class HiveProcessBuilder(
 
   private val hiveHome: String = getEngineHome("hive")
 
+  private val HIVE_HADOOP_CLASSPATH: String = "HIVE_HADOOP_CLASSPATH"
```
yaooqinn, line 42: nit: unnecessary val?

yaooqinn:
Maybe we shall add `HIVE_HADOOP_CLASSPATH` to kyuubi-env.sh.template

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/hive/HiveProcessBuilder.scala
deadwind4: It's usefull. I change all "HIVE_HADOOP_CLASSPATH" literal to this constant. In the future, we could move all similar constants to a scala file.

The following code uses this constant.
https://github.com/deadwind4/incubator-kyuubi/blob/9f9fcdac5b6f77eb03e20d6e41f95b03a1dc9568/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/hive/HiveProcessBuilder.scala#L82-L95

deadwind4:
> Maybe we shall add `HIVE_HADOOP_CLASSPATH` to kyuubi-env.sh.template

I added it in template @yaooqinn

conf/kyuubi-env.sh.template
```diff
@@ -46,6 +46,10 @@
 # - FLINK_HOME              Flink distribution which you would like to use in Kyuubi.
 # - FLINK_CONF_DIR          Optional directory where the Flink configuration lives.
 #                           (Default: $FLINK_HOME/conf)
+# - HIVE_HOME               Hive distribution which you would like to use in Kyuubi.
+# - HIVE_CONF_DIR           Optional directory where the Hive configuration lives.
+#                           (Default: $HIVE_HOME/conf)
+# - HIVE_HADOOP_CLASSPATH   Required hadoop jars when you use Kyuubi Hive engine on Yarn.
```
yaooqinn, line 52: Not only for Yarn?

yaooqinn (approved):

PR #2448, opened 2022-04-23
Log the engine id when opening Kyuubi connection

turboFei:

Log the engine id when opening Kyuubi connection, it is more user friendly.


pan3793 (approved):

turboFei:
thanks, merging to master

Issue #2439, opened 2022-04-22
[Bug] kyuubi-tpcds will failed with "error=26, Text file busy"


bestbugwriter:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I want to use incubator-kyuubi/dev/kyuubi-tpcds to generate test datas. but I got an exception：
```
Caused by: java.io.IOException: Cannot run program "./dsdgen" (in directory "/tmp/spark-fee2eca2-caae-476a-8b53-2892ecb1045a/tpcds-8e81c7be-86c3-4e7c-a186-f31edee05210"): error=26, Text file busy
```
This problem is inevitable. 

I refer to this link for operation  https://github.com/apache/incubator-kyuubi/tree/master/dev/kyuubi-tpcds

my command: 
```
$SPARK_HOME/bin/spark-submit --driver-memory 16G --class org.apache.kyuubi.tpcds.DataGenerator /home/lpz/incubator-kyuubi/dev/kyuubi-tpcds/target/kyuubi-tpcds_2.12-1.6.0-SNAPSHOT.jar --db tpcds_test1 --scaleFactor 1 --format parquet --parallel 1
```

my hardware:
```
cpu: i5-8500
memory: 24G ddr4 2666
```

my os:
```
Linux U 5.13.0-40-generic #45~20.04.1-Ubuntu SMP Mon Apr 4 09:38:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

my java version:
```
openjdk version "1.8.0_312"
OpenJDK Runtime Environment (build 1.8.0_312-8u312-b07-0ubuntu1~20.04-b07)
OpenJDK 64-Bit Server VM (build 25.312-b07, mixed mode
```

my spark version:
```
spark-3.2.1-bin-hadoop3.2
```

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk
Driver stacktrace:
	at org.apache.spark.scheduler.DAGScheduler.failJobAndIndependentStages(DAGScheduler.scala:2454)
	at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2(DAGScheduler.scala:2403)
[...]

PR #2449, opened 2022-04-23
[KYUUBI #2439] Using Pure Java TPC-DS generator

pan3793:

This PR proposes change the Kyuubi TPC-DS generator to pure Java implementation instead of the original C binary.

The new pure Java TPC-DS generator is under Apache License, and in fact, I don't know the original C binary License, so we exclude them from release in the past.

Since the change removes the License issue of Kyuubi TPC-DS module, we can bundle the TPC-DS tool in the future release.

And after migration, I haven't see "error=26, Text file busy" described in #2439 any more.

1. Use old C binary based TPC-DS generator generate 1GB data under database `tpcds_s1`
2. Use new pure Java base TPC-DS generator generate 1GB data under database `new_tpcds_sf1`
3. Compare results of `select count(*)`, and `select sum(hash(*))`

```
spark-sql> select count(*) from tpcds_s1.inventory;
11745000
Time taken: 0.161 seconds, Fetched 1 row(s)
spark-sql> select count(*) from new_tpcds_sf1.inventory;
11745000
Time taken: 0.141 seconds, Fetched 1 row(s)
spark-sql> select sum(hash(*)) from tpcds_s1.inventory;
-556768665838
Time taken: 0.252 seconds, Fetched 1 row(s)
spark-sql> select sum(hash(*)) from new_tpcds_sf1.inventory;
-556768665838
Time taken: 0.232 seconds, Fetched 1 row(s)
```


yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #1936, opened 2022-02-18
[Bug] Engine token expired issue after idle for long time


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

We have enabled the `HadoopCredentialsManager` to refresh delegation token, but met token expired issue when the backend engine is idle for long time.

I think that we can refresh the token when establishing a connection to prevent this issue.
```
22/02/18 00:58:44 INFO ExecuteStatement: 
           Spark application name: kyuubi_apollorno_USER_SPARK_SQL_b_stf_default_06dc1722-f0ff-4535-bea1-d2abe5d3ebf2
                 application ID: application_1635906065713_7300973
                 application web UI: https://apollo-rno-rm-1.vip.hadoop.ebay.com:50030/proxy/application_1635906065713_7300973,https://apollo-rno-rm-2.vip.hadoop.ebay.com:50030/proxy/application_1635906065713_7300973
                 master: yarn
                 deploy mode: cluster
                 version: 3.1.1.0.6.0
           Start time: 2022-02-15T05:07:49.417
           User: b_stf
22/02/18 00:58:45 WARN Client: Exception encountered while connecting to the server : org.apache.hadoop.ipc.RemoteException(org.apache.hadoop.security.token.SecretManager$InvalidToken): token (HDFS_DELEGATION_TOKEN token 22821449 for b_stf) can't be found in cache
22/02/18 00:58:45 INFO DAGScheduler: Asked to cancel job group 292fd765-4f99-44f8-afb3-a35fcc96e1a6
22/02/18 00:58:45 INFO ExecuteStatement: Processing b_stf's query[292fd765-4f99-44f8-afb3-a35fcc96e1a6]: RUNNING_STATE -> ERROR_STATE, statement: select count(*) from ddm_notiplat_t.nss_event_queue where src_cre_dt>='2022-02-16 00:00:00', time taken: 1.108 seconds
22/02/18 00:58:45 ERROR ExecuteStatement: Error operating EXECUTE_STATEMENT: org.apache.hadoop.ipc.RemoteException(org.apache.hadoop.security.token.SecretManager$InvalidToken): token (HDFS_DELEGATION_TOKEN token 22821449 for b_stf) can't be found in cache
[...]

PR #2459, opened 2022-04-24
[KYUUBI #1936][FOLLOWUP] Stop updating credentials when credentials are expired

lightning-L:

This is the follow-up for the comments in https://github.com/apache/incubator-kyuubi/pull/2210 

------
But considering the following scenario, I think we should also check whether userRef in `userCredentialsRefMap` has been replaced by a new one .

`userRef` was waiting for next renewal.
`userRef` expired and was removed from `userCredentialsRefMap`
A new `userRef` was created because of user visiting.
Renewal of old userRef started again.
If we do not check, both old and new userRef will renew periodically.

modified the code and added test case for the scenario.

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
@@ -158,6 +158,18 @@ class HadoopCredentialsManagerSuite extends KyuubiFunSuite {
       eventually(timeout(9000.milliseconds), interval(100.milliseconds)) {
         assert(manager.userCredentialsRefMap.size == 0)
       }
+
+      // New userRef is created
+      val newUserRef = manager.getOrCreateUserCredentialsRef(appUser)
+      assert(manager.userCredentialsRefMap.size == 1)
+
+      // Old renewal schedule is stopped
+      val epoch = userRef.getEpoch
+      Thread.sleep(2000L)
+      assert(userRef.getEpoch == epoch)
+
+      // New renewal schedule is running
+      assert(newUserRef.getEpoch >= 1)
     }
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/CredentialsRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/CredentialsRef.scala
@@ -49,6 +49,10 @@ class CredentialsRef(appUser: String) {
 
   def getLastAccessTime: Long = lastAccessTime
 
+  def getNoOperationTime: Long = {
+    System.currentTimeMillis() - lastAccessTime
+  }
+
   def getEpoch: Long = epoch
 
   def getAppUser: String = appUser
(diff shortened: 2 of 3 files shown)


lightning-L:
Sorry for the late commit.

This is the follow-up for the comments in https://github.com/apache/incubator-kyuubi/pull/2210 
------
But considering the following scenario, I think we should also check whether userRef in `userCredentialsRefMap` has been replaced by a new one .

`userRef` was waiting for next renewal.
`userRef` expired and was removed from `userCredentialsRefMap`
A new `userRef` was created because of user visiting.
Renewal of old userRef started again.
If we do not check, both old and new userRef will renew periodically.

modified the code and added test case for the scenario.

turboFei:
cc @zhouyifan279

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
```diff
@@ -158,6 +158,49 @@ class HadoopCredentialsManagerSuite extends KyuubiFunSuite {
...
+      val userRef = manager.getOrCreateUserCredentialsRef(appUser)
+      assert(manager.userCredentialsRefMap.size == 1)
+
+      // Credentials are expired
+      eventually(timeout(9000.milliseconds), interval(100.milliseconds)) {
+        assert(manager.userCredentialsRefMap.size == 0)
+      }
+
+      // New userRef is created
+      val userRef_new = manager.getOrCreateUserCredentialsRef(appUser)
```
turboFei, line 191: nit: userRef_new -> newUserRef

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
```diff
@@ -158,6 +158,49 @@ class HadoopCredentialsManagerSuite extends KyuubiFunSuite {
       eventually(timeout(9000.milliseconds), interval(100.milliseconds)) {
         assert(manager.userCredentialsRefMap.size == 0)
```
turboFei, line 159: nit:
```    
  eventually(timeout(9000.milliseconds), interval(100.milliseconds)) {
        assert(manager.userCredentialsRefMap.size == 0)
        val epoch = userRef.getEpoch
        Thread.sleep(2000L)
        assert(userRef.getEpoch == epoch)
  }
```

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HadoopCredentialsManagerSuite.scala
```diff
@@ -158,6 +158,49 @@ class HadoopCredentialsManagerSuite extends KyuubiFunSuite {
...
+      assert(manager.userCredentialsRefMap.size == 1)
+
+      // Credentials are expired
+      eventually(timeout(9000.milliseconds), interval(100.milliseconds)) {
+        assert(manager.userCredentialsRefMap.size == 0)
+      }
+
+      // New userRef is created
+      val userRef_new = manager.getOrCreateUserCredentialsRef(appUser)
+      assert(manager.userCredentialsRefMap.size == 1)
```
turboFei, line 192: I think you can move this part into original UT.

turboFei (changes requested):

turboFei (approved):

zhouyifan279:
LGTM

turboFei:
thanks, merged to master

PR #2462, opened 2022-04-24
[KYUUBI #2461] Use the original host argument

### LINKED ISSUE #2461: [Bug] `bin/kyuubi-ctl` cannot delete server instance when kuubi server binds ip address.
### Code of Conduct

### Search before asking

### Describe the bug

server conf:
```
kyuubi.frontend.bind.host       196.168.10.148
```

bin/kyuubi-ctl list
```
/kyuubi 196.168.10.148    10015   1.6.0-SNAPSHOT
```

`bin/kyuubi-ctl delete -s 196.168.10.148 -p 10015` can't delete the server instance

`bin/kyuubi-ctl get -s 196.168.10.148 -p 10015` can't get the server list.

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

wForget:

close #2461

### DIFF
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
@@ -249,7 +249,8 @@ class ServiceControlCliArguments(args: Seq[String], env: Map[String, String] = s
     }
 
     try {
-      cliArgs = cliArgs.copy(host = InetAddress.getByName(cliArgs.host).getCanonicalHostName)
+      InetAddress.getByName(cliArgs.host)
+      cliArgs = cliArgs.copy(host = cliArgs.host)
     } catch {
       case _: Exception =>
         fail(s"Unknown host: ${cliArgs.host}")


kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/ServiceControlCliArguments.scala
```diff
@@ -249,7 +249,8 @@ class ServiceControlCliArguments(args: Seq[String], env: Map[String, String] = s
     }
 
     try {
-      cliArgs = cliArgs.copy(host = InetAddress.getByName(cliArgs.host).getCanonicalHostName)
+      InetAddress.getByName(cliArgs.host)
```
cxzl25, line 252: Because `kyuubi.frontend.connection.url.use.hostname` is configured to determine whether the node registered to zookeeper is hostname or ip, would it be better to use the original value passed in with `--host` here?

wForget: > Because `kyuubi.frontend.connection.url.use.hostname` is configured to determine whether the node registered to zookeeper is hostname or ip, would it be better to use the original value passed in with `--host` here?

When we configure `kyuubi.frontend.thrift.binary.bind.host`, the `kyuubi.frontend.connection.url.use.hostname` config does not seem to take effect, we use the original host argument will be more flexible.

cxzl25: The server is configured with `kyuubi.frontend.thrift.binary.bind.host`, so `kyuubi.frontend.connection.url.use.hostname` does not take effect.
https://github.com/apache/incubator-kyuubi/blob/deb0e6203b5b56a1c61162f793d3601be0395118/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala#L109-L119


Because there may be the ip used by the server and the hostname used by the engine, it is sometimes embarrassing to convert `--host`.

ulysses-you:
shall we just pass the input string which respect user specified ? It seems it's unnecessary to parse it to hostname

wForget:
> shall we just pass the input string which respect user specified ? It seems it's unnecessary to parse it to hostname

I think this should be done, this pr is like this.

wForget:
`InetAddress.getByName(cliArgs.host)` is only used to validate the host argument.

ulysses-you (approved):

ulysses-you:
I guess @cxzl25 also misunderstand the behavior according to the comments, isn't @cxzl25 ?

cxzl25 (approved):

ulysses-you:
thanks, merging to master

Issue #2471, opened 2022-04-25
[Bug] Failure to load the package correctly at kyuubi engine runtime with scala code mode


iodone:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

in scala mode:
```
spark.sql("add jar /path/to/test.jar");
import test._
```
Dynamically importing external dependency packages and not getting them loaded correctly

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2475, opened 2022-04-25
[KYUUBI #2471] Fix the bug of dynamically loading external packages

iodone:

In scala code mode implementation, we need to call `repl.addUrlsToClassPath` to set the classpath of the current thread to repl before entering `repl`


yaooqinn:
Can you verify using spark.sharedState.jarClassLoader at https://github.com/apache/incubator-kyuubi/blob/c29c4430aa0fb15de8984a24fa61490ec2cf5bdc/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala#L47

iodone:
> Can you verify using spark.sharedState.jarClassLoader at
> 
> https://github.com/apache/incubator-kyuubi/blob/c29c4430aa0fb15de8984a24fa61490ec2cf5bdc/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala#L47

OK， i will try it.

iodone:
@yaooqinn, Using `spark.sharedState.jarClassLoader` does not work, because this setting is in the process of initialization and will only be set once?

yaooqinn:
> @yaooqinn, Using `spark.sharedState.jarClassLoader` does not work, because this setting is in the process of initialization and will only be set once?

yes. looks we can get added jars from sc.listJars and resolve them accordingly？

iodone:
> 



> > @yaooqinn, Using `spark.sharedState.jarClassLoader` does not work, because this setting is in the process of initialization and will only be set once?
> yes. looks we can get added jars from sc.listJars and resolve them accordingly？

Can you explain further? I cannot get to your point, here is my understanding: 
When `spark.sql(add jar)` is executed, the added jar of`session.sharedState.jarClassLoader` is available and the classpath of the current thread is set. However, the classpath does not exist in the repl's runtime, and the path of the add jar needs to be added via `repl.addUrlsToClassPath` after each execution of the statement.

yaooqinn:
```scala
      val jars = spark.sparkContext.listJars().map { jar =>
        val jarStr = SparkFiles.get(new File(jar).getName)
        val jarUri = new Path(jarStr).toUri

        if (jarUri.getScheme == null) {
          new File(jarStr).toURI.toURL
        } else {
          jarUri.toURL
        }
      }
      repl.addUrlsToClassPath(jars: _*)
```

yaooqinn:
more simple way

```scala
      val jars = spark.sparkContext.listJars().map { jar =>
        spark.sharedState.jarClassLoader.getResource(new File(jar).getName)
      }
      repl.addUrlsToClassPath(jars: _*)
```

turboFei:
It seems that we add test resource `test-functions.jar` for four times.

Is it possible that other modules depends kyuubi-common:tests:jar?

iodone:
> It seems that we add test resource `test-functions.jar` for four times.
> 
> Is it possible that other modules depends kyuubi-common:tests:jar?

I have spent a lot of time solving this problem recently, and have tried many methods, but since each module is tested separately, putting it in another module will result in not finding the `test-function.jar` path. And this does not let the k8s integration test pass. I need some help, can you give me some advice?

turboFei:
> > It seems that we add test resource `test-functions.jar` for four times.
> > Is it possible that other modules depends kyuubi-common:tests:jar?
> 
> I have spent a lot of time solving this problem recently, and have tried many methods, but since each module is tested separately, putting it in another module will result in not finding the `test-function.jar` path. And this does not let the k8s integration test pass. I need some help, can you give me some advice?

Got it. It seems that spark does not support the jar with schema 'jar:fille:'.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteScala.scala
```diff
@@ -64,6 +68,27 @@ class ExecuteScala(
       if (legacyOutput.nonEmpty) {
         warn(s"Clearing legacy output from last interpreting:\n $legacyOutput")
       }
+
+      val jars = spark.sparkContext.listJars().flatMap { jar =>
+        val jarFile = new File(jar)
+        spark.sharedState.jarClassLoader.getURLs.groupBy(_.getFile.split("/").last).get(
+          jarFile.getName) match {
```
turboFei, line 75: Should we use URL?
```
        val jarName = new URL(jar).getFile.split("/").last
        spark.sharedState.jarClassLoader.getURLs.groupBy(_.getFile.split("/").last).get(
          jarName) match
```

iodone: `SparkContext.listJars` are of type `spark://` scheme, URL that does not support this type

turboFei: got it

turboFei:
```

**Comment by turboFei** [comment]
```
- scala code with loading external package at runtime  *** FAILED ***
  org.apache.kyuubi.shade.org.apache.hive.service.cli.HiveSQLException: Error operating EXECUTE_STATEMENT: java.lang.NullPointerException
	at scala.tools.nsc.Global.assoc$1(Global.scala:885)
	at scala.tools.nsc.Global.$anonfun$invalidateClassPathEntries$2(Global.scala:900)
	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:293)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at scala.collection.TraversableLike.flatMap(TraversableLike.scala:293)
	at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:290)
	at scala.collection.AbstractTraversable.flatMap(Traversable.scala:108)
	at scala.tools.nsc.Global.invalidateClassPathEntries(Global.scala:900)
	at scala.tools.nsc.Global.extendCompilerClassPath(Global.scala:835)
	at scala.tools.nsc.interpreter.IMain.addUrlsToClassPath(IMain.scala:254)
	at org.apache.kyuubi.engine.spark.operation.ExecuteScala.$anonfun$runInternal$1(ExecuteScala.scala:86)
	at org.apache.kyuubi.engine.spark.operation.SparkOperation.withLocalProperties(SparkOperation.scala:88)
	at org.apache.kyuubi.engine.spark.operation.ExecuteScala.runInternal(ExecuteScala.scala:60)
	at org.apache.kyuubi.operation.AbstractOperation.run(AbstractOperation.scala:157)
	at org.apache.kyuubi.session.AbstractSession.runOperation(AbstractSession.scala:94)
	at org.apache.kyuubi.engine.spark.session.SparkSessionImpl.runOperation(SparkSessionImpl.scala:65)
	at org.apache.kyuubi.session.AbstractSession.$anonfun$executeStatement$1(AbstractSession.scala:123)
	at org.apache.kyuubi.session.AbstractSession.withAcquireRelease(AbstractSession.scala:75)
	at org.apache.kyuubi.session.AbstractSession.executeStatement(AbstractSession.scala:120)
	at org.apache.kyuubi.service.AbstractBackendService.executeStatement(AbstractBackendService.scala:66)
	at org.apache.kyuubi.service.TFrontendService.ExecuteStatement(TFrontendService.scala:233)
	at org.apache.kyuubi.shade.org.apache.hive.service.rpc.thrift.TCLIService$Processor$ExecuteStatement.getResult(TCLIService.java:1557)
	at org.apache.kyuubi.shade.org.apache.hive.service.rpc.thrift.TCLIService$Processor$ExecuteStatement.getResult(TCLIService.java:1542)
	at org.apache.kyuubi.shade.org.apache.thrift.ProcessFunction.process(ProcessFunction.java:39)
	at org.apache.kyuubi.shade.org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:39)
	at org.apache.kyuubi.service.authentication.TSetIpAddressProcessor.process(TSetIpAddressProcessor.scala:36)
	at org.apache.kyuubi.shade.org.apache.thrift.server.TThreadPoolServer$WorkerProcess.run(TThreadPoolServer.java:286)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

**Comment by turboFei** [comment]
cc @pan3793 
Do you know the difference of k8s integration test?

**Comment by turboFei** [comment]
It seems that for integration test, the `test-functions.jar` is  an inner resource of kyuubi-common:tests jar.

So the UT will fail.

File: kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkQueryTests.scala [hunk 2]
```diff
@@ -608,6 +610,34 @@ trait SparkQueryTests extends HiveJDBCTestHelper {
...
   }
 
+  test("scala code with loading external package at runtime ") {
+    val jarDir = Utils.createTempDir().toFile
+
+    withJdbcStatement() { statement =>
+      statement.execute("SET kyuubi.operation.language=scala")
+
+      val jarFileInputStream =
+        Thread.currentThread().getContextClassLoader.getResourceAsStream("test-function.jar")
```
turboFei, line 620: It seems that the `test-function.jar` must be a File instead of resource.

turboFei: let me raise a PR based on yours and have a try.

turboFei:
```
error file=/home/runner/work/incubator-kyuubi/incubator-kyuubi/kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkQueryTests.scala message=Line contains a tab line=626 column=11
error file=/home/runner/work/incubator-kyuubi/incubator-kyuubi/kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkQueryTests.scala message=Line contains a tab line=628 column=11
error file=/home/runner/work/incubator-kyuubi/incubator-kyuubi/kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkQueryTests.scala message=Line contains a tab line=629 column=11
error file=/home/runner/work/incubator-kyuubi/incubator-kyuubi/kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkQueryTests.scala message=Line contains a tab line=630 column=11
```
You can check scala style with below command:
```
build/mvn  scalastyle:check 

```

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkQueryTests.scala
```diff
@@ -608,6 +613,73 @@ trait SparkQueryTests extends HiveJDBCTestHelper {
     }
   }
 
+  test("scala code with loading external package at runtime ") {
+    val jarDir = Utils.createTempDir().toFile
+
+    withJdbcStatement() { statement =>
+      statement.execute("SET kyuubi.operation.language=scala")
+      val udfCode =
```
turboFei, line 621: or add 

```
      // scalastyle:off
      val udfCode =
        """
          |package test.utils
          |
          |object Math {
          |	def add(x: Int, y: Int): Int = x + y
          |
          |	def main(args: Array[String]): Unit = {
          |		println(add(1, 2))
          |	}
          |}
          |
          |""".stripMargin
      // scalastyle:on
```

iodone:
> It seems that for integration test, the `test-functions.jar` is an inner resource of kyuubi-common:tests jar.
> 
> So the UT will fail.

Checked out the IT process and found some differences from UT.
image
This latest commit is tested by loading a jar package from external in K8s IT, similar to local file loading and OnYarn from HDFS loading. The jar file generated on the Client side needs a shared storage similar to HDFS to ensure that the Spark Engine Pod can access it. 
Above I think this is the reason why IT test can not pass，@turboFei WDYT?

ulysses-you:
So the blocker is: the new added test depends on a local jar but the k8s-it failed since the pod can not aware of the jar location ?

We can add the test in `SparkOperationSuite` first. In real world, the location of added jar should always at remote.

iodone:
> So the blocker is: the new added test depends on a local jar but the k8s-it failed since the pod can not aware of the jar location ?
> 
> We can add the test in `SparkOperationSuite` first. In real world, the location of added jar should always at remote.

Yes, I think this problem is causing IT blocker. I'll try it to follow your suggestion, thanks!

integration-tests/kyuubi-kubernetes-it/pom.xml
```diff
@@ -80,6 +80,18 @@
...
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-compiler</artifactId>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
iodone: yes, IT tests extend from `SparkQueryTests`. `SparkQueryTests` depends on the above two packages.

ulysses-you (approved):

ulysses-you:
thanks, merging to master

yaooqinn:
Late + 1

PR #2477, opened 2022-04-25
Change state early on stopping

pan3793:

Change state early on stopping to avoid unnecessary invoke.


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -48,8 +48,8 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
   override val backendService = new SparkSQLBackendService(spark)
   override val frontendServices = Seq(new SparkTBinaryFrontendService(this))
 
-  @volatile private var shutdown = false
-  @volatile private var deregistered = false
+  private val shutdown = new AtomicBoolean(false)
+  private val deregistered = new AtomicBoolean(false)
```
ulysses-you, line 52: as we are here, we can move `deregistered` inside `startLifetimeTerminatingChecker`

ulysses-you:
thanks, merging to master

PR #2485, opened 2022-04-26
[KYUUBI #2484] Add conf to SessionEvent and display it in EngineSessionPage

### LINKED ISSUE #2484: [FEATURE] show session level runtime config for SparkSql engine
### Code of Conduct

### Search before asking

### Describe the feature

show session level runtime config for SparkSqlEngine. 

### Motivation

> For now, a SparkSQLEngine can be shared by USER/POOL. In this case, each connection holds an isolated spark session instance. And each isolated spark session instance inherits its own runtime configs from the SparkSQLEngine.

> we want to know the runtime configs of each SQL, especially when a SQL throws errors or has a bad performance.

> if community agree with that, I will create a PR for this feature 

### Describe the solution

My proposal is that we add a runtime config  page to `Kyuubi Query Engine Session` tab 

image

### Additional context

_No response_

wForget:

close #2484

image

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/SessionEvent.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/SessionEvent.scala
@@ -43,6 +43,7 @@ case class SessionEvent(
     username: String,
     ip: String,
     serverIp: String,
+    conf: Map[String, String],
     startTime: Long,
     var endTime: Long = -1L,
     var totalOperations: Int = 0) extends KyuubiEvent with SparkListenerEvent {
@@ -70,6 +71,7 @@ object SessionEvent {
       session.user,
       session.ipAddress,
       session.serverIpAddress,
+      session.conf,
       session.createTime)
   }
 }
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineSessionPage.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineSessionPage.scala
@@ -23,6 +23,7 @@ import javax.servlet.http.HttpServletRequest
 import scala.xml.Node
 
 import org.apache.spark.internal.Logging
+import org.apache.spark.internal.config.SECRET_REDACTION_PATTERN
 import org.apache.spark.ui.UIUtils._
 import org.apache.spark.util.Utils
 
@@ -30,6 +31,10 @@ import org.apache.spark.util.Utils
 case class EngineSessionPage(parent: EngineTab)
   extends WebUIPage("session") with Logging {
   val store = parent.store
+
+  private def propertyHeader = Seq("Name", "Value")
+  private def headerClasses = Seq("sorttable_alpha", "sorttable_alpha")
+  private def propertyRow(kv: (String, String)) = <tr><td>{kv._1}</td><td>{kv._2}</td></tr>
 
   /** Render the page */
   def render(request: HttpServletRequest): Seq[Node] = {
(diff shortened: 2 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/SessionEvent.scala
```diff
@@ -43,6 +43,7 @@ case class SessionEvent(
     username: String,
     ip: String,
     serverIp: String,
+    conf: Map[String, String],
```
gabrywu, line 46: I think this is connection initializing config map, how about the configs from `set` command

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineSessionPage.scala
```diff
@@ -49,6 +82,7 @@ case class EngineSessionPage(parent: EngineTab)
           Session created at {formatDate(sessionStat.startTime)},
           Total run {sessionStat.totalOperations} SQL
         </h4> ++
+        sessionPropertiesTable ++
```
gabrywu, line 85: Is it better to be shown after SQL stats?

gabrywu (commented):
suggestions

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/SessionEvent.scala
wForget: > I think this is connection initializing config map, how about the configs from `set` command

The set statement will be displayed in the statement list, we don't need to override it here, it might be better to show the initial configurations.

wForget:
> need more unit tests

Thanks, I'll add it later

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineSessionPage.scala
gabrywu: OK, let keep it before `SQL stats`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/SessionEvent.scala
gabrywu: Yes, can't agree more. Comment on purpose 😁

turboFei:
nice feature, thanks

gabrywu (commented):
🐂

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineSessionPage.scala
```diff
@@ -40,6 +45,34 @@ case class EngineSessionPage(parent: EngineTab)
...
+        case Some(ui) => Some(ui.conf.get(SECRET_REDACTION_PATTERN))
+        case None => SECRET_REDACTION_PATTERN.defaultValue
+      }
+
+      val sessionPropertiesTable =
+        if (sessionStat.conf != null && !sessionStat.conf.isEmpty) {
+          val table = UIUtils.listingTable(
+            propertyHeader,
+            propertyRow,
+            Utils.redact(redactionPattern, sessionStat.conf.toArray),
```
ulysses-you, line 58: nit: sort the conf to make the ordering more readable

ulysses-you (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineSessionPage.scala
wForget: done

cfmcgrady (approved):
LGTM

turboFei (approved):

turboFei:
thanks, merged to master and branch-1.5

PR #2510, opened 2022-04-28
Fix NPE when invoking YarnApplicationOperation::getApplicationInfoByTag

turboFei:

null value might be returned for YarnApplicationOperation::getApplicationInfoByTag

I met NPE issue yesterday.

image
image

Here the state might be null, because Some(null) is nonEmpty.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
@@ -78,7 +78,9 @@ class KyuubiApplicationManager extends AbstractService("KyuubiApplicationManager
   def getApplicationInfo(
       clusterManager: Option[String],
       tag: String): Option[Map[String, String]] = {
-    operations.find(_.isSupported(clusterManager)).map(_.getApplicationInfoByTag(tag))
+    operations.find(_.isSupported(clusterManager)).flatMap { m =>
+      Option(m.getApplicationInfoByTag(tag))
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


yaooqinn:
I don't understand，can you add a test

turboFei:
I will try to add UT, I met NPE issue yesterday.

image
image

Here the state might be null, because Some(null) is nonEmpty.

yaooqinn:
the state can be null, how about the status?

turboFei:
image

If the applicationStatus is Some(null).

Then NPE thrown.

```
      val state = applicationStatus.get(ApplicationOperation.APP_STATE_KEY)
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -78,7 +78,9 @@ class KyuubiApplicationManager extends AbstractService("KyuubiApplicationManager
   def getApplicationInfo(
       clusterManager: Option[String],
       tag: String): Option[Map[String, String]] = {
-    operations.find(_.isSupported(clusterManager)).map(_.getApplicationInfoByTag(tag))
+    operations.find(_.isSupported(clusterManager)).flatMap { m =>
```
yaooqinn, line 81: Nit:
```scala
  val option = operations.find(_.isSupported(clusterManager))
    option match {
      case Some(op) => Option(op.getApplicationInfoByTag(tag))
      case None => None
    }
```

yaooqinn (approved):

turboFei:
thanks, merging to master

PR #2517, opened 2022-04-29
Rename ZorderSqlAstBuilder to KyuubiSparkSQLAstBuilder

pan3793:

Rename `ZorderSqlAstBuilder` to a generic name and then we can extend other syntax than z-order, e.g. `SHOW CATALOGS` for Spark 3.1/3.2

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-common/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSparkSQL.g4
+++ b/extensions/spark/kyuubi-extension-spark-common/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSparkSQL.g4
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-grammar KyuubiSqlExtensions;
+grammar KyuubiSparkSQL;
 
 @members {
    /**
--- a/extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLParser.scala
+++ b/extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLParser.scala
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.kyuubi.sql.zorder
+package org.apache.kyuubi.sql
 
 import java.time.LocalDate
 import java.util.Locale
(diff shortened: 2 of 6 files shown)


extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLParser.scala
```diff
@@ -40,10 +40,10 @@ import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
 
-import org.apache.kyuubi.sql.KyuubiSQLConf
-import org.apache.kyuubi.sql.zorder.ZorderSqlExtensionsParser.{BigDecimalLiteralContext, BigIntLiteralContext, BooleanLiteralContext, DecimalLiteralContext, DoubleLiteralContext, IntegerLiteralContext, LogicalBinaryContext, MultipartIdentifierContext, NullLiteralContext, NumberContext, OptimizeZorderContext, PassThroughContext, QueryContext, SingleStatementContext, SmallIntLiteralContext, StringLiteralContext, TinyIntLiteralContext, TypeConstructorContext, ZorderClauseContext}
+import org.apache.kyuubi.sql.KyuubiSqlExtensionsParser._
+import org.apache.kyuubi.sql.zorder.{OptimizeZorderStatement, OptimizeZorderStatementBase, Zorder, ZorderBase}
 
-abstract class ZorderSqlAstBuilderBase extends ZorderSqlExtensionsBaseVisitor[AnyRef] {
+abstract class KyuubiSQLParserBase extends KyuubiSqlExtensionsBaseVisitor[AnyRef] {
```
pan3793, line 46: How about rename `KyuubiSqlExtensions` to `KyuubiSQL` as we use the upper case `SQL` everywhere other than this place.

extensions/spark/kyuubi-extension-spark-common/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSparkSQL.g4
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-grammar ZorderSqlExtensions;
+grammar KyuubiSparkSQL;
```
pan3793, line 18: I rename it to `KyuubiSparkSQL` because 1) we use the upper case `SQL` anywhere other than this place 2) we may introduce `KyuubiSQL` in server module in the future.

ulysses-you (approved):

pan3793:
Thanks, the failed test does not related, merging to master

PR #2520, opened 2022-04-29
Support SHOW CATALOGS syntax in Spark extensions

pan3793:

Spark support multi-catalogs since 3.0, and it's important to Kyuubi as Kyuubi aim to be a Gateway of LakeHouse.

The implementation is different from SPARK-35973 (available since 3.3.0).

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/spark/sql/catalog/ShowCatalogsCommand.scala
+++ b/extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/spark/sql/catalog/ShowCatalogsCommand.scala
@@ -19,15 +19,16 @@ package org.apache.spark.sql.catalog
 
 import org.apache.spark.sql.{Row, SparkSession}
 import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.util.StringUtils
 import org.apache.spark.sql.connector.catalog.CatalogManager
-import org.apache.spark.sql.execution.command.LeafRunnableCommand
+import org.apache.spark.sql.execution.command.RunnableCommand
 import org.apache.spark.sql.types.StringType
 
 /**
  * The command for `SHOW CATALOGS`.
  */
-case class ShowCatalogsCommand(pattern: Option[String]) extends LeafRunnableCommand {
+case class ShowCatalogsCommand(pattern: Option[String]) extends RunnableCommand {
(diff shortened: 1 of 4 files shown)


turboFei:
Looks that we can also add a `show sessions` syntax to show the  openSessions in current spark engine?

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/spark/sql/catalog/ShowCatalogsCommand.scala
```diff
@@ -0,0 +1,48 @@
...
+case class ShowCatalogsCommand(pattern: Option[String]) extends RunnableCommand {
+  override val output: Seq[Attribute] = Seq(
+    AttributeReference("catalog", StringType, nullable = false)())
+
+  // The implementation use eager strategy to list catalog, which is different from SPARK-35973
+  override def run(sparkSession: SparkSession): Seq[Row] = {
+    val configuredCatalogs = sparkSession.sessionState.conf.getAllConfs.keys
+      .filter { _ startsWith "spark.sql.catalog." }
+      .map { _ stripPrefix "spark.sql.catalog." }
+      .filterNot { _ contains "." }
```
ulysses-you, line 40: what if the catalog has multi-identifier, a.b.c

pan3793: good question, after reading spark code, I think `a.b.c` is a legal catalog name. let me try

pan3793:
We have a previous discussion https://github.com/apache/incubator-kyuubi/issues/929, eagerly load catalogs is more intuitive

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/spark/sql/catalog/ShowCatalogsCommand.scala
pan3793: it's legal,   as long as we use back quote "use \`p.g\`"

turboFei: but it seems that `spark.sql.catalog.`p.g`` also contains "."

pan3793: yea, i see. i'm thinking if we can do a light check and cache the result to implement the eager load.

PR #2536, opened 2022-05-02
Verify whether yarn client is valid in YarnApplicationOperation initialization

turboFei:

For UT, without `MiniYarnService`, yarn cluster manager is not supported.

If I specify the spark.master to yarn, it will use YarnApplicationOperation to killApplication.

In fact, the yarn client in YarnApplicationOperation is invalid, so the killApplication will stuck for about 20 minutes and then fail.


pan3793 (approved):

turboFei:
```
# yarn client timeout parameters, for YarnApplicationOperation
# Using KyuubiConf().loadFileDefaults to load this.
ipc.client.connect.max.retries=1
yarn.client.failover-max-attempts=1
```

PR #2548, opened 2022-05-03
Prevent dead loop if the batch job submission process it not alive

turboFei:

If the batch job submission process is not alive, out of the get applicationStatus loop


pan3793 (approved):

turboFei:
thanks, merged to master

PR #2552, opened 2022-05-05
[KYUUBI #2500] Command OptionParser for launching Flink Backend Engine

### LINKED ISSUE #2500: [Improvement] Command OptionParser for launching Flink Backend Engine
### Code of Conduct

### Search before asking

after 
https://github.com/apache/incubator-kyuubi/pull/2418
finished
we should change Flink Engine as  hive does in
https://github.com/apache/incubator-kyuubi/pull/2356 

### How should we improve?

_No response_

jiaoqingbo:

fix #2500 

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
@@ -46,7 +46,7 @@ class FlinkProcessBuilder(
   override protected def mainClass: String = "org.apache.kyuubi.engine.flink.FlinkSQLEngine"
 
   override protected def childProcEnv: Map[String, String] = conf.getEnvs +
-    ("FLINK_CONF_DIR" -> s"$FLINK_HOME/conf")
+    ("FLINK_CONF_DIR" -> env.getOrElse("FLINK_CONF_DIR", s"$FLINK_HOME${File.separator}conf"))
 
   override protected def commands: Array[String] = {
     val buffer = new ArrayBuffer[String]()
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
@@ -46,30 +46,17 @@ class FlinkProcessBuilder(
   override protected def mainClass: String = "org.apache.kyuubi.engine.flink.FlinkSQLEngine"
 
   override protected def childProcEnv: Map[String, String] = conf.getEnvs +
-    ("FLINK_HOME" -> FLINK_HOME) +
-    ("FLINK_CONF_DIR" -> s"$FLINK_HOME/conf") +
-    ("_FLINK_HOME_DETERMINED" -> s"1")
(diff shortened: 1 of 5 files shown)


jiaoqingbo:
cc @yaooqinn

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
```diff
@@ -54,27 +54,22 @@ class FlinkProcessBuilder(
     val buffer = new ArrayBuffer[String]()
     buffer += s"bash"
     buffer += s"-c"
-    val commandStr = new StringBuilder()
+    val commandSet = new LinkedHashSet[String]()
```
yaooqinn, line 57: why not use the `buffer`

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
```diff
@@ -54,27 +54,22 @@ class FlinkProcessBuilder(
...
-    commandStr.append(s"source $FLINK_HOME${File.separator}bin" +
+    commandSet.add(s"source $FLINK_HOME${File.separator}bin" +
       s"${File.separator}config.sh && $executable")
 
     // TODO: How shall we deal with proxyUser,
     // user.name
     // kyuubi.session.user
     // or just leave it, because we can handle it at operation layer
-    commandStr.append(s" -D$KYUUBI_SESSION_USER_KEY=$proxyUser ")
+    commandSet.add(s"-D$KYUUBI_SESSION_USER_KEY=$proxyUser")
```
yaooqinn, line 66: why not pass via --conf

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
jiaoqingbo: I need to treat the contents contained in the commandSet as a whole string, while separating the contents with spaces. this string as an argument to bash -C

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
jiaoqingbo: I found that hive is passed through the -D parameter, so it remains consistent with hive. Do I also need to change hive's?

yaooqinn: yes

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
yaooqinn: i don't quite get it

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
jiaoqingbo: OK I will change it later

yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #2522, opened 2022-04-29
Flaky Test: KyuubiBatchYarnClusterSuite - open batch session


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

```
KyuubiBatchYarnClusterSuite:
- open batch session *** FAILED ***
  The code passed to eventually never returned normally. Attempted 207 times over 10.023100855000001 seconds. Last failure message: RUNNING did not equal ERROR. (KyuubiBatchYarnClusterSuite.scala:83)
```

### Affects Version(s)

1.6.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

turboFei:
```
KyuubiBatchYarnClusterSuite:
- open batch session *** FAILED ***
  The code passed to eventually never returned normally. Attempted 3578 times over 3.0008272455666667 minutes. Last failure message: FINISHED did not equal ERROR. (KyuubiBatchYarnClusterSuite.scala:87)
```

Related code:
```
    assert(appInfo("state") === "KILLED")

    eventually(timeout(3.minutes), interval(50.milliseconds)) {
      assert(batchJobSubmissionOp.getStatus.state === ERROR)
    }
```

It seems that, in yarn-client mode, even we kill the yarn application, the local process exit code is still 0.

So, I think that, we need check the application state according to resource manager state.

PR #2557, opened 2022-05-05
[KYUUBI #2522] Even the process exit code is zero, also check the application state from resource manager

turboFei:

In yarn-client mode, even we kill the application, the local process exit code also could be `0`.

So, we need check the application state from resource manager to have a double check.


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -118,6 +118,13 @@ class BatchJobSubmission(session: KyuubiBatchSessionImpl, batchRequest: BatchReq
         if (process.exitValue() != 0) {
           throw new KyuubiException(s"Process exit with value ${process.exitValue()}")
         }
+
+        val finalApplicationStatus = currentApplicationState
+        if (finalApplicationStatus.map(_.get(ApplicationOperation.APP_STATE_KEY)).filter(s =>
```
yaooqinn, line 123: .exists(s => s.contains("KILLED") || s.contains("FAILED"))

ulysses-you:
we should tune back the timeout to see the behavior

turboFei:
> #2522

Before,  It is` RUNNING did not equal ERROR`.
```
KyuubiBatchYarnClusterSuite:
- open batch session *** FAILED ***
  The code passed to eventually never returned normally. Attempted 207 times over 10.023100855000001 seconds. Last failure message: RUNNING did not equal ERROR. (KyuubiBatchYarnClusterSuite.scala:83)
``` 

Now, it is `FINISHED did not equal ERROR.`

turboFei:
> we should tune back the timeout to see the behavior

I think that, we can set `spark.submit.deployMode` to cluster. It should terminate fast.

yaooqinn:
So, please do not merge a PR if it is well explained. please ask the author for an explanation over and over until you understand what the PR is doing. @ulysses-you @turboFei

yaooqinn (approved):

turboFei:
have discussed with @ulysses-you  offline.

We think that it is better to check the applicationState periodically and destroy the process once the application state is failed.

It might increase the rpc call to resource manager.

How do you think about this?
@yaooqinn

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -99,25 +100,34 @@ class BatchJobSubmission(session: KyuubiBatchSessionImpl, batchRequest: BatchReq
...
         process.destroyForcibly()
         throw new RuntimeException("Batch job failed:" + applicationStatus.get.mkString(","))
       } else {
-        process.waitFor()
         if (process.exitValue() != 0) {
           throw new KyuubiException(s"Process exit with value ${process.exitValue()}")
         }
+
+        applicationStatus = currentApplicationState
+        if (applicationFailed(applicationStatus)) {
```
ulysses-you, line 127: seems this branch is unnecessay ? we have checked status in if branch

turboFei: the while loop might terminate because the process is not alive.

So, the application status might not be latest.

So, It is better to get final application status at last.

ulysses-you: I think it's minor .. the check of `isActive` is a lightweight operation so the status should always be same

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -99,25 +100,34 @@ class BatchJobSubmission(session: KyuubiBatchSessionImpl, batchRequest: BatchReq
...
       }
 
-      if (applicationStatus.map(_.get(ApplicationOperation.APP_STATE_KEY)).filter(s =>
-          s == Some("KILLED") || s == Some("FAILED")).isDefined) {
+      if (applicationFailed(applicationStatus)) {
         process.destroyForcibly()
         throw new RuntimeException("Batch job failed:" + applicationStatus.get.mkString(","))
       } else {
-        process.waitFor()
         if (process.exitValue() != 0) {
```
ulysses-you, line 122: `exitValue` can throw exception is the process is still alive.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: ok

turboFei:
cc @yaooqinn  because there is code change

ulysses-you (approved):

turboFei:
thanks for your review, merging to master to fix the flaky test

PR #2551, opened 2022-05-04
[KYUUBI #2550] Fix swagger does not show the request/response schema issue

turboFei:
### _Why are the changes needed?_

To close #2550

Refer the comments: https://github.com/apache/incubator-kyuubi/pull/1658#issuecomment-1003244581

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [x] Add screenshots for manual tests if appropriate
image

image

- [x] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #2567, opened 2022-05-06
[KYUUBI #2378] Implement BatchesResource GET /batches/${batchId}/log

ulysses-you:

Support fetch operation log with rest batch mode.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
@@ -21,8 +21,9 @@ import java.io.{BufferedReader, IOException}
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Path, Paths}
-import java.util.{ArrayList => JArrayList}
+import java.util.{ArrayList => JArrayList, List => JList}
 
+import scala.collection.JavaConverters._
 import scala.collection.mutable.ListBuffer
 
 import org.apache.hive.service.rpc.thrift.{TColumn, TRow, TRowSet, TStringColumn}
@@ -86,11 +87,15 @@ class OperationLog(path: Path) {
   private lazy val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
   private lazy val reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)
 
+  private lazy val extraPaths: ListBuffer[Path] = ListBuffer()
   private lazy val extraReaders: ListBuffer[BufferedReader] = ListBuffer()
+  private var lastReadPos = 0
+  private var seekableReader: SeekableBufferedReader = _
 
   def addExtraLog(path: Path): Unit = synchronized {
     try {
       extraReaders += Files.newBufferedReader(path, StandardCharsets.UTF_8)
+      extraPaths += path
     } catch {
       case _: IOException =>
     }
@@ -130,6 +135,13 @@ class OperationLog(path: Path) {
     }
   }
 
+  private def toRowSet(logs: JList[String]): TRowSet = {
+    val tColumn = TColumn.stringVal(new TStringColumn(logs, ByteBuffer.allocate(0)))
+    val tRow = new TRowSet(0, new JArrayList[TRow](logs.size()))
+    tRow.addToColumns(tColumn)
+    tRow
+  }
+
   /**
    * Read to log file line by line
    *
@@ -144,10 +156,29 @@ class OperationLog(path: Path) {
       logs.addAll(extraLogs)
     }
 
-    val tColumn = TColumn.stringVal(new TStringColumn(logs, ByteBuffer.allocate(0)))
-    val tRow = new TRowSet(0, new JArrayList[TRow](logs.size()))
-    tRow.addToColumns(tColumn)
-    tRow
+    toRowSet(logs)
+  }
+
+  def read(from: Int, size: Int): TRowSet = synchronized {
+    var pos = from
+    if (pos < 0) {
+      // just fetch forward
+      pos = lastReadPos
+    }
+    if (seekableReader == null) {
+      seekableReader = new SeekableBufferedReader(Seq(path) ++ extraPaths)
+    } else {
+      // if from < last pos, we should reload the reader
+      // otherwise, we can reuse the existed reader for better performance
+      if (pos < lastReadPos) {
+        seekableReader.close()
+        seekableReader = new SeekableBufferedReader(Seq(path) ++ extraPaths)
+      }
+    }
+
+    val it = seekableReader.readLine(pos, size)
+    lastReadPos = pos + size
+    toRowSet(it.toList.asJava)
   }
 
   def close(): Unit = synchronized {
@@ -158,6 +189,13 @@ class OperationLog(path: Path) {
     }
     trySafely {
       writer.close()
+    }
+
+    if (seekableReader != null) {
+      lastReadPos = 0
+      trySafely {
+        seekableReader.close()
+      }
     }
 
     trySafely {
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
@@ -89,7 +89,7 @@ class OperationLog(path: Path) {
 
   private lazy val extraPaths: ListBuffer[Path] = ListBuffer()
   private lazy val extraReaders: ListBuffer[BufferedReader] = ListBuffer()
-  private var lastReadPos = 0
+  private var lastSeekReadPos = 0
   private var seekableReader: SeekableBufferedReader = _
 
   def addExtraLog(path: Path): Unit = synchronized {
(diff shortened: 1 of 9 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -41,7 +41,7 @@ class BatchJobSubmission(session: KyuubiBatchSessionImpl, batchRequest: BatchReq
 
   override def shouldRunAsync: Boolean = true
 
-  private lazy val _operationLog = OperationLog.createOperationLog(session, getHandle)
```
ulysses-you, line 44: it's unnecessary since we always eagerly materialize SparkBatchProcessBuilder

ulysses-you:
cc @turboFei @zhaomin1423

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -71,23 +87,38 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+  @ApiResponse(
+    responseCode = "200",
+    content = Array(new Content(
+      mediaType = MediaType.APPLICATION_JSON)),
+    description = "get operation log")
+  @GET
+  @Path("{batchId}/log")
+  def getBatchLog(
+      @PathParam("batchId") batchId: String,
+      @QueryParam("maxRows") maxRows: Int): OperationLog = {
```
turboFei, line 107: for restful api, it is stateless.

So I think it is better that we can specify the from offset when fetching logs.
If the from offset is not specified, we can always forward.

turboFei: cc @yaooqinn

ulysses-you: Looks fine to me, we can do this in another pr. The rest non-batch code path can also support this feature.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -71,23 +87,38 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
-    Batch(
-      batchOp.batchId,
-      batchOp.batchType,
-      batchOp.currentApplicationState.getOrElse(Map.empty),
-      fe.connectionUrl,
-      batchOp.getStatus.state.toString)
+  @ApiResponse(
+    responseCode = "200",
+    content = Array(new Content(
+      mediaType = MediaType.APPLICATION_JSON)),
```
turboFei, line 101: please rebase on #2551

```
schema = new Schema(....)
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
zhaomin1423: https://github.com/apache/incubator-kyuubi/discussions/2304

ulysses-you: @zhaomin1423 thank you for the reminder !

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
ulysses-you: rebased

ulysses-you:
let's wait https://github.com/apache/incubator-kyuubi/pull/2573

ulysses-you:
@turboFei @zhaomin1423 it's ready to review

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -133,6 +133,16 @@ class BatchJobSubmission(session: KyuubiBatchSessionImpl, batchRequest: BatchReq
     }
   }
 
+  def getOperationLogRowSet(
+      order: FetchOrientation,
+      from: Int,
+      size: Int): TRowSet = {
+    val operationLog = getOperationLog
+    operationLog.map(_.read(from, size)).getOrElse {
```
turboFei, line 141: It seems that, if the from < 0, we can invoke `def read(maxRows: Int): TRowSet` to read forward and `seekableReader` is not needed.

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -144,10 +156,29 @@ class OperationLog(path: Path) {
...
-    tRow.addToColumns(tColumn)
-    tRow
+    toRowSet(logs)
+  }
+
+  def read(from: Int, size: Int): TRowSet = synchronized {
+    var pos = from
+    if (pos < 0) {
+      // just fetch forward
+      pos = lastReadPos
```
turboFei, line 166: if pos <0, could we invoke `def read(maxRows: Int): TRowSet` directly?

turboFei: I see,
Now we have reader/extraReaders for read forward and seekableReader for read from offset.

And the lastReadPosition is not consistent.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -133,6 +133,16 @@ class BatchJobSubmission(session: KyuubiBatchSessionImpl, batchRequest: BatchReq
     }
   }
 
+  def getOperationLogRowSet(
+      order: FetchOrientation,
```
turboFei, line 137: the order is never used

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
turboFei: It seems not matter, only one more seekableReader will be created.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
ulysses-you: I considered the hybrid case that the first request is [-1, 100), the second request is [100, 110). So it would have better performance if we always use seekable reader.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -102,14 +116,33 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
-      batchOp.batchType,
-      batchOp.currentApplicationState.getOrElse(Map.empty),
-      fe.connectionUrl,
-      batchOp.getStatus.state.toString)
+  @ApiResponse(
+    responseCode = "200",
+    content = Array(new Content(
+      mediaType = MediaType.APPLICATION_JSON,
+      schema = new Schema(implementation = classOf[OperationLog]))),
+    description = "get operation log")
```
turboFei, line 124: get the log lines from this batch

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
ulysses-you: yea.. I just leave it here since the thrift code path did so

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: got it

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
ulysses-you: yea, we should use the one reader to track the last position, otherwise it become complex

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -144,10 +156,29 @@ class OperationLog(path: Path) {
...
+      // if from < last pos, we should reload the reader
+      // otherwise, we can reuse the existed reader for better performance
+      if (pos < lastReadPos) {
+        seekableReader.close()
+        seekableReader = new SeekableBufferedReader(Seq(path) ++ extraPaths)
+      }
+    }
+
+    val it = seekableReader.readLine(pos, size)
+    lastReadPos = pos + size
```
turboFei, line 180: the it.size may be small than size.

```
    lastReadPos = pos + it.size
```

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -86,11 +87,15 @@ class OperationLog(path: Path) {
   private lazy val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
   private lazy val reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)
 
+  private lazy val extraPaths: ListBuffer[Path] = ListBuffer()
   private lazy val extraReaders: ListBuffer[BufferedReader] = ListBuffer()
+  private var lastReadPos = 0
```
turboFei, line 92: maybe `lastSeekReadPos` is better?

turboFei:
Seems a new flaky test:
```
- JpsApplicationOperation with spark local mode *** FAILED ***
  response._1 was false APPLICATION_NOT_FOUND (JpsApplicationOperationSuite.scala:87)
```

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
ulysses-you: good point, use materialized list size instead of the required size

turboFei (approved):

ulysses-you:

PR #2573, opened 2022-05-06
[KPIP-4][SUB-TASK] Add a seekable buffered reader for random access operation log

ulysses-you:

A util class for random access operation log

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/SeekableBufferedReader.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/SeekableBufferedReader.scala
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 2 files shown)


ulysses-you:
cc @turboFei @yaooqinn this is a pre-work for rest get log

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/SeekableBufferedReader.scala
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * A seekable buffer reader, if the previous file path lines do not satisfy the limit size
+ * the reader will fetch line from next file path.
+ *
+ * Note that, call side should always new instance if want to replay the lines which has been
+ * fetched.
+ */
+class SeekableBufferedReader(paths: Seq[Path])
+  extends Closeable {
```
turboFei, line 32: nit: seems that new line is not needed

turboFei (approved):

ulysses-you:
thanks, merging to master

Issue #2553, opened 2022-05-05
[Subtask] Kyuubi Spark TPC-DS Connector - Add tiny scale


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

Current available scales: 0, 1, 10, 100, 300, 1000, 3000, 10000, 30000, 100000.

Event scale 1G is a little bit large for demo on laptop, we can introduce a scale 0.01 and name the database as `tiny` for demo purpose.

### Parent issue

https://github.com/apache/incubator-kyuubi/issues/2538

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yihua:
I’m interested in taking this up.

pan3793:
Assigned, thx.

PR #2576, opened 2022-05-06
Bump Hudi 0.11.0

cxzl25:
https://hudi.apache.org/releases/release-0.11.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,7 +126,6 @@
         <hive.archive.mirror>${apache.archive.dist}/hive/hive-${hive.engine.hive.version}</hive.archive.mirror>
         <hive.archive.download.skip>false</hive.archive.download.skip>
         <hudi.version>0.11.0</hudi.version>
-        <hudi.name>hudi-spark3.2-bundle_${scala.binary.version}</hudi.name>
         <iceberg.name>iceberg-spark-runtime-3.2_${scala.binary.version}</iceberg.name>
         <iceberg.version>0.13.1</iceberg.version>
         <jackson.version>2.13.1</jackson.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -964,27 +964,6 @@
...
-
-              - org.apache.hudi:hudi-common:jar:0.10.0:test
-                - org.apache.hbase:hbase-server:jar:1.2.3:test
-                  - org.apache.hadoop:hadoop-common:jar:2.5.1:test
-                    - org.apache.hadoop:hadoop-annotations:jar:2.5.1:test
-                      - jdk.tools:jdk.tools:jar:1.6:system
-            -->
-            <dependency>
-                <groupId>org.apache.hadoop</groupId>
-                <artifactId>hadoop-common</artifactId>
```
cxzl25, line 984: [HUDI-1180](https://issues.apache.org/jira/browse/HUDI-1180) Upgrade HBase from 1.2.3 to 2.4.9.
[HBASE-18640](https://issues.apache.org/jira/browse/HBASE-18640) Fixes leaks `jdk.tools` dependency in version 2.x.
So I try to remove `hadoop-common` dependency.

pan3793: Hudi 0.11 has a different module dependency graph, the exclusion may need to update. 

BTW, seems that Hudi 0.11 reduced the bundle jar dependencies, but I failed to pass the ut after switching to bundle jar w/ spark 3.2 in my previous try.

cxzl25: I first tried to pass the existing UT based on Spark3.1, so I changed hudi-spark3 to hudi-spark3.1.x, because hudi-spark3 uses spark3.2.

The next step is to plan to support hudi integration test in Spark 3.2.

I'll check for redundant exclusions later.

pom.xml
```diff
@@ -963,138 +964,10 @@
...
-
-            <dependency>
-                <groupId>org.apache.spark</groupId>
-                <artifactId>spark-avro_${scala.binary.version}</artifactId>
-                <version>${spark.version}</version>
-            </dependency>
-
-            <dependency>
-                <groupId>org.apache.hudi</groupId>
-                <artifactId>hudi-spark-common_${scala.binary.version}</artifactId>
```
cxzl25, line 1002: Because the `nodePatterns` field is added to the `Command` of Spark3.2.0 ([SPARK-35544](https://issues.apache.org/jira/browse/SPARK-35544)) , and `hudi-spark-common_2.12` is compiled with the default `Spark2.4.4`, there will be an `AbstractMethodError` problem here.

## How to reproduce this problem
```xml
        <dependency>
            <groupId>org.apache.spark</groupId>
            <artifactId>spark-sql_2.12</artifactId>
            <version>3.2.1</version>
        </dependency>

        <dependency>
            <groupId>org.apache.hudi</groupId>
            <artifactId>hudi-spark-common_2.12</artifactId>
            <version>0.11.0</version>
        </dependency>
```
```scala
DropHoodieTableCommand(TableIdentifier("", Option("")), false, false, false)
```
```
Exception in thread "main" java.lang.AbstractMethodError: Method org/apache/spark/sql/hudi/command/DropHoodieTableCommand.org$apache$spark$sql$catalyst$plans$logical$Command$_setter_$nodePatterns_$eq(Lscala/collection/Seq;)V is abstract
	at org.apache.spark.sql.hudi.command.DropHoodieTableCommand.org$apache$spark$sql$catalyst$plans$logical$Command$_setter_$nodePatterns_$eq(DropHoodieTableCommand.scala)
	at org.apache.spark.sql.catalyst.plans.logical.Command.$init$(Command.scala:38)
	at org.apache.spark.sql.hudi.command.DropHoodieTableCommand.<init>(DropHoodieTableCommand.scala:38)
```
## Use bundle
If we use spark bundle, we don't have this problem.

```xml
        <dependency>
            <groupId>org.apache.spark</groupId>
            <artifactId>spark-sql_2.12</artifactId>
            <version>3.2.1</version>
        </dependency>

        <dependency>
            <groupId>org.apache.hudi</groupId>
            <artifactId>hudi-spark3.2-bundle_2.12</artifactId>
            <version>0.11.0</version>
        </dependency>
```

pan3793:
Thanks, merging to master

yaooqinn:
seems 10-30% test time increased, since this merged

cxzl25:
> seems 10-30% test time increased, since this merged

Hudi's test shouldn't take long, let me investigate.

cxzl25:
> seems 10-30% test time increased, since this merged

After triggering GA N times, I have the following data.

## Build  \5ae7c9c\  2022/4/24

image


## Build \5b788a13\ 2022/5/7

image

## Build \cb5f49e3\ Bump Hudi 0.11.0  2022/05/11

image

## Master

image


The test time after upgrading Hudi version is close to the same as Master.
Then I analyzed the slowest suites that started to rise before 2022/5/7 and the slowest master suites.


image

This is a rough statistic. I just grabbed the log and calculated the interval between the output times of the two Suites.

Because `scalatest-maven-plugin` does not support slowpokeParams configuration, otherwise we can find slower UT through configuration.

https://www.scalatest.org/user_guide/using_scalatest_with_sbt
Slowpoke notifications
[[SPARK-33674](https://issues.apache.org/jira/browse/SPARK-33674)][TEST] Show Slowpoke notifications in SBT tests

https://www.scalatest.org/user_guide/using_the_scalatest_maven_plugin

yaooqinn:
better to separate tpcds tests

yaooqinn:
cc @pan3793

pan3793:
https://github.com/apache/incubator-kyuubi/issues/2553 would help to reduce the time significantly, @cxzl25 do you interested in it?

cxzl25:
> #2553 would help to reduce the time significantly, @cxzl25 do you interested in it?

OK, I can refactor the scale type, now it is int and needs to be changed to double.

Issue #2375, opened 2022-04-15
[SUB-TASK][KPIP-4] Implement BatchesResource GET /batches 


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

#2240

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

turboFei:
@zhaomin1423  would you like to take this at first?

zhaomin1423:
sure, thank you.



| |
***@***.***
|
|
***@***.***
|




---- 回复的原邮件 ----
| 发件人 | Fei ***@***.***> |
| 日期 | 2022年04月16日 11:34 |
| 收件人 | ***@***.***> |
| 抄送至 | Xiao ***@***.******@***.***> |
| 主题 | Re: [apache/incubator-kyuubi] [SUB-TASK][KPIP-4] Implement BatchesResource GET /batches (Issue #2375) |

@zhaomin1423 would you like to take this at first?

PR #2583, opened 2022-05-07
[KYUUBI #2375][FOLLOWUP] Implement BatchesResource GET /batches

zhaomin1423:

To return all types of batch when empty batch type specified.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
@@ -156,12 +156,17 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
   }
 
   def getBatchSessionList(batchType: String, from: Int, size: Int): Seq[Session] = {
-    allSessions().filter {
-      case batchSession: KyuubiBatchSessionImpl if batchType != null =>
-        batchSession.batchJobSubmissionOp.batchType.equalsIgnoreCase(batchType)
-      case _: KyuubiBatchSessionImpl => true
-      case _ => false
-    }.toSeq.sortBy(_.handle.identifier.toString).slice(from, from + size)
+    val sessions =
+      if (batchType == null) {
+        allSessions().filter(_.isInstanceOf[KyuubiBatchSessionImpl])
+      } else {
+        allSessions().filter {
+          case batchSession: KyuubiBatchSessionImpl =>
+            batchSession.batchJobSubmissionOp.batchType.equalsIgnoreCase(batchType)
+          case _ => false
+        }
+      }
+    sessions.toSeq.sortBy(_.handle.identifier.toString).slice(from, from + size)
   }
 
   override def start(): Unit = synchronized {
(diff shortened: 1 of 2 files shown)


zhaomin1423:
please help me to review it. @turboFei @ulysses-you

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -157,8 +157,9 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
 
   def getBatchSessionList(batchType: String, from: Int, size: Int): Seq[Session] = {
     allSessions().filter {
-      case batchSession: KyuubiBatchSessionImpl =>
+      case batchSession: KyuubiBatchSessionImpl if batchType != null =>
         batchSession.batchJobSubmissionOp.batchType.equalsIgnoreCase(batchType)
+      case _: KyuubiBatchSessionImpl => true
       case _ => false
     }.toSeq.sortBy(_.handle.identifier.toString).slice(from, from + size)
```
ulysses-you, line 164: we can pull out the condition:

```
val sessions = if (isnull) all else ...
session.toSeq.sortBy..
```

turboFei: > we can pull out the condition:
> 
> ```
> val sessions = if (isnull) all else ...
> session.toSeq.sortBy..
> ```

+1

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -157,8 +157,9 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
 
   def getBatchSessionList(batchType: String, from: Int, size: Int): Seq[Session] = {
     allSessions().filter {
-      case batchSession: KyuubiBatchSessionImpl =>
+      case batchSession: KyuubiBatchSessionImpl if batchType != null =>
```
turboFei, line 160: by the way, I am not sure if no specified batchType, its default values is `null` or empty string?

```
Option(batchType).exists(_.nonEmpty)
```

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
zhaomin1423: ignore from and size?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
zhaomin1423: it is null

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
zhaomin1423: updated

ulysses-you (approved):

turboFei (approved):

turboFei:
merging to master, thanks

Issue #2596, opened 2022-05-08
[FEATURE] Add RangerSparkCatalogShim to allow getDatabases/getTables/getColumns to support ranger permission control


zhangrenhua:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add RangerSparkCatalogShim to allow getDatabases/getTables/getColumns to support ranger permission control.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

bowenliang123:
It sounds good.

PR #2597, opened 2022-05-08
[KYUUBI #2596] Support to control the viewing permissions of databases/tables/views/columns

zhangrenhua:

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleReplaceShowObjectCommands.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleReplaceShowObjectCommands.scala
@@ -17,21 +17,90 @@
 
 package org.apache.kyuubi.plugin.spark.authz.ranger
 
+import scala.collection.mutable.ArrayBuffer
+
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.spark.sql.{Row, SparkSession}
+import org.apache.spark.sql.catalyst.analysis.ResolvedNamespace
 import org.apache.spark.sql.catalyst.expressions.Attribute
-import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, ShowNamespaces}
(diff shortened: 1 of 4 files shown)


extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -261,6 +261,19 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
```
yaooqinn, line 264: Not a good experience to introduce spark sql engine module here

zhangrenhua: I didn't find the general module of spark-sql-engine, can you provide ideas?

zhangrenhua: I should write a ranger-based OperationManager, thanks!

zhangrenhua: In order to avoid too much repetitive code and support more subsequent extensions, I think it is necessary to write the kyuubi-spark-sql-common module to store abstract interfaces and tool classes.

yaooqinn: we can format arguments to sql command, then run with spark.sql("command here"). with the result it produces, which is filtered ahead by auth module, we call catalog APIs to get detailed information

zhangrenhua: OK, I will adjust according to this idea.

zhangrenhua: If a large number of tables are queried, the performance loss of "SHOW COLUMNS" will be severe.

yaooqinn: this case is very rare, most meta APIs are called lazily, that is, when getColums is called, the `catalog/schema/table` part is specific

zhangrenhua: This situation is not uncommon. If we regard kyuubi as a database, does the application generally obtain table information through jdbc meta?

zhangrenhua:
Code submitted

extensions/spark/kyuubi-spark-authz/pom.xml
zhangrenhua: sorry, I resubmitted a pr (https://github.com/apache/incubator-kyuubi/pull/2648), the kyuubi code update frequency is quite high, praise!

PR #2609, opened 2022-05-09
Set Kyuubi server thrift client socket timeout to inf

pan3793:

We can detect engine dead cases after https://github.com/apache/incubator-kyuubi/pull/2122, so set thrift client socket timeout to inf to avoid socket timeout in blocking case, e.g. retrieve data except to 1st partition in increment collection mode, spark scale mode.


yaooqinn (approved):

turboFei:
update AllKyuubiConfigurationSuite

turboFei (approved):

Issue #2614, opened 2022-05-10
[Bug] Missing classes for beeline after upgrading Hive 3.1


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

```
➜  apache-kyuubi git:(master) beeline
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/commons/io/output/StringBuilderWriter
	at org.apache.hive.beeline.BeeLine.<init>(BeeLine.java:177)
	at org.apache.hive.beeline.KyuubiBeeLine.<init>(KyuubiBeeLine.java:44)
	at org.apache.hive.beeline.KyuubiBeeLine.<init>(KyuubiBeeLine.java:40)
	at org.apache.hive.beeline.KyuubiBeeLine.mainWithInputRedirection(KyuubiBeeLine.java:80)
	at org.apache.hive.beeline.KyuubiBeeLine.main(KyuubiBeeLine.java:67)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.io.output.StringBuilderWriter
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:419)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:352)
	... 5 more
```

### Affects Version(s)

master

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pan3793:
cc @gabrywu

gabrywu:
> cc @gabrywu

I will try to fix it, and don't upgrade beeline yet

PR #2618, opened 2022-05-10
[KYUUBI #2614] Add commons-io to beeline module since jdbc upgraded to 3.1.3

gabrywu:

close https://github.com/apache/incubator-kyuubi/issues/2614

add commons-io to beeline module since jdbc upgraded to 3.1.3

using existing UTs to test it

### DIFF
--- a/kyuubi-hive-beeline/pom.xml
+++ b/kyuubi-hive-beeline/pom.xml
@@ -79,6 +79,17 @@
         </dependency>
 
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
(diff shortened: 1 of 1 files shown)


gabrywu:
cc @pan3793

kyuubi-hive-beeline/pom.xml
```diff
@@ -78,6 +78,17 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <exclusions>
```
pan3793, line 84: The `commons-io` does not has transitive dependencies, we don't need this exclusion

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #2621, opened 2022-05-10
Always use Hadoop shaded client

pan3793:

There are some deps has low version hadoop deps, we should exclude them and use hadoop shaded client instead.

For engine modules, the shaded client will not be packaged into shaded jar, then nothing changes in the final shaded jar.


yaooqinn (approved):

pan3793:
Thanks, merging to master

PR #2623, opened 2022-05-10
[KYUUBI #2438] Ranger AccessResource get database from spark catalog when database isnull

### LINKED ISSUE #2438: [Improvement] Shall We Add a SparkSession In AccessResource?
### Code of Conduct

### Search before asking

The current constructor of AccessResource is :
```scala
object AccessResource {
  def apply(
      objectType: ObjectType,
      firstLevelResource: String,
      secondLevelResource: String,
      thirdLevelResource: String) {
    ...
  }
}
```
In some cases, firstLevelResource maybe null, for example, in spark 2.4,`DESCRIBE TABLE table_name`, will produce a `DescribeTableCommand` without database info, so the correspnding AccessResource's `firstLevelResource` is null. Then it will generate a invaild RangerRequest, because of the databasse field is null.

I'm not sure that when the database is not explicitly specified in the sql statement, there will be other commands without database information.

### How should we improve?

Therefore, I think we should add a sparksession in AccessResource, in the case of databasse is null. When the `firstLevelResource` is null or empty, get `firstLevelResource` from catalog's current database.

packyan:

To close #2438.
Since there are some commands without database information, when `firstLevelResource` is null or empty, it's reasonable to fallback to using the current db as `firstLevelResource`.


extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AccessResource.scala
```diff
@@ -40,11 +41,20 @@ object AccessResource {
...
       secondLevelResource: String,
-      thirdLevelResource: String): AccessResource = {
+      thirdLevelResource: String,
+      spark: SparkSession = null): AccessResource = {
     val resource = new AccessResource(objectType)
 
+    val firstLevelResource =
+      if (spark != null
+        && (firstLevelResourceOrNull == null || firstLevelResourceOrNull.isEmpty)) {
+        spark.catalog.currentDatabase
```
yaooqinn, line 53: why session catalog？how about other catalog implementations

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AccessResourceSuite.scala
```diff
@@ -57,3 +59,95 @@ class AccessResourceSuite extends AnyFunSuite {
...
+        val operationType = OperationType(plan.nodeName)
+        val tuple = PrivilegesBuilder.build(plan)
+
+        val resourceWithoutSparkSession = AccessResource(tuple._1.head, operationType)
+        assert(resourceWithoutSparkSession.objectType === TABLE)
+        assert(resourceWithoutSparkSession.getDatabase === dbName)
+        assert(resourceWithoutSparkSession.getTable equalsIgnoreCase tableName)
+
+        val resourceWithSparkSession = AccessResource(tuple._1.head, operationType, spark = spark)
+        assert(resourceWithSparkSession.objectType === TABLE)
```
yaooqinn, line 134: try view without database？

yaooqinn: desc ${viewName}

yaooqinn:
I still think we shall handle the commands where the database is optional, and set currentDb accordingly, instead of setting it blindly in AccessResource

packyan:
> 

Ok,  if we can handle the commands where the database is optional, it’s definitely better than setting it blindly in AccessResource. My solution may be too violent.

Issue #2631, opened 2022-05-11
[Improvement]  Rename high availability config key to support muti discovery client


hddong:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Rename high availability config key to support muti discovery client

### How should we improve?

To support muti discovery client

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2633, opened 2022-05-11
[KYUUBI #2631] Rename high availability config key to support multi discovery client

hddong:

Rename high availability config key to support muti discovery client


hddong:
cc @yaooqinn @pan3793 
changes:
`kyuubi.ha.zookeeper.namespace` -> `kyuubi.ha.discovery.deploy.namespace`
`kyuubi.ha.zookeeper.quorum` -> `kyuubi.ha.discovery.endpoints`
`kyuubi.ha.service.discovery.client.class` -> `kyuubi.ha.discovery.client.class`

yaooqinn:
kyuubi.ha.zookeeper.namespace -> kyuubi.ha.namespace
kyuubi.ha.zookeeper.quorum -> kyuubi.ha.endpoints
kyuubi.ha.service.discovery.client.class -> kyuubi.ha.client.class

pan3793:
WDYT the following style?
```
kyuubi.ha.type=<zookeeper|etcd|custom>

kyuubi.ha.zookeeper.k1=
kyuubi.ha.zookeeper.k2=
kyuubi.ha.etcd.k1=
kyuubi.ha.etcd.k2=
kyuubi.ha.custom.class=
kyuubi.ha.custom.k1=
kyuubi.ha.custom.k2=
```

hddong:
> kyuubi.ha.zookeeper.namespace -> kyuubi.ha.namespace
> kyuubi.ha.zookeeper.quorum -> kyuubi.ha.addresses
>kyuubi.ha.service.discovery.client.class -> kyuubi.ha.client.class

Changed

>`kyuubi.ha.type=<zookeeper|etcd|custom>

This change had put in the next PR(supoort etcd) and will push resently, we can unify the name of some common configs first.

yaooqinn:
is kyuubi.ha.type duplicated with kyuubi.ha.client.class?

hddong:
> is kyuubi.ha.type duplicated with kyuubi.ha.client.class?

It wil used when type is custom, and be option

yaooqinn:
> > is kyuubi.ha.type duplicated with kyuubi.ha.client.class?
> 
> It will be used when type is custom, and be option

if the class is provided, why do we need an extra type to specify it?

hddong:
> if the class is provided, why do we need an extra type to specify it?

IMO, in project ,we always use type to choice built in classes (such as: auth, evnets). 
https://github.com/apache/incubator-kyuubi/blob/33c816240d3c1eee744bad351c5ae25c54f64139/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala#L56-L59
And in `KyuubiServer`, we need `create chroot path if necessary` when ues `ZK`. `ha.type` match may better than `client.class` equals to check if discovery client is zk.

yaooqinn:
> > if the class is provided, why do we need an extra type to specify it?
> 
> IMO, in project ,we always use type to choice built in classes (such as: auth, evnets).
> 
> https://github.com/apache/incubator-kyuubi/blob/33c816240d3c1eee744bad351c5ae25c54f64139/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala#L56-L59
> 
> 
> And in `KyuubiServer`, we need `create chroot path if necessary` when ues `ZK`. `ha.type` match may better than `client.class` equals to check if discovery client is zk.

the code path your provide, I don't know very well
but the client.class is duplicated with type

hddong:
> the code path your provide, I don't know very well but the client.class is duplicated with type

How about
```
kyuubi.ha. type=<zk|etcd|custom>
kyuubi.ha.custom.class=Optional[String]
```

yaooqinn:
why is zk so special?

yaooqinn:
check the code in kyuubi server again, it looks like it can be removed(it's a quite rare and odd case) or share with all discovery clients

hddong:
> check the code in kyuubi server again, it looks like it can be removed(it's a quite rare and odd case) or share with all discovery clients

Ok, will remove it when support etcd.

yaooqinn:
less is more, I prefer to remain less configurations

hddong:
> less is more, I prefer to remain less configurations

Had revert it.
now:
```scala
kyuubi.ha.namespace
kyuubi.ha.addresses
kyuubi.ha.client.class
```

yaooqinn (approved):

hddong:
@yaooqinn @pan3793 thanks for review, merging to master

PR #2641, opened 2022-05-12
Client should not assume launch engine has completed on exception

pan3793:

`LaunchEngineOp` will never close until session is closed, the client should not assume launch engine has completed on exception.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/Operation.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/Operation.scala
@@ -49,4 +49,7 @@ trait Operation {
 object Operation {
   val DEFAULT_FETCH_ORIENTATION_SET: Set[FetchOrientation] =
     Set(FetchOrientation.FETCH_NEXT, FetchOrientation.FETCH_FIRST, FetchOrientation.FETCH_PRIOR)
+
+  val LAUNCH_ENGINE_GUID = "kyuubi.session.engine.launch.handle.guid"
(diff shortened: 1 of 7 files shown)


pan3793:
cc @gabrywu

turboFei (approved):

yaooqinn:
shall we test this?

gabrywu (approved):
+1

pan3793:
> shall we test this?

Actually, we have a test for this case, and I enhance it in this PR

https://github.com/apache/incubator-kyuubi/blob/8de2f5f17fa909979294a879503700bd3b083df0/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala#L116-L130

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/Operation.scala
```diff
@@ -49,4 +49,7 @@ trait Operation {
 object Operation {
   val DEFAULT_FETCH_ORIENTATION_SET: Set[FetchOrientation] =
     Set(FetchOrientation.FETCH_NEXT, FetchOrientation.FETCH_FIRST, FetchOrientation.FETCH_PRIOR)
+
+  val LAUNCH_ENGINE_GUID = "kyuubi.session.engine.launch.handle.guid"
```
yaooqinn, line 53: what about add them to KyuubiReservedKeys

yaooqinn (approved):

PR #2647, opened 2022-05-13
[SUB-TASK][KPIP-4] Unify the conf between Session and Batch

ulysses-you:

closes https://github.com/apache/incubator-kyuubi/issues/2646

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
@@ -36,7 +36,7 @@ class SparkBatchProcessBuilder(
 
   override def mainResource: Option[String] = Option(batchRequest.resource)
 
-  override protected val commands: Array[String] = {
+  override val commands: Array[String] = {
     val buffer = new ArrayBuffer[String]()
     buffer += executable
     Option(mainClass).foreach { cla =>
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
@@ -53,8 +53,11 @@ class SparkBatchProcessBuilder(
       buffer += s"$k=$v"
     }
 
-    buffer += PROXY_USER
-    buffer += proxyUser
+    // iff the keytab is specified, PROXY_USER is not supported
+    if (!useKeytab()) {
+      buffer += PROXY_USER
+      buffer += proxyUser
+    }
 
     assert(mainResource.isDefined)
     buffer += mainResource.get
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
@@ -16,6 +16,8 @@
  */
 package org.apache.kyuubi.session
 
+import scala.collection.JavaConverters._
+
 import org.apache.hive.service.rpc.thrift.TProtocolVersion
 
 import org.apache.kyuubi.events.KyuubiSessionEvent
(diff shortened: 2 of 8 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
```diff
@@ -29,6 +31,17 @@ abstract class KyuubiSession(
     sessionManager: KyuubiSessionManager)
   extends AbstractSession(protocol, user, password, ipAddress, conf, sessionManager) {
 
-  def getSessionEvent: Option[KyuubiSessionEvent]
+  private[kyuubi] lazy val optimizedConf: Map[String, String] = {
+    val confOverlay = sessionManager.sessionConfAdvisor.getConfOverlay(
```
turboFei, line 35: In my opinion, maybe we need a new method for batch conf overlay.

turboFei: by defaults, the batch conf overlay could be same with the original one, but it is better to have two seprate methods.

turboFei:
```
  /** The returned conf will overwrite the batch session conf */
  default Map<String, String> getBatchConfOverlay(String user, Map<String, String> sessionConf) {
    return getConfOverlay(user, sessionConf);
  }
```

zhaomin1423: +1. there may be some differences.

ulysses-you: yeah, I considered this. I do not against add a new method for batch. cc @wForget @yaooqinn what do you think ?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
```diff
@@ -53,8 +53,11 @@ class SparkBatchProcessBuilder(
       buffer += s"$k=$v"
     }
 
-    buffer += PROXY_USER
-    buffer += proxyUser
+    // iff the keytab is specified, PROXY_USER is not supported
```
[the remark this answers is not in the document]
ulysses-you: I copied from `SparkProcessBuilder`, it should mean `ony if` ..

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
yaooqinn: I am not sure it's a good idea. if so, we shall add APIs for different resource managers, engines, workloads, etc.

ulysses-you: How about inject the type of request to the `sessionConf` ? For example: `kyuubi.session.type -> batch` then the plugin can catch the key and do something.

ulysses-you: The interface of `sessionConf` is flexible, user can inject any configs by theirself and catch it inside plugin. So we can leave it as is.

turboFei: > The interface of sessionConf is flexible, user can inject any configs by theirself and catch it inside plugin. So we can leave it as is.

Agree with that.
The sessionConfAdvisor should be developed by kyuubi admin and they can inject some reserved key words for the sessionConfAdvisor.

We can add some comments in the interface.

ulysses-you: sure, enriched the interface comment

kyuubi-server/src/test/scala/org/apache/kyuubi/session/KyuubiBatchSessionSuite.scala
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package org.apache.kyuubi.session
+
+import org.apache.kyuubi.{Utils, WithKyuubiServer}
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.engine.spark.SparkBatchProcessBuilder
+import org.apache.kyuubi.server.api.v1.{BatchesResource, BatchRequest}
+
+class KyuubiBatchSessionSuite extends WithKyuubiServer {
```
yaooqinn, line 25: use e2e tests in BatchesResourceSuite?

ulysses-you: this test is for checking the internal config propagation between request -> session -> builder. e2e test can not catch the details of each class, so the unit test should be good ?

yaooqinn: so why do we need a WithKyuubiServer injection?

yaooqinn: for example, the tests you added behave complete different in batch or SQL sessions. For SQL, k1/k2 will be sent to SparkSQLEngine with the opening session, but for batch, there are spark.k1/spark.v2 only

ulysses-you: Without `WithKyuubiServer`, the event framwork will fail since the kyuubi server is not initialized..

I added some more test for checking the commands result.

PR #2661, opened 2022-05-16
[SUB-TASK][KPIP-4] Rename GET /batches/$batchId/log to GET /batches/$batchId/localLog

turboFei:

Now the log got is local log, refactor the api to make it more clear.


yaooqinn (approved):

ulysses-you (approved):
we should also update the docs

turboFei:
> we should also update the docs

thanks, updated

turboFei:
thanks, merged to master

Issue #2285, opened 2022-04-06
[Improvement] Trino engine will save the result set all in memory, even if the incremental collect mode selected


iodone:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

The advantage of the trino engine is the ability to stream returned data, kyuubi as a proxy should not remove this feature, storing the result set all in memory will lead to performance and OOM problems. 
https://github.com/apache/incubator-kyuubi/blob/6a231519454c5d75aca2120de140c06febf7ae54/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala#L81-L135

The iterator in this code is a direct conversion of ArrayBuffer, not a real stream iterator, and in each call to getNextResultSet, there is no real call to `Trino.client `to get the next result each time. @hddong WDYT?

### How should we improve?

1. Trino's proxy implementation should be based entirely on the incremental collect model.
2. Based on `IterableFetchIterator` to implement Trino DataResult Iterator, we can refer to Trino-JDBC's `ResultSet` implementation.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

yaooqinn:
I suggest we shall always not hold data at engine side in any mode

PR #2678, opened 2022-05-17
[KYUUBI #2285] trino's result fetching method is changed to a streaming iterator mode to avoid hold data at server side

iodone:

1. Change to a true iterator implementation to get the results of Trino Server to avoid caching data in memory.
2. Supports FETCH_NEXT mode only, consistent with Trino Server's capabilities

### DIFF
--- a/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
+++ b/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
@@ -17,16 +17,11 @@
 
 package org.apache.kyuubi.engine.trino
 
-import java.util.ArrayList
-import java.util.concurrent.ArrayBlockingQueue
 import java.util.concurrent.Executors
 
+import scala.annotation.tailrec
 import scala.collection.JavaConverters._
-import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
-import scala.concurrent.duration
-import scala.concurrent.duration.Duration
 
 import com.google.common.base.Verify
 import io.trino.client.ClientSession
@@ -38,7 +33,6 @@ import org.apache.kyuubi.KyuubiSQLException
 import org.apache.kyuubi.Logging
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.trino.TrinoConf.DATA_PROCESSING_POOL_SIZE
-import org.apache.kyuubi.engine.trino.TrinoStatement._
 
 /**
  * Trino client communicate with trino cluster.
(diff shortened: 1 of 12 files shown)


pan3793:
> Supports FETCH_NEXT mode only, consistent with Trino Server's capabilities

IIRC, it may break HUE export data functionality. Can we support both?

iodone:
> > Supports FETCH_NEXT mode only, consistent with Trino Server's capabilities
> 
> IIRC, it may break HUE export data functionality. Can we support both?

IMO, trino is designed to return results based on the pipeline model and does not have the capability of other fetch order types. If we want to support it, need to cache the trino results on the Server side, but Kyuubi Server is not suitable for caching results as the Gateway, because it may lead to OOM.

image
https://trino.io/docs/current/sql/select.html?highlight=fetch#limit-or-fetch-first-clause 

It was found that Trino SQL originally supported these order modes.

pan3793:
I think we can give a option to cache the result on engine side, it will not affect Kyuubi server even if engine crash by OOM. Another idea is re-run query if the client request the fetch FIRST after consuming, it's usually for export cases.

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/FetchIterator.scala
```diff
@@ -109,3 +109,26 @@ class IterableFetchIterator[A](iterable: Iterable[A]) extends FetchIterator[A] {
     }
   }
 }
+
+object FetchIterator {
+
+  def from[A](iter: Iterator[A]): FetchIterator[A] = new FetchIterator[A] {
```
yaooqinn, line 115: is this so common to have this name?

yaooqinn (approved):
LGTM, only one comment

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/FetchIterator.scala
iodone: OK, I modified it to `fromIterator`, just like the `Future.fromTry` API

yaooqinn: shouldn't it be as same as IterableFetchIterator?

hddong:
> > I think we can give an option to cache the result on engine side, it will not affect Kyuubi server even if engine crash by OOM. Another idea is re-run query if the client request the fetch FIRST after consuming, it's usually for export cases.
> 
> Can we consider implementing FETCH_FIRST in KyuybiJDBC Client, so that keep the immediate result return feature of trino?

Or we just Keep the previous `execute(): Iterable[List[Any]]` as `executeCollect(): Iterable[List[Any]]` , and run:
```
iter =
    if (incrementalCollect) {
      info("Execute in incremental collect mode")
      FetchIterator.fromIterator(trinoStatement.execute())
    } else {
      info("Execute in full collect mode")
      new ArrayFetchIterator(trinoStatement.executeCollect())
    }
```

pan3793:
> Can we consider implementing FETCH_FIRST in KyuubiJDBC Client, so that keep the immediate result return feature of trino?

There are many language client for Kyuubi, I prefer to implement this feature in engine, like Spark full collect mode and incremental collect mode

iodone:
> > > I think we can give an option to cache the result on engine side, it will not affect Kyuubi server even if engine crash by OOM. Another idea is re-run query if the client request the fetch FIRST after consuming, it's usually for export cases.
> > 
> > 
> > Can we consider implementing FETCH_FIRST in KyuybiJDBC Client, so that keep the immediate result return feature of trino?
> 
> Or we just Keep the previous `execute(): Iterable[List[Any]]` as `executeCollect(): Iterable[List[Any]]` , and run:
> 
> ```
> iter =
>     if (incrementalCollect) {
>       info("Execute in incremental collect mode")
>       FetchIterator.fromIterator(trinoStatement.execute())
>     } else {
>       info("Execute in full collect mode")
>       new ArrayFetchIterator(trinoStatement.executeCollect())
>     }
> ```

Good advice, I'll try it, thanks

externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
```diff
@@ -80,63 +74,34 @@ class TrinoStatement(
...
-        if (data != null) {
-          data.asScala.map(_.asScala.toList)
-            .foreach(e => putOrThrow(rowQueue, e))
+  def execute(): Iterator[List[Any]] = {
+    Iterator.continually {
+      @tailrec
+      def getData(): (Boolean, List[List[Any]]) = {
+        if (trino.isRunning) {
+          val data = trino.currentData().getData()
+          trino.advance()
```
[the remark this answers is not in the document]
iodone: `trino.advance` must be called each time to ensure that the results obtained each time is the latest, if put into 87line will lead to keep getting the same data

hddong (approved):

externals/kyuubi-trino-engine/src/test/scala/org/apache/kyuubi/engine/trino/TrinoStatementSuite.scala
```diff
@@ -52,16 +52,16 @@ class TrinoStatementSuite extends WithTrinoContainerServer {
       assert(this.schema === trinoStatement.getCurrentDatabase)
 
       val trinoStatement2 = TrinoStatement(trinoContext, kyuubiConf, "use sf1")
-      trinoStatement2.execute()
-
+      // if trinoStatement.execute return iterator is lazy, call toArray to strict evaluation
```
SteNicholas, line 55: How does the end user knows whether `trinoStatement.execute` is lazy?

externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/TrinoStatement.scala
```diff
@@ -80,63 +74,34 @@ class TrinoStatement(
...
+            (true, data.asScala.toList.map(_.asScala.toList))
+          } else {
+            getData()
+          }
+        } else {
+          Verify.verify(trino.isFinished)
+          val finalStatus = trino.finalStatusInfo()
+          if (finalStatus.getError() != null) {
+            throw KyuubiSQLException(
+              s"Query ${finalStatus.getId} failed: ${finalStatus.getError.getMessage}")
```
SteNicholas, line 94: If the user invokes the `getQueryId` interface, is this query id same as the returned id of `getQueryId`?

iodone: yes，the `queryId` is generated by trinoServer at the time of the first query and is unique throughout the query process

yaooqinn:
Hi @hddong, ready to merge?

yaooqinn:
thanks, merged to master

PR #2685, opened 2022-05-18
[KYUUBI #2594] Fix flaky Test - support engine alive probe to fast fail on engine broken

### LINKED ISSUE #2594: Flaky Test: KyuubiOperationPerConnectionSuite - support engine alive probe to fast fail on engine broken
### Code of Conduct

### Search before asking

### Describe the bug

https://github.com/apache/incubator-kyuubi/runs/6334737100?check_suite_focus=true#step:5:1603

```
KyuubiOperationPerConnectionSuite:
- KYUUBI #2102 - support engine alive probe to fast fail on engine broken *** FAILED ***
  1253 was not greater than 3000 (KyuubiOperationPerConnectionSuite.scala:241)
```

### Affects Version(s)

1.6.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

turboFei:

To close #2594 

For the liveness probe, it marks the remote engine broken if the last alive time detection is timeout.

```
                if (now - engineLastAlive > engineAliveTimeout) {
                  error("Mark the engine not alive with no recent alive probe success:" +
                    s" ${now - engineLastAlive} ms exceeds timeout $engineAliveTimeout ms")
                  remoteEngineBroken = true
                }
```
But now the engineLastAlive variable is not volatile.
So we set the engineLastAlive as volatile variable in this pr to prevent flaky test.

And now we are using `scheduleAtFixedRate`, it will trigger the liveness detection in fixed rate and regardless whether the last one has completed.

For scheduleWithFixedDelay,
```
Creates and executes a periodic action that becomes enabled first after the given initial delay, and subsequently with the given delay between the termination of one execution and the commencement of the next. If any execution of the task encounters an exception, subsequent executions are suppressed. Otherwise, the task will only terminate via cancellation or termination of the executor.
```

For scheduleAtFixedRate,
```
Creates and executes a periodic action that becomes enabled first after the given initial delay, and subsequently with the given period; that is executions will commence after initialDelay then initialDelay+period, then initialDelay + 2 * period, and so on. If any execution of the task encounters an exception, subsequent executions are suppressed. Otherwise, the task will only terminate via cancellation or termination of the executor. If any execution of this task takes longer than its period, then subsequent executions may start late, but will not concurrently execute.
```

We should use scheduleWithFixedDelay, so that the liveness probe detections are in sequence.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
@@ -53,7 +53,7 @@ class KyuubiSyncThriftClient private (
   @volatile private var remoteEngineBroken: Boolean = false
   private val engineAliveProbeClient = engineAliveProbeProtocol.map(new TCLIService.Client(_))
   private var engineAliveThreadPool: ScheduledExecutorService = _
-  private var engineLastAlive: Long = _
+  @volatile private var engineLastAlive: Long = _
 
   private def startEngineAliveProbe(): Unit = {
     engineAliveThreadPool = ThreadUtils.newDaemonSingleThreadScheduledExecutor(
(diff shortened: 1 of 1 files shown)


turboFei:
image
GA passed, re-run again.

turboFei:
cc @wForget

wForget (approved):
LGTM

yaooqinn (approved):

turboFei:
thanks for your review, merging to master

PR #2691, opened 2022-05-18
[KYUUBI #2690] Make ProcessBuilder.commands immutable

### LINKED ISSUE #2690: [Bug] SparkProcessBuilder.commands is immutable causing tag not to take effect
### Code of Conduct

### Search before asking

### Describe the bug

When `SparkProcessBuilder` is constructed, the commands are generated according to `conf` and are immutable, so the tags will not take effect if they are set later.
This is introduced by #2632.

### Affects Version(s)

1.6.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

cxzl25:
close #2690

When `SparkProcessBuilder` is constructed, the commands are generated according to `conf` and are immutable, so the tags will not take effect if they are set later.

- Make several existing `ProcessBuilder` commands immutable.
- Set tags when building commands.
- Fix Flink tags may not be set to the right value.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
@@ -100,9 +100,9 @@ object KyuubiApplicationManager {
   }
 
   private def setupFlinkK8sTag(tag: String, conf: KyuubiConf): Unit = {
-    val originalTag = conf.getOption(FlinkProcessBuilder.APP_KEY).map(_ + ",")
+    val originalTag = conf.getOption(FlinkProcessBuilder.TAG_KEY).map(_ + ",")
     val newTag = s"${originalTag}KYUUBI" + Some(tag).map("," + _).getOrElse("")
-    conf.set(FlinkProcessBuilder.APP_KEY, newTag)
+    conf.set(FlinkProcessBuilder.TAG_KEY, newTag)
   }
 
   /**
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
@@ -91,7 +91,7 @@ class KyuubiApplicationManager extends AbstractService("KyuubiApplicationManager
 object KyuubiApplicationManager {
   private def setupSparkYarnTag(tag: String, conf: KyuubiConf): Unit = {
     val originalTag = conf.getOption(SparkProcessBuilder.TAG_KEY).map(_ + ",").getOrElse("")
-    val newTag = s"${originalTag}KYUUBI" + Some(tag).map("," + _).getOrElse("")
+    val newTag = s"${originalTag}KYUUBI" + Some(tag).filterNot(_.isEmpty).map("," + _).getOrElse("")
     conf.set(SparkProcessBuilder.TAG_KEY, newTag)
   }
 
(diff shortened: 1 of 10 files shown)


yaooqinn (approved):

turboFei (approved):

cxzl25:
> is it possible to resolve te `// TODO: Better to do this inside ProcBuilder` ? it's confused we update the conf outside the builder. So the result of commands can be immuable.

So can we do this?

Add the parameter `engineRefId` when constructing `ProcessBuilder`, and call `tagApplication` when `ProcessBuilder` build commands.

ulysses-you:
yeah I think we can, the engineRefId should be bound to the engine builder

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -98,10 +100,9 @@ object KyuubiApplicationManager {
   }
 
   private def setupFlinkK8sTag(tag: String, conf: KyuubiConf): Unit = {
-    // TODO: yarn.tags or flink.yarn.tags, the mess of flink settings confuses me now.
```
cxzl25, line 101: Check the Flink configuration documentation, there is no such configuration as `flink.yarn.tags`.

Although Flink supports converting `flink.yarn.xx` to `yarn.xx`, there is no `yarn.tags` configuration in yarn.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilder.scala
```diff
@@ -37,7 +37,8 @@ import org.apache.kyuubi.operation.log.OperationLog
 class FlinkProcessBuilder(
     override val proxyUser: String,
     override val conf: KyuubiConf,
-    val extraEngineLog: Option[OperationLog] = None)
+    val extraEngineLog: Option[OperationLog] = None,
+    val engineRefId: String = "")
```
ulysses-you, line 41: the default value is only for simplify test code ? if so, we can add a new constructor with comments for that to be clear

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilderSuite.scala
```diff
@@ -32,6 +32,7 @@ class FlinkProcessBuilderSuite extends KyuubiFunSuite {
     .set(
       ENGINE_FLINK_JAVA_OPTIONS,
       "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005")
+    .set("yarn.tags", "KYUUBI")
```
ulysses-you, line 35: why add this config ?

cxzl25: Because `FlinkProcessBuilde`r will write the `yarn.tags` configuration when generating commands, the previous test is broken.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -88,20 +90,19 @@ class KyuubiApplicationManager extends AbstractService("KyuubiApplicationManager
...
     conf.set("spark.kubernetes.driver.label.kyuubi_unique_tag", tag)
   }
 
   private def setupFlinkK8sTag(tag: String, conf: KyuubiConf): Unit = {
-    // TODO: yarn.tags or flink.yarn.tags, the mess of flink settings confuses me now.
-    val originalTag = conf.getOption("yarn.tags").map(_ + ",")
-    val newTag = s"${originalTag}KYUUBI,$tag"
-    conf.set("yarn.tags", newTag)
+    val originalTag = conf.getOption(FlinkProcessBuilder.TAG_KEY).map(_ + ",")
+    val newTag = s"${originalTag}KYUUBI" + Some(tag).map("," + _).getOrElse("")
```
ulysses-you, line 104: is it a long time bug ? the type of `originalTag` is a Option not a String ..

ulysses-you:
thanks, merging to master

PR #2697, opened 2022-05-18
[SUB-TASK][KPIP-4] Introduce state store interface and default InMemoryStateStore

turboFei:

Introduce the basic state store interface used for batches ha & recovery.
Refer the docs: https://docs.google.com/document/d/1_BMekMT88P2vCHCl4ZacWEeWslpIv4pxsJH32PAnWSI/edit#

And add a default InMemoryStateStore.

TODO:
- Introduce mysql based state store.
- Support batch recovery.
- Support to redirect the rest request to the kyuubi instance that created the batch.


turboFei:
will add a default in memory state store.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/statestore/SessionStateStore.scala
```diff
@@ -0,0 +1,133 @@
...
+      batchOwner: String,
+      conf: Map[String, String],
+      batchRequest: BatchRequest): Unit = {
+    val batchState = BatchState(
+      batchId,
+      batchRequest.getBatchType,
+      batchOwner,
+      conf,
+      restInstance,
+      "PENDING",
```
turboFei, line 68: waiting for #2699

turboFei: and shall update the state to `RUNNING` if applicationState got at the first time.

turboFei:
also cc @pan3793

turboFei:
I just know that derby can support memory db, so the InMemoryStaeStore is not needed, close this one.

PR #2699, opened 2022-05-18
Introduce batch application state and rename BatchJobSubmission to SubmitBatchApp

turboFei:

Introduce batch application submission state.

The initialized state is `PENDING`, when it gets the applicationStatus at first time, it should update the status to `RUNNIGN`.

And when the batch operation terminates, update the final state to `FINISHED` or `FAILED` or `KILLED`.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
@@ -20,5 +20,5 @@ package org.apache.kyuubi.operation
 object BatchAppSubmissionState extends Enumeration {
   type BatchAppSubmissionState = Value
 
-  val PENDING, RUNNING, FINISHED, FAILED, KILLED = Value
+  val PENDING, RUNNING, SUBMITTED, FAILED, KILLED = Value
 }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
@@ -17,8 +17,8 @@
 
 package org.apache.kyuubi.operation
 
-object BatchAppState extends Enumeration {
-  type BatchAppState = Value
+object BatchAppSubmissionState extends Enumeration {
+  type BatchAppSubmissionState = Value
 
   val SUBMITTING, RUNNING, FINISHED, FAILED, KILLED = Value
 }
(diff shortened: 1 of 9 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+object BatchAppSubmissionState extends Enumeration {
+  type BatchAppSubmissionState = Value
+
+  val SUBMITTING, RUNNING, FINISHED, FAILED, KILLED = Value
```
yaooqinn, line 23: SUBMITTING -> PENDING?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+object BatchAppSubmissionState extends Enumeration {
```
yaooqinn, line 20: ApplicationState?

turboFei: it is not absolute same with the state of ApplicationReport.

So, I am thinking how to name it.

turboFei: For example, the `BatchAppSubmissionState` is FINISHED and the application state is `RUNNING`.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
turboFei: updated

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchAppSubmissionState.scala
yaooqinn: we shall call it submitted...

yaooqinn:
OperationState seems can cover all scenarios of submission and app state

turboFei:
> OperationState seems can cover all scenarios of submission and app state

I see.

Issue #2100, opened 2022-03-11
[Umbrella] Flink Engine Improvement and Quality Assurance


yaooqinn:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the proposal

We introduced the Flink engine in https://github.com/apache/incubator-kyuubi/issues/1322.

In this ticket, we collect feedback, improvements, bugfixes, aim to make it production-ready 

### Improvement
- [ ] #2002 @link3280 
- [ ] #1652 @yanghua 
- [ ] #1865 @SteNicholas 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

SteNicholas:
@yaooqinn, the module label should be flink, not hive.

yaooqinn:
> @yaooqinn, the module label should be flink, not hive.

oops..

SteNicholas:
@yaooqinn, could you please help to PIN ths issue?

link3280:
@yaooqinn shall we make this a KPIP and let the corresponding issues follow the naming pattern like `[SUBTASK][KPIP-X]`?

yaooqinn:
I am not sure that we can propose a [KPIP](https://kyuubi.apache.org/improvement-proposals.html) on the status of this ticket, which seems not to meet the requirement of a KPIP.

In fact, we shall not create subtasks for KPIP-2 as it has been resolved. [SUBTASK][#2100] may be enough?

link3280:
@yaooqinn LGTM

pan3793:
Postpone to 1.8, because this feature is not under rapid development, and it's not supposed to be accomplished in a short time.

waywtdcc:
The jdbc interface supports asynchronous real-time tasks to obtain results. Can this be done?   @pan3793

pan3793:
@waywtdcc technically, I don't think there is any blocker in Kyuubi framework, the JDBC driver retrieve result from Kyuubi Server in mini-batch, and we do similar thing in Spark which called incremental collection.

So it could be true if the Flink engine can return the streaming data in an `Iterator`.

cc the Flink experts @SteNicholas @link3280 @yanghua

pan3793:
[...]

PR #2718, opened 2022-05-22
[KYUUBI#2405] Support Flink StringData Data Type

link3280:
Currently, Flink uses its legacy data type system in CollectSink, but sooner would move to the new type system (see https://issues.apache.org/jira/browse/FLINK-12251). Kyuubi should adapt to the new data type system beforehand.

This PR supports StringData in Flink.

This is a subtask of #2100 .

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
@@ -174,7 +174,7 @@ object RowSet {
     // if there's zero row, set empty as the column
     var sampleField: Object = null
     var i = -1
-    while (sampleField == null && i < rows.length) {
+    while (sampleField == null && rows.nonEmpty && i < rows.length) {
       i += 1
       sampleField = rows(i).getField(ordinal)
     }
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
@@ -173,7 +173,7 @@ object RowSet {
     // if there's zero row, set empty as the column
     var sampleField: Object = null
     var i = -1
-    while (sampleField == null && rows.nonEmpty && i < rows.length) {
+    while (sampleField == null && i + 1 < rows.length) {
       i += 1
       sampleField = rows(i).getField(ordinal)
     }
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
@@ -29,6 +29,7 @@ import scala.collection.mutable.ListBuffer
 import scala.language.implicitConversions
 
 import org.apache.flink.table.catalog.Column
+import org.apache.flink.table.data.StringData
 import org.apache.flink.table.types.logical._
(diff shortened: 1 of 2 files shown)


pan3793:
Thanks @link3280, could you please add the FLINK JIRA ticket link about the change you described?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
```diff
@@ -136,18 +137,22 @@ object RowSet {
...
-      case _: CharType =>
-        val tStringValue = new TStringValue
-        if (row.getField(ordinal) != null) {
-          val stringValue = row.getField(ordinal).asInstanceOf[String]
-          tStringValue.setValue(stringValue)
+        val fieldValue = row.getField(ordinal)
+        fieldValue match {
+          case _: String =>
+            val stringValue = fieldValue.asInstanceOf[String]
+            tStringValue.setValue(stringValue)
```
pan3793, line 146:
```suggestion
          case value: String =>
            tStringValue.setValue(value)
```

link3280:
@pan3793 Sure. Updated.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
```diff
@@ -165,6 +169,14 @@ object RowSet {
...
   private def toTColumn(rows: Seq[Row], ordinal: Int, logicalType: LogicalType): TColumn = {
     val nulls = new java.util.BitSet()
+    // for each column, determine the conversion class by sampling the first row
+    // if there's zero row, set empty as the column
+    var sampleField: Object = null
+    var i = -1
+    while (sampleField == null && i + 1 < rows.length) {
+      i += 1
+      sampleField = rows(i).getField(ordinal)
+    }
```
yanghua, line 179: Can we simply this code snippet via an `if/else` block?

link3280: @yanghua I'd like to. Do you have a good idea? The problem is that we can't get the underlying conversion class unless we read one value.

yanghua: > @yanghua I'd like to. Do you have a good idea? I think the key is that we can't get the underlying conversion class unless we read one value.

I mean simplifying the way that we set the value for `sampleField ` variable.

The pseudo-code looks like this:

```
if (rows.length > 0) {
  sampleField = rows(0)..getField(ordinal)
}
```

Can we remove the `while` block and `i` counter?

link3280: But if we get a null value, we still can't determine its conversion class. We need a non-null value here, that's what the loop is for.

yanghua: > But if we get a null value, we still can't determine its conversion class. We need a non-null value here, that's what the loop is for.

Oh... I misunderstood, based on your comment. Now, I suggest, you change your comment:

`by sampling the first row` -> `by sampling the first non-null row`.

So the suggestion may be to simplify the code style about how to find the first non-null value right?

May you reference this [answer](https://stackoverflow.com/a/8715730) in the StackOverflow?

link3280: Sorry for the misleading outdated comment. Updating the comment and the codes.

yanghua (approved):
LGTM

pan3793 (approved):

PR #2720, opened 2022-05-22
Fix KyuubiDatabaseMetaData#supportsCatalogs*

pan3793:

`KyuubiDatabaseMetaData#supportsCatalogs*` should return `true` as Kyuubi support concept `catalog`, otherwise, some GUI tools like DBeaver does not work as exptected.

image

image

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiDatabaseMetaData.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiDatabaseMetaData.java
@@ -106,7 +106,7 @@ public class KyuubiDatabaseMetaData implements DatabaseMetaData {
   }
 
   public String getCatalogTerm() throws SQLException {
-    return "instance";
+    return "catalog";
   }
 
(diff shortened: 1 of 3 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiDatabaseMetaData.java
```diff
@@ -106,7 +106,7 @@ public String getCatalogSeparator() throws SQLException {
   }
 
   public String getCatalogTerm() throws SQLException {
-    return "instance";
+    return "catalog";
```
ulysses-you, line 109: what's the reason of rename this term ?

pan3793: Currently, Kyuubi supports Spark, Flink, Trino, Hive engines, except to Hive, all other engines use term `catalog`.

ulysses-you (approved):

pan3793:
Thanks, merging to master/1.5

Issue #2238, opened 2022-03-28
[FEATURE] Support Flink 1.15


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Flink 1.15.0 is out, Kyuubi needs to support it.

BTW, considering the frequent changes of the Flink API and the resource limitations of Kyuubi, we need to discuss Kyuubi's support for Flink's version strategy. 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pan3793:
cc @SteNicholas @yanghua @link3280 @deadwind4

deadwind4:
I will research Apache Zeppelin's code of supporting the Flink multiple version.

link3280:
Flink adopts a time-based release strategy and releases a new version every 3-4 months. However, most companies wouldn't follow very closely and might stick to some favorable version for some time (I think 1.11/1.12 still has the most user base). Thus, I think it's good to support the latest 3 versions if possible (Flink officially maintained versions plus 1), to leave some time for users to upgrade.

KenjiFujima:
@pan3793 @deadwind4 @link3280, thanks for the feedback of the Flink engine support. Flink 1.15 version is delayed and will release for some time. IMO, the multiple Flink version support could refer to the implementation of the Zeppelin, which is currently supported by @zjffdu.
@zjffdu, could you please give some suggestion for the multiple Flink version support?

yanghua:
> Flink adopts a time-based release strategy and releases a new version every 3-4 months. However, most companies wouldn't follow very closely and might stick to some favorable version for some time (I think 1.11/1.12 still has the largest user base). Thus, I think it's good to support the latest 3 versions if possible (Flink officially maintained versions plus 1), to leave some time for users to upgrade.

+1 for maintaining multiple Flink versions (I am ok about 2 or 3).

wolfboys:
Hello everyone, Regarding the support for multiple versions of flink, I have implemented it in the StreamX project, have relevant experience, and have supported the latest flink 1.15 support, and can switch flink versions freely. I am very interested in this issue, can you please assign to me, 😊

wolfboys:
[...]

PR #2724, opened 2022-05-23
[KYUUBI #2238] [WIP] Support Flink 1.15

wolfboys:

Flink 1.15.0 is out, Kyuubi needs to support it.

Please clarify why the changes are needed. For instance

1. Rename the original `kyuubi-flink-sql-engine` to `kyuubi-flink-sql-engine-1.14`, the code inside remains unchanged
2. New `kyuubi-flex-sql-engine-1.15` module, implementation reference ` kyuubi-flex-sql-engine-1.14`
3. Removed the dependency declaration of Flink from the parent module,  because the dependency declaration of Flink is placed in the specific sub module of `kyuubi-flink-sql-engine`

`kyuubi-flex-sql-engine-1.15` part of the work is still in progress, At present, 28 test cases have passed and 9 have failed, You can review, put forward suggestions for improvement, I will continue to improve,  After this is passed, a Kyuubi flex SQL engine base module will be pulled out, and the public APIs that are not affected by the version will be put into this module. In the next step, the coexistence and isolation of multiple versions will be realized.

- [x]  Add some test cases that check the changes thoroughly including negative and positive cases if possible


externals/kyuubi-flink-sql-engine/pom.xml
```diff
@@ -61,19 +61,19 @@
 
         <dependency>
             <groupId>org.apache.flink</groupId>
-            <artifactId>flink-streaming-java_${scala.binary.version}</artifactId>
+            <artifactId>flink-streaming-java</artifactId>
```
pan3793, line 64: Based on the discussion, Kyuubi should support 2 or 3 Flink versions, this change will break support for Flink 1.14

pom.xml
```diff
@@ -109,7 +109,7 @@
         <curator.version>2.12.0</curator.version>
         <delta.version>1.2.1</delta.version>
         <fb303.version>0.9.3</fb303.version>
-        <flink.version>1.14.4</flink.version>
+        <flink.version>1.15.0</flink.version>
```
pan3793, line 112: We need add a `flink-1.15` maven profile to overwrite the `flink.version` property rather than upgrade it in place, and once we implement support everything on the new Flink version, we can change the default property here.

wolfboys: thx for you review. this time only flink 1.15 is supported. If flink 1.15 and multiple versions are supported at the same time, the development and review time of pr will be extended. For the time being, it can be split into two sub-tasks. This time, flink 1.15 will be supported first, and we will wait for this pr After passing, you can open a file about flink multi-version support.

externals/kyuubi-flink-sql-engine/pom.xml
link3280: That's indeed a critical issue. Kyuubi uses some non-public Flink APIs, which have no backward compatibility guarantee. We may need a multi-version solution like FlinkShims in Zeeplin. 

@wolfboys Does StreamX use Flink internal APIs as well? How does it achieve multi-version support? Thanks!

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
```diff
@@ -349,6 +347,8 @@ class FlinkOperationSuite extends WithFlinkSQLEngine with HiveJDBCTestHelper {
     withJdbcStatement() { statement =>
       val metaData = statement.getConnection.getMetaData
       var resultSet = metaData.getSchemas(null, null)
+      val DEFAULT_BUILTIN_CATALOG = "default_catalog"
```
link3280, line 350: How about `TableConfigOptions.TABLE_CATALOG_NAME.defaultValue()` instead?

link3280 (commented):
@wolfboys Thanks a lot for your PR! I left some comments PTAL. Thanks!

externals/kyuubi-flink-sql-engine/pom.xml
wolfboys: > That's indeed a critical issue. Kyuubi uses some non-public Flink APIs, which have no backward compatibility guarantee. We may need a multi-version solution like FlinkShims in Zeeplin.
> 
> @wolfboys Does StreamX use Flink internal APIs as well? How does it achieve multi-version support? Thanks!

Since I am not familiar with the Kyuubi project, I originally wanted to upgrade the version to Flink 1.15, implement this, and then consider multi version support. Now it seems that upgrading directly to 1.15 will affect the support of the released Flink 1.14. I will bring up the PR supported by multiple versions later

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -129,11 +137,12 @@ class ExecuteStatement(
       val resultDescriptor = executor.executeQuery(sessionId, operation)
 
       resultId = resultDescriptor.getResultId
+      val dataTypes = resultDescriptor.getResultSchema.getColumnDataTypes
 
       val rows = new ArrayBuffer[Row]()
-      var loop = true
+      val loop = new AtomicBoolean(true)
```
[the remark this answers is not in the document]
wolfboys: > @wolfboys IIUC, `loop` is a local variable that's only used in one thread. Could you please clarify why it needs to be an `AtomicBoolen`? Thanks!

This is only from the perspective of the characteristics of Scala language and Functional Programming, Variables are declared. It is best not to change, and this principle is followed.  my point may be is not true, this is my personal programming habit. avoid using var in scala

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
wolfboys: good idea. I agree with you.

pom.xml
```diff
@@ -1281,83 +1281,63 @@
...
+                <artifactId>flink-sql-parser</artifactId>
                 <version>${flink.version}</version>
             </dependency>
 
             <dependency>
                 <groupId>org.apache.flink</groupId>
-                <artifactId>flink-test-utils_${scala.binary.version}</artifactId>
+                <artifactId>flink-test-utils</artifactId>
                 <version>${flink.version}</version>
-                <exclusions>
```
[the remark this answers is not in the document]
wolfboys: > @wolfboys Could you clarify why remove these exclusions? Thanks!

`flink-test-utils` just a statement here. When it is really used in `kyuubi-flink-sql-engine` , dependency's scope is test, and other dependencies will not be attached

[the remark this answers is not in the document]
wolfboys: > @wolfboys Could you clarify why remove these exclusions? Thanks!

flink-test-utils just a declare in parent module. When it is really used in `kyuubi-flink-sql-engine` module , dependency's scope is test, and other dependencies will not be attached

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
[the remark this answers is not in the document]
link3280: @pan3793 WDYT? It's that recommended in Kyuubi's code style?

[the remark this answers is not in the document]
pan3793: I agree with @link3280, it's overkill to use AtomicBoolean here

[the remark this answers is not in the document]
wolfboys: > I agree with @link3280, it's overkill to use AtomicBoolean here

😄 I'll revise it later

yanghua:
@wolfboys please rebase this PR, thanks

pan3793:
I'm a little bit surprised that you create dedicated modules for every Flink minor version. Actually, in Kyuubi project, we try our best to avoid maintaining multi-version support in this way, we only adopt this approach in Kyuubi Spark Extension modules because all APIs that we used are private APIs.

IMO, to support both Flink 1.14 & 1.15, even if we access some Flink private API in Flink engine module, it should be handled by reflection properly. In detail, my idea is,

1) Introduce `flink-1.14` and `flink-1.15` maven profiles to handle name changes of some modules.
```
<profile>
  <id>flink-1.14</id>
  <properties>
    <flink.module.scala.suffix></flink.module.scala.suffix>
  </properties>
</profile>
<profile>
  <id>flink-1.15</id>
  <properties>
    <flink.module.scala.suffix>${scala.binary.version}</flink.module.scala.suffix>
  </properties>
</profile>
```
2)  The code should work with `flink-1.14` but may not `flink-1.15`, find them and extract a `FlinkShim` trait with implementations `FlinkShim_1_14` and `FlinkShim_1_15`, to detect and handle the Flink version in runtime.
3) Enable `flink-1.14` and `flink-1.15` in CI
4) Change the default Flink version to 1.15

wolfboys:
> 

In fact, my approach is to refer to the implementation of shims in zeppelin and streamx. If a module is based on a specific flink version such as (flink 1.14) to support multiple different flink versions, use reflection In fact, it will be more troublesome to implement. In addition to the incompatibility of API changes, some versions such as (flink 1.15) will even use new APIs (new APIs than flink 1.14),   It's not easy to do by reflection

The current method is only an initial version, and in the final implementation, the common classes and codes in each version will be extracted upward, and only the code compatible with that version will be placed in each shims package.

wolfboys:
> I'm a little bit surprised that you create dedicated modules for every Flink minor version. Actually, in Kyuubi project, we try our best to avoid maintaining multi-version support in this way, we only adopt this approach in Kyuubi Spark Extension modules because all APIs that we used are private.
> 
> IMO, to support both Flink 1.14 & 1.15, even if we access some Flink private API in Flink engine module, it should be handled by reflection properly. In detail, my idea is,
> 
> 1. Introduce `flink-1.14` and `flink-1.15` maven profiles to handle version/name changes of some modules.
> 
> ```
> <profile>
>   <id>flink-1.14</id>
>   <properties>
>     <flink.version>1.14.4</flink.version>
>     <flink.module.scala.suffix>_${scala.binary.version}</flink.module.scala.suffix>
>   </properties>
> </profile>
> <profile>
>   <id>flink-1.15</id>
>   <properties>
>     <flink.version>1.15.0</flink.version>
>     <flink.module.scala.suffix></flink.module.scala.suffix>
>   </properties>
> </profile>
> ```
> 
> 2. The code should work with `flink-1.14` but may not `flink-1.15`, find them and extract a `FlinkShim` trait with implementations `FlinkShim_1_14` and `FlinkShim_1_15`, to detect and handle the Flink version in runtime.
> 3. Enable `flink-1.14` and `flink-1.15` in CI
> 4. Change the default Flink version to 1.15
> 
> On other hand, if the API changes are huge enough, I agree with you that dedicated modules for each version is a better way.

Thank you for your review and so great point of view, I will seriously consider how to achieve it

link3280:
@wolfboys Thanks for your work! Looking forward to your upcoming PR.

PR #2760, opened 2022-05-26
Add adapter layer in Kyuubi Hive JDBC module

cxzl25:
Separate noisy code in an adapter layer to make code maintainable.


pan3793 (approved):

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/adapter/SQLConnection.java
```diff
@@ -0,0 +1,305 @@
...
+    throw new SQLFeatureNotSupportedException("Method not supported");
+  }
+
+  @Override
+  default void rollback() throws SQLException {
+    throw new SQLFeatureNotSupportedException("Method not supported");
+  }
+
+  @Override
+  default void close() throws SQLException {
```
pan3793, line 68: We can remove the default method in adaptor layer for those implemented methods.

pan3793:
Thanks, merging to master

PR #2762, opened 2022-05-27
[KYUUBI #2742] Introduce admin resource for service admin - refresh frontend hadoop conf without restart

### LINKED ISSUE #2742: [FEATURE]  Support refreshSuperUserGroupsConfiguration without restart kyuubi server
### Code of Conduct

### Search before asking

### Describe the feature
Support to renew the hadoop configuration for TFrontEndService.

### Motivation

For impersonation, it depends on local hadoop configuration.

If we want to refresh the super user group, we have to restart kyuubi server.

And hadoop namenode/resource manager support refresh the super user groups without restart service.

So, it is great that kyuubi also support `refreshSuperUserGroupsConfiguration` function.

### Describe the solution

_No response_

### Additional context

_No response_

turboFei:

To close #2742
For impersonation, it depends on local hadoop configuration.

If we want to refresh the super user group, we have to restart kyuubi server.

And hadoop namenode/resource manager support refresh the super user groups without restart service.

So, it is great that kyuubi also support refreshSuperUserGroupsConfiguration function.

With this pr, we can refresh the frontend service hadoop conf(used for proxyuser verfication) without restarting kyuubi server.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
@@ -31,6 +31,7 @@ import org.apache.thrift.transport.TTransport
 
 import org.apache.kyuubi.{KyuubiSQLException, Logging, Utils}
 import org.apache.kyuubi.Utils.stringifyException
+import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf.{FRONTEND_CONNECTION_URL_USE_HOSTNAME, FRONTEND_THRIFT_BINARY_BIND_HOST}
 import org.apache.kyuubi.operation.{FetchOrientation, OperationHandle}
 import org.apache.kyuubi.service.authentication.KyuubiAuthenticationFactory
@@ -46,7 +47,7 @@ abstract class TFrontendService(name: String)
   extends AbstractFrontendService(name) with TCLIService.Iface with Runnable with Logging {
   import TFrontendService._
   private val started = new AtomicBoolean(false)
-  private lazy val hadoopConf: Configuration = KyuubiHadoopUtils.newHadoopConf(conf)
+  private var hadoopConf: Configuration = _
   private lazy val serverThread = new NamedThreadFactory(getName, false).newThread(this)
   private lazy val serverHost = conf.get(FRONTEND_THRIFT_BINARY_BIND_HOST)
 
@@ -56,6 +57,11 @@ abstract class TFrontendService(name: String)
   protected lazy val serverSocket = new ServerSocket(portNum, -1, serverAddr)
   protected lazy val authFactory: KyuubiAuthenticationFactory =
     new KyuubiAuthenticationFactory(conf, isServer())
+
+  override def initialize(conf: KyuubiConf): Unit = {
+    super.initialize(conf)
+    hadoopConf = KyuubiHadoopUtils.newHadoopConf(conf)
+  }
 
   /**
    * Start the service itself(FE) and its composited (Discovery service, DS) in the order of:
@@ -78,6 +84,10 @@ abstract class TFrontendService(name: String)
         stopInternal()
         throw e
     }
+  }
+
+  override private[kyuubi] def reloadHadoopConf(): Unit = synchronized {
+    hadoopConf = KyuubiHadoopUtils.newHadoopConf(conf)
   }
 
   protected def stopServer(): Unit
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
@@ -31,7 +31,6 @@ import org.apache.thrift.transport.TTransport
 
 import org.apache.kyuubi.{KyuubiSQLException, Logging, Utils}
 import org.apache.kyuubi.Utils.stringifyException
-import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf.{FRONTEND_CONNECTION_URL_USE_HOSTNAME, FRONTEND_THRIFT_BINARY_BIND_HOST}
 import org.apache.kyuubi.operation.{FetchOrientation, OperationHandle}
 import org.apache.kyuubi.service.authentication.KyuubiAuthenticationFactory
@@ -47,7 +46,7 @@ abstract class TFrontendService(name: String)
   extends AbstractFrontendService(name) with TCLIService.Iface with Runnable with Logging {
   import TFrontendService._
   private val started = new AtomicBoolean(false)
-  private var hadoopConf: Configuration = _
+  private lazy val _hadoopConf: Configuration = KyuubiHadoopUtils.newHadoopConf(conf)
   private lazy val serverThread = new NamedThreadFactory(getName, false).newThread(this)
   private lazy val serverHost = conf.get(FRONTEND_THRIFT_BINARY_BIND_HOST)
 
@@ -58,10 +57,7 @@ abstract class TFrontendService(name: String)
   protected lazy val authFactory: KyuubiAuthenticationFactory =
     new KyuubiAuthenticationFactory(conf, isServer())
 
-  override def initialize(conf: KyuubiConf): Unit = {
-    super.initialize(conf)
-    hadoopConf = KyuubiHadoopUtils.newHadoopConf(conf)
-  }
+  protected def hadoopConf: Configuration = _hadoopConf
 
   /**
    * Start the service itself(FE) and its composited (Discovery service, DS) in the order of:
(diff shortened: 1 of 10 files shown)


turboFei:
thanks, updated

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.kyuubi.Logging
+import org.apache.kyuubi.server.KyuubiServer
+import org.apache.kyuubi.server.api.ApiRequestContext
+import org.apache.kyuubi.server.http.authentication.AuthenticationFilter
+import org.apache.kyuubi.service.AbstractFrontendService
+
+@Tag(name = "Admin")
+@Produces(Array(MediaType.APPLICATION_JSON))
+private[v1] class AdminResource extends ApiRequestContext with Logging {
+  private def fes: Seq[AbstractFrontendService] = KyuubiServer.kyuubiServer.frontendServices
```
yaooqinn, line 37: why we need fes?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -40,6 +41,7 @@ import org.apache.kyuubi.zookeeper.EmbeddedZookeeper
 object KyuubiServer extends Logging {
   private val zkServer = new EmbeddedZookeeper()
   private[kyuubi] var kyuubiServer: KyuubiServer = _
+  private[kyuubi] var hadoopConf: Configuration = _
```
yaooqinn, line 44: volatile

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
turboFei: removed

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
turboFei: done

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/AdminResourceSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Base64
+
+import org.apache.hadoop.security.UserGroupInformation
+
+import org.apache.kyuubi.{KyuubiFunSuite, RestFrontendTestHelper}
+import org.apache.kyuubi.server.http.authentication.AuthenticationHandler.AUTHORIZATION_HEADER
+
+class AdminResourceSuite extends KyuubiFunSuite with RestFrontendTestHelper {
+  test("test refresh frontend services hadoop conf") {
```
yaooqinn, line 28:
```suggestion
  test("refresh Hadoop configuration of the kyuubi server") {
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -33,7 +33,7 @@ import org.apache.kyuubi.server.http.authentication.AuthenticationFilter
 @Tag(name = "Admin")
 @Produces(Array(MediaType.APPLICATION_JSON))
 private[v1] class AdminResource extends ApiRequestContext with Logging {
-  private lazy val adminUser = UserGroupInformation.getCurrentUser.getShortUserName
+  private lazy val administrator = UserGroupInformation.getCurrentUser.getShortUserName
```
yaooqinn, line 36: Utils.currentUser

yaooqinn:
cc @yanghua @cfmcgrady @ulysses-you

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -90,9 +92,19 @@ object KyuubiServer extends Logging {
...
+    hadoopConf = KyuubiHadoopUtils.newHadoopConf(conf)
+    UserGroupInformation.setConfiguration(hadoopConf)
     startServer(conf)
   }
+
+  private[kyuubi] def getHadoopConf(): Configuration = {
+    hadoopConf
+  }
+
+  private[kyuubi] def reloadHadoopConf(): Unit = synchronized {
```
cxzl25, line 104: Not related to this PR.
Can we also implement `reloadkyuubiConf`?
Because some configurations are actually for the engine, when the engine parameters related to `kyuubi-defaults.conf` are modified or added, the new engine parameters submitted by the server will not take effect, and only after restarting the server will it take effect.

turboFei: kyuubi conf is widely used and you might be re-initialize many components, so it seems more complex.

turboFei: In this pr, we only refresh the hadoop conf for frontend service for proxy user verification.

For HadoopCredentialsManager conf, it is not refreshed.

turboFei: @cxzl25 

Maybe we can introduce api to refresh the conf for KyuubiSessionManager and only take affect for engine.

cxzl25: I agree with you. @turboFei 
Reloading `kyuubiConf` rashly may still have side effects.

Configuration refresh is also implemented in HDFS, using the `hdfs dfsadmin -reconfig` command, it will only refresh the configuration items marked refreshable to protect the configuration.

Because `SessionConfAdvisor` can only set session-level parameters, and then in JDBC mode, the cost of server restart is relatively high. 

Maybe we can design how to achieve better dynamic configuration?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -40,8 +41,10 @@ import org.apache.kyuubi.zookeeper.EmbeddedZookeeper
 object KyuubiServer extends Logging {
   private val zkServer = new EmbeddedZookeeper()
   private[kyuubi] var kyuubiServer: KyuubiServer = _
+  @volatile private[kyuubi] var hadoopConf: Configuration = _
 
   def startServer(conf: KyuubiConf): KyuubiServer = {
+    hadoopConf = KyuubiHadoopUtils.newHadoopConf(conf)
```
turboFei, line 47: Here is the entrance for UT.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -46,7 +46,7 @@ abstract class TFrontendService(name: String)
   extends AbstractFrontendService(name) with TCLIService.Iface with Runnable with Logging {
   import TFrontendService._
   private val started = new AtomicBoolean(false)
-  private lazy val hadoopConf: Configuration = KyuubiHadoopUtils.newHadoopConf(conf)
+  private lazy val _hadoopConf: Configuration = KyuubiHadoopUtils.newHadoopConf(conf)
```
ulysses-you, line 49: why thrift server should not reload hadoop conf ?

turboFei: Only KyuubiTBinaryFrontendService will reload the hadoop conf, the engine will not.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -0,0 +1,57 @@
...
+  private lazy val administrator = Utils.currentUser
+
+  @ApiResponse(
+    responseCode = "200",
+    content = Array(new Content(
+      mediaType = MediaType.APPLICATION_JSON)),
+    description = "refresh the Kyuubi server hadoop conf, note that, " +
+      "it only takes affect for frontend services now")
+  @POST
+  @Path("{refreshServerHadoopConf}")
```
yanghua, line 44: I would not suggest using the camelCase-style in the URI.

yanghua, line 44: Do not need the brace? It seems it's not a `PathParam` right?

yanghua (commented):
Left two comments.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
turboFei: nice catch

turboFei: /admin/refresh/hadoopconf?

yanghua: lgtm

turboFei:
cc @ulysses-you @pan3793 @yanghua

turboFei:
thanks, merging to master

Issue #2643, opened 2022-05-13
[SUB-TASK][KPIP-4] Implement kyuubi rest sdk for batch job submission 


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

sub task of #2240 

We should introduce a new module for kyuubi rest sdk to submit batch job.

It should be written in java for better compatibilities.
![ccoe_review-Page-4 drawio](https://user-images.githubusercontent.com/6757692/168195386-035ee597-562a-4d41-a69f-233723be9508.png)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

turboFei:
Thanks @lightning-L  for taking this.

PR #2779, opened 2022-05-30
[KYUUBI #2643][FOLLOWUP] Refine the rest sdk

ulysses-you:

1. Add a static method `builder` in `KyuubiRestClient` to make it easy to use
2. make the default constructor private with `KyuubiRestClient` and `BatchRestApi`
3. enable surefire plugin test for java ut

### DIFF
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
@@ -32,6 +32,8 @@ public class BatchRestApi {
 
   private static final String API_BASE_PATH = "batches";
 
+  private BatchRestApi() {}
+
   public BatchRestApi(KyuubiRestClient client) {
     this.client = client;
   }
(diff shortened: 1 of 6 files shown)


ulysses-you:
cc @turboFei @lightning-L

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
```diff
@@ -32,6 +32,8 @@ public class BatchRestApi {
 
   private static final String API_BASE_PATH = "batches";
 
+  private BatchRestApi() {}
```
turboFei, line 35: why we need this?

turboFei: to disable user invoke the default constructor?

ulysses-you: The default constructor will cause npe since the client is not set, that means we must call the constructor which hold the KyuubiRestClient

turboFei: got it, thanks

turboFei (approved):
LGTM

turboFei:
thanks, merging to master

PR #2780, opened 2022-05-30
Refine stylecheck

cfmcgrady:

1. reformat code should include all of the modules in the project.
2. stylecheck should include all of the modules in the project.
3. there is a bug in the stage `Upload style check report`.

Since the `grep` returns code 1 when the pattern is not found, the upload report does not work before this pr.

```
> echo "a" > /tmp/a; grep abc /tmp/a; echo $?
1
```

Before this PR
```
Run for log in `find * -name "scalastyle-output.xml"`;  do echo "=========$log========="; grep "error" $log; done
=========dev/kyuubi-tpcds/target/scalastyle-output.xml=========
Error: Process completed with exit code 1.
```

After this PR
```
Run for log in `find * -name "scalastyle-output.xml"`;
<?xml version="1.0" encoding="UTF-8"?>
<checkstyle version="5.0">
 <file name="/home/runner/work/incubator-kyuubi/incubator-kyuubi/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/KyuubiSparkSQLExtension.scala">
  <error column="0" line="21" source="org.scalastyle.scalariform.ImportOrderChecker" severity="error" message="There should at least one a single empty line separating groups 3rdParty and kyuubi."/>
 </file>
</checkstyle>
```

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -36,10 +36,6 @@ jobs:
   linter:
     name: Style check
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        profiles:
-          - '-Ptpcds -Pspark-block-cleaner -Pkubernetes-it -Pspark-3.1 -Pspark-3.2'
     steps:
       - uses: actions/checkout@v2
       - name: Setup JDK 8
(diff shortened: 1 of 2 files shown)


.github/workflows/style.yml
```diff
@@ -50,9 +46,15 @@ jobs:
           cache: 'maven'
           check-latest: false
       - name: Scalastyle with maven
-        run: build/mvn scalastyle:check ${{ matrix.profiles }} -pl '!:kyuubi-codecov_2.12,!:kyuubi-assembly_2.12'
+        run: build/mvn help:all-profiles -pl . | grep "Profile Id" | awk '{print $3}' | tr '\n' ',' | awk '{print "-P"$0}' | xargs build/mvn scalastyle:check
       - name: Upload scalastyle report
```
ulysses-you, line 50: the name should be `print scalastyle error report` ?

pan3793:
```
➜  apache-kyuubi git:(master) mvn help:all-profiles -pl . | grep "Profile Id" | awk '{print $3}' | tr '\n' ',' | awk '{print "-P"$0}'
-Pjava-11,apache-release,tpcds,java-8,hive-provided,spark-provided,spark-block-cleaner,flink-provided,mirror-cdn,spark-3.3,spark-3.2,spark-3.1,spark-3.0,flink-1.15,spark-master,flink-1.14,
```
Looks like using all profiles is overkill

cfmcgrady:
> Looks like using all profiles is overkill

It's for finding out all of the modules in the project, do you have a better idea?

pan3793:
I prefer to use explicit profiles, because I don't know what will happen if we use both `java-8` `java-11`

cfmcgrady:
> I prefer to use explicit profiles,

SGTM, updated.

ulysses-you:
thanks, merging to master

PR #2787, opened 2022-05-31
[KYUUBI #2643][FOLLOWUP] Using javax AuthenticationException instead of hadoop AuthenticationException

turboFei:

Using javax.security.sasl.AuthenticationException instead of hadoop AuthenticationException


turboFei:
> The UserGroupInformation may also be replaced by jdk Subject?
Seems difficult to do that if user use `UserGroupInformation.setLoginUser(loginUser)`.

ulysses-you:
seems we can remove hadoop dependency in pom ?

turboFei:
> seems we can remove hadoop dependency in pom ?

It is still used here.
https://github.com/apache/incubator-kyuubi/pull/2787/files#diff-282b1e8c3fee6a8b8b6814d201ba42ad60fc1a1b5546fcc2a31150f3e5ff89abR46

ulysses-you:
thanks, merging to master

turboFei:
thanks for your review @ulysses-you

PR #2800, opened 2022-06-01
Refine batch mode code path

ulysses-you:

1. rename `Metadata` to `SessionMetadata`
2. wrap `CloseBatchResponse` for closeBatchSession
3. remove killApp in closeBatchSession, so every request will kill the app by default
4. correct when to update metadata
5. cleanupMetadataByAge should check state and only cleanup terminated metadata
6. correct the batch session state, we should never get a closed state


kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/CloseBatchResponse.java
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.client.api.v1.dto;
+
+import java.util.Objects;
+
+public class CloseBatchResponse {
```
yanghua, line 22: There is a container file named `dto` that packaged many `case class`. Since you introduce a `dto` directory. So can we refactor the old dto?

ulysses-you: yeah, it should be done at https://github.com/apache/incubator-kyuubi/pull/2668. We can do further clean up if you find some more miss rewrite class.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -17,23 +17,40 @@
...
+ *
+ *                        /  ERROR
+ * PENDING  ->  RUNNING  ->  FINISHED
+ *                        \  CANCELED (CLOSED)
+ *
+ * We can not change FINISHED/ERROR/CANCELED to CLOSED, and it's different with other operation
+ * which final status is always CLOSED, so we do not use CLOSED state in this class.
+ * To compatible with kill application we combine the semantics of `cancel` and `close`, so if
+ * user close the batch session that means the final status is CANCELED.
+ */
```
ulysses-you, line 53: After some thought, the state lifecycle seems quite different between batch and sql. So I do this change in this pr. what do you think ? @turboFei @lightning-L @yaooqinn @pan3793 @yanghua @zhaomin1423

yanghua: I am thinking of `ERROR` or `FAILED`.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -86,8 +106,27 @@ class BatchJobSubmission(
...
+  private def updateBatchMetadata(): Unit = {
+    val endTime =
+      if (isTerminalState(getStatus.state)) {
+        getStatus.lastModified
+      } else {
+        0L
+      }
+    session.sessionManager.updateBatchMetadata(
+      batchId,
+      getStatus.state,
```
turboFei, line 118: getStatus.state => state

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -86,8 +106,27 @@ class BatchJobSubmission(
   private val applicationCheckInterval =
     session.sessionConf.get(KyuubiConf.BATCH_APPLICATION_CHECK_INTERVAL)
 
+  private def updateBatchMetadata(): Unit = {
+    val endTime =
+      if (isTerminalState(getStatus.state)) {
+        getStatus.lastModified
```
turboFei, line 112: getStatus.lastModified => lastAccessTime

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -86,8 +106,27 @@ class BatchJobSubmission(
...
+      } else {
+        0L
+      }
+    session.sessionManager.updateBatchMetadata(
+      batchId,
+      getStatus.state,
+      applicationStatus.getOrElse(Map.empty),
```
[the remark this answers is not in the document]
turboFei: I am thinking about how can we provide more prompt for customer if it failed before submitted to resource manager.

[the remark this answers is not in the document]
ulysses-you: Seems hard, user should get cause from log. And we should provide remote log if local log is not enough.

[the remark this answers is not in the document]
turboFei: what is remote log? do you mean that uploading the local log file to global file system?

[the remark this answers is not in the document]
ulysses-you: remote logs can be fetched through resourse manager, e.g. the cluster mode  driver log from yarn, or pod log from k8s

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -86,35 +106,60 @@ class BatchJobSubmission(
...
     }
+
     try {
       val opHandle = session.sessionManager.submitBackgroundOperation(asyncOperation)
       setBackgroundHandle(opHandle)
-    } catch onError("submitting batch job submission operation in background, request rejected")
+    } catch {
+      onError("submitting batch job submission operation in background, request rejected")
+    } finally {
+      updateBatchMetadata()
```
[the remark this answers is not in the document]
turboFei: how about:

```
if (isTerminalState(getStatus.state)) {
  updateBatchMetadata()
}

```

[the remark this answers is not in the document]
ulysses-you: make sense

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -150,12 +150,11 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
     content = Array(new Content(
       mediaType = MediaType.APPLICATION_JSON)),
-    description = "close a batch session")
+    description = "close and cancel a batch session")
   @DELETE
   @Path("{batchId}")
   def closeBatchSession(
       @PathParam("batchId") batchId: String,
-      @QueryParam("killApp") killApp: Boolean,
       @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String): Response = {
```
turboFei, line 158: should we set the return type to `CloseBatchResponse`?

turboFei: so that the response schema is visible in swagger.

turboFei (commented):
left minor comments, LGTM overall

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
ulysses-you: fixed

turboFei: not sure whether it is needed, you can check the swagger locally


```
      schema = new Schema(implementation = classOf[CloseBatchResponse]))),
```

ulysses-you: fixed

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -244,8 +244,21 @@ class BatchJobSubmission(
...
+      } catch {
+        case e: Exception => warn("failed to close engine process builder", e)
+      }
+      MetricsSystem.tracing(_.decCount(
+        MetricRegistry.name(OPERATION_OPEN, statement.toLowerCase(Locale.getDefault))))
+
+      // fast fail
+      if (isTerminalState(state)) {
+        killMessage = (false, s"batch $batchId is already terminal so can not kill it.")
+        return
```
turboFei, line 257: seems that we can not close builder before `killBatchApplication`.

For yarn-cluster, if we close builder, then the async thread will terminate because the builder process is not alive.

And then the state is terminalState, but the application is still running.

ulysses-you: good catch I considered before but missed it last commit ..

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -86,35 +106,62 @@ class BatchJobSubmission(
   private val applicationCheckInterval =
     session.sessionConf.get(KyuubiConf.BATCH_APPLICATION_CHECK_INTERVAL)
 
+  private def updateBatchMetadata(): Unit = {
+    val endTime =
+      if (isTerminalState(getStatus.state)) {
```
turboFei, line 111: getStatus.state => state

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -189,12 +233,46 @@ class BatchJobSubmission(
     }.getOrElse(ThriftUtils.EMPTY_ROW_SET)
   }
 
-  override def close(): Unit = {
+  override def close(): Unit = state.synchronized {
     if (!isClosedOrCanceled) {
-      if (builder != null) {
+      try {
+        // For launch engine operation, we use OperationLog to pass engine submit log but
+        // at that time we do not have remoteOpHandle
```
turboFei, line 240: we can remove the comment or modify it(launch engine operation).

turboFei (approved):

turboFei:
thanks, merging to master

ulysses-you:
thank you all !

PR #2802, opened 2022-06-02
Retry opening the TSocket in KyuubiSyncThriftClient

mahmoudbahaa:

the new option: **kyuubi.operation.thrift.client.request.max.attempts** is not respected in createTProtocol in KyuubiSyncThriftClient thus it fails from the first time and doesn't retry even if **kyuubi.operation.thrift.client.request.max.attempts** was geater than 1. so adjusted the code to allow retry for that case. (My scala is kinda rusty so forgive if there is some uncleanness in the code)

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
@@ -106,23 +106,14 @@ class KyuubiSyncThriftClient private (
   }
 
   private def withRetryingRequest[T](block: => T, request: String): T = withLockAcquired {
-    var attemptCount = 1
+    val (resp, shouldResetEngineBroken) = KyuubiSyncThriftClient.withRetryingRequestNoLock(
+      block,
+      request,
+      maxAttempts,
+      remoteEngineBroken,
+      isConnectionValid)
 
-    var resp: T = null.asInstanceOf[T]
-    while (attemptCount <= maxAttempts && resp == null) {
-      try {
-        resp = block
-        remoteEngineBroken = false
-      } catch {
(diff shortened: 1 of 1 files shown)


turboFei:
FYI:
You can run below command to format the code style.
```
dev/reformat
```

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -106,23 +106,13 @@ class KyuubiSyncThriftClient private (
...
-          throw e
-      }
-    }
+    val (resp, remoteEngineBrokenRetValue) = KyuubiSyncThriftClient.withRetryingRequestNoLock(
+      block,
+      request,
+      maxAttempts,
+      remoteEngineBroken,
+      isConnectionValid)
+    remoteEngineBroken = remoteEngineBrokenRetValue;
```
ulysses-you, line 115: The semantics seems a little different with previous. Before we only reset `remoteEngineBroken` if we succeeded to do request, and if failed the `remoteEngineBroken` depend on the probe so we do not change it.

So, we do not pass `remoteEngineBroken` to the new method `withRetryingRequestNoLock`, and change the semantics of the returned value `remoteEngineBrokenRetValue` to `shouldResetRemoteEngineBroken`, then if return true, we change  `remoteEngineBroken` to false, else we do nothing which is same with previous.

mahmoudbahaa: @ulysses-you can you please re-check

ulysses-you (approved):
lgtm, thank you @mahmoudbahaa !

turboFei (approved):
LGTM

turboFei:
thanks, merged to master

PR #2811, opened 2022-06-02
register java spark udfs through spark custom configurations

mahmoudbahaa:

As far as I can tell there is no way to register spark java UDFs (through jars) through SQL commands the same way as hive UDFs. https://spark.apache.org/docs/latest/sql-ref-functions-udf-hive.html The create command passes without problem but when using the custom UDF it gives the error: **org.apache.spark.sql.AnalysisException: No handler for UDF/UDAF/UDTF** so as a workaround added number of custom kyuubi options as follow:

for each udf add an entry as follow
kyuubi.spark.udf.udfName=class,numArgs,returnTypes

return types have to be of type DataType defined in org.apache.spark.sql.types.DataTypes: possible values: StringType, BinaryType, BooleanType, DateType, TimestampType, CalendarIntervalType, DoubleType, ByteType, IntegerType, LongType, ShortType, NullType

Notes:

1. jars for the classes can be added using spark.jar option

example:
kyuubi.spark.udf.udfName1=com.exmaple.UDF1,2,DoubleType
kyuubi.spark.udf.udfName2=com.example.UDF2,5,StringType
kyuubi.spark.udf.udfName3=com.example.UDF3,8,BooleanType

Tried on 23 UDFs and they got loaded properly example is below

Screenshot_20220602_143958
Screenshot_20220602_151310

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
@@ -21,16 +21,22 @@ import scala.collection.mutable.ArrayBuffer
 
 import org.apache.spark.{SparkEnv, TaskContext}
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.api.java._
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import org.apache.spark.sql.functions.udf
+import org.apache.spark.sql.types.{DataType, DataTypes}
 
 import org.apache.kyuubi.{KYUUBI_VERSION, Utils}
+import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiReservedKeys.KYUUBI_SESSION_USER_KEY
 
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/udf/KDFRegistry.scala
```diff
@@ -88,5 +106,318 @@ object KDFRegistry {
     for (func <- registeredFunctions) {
       spark.udf.register(func.name, func.udf)
     }
+
+    if (extraUdfsNames.length != 0 && extraUdfsClasses.length != 0 &&
```
yaooqinn, line 110: we can get these from `spark.conf`

mahmoudbahaa: I don't get what you mean can you explain more please ?

yaooqinn: replace `SparkEnv.get.conf ` with `SparkSession.conf`

PR #2812, opened 2022-06-02
[SUB-TASK][KPIP-4] Refine the batch info response

turboFei:

Return more batch info for insight: `user`, `name`, `createTime` and `endTime`.

Support to list batches with filter conditions:
- batchState
- batchUser
- createTime
- endTime

### DIFF
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
@@ -48,9 +48,24 @@ public class BatchRestApi {
     return this.getClient().get(path, null, Batch.class, client.getAuthHeader());
   }
 
-  public GetBatchesResponse listBatches(String batchType, int from, int size) {
+  public GetBatchesResponse listBatches(
+      String batchType,
+      String batchUser,
+      String batchState,
+      Long createTime,
+      Long endTime,
+      int from,
+      int size) {
     Map<String, Object> params = new HashMap<>();
     params.put("batchType", batchType);
+    params.put("batchUser", batchUser);
+    params.put("batchState", batchState);
+    if (null != createTime && createTime >= 0) {
+      params.put("createTime", createTime);
+    }
+    if (null != endTime && endTime >= 0) {
+      params.put("endTime", endTime);
+    }
     params.put("from", from);
     params.put("size", size);
     return this.getClient()
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -31,9 +31,9 @@ import org.apache.hive.service.rpc.thrift.TProtocolVersion
 
 import org.apache.kyuubi.Logging
 import org.apache.kyuubi.client.api.v1.dto._
-import org.apache.kyuubi.operation.FetchOrientation
+import org.apache.kyuubi.operation.{FetchOrientation, OperationState}
 import org.apache.kyuubi.server.api.ApiRequestContext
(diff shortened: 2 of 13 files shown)


ulysses-you:
As we are here, we can return the start time and end time in Batch. Includes get batch and list bathes interface.

turboFei:
> As we are here, we can return the start time and end time in Batch. Includes get batch and list bathes interface.

sure

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
```diff
@@ -47,9 +47,12 @@ public Batch getBatchById(String batchId) {
     return this.getClient().get(path, null, Batch.class, client.getAuthHeader());
   }
 
-  public GetBatchesResponse listBatches(String batchType, int from, int size) {
+  public GetBatchesResponse listBatches(
+      String batchType, String batchUser, String batchState, int from, int size) {
     Map<String, Object> params = new HashMap<>();
     params.put("batchType", batchType);
+    params.put("batchUser", batchUser);
+    params.put("batchState", batchState);
```
ulysses-you, line 55: how about supporting filter by createTime and endTime ?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -164,7 +164,7 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
   }
 
   def getBatchSessionImpl(sessionHandle: SessionHandle): KyuubiBatchSessionImpl = {
-    getSession(sessionHandle).asInstanceOf[KyuubiBatchSessionImpl]
+    Option(handleToSession.get(sessionHandle)).map(_.asInstanceOf[KyuubiBatchSessionImpl]).orNull
```
ulysses-you, line 167: we can add a new method `getSessionOption` in parent class instead of exposing `handleToSession`.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
turboFei: seems not needed, now the createTime and endTime are all LongType and it is difficult to filter now.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
turboFei: thanks

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
turboFei: let me have a try

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -95,9 +96,14 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
   @Path("{batchId}")
   def batchInfo(@PathParam("batchId") batchId: String): Batch = {
-    Option(sessionManager.getBatch(batchId)).getOrElse {
-      error(s"Invalid batchId: $batchId")
-      throw new NotFoundException(s"Invalid batchId: $batchId")
+    val sessionHandle = normalizedBatchSessionHandle(batchId)
+    Option(sessionManager.getBatchSessionImpl(sessionHandle)).map { batchSession =>
+      buildBatch(batchSession)
+    }.getOrElse {
+      Option(sessionManager.getBatch(batchId)).getOrElse {
```
ulysses-you, line 103: we can rename `getBatch` to `getBatchFromStateStore`

ulysses-you (approved):

PR #2822, opened 2022-06-07
[GA] Set log level to info

pan3793:

Have seen run out of disk spaces in GitHub Action, set log level to info to reduce log size

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/test/resources/log4j2-test.properties
+++ b/externals/kyuubi-flink-sql-engine/src/test/resources/log4j2-test.properties
@@ -16,7 +16,7 @@
 #
 
 # Set everything to be logged to the file target/unit-tests.log
-rootLogger.level = debug
+rootLogger.level = info
 rootLogger.appenderRef.stdout.ref = STDOUT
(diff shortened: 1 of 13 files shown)


externals/kyuubi-flink-sql-engine/src/test/resources/log4j2-test.properties
```diff
@@ -16,7 +16,7 @@
 #
 
 # Set everything to be logged to the file target/unit-tests.log
-rootLogger.level = debug
+rootLogger.level = info
```
cfmcgrady, line 19: Shall we also update L57?

pan3793:
One CI job failed, for a reason
```
- test select *** FAILED ***
  java.sql.SQLException: org.apache.kyuubi.KyuubiSQLException: Error operating EXECUTE_STATEMENT: java.sql.SQLException: errCode = 2, detailMessage = Failed to find enough host with storage medium and tag(HDD/{"location" : "default"}) in all backends. need: 1
```
But the output log size dramatically reduces from ~1GB to ~30MB

pan3793:
Thanks, merging to master

PR #2824, opened 2022-06-07
[TEST] Replace test tag ExtendedSQLTest by Slow

pan3793:

Currently, we define a `ExtendedSQLTest` in `kyuubi-common` test sources, and use it to mark some slow test. This PR propose to replace `ExtendedSQLTest` by `Slow` which is provided by scalatest, then we can use the same tag to mark slow tests in some modules which does not depends on `kyuubi-common`, and then seperate them into independent CI job to reduce the CI time


pan3793:
cc @cxzl25

cxzl25 (approved):
LGTM

PR #2829, opened 2022-06-08
Make secret id static and remove thrift protocol from RPC handles

yaooqinn:

1. remove `thrift protocol version` from REST req and resp.
   - it's useless for this REST. It's not used for version check and not for result set determination.
   - it's not a part of TSessionHandle/TOperationHandle either. The protocol version is determined in open session phase and used both by the session and the within operations. It should not be set per request. 
2. Make secret-id static across rest/batch/jdbc APIs


turboFei:
```
Error:  /home/runner/work/incubator-kyuubi/incubator-kyuubi/kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/BatchesResourceSuite.scala:438: value getBatchSessionHandle is not a member of org.apache.kyuubi.session.KyuubiSessionManager

```

ulysses-you:
Seems clear to remove thrift protocol from RPC handles

turboFei (approved):
LGTM, thanks

yaooqinn:
thanks, merged to master

Issue #886, opened 2021-08-02
[Umbrella] Add HTTP transport mode support to KYUUBI


RamakrishnaChilaka:
### 1. Describe the feature
Access KYUUBI HTTP calls.

### 2. Motivation
if we can support submitting queries via HTTP mode,  this will enable us to use this via HAPROXY/NGINX reverse proxys.
We can use Jupyter Notebooks and backend servers to query spark sql as a service.

### 3. Tasks

- [ ] #1761
- [ ] Implement a httpserver by jetty

RamakrishnaChilaka:
https://github.com/apache/incubator-kyuubi/issues/32 this already talks about HTTP frontend. Closing this.

yaooqinn:
that ticket is not exactly the same，you can work on this if you are interested

RamakrishnaChilaka:
I am interested @yaooqinn , can you point me to the codebase documentation ?

yaooqinn:
docs/ ？

iodone:
@RamakrishnaChilaka This mode is just like Hive Thrift Server HttpMode?  The protocol for client interaction with Kyubi is still Thrift, and the communication protocol is HTTP.

RamakrishnaChilaka:
Yes @iodone .

SteNicholas:
@yaooqinn, could you please assign this ticket to me?

PR #2833, opened 2022-06-08
[KYUUBI #886] Add HTTP transport mode support to KYUUBI - no Kerberos support

mahmoudbahaa:
Same as https://github.com/apache/incubator-kyuubi/pull/2815 but with Kerberos auth removed as was not complete nor tested. so thought to make a separated PR with no Kerberos support then we can add Kerberos support later on

Add http transport protocol support to Kyuubi
main code was taken from:

1. https://github.com/SteNicholas/incubator-kyuubi/commit/25ac0731d7331616988026996ca918793af151e9
2. https://github.com/apache/hive/blob/branch-3.1/service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpCLIService.java
2. https://github.com/apache/hive/blob/branch-3.1/service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java


yaooqinn:
we can remove the current tests in this PR and add a test like this, which covers both client/server behavior.

```scala
package org.apache.kyuubi.operation.thift.http

class KyuubiOperationPerUserSuite extends WithKyuubiServer with SparkQueryTests {

  override protected val conf: KyuubiConf = KyuubiConf()

  override protected val frontendProtocols: Seq[KyuubiConf.FrontendProtocols.Value] =
    FrontendProtocols.THRIFT_HTTP :: Nil

  override protected def getJdbcUrl: String =
    s"jdbc:hive://${server.frontendServices.head.connectionUrl}/;transportMode=http"

  override protected def jdbcUrl: String = getJdbcUrl
}
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/http/util/HttpAuthUtils.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import org.apache.kyuubi.Logging
+
+object HttpAuthUtils {
+  val WWW_AUTHENTICATE = "WWW-Authenticate"
+  val AUTHORIZATION = "Authorization"
+  val BASIC = "Basic"
+  val NEGOTIATE = "Negotiate"
+
+  private val logger = new Logger
```
yaooqinn, line 34:
```suggestion
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/http/util/HttpAuthUtils.scala
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Parses a cookie token to retrieve client user name.
+   *
+   * @param tokenStr Token String.
+   * @return A valid user name if input is of valid format, else returns null.
+   */
+  def getUserNameFromCookieToken(tokenStr: String): String = {
+    val map = splitCookieToken(tokenStr)
+    if (!(map.keySet == COOKIE_ATTRIBUTES)) {
+      logger.error("Invalid token with missing attributes " + tokenStr)
```
yaooqinn, line 67:
```suggestion
      error("Invalid token with missing attributes " + tokenStr)
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/http/util/HttpAuthUtils.scala
```diff
@@ -0,0 +1,100 @@
...
+        return null
+      }
+      val key = part.substring(0, separator)
+      val value = part.substring(separator + 1)
+      map.put(key, value)
+    }
+    map
+  }
+
+  private class Logger extends Logging
```
yaooqinn, line 99:
```suggestion
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/http/util/SessionManager.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  def getProxyUserName: String = {
+    return threadLocalProxyUserName.get
+  }
+
+  def clearProxyUserName(): Unit = {
+    threadLocalProxyUserName.remove()
+  }
+
+  private class Logger extends Logging
```
yaooqinn, line 92:
```suggestion
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  override protected def getIpAddress: String = {
+    SessionManager.getIpAddress
+  }
+
+  override protected def getUserName(req: TOpenSessionReq): String = {
+    var userName: String = SessionManager.getUserName
+    if (userName == null) userName = req.getUsername
+    userName = getShortName(userName)
```
yaooqinn, line 267: we can use org.apache.kyuubi.service.ServiceUtils#getShortName

yaooqinn (approved):

pan3793:
Thanks, merging to master

PR #2842, opened 2022-06-09
[TEST] Optimize the output of ExceptionThrowingDelegationTokenProvider in the Test

cxzl25:

Now kyuubi-server runs some tests that are not `HadoopCredentialsManagerSuite`, `ExceptionThrowingDelegationTokenProvider` still throws exception.
```
Caused by: java.lang.IllegalArgumentException
	at org.apache.kyuubi.credentials.ExceptionThrowingDelegationTokenProvider.<init>(HadoopCredentialsManagerSuite.scala:229) 
```

In this PR, I limit it to `HadoopCredentialsManagerSuite` to allow exceptions to be thrown. Other suites can load this provider by default, but it does not affect other suite tests.


yaooqinn:
thanks, merged to master

PR #2848, opened 2022-06-10
Global temp view should only exist in session catalog

pan3793:

DS v2 does not support view yet.
image

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v3_0.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v3_0.scala
@@ -129,10 +129,13 @@ class CatalogShim_v3_0 extends CatalogShim_v2_4 {
       spark: SparkSession,
       catalogName: String,
       schemaPattern: String): Seq[Row] = {
-    val viewMgr = getGlobalTempViewManager(spark, schemaPattern)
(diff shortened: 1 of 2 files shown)


pan3793:
It's a long term bug, I mark it as a blocker for the ongoing release v1.5.2

pan3793:
cc @ulysses-you

ulysses-you:
Accroiding to the comments of how to resolve global_temp, we should mark global_temp as a special catalog. But given Spark has not supported v2 catalog for view, we can mark global_temp as a special table in spark built-in catalog spark_catalog.

https://github.com/apache/spark/blob/b5c7f34b576d25aec292c65e7565360d67142227/sql/catalyst/src/main/scala/org/apache/spark/sql/connector/catalog/LookupCatalog.scala#L113-L119

So the snapshot after correct the issue will be:
```
tpcd
  - sf0
  - tiny
  - sf1
  - sf10
  - sf30

spark_catalog
  - default
  - global_temp

......
```

If Spark supports v2 catalog for view in future, we can change it to:
```
tpcd
  - sf0
  - tiny
  - sf1
  - sf10
  - sf30

spark_catalog
  - default

global_temp

......
```

ulysses-you:
cc @yaooqinn

yaooqinn:
oh, I made this minor bug with the purpose to see if anyone use multi-cat feature

yaooqinn:
An issue is better not to be a blocker when it is trivial or long-standing

yaooqinn (approved):

pan3793:
Thanks, merging to master/1.5

PR #2849, opened 2022-06-10
Close the engine alive pool gracefully

turboFei:

Close the engine alive thread pool and wait the in flight task finished.
And try to close the engine alive probe client anyway.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
@@ -21,6 +21,7 @@ object KyuubiReservedKeys {
   final val KYUUBI_SESSION_USER_KEY = "kyuubi.session.user"
   final val KYUUBI_STATEMENT_ID_KEY = "kyuubi.statement.id"
   final val KYUUBI_ENGINE_SUBMIT_TIME_KEY = "kyuubi.engine.submit.time"
+  final val KYUUBI_ENGINE_SESSION_TAG_KEY = "kyuubi.engine.session.tag"
   final val KYUUBI_SESSION_ENGINE_LAUNCH_HANDLE_GUID =
     "kyuubi.session.engine.launch.handle.guid"
   final val KYUUBI_SESSION_ENGINE_LAUNCH_HANDLE_SECRET =
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
@@ -21,7 +21,6 @@ object KyuubiReservedKeys {
   final val KYUUBI_SESSION_USER_KEY = "kyuubi.session.user"
   final val KYUUBI_STATEMENT_ID_KEY = "kyuubi.statement.id"
   final val KYUUBI_ENGINE_SUBMIT_TIME_KEY = "kyuubi.engine.submit.time"
-  final val KYUUBI_ENGINE_SESSION_TAG_KEY = "kyuubi.engine.session.tag"
   final val KYUUBI_SESSION_ENGINE_LAUNCH_HANDLE_GUID =
     "kyuubi.session.engine.launch.handle.guid"
   final val KYUUBI_SESSION_ENGINE_LAUNCH_HANDLE_SECRET =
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -161,8 +164,10 @@ class KyuubiSyncThriftClient private (
       case e: Exception =>
         throw KyuubiSQLException("Error while cleaning up the engine resources", e)
     } finally {
-      Option(engineAliveThreadPool).foreach(_.shutdown())
-      if (_aliveProbeSessionHandle != null && !remoteEngineBroken) {
+      Option(engineAliveThreadPool).foreach { pool =>
+        ThreadUtils.shutdown(pool, Duration(engineAliveProbeInterval, TimeUnit.MILLISECONDS))
+      }
+      if (_aliveProbeSessionHandle != null) {
```
turboFei, line 170: try to close it anyway

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -161,8 +164,10 @@ class KyuubiSyncThriftClient private (
       case e: Exception =>
         throw KyuubiSQLException("Error while cleaning up the engine resources", e)
     } finally {
-      Option(engineAliveThreadPool).foreach(_.shutdown())
-      if (_aliveProbeSessionHandle != null && !remoteEngineBroken) {
+      Option(engineAliveThreadPool).foreach { pool =>
+        ThreadUtils.shutdown(pool, Duration(engineAliveProbeInterval, TimeUnit.MILLISECONDS))
```
turboFei, line 168: wait the current task finished to prevent out of sequence issue.

pan3793 (approved):

pan3793:
BTW `KyuubiSyncThriftClient` take more responsibilities than `sync`, maybe we can rename it to `KyuubiThriftClient`

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
ulysses-you: engineAliveProbeInterval -> connectionTimeout ?

turboFei: do you mean rename the `engineAliveProbeInterval` variable?

ulysses-you: If the probe request is in flight, the max flight time should not be `engineAliveProbeInterval`. We should use a more accurate time.

turboFei: engineAliveProbeInterval is also the requestTimeout of the alive probe thrift client.

```
          createTProtocol(user, passwd, host, port, aliveProbeInterval, loginTimeout),
```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
```diff
@@ -21,6 +21,7 @@ object KyuubiReservedKeys {
   final val KYUUBI_SESSION_USER_KEY = "kyuubi.session.user"
   final val KYUUBI_STATEMENT_ID_KEY = "kyuubi.statement.id"
   final val KYUUBI_ENGINE_SUBMIT_TIME_KEY = "kyuubi.engine.submit.time"
+  final val KYUUBI_ENGINE_SESSION_TAG_KEY = "kyuubi.engine.session.tag"
```
yaooqinn, line 24: what is it for?

yaooqinn (changes requested):
2 features seem not related, break them into 2 PRs

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
ulysses-you: ah I see, then it's ok

yaooqinn:
> Add the tag to make the session info more clear.

we already have session name as a kyuubi conf, using it for logging?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
turboFei: will raise a new pr.

turboFei:
> BTW KyuubiSyncThriftClient take more responsibilities than sync, maybe we can rename it to KyuubiThriftClient

How do you think about this? @yaooqinn

yaooqinn:
> > BTW KyuubiSyncThriftClient take more responsibilities than sync, maybe we can rename it to KyuubiThriftClient
> 
> How do you think about this? @yaooqinn

I checked the KyuubiSyncThriftClient, name changing here is not necessary to me

turboFei:
thanks, merging to master

PR #2851, opened 2022-06-10
Log session name when opening/closing session

turboFei:

Log session name when opening/closing session

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/Session.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/Session.scala
@@ -32,6 +32,7 @@ trait Session {
   def user: String
   def password: String
   def ipAddress: String
+  def name: String
 
   def createTime: Long
   def lastAccessTime: Long
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/Session.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/Session.scala
@@ -26,7 +26,7 @@ trait Session {
 
   def protocol: TProtocolVersion
   def handle: SessionHandle
-  def name: String
+  def name: Option[String]
 
   def conf: Map[String, String]
 
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/Session.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/Session.scala
@@ -26,13 +26,13 @@ trait Session {
 
   def protocol: TProtocolVersion
   def handle: SessionHandle
+  def name: String
 
   def conf: Map[String, String]
 
   def user: String
   def password: String
   def ipAddress: String
-  def name: String
 
   def createTime: Long
   def lastAccessTime: Long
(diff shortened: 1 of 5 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/session/Session.scala
```diff
@@ -32,6 +32,7 @@ trait Session {
   def user: String
   def password: String
   def ipAddress: String
+  def name: String
```
yaooqinn, line 35: mv to right behind handle

yaooqinn (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -139,7 +139,9 @@ class KyuubiSyncThriftClient private (
       .map(_.get("kyuubi.engine.id"))
 
     engineAliveProbeClient.foreach { aliveProbeClient =>
+      val sessionName = SessionHandle.apply(_remoteSessionHandle).identifier + "_aliveness_probe"
```
yaooqinn, line 142: what is this for？

turboFei: to bound with the main thrift session.

yaooqinn:
cc @wForget

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
yaooqinn: ok

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -121,7 +121,7 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
     if (session == null) {
       throw KyuubiSQLException(s"Invalid $sessionHandle")
     }
-    info(s"$sessionHandle is closed, current opening sessions $getOpenSessionCount")
+    info(s"$sessionHandle/${session.name} is closed, current opening sessions $getOpenSessionCount")
```
wForget, line 124: ditto

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -100,7 +100,7 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
       val handle = session.handle
       session.open()
       setSession(handle, session)
-      info(s"$user's session with $handle is opened, current opening sessions" +
+      info(s"$user's session with $handle/${session.name} is opened, current opening sessions" +
```
wForget, line 103: nit: session.name is not always configured, which may make the log dirty.

wForget (approved):
LGTM, thanks

turboFei:
thanks, merged to master

PR #2856, opened 2022-06-10
[KYUUBI #2708] Open engine session and renew engine credentials in the one

### LINKED ISSUE #2708: [Improvement] Open engine session and send credentials in the one
### Code of Conduct

### Search before asking

This is a follow up of #2059

In that pr, we send credentials when opening session.

But openEngineSession and sendCredentials are not in one rpc request.

We might meet corner case when hive delegation is expired.

For example:

If hive delegation token is expired, we will lose the shared metastore client.

And when opening engine session, it will create a sparkSession, which will recreate the shared metastore client.

But it will fail because the old hive delegation token is expired, but it has not received the new hive delegation token yet.

So, the openEngineSession operation will fail.

So, we need gather openEngineSession and sendCredentials in the one to fix the corner case.

### How should we improve?

_No response_

turboFei:

To close #2708 

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -179,8 +179,9 @@ object SparkSQLEngine extends Logging {
   }
 
   def createSpark(): SparkSession = {
-    val engineCredentials = _sparkConf.getOption(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
-    _sparkConf.remove(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
+    val engineCredentials = kyuubiConf.getOption(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
+    kyuubiConf.unset(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
+    _sparkConf.remove(s"spark.${KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY}")
 
     val session = SparkSession.builder.config(_sparkConf).getOrCreate
 
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -31,7 +31,7 @@ import org.apache.spark.sql.SparkSession
 
 import org.apache.kyuubi.{KyuubiException, Logging, Utils}
 import org.apache.kyuubi.Utils._
-import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.config.{KyuubiConf, KyuubiReservedKeys}
 import org.apache.kyuubi.config.KyuubiConf._
 import org.apache.kyuubi.config.KyuubiReservedKeys.KYUUBI_ENGINE_SUBMIT_TIME_KEY
 import org.apache.kyuubi.engine.spark.SparkSQLEngine.{countDownLatch, currentEngine}
@@ -179,7 +179,15 @@ object SparkSQLEngine extends Logging {
   }
 
   def createSpark(): SparkSession = {
+    val engineCredentials = _sparkConf.getOption(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
+    _sparkConf.remove(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
+
     val session = SparkSession.builder.config(_sparkConf).getOrCreate
+
+    engineCredentials.filter(_.nonEmpty).foreach { credentials =>
+      SparkTBinaryFrontendService.renewDelegationToken(session.sparkContext, credentials)
+    }
+
     KyuubiSparkUtil.initializeSparkSession(
       session,
       kyuubiConf.get(ENGINE_INITIALIZE_SQL) ++ kyuubiConf.get(ENGINE_SESSION_INITIALIZE_SQL))
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
@@ -17,8 +17,6 @@
 
 package org.apache.kyuubi.engine.spark
 
-import scala.collection.JavaConverters._
-
 import org.apache.hadoop.io.Text
 import org.apache.hadoop.security.{Credentials, UserGroupInformation}
 import org.apache.hadoop.security.token.{Token, TokenIdentifier}
@@ -26,7 +24,7 @@ import org.apache.hive.service.rpc.thrift.{TOpenSessionReq, TOpenSessionResp, TR
 import org.apache.spark.kyuubi.SparkContextHelper
 
 import org.apache.kyuubi.KyuubiSQLException
-import org.apache.kyuubi.config.KyuubiReservedKeys
+import org.apache.kyuubi.config.KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY
 import org.apache.kyuubi.ha.client.{EngineServiceDiscovery, ServiceDiscovery}
 import org.apache.kyuubi.service.{Serverable, Service, TBinaryFrontendService}
 import org.apache.kyuubi.service.TFrontendService._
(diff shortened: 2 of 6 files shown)


turboFei:
cc @lightning-L  @cxzl25

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
```diff
@@ -75,6 +64,13 @@ class SparkTBinaryFrontendService(
       val respConfiguration = new java.util.HashMap[String, String]()
       respConfiguration.put("kyuubi.engine.id", sc.applicationId)
 
+      if (req.getConfiguration != null) {
```
cxzl25, line 67: Now the token is updated using serialized credentials in `openSession`, but the initial engine initialization sql (`kyuubi.engine.initialize.sql`) may fail to execute. 
For example, in master local mode, the engine does not have a Hive delegation token.

turboFei: cc @yaooqinn

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
```diff
@@ -65,10 +63,8 @@ class SparkTBinaryFrontendService(
       respConfiguration.put("kyuubi.engine.id", sc.applicationId)
 
       if (req.getConfiguration != null) {
-        req.getConfiguration.asScala.get(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY).filter(
-          _.nonEmpty).foreach { credentials =>
-          renewDelegationToken(credentials)
-        }
+        val credentials = req.getConfiguration.remove(KYUUBI_ENGINE_CREDENTIALS_KEY)
```
turboFei, line 66: Not all Map support `remove` operation, but here, it seems that the thrift request supports `remove`.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -106,7 +108,9 @@ class KyuubiSessionImpl(
           Option(password).filter(_.nonEmpty).getOrElse("anonymous")
         }
       try {
-        _client = KyuubiSyncThriftClient.createClient(user, passwd, host, port, sessionConf)
+        val engineSessionConf = sessionConf.clone
+        engineSessionConf.set(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY, engineCredentials)
```
wForget, line 112: The `engineSessionConf` doesn't seem to be passed to engine, put it in `optimizedConf`?

turboFei: Seems we can not put it into optimizedConf, it might be passed to SparkProcessBuilder and will be passed to engine.

turboFei: How about:
```
        sessionConf.set(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY, engineCredentials)
        _client = KyuubiSyncThriftClient.createClient(user, passwd, host, port, sessionConf)
```
cc @yaooqinn

turboFei: got it, thanks

wForget: Sorry, I didn't express my idea exactly. I mean that we need to pass this configuration through the 'openSession' method instead of the 'createClient'`.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
yaooqinn: we can also pass token via startup spark configuations

ulysses-you: @turboFei seems this comment has not been addressed, can you look at it ?

turboFei: ok

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -179,7 +179,13 @@ object SparkSQLEngine extends Logging {
   }
 
   def createSpark(): SparkSession = {
+    val credentials = _sparkConf.getOption(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
+    _sparkConf.remove(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
+
     val session = SparkSession.builder.config(_sparkConf).getOrCreate
+
+    credentials.foreach(SparkTBinaryFrontendService.renewDelegationToken(session.sparkContext, _))
```
turboFei, line 187: cc @yaooqinn @ulysses-you

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -179,7 +179,15 @@ object SparkSQLEngine extends Logging {
   }
 
   def createSpark(): SparkSession = {
+    val engineCredentials = _sparkConf.getOption(KyuubiReservedKeys.KYUUBI_ENGINE_CREDENTIALS_KEY)
```
yaooqinn, line 182: this is wrong, you should get the creds from kyuubiConf

turboFei: thanks, fixed

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -67,6 +68,9 @@ class KyuubiSessionImpl(
     case (key, value) => sessionConf.set(key, value)
   }
 
+  private val engineCredentials = renewEngineCredentials()
+  sessionConf.set(KYUUBI_ENGINE_CREDENTIALS_KEY, engineCredentials)
```
ulysses-you, line 72: unnecessary ?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -107,7 +111,11 @@ class KyuubiSessionImpl(
         }
       try {
         _client = KyuubiSyncThriftClient.createClient(user, passwd, host, port, sessionConf)
-        _engineSessionHandle = _client.openSession(protocol, user, passwd, optimizedConf)
+        _engineSessionHandle = _client.openSession(
+          protocol,
+          user,
+          passwd,
+          optimizedConf ++ Map(KYUUBI_ENGINE_CREDENTIALS_KEY -> engineCredentials))
```
ulysses-you, line 118: I think a more safer way is that check `engineCredentials` is vaildated before inject into engine, so if engine get the KYUUBI_ENGINE_CREDENTIALS_KEY that means it must has credentials.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
ulysses-you: ah, I missed read the code.. nvm

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -98,6 +98,19 @@ class KyuubiSessionImpl(
...
+      val engineCredentials = renewEngineCredentials()
+      var openEngineSessionConf = optimizedConf
+      if (engineCredentials.nonEmpty) {
+        sessionConf.set(KYUUBI_ENGINE_CREDENTIALS_KEY, engineCredentials)
+        openEngineSessionConf =
+          optimizedConf ++ Map(KYUUBI_ENGINE_CREDENTIALS_KEY -> engineCredentials)
+      }
+
+      engine = new EngineRef(
+        sessionConf,
```
turboFei, line 110: how about this?

@ulysses-you 
we need transfer the config item to EngineRef

ulysses-you: can we poll out renewEngineCredentials ? so we do not need change the initialization of EngineRef. Then the ordering looks like:

renewEngineCredentials -> new EngineRef -> openEngineSession

ulysses-you (approved):

turboFei:
thanks, merged to master

turboFei:
The kyuubi.engine.credentials is still visible in spark ui.

Issue #2628, opened 2022-05-11
[SUB-TASK][KPIP-4] Implement restClient for batch job submission


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

sub-task of #2240 

Thanks the docs drafted by @ulysses-you 
https://docs.google.com/document/d/1GQQXzC0GdXdKwGI0DiO04pa9N3AAixuuo1B4j6EI2AA/edit#heading=h.e8opx7al2u29

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

ulysses-you:
@turboFei do you want add a rest client module for the functionality of sdk ?

turboFei:
@turboFei do you want add a rest client module for the functionality of sdk ?

Yeah.

ulysses-you:
I'm fine with this, and for user-faced interface, we need better docs. cc @yaooqinn

turboFei:
cc @lightning-L are you interested in this?

lightning-L:
> cc @lightning-L are you interested in this?

of course..

turboFei:
let us wait for more discussions on this.

turboFei:
Please refer the docs: https://docs.google.com/document/d/1GQQXzC0GdXdKwGI0DiO04pa9N3AAixuuo1B4j6EI2AA/edit#heading=h.e8opx7al2u29

```
  Support kyuubi batch command in existing kyuubi-ctl
  From:
   Usage: kyuubi-ctl [create|get|delete|list] [server|engine] [options]

   To:

   Usage: kyuubictl [create|get|delete|list|logs|submit] [resource] [options]
   resources: [server|engine|batch]
```

PR #2892, opened 2022-06-16
[KYUUBI #2628][FOLLOWUP] Reuse the kyuubi-ctl batch commands for SubmitBatchCommand

turboFei:

For SubmitBatchCommand, reuse the CreateBatchCommand and LogBatchCommand.


kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/BatchCommand.scala
```diff
@@ -0,0 +1,33 @@
...
+
+abstract class BatchCommand(cliConfig: CliConfig) extends Command(cliConfig) {
+
+  override def run(): Unit = {
+    doRunAndReturnBatchReport.foreach { batch =>
+      info(Render.renderBatchInfo(batch))
+    }
+  }
+
+  def doRunAndReturnBatchReport(): Option[Batch]
```
turboFei, line 32: now ListBatchesCommand does not extend `BatchCommand`

turboFei: cc @ulysses-you

turboFei: ListBatchesCommand should be `BatchesCommand`

turboFei:
cc @lightning-L 
Hope it can simplify the following optimization for SubmitBatchCommand and LogBatchCommand.

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/delete/DeleteBatchCommand.scala
```diff
@@ -35,16 +36,22 @@ class DeleteBatchCommand(cliConfig: CliConfig) extends Command(cliConfig) {
...
-        }
+      if (result.isSuccess) {
+        null
+      } else {
+        batchRestApi.getBatchById(batchId)
       }
     }
   }
 
+  override def run(): Unit = {
```
ulysses-you, line 47: can not we do this inside `doRunAndReturnBatchReport ` ?

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/BatchCommand.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import org.apache.kyuubi.client.api.v1.dto.Batch
+import org.apache.kyuubi.ctl.CliConfig
+import org.apache.kyuubi.ctl.util.Render
+
+abstract class BatchCommand(cliConfig: CliConfig) extends Command(cliConfig) {
+
+  def doRunAndReturnBatchReport(): Batch
+
+  override def run(): Unit = {
```
ulysses-you, line 28: shall we make it `final` ? otherwise it just add complexity ..

turboFei: ok

lightning-L:
I agree that we need to reuse the code in CreateBatchCommand and LogBatchCommand for submit batch operation.
But I am wondering if it is necessary to add a BatchCommand.

I think the main point is that we need to separate the calling api(do Run) part and the render part. Currently, those are stuck together in run() method.

For all the Command, the operation is: validate, doRun, and render.
We could separate the current run() method into doRun and render.
We could let doRun return object, and let render print the object.
It will apply to all kinds of resources. So we could do it in Command.scala instead of adding a new BatchCommand.scala?

turboFei:
> I agree that we need to reuse the code in CreateBatchCommand and LogBatchCommand for submit batch operation. But I am wondering if it is necessary to add a BatchCommand.
> 
> I think the main point is that we need to separate the calling api(do Run) part and the render part. Currently, those are stuck together in run() method.
> 
> For all the Command, the operation is: validate, doRun, and render. We could separate the current run() method into doRun and render. We could let doRun return object, and let render print the object. It will apply to all kinds of resources. So we could do it in Command.scala instead of adding a new BatchCommand.scala?

thanks @lightning-L

turboFei:
thanks @lightning-L 
could you have a review again?

ulysses-you (approved):

turboFei:
thanks, merging to master

PR #2908, opened 2022-06-18
[SUB-TASK][KPIP-4] Get batch application status from resource manager for across instances get batch request

turboFei:

Now, we do not update the metadata if there is no batch operation state change.

For the batch application state change, it might no be updated in time.
For example:
-  SUBMITTED => ACCEPTED
-  SUBMITTED => RUNNING

So, it is better to get batch application status from resource manager for get batch request.

BTW, there is some corner case.

For example:
We deploy kyuubi on k8s, if the pod re-schedule and it stuck in provision phase caused by resource insufficient or other issue.

The kyuubi server can not recover in short-time.

We need fetch the application status from resource manager and return the correct batch report to prevent to block the batch client, which might wait the batch finished.

```
kyuubi-ctl submit batch -f batch.yaml --waitCompletion=true
```

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -76,7 +76,7 @@ class BatchJobSubmission(
 
   private[kyuubi] val batchId: String = session.handle.identifier.toString
 
-  private var applicationStatus: Option[Map[String, String]] = None
+  private[kyuubi] var applicationStatus: Option[Map[String, String]] = None
 
   private var killMessage: KillResponse = (false, "UNKNOWN")
   def getKillMessage: KillResponse = killMessage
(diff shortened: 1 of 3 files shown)


turboFei:
> We deploy kyuubi on k8s, if the pod re-schedule and it stuck in provision phase caused by resource insufficient or other issue.
> The kyuubi server can not recover in short-time.


BTW, for this case, we need support to kill the batch job on the kyuubi instance(stuck in pod provision).

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -246,6 +238,7 @@ class BatchJobSubmission(
       throw new RuntimeException(s"$batchType batch[$batchId] job failed:" +
         applicationStatus.get.mkString(","))
     } else {
+      updateBatchMetadata()
```
turboFei, line 241: if the batch is recovered and before its state is PENDING.

We need update the state to RUNNING if get applicationId from resource manager before recovery.

turboFei:
> > We deploy kyuubi on k8s, if the pod re-schedule and it stuck in provision phase caused by resource insufficient or other issue.
> > The kyuubi server can not recover in short-time.
> 
> BTW, for this case, we need support to kill the PENDING state batch job on the kyuubi instance(stuck in pod provision).


How do you think about? @ulysses-you

turboFei:
move to #2911

PR #2910, opened 2022-06-18
[SUB-TASK][KPIP-4] Support to mark the batch session closed by remote kyuubi instance if it is not reachable

turboFei:

For batch session, it is not stateless, and some operations, such as close PENDING batch session, must be processed in the local kyuubi instance.
 But sometimes, the kyuubi instance might be unreachable, we need mark the batch session be CLOSED state in remote kyuubi instance.

In this pr, I introduce a new field `remoteClosed` to present whether it has been marked as CLOSED state by remote kyuubi instance.

The `remoteClosed` should only be updated by remote kyuubi instance and set its value to true, so there is no race condition.

- When opening the batch session in recovery mode, if `remoteClosed` is true, set the state to CANCELED directly
- check the local batch sessions that is marked as remoteClosed periodically and close them.


turboFei:
close this one, move to #2911

PR #2911, opened 2022-06-19
[SUB-TASK][KPIP-4] If the kyuubi instance unreachable, support to backfill state from resource manager and mark batch closed by remote kyuubi instance 

turboFei:

Now for the batch state, we rely on the metadata store, which is only updated by the local kyuubi instance.

But if the kyuubi instance is down or unreachable sometimes, the metadata state will not change and we can not get the correct state, it might cause the client side stuck if it is waiting the batch completion.

So, if the kyuubi instance is not reachable, we need backfill the batch state according to resource manager batch application state it the batch was in RUNNING state.

If the batch was stuck in PENDING state and the kyuubi instance can not get recovered in short-time, we need support to close the batch session and prevent the batch job re-submitted if it get recovered.

In this pr, I introduce a new field `remoteClosed` to present whether it has been marked as CLOSED state by remote kyuubi instance.

The `remoteClosed` should only be updated by remote kyuubi instance and set its value to true, so there is no race condition.

- When opening the batch session in recovery mode, if `remoteClosed` is true, set the state to CANCELED directly
- check the local batch sessions that is marked as remoteClosed periodically and close them.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -64,6 +64,7 @@ class BatchJobSubmission(
     batchArgs: Seq[String],
     recoveryMetadata: Option[Metadata])
   extends KyuubiOperation(OperationType.UNKNOWN_OPERATION, session) {
+  import BatchJobSubmission._
 
   override def statement: String = "BATCH_JOB_SUBMISSION"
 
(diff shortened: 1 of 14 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -246,6 +241,7 @@ class BatchJobSubmission(
       throw new RuntimeException(s"$batchType batch[$batchId] job failed:" +
         applicationStatus.get.mkString(","))
     } else {
+      updateBatchMetadata()
```
turboFei, line 244: if the batch is recovered and before its state is PENDING.

We need update the state to RUNNING if get applicationId from resource manager before recovery.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -77,6 +79,34 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       batchOpStatus.completed)
   }
 
+  private def buildBatch(
+      metadata: Metadata,
+      batchAppStatus: Option[Map[String, String]]): Batch = {
+    batchAppStatus.map { appStatus =>
+      val currentBatchState =
+        if (BatchJobSubmission.applicationFailed(batchAppStatus)) {
```
turboFei, line 87: backfill the batch state according to resource manager state.

yaooqinn (approved):

ulysses-you (approved):

turboFei:
thanks, merging to master

Issue #2924, opened 2022-06-21
[Bug] Correct the frontend server start state


ulysses-you:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

CI has a lot of flaky test, it is because we will mark server is started when we call thread.start, but the server (bianry or http) may not be work at that moment. As a result, the client can not connect a started server.

### Affects Version(s)

master/1.5

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #2925, opened 2022-06-21
[KYUUBI #2924] Correct the frontend server start state

ulysses-you:

CI has a lot of flaky test, it is because we will mark server is started when we call thread.start, but the server (bianry or http) may not be work at that moment. As a result, the client can not connect a started server.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
@@ -34,7 +34,7 @@ import org.eclipse.jetty.util.security.Constraint
 import org.eclipse.jetty.util.ssl.SslContextFactory
 import org.eclipse.jetty.util.thread.ExecutorThreadPool
 
-import org.apache.kyuubi.{KyuubiException, Utils}
+import org.apache.kyuubi.KyuubiException
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf._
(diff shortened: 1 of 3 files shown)


yaooqinn:
we can simply sleep 3-5s for tests?

ulysses-you:
>  we can simply sleep 3-5s for tests?

@yaooqinn this pr is also not very complex. I'm afraid the temporary fix will not fix the flaky in future.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
```diff
@@ -215,6 +215,10 @@ final class KyuubiTHttpFrontendService(
         info(s"Starting and exposing JDBC connection at: jdbc:hive2://$connectionUrl/")
       }
       server.foreach(_.start())
+      // wait a moment for server thread starting ..
+      if (Utils.isTesting) {
+        Thread.sleep(3000)
```
ulysses-you, line 220: @yaooqinn  The jetty server state seems also not very accurate in a simple way.. I'm going back to add some sleep for test code path with HTTP mode. I'm not add this for binary mode for now and we can add if we see other flaky test in future.

yaooqinn: you can do it in test..

ulysses-you: changed..

turboFei (approved):

ulysses-you:
thanks, merging to master

PR #2934, opened 2022-06-22
[KYUUBI #2927][1.5] Fix the thread in ScheduleThreadExecutorPool can't be shutdown immediately

### LINKED ISSUE #2927: [Bug] `shutdownTimeoutMonitor` thread can't be shut down immediately when query statement is finished
### Code of Conduct

### Search before asking

### Describe the bug

``` 
protected def addTimeoutMonitor(queryTimeout: Long): Unit = {
    if (queryTimeout > 0) {
      val timeoutExecutor =
        ThreadUtils.newDaemonSingleThreadScheduledExecutor("query-timeout-thread")
      val action: Runnable = () => cleanup(OperationState.TIMEOUT)
      timeoutExecutor.schedule(action, queryTimeout, TimeUnit.SECONDS)
      statementTimeoutCleaner = Some(timeoutExecutor)
    }
  }

protected def shutdownTimeoutMonitor(): Unit = {
    statementTimeoutCleaner.foreach(_.shutdown())
}
```
in `AbstractionOperation` which be used in spark/trino/jdbc/hive engine's `ExecuteStatement`. 
`shutdownTimeoutMonitor` is a `ScheduleThreadExecutorPool`, if the  `queryTimeout`  exceeds the execution time of the statement, even if the `shutDown` is called, it will wait for the delayed task to be executed and then exit. 
In the scenario of high concurrency and short sql execution, this can lead to a temporary leak of resources in the statement thread pool.

Refer to `ScheduleThreadExecutorPool` of **setExecuteExistingDelayedTasksAfterShutdownPolicy** method: 
Sets the policy on whether to execute existing delayed tasks even when this executor has been shutdown. In this case, these tasks will only terminate upon shutdownNow, or after setting the policy to false when already shutdown. **This value is by default true**.

We can set the policy to false or using `shutdownNow` to replace `shutdown` to fix it.

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

iodone:
Close #2927 for branch-1.5

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -179,7 +179,7 @@ class ExecuteStatement(
   private def addTimeoutMonitor(): Unit = {
     if (queryTimeout > 0) {
       val timeoutExecutor =
-        ThreadUtils.newDaemonSingleThreadScheduledExecutor("query-timeout-thread")
+        ThreadUtils.newDaemonSingleThreadScheduledExecutor("query-timeout-thread", false)
       val action: Runnable = () => cleanup(OperationState.TIMEOUT)
       timeoutExecutor.schedule(action, queryTimeout, TimeUnit.SECONDS)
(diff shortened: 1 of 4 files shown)


ulysses-you (approved):

ulysses-you:
thanks, merging to branch-1.5

PR #2946, opened 2022-06-25
Add retrying kyuubi sync thrift client for engine connection

turboFei:


turboFei:
test passed.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1540,6 +1525,27 @@ object KyuubiConf {
...
+    buildConf("kyuubi.session.reserve.on.broken.enabled")
+      .internal
+      .doc("Whether to reserve the session on the transport protocol broken. Now we support" +
+        " RetryingKyuubiSyncThriftClient, it will close the recreate the protocol for thrift" +
+        " request retry, and we need reserve the session in engine side for that.")
+      .version("1.6.0")
+      .booleanConf
+      .createWithDefault(false)
+
+  val SESSION_RESERVE_ON_BROKEN_TIMEOUT: ConfigEntry[Long] =
```
yaooqinn, line 1538: `kyuubi.session.reserve.*` are unnecessary as a retry logic

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -160,13 +163,32 @@ abstract class TFrontendService(name: String)
...
-      userName,
-      req.getPassword,
-      ipAddress,
-      configuration)
-    sessionHandle
+
+    val sessionHandleGuid =
+      configuration.get(KyuubiReservedKeys.KYUUBI_SESSION_HANDLE_GUID)
+    val sessionHandleSecret =
+      configuration.get(KyuubiReservedKeys.KYUUBI_SESSION_HANDLE_SECRET)
```
yaooqinn, line 170: in which case, we need these?

turboFei: recreate the protocol and connect to the old session

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: it is used to tell the engine front end service that do not close the session on protocol broken.

Otherwise, the session will be closed.

yaooqinn: the session will be closed on disconnected with these..

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
yaooqinn: in which case we need the original session?

turboFei: the original session means original context, If we create a new session, the context is loss

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: yeah, so we need this config to reserve the old context.

turboFei:
seems not needed, close it

PR #2948, opened 2022-06-27
Remove thrift request timeout for KyuubiSyncThriftClient

turboFei:

Remove thrift request timeout for KyuubiSyncThriftClient

### DIFF
--- a/docs/deployment/settings.md
+++ b/docs/deployment/settings.md
@@ -422,7 +422,6 @@ kyuubi.session.engine.initialize.timeout|PT3M|Timeout for starting the backgroun
 kyuubi.session.engine.launch.async|true|When opening kyuubi session, whether to launch backend engine asynchronously. When true, the Kyuubi server will set up the connection with the client without delay as the backend engine will be created asynchronously.|boolean|1.4.0
 kyuubi.session.engine.log.timeout|PT24H|If we use Spark as the engine then the session submit log is the console output of spark-submit. We will retain the session submit log until over the config value.|duration|1.1.0
 kyuubi.session.engine.login.timeout|PT15S|The timeout of creating the connection to remote sql query engine|duration|1.0.0
-kyuubi.session.engine.request.timeout|PT0S|The timeout of awaiting response after sending request to remote sql query engine|duration|1.4.0
 kyuubi.session.engine.share.level|USER|(deprecated) - Using kyuubi.engine.share.level instead|string|1.0.0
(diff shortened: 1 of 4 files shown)


yaooqinn (approved):

turboFei:
thanks, merged to master

risyomei:
@yaooqinn @turboFei 
Hello, Sorry to disturb you while you are busy.

May I know the motivation for deleting the timeout?

turboFei:
> @yaooqinn @turboFei 
> 
> Hello, Sorry to disturb you while you are busy.
> 
> 
> 
> May I know the motivation for deleting the timeout?

Remove it to prevent out of sequence issue.

turboFei:
The thrift protocol does not support retry.

risyomei:
@turboFei 
Thank you very much for your response.
Do you have a ticket to describe the "out of sequence issue"?

turboFei:
> Do you have a ticket to describe the "out of sequence issue"?

about `out of sequence issue`.
```
In the context of Apache Thrift, "out of sequence issue" can refer to a situation where a Thrift server receives a message from a client that is out of order, meaning the server receives a message with a sequence number that is lower than expected.

This can happen if there is a delay or a disruption in the network communication between the client and the server, causing messages to be delivered out of order. If the server is not able to handle such out of sequence messages properly, it can result in incorrect behavior or even crashes.

To address this issue, Thrift includes a sequence number in each message to help the server determine if a message is out of order. If a server receives an out of sequence message, it can respond with an error message indicating the issue, or it can buffer the message until the expected sequence number is received.

Thrift also provides an option for clients to request ordered delivery of messages, which instructs the server to buffer and deliver messages in the correct order, even if they arrive out of sequence. This option can be useful in situations where network disruptions are common, but it can also impact performance, as buffering and reordering messages can introduce additional latency.
```

For example: 

1. client send fetchResult request  -> timeout , then client does not receive the fetchResult response.
2. client send getOpStatus request -> then receive the last `fetchResult response`(out of sequence because expected result is getOpStatus response`

risyomei:
@turboFei 
Thank you very much for your explanation.
It's very easy to follow!

> client send fetchResult request -> timeout , then client does not receive the fetchResult response.
client send getOpStatus request -> then receive the last fetchResult response(out of sequence because expected result is getOpStatus response`

By the way, I have found a possible issue (https://github.com/apache/kyuubi/issues/4457) that are kind of related to this issue.
It will be very helpful if you can take a look at that as well 🙇‍♂️

PR #2952, opened 2022-06-27
Remove OperationType from OperationHandle for simplification

yaooqinn:

The Operation Type is only used for converting `TOperationType` through specific Thrift Operation APIs. So we can TOperationType in these APIs directly, such as, during calling TFrontendService.ExecuteStatement, we can set TOperationType.EXECUTE_STATEMENT at the end. There is no need to add it to OperationType.

This PR removes `OperationType` completely to simply `OperationHandle` which benefits other `kyuubi.frontend.protocols`.

The thrift protocol shall be able to be removed in following PRs.


yaooqinn:
cc @turboFei @ulysses-you

cxzl25 (approved):
Checked the PR locally. LGTM

turboFei (approved):
LGTM

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
```diff
@@ -30,17 +30,13 @@ import org.apache.kyuubi.metrics.MetricsConstants.{OPERATION_FAIL, OPERATION_OPE
...
-abstract class KyuubiOperation(opType: OperationType, session: Session)
-  extends AbstractOperation(opType, session) {
+abstract class KyuubiOperation(session: Session)
+  extends AbstractOperation(session) {
 
-  private val opTypeName = (opType match {
-    case OperationType.UNKNOWN_OPERATION => statement
-    case _ => opType.toString
-  }).toLowerCase
+  private val opTypeName = getClass.getSimpleName
```
ulysses-you, line 39: nit: `super.opType`

ulysses-you (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
yaooqinn: removed this and use opType directly

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationHandle.scala
```diff
@@ -40,35 +38,29 @@ case class OperationHandle(
...
     tOperationHandle.setOperationId(Handle.toTHandleIdentifier(identifier))
-    tOperationHandle.setOperationType(OperationType.toTOperationType(typ))
     tOperationHandle.setHasResultSet(_hasResultSet)
     tOperationHandle
   }
 
-  override def hashCode(): Int = {
-    (Objects.hashCode(identifier) + 31) * 31 + Objects.hashCode(typ)
-  }
+  override def hashCode(): Int = Objects.hashCode(identifier) + 31
```
cfmcgrady, line 45: unnecessary override?

cfmcgrady: unnecessary `+ 31`?

yaooqinn: updated

yaooqinn:
@cxzl25 @turboFei @ulysses-you @cfmcgrady thanks for the review. merged to master

PR #2953, opened 2022-06-27
Support to interrupt the thrift request if remote engine is broken

turboFei:

The engine request timeout was removed by #2948

We need rely on the aliveness probe to interrupt the thrift request if remote engine is broken.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
@@ -38,6 +38,7 @@ import org.apache.kyuubi.session.SessionHandle
 import org.apache.kyuubi.util.{ThreadUtils, ThriftUtils}
 
 class KyuubiSyncThriftClient private (
+    sessionHandle: SessionHandle,
     protocol: TProtocol,
     maxAttempts: Int,
     engineAliveProbeProtocol: Option[TProtocol],
@@ -57,7 +58,7 @@ class KyuubiSyncThriftClient private (
   @volatile private var engineLastAlive: Long = _
 
   private lazy val asyncRequestExecutor =
-    ThreadUtils.newDaemonSingleThreadScheduledExecutor("async-request-pool-" + _remoteSessionHandle)
+    ThreadUtils.newDaemonSingleThreadScheduledExecutor("async-request-executor-" + sessionHandle)
 
   private def startEngineAliveProbe(): Unit = {
     engineAliveThreadPool = ThreadUtils.newDaemonSingleThreadScheduledExecutor(
@@ -440,6 +441,7 @@ private[kyuubi] object KyuubiSyncThriftClient extends Logging {
   }
 
   def createClient(
+      sessionHandle: SessionHandle,
       user: String,
       password: String,
       host: String,
@@ -471,6 +473,7 @@ private[kyuubi] object KyuubiSyncThriftClient extends Logging {
         None
       }
     new KyuubiSyncThriftClient(
+      sessionHandle,
       tProtocol,
       requestMaxAttempts,
       aliveProbeProtocol,
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
@@ -17,13 +17,11 @@
 
 package org.apache.kyuubi.client
 
-import java.util.concurrent.{ScheduledExecutorService, TimeUnit}
+import java.util.concurrent.{Future, ScheduledExecutorService, TimeUnit}
 import java.util.concurrent.locks.ReentrantLock
 
 import scala.collection.JavaConverters._
-import scala.concurrent.Promise
 import scala.concurrent.duration.Duration
-import scala.util.Try
 
 import org.apache.hive.service.rpc.thrift._
 import org.apache.thrift.TException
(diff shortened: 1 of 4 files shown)


yaooqinn:
can we test this?

turboFei:
> can we test this?

The code should has been covered by UT.

Do you mean integration testing?

yaooqinn:
The code should has been covered by UT.

which one？w/ or w/o this feature, it passes?

turboFei:
How can we simulate the case that the network between kyuubi server and kyuubi engine break?

yaooqinn:
> How can we simulate the case that the network between kyuubi server and kyuubi engine break?

`select java_method('System.exit',...`

turboFei:
https://github.com/apache/incubator-kyuubi/blob/994dc6ce849f4502e2a9060cc4e5246967eb03d3/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerConnectionSuite.scala#L50-L80

It has been defined in KyuubiOperationPerConnectionSuite

turboFei:
this pr is to address the case:
- the network break and the client side does not aware that and block until os level  TPC keepalive timeout

yaooqinn:
Not exactly, the test is not for your PR, you need to make it crash in the middle of an RPC call

turboFei:
UT added.

turboFei:
should we enable alive probe by defaults? @yaooqinn

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
```diff
@@ -153,4 +154,44 @@ class KyuubiOperationPerUserSuite extends WithKyuubiServer with SparkQueryTests
...
+        val preReq = new TExecuteStatementReq()
+        preReq.setStatement("select engine_name()")
+        preReq.setSessionHandle(handle)
+        preReq.setRunAsync(false)
+        client.ExecuteStatement(preReq)
+
+        new Thread("async-exit") {
+          override def run(): Unit = {
+            withSessionHandle { (client2, handle2) =>
+              Thread.sleep(2000)
```
yaooqinn, line 175: Instead of `Thread.sleep(2000) / java.lang.Thread', 'sleep', 3600000l)` with flakiness, I guess we can find a better way

turboFei: how about now?

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
```diff
@@ -153,4 +154,38 @@ class KyuubiOperationPerUserSuite extends WithKyuubiServer with SparkQueryTests
...
+        systemExitReq.setRunAsync(true)
+        client.ExecuteStatement(systemExitReq)
+
+        val executeStmtReq = new TExecuteStatementReq()
+        executeStmtReq.setStatement("SELECT java_method('java.lang.Thread', 'sleep', 30000l)")
+        executeStmtReq.setSessionHandle(handle)
+        executeStmtReq.setRunAsync(false)
+        val startTime = System.currentTimeMillis()
+        val executeStmtResp = client.ExecuteStatement(executeStmtReq)
+        assert(executeStmtResp.getStatus.getStatusCode === TStatusCode.ERROR_STATUS)
```
yaooqinn, line 185: how about we check the error too

turboFei:
https://community.cloudera.com/t5/Support-Questions/Hive-http-transport-mode-problem/td-p/136746

Invalid status 72

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationPerUserSuite.scala
```diff
@@ -153,4 +154,44 @@ class KyuubiOperationPerUserSuite extends WithKyuubiServer with SparkQueryTests
...
+          executeStmtReq.setRunAsync(false)
+          val startTime = System.currentTimeMillis()
+          val executeStmtResp = client.ExecuteStatement(executeStmtReq)
+          assert(executeStmtResp.getStatus.getStatusCode === TStatusCode.ERROR_STATUS)
+          assert(executeStmtResp.getStatus.getErrorMessage.contains(
+            "java.net.SocketException: Connection reset") ||
+            executeStmtResp.getStatus.getErrorMessage.contains(
+              "Caused by: java.net.SocketException: Broken pipe (Write failed)"))
+          val elapsedTime = System.currentTimeMillis() - startTime
+          assert(elapsedTime < 20 * 1000)
```
turboFei, line 192: seems difficult to simulate the socket stuck case.

@yaooqinn  should we remove this UT?

yaooqinn: create a "socket read timeout" err?

turboFei: ok

turboFei: tried to close the underlying transport socket, but the result is not expected.

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -56,6 +58,24 @@ class KyuubiSyncThriftClient private (
   private var engineAliveThreadPool: ScheduledExecutorService = _
   @volatile private var engineLastAlive: Long = _
 
+  private var asyncRequestExecutor: ScheduledExecutorService = _
```
yaooqinn, line 61: Use something like com.google.common.util.concurrent.MoreExecutors#sameThreadExecutor？

turboFei: seems that we can interrupt the task running in the same thread executor.

yaooqinn (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
yaooqinn: maybe we don't need the Scheduled interfaces

turboFei: thanks, addressed.

turboFei:
thanks, merged to master

PR #2957, opened 2022-06-28
[SUB-TASK][KPIP-4] Use canonical host name for kyuubi instance

turboFei:

We have to use canonical host name as kyuubi instance, otherwise, the ip is always change after pod restart, and the kyuubi instance can not get recovered.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
@@ -56,7 +56,13 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
 
   override def initialize(conf: KyuubiConf): Unit = synchronized {
     val host = conf.get(FRONTEND_REST_BIND_HOST)
-      .getOrElse(Utils.findLocalInetAddress.getCanonicalHostName)
+      .getOrElse {
+        if (conf.get(KyuubiConf.FRONTEND_CONNECTION_URL_USE_HOSTNAME)) {
+          Utils.findLocalInetAddress.getCanonicalHostName
+        } else {
+          Utils.findLocalInetAddress.getHostAddress
+        }
+      }
     server = JettyServer(getName, host, conf.get(FRONTEND_REST_BIND_PORT))
     super.initialize(conf)
   }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
@@ -56,7 +56,7 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
 
   override def initialize(conf: KyuubiConf): Unit = synchronized {
     val host = conf.get(FRONTEND_REST_BIND_HOST)
-      .getOrElse(Utils.findLocalInetAddress.getHostAddress)
+      .getOrElse(Utils.findLocalInetAddress.getCanonicalHostName)
     server = JettyServer(getName, host, conf.get(FRONTEND_REST_BIND_PORT))
     super.initialize(conf)
   }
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -56,7 +56,7 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
 
   override def initialize(conf: KyuubiConf): Unit = synchronized {
     val host = conf.get(FRONTEND_REST_BIND_HOST)
-      .getOrElse(Utils.findLocalInetAddress.getHostAddress)
+      .getOrElse(Utils.findLocalInetAddress.getCanonicalHostName)
```
ulysses-you, line 59: shall we respect the config `FRONTEND_CONNECTION_URL_USE_HOSTNAME` ? and it seems `getHostName` is enough since the StatefulSet ensures the each pod hostname.

turboFei: no,  I deploy kyuubi in multiple k8s clusters.

It might be:
kyuubidev-0.35.k8s.io
kyuubidev-0.28.k8s.io

turboFei: > getHostName

Seems `getHostName` is a bug for me...

turboFei: shall we modify the `getHostName` to `getCanonicalHostName`? @ulysses-you

cxzl25: `TFrontendService` should have a similar problem.
We encountered this problem when upgrading 1.5 and changed to `getCanonicalHostName`
I see that `KyuubiMySQLFrontendService` still uses `getCanonicalHostName`.

turboFei: > We encountered this problem when upgrading 1.5 and changed to getCanonicalHostName

I also meet that issue and I set `kyuubi.engine.connection.url.use.hostname=false` manually.

> shall we respect the config FRONTEND_CONNECTION_URL_USE_HOSTNAME
I think that we should always use CanonicalHostName for the kyuubiInstance url in metastore.

Because I set `kyuubi.engine.connection.url.use.hostname=false` before, if I respect FRONTEND_CONNECTION_URL_USE_HOSTNAME, I will use ipaddress for kyuubiInstance.

turboFei: the connection url for rest fe is special.

kyuubi-zookeeper/src/main/scala/org/apache/kyuubi/zookeeper/EmbeddedZookeeper.scala
```diff
@@ -43,7 +43,7 @@ class EmbeddedZookeeper extends AbstractService("EmbeddedZookeeper") {
     val maxClientCnxns = conf.get(ZK_MAX_CLIENT_CONNECTIONS)
     val minSessionTimeout = conf.get(ZK_MIN_SESSION_TIMEOUT)
     val maxSessionTimeout = conf.get(ZK_MAX_SESSION_TIMEOUT)
-    host = conf.get(ZK_CLIENT_PORT_ADDRESS).getOrElse(findLocalInetAddress.getHostName)
+    host = conf.get(ZK_CLIENT_PORT_ADDRESS).getOrElse(findLocalInetAddress.getCanonicalHostName)
```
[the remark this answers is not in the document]
turboFei: seems the zk issue is gone.

cxzl25 (approved):

turboFei:
thanks @ulysses-you @pan3793 @zwangsheng @cxzl25 

merging to master

PR #2963, opened 2022-06-29
Bump Delta 2.0.0rc1

cfmcgrady:

Test against delta 2.0.0rc1

https://github.com/delta-io/delta/releases/tag/v2.0.0rc1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,7 @@
         <commons-lang.version>2.6</commons-lang.version>
         <commons-lang3.version>3.10</commons-lang3.version>
         <curator.version>2.12.0</curator.version>
-        <delta.version>1.2.1</delta.version>
+        <delta.version>2.0.0rc1</delta.version>
         <fb303.version>0.9.3</fb303.version>
         <flink.version>1.14.4</flink.version>
         <flink.module.scala.suffix>_${scala.binary.version}</flink.module.scala.suffix>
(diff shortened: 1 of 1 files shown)


pan3793:
Does DeltaLake provide some new features on 2.0? It's better to add some tests for those features as well

cfmcgrady:
> Does DeltaLake provide some new features in 2.0? It's better to add some tests for those features as well

The key features in the changelog are shown as follows


1. Support Change Data Feed on Delta tables. Change Data Feed represents the row level changes between different versions of the table. When enabled, additional information is recorded regarding row level changes for every write operation on the table. See the documentation for more details.

2. Support Z-Order clustering of data to reduce the amount of data read. Z-Ordering is a technique to colocate related information in the same set of files. This data clustering allows column stats (released in Delta 1.2) to be more effective in skipping data based on filters in a query. See the documentation for more details.

3. Support for idempotent writes to Delta tables to enable fault-tolerant retry of Delta table writing jobs without writing the data multiple times to the table. See the documentation for more details.

4. Support for dropping columns in a Delta table as a metadata change operation. This command drops the column from metadata and not the column data in underlying files. See documentation for more details.

5. Support for dynamic partition overwrite. Overwrite only the partitions with data written into them at runtime. See documentation for details.

6. Experimental support for multi-part checkpoints to split the Delta Lake checkpoint into multiple parts to speed up writing the checkpoints and reading. See documentation for more details.

7. Python and Scala API support for OPTIMIZE file compaction and Z-order by.

We can add the tests in a separate PR, WDYT?

pan3793:
> We can add the tests in a separate PR, WDYT?

SGTM, thanks, merging to master

PR #2972, opened 2022-06-29
Using stdout for the output of kyuubi-ctl

turboFei:

Using stdout as the output of kyuubi-ctl.

So that we can using the shell command
```
result=$(kyuubi-ctl action resource)
```

Screen Shot 2022-06-30 at 07 46 42

Screen Shot 2022-06-30 at 07 48 52


pan3793 (approved):

pan3793:
Thanks, merging to master

PR #2975, opened 2022-06-30
Code improvement in rest client

pan3793:

Small code improvements and clean up in rest client

### DIFF
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/HttpClientFactory.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/HttpClientFactory.java
@@ -25,6 +25,7 @@ import org.apache.http.conn.ssl.TrustStrategy;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.ssl.SSLContexts;
+import org.apache.kyuubi.client.exception.KyuubiRestException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -45,7 +46,7 @@ public class HttpClientFactory {
       sslSocketFactory = new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE);
     } catch (Exception e) {
       LOG.error("Error: ", e);
-      throw new RuntimeException(e);
+      throw new KyuubiRestException("Failed to create HttpClient", e);
     }
 
     return HttpClientBuilder.create()
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/IRestClient.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/IRestClient.java
@@ -21,15 +21,15 @@ import java.util.Map;
 
 /** A underlying http client interface for common rest request. */
 public interface IRestClient extends AutoCloseable {
-  public <T> T get(String path, Map<String, Object> params, Class<T> type, String authHeader);
+  <T> T get(String path, Map<String, Object> params, Class<T> type, String authHeader);
 
-  public String get(String path, Map<String, Object> params, String authHeader);
+  String get(String path, Map<String, Object> params, String authHeader);
 
-  public <T> T post(String path, String body, Class<T> type, String authHeader);
+  <T> T post(String path, String body, Class<T> type, String authHeader);
 
-  public String post(String path, String body, String authHeader);
+  String post(String path, String body, String authHeader);
 
-  public <T> T delete(String path, Map<String, Object> params, Class<T> type, String authHeader);
+  <T> T delete(String path, Map<String, Object> params, Class<T> type, String authHeader);
 
-  public String delete(String path, Map<String, Object> params, String authHeader);
+  String delete(String path, Map<String, Object> params, String authHeader);
 }
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/RestClient.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/RestClient.java
@@ -17,7 +17,6 @@
 
 package org.apache.kyuubi.client;
 
-import java.io.IOException;
 import java.net.ConnectException;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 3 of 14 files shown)


kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/HttpClientFactory.java
```diff
@@ -45,7 +46,7 @@ public static CloseableHttpClient createHttpClient(RestClientConf conf) {
       sslSocketFactory = new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE);
     } catch (Exception e) {
       LOG.error("Error: ", e);
-      throw new RuntimeException(e);
+      throw new KyuubiRestException("Failed to create HttpClient", e);
```
pan3793, line 49: Always throw `KyuubiRestException`

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/IRestClient.java
```diff
@@ -21,15 +21,15 @@
 
 /** A underlying http client interface for common rest request. */
 public interface IRestClient extends AutoCloseable {
-  public <T> T get(String path, Map<String, Object> params, Class<T> type, String authHeader);
+  <T> T get(String path, Map<String, Object> params, Class<T> type, String authHeader);
```
pan3793, line 24: `public` is redundant here hence nothing other than `public` is allowed in `interface`

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/auth/SpnegoAuthHeaderGenerator.java
```diff
@@ -55,45 +57,43 @@ public String generateAuthHeader() {
...
+    Subject subject;
+    try {
+      Class<?> ugiClz = Class.forName(UGI_CLASS);
+      Method ugiGetCurrentUserMethod = ugiClz.getDeclaredMethod("getCurrentUser");
+      Object ugiCurrentUser = ugiGetCurrentUserMethod.invoke(null);
+      LOG.debug("The user credential is {}", ugiCurrentUser);
+      Field ugiSubjectField = ugiCurrentUser.getClass().getDeclaredField("subject");
+      subject = (Subject) ugiSubjectField.get(ugiCurrentUser);
+    } catch (ClassNotFoundException e) {
+      // TODO do kerberos authentication using JDK class directly
```
pan3793, line 72: We actually can do Kerberos login using JDK classes directly, as long as the user provides principle and keytab.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/exception/RetryableKyuubiRestException.java
```diff
@@ -21,8 +21,8 @@
  * A retryable exception which is thrown by underlying rest client. The call side can do retry by
  * catching this exception.
  */
-public class KyuubiRetryableException extends KyuubiRestException {
-  public KyuubiRetryableException(String message, Throwable cause) {
+public class RetryableKyuubiRestException extends KyuubiRestException {
```
pan3793, line 24: A simple rename

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/util/JsonUtils.java
```diff
@@ -33,7 +33,7 @@ public static String toJson(Object object) {
     }
   }
 
-  public static <T> T toObject(String json, Class<T> clazz) {
+  public static <T> T fromJson(String json, Class<T> clazz) {
```
pan3793, line 36: A simple rename, and this way can not cover generic classes, e.g. `List<String>` because of the restrictions on Java generics system. But it's OK for now.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/RetryableRestClient.java
```diff
@@ -57,7 +57,7 @@ public static IRestClient getRestClient(List<String> uris, RestClientConf conf)
             client);
   }
 
-  private void newRestClient() {
+  private synchronized void newRestClient() {
```
pan3793, line 60: the client may be used in multi threads

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/RestClient.java
```diff
@@ -99,8 +97,7 @@ public String delete(String path, Map<String, Object> params, String authHeader)
   }
 
   private String doRequest(URI uri, String authHeader, RequestBuilder requestBuilder) {
-    String response = "";
-    CloseableHttpResponse httpResponse = null;
```
pan3793, line 103: this class is not used at all

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/auth/SpnegoAuthHeaderGenerator.java
pan3793: Some further thoughts: we can add `principle` and `keytabPath` in RestClientConf, and use them to do Kerberos authentication if the configuration is provided by the user, then fall back to using Hadoop UGI.
And we can rework the Kerberos auth in kyuubi jdbc driver in the same way.
cc @zhouyifan279

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/RetryableRestClient.java
ulysses-you: it will not be used in multi threads.. we can improve the class docs, each thread should hold its own instance of RetryableRestClient

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/auth/SpnegoAuthHeaderGenerator.java
ulysses-you: hmm it's a big change, can we make a new pr for this ?

pan3793: It's a plan, the current change is a simple refactor.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/RetryableRestClient.java
pan3793: In practice, an HTTP-based client should be thread-safe.

ulysses-you (approved):

pan3793:
Thanks, merging to master

PR #2978, opened 2022-06-30
[SUB-TASK][KPIP-4] If batch app status not found from cluster manager, fall back to metadata store

turboFei:

Because the kyuubi batch is not closed once the session completed(reserve the batch log for sometime).

If a batch has been finished for some time, it might can not be found from RM, which has max reserved application list limitation, for this case, we need fall back to metadata for batch info response.

Before:
image


turboFei:
thanks, merged to master

PR #2982, opened 2022-07-01
[KYUUBI #2977] [BATCH] Using KyuubiApplicationManger#tagApplication help tag batch application

### LINKED ISSUE #2977: [Improvement] SparkBatchProcessBuilder build commands should use KyuubiApplicationManger#tagApplication
### Code of Conduct

### Search before asking

1) SparkBatchProcessBuilder tag spark batch job with duplicate code & logic.
We can use `KyuubiApplicationManager#tagApplication` instead.

### How should we improve?

_No response_

zwangsheng:
close #2977 

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
@@ -17,7 +17,6 @@
 
 package org.apache.kyuubi.engine.spark
 
-import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.kyuubi.config.KyuubiConf
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
@@ -49,7 +49,7 @@ class SparkBatchProcessBuilder(
     // tag batch application
     KyuubiApplicationManager.tagApplication(batchId, "spark", clusterManager(), batchKyuubiConf)
 
-    batchKyuubiConf.getAll.foreach { case (k, v) =>
+    (batchKyuubiConf.getAll ++ sparkAppNameConf()).foreach { case (k, v) =>
       buffer += CONF
       buffer += s"$k=$v"
     }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
@@ -46,7 +46,7 @@ class SparkBatchProcessBuilder(
     }
 
     val batchKyuubiConf = new KyuubiConf(false)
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
```diff
@@ -44,16 +45,14 @@ class SparkBatchProcessBuilder(
       buffer += cla
     }
 
-    // tag for YARN
-    val batchJobTag = batchConf.get(TAG_KEY).map(_ + ",").getOrElse("") + batchId
-    var allConf = batchConf ++ Map(TAG_KEY -> batchJobTag) ++ sparkAppNameConf()
+    // session conf + batch conf
+    // user set batch conf should override session conf
+    batchConf.asJava.entrySet().forEach(entry => conf.set(entry.getKey, entry.getValue))
```
turboFei, line 50: we should not use the sessionConf(KyuubiConf), it might contain the spark parameters set for interactive session.

turboFei: and it will also introduce many kyuubi parameters for the batch job

zwangsheng: I see, so it means we have to modify func `KyuubiApplicationManager#tagApplication` for supporting Map type. Or we can add new func to support Map params.

turboFei: how about:
```
    val tagConf = new KyuubiConf(false)
    KyuubiApplicationManager.tagApplication(batchId, "spark", clusterManager(), tagConf)
    val allConf = batchConf ++ tagConf.getAll
    allConf.foreach { case (k, v) =>
      buffer += CONF
      buffer += s"$k=$v"
    }

```

zwangsheng: This also seems to introduce Kyuubi configurations. 
I have made a new change, please check whether this is reasonable.

turboFei: seems the settings of `new KyuubiConf(false)` is empty

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
```diff
@@ -44,16 +44,12 @@ class SparkBatchProcessBuilder(
       buffer += cla
     }
 
-    // tag for YARN
-    val batchJobTag = batchConf.get(TAG_KEY).map(_ + ",").getOrElse("") + batchId
-    var allConf = batchConf ++ Map(TAG_KEY -> batchJobTag) ++ sparkAppNameConf()
```
turboFei, line 49: the `sparkAppNameConf()` is missed

zwangsheng: Done

turboFei (approved):
LGTM

turboFei:
thanks, merged to master

PR #2991, opened 2022-07-02
[KYUUBI #2801] Add local dir allow list and check the application access path URI

### LINKED ISSUE #2801: [FEATURE] Customer behavior audit - forbid to upload local files
### Code of Conduct

### Search before asking

### Describe the feature

We should forbid the customer to upload local files and throw exception if there is any extra local file involved in the engine.

Such as `spark.files=/etc/security/keytabs/kyuubi.keytab`.

### Motivation

It is not security for kyuubi server.

### Describe the solution

_No response_

### Additional context

_No response_

- [x] Yes I am willing to submit a PR!

turboFei:

To close #2801 

Customer might set some parameters to upload the local files, such as spark.files=/etc/security/keytabs/kyuubi.keytab. It is not security for kyuubi server.

In this pr, I introduce a local dir allow list and customer can only upload the files under these dirs.


pan3793:
Prefer to use allow/deny instead of white/black

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -168,6 +168,13 @@ object SparkProcessBuilder {
   final val APP_KEY = "spark.app.name"
   final val TAG_KEY = "spark.yarn.tags"
   final val MASTER_KEY = "spark.master"
+  final val PATH_CONFIGS = Seq(
+    SPARK_FILES,
+    "spark.jars",
+    "spark.archives",
+    "spark.yarn.dist.files",
+    "spark.yarn.dist.jars",
+    "spark.yarn.dist.archives")
```
ulysses-you, line 177: cc @zwangsheng for k8s code path

turboFei: maybe there is no such parameter for k8s？

these config items are from SparkSubmit.

zwangsheng: Well, I'm not sure if the parameters (etc. `spark.kubernetes.kerberos.krb5.path`) for uploading functional files are in your range.

turboFei: seems that spark.kubernetes.kerberos.krb5.path is not uploaded during spark-submit?

It should be including in the image?

turboFei: image

The config item I use is from SparkSubmit

zwangsheng: This conf `spark.kubernetes.kerberos.krb5.path` used in `KerberosConfDriverFeatureStep` which build driver pod.
This file is uploaded when the Driver Pod Resource is built after Submit and becomes a mount to DriverPod

turboFei: I see, I did not enable `kubenertes` profile in my spark project, let me have a check.

ulysses-you: BTW, we can add some comments of where the spark code reference it to help recall in future

turboFei:
```
spark.kerberos.keytab
spark.yarn.keytab
spark.kubernetes.kerberos.krb5.path
```

turboFei: ok, have added above config item and add some comments

zwangsheng: spark.kubernetes.file.upload.path

zwangsheng: What is our attitude on this `spark.kubernetes.file.upload.path `, in Kubernetes Cluster mode, that this configuration is important for uploading jars.

zwangsheng: We also need to check  kubernetes authenticate etc. `spark.kubernetes.authenticate.driver.caCertFile`.

turboFei: Ok, we can also add it into path configs.

turboFei: Thanks @zwangsheng , is there any other parameter for k8s use case?

zwangsheng: This is the relevant configuration I can find, based on the situations I have used.

turboFei: ok, we can complete it if we meet another case in the future.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -106,6 +108,22 @@ object KyuubiApplicationManager {
     conf.set(FlinkProcessBuilder.TAG_KEY, newTag)
   }
 
+  private def checkSparkPathURIs(
+      appConf: Map[String, String],
+      sessionManager: KyuubiSessionManager): Unit = {
```
ulysses-you, line 113: it seems we can pass kyuubi conf and move `sessionManager.checkSessionAccessPathURI` into `KyuubiApplicationManager` ?

turboFei: the `SESSION_LOCAL_DIR_ALLOW_LIST` is server side only conf, so it can not passed to sessionConf.

turboFei: let me try.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -96,6 +103,8 @@ class KyuubiSessionImpl(
     // we should call super.open before running launch engine operation
     super.open()
```
ulysses-you, line 104: nit: we can move `checkSessionAccessPathURIs`  before the `super.open()` to avoid creating operation log if conf is illegal.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -126,6 +134,8 @@ class KyuubiBatchSessionImpl(
     // we should call super.open before running batch job submission operation
     super.open()
 
+    checkSessionAccessPathURIs()
```
ulysses-you, line 137: ditto

ulysses-you (approved):

turboFei:
thanks @ulysses-you @zwangsheng @pan3793 
merged to master

PR #3004, opened 2022-07-04
Clean up JDBC shaded client pom and license

pan3793:

Clean up JDBC shaded client pom and license because we removed some dependencies.

This PR also disables `minimizeJar` in JDBC shaded client, because it may cause unexpected "class not found" issues.

The current JDBC shaded client jar size is 11MB.


cxzl25 (approved):

pan3793:
Thanks, merging to master

PR #3023, opened 2022-07-06
Kyuubi Hive JDBC: Replace UGI-based Kerberos authentication w/ JAAS

pan3793:

The current JDBC Kerberos authentication and Hadoop UGI classes are strongly coupled, it's not friendly for downstream projects which do not have Hadoop dependencies, and users who are not familiar w/ Hadoop security mechanism.
 
This PR proposes to replace the UGI-based Kerberos authentication with JAAS-based.

The main logic is

```
if (param.auth == 'noSasl') {
    // no-sasl
    raw socket/http requrest
} else if (!param.principal) {
    // sasl/plain
    auth by param.username, param.password
} else if (param.principal && param.kerberosAuthType == 'fromSubject') {
    [0]
} else if (param.principal && param.kyuubiClientPrincipal && param.kyuubiClientKeytab) {
    [1]
} else if (param.principal && !param.kyuubiClientPrincipal && !param.kyuubiClientKeytab) {
    [3]
}

- [0][JDK] subject from context
- [1][JDK] principle + keytab
- [2][UGI] principle + keytab (no required, use [1] instead)
- [3][JDK] principle + tgt cache
- [4][UGI] principle + tgt cache (no required, use [3] instead)
```

To achieve this, we need to introduce new JDBC parameters `kyuubiClientPrincipal` and `kyuubiClientKeytab`.

The above description is pure JDK-based and totally removed Hadoop dependencies from Kyuubi Hive JDBC driver.

To minimize the scope of this PR, I separate the support of delegation token to a follow-up PR #3096, and it still needs UGI.


pan3793:
cc @yaooqinn @zhouyifan279 @cxzl25 @turboFei

pan3793:
@yaooqinn for review convince, this PR has been split into multiple parts, I updated the PR description to reflect the changes, please take a look when you have time.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -83,7 +82,6 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
   public static final Logger LOG = LoggerFactory.getLogger(KyuubiConnection.class.getName());
   public static final String BEELINE_MODE_PROPERTY = "BEELINE_MODE";
   public static final String HS2_PROXY_USER = "hive.server2.proxy.user";
-  public static final String HS2_CLIENT_TOKEN = "hiveserver2ClientToken";
```
cxzl25, line 86: n this PR we do not support the `delegationToken` authentication method, we will add it back in the subsequent PR #3096, right?

pan3793: yea

pan3793:
Since #3096 got approved, I'm going to merge this PR, thanks all for helping reviews.

turboFei:
hi @pan3793 

see that ugi.doAs does not work after this change if KRB5CCNAME env is set and does not set `kerberosAuthType=fromSubject` explicitly.

turboFei:
cc @gabrywu

pan3793:
@turboFei @gabrywu thanks for reporting this issue, raised https://github.com/apache/kyuubi/pull/4479 to fix it, please take a look

PR #3047, opened 2022-07-12
Fallback krb5 conf to OS if not configured

pan3793:

We should not assume that the default `krb5.conf` is located at `/etc/krb5.conf`, it OS-depended.


yaooqinn (approved):

pan3793:
Thanks, merging to master

PR #3048, opened 2022-07-12
[KYUUBI #3051] Support to get the  real client ip address for thrift connection when using VIP as kyuubi server load balancer

### LINKED ISSUE #3051: [FEATURE] Support to get real ip for thrift connection when using VIP as kyuubi server LB
### Code of Conduct

### Search before asking

### Describe the feature

We deploy kyuubi server on k8s and using k8s lb and the connection path is:

client -> lb -> kyuubi server.

Now the client ip address is always the lb ip and is meaning less.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

turboFei:

Close #3051 

We deploy kyuubi server on k8s and using k8s lb and the connection path is:

client -> lb -> kyuubi server.

Now the client ip address is always the lb ip and is meaning less.

### DIFF
--- a/externals/kyuubi-hive-sql-engine/src/main/scala/org/apache/kyuubi/engine/hive/session/HiveSessionManager.scala
+++ b/externals/kyuubi-hive-sql-engine/src/main/scala/org/apache/kyuubi/engine/hive/session/HiveSessionManager.scala
@@ -28,7 +28,6 @@ import org.apache.hive.service.cli.session.{HiveSessionImplwithUGI => ImportedHi
 import org.apache.hive.service.rpc.thrift.TProtocolVersion
 
 import org.apache.kyuubi.config.KyuubiConf.ENGINE_SHARE_LEVEL
-import org.apache.kyuubi.config.KyuubiReservedKeys.KYUUBI_CLIENT_IP_KEY
 import org.apache.kyuubi.engine.ShareLevel
 import org.apache.kyuubi.engine.hive.HiveSQLEngine
 import org.apache.kyuubi.engine.hive.operation.HiveOperationManager
@@ -74,7 +73,6 @@ class HiveSessionManager(engine: HiveSQLEngine) extends SessionManager("HiveSess
       ipAddress: String,
       conf: Map[String, String]): Session = {
     val sessionHandle = SessionHandle()
-    val clientIp = conf.getOrElse(KYUUBI_CLIENT_IP_KEY, ipAddress)
     val hive = {
       val sessionWithUGI = new ImportedHiveSessionImpl(
         new ImportedSessionHandle(sessionHandle.toTSessionHandle, protocol),
@@ -84,7 +82,7 @@ class HiveSessionManager(engine: HiveSQLEngine) extends SessionManager("HiveSess
         HiveSQLEngine.hiveConf,
         ipAddress,
         null,
-        Seq(clientIp).asJava)
+        Seq(ipAddress).asJava)
       val proxy = HiveSessionProxy.getProxy(sessionWithUGI, sessionWithUGI.getSessionUgi)
       sessionWithUGI.setProxySession(proxy)
       proxy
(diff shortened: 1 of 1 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
```diff
@@ -19,6 +19,7 @@ package org.apache.kyuubi.config
 
 object KyuubiReservedKeys {
   final val KYUUBI_SESSION_USER_KEY = "kyuubi.session.user"
+  final val KYUUBI_SESSION_IP_KEY = "kyuubi.session.ip"
```
yaooqinn, line 22: we have `kyuubi.client.ip` already, how about using this?

maybe  kyuubi.client.ipAddress is better

turboFei: thanks for the information

kyuubi-common/src/main/scala/org/apache/kyuubi/session/package.scala
```diff
@@ -28,5 +28,5 @@ package object session {
 
   val SPARK_PREFIX = "spark."
 
-  val CLIENT_IP_KEY = "kyuubi.client.ip"
```
yaooqinn, line 31: move it to reserved keys class

yaooqinn:
for authentication, this introduce a CVE

turboFei:
> for authentication, this introduces a CVE

Seems difficult to avoid it....

turboFei:
In fact, now with the proxy lb, the client ip address is also not the real ip and maybe not in the ip white list.

yaooqinn:
> In fact, now with the proxy lb, the client ip address is also not the real ip and maybe not in the ip white list.

that’s lb's duty..

turboFei:
make sense

yaooqinn:
if we just want to track the IP, not for authentication, we may just leave it in the configuration and get it whenever we want

turboFei:
> if we just want to track the IP, not for authentication, we may just leave it in the configuration and get it whenever we want

How about that,
we define another reserved key for this?

Otherwise, the kyuubi.client.ip will be overwrite.

turboFei:
It will be overwritten here:

https://github.com/apache/incubator-kyuubi/blob/99934591f07be5bc8abc1955848e4fc291b8569b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala#L60-L65

yaooqinn (approved):

turboFei:
thank a lot @yaooqinn

externals/kyuubi-hive-sql-engine/src/main/scala/org/apache/kyuubi/engine/hive/session/HiveSessionImpl.scala
```diff
@@ -33,7 +33,6 @@ class HiveSessionImpl(
     protocol: TProtocolVersion,
     user: String,
     password: String,
-    serverIpAddress: String,
```
pan3793, line 36: this change may breaks https://github.com/apache/incubator-kyuubi/pull/1530/files#diff-759a09fe90c07b0ef78f6f4dc28bac9e3d0785d4b922cbe33422524e8a2520d4R51

pan3793: cc @turboFei @wForget

wForget: [SparkSessionImpl.scala#L51](https://github.com/turboFei/incubator-kyuubi/blob/97f9b00417f211cc547998821ba421b15edb6c76/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSessionImpl.scala#L51) becomes redundant, thanks, I will remove it.

```
  def serverIpAddress(): String = serverIpAddress
```

PR #3053, opened 2022-07-13
[KYUUBI #3052] Support to get the real client ip address for http connection when using VIP as kyuubi server load balancer

### LINKED ISSUE #3052: [FEATURE] Support to get real ip for http connection when using VIP as kyuubi server LB #3051
### Code of Conduct

### Search before asking

### Describe the feature

We deploy kyuubi server on k8s and using k8s lb and the connection path is:

client -> lb -> kyuubi server.

Now the client ip address is always the lb ip and is meaning less.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

turboFei:

To close #3052 

If your server is behind a load balancer or other proxy, the server will see this  load balancer or proxy IP address as the client IP address, to get around this common issue, most load balancers or proxies offer the ability to record the real remote IP address in an HTTP herader that will be added to the request for other devices to use.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -580,6 +580,17 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val FRONTEND_PROXY_HTTP_CLIENT_IP_HEADER: ConfigEntry[String] =
+    buildConf("kyuubi.frontend.proxy.http.client.ip.header")
+      .doc("The http header to record the real client ip address. If your server is behind a load" +
+        " balancer or other proxy, the server will see this load balancer or proxy IP address as" +
+        " the client IP address, to get around this common issue, most load balancers or proxies" +
+        " offer the ability to record the real remote IP address in an HTTP herader that will be" +
+        " added to the request for other devices to use.")
+      .version("1.6.0")
+      .stringConf
+      .createWithDefault("X-Real-IP")
+
   val AUTHENTICATION_METHOD: ConfigEntry[Seq[String]] = buildConf("kyuubi.authentication")
     .doc("A comma separated list of client authentication types.<ul>" +
       " <li>NOSASL: raw transport.</li>" +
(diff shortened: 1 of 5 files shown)


pan3793 (approved):
The change LGTM, please attach a manual test result in the PR description.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -580,6 +580,17 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val FRONTEND_PROXY_HTTP_CLIENT_IP_HEADER: ConfigEntry[String] =
```
yaooqinn, line 583: why we need this conf?

yaooqinn:
this is odd, shall it be `forwardedAddresses/X_FORWARDED_FOR_HEADER`

yaooqinn (changes requested):
this is a risk and nonstandard change for me

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: It is a config for the proxy.

I am using ebay hlb, the header is `X-eBay-Web-Tier-IP`.

turboFei:
I tried to 

> this is odd, shall it be `forwardedAddresses/X_FORWARDED_FOR_HEADER`
> 
> > X-Forwarded-For
> > The X-Forwarded-For (XFF) request header is a de-facto standard header for identifying the originating IP address of a client connecting to a web server through a proxy server.

In fact, it depends on the proxy setting.

For the ebay HLB, there is no ` X-Forwarded-For` header.

And the header to present the real client ip is `X-eBay-Web-Tier-IP` instead of `X-Real-IP`.

turboFei:
FYI:

https://docs.fastly.com/signalsciences/faq/real-client-ip-addresses/#setting-alternative-headers-directly-in-the-agent


Because the LB I am using does not have `X-Forwarded-For`, so I can not make manual testing for that.

How about do it in another pr.

yaooqinn:
how about making `X-Forwarded-For/X-Real-IP` in one and delivering with forwardedAddresses

we need the IP of the direct accessor if it is not only proxying address but also proxying user

yaooqinn (approved):

yaooqinn:
on second thought, it makes sense. please add a ut

turboFei:
ok, I will add UT for rest connection.

turboFei:
thanks @yaooqinn @pan3793 
merged to master

PR #3055, opened 2022-07-13
Expose client ip address into batch request conf

turboFei:

Append the client ip address into batch conf, so that we can know that where the batch request come from on spark UI.


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -114,6 +114,7 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       batchRequest: BatchRequest,
       recoveryMetadata: Option[Metadata] = None): KyuubiBatchSessionImpl = {
     val username = Option(user).filter(_.nonEmpty).getOrElse("anonymous")
+    batchRequest.getConf.put(CLIENT_IP_KEY, ipAddress)
```
ulysses-you, line 117: where we use it ?

turboFei: we can check the client ip in spark-ui.

ulysses-you (approved):

turboFei:
thanks, merged to master

PR #3058, opened 2022-07-13
decouple jackson from project

zhaomin1423:

Decouple jackson from project, using it by json. If need, we can replace it conveniently in the future.


pan3793:
It's not much necessary I think.

zhaomin1423:
> It's not much necessary I think.

ok, closed it.

PR #3065, opened 2022-07-14
Support to retry the killApplicationByTag for JpsApplicationOperation

turboFei:

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/JpsApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/JpsApplicationOperation.scala
@@ -58,7 +58,7 @@ class JpsApplicationOperation extends ApplicationOperation {
     }
   }
 
-  override def killApplicationByTag(tag: String): KillResponse = {
+  private def killJpsApplicationByTag(tag: String, retryable: Boolean): KillResponse = {
     val commandOption = getEngine(tag)
     if (commandOption.nonEmpty) {
       val idAndCmd = commandOption.get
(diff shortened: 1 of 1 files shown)


turboFei:
cc @ulysses-you @yaooqinn

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/JpsApplicationOperation.scala
```diff
@@ -68,13 +68,20 @@ class JpsApplicationOperation extends ApplicationOperation {
         (true, s"Succeeded to terminate: $idAndCmd")
       } catch {
         case e: Exception =>
+          if (retryable && getEngine(tag).nonEmpty) {
```
ulysses-you, line 71: we can leave some comments why we retry

ulysses-you (dismissed):

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/JpsApplicationOperation.scala
```diff
@@ -68,13 +68,20 @@ class JpsApplicationOperation extends ApplicationOperation {
...
+            killJpsApplicationByTag(tag, false)
```
yaooqinn, line 72: ..., else

yaooqinn (changes requested):

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/JpsApplicationOperation.scala
ulysses-you: shit.. we should return the last attempt result

yaooqinn: **PLEASE do follow the Code of Conduct of the Kyuubi community.*

turboFei: thanks, it is a little rush recently. will follow it.

yaooqinn (approved):

turboFei:
thanks @yaooqinn @ulysses-you merging to master

PR #3078, opened 2022-07-15
[KYUUBI #3052][FOLLOWUP] Do not use the ip in proxy http header for authentication to prevent CVE

turboFei:

Because the client can specify any ip in http header, to prevent CVE issue, we do not use it for authentication.


yaooqinn (approved):

turboFei:
thanks, merged to master

PR #3087, opened 2022-07-15
Convert the kyuubi batch conf with `spark.` prefix so that spark could identify

turboFei:

We might need transfer some parameters like `kyuubi.client.ipAddress`, but it can not be identified by spark.

So we need convert them.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
@@ -50,16 +50,8 @@ class SparkBatchProcessBuilder(
     KyuubiApplicationManager.tagApplication(batchId, "spark", clusterManager(), batchKyuubiConf)
 
     (batchKyuubiConf.getAll ++ sparkAppNameConf()).foreach { case (k, v) =>
-      val newKey = {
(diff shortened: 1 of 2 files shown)


turboFei:
cc @ulysses-you 
Sorry for the mistake in #3055, it can be identified by spark.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
```diff
@@ -50,8 +50,16 @@ class SparkBatchProcessBuilder(
     KyuubiApplicationManager.tagApplication(batchId, "spark", clusterManager(), batchKyuubiConf)
 
     (batchKyuubiConf.getAll ++ sparkAppNameConf()).foreach { case (k, v) =>
+      val newKey = {
+        // convert the kyuubi config with `spark` prefix, so that spark could identify
+        if (k.startsWith("kyuubi.")) {
+          "spark." + k
```
ulysses-you, line 56: shall we reuse the code in `SparkProcessBuilder` ? so we can keep the same behavior on spark conf and hadoop conf.

turboFei: sorry for the late reply, updated.

ulysses-you (approved):

turboFei:
thanks, merged to master

PR #3096, opened 2022-07-18
Kyuubi Hive JDBC: UGI-based delegation token authentication

pan3793:

This is a follow-up to #3023.

```
if (param.auth = 'delegationToken') {
  // delegation token
  ensure hadoopClassAvaiable, but maybe we can do it w/o hadoop classes in the future
  UGI-based delegation token authentication
}
```

To avoid depending on Hadoop classes in compile phase, this PR introduces a `HadoopShim` to use reflection to invoke Hadoop classes for delegation token authentication.

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/auth/HttpAuthUtils.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/auth/HttpAuthUtils.java
@@ -32,8 +32,8 @@ public final class HttpAuthUtils {
 
   /** @return Stringified Base64 encoded kerberosAuthHeader on success */
   public static String getKerberosServiceTicket(
-      String serverPrinciple, String host, Subject loggedInSubject) throws Exception {
-    String spn = KerberosUtils.canonicalPrincipal(serverPrinciple, host);
+      String serverPrincipal, String host, Subject loggedInSubject) throws Exception {
+    String spn = KerberosUtils.canonicalPrincipal(serverPrincipal, host);
     return Subject.doAs(loggedInSubject, new HttpKerberosClientAction(spn));
   }
 
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -82,6 +82,7 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
   public static final Logger LOG = LoggerFactory.getLogger(KyuubiConnection.class.getName());
   public static final String BEELINE_MODE_PROPERTY = "BEELINE_MODE";
   public static final String HS2_PROXY_USER = "hive.server2.proxy.user";
+  public static final String HS2_CLIENT_TOKEN = "hiveserver2ClientToken";
   public static int DEFAULT_ENGINE_LOG_THREAD_TIMEOUT = 10 * 1000;
 
   private String jdbcUriString;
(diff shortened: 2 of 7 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/auth/HttpAuthUtils.java
```diff
@@ -31,21 +30,11 @@ public final class HttpAuthUtils {
...
-    String serverPrincipal = HadoopThriftAuthBridge.getBridge().getServerPrincipal(principal, host);
-    if (loggedInSubject != null) {
-      return Subject.doAs(loggedInSubject, new HttpKerberosClientAction(serverPrincipal));
-    } else {
-      // JAAS login from ticket cache to setup the client UserGroupInformation
-      UserGroupInformation clientUGI =
-          HadoopThriftAuthBridge.getBridge().getCurrentUGIWithConf("kerberos");
-      return clientUGI.doAs(new HttpKerberosClientAction(serverPrincipal));
-    }
+      String serverPrinciple, String host, Subject loggedInSubject) throws Exception {
```
cxzl25, line 35: serverPrinciple -> serverPrincipal

kyuubi-hive-jdbc-shaded/pom.xml
```diff
@@ -44,12 +44,6 @@
             <artifactId>jsr305</artifactId>
             <optional>true</optional>
         </dependency>
-
-        <!-- TODO: mark as optional once provide JAAS-based Kerberos authentication   -->
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-client-api</artifactId>
```
cxzl25, line 51: If the user needs to use the `delegationToken` authentication method, the user needs to manually add `hadoop-client-api` to the class path, right?

pan3793: yes, either `hadoop-comon` or `hadoop-client-api` is fine

pan3793:
Kindly ping @yaooqinn

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationKerberosAndPlainAuthSuite.scala
```diff
@@ -65,8 +68,29 @@ class KyuubiOperationKerberosAndPlainAuthSuite extends WithKyuubiServer with Ker
...
+    val conn = DriverManager.getConnection(jdbcUrlWithConf(kerberosTgtJdbcUrl), user, "")
+    try {
+      val statement = conn.createStatement()
+      val resultSet = statement.executeQuery("select engine_name()")
+      assert(resultSet.next())
+      assert(resultSet.getString(1).nonEmpty)
+    } finally {
+      conn.close()
+    }
+  }
```
zhouyifan279, line 89: Better to call `kdestroy` to avoid potential effects to other tests

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/auth/CachingKerberosAuthentication.java
```diff
@@ -0,0 +1,63 @@
...
+      subject =
+          requireNonNull(
+              kerberosAuthentication.getSubject(), "kerberosAuthentication.getSubject() is null");
+      KerberosTicket tgtTicket = KerberosUtils.getTgt(subject);
+      nextRefreshTime = KerberosUtils.getTgtRefreshTime(tgtTicket);
+    }
+    return subject;
+  }
+
+  public synchronized void reauthenticateIfSoonWillBeExpired() {
```
zhouyifan279, line 51: This method is not used. Will we spawn a thread like `UserGroupInformation#spawnAutoRenewalThreadForUserCreds`  to refresh TgtCache ?

zhouyifan279 (approved):

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationKerberosAndPlainAuthSuite.scala
pan3793: Thanks, added

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/auth/CachingKerberosAuthentication.java
pan3793: It's a good point, since `CachingKerberosAuthentication#getSubject` is called on creating each connection, then in the current implementation, it would be a problem if the user creates a connection and keeps it for a long time, especially when the user uses the connection pool.
Considering this PR is large enough, I prefer to implement the `spawnAutoRenewalThreadForUserCreds` in the following PR.

PR #3106, opened 2022-07-20
Correct `RelMetadataProvider` used in flink-sql-engine

liaoyt:

Currently flink-sql-engine can not run sql statement like followed, which is a normal case in flink sql.
```sql
select count(*) from tbl_src;
```
In Flink the default RelMetadataProvider is `org.apache.flink.table.planner.plan.metadata.FlinkDefaultRelMetadataProvider`, and in Kyuubi we use `org.apache.calcite.rel.metadata.DefaultRelMetadataProvider`.
We should change the RelMetadataProvider to the flink one to support all flink sql syntax.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -24,13 +24,14 @@ import java.util.concurrent.RejectedExecutionException
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 
-import org.apache.calcite.rel.metadata.{DefaultRelMetadataProvider, JaninoRelMetadataProvider, RelMetadataQueryBase}
+import org.apache.calcite.rel.metadata.{JaninoRelMetadataProvider, RelMetadataQueryBase}
 import org.apache.flink.table.api.ResultKind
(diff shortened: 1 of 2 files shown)


SteNicholas (approved):
@liaoyt, thanks for the contribution to fix the `RelMetadataProvider`. LGTM. I wll merge after the CI passed.

zhaomin1423:
You can run dev/reformat to format all Java and Scala code.

liaoyt:
> You can run dev/reformat to format all Java and Scala code. There are some information that can help you. https://github.com/apache/incubator-kyuubi/blob/f1312ea439d288009200a105db36aa8431f27f8a/docs/develop_tools/developer.md

Thanks for the help, I will try it later.
I found that some unit test case failed with flink version 1.15 due to the difference between flink 1.14 and 1.15.
I'm thinking about whether I should fix it in this PR or open another issue and fix it in the future.
Any suggestions? @SteNicholas @zhaomin1423

pan3793:
> I found that some unit test case failed with flink version 1.15 due to the difference between flink 1.14 and 1.15.

This is because Kyuubi engine uses vanilla `org.apache.calcite` classes, it's not a correct way, we need to use reflect to access the Flink runtime shaded calcite classes, and fallback to vanilla one(only for ut.)

> I'm thinking about whether I should fix it in this PR or open another issue and fix it in the future.

Either is fine. Choose the one you like.

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #3111, opened 2022-07-21
Replace HashMap with singletonMap

gabrywu:

replace HashMap with singletonMap, since it only has one element


yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #3139, opened 2022-07-26
Override toString method for rest dto classes

turboFei:

It is helpful if we want to log some info when using rest sdk.

Pass existing UT.

image

### DIFF
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/Batch.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/Batch.java
@@ -18,6 +18,8 @@
 package org.apache.kyuubi.client.api.v1.dto;
 
 import java.util.Objects;
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
 
 public class Batch {
   private String id;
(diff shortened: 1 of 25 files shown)


kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/Batch.java
```diff
@@ -170,4 +170,44 @@ public boolean equals(Object o) {
...
+        + kyuubiInstance
+        + '\''
+        + ", state='"
+        + state
+        + '\''
+        + ", createTime="
+        + createTime
+        + ", endTime="
+        + endTime
+        + '}';
```
ulysses-you, line 211: is it generated by idea ? why looks so complex..

turboFei: yes, all the toString methods are generated by IDEA.

And then execute `dev/reformat` to format the code style.

ulysses-you: magic style..

ulysses-you (approved):

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/Batch.java
turboFei: @pan3793  suggested to use ReflectionToStringBuilder, let me have a try.

turboFei: image

pan3793 (approved):

turboFei:
thanks @ulysses-you @pan3793 , merged to master

PR #3142, opened 2022-07-26
[KYUUBI #3141] Trino engine etcd support

hddong:

Resolved conflict jar for trino engine used etcd

### DIFF
--- a/externals/kyuubi-trino-engine/pom.xml
+++ b/externals/kyuubi-trino-engine/pom.xml
@@ -64,11 +64,6 @@
         </dependency>
 
         <dependency>
-            <groupId>net.jodah</groupId>
-            <artifactId>failsafe</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.kyuubi</groupId>
             <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
             <version>${project.version}</version>
--- a/externals/kyuubi-trino-engine/pom.xml
+++ b/externals/kyuubi-trino-engine/pom.xml
@@ -64,6 +64,11 @@
         </dependency>
 
         <dependency>
+            <groupId>net.jodah</groupId>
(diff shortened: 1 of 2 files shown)


hddong:
cc @yaooqinn @pan3793

externals/kyuubi-trino-engine/src/test/scala/org/apache/kyuubi/engine/trino/ShareLevelTrinoEngineTests.scala
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.kyuubi.engine.ShareLevel
+import org.apache.kyuubi.engine.ShareLevel.ShareLevel
+import org.apache.kyuubi.operation.HiveJDBCTestHelper
+import org.apache.kyuubi.service.ServiceState
+
+/**
+ * This suite is to test some behavior with Trino engine in different share level.
+ * e.g. cleanup discovery service before stop.
+ */
+trait ShareLevelTrinoEngineTests
```
yaooqinn, line 31: we don't need to copy tests from other modules

yaooqinn:
ha behavior should be the same across engine types, so the tests here are unnecessary to me

externals/kyuubi-trino-engine/src/test/scala/org/apache/kyuubi/engine/trino/ShareLevelTrinoEngineTests.scala
hddong: > we don't need to copy tests from other modules

Get it, removed the tests.

yaooqinn (approved):

hddong:
@pan3793 As discussed, used explicit declaration instead exclusion.
cc @yaooqinn

yaooqinn:
thanks, merged to master

PR #1147, opened 2021-09-24
[KYUUBI #1131] Rename KyuubiDriver to KyuubiHiveDriver

pan3793:
### _Why are the changes needed?_

This PR is based on discussion https://www.mail-archive.com/dev@kyuubi.apache.org/msg00977.html

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.readthedocs.io/en/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #3144, opened 2022-07-26
Remove deprecated KyuubiDriver in services manifest

pan3793:

`KyuubiDriver` has been marked deprecated and replaced by `KyuubiHiveDriver` in 1.4.0, see detail in #1147.

This PR removes the deprecated KyuubiDriver in services manifest so it would not be discovered by the service loader, users still can use this deprecated driver before removing the class itself.


cxzl25 (approved):

pan3793:
Thanks, merging to master

PR #3146, opened 2022-07-26
[KYUUBI #3138] [Subtask] DorisSQLEngine - Add jdbc engine to dist

zhaomin1423:

close https://github.com/apache/incubator-kyuubi/issues/3138#issue-1316764571

### DIFF
--- a/build/dist
+++ b/build/dist
@@ -217,6 +217,7 @@ mkdir -p "$DISTDIR/externals/engines/flink"
 mkdir -p "$DISTDIR/externals/engines/spark"
 mkdir -p "$DISTDIR/externals/engines/trino"
 mkdir -p "$DISTDIR/externals/engines/hive"
+mkdir -p "$DISTDIR/externals/engines/jdbc"
 mkdir -p "$DISTDIR/beeline-jars"
 echo "Kyuubi $VERSION $GITREVSTRING built for" > "$DISTDIR/RELEASE"
 echo "Java $JAVA_VERSION" >> "$DISTDIR/RELEASE"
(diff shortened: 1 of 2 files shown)


build/dist
```diff
@@ -256,6 +257,9 @@ cp -r "$KYUUBI_HOME/externals/kyuubi-trino-engine/target/scala-$SCALA_VERSION/ja
 # Copy hive engines
 cp "$KYUUBI_HOME/externals/kyuubi-hive-sql-engine/target/kyuubi-hive-sql-engine_${SCALA_VERSION}-${VERSION}.jar" "$DISTDIR/externals/engines/hive"
 
+# Copy jdbc engines
+cp "$KYUUBI_HOME/externals/kyuubi-jdbc-engine/target/kyuubi-jdbc-engine_${SCALA_VERSION}-${VERSION}.jar" "$DISTDIR/externals/engines/jdbc"
```
pan3793, line 261: does the JDBC engine use an all-in-one fat jar? do you consider putting all jars into this folder instead of a fat jar?

zhaomin1423: Yes, https://github.com/apache/incubator-kyuubi/pull/3125. I think that putting jars to the folder is ok, but  I want to know your concerns.

pan3793: Shade should always be the last solution because it breaks the gather all of the stuff into one fat-jar, which makes it hard to distinguish where the class comes from when we meet issues, e.g. the relocation may introduce runtime linkage risks w/ reflection, and relocation does not work well w/ native dynamic libraries.

For kyuubi spark engine and flink engine, we choose to use a shaded jar because our classes will be submitted to spark/flink cluster alongside w/ the spark/flink runtime jars and user resources(e.g. UDF), we should make sure that no class confliction w/ them. In other words, the engine plays a "plugin" role in these cases.

For trino and jdbc engine, we just invoke the client to communicate to the server, it has less opportunity that the user will add many jars on the engine side. If they do, for example, add JDBC driver into JDBC engine classpath, then they have the responsibility to align the library version of us or all shade classes to make sure no class confliction occurs.  In other words, the engine plays a "framework" role in these cases.

zhaomin1423: Thanks, make sense. I will update it.

zhaomin1423:
Add all jars to /external/jdbc/jars. @pan3793

externals/kyuubi-jdbc-engine/pom.xml
```diff
@@ -85,62 +79,17 @@
...
-                            <goal>shade</goal>
+                            <goal>test-jar</goal>
                         </goals>
                     </execution>
                 </executions>
             </plugin>
         </plugins>
     </build>
-
 </project>
```
pan3793, line 95: nit: add a newline

zhaomin1423: done

pan3793:
Thanks, merging to master

PR #3152, opened 2022-07-26
Introduce JDBC parameters to control connection timeout

pan3793:

There are two concepts `connect timeout` and `so timeout` in the socket, currently, the Kyuubi Hive JDBC driver uses `DriverManager#loginTimeout` as both of them, it does not make sense, and will cause unexpected "read timeout" exceptions if `DriverManager#loginTimeout` was set a small value, e.g. Spring Boot set it to 30s in default.

This PR proposes to introduce two dedicated JDBC parameters to control them.

See also https://github.com/apache/hive/pull/3379

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -37,7 +37,6 @@ import java.security.SecureRandom;
 import java.sql.*;
 import java.util.*;
 import java.util.Map.Entry;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
@@ -51,6 +50,7 @@ import org.apache.http.HttpResponse;
 import org.apache.http.NoHttpResponseException;
 import org.apache.http.client.CookieStore;
 import org.apache.http.client.ServiceUnavailableRetryStrategy;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.config.Registry;
 import org.apache.http.config.RegistryBuilder;
 import org.apache.http.conn.socket.ConnectionSocketFactory;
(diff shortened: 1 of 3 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -856,13 +857,24 @@ private String getSessionValue(String varName, String varDefault) {
     return varValue;
   }
 
-  // copy loginTimeout from driver manager. Thrift timeout needs to be in millis
-  private void setupLoginTimeout() {
-    long timeOut = TimeUnit.SECONDS.toMillis(DriverManager.getLoginTimeout());
```
cxzl25, line 861: Do we need to consider the scenario where only `loginTimeout` is set, but `connectTimeout` and `socketTimeout` are not set?
It looks like this change is not intended to use JDBC's loginTimeout.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -856,13 +857,24 @@ private String getSessionValue(String varName, String varDefault) {
...
-    if (timeOut > Integer.MAX_VALUE) {
-      loginTimeout = Integer.MAX_VALUE;
-    } else {
-      loginTimeout = (int) timeOut;
+  private void setupTimeout() {
+    if (sessConfMap.containsKey(CONNECT_TIMEOUT)) {
+      String loginTimeoutStr = sessConfMap.get(CONNECT_TIMEOUT);
```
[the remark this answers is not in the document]
pan3793: fixed.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
pan3793: It is based on the discussion https://github.com/apache/hive/pull/3379#discussion_r905980870, do you have different ideas?

cxzl25: After reading the discussion on several PRs, this change looks good to me.
[HIVE-12371](https://issues.apache.org/jira/browse/HIVE-12371)
[HIVE-22196](https://issues.apache.org/jira/browse/HIVE-22196)
[HIVE-26336](https://issues.apache.org/jira/browse/HIVE-26336)

cxzl25 (approved):

pan3793:
Thanks, merging to master

Issue #3180, opened 2022-08-04
[Improvement] Add component version util class


deadwind4:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Sometimes Kyuubi needs to check the component version. For example, check Spark >= 3.2 or Flink >= 1.14 or Kyuubi > 1.16 .
We need a unity class util.

### How should we improve?

Extract the spark version method in AuthZUtils class to a new unity class in the `kyuubi-common` module.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3181, opened 2022-08-04
[KYUUBI #3180] Add component version util

deadwind4:

Add component version util class

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz/pom.xml
@@ -40,6 +40,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.ranger</groupId>
             <artifactId>ranger-plugins-common</artifactId>
             <version>${ranger.version}</version>
--- a/extensions/spark/kyuubi-spark-authz/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz/pom.xml
@@ -40,11 +40,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.apache.kyuubi</groupId>
-            <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
             <groupId>org.apache.ranger</groupId>
             <artifactId>ranger-plugins-common</artifactId>
             <version>${ranger.version}</version>
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/ComponentVersion.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/ComponentVersion.scala
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.util
+
+/**
+ * Encapsulate a component (Kyuubi/Spark/Hive/Flink) version for the convenience of version checks.
+ * Copy from org.apache.kyuubi.engine.ComponentVersion
+ */
+class ComponentVersion(versionString: String) {
+
+  val (majorVersion, minorVersion, patchVersion) =
+    """^(\d+)\.(\d+)\.(.*?)$""".r.findFirstMatchIn(versionString) match {
+      case Some(m) =>
+        (m.group(1).toInt, m.group(2).toInt, m.group(3).toInt)
+      case None =>
+        throw new IllegalArgumentException(s"Tried to parse '$versionString' as a project" +
+          s" version string, but it could not find the major, minor and patch version numbers.")
+    }
+}
+
+object ComponentVersion {
+
+  def isVersionAtMost(targetVersionString: String, runtimeVersionString: String): Boolean = {
+    compareVersion(
+      targetVersionString,
+      runtimeVersionString,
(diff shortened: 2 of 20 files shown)


extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -39,6 +39,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
```
yaooqinn, line 42: we can duplicate the code, authz aims to be independent

kyuubi-common/src/test/scala/org/apache/kyuubi/engine/ComponentVersionSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+    assert(version.majorVersion === 1)
+    assert(version.minorVersion === 12)
+    assert(version.patchVersion === "4")
+  }
+
+  test("parse snapshot version") {
+    val version = new ComponentVersion("2.14.8-SNAPSHOT")
+    assert(version.majorVersion === 2)
+    assert(version.minorVersion === 14)
+    assert(version.patchVersion === "8-SNAPSHOT")
```
pan3793, line 35: I don't think this is right, patch version should be "8".

In practice, when we fork and maintain a internal version of an open-source project, we add additional suffix in version, e.g. the upstream version is `1.2.3`, the internel version may be `1.2.3.1`, `1.2.3.20220781`, `1.2.3-org`, and those version should be compatible w/ `1.2.3`

deadwind4: The patch version is in various forms. I delete the patch version in this class and think about it in the future.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ComponentVersion.scala
```diff
@@ -0,0 +1,80 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine
+
+/**
+ * Encapsulate a component (Kyuubi/Spark/Hive/Flink) version for the convenience of version checks.
+ */
+class ComponentVersion(versionString: String) {
```
yaooqinn, line 23: case class TernaryVersion(major: Int, minor: Int, patch: Int)

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ComponentVersion.scala
```diff
@@ -0,0 +1,80 @@
...
+        (m.group(1).toInt, m.group(2).toInt, m.group(3))
+      case None =>
+        throw new IllegalArgumentException(s"Tried to parse '$versionString' as a project" +
+          s" version string, but it could not find the major, minor and patch version numbers.")
+    }
+}
+
+object ComponentVersion {
+
+  def isVersionAtMost(targetVersionString: String, runtimeVersionString: String): Boolean = {
```
yaooqinn, line 37: define `def isVersionAtMost(targetVersionString: String)` in the companion class

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ComponentVersion.scala
```diff
@@ -0,0 +1,80 @@
...
+      targetVersionString,
+      runtimeVersionString,
+      (targetMajor: Int, targetMinor: Int, runtimeMajor: Int, runtimeMinor: Int) => {
+        (runtimeMajor > targetMajor) || {
+          runtimeMajor == targetMajor && runtimeMinor >= targetMinor
+        }
+      })
+  }
+
+  def isVersionEqualTo(targetVersionString: String, runtimeVersionString: String): Boolean = {
```
yaooqinn, line 59: ditto

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/ComponentVersion.scala
```diff
@@ -0,0 +1,81 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.util
+
+/**
+ * Encapsulate a component (Kyuubi/Spark/Hive/Flink) version for the convenience of version checks.
```
yaooqinn, line 21: this is a spark oriented module, should not have opportunities to work with hive flink

yaooqinn:
there are also some duplicated functions in Utils.class

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ComponentVersion.scala
deadwind4: I will try to implement `case class BinaryVersion(major: Int, minor: Int)`. IMO, The patch version is not considered at present.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/ComponentVersion.scala
yaooqinn: we can leave this module alone in this PR.

deadwind4: Should I revert this module in this PR?

yaooqinn: it's up to you, i am ok with both

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ComponentVersion.scala
pan3793: How about naming it `SemanticVersion`?

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/ComponentVersion.scala
pan3793: Agreed. 

The following API is more intuitive.
```
ComponentVersion("3.2.2").isVersionAtMost("3.3")
```

extensions/spark/kyuubi-spark-connector-common/src/main/scala/org/apache/kyuubi/spark/connector/common/SparkUtils.scala
```diff
@@ -21,58 +21,15 @@ import org.apache.spark.SPARK_VERSION
 
 object SparkUtils {
```
deadwind4, line 22: Should I delete this `SparkUtils` class?

pan3793 (approved):

extensions/spark/kyuubi-spark-connector-common/src/main/scala/org/apache/kyuubi/spark/connector/common/SparkUtils.scala
pan3793: I'm neutral, either is fine.

pan3793:
Thanks, merging to master

Issue #3194, opened 2022-08-08
[Subtask] Refine deprecated config


cfmcgrady:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

Refine deprecated config to fix the error

```
[ERROR] /home/chenfu/workspace/incubator-kyuubi/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala:353: annotation argument needs to be a constant; found: "using ".+(KyuubiConf.this.FRONTEND_THRIFT_BINARY_BIND_PORT.key).+(" instead")
```

### Parent issue

#3163 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3195, opened 2022-08-08
[KYUUBI #3194][Scala-2.13] Refine deprecated config

cfmcgrady:

1. add a warning message when the user sets the deprecated configuration
```
10:49:58.558 WARN org.apache.kyuubi.config.KyuubiConf: The Kyuubi config 'kyuubi.frontend.bind.port' has been deprecated in Kyuubi v1.4.0 and may be removed in the future. Use kyuubi.frontend.thrift.binary.bind.port instead
```

2. make the annotation `@deprecated` arguments only contain constants to fix the error when compiling for scala-2.13

```
[ERROR] /home/chenfu/workspace/incubator-kyuubi/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala:353: annotation argument needs to be a constant; found: "using ".+(KyuubiConf.this.FRONTEND_THRIFT_BINARY_BIND_PORT.key).+(" instead")
```

Pass CI.


pan3793 (approved):

pan3793:
Thanks, merging to master

PR #3207, opened 2022-08-09
[KYUUBI #3206] Change Flink default version to 1.15

deadwind4:

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilderSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilderSuite.scala
@@ -23,7 +23,7 @@ import java.util
 import scala.collection.JavaConverters._
 import scala.collection.immutable.ListMap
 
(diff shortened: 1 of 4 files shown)


pan3793:
Please update the corresponding GitHub workflow, we should verify the default Flink version with which build should work on all supported versions of Flink distributions.

pan3793 (approved):

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/flink/FlinkProcessBuilderSuite.scala
```diff
@@ -79,7 +80,7 @@ class FlinkProcessBuilderSuite extends KyuubiFunSuite {
 
   private val javaPath = s"${envDefault("JAVA_HOME")}${File.separator}bin${File.separator}java"
   private val flinkSqlClientJarPathSuffix = s"${File.separator}opt${File.separator}" +
-    s"flink-sql-client_$SCALA_COMPILE_VERSION-$FLINK_COMPILE_VERSION.jar"
+    s"flink-sql-client-$FLINK_COMPILE_VERSION.jar"
```
pan3793, line 83: As Kyuubi claims to support both Flink 1.14 and 1.15, we should support both of them.

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #3216, opened 2022-08-10
[KYUUBI #3214] Plan only mode should unset when mode value is incorrect

### LINKED ISSUE #3214: [Bug] Plan only mode should unset when mode value is incorrect
### Code of Conduct

### Search before asking

### Describe the bug

if set `kyuubi.operation.plan.only.mode` an Incorrect value,no matter what sql execute ,return the same error,we can not  set it a correct value anymore
```sql
set kyuubi.operation.plan.only.mode=parser
```
image

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

lsm1:

close #3214 

### manual tests
when we set `kyuubi.operation.plan.only.mode` an incorrect value,unset plan only mode
image
so we can set a correct value 

image


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
```diff
@@ -69,6 +69,11 @@ class SparkSQLOperationManager private (name: String) extends OperationManager(n
       OperationLanguages.withName(lang.toUpperCase(Locale.ROOT)) match {
         case OperationLanguages.SQL =>
           val mode = spark.conf.get(OPERATION_PLAN_ONLY_MODE.key, operationModeDefault)
+          if (!OperationModes.values.exists(_.toString.equalsIgnoreCase(mode))) {
+            spark.conf.unset(OPERATION_PLAN_ONLY_MODE.key)
+            throw new NoSuchElementException(s"Unsupported operation mode: $mode," +
+              s" unset kyuubi.operation.plan.only.mode")
```
cxzl25, line 75: s"unset ${OPERATION_PLAN_ONLY_MODE.key}"

cxzl25:
`FlinkSQLOperationManager` should have a similar problem.

wForget:
unset conf doesn't seem to be a good way:
```
set kyuubi.operation.plan.only.mode=parser; // success
select 1; // exception
select 1; // success without plan mode
```

we can add `UNKNOWN` type to `OperationModes` and throw an exception in `PlanOnlyStatement#runInternal`.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/PlanOnlyOperationSuite.scala
```diff
@@ -114,6 +114,24 @@ class PlanOnlyOperationSuite extends WithKyuubiServer with HiveJDBCTestHelper {
...
+    withSessionConf()(Map(KyuubiConf.OPERATION_PLAN_ONLY_MODE.key -> "parse"))(Map.empty) {
+      withJdbcStatement() { statement =>
+        statement.executeQuery(s"set ${KyuubiConf.OPERATION_PLAN_ONLY_MODE.key}=parser")
+        try {
+          statement.executeQuery("select 1")
+        } catch {
+          case e: Exception =>
+            assert(e.getMessage.contains("The operation mode UNKNOWN doesn't support"))
+        }
+        statement.executeQuery(s"set ${KyuubiConf.OPERATION_PLAN_ONLY_MODE.key}=parse")
```
yaooqinn, line 127: this test works because of `kyuubi.operation.plan.only.excludes`.

So this is actually a side-effect and indeterministic.

kyuubi-server/src/test/scala/org/apache/kyuubi/operation/PlanOnlyOperationSuite.scala
```diff
@@ -114,6 +114,24 @@ class PlanOnlyOperationSuite extends WithKyuubiServer with HiveJDBCTestHelper {
...
   }
 
+  test("kyuubi #3214: Plan only mode with an incorrect value") {
+    withSessionConf()(Map(KyuubiConf.OPERATION_PLAN_ONLY_MODE.key -> "parse"))(Map.empty) {
+      withJdbcStatement() { statement =>
+        statement.executeQuery(s"set ${KyuubiConf.OPERATION_PLAN_ONLY_MODE.key}=parser")
+        try {
+          statement.executeQuery("select 1")
+        } catch {
+          case e: Exception =>
```
yaooqinn, line 124: use intercept instead of try-catch

yaooqinn:
I suggest 
- we implement it in `OperationModes` for reuse in different engines.
- redirect unrecognized modes to NONE with a warning message, it has a deterministic behavior which does not rely on kyuubi.operation.plan.only.excludes

```scala
  object OperationModes extends Enumeration with Logging {
    type OperationMode = Value
    val PARSE, ANALYZE, OPTIMIZE, PHYSICAL, EXECUTION, NONE = Value

    def apply(mode: String): OperationMode = {
      mode.toUpperCase(Locale.ROOT) match {
        case "PARSE" => PARSE
        case "ANALYZE" => ANALYZE
        case "OPTIMIZE" => OPTIMIZE
        case "PHYSICAL" => PHYSICAL
        case "EXECUTION" => EXECUTION
        case "NONE" => NONE
        case other =>
          warn(s"Unsupported operation mode: $mode, using NONE instead")
          NONE
      }
    }
  }
```

Accordingly in OperationManagers,
```scala
       case OperationLanguages.SQL =>
          val mode =
            OperationModes(spark.conf.get(OPERATION_PLAN_ONLY_MODE.key, operationModeDefault))
          mode match {
            case NONE =>
              val incrementalCollect = spark.conf.getOption(OPERATION_INCREMENTAL_COLLECT.key)
                .map(_.toBoolean).getOrElse(operationIncrementalCollectDefault)
              new ExecuteStatement(session, statement, runAsync, queryTimeout, incrementalCollect)
            case mode =>
              new PlanOnlyStatement(session, statement, mode)
          }

```

wForget:
> redirect unrecognized modes to NONE with a warning message, it has a deterministic behavior which does not rely on kyuubi.operation.plan.only.excludes

When we set a wrong plan mode he won't have any exception, then we execute the insert command successfully, which seems dangerous.

> this test works because of kyuubi.operation.plan.only.excludes.
> So this is actually a side-effect and indeterministic.

It looks like this, but if we need to switch plan mode normally, we also rely on `kyuubi.operation.plan.only.excludes`:
```
set kyuubi.operation.plan.only.mode=parse;
set kyuubi.operation.plan.only.mode=none;
```

yaooqinn:
make sense, let's add an unknown type here

yaooqinn:
Modified a bit according to @wForget 's comments

> I suggest
> 
> * we implement it in `OperationModes` for reuse in different engines.
> * redirect unrecognized modes to NONE with a warning message, it has a deterministic behavior which does not rely on kyuubi.operation.plan.only.excludes
> 
> ```scala
>   object OperationModes extends Enumeration with Logging {
>     type OperationMode = Value
>     val PARSE, ANALYZE, OPTIMIZE, PHYSICAL, EXECUTION, NONE, UNKNOWN = Value
> 
>     def apply(mode: String): OperationMode = {
>       mode.toUpperCase(Locale.ROOT) match {
>         case "PARSE" => PARSE
>         case "ANALYZE" => ANALYZE
>         case "OPTIMIZE" => OPTIMIZE
>         case "PHYSICAL" => PHYSICAL
>         case "EXECUTION" => EXECUTION
>         case "NONE" => NONE
>         case other =>
>           warn(s"Unsupported operation mode: $other, using UNKNOWN instead")
>           UNKNOWN
>       }
>     }
>   }
> ```
> 
> Accordingly in OperationManagers,
> 
> ```scala
>        case OperationLanguages.SQL =>
>           val mode =
>             OperationModes(spark.conf.get(OPERATION_PLAN_ONLY_MODE.key, operationModeDefault))
>           spark.conf.set(OPERATION_PLAN_ONLY_MODE.key, mode.toString)

>           mode match {
>             case NONE =>
>               val incrementalCollect = spark.conf.getOption(OPERATION_INCREMENTAL_COLLECT.key)
>                 .map(_.toBoolean).getOrElse(operationIncrementalCollectDefault)
>               new ExecuteStatement(session, statement, runAsync, queryTimeout, incrementalCollect)
>             case mode =>
>               new PlanOnlyStatement(session, statement, mode)
>           }
> ```

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -224,6 +224,7 @@
                             <include>org.apache.curator:curator-recipes</include>
                             <include>org.apache.hive:hive-service-rpc</include>
                             <include>org.apache.thrift:*</include>
+                            <include>org.apache.logging.log4j:*</include>
```
cxzl25, line 227: Do we need to package log4j?

lsm1: without log4j, we get a ClassNotFoundException
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/logging/log4j/Level
        at java.lang.Class.privateGetPublicMethods(Class.java:2902)
        at java.lang.Class.privateGetPublicMethods(Class.java:2911)
        at java.lang.Class.getMethods(Class.java:1615)
        at scala.Enumeration.populateNameMap(Enumeration.scala:193)
        at scala.Enumeration.$anonfun$nameOf$1(Enumeration.scala:212)
Caused by: java.lang.ClassNotFoundException: org.apache.logging.log4j.Level
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
```

pan3793: it should not happen, can you provide a reproducable steps?

lsm1: ok，remove this line of code，build and run ut such as  `org.apache.kyuubi.operation.PlanOnlyOperationSuite`，we will get this error. if the warn string in `OperationModes`  is not import，maybe we can remove it？

pan3793: > build and run ut such as ...

Could you elaborate more about it? e.g. using JDK 11 on macOS, run command "build/mvn xxxx" could cause xxx error

lsm1: ok, using JDK 8 on Ubuntu 20, run commond `build/mvn install -Pspark-3.2 -pl dev/kyuubi-codecov -am -Dtest=none -DwildcardSuites=org.apache.kyuubi.operation.PlanOnlyOperationSuite` could cause `ClassNotFoundException: org.apache.logging.log4j.Level`

cxzl25 (approved):

yaooqinn (approved):

yaooqinn:
LGTM

pan3793:
Thanks, merging to master

PR #3227, opened 2022-08-14
SparkConfParser supports parse bytes and time

pan3793:

SparkConfParser supports parse bytes and time for better user experience.


extensions/spark/kyuubi-spark-connector-common/src/main/scala/org/apache/kyuubi/spark/connector/common/SparkConfParser.scala
```diff
@@ -59,11 +61,19 @@ case class SparkConfParser(
...
+  class TimeConfParser extends ConfParser[Long] {
+    override protected def conversion(value: String): Long = JavaUtils.timeStringAsMs(value)
+  }
+
   abstract class ConfParser[T]() {
     private var optionName: Option[String] = None
     private var sessionConfName: Option[String] = None
     private var tablePropertyName: Option[String] = None
-    private var defaultValue: Option[T] = None
+    private var defaultValue: Option[String] = None
```
wForget: Do we need to add another method for the default value of type string?
like spark:
3d06f5d6a8a08d05c612d30316eeddc

wForget (approved):

pan3793:
Thanks, merging to master

Issue #3170, opened 2022-08-01
[Subtask] Expose thrift binary connection metrics #3156


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

expose connection metrics for thrift binary connection.

### Parent issue

#3169 

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

lightning-L:
reference: https://github.com/apache/hive/blob/9ad5a0b28b136521ded27b4b5418f61b610da788/service/src/java/org/apache/hive/service/cli/thrift/ThriftBinaryCLIService.java

PR #3229, opened 2022-08-14
[KYUUBI #3170] Expose thrift binary connection metrics

lightning-L:

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
@@ -32,6 +32,7 @@ object MetricsConstants {
 
   final private val CONN = KYUUBI + "connection."
   final private val THRIFT_HTTP_CONN = KYUUBI + "thrift.http.connection."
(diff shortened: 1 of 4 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
```diff
@@ -32,6 +32,7 @@ object MetricsConstants {
 
   final private val CONN = KYUUBI + "connection."
   final private val THRIFT_HTTP_CONN = KYUUBI + "thrift.http.connection."
+  final private val BINARY_FRONTEND_CONNECTION = KYUUBI + "binary.frontend.connection."
```
pan3793, line 35:
```suggestion
final private val THRIFT_BINARY_CONN = KYUUBI + "thrift.binary.connection."
```

pan3793 (approved):

lightning-L:
@turboFei

lightning-L:
reference: https://github.com/apache/hive/blob/9ad5a0b28b136521ded27b4b5418f61b610da788/service/src/java/org/apache/hive/service/cli/thrift/ThriftBinaryCLIService.java

pan3793:
Thanks, merging to master

Issue #3222, opened 2022-08-12
[FEATURE] JDBC Authentication Provider for server


bowenliang123:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add JDBC authentication provider as implementation of PasswdAuthenticationProvider out of box.

Compared to currently support authentication methods like Kerberos and LDAP, JDBC source is a much easy and stable source for practical deployment. 

The solution should provide:
- easy to use and config jdbc connection details
- handy to customize query statements for authentication
- provide efficient db connection pooling (optional)

### Describe the solution

Add `JDBC` to `AuthMethods` enum and `JDBCPasswdAuthenticationProvider` into package `org.apache.kyuubi.service.authentication` with listed features:

- specify JDBC driver name and load the driver class
- configs of JDBC url, username and password
- select query with placeholders for checking user and password , like `SELECT * from user_pass_hash where username = ${user} and password = MD5(${password})`
- cache and reuse jdbc connections (optional)

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3235, opened 2022-08-15
[KYUUBI #3222]  JDBC Authentication Provider for server 

bowenliang123:

Add JDBC authentication provider as implementation of PasswdAuthenticationProvider out of box.

Compared to currently support authentication methods like Kerberos and LDAP, JDBC source is a much easy and stable source for practical deployment.

The solution should provide:

easy to use and config jdbc connection details
handy to customize query statements for authentication

Adding JDBC to AuthMethods enum and JDBCPasswdAuthenticationProvider into package org.apache.kyuubi.service.authentication with listed features:

- specify JDBC driver name and load the driver class
- configs of JDBC url, username and password
- select query with placeholders for checking user and password , like `SELECT 1 from user_pass_hash where username = ${user} and password = MD5(${password})`


pan3793:
Can we add a derby/sqlite based ut?

bowenliang123:
> Can we add a derby/sqlite based ut?

Is it suggesting an embedded solution with derby or sqllite for JDBC auth? (and what does 'ut' mean, i didn't get it.)

The initial idea of PR is intended to provide an basic implementation for JDBC auth out of box without depending on other packages other than driver and db itself.

Maybe we could bring derby/sqlite embeded implementation in future PRs.

pan3793:
"ut" is abbr of "unit testing", we need to add unit testing to verify the new feature works as expected and avoid breaking accidentally by future code changes.

bowenliang123:
yes, will add the test suite for it referring the LdapAuthenticationProviderImplSuite and other test cases with derby.

bowenliang123:
Some unit tests added with derby embedded server. Will add more config docs for JdbcAuthenticationProviderImpl.

bowenliang123:
unit tests on derby added.

pan3793:
Hi @bowenliang123, you can use `dev/reformat` to fix the code style issue quickly.

bowenliang123:
@pan3793 Thanks for the reminding. And struggling with the differences with IDEA default code format actions and project style checks. I will fix them very soon. The main part of code, test and docs should be okay.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
```diff
@@ -0,0 +1,232 @@
...
+        throw e
+      case e: Exception =>
+        error("Cannot get user info", e);
+        throw e
+    } finally {
+      closeDbConnection(connection, queryStatement)
+    }
+  }
+
+  private def checkConfigs: Unit = {
```
pan3793, line 97: let add `()` if the function has side-effect.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
```diff
@@ -0,0 +1,232 @@
...
+    def configLog(config: String, value: String): String = s"JDBCAuthConfig: $config = '$value'"
+
+    debug(configLog("Driver Class", dbDriver.orNull))
+    debug(configLog("JDBC URL", dbUrl.orNull))
+    debug(configLog("Database Username", dbUserName.orNull))
+    debug(configLog("Database Password", dbPassword.orNull))
```
pan3793, line 103: it's sensitive data

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
```diff
@@ -0,0 +1,232 @@
...
+    }
+    if (!querySqlInLowerCase.contains("where")) {
+      warn("Query SQL does not contains \"WHERE\" keyword");
+    }
+    if (!querySqlInLowerCase.contains("${username}")) {
+      warn("Query SQL does not contains \"${username}\" placeholder");
+    }
+  }
+
+  private def loadJdbcDriverClass: Unit = {
```
pan3793, line 135: HikariDataSource can be used here to simplify JDBC code
`org.apache.kyuubi.server.metadata.jdbc.JDBCMetadataStore`

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
```diff
@@ -0,0 +1,232 @@
...
+
+    checkConfigs
+
+    loadJdbcDriverClass
+
+    var connection: Connection = null
+    var queryStatement: PreparedStatement = null
+
+    try {
+      connection = DriverManager.getConnection(dbUrl.get, dbUserName.orNull, dbPassword.orNull)
```
pan3793, line 72: prefer to use DataSource rather than DriverManager to retrieve connection

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: It supposed to be view in debug logging and only accessed by limited user. Let me change to log password length only.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: I didn't use HikariCP here for minimum depedency changes, since I have noticed kyuubi-common module does not depend on HikariCP (while kyuubi-server does).

In issue https://github.com/apache/incubator-kyuubi/issues/3222, the pooling feature was mentioned to provide.

Or let me follow your idea and change to use HikariDataSource here.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: thx and added.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
pan3793: It's ok to add HikariCP in kyuubi-common module.
Hikari provides both pooled `HikariDataSource` and unpooled `DriverDataSource` implementations of `DataSource`, which can help us to narrow code differences.

bowenliang123: migrated to use HikariDataSource.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: migrated to use HikariDataSource for pooling and connection retrieving.

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+    datasourceProperties.put("password", dbPasswd)
+
+    authDb = Utils.createTempDir(namePrefix = getClass.getSimpleName)
+    authDb.toFile.delete()
+
+    jdbcUrl = s"jdbc:derby:;databaseName=$authDb;create=true"
+    conn = DriverManager.getConnection(
+      jdbcUrl
+        + ";user=" + dbUser
+        + ";password=" + dbPasswd,
```
pan3793, line 55:
```suggestion
      s"$jdbcUrl;user=$dbUser;password=$dbPasswd",
```

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+      datasourceProperties)
+
+    conn.prepareStatement("CREATE SCHEMA " + dbUser).execute();
```
pan3793, line 58:
```suggestion
    conn.prepareStatement("CREATE SCHEMA $dbUser").execute();
```

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+    super.beforeAll()
+
+    val datasourceProperties = new Properties()
+    datasourceProperties.put("user", dbUser)
+    datasourceProperties.put("password", dbPasswd)
+
+    authDb = Utils.createTempDir(namePrefix = getClass.getSimpleName)
+    authDb.toFile.delete()
+
+    jdbcUrl = s"jdbc:derby:;databaseName=$authDb;create=true"
```
pan3793, line 51: a memory db can be used here to simplify the code
`jdbc:derby:memory:<db_name>;create=true`

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  protected val authUser: String = "liangtiancheng"
+  protected val authPasswd: String = "liangtiancheng"
+
+  protected var conf = new KyuubiConf()
+  var conn: Connection = _
+  var authDb: Path = _
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
```
pan3793, line 42: How about

```
override def beforeAll(): Unit = {
    // init db
    super.beforeAll()
}

override def afterAll(): Unit = {
    super.afterAll()
    // clean up db
}
```
?

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+      case e: Throwable =>
+    }
+  }
+
+  test("authenticate tests") {
+    var providerImpl = new JdbcAuthenticationProviderImpl(conf)
+
+    providerImpl.authenticate(authUser, authPasswd)
+
+    val e1 = intercept[AuthenticationException](providerImpl.authenticate("", ""))
```
pan3793, line 89:
```suggestion
    val e1 = intercept[AuthenticationException] {
      providerImpl.authenticate("", "")
    }
```

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+        " VALUES ('demouser','demopassword'); ")
+    val e9 = intercept[IllegalArgumentException](new JdbcAuthenticationProviderImpl(conf))
+    assert(e9.getMessage.contains("Query SQL must start with \"SELECT\""))
+
+    conf.unset(AUTHENTICATION_JDBC_URL)
+    val e10 = intercept[IllegalArgumentException](new JdbcAuthenticationProviderImpl(conf))
+    assert(e10.getMessage.contains("JDBC url is not configured"))
+  }
+
+  private def genJdbcAuthConfigs: KyuubiConf = {
```
pan3793, line 131: let's init the memeber variable `conf` and use `conf.copy()` instead of call this method every time

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
bowenliang123: agree.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
```diff
@@ -0,0 +1,235 @@
...
+    stmt.setMaxRows(1)
+
+    stmt
+  }
+
+  /**
+   * Gracefully close DB connection
+   *
+   * @param connection
+   * @param statement
```
pan3793, line 200: please remove the unnecessary dummy scaladoc, it helpless

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
bowenliang123: agree. thx for thsi hint.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
```diff
@@ -0,0 +1,235 @@
...
+
+      queryStatement = getAndPrepareQueryStatement(connection, user, password)
+
+      val resultSet = queryStatement.executeQuery()
+
+      if (resultSet == null || !resultSet.next()) {
+        // Auth failed
+        throw new AuthenticationException(s"Password does not match or no such user. user:" +
+          s" $user , password length: ${password.length}")
+      }
```
pan3793, line 84: is the `resultSet` leak?

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
```diff
@@ -0,0 +1,235 @@
...
+    val stmt = connection.prepareStatement(getPreparedSql(authQuerySql.get))
+
+    // Extract placeholder list and use its order to pass parameters
+    val placeholderList: List[String] = getPlaceholderList(authQuerySql.get)
+    for (i <- placeholderList.indices) {
+      val param = placeholderList(i) match {
+        case USERNAME_SQL_PLACEHOLDER => user
+        case PASSWORD_SQL_PLACEHOLDER => password
+        case otherPlaceholder =>
+          error(s"Unrecognized Placeholder In Query SQL: $otherPlaceholder")
```
pan3793, line 182: duplicated error message, the exeception will be propagate and logged

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: dont think so? 
ref to doc of `java.sql.Statement`, `When a Statement object is closed, its current ResultSet object, if one exists, is also closed.`

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -645,6 +647,51 @@ object KyuubiConf {
...
+        "<li>`${username}`</li>" +
+        "<li>`${password}`</li></ul>" +
+        "eg.: query sql `SELECT 1 FROM auth_table WHERE user=${username} AND " +
+        "passwd=MD5(CONCAT(salt,${password}));` " +
+        "will be prepared as: `SELECT 1 FROM auth_table " +
+        "WHERE user=? AND passwd=MD5(CONCAT(salt,?));`" +
+        " with value replacement of `username` and `password` in string type.")
+      .version("1.6.0")
+      .stringConf
+      .createOptional
```
pan3793, line 693: Maybe a default SQL is better. cc @yaooqinn @turboFei WDYT?

`SELECT 1 FROM kyuubi_user WHERE user=${username} AND passwd=MD5(${password})`

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: It is supposed to an error for server admin to notice using wrong/unspported placeholders. are you suggesting just throw an exception here without explicit error logs?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
bowenliang123: A default sql may be not a proper way to suggest. Explict query sql is the key to make use of db , including the hash or encrypt/decrypt funcion, the table(s) for authentication.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
pan3793: Then it's fine, thanks for clarification

pan3793: cc @turboFei, we can add a `JdbcUtils` in common modules to share some common codes, since Jdbc is used in several places.

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
bowenliang123: ok. changed to your suggested style. the previous style refered to `LdapAuthenticationProviderImplSuite`.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: agree.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
pan3793: I'm neutral here.
But for SQL dialect concerns, md5 is a common function supported by most of dbms, ref: https://github.com/apache/spark/pull/37169. But in other hand, md5 w/o salt is not security

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
pan3793: yes, especially for logging same message w/ exception content. the stacktrace will indicate where the error happens.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
bowenliang123: yes, got your idea. Thus the compatibility is still limited , eg. Md5 is not supported in Derby. Plus we are not suggesting users to choose in hash or encryptions, but rather than encourage them to maximize usage in conditions, rules, joins and methods in their senarios, eg. limit max 5 times logs in 1 min with specific prefix in username or different types of password or tokens.

pan3793: ok, let's leave it then.

bowenliang123: > I'm neutral here. For SQL dialect concerns, md5 is a common function supported by most of dbms, ref: [apache/spark#37169](https://github.com/apache/spark/pull/37169). But in other hand, md5 w/o salt is not security

alright. and thx for the link to [apache/spark#37169](https://github.com/apache/spark/pull/37169). it is clear and helpful.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
pan3793: let's do such cleanup on other method as well

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: agree.

kyuubi-common/src/test/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImplSuite.scala
bowenliang123: agree.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/JdbcAuthenticationProviderImpl.scala
bowenliang123: sure. please have a check in latest commit.

pan3793 (approved):

pan3793:
Thanks, merging to master

bowenliang123:
Thanks for all your patience, response and guidance. @pan3793 
I will try to do it better next time with less doc/code/scala style bugs and unhelpful docs. And focus more on ideas and approaches.

Also thanks to my colleagues from Guangfa Securities for feature discussion and testing.  @jeanlyn @Ero98

pan3793:
Thanks for contribution!

bowenliang123:
Glad to have this feature metioned as one of highlights in v1.6.0-incubating release notes. 🎉
Also thanks to my colleague @lxbalex at Guangfa Securities for initial idea of this feature and code samples. 😄

yaooqinn:
👍

PR #3243, opened 2022-08-16
Move trait Logging#initializeLogging to object Logging

cxzl25:
If we use scala enumeration, we need to use logging, and an error will be reported in an environment where log4j2 is not packaged.

This PR just moved the `initializeLogging` method of trait Logging to object Logging.

```scala
XX extends Enumeration with Logging
```

```
        at java.lang.Class.privateGetPublicMethods(Class.java:2902)
        at java.lang.Class.privateGetPublicMethods(Class.java:2911)
        at java.lang.Class.getMethods(Class.java:1615)
        at scala.Enumeration.populateNameMap(Enumeration.scala:193)
        at scala.Enumeration.$anonfun$nameOf$1(Enumeration.scala:212)
Caused by: java.lang.ClassNotFoundException: org.apache.logging.log4j.Level
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
```

Scala enumeration uses reflection, resulting in class not found in some scenarios.

scala.Enumeration#populateNameMap

https://github.com/scala/scala/blob/685017051a2dacf40ee7d8d2636721c1bf00182b/src/library/scala/Enumeration.scala#L188-L193

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Logging.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Logging.scala
@@ -88,81 +88,10 @@ trait Logging {
     if (!Logging.initialized) {
       Logging.initLock.synchronized {
         if (!Logging.initialized) {
-          initializeLogging(isInterpreter)
+          Logging.initializeLogging(isInterpreter, loggerName, logger)
         }
       }
     }
-  }
-
-  private def initializeLogging(isInterpreter: Boolean): Unit = {
-    if (ClassUtils.classIsLoadable("org.slf4j.bridge.SLF4JBridgeHandler")) {
-      // Handles configuring the JUL -> SLF4J bridge
-      SLF4JBridgeHandler.removeHandlersForRootLogger()
-      SLF4JBridgeHandler.install()
-    }
-
-    if (Logging.isLog4j2) {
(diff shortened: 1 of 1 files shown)


cxzl25:
Related PR :  https://github.com/apache/incubator-kyuubi/pull/3216#discussion_r944061488

kyuubi-common/src/main/scala/org/apache/kyuubi/Logging.scala
```diff
@@ -88,13 +88,55 @@ trait Logging {
...
+      rootLogger.getLevel == Level.ERROR &&
+      LogManager.getContext.asInstanceOf[LoggerContext]
+        .getConfiguration.isInstanceOf[DefaultConfiguration])
+  }
 
-  private def initializeLogging(isInterpreter: Boolean): Unit = {
+  private def initializeLogging(
+      isInterpreter: Boolean,
+      loggerName: String,
+      logger: () => Logger): Unit = {
```
pan3793, line 139: thuck could be used here to simplify code
```suggestion
      logger: => Logger): Unit = {
```

pan3793 (approved):

pan3793:
Thanks, merging to master for 1.6

PR #3244, opened 2022-08-16
Bump Hudi 0.12.0

cxzl25:

https://lists.apache.org/thread/pb72xpf3wztx2y1mctd3nwfx4tr5lqjt

https://github.com/apache/hudi/compare/release-0.11.1...release-0.12.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,7 +138,7 @@
         <hive.archive.download.skip>false</hive.archive.download.skip>
         <httpclient.version>4.5.13</httpclient.version>
         <httpcore.version>4.4.15</httpcore.version>
-        <hudi.version>0.11.1</hudi.version>
+        <hudi.version>0.12.0</hudi.version>
         <iceberg.version>0.14.0</iceberg.version>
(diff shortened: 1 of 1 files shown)


pan3793:
What the Hudi 0.12 default build Spark version? I'd prefer to use the same Spark version for testing

cxzl25:
> What's the Hudi 0.12 default build Spark version? I'd prefer to use the same Spark version for testing

Hudi does not have active spark3 profile by default, and the default spark version of spark3 profile is 3.3.0.

https://github.com/apache/hudi/blob/master/pom.xml#L1653-L1724

I verified that Hudi 0.12.0 still has this problem in Spark3.3.0.  https://github.com/apache/incubator-kyuubi/pull/2576#discussion_r867464389

If the problem of hudi-spark-common_2.12 is not solved, we can only introduce hudi bundle to test each spark version, but once the dependencies conflict, we will be in more trouble.

pan3793 (approved):

cxzl25:
Thanks, merging to master/1.6

PR #3252, opened 2022-08-17
Fix the problem that |release| in the document was not replaced correctly

cxzl25:
Fix the problem that `|release|` in the document was not replaced correctly.

Current

image

Fix

image


pan3793 (approved):
Sorry, I missed this one, thanks for fixing.

pan3793:
Thanks, merging to master

Issue #3259, opened 2022-08-17
[FEATURE] Support Hive V2 DataSource in Kyuubi


yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Hey everyone,

In a modern database architecture, users may have a strong need for federated queries. Since there are a large number of Hive warehouse in the history database, we tried to implement the Hive V2 Datasource based on Spark Datasource V2 to meet this need.  for the discussion, see :https://lists.apache.org/thread/fq8ywr58rzf9bycflj1q4fl1xyz2rq2w

To do this, we have some subtask as follow:
- initial implement
- support read code path
- support write code path
- document
- add more tests
- support hive client pool
- support convert DatasourceV2Relation(a.k.a HiveRelationV2) to LocalRelation to improbe performance
- support more V2 API

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

yikf:
cc @yaooqinn @pan3793 @zhaomin1423 @permanentstar

yaooqinn:
where do the Catalog V2 APIs lay in your roadmap?

yikf:
> where do the Catalog V2 APIs lay in your roadmap?

Confused: Does V2API refer to things like supportNamespaces for Catalog or supportPushDown for fileScan?

yaooqinn:
yes

yikf:
My idea is to plan for iterative support, especially after Apache Spark 3.3 is released, and i'll add `support more V2 API` to roadmap, what do you think

yaooqinn:
OK. Additional questions, 
1) What are we planning for multiple spark version support for this feature?
2) How do we support HMS with a lower version of spark built-in HMS client?
3) For file scanning, what APIs will we choose, from spark, hive, or raw implementations of file format?

yikf:
> 1. What are we planning for multiple spark version support for this feature?

yes, but the plan is to start support with Spark 3.2, since the 3.2 V2 API is closer to complete,  no plans for an older version

> 2. How do we support HMS with a lower version of spark built-in HMS client?

catalogPlugin support user cusmot option, like our catalog is v2.3, built-in catalog is v1.2,we can set
```
[...]

PR #3260, opened 2022-08-17
[KYUUBI #3259] Initial implementation of the Hive Connector based on the Spark datasource V2

yikf:

In a modern database architecture, users may have a strong need for federated queries. Since there are a large number of Hive warehouse in the history database, we tried to implement the Hive V2 Datasource based on Spark Datasource V2 to meet this need. for the discussion, see :https://lists.apache.org/thread/fq8ywr58rzf9bycflj1q4fl1xyz2rq2w

This PR is the first step in fixing https://github.com/apache/incubator-kyuubi/issues/3259, having 
- initialization implementation 
- support read code path

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/pom.xml
+++ b/extensions/spark/kyuubi-spark-connector-hive/pom.xml
@@ -1,0 +1,216 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~    http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.apache.kyuubi</groupId>
+        <artifactId>kyuubi-parent</artifactId>
+        <version>1.7.0-SNAPSHOT</version>
+        <relativePath>../../../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>kyuubi-spark-connector-hive</artifactId>
+    <name>Kyuubi Spark Hive Connector</name>
+    <description>A Kyuubi hive connector based on Spark V2 DataSource</description>
+    <packaging>jar</packaging>
+    <url>https://kyuubi.apache.org/</url>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-spark-connector-common_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-spark-connector-common_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-library</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_${scala.binary.version}</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-core_${scala.binary.version}</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-catalyst_${scala.binary.version}</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-hive_${scala.binary.version}</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.scalatestplus</groupId>
+            <artifactId>scalacheck-1-15_${scala.binary.version}</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_${scala.binary.version}</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-common_${scala.binary.version}</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client-api</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client-runtime</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+
+        <!--
+          Spark requires `commons-collections` and `commons-io` but got them from transitive
+          dependencies of `hadoop-client`. As we are using Hadoop Shaded Client, we need add
+          them explicitly. See more details at SPARK-33212.
+          -->
+        <dependency>
+            <groupId>commons-collections</groupId>
+            <artifactId>commons-collections</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>jakarta.xml.bind</groupId>
+            <artifactId>jakarta.xml.bind-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <outputDirectory>target/scala-${scala.binary.version}/classes</outputDirectory>
+        <testOutputDirectory>target/scala-${scala.binary.version}/test-classes</testOutputDirectory>
+
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <configuration>
+                    <shadedArtifactAttached>false</shadedArtifactAttached>
+                    <artifactSet>
+                        <includes>
+                            <include>com.google.guava:guava</include>
+                            <include>org.apache.kyuubi:kyuubi-spark-connector-common_${scala.binary.version}</include>
+                        </includes>
+                    </artifactSet>
+                    <relocations>
+                        <relocation>
+                            <pattern>com.google.common</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.com.google.common</shadedPattern>
+                            <includes>
+                                <include>com.google.common.**</include>
+                            </includes>
+                        </relocation>
+                    </relocations>
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>prepare-test-jar</id>
+                        <phase>test-compile</phase>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveFileIndex.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveFileIndex.scala
@@ -1,0 +1,256 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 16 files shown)


ulysses-you:
thank you for the contribution ! @yikf

ulysses-you:
to pass style check, you can run `./dev/reformat`

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -0,0 +1,410 @@
...
+      hadoopConf = hadoopConf,
+      parser = sessionState.sqlParser,
+      functionResourceLoader = sessionState.resourceLoader)
+  }
+
+  /**
+   * A catalog that interacts with external systems.
+   */
+  lazy val externalCatalog: ExternalCatalogWithListener = {
+    val externalCatalog = new HiveExternalCatalog(sparkConf, hadoopConf)
```
permanentstar, line 117: How do we deliver different conf variables for different hive clusters here, I think the variables should already been merged into the current hadoopConf ?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -0,0 +1,410 @@
...
+
+    // Wrap to provide catalog events
+    val wrapped = new ExternalCatalogWithListener(externalCatalog)
+
+    // Make sure we propagate external catalog events to the spark listener bus
+    wrapped.addListener((event: ExternalCatalogEvent) => postExternalCatalogEvent(sc, event))
```
permanentstar, line 123: Good consideration, but should we need to define a new Event subclass to identify which catalog the event came from ?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -0,0 +1,410 @@
...
+          .listTables(db)
+          .map(ident => Identifier.of(ident.database.map(Array(_)).getOrElse(Array()), ident.table))
+          .toArray
+      case _ =>
+        throw new NoSuchNamespaceException(namespace)
+    }
+  }
+
+  override def loadTable(ident: Identifier): Table = {
+    HiveTable(sparkSession, catalog.getTableMetadata(ident.asTableIdentifier), this)
```
permanentstar, line 143: catalog.getTableMetadata may throw NoSuchTableException(db.table) but can not identify catalog name, maybe better to try catch and make up the catalog info

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -0,0 +1,410 @@
...
+  override def createTable(
+      ident: Identifier,
+      schema: StructType,
+      partitions: Array[Transform],
+      properties: util.Map[String, String]): Table = {
+    val (partitionColumns, maybeBucketSpec) = convertTransforms(partitions.toSeq)
+    val provider = properties.getOrDefault(TableCatalog.PROP_PROVIDER, conf.defaultDataSourceName)
+    val tableProperties = properties.asScala
+    val location = Option(properties.get(TableCatalog.PROP_LOCATION))
+    val storage = DataSource.buildStorageFormatFromOptions(toOptions(tableProperties.toMap))
```
permanentstar, line 155: I think create table a.b.c (...) stored as orc should be also supported in hive dsv2, shall we?

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveConnectorSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+
+package org.apache.kyuubi.spark.connector.hive
+
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.{Row, SparkSession}
+
+import org.apache.kyuubi.KyuubiFunSuite
+import org.apache.kyuubi.spark.connector.common.LocalSparkSession.withSparkSession
+
+class HiveConnectorSuite extends KyuubiFunSuite {
```
pan3793, line 26: We need to supply more test cases to make sure the behavior match v1 (or changes are expected)

extensions/spark/kyuubi-spark-connector-hive/pom.xml
```diff
@@ -0,0 +1,218 @@
...
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.apache.kyuubi</groupId>
+        <artifactId>kyuubi-parent</artifactId>
+        <version>1.6.0-SNAPSHOT</version>
```
pan3793, line 25: This feature is exptect for 1.7, branch-1.6 has been cut, and the master branch has go ahead to 1.7.0-SNAPSHOT, please rebease your PR on master and change the version.

BTW, do you want to support old Spark versions, or just for 3.3 and above?

permanentstar: same concern about spark version, because support of some basic commands diffently amoung spark 3.1/3.2/3.3, ex: `show create table` only works >=3.2, `set/use catalog` only works >=3.3,  if we consider downward compatibility, do we need to  sync up this capabilities?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -0,0 +1,410 @@
...
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+
+import org.apache.kyuubi.spark.connector.hive.HiveTableCatalog.{toCatalogDatabase, CatalogDatabaseHelper, IdentifierHelper, NamespaceHelper}
+
+/**
+ * A [[TableCatalog]] that wrap HiveExternalCatalog to as V2 CatalogPlugin instance to access Hive.
+ */
+class HiveTableCatalog(sparkSession: SparkSession)
+  extends TableCatalog with SQLConfHelper with SupportsNamespaces with Logging {
```
permanentstar, line 51: mybe better if we consider using `StagingTableCatalog` instead of TableCatalog, or plan it as a following improvement

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
yikf: core-site.xml & hive-site.xml is the same in different hiveCatalogs, if have different hive clusters, the exclusive catalog configuration needs to be configured in spark-default.conf(like `spark.sql.catalog.hivev2catalog.spark.sql.hive.metastore.version`), which overrides the default catalog configuration

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
yikf: That's great, but seems `ExternalCatalogWithListener` post spark event(is subclass of ExternalCatalogEvent), not custom event, unless we rewrite `ExternalCatalogWithListener` to post our event, which seems a bit expensive (if it's just for signs), probably

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
yikf: good catch, will add to log these infomations

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveConnectorSuite.scala
yikf: more tests will be added next

extensions/spark/kyuubi-spark-connector-hive/pom.xml
yikf: consent is supported only in version 3.3 or above, as in https://github.com/apache/incubator-kyuubi/issues/3259#issuecomment-1219243969

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
yikf: thanks for reminding, I'll look up the difference between the two

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveFileIndex.scala
```diff
@@ -0,0 +1,269 @@
...
+    val partDirToHivePart = fileIndex.partDirToBindHivePartMap()
+    (partDirs, partDirToHivePart)
+  }
+
+  def filterPartitions(filters: Seq[Expression]): HiveInMemoryFileIndex = {
+    if (table.partitionColumnNames.nonEmpty) {
+      val startTime = System.nanoTime()
+      val selectedPartitions: Seq[BindPartition] = {
+        val partitions =
+          listPartitionsByFilter(sparkSession.sessionState.conf, hiveCatalog, table, filters)
```
permanentstar, line 71: IIUC, this can be simply used by `ExternalCatalogUtils.listPartitionsByFilter(sparkSession.sessionState.conf, hiveCatalog.catalog, table, filters)`

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HivePartitionReaderFactory.scala
```diff
@@ -0,0 +1,197 @@
...
+      if (bindPartition != null) {
+        Utilities.getPartitionDesc(bindPartition)
+      } else null
+
+    val ifc =
+      if (partDesc == null) {
+        hiveTable.getInputFormatClass
+          .asInstanceOf[java.lang.Class[InputFormat[Writable, Writable]]]
+      } else {
+        partDesc.getInputFileFormatClass
```
permanentstar, line 116: I have some doubt here
Runtime code will always reach here since the in `HiveFileIndex::filterPartitions` all filtered partitions will tranforms into `BindPartition`, right?
Does the code decision here is to handle the inconsistent format between  partition and table?  Can you explain how does this situation will happen and if we dynamically handle the case here, what strategy should be used in (dynamic/static) partition overwrite.

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HivePartitionedReader.scala
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{Attribute, SpecificInternalRow}
+import org.apache.spark.sql.catalyst.util.DateTimeUtils
+import org.apache.spark.sql.connector.read.PartitionReader
+import org.apache.spark.sql.execution.datasources.PartitionedFile
+import org.apache.spark.sql.hive.kyuubi.connector.HiveConnectorHelper.{hadoopTableReader, hiveShim}
+import org.apache.spark.unsafe.types.UTF8String
+import org.apache.spark.util.SerializableConfiguration
+
+case class HivePartitionedReader(
```
permanentstar, line 40: We can reuse HadoopTableReader::fillObject to reduce the complexity and code burden, WDUT?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/FilePartitionReader.scala
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.connector.read.PartitionReader
+import org.apache.spark.sql.execution.datasources.SchemaColumnConvertNotSupportedException
+import org.apache.spark.sql.hive.kyuubi.connector.HiveConnectorHelper.inputFileBlockHolder
+import org.apache.spark.sql.internal.SQLConf
+
+// scalastyle:off line.size.limit
+// copy from https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/FilePartitionReader.scala
+// scalastyle:on line.size.limit
+class FilePartitionReader[T](readers: Iterator[HivePartitionedFileReader[T]])
```
permanentstar, line 32: It's same compared with the original one,  why not use the source?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HivePartitionReaderFactory.scala
yikf: > I have some doubt here
> Runtime code will always reach here since the in `HiveFileIndex::filterPartitions` all filtered partitions will tranforms into `BindPartition`, right?

No, if the table is't a partitioned table, partDesc will be null



> Does the code decision here is to handle the inconsistent format between partition and table? Can you explain how does this situation will happen and if we dynamically handle the case here, what strategy should be used in (dynamic/static) partition overwrite.

Hive table partitions have unique metadata, this may be inconsistent with the metadata of the table, Refs: https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-alter-table.html

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/FilePartitionReader.scala
yikf: this is due to generic inconsistencies, the separate classes seem to me to be a little bit cleaner 😄, but i'm ok to follow `FilePartitionReaderFactory ` too

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
yikf: there appear to be additional `staging` implementations that can be implemented in subsequent optimizations

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HivePartitionedReader.scala
yikf: The original implementation does some extra work on each row of data processing, which can actually be public, maybe it's better to initialize it once, WDYT?

pom.xml
```diff
@@ -69,6 +69,7 @@
         <module>extensions/server/kyuubi-server-plugin</module>
         <module>extensions/spark/kyuubi-spark-authz</module>
         <module>extensions/spark/kyuubi-spark-connector-common</module>
+        <module>extensions/spark/kyuubi-spark-connector-hive</module>
```
pan3793, line 72: please refer kudu connector to add it as a spark 3.3 specific module since we decided to support 3.3+ only

yikf: ok~  and it looks like have a flaky test not related to this pr? please take a look, you can click [log](https://pipelines.actions.githubusercontent.com/serviceHosts/618b7dd5-55a0-4451-951d-ef3929531f08/_apis/pipelines/1/runs/27921/signedlogcontent/5?urlExpires=2022-08-22T06%3A37%3A02.1086683Z&urlSigningMethod=HMACV1&urlSignature=LKth2y9vIVX%2B4Yhhza8oaNjxpIkM6f4yVFZUMN6WAww%3D) and search keyword of the `support to interrupt the thrift request if remote engine is broken *** FAILED ***`

pom.xml
```diff
@@ -2186,6 +2186,7 @@
                 <module>extensions/spark/kyuubi-extension-spark-common</module>
                 <module>extensions/spark/kyuubi-extension-spark-3-3</module>
                 <module>extensions/spark/kyuubi-spark-connector-kudu</module>
+                <module>extensions/spark/kyuubi-spark-connector-hive</module>
```
pan3793, line 2189: nit: sort alphabetically

pan3793 (approved):
LGTM, I think it's already in good shape to get in as an inital implement.

extensions/spark/kyuubi-spark-connector-hive/pom.xml
```diff
@@ -0,0 +1,218 @@
...
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-catalyst_${scala.binary.version}</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-hive_${scala.binary.version}</artifactId>
+            <version>${spark.version}</version>
```
pan3793, line 92: nit: manage all deps version in `dependencyManagement` scope of root pom

extensions/spark/kyuubi-spark-connector-hive/pom.xml
```diff
@@ -0,0 +1,218 @@
...
+        <dependency>
+            <groupId>org.scalatestplus</groupId>
+            <artifactId>scalacheck-1-15_${scala.binary.version}</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_${scala.binary.version}</artifactId>
+            <version>${spark.version}</version>
```
pan3793, line 104: ditto

pan3793:
Thanks @yikf and @permanentstar, I'm going to merge this PR first to move forward.

Issue #3226, opened 2022-08-12
[Bug]  RuleAuthorization.checkPrivileges execute multiple times


melin:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

RuleAuthorization.checkPrivileges execute multiple times

### Affects Version(s)

1.4

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3262, opened 2022-08-17
[KYUUBI #3226] Privileges should be checked only once in `RuleAuthorization`

yikf:

fix https://github.com/apache/incubator-kyuubi/issues/3226

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
@@ -22,19 +22,26 @@ import scala.collection.mutable.ArrayBuffer
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.catalyst.trees.TreeNodeTag
 
 import org.apache.kyuubi.plugin.spark.authz.{ObjectType, _}
 import org.apache.kyuubi.plugin.spark.authz.ObjectType._
+import org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization.KYUUBI_AUTHZ_TAG
 import org.apache.kyuubi.plugin.spark.authz.util.AuthZUtils._
 
 class RuleAuthorization(spark: SparkSession) extends Rule[LogicalPlan] {
-  override def apply(plan: LogicalPlan): LogicalPlan = {
-    RuleAuthorization.checkPrivileges(spark, plan)
-    plan
+  override def apply(plan: LogicalPlan): LogicalPlan = plan match {
+    case p if !plan.getTagValue(KYUUBI_AUTHZ_TAG).contains(true) =>
+      RuleAuthorization.checkPrivileges(spark, p)
+      p.setTagValue(KYUUBI_AUTHZ_TAG, true)
+      p
+    case p => p // do nothing if checked privileges already.
   }
 }
 
 object RuleAuthorization {
+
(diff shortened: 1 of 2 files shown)


ulysses-you (approved):
cc @yaooqinn

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
```diff
@@ -22,20 +22,30 @@ import scala.collection.mutable.ArrayBuffer
...
 import org.apache.kyuubi.plugin.spark.authz.{ObjectType, _}
 import org.apache.kyuubi.plugin.spark.authz.ObjectType._
+import org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization.KYUUBI_AUTH_TAG
 import org.apache.kyuubi.plugin.spark.authz.util.AuthZUtils._
 
 class RuleAuthorization(spark: SparkSession) extends Rule[LogicalPlan] {
   override def apply(plan: LogicalPlan): LogicalPlan = {
-    RuleAuthorization.checkPrivileges(spark, plan)
+    plan match {
+      case p if !p.getTagValue(KYUUBI_AUTH_TAG).getOrElse(false) =>
```
yaooqinn, line 35: what if this plan is cached？ for group mode, multiple client users may share the same cache manager, can we add a unit test to verify

yikf: good catch, will add test suites to verify this.

yikf:
@yaooqinn tests is added to verify the multi-session & cache plan scenario;

Even if plan is cached, check is performed for every new query, this is because `QueryExecution.optimizedPlan` did not copy PlanTag when `withCachedData.clone()` [1], but it seems to be in line with our expectations

[1] [Spark QueryExecution base code](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/QueryExecution.scala#L126)

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -59,6 +61,63 @@ abstract class RangerSparkExtensionSuite extends AnyFunSuite
     s"Permission denied: user [$user] does not have [$privilege] privilege on [$resource]"
   }
 
+  test("RuleAuthorization: Should check privileges once only.") {
```
yaooqinn, line 64: let's add `[KYUUBI #3226]` here

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -59,6 +61,63 @@ abstract class RangerSparkExtensionSuite extends AnyFunSuite
...
+    val rule = new RuleAuthorization(spark)
+
+    (1 until 10).foreach { ignore =>
+      rule.apply(logicalPlan)
+    }
+
+    assert(logicalPlan.getTagValue(KYUUBI_AUTH_TAG).getOrElse(0) == 1)
+  }
+
+  test("RuleAuthorization: Another session should also check even if the plan is cached.") {
```
yaooqinn, line 75: ditto

yaooqinn (commented):
[KYUUBI #3226] needs to be added to title

yikf:
thanks @yaooqinn @ulysses-you  for review, addressed comment

yaooqinn (approved):

yaooqinn:
please fix code style

ulysses-you:
thanks, merging to master

PR #3293, opened 2022-08-21
[KYUUBI #3226][FOLLOWUP] Revert - Privileges should be checked only once in RuleAuthorization

yikf:

This pr aims to revert https://github.com/apache/incubator-kyuubi/pull/3262, this is due to the fact that the patch does not support Spark 2.4, but authZ does.


yikf:
we should revert this commit first, then I will have a new PR (more elegant) to solve the authentication problem.

yikf:
and we should backport this revert to branch-1.6

Issue #3295, opened 2022-08-21
[FEATURE] Filter other engine configs when build engine process builder


zhaomin1423:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When we build engine process builder, using all configs, but there are some other engine configs, it is unnecessary.
![Uploading image.png…]()

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3296, opened 2022-08-21
[KYUUBI #3295] [FEATURE] Filter other engine configs when build engine process builder

zhaomin1423:

close https://github.com/apache/incubator-kyuubi/issues/3295#issue-1345447015
image


ulysses-you:
This constraint should be added at `KyuubiConf.getCurrentEngineConfigs` for the default Kyuubi session config.

We can not remove the config which is added by user.

Issue #3298, opened 2022-08-22
[Improvement] Unify the approach of get classLoader in Kyuubi


yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Unify the approach of get classLoader in Kyuubi to `Utils.getContextOrKyuubiClassLoader`

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3299, opened 2022-08-22
[KYUUBI #3298] Unify the approach of get classLoader in Kyuubi

yikf:

fix https://github.com/apache/incubator-kyuubi/issues/3298

Curently, a serval snippets use `getClass.getClassLoader` to get the classLoader in Apache Kyuubi, this pr aims to unify this behavior in `Utils.getContextOrKyuubiClassLoader`

### DIFF
--- a/dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/benchmark/Benchmark.scala
+++ b/dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/benchmark/Benchmark.scala
@@ -17,13 +17,6 @@
 
 package org.apache.kyuubi.tpcds.benchmark
 
-import scala.concurrent._
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.duration.DurationInt
-import scala.language.implicitConversions
-import scala.util.{Failure => SFailure, Success, Try}
-import scala.util.control.NonFatal
-
 import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
 
 // scalastyle:off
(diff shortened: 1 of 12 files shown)


dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/benchmark/Benchmark.scala
```diff
@@ -17,13 +17,6 @@
 
 package org.apache.kyuubi.tpcds.benchmark
 
-import scala.concurrent._
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.duration.DurationInt
-import scala.language.implicitConversions
-import scala.util.{Failure => SFailure, Success, Try}
-import scala.util.control.NonFatal
```
yikf, line 25: `UnusedImports` are removed by spotless

pan3793 (approved):

pan3793:
Hi @yikf, let's make the PR description contains more details, it will be included into commit message, and it's not necessary create an issue for it, a direct PR is fine as well.

yikf:
> Hi @yikf, let's make the PR description contains more details, it will be included into commit message, and it's not necessary create an issue for it, a direct PR is fine as well.

got it, thanks

zhaomin1423 (approved):

pan3793:
Thanks, merging to master

dev/kyuubi-tpcds/src/main/scala/org/apache/kyuubi/tpcds/benchmark/Benchmark.scala
pan3793: @Yikf it's not unused imports, and it breaks the CI.

pan3793: could you please send a follow-up PR to recover the build? otherwise, I need to revert this one.

pan3793: FYI: you can use `./build/mvn clean package -Ptpcds -pl dev/kyuubi-tpcds -am` to verify if it works

PR #3301, opened 2022-08-22
Construct lifetimeTerminatingChecker only when needed

cxzl25:
Construct `lifetimeTerminatingChecker` only when needed.
The default value of `kyuubi.session.engine.spark.max.lifetime` is 0, so the `lifetimeTerminatingChecker` is not enabled, there is no need to construct a `ThreadScheduledExecutor`, and `shutdown` can also be avoided when stopping.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -18,7 +18,7 @@
 package org.apache.kyuubi.engine.spark
 
 import java.time.Instant
-import java.util.concurrent.{CountDownLatch, TimeUnit}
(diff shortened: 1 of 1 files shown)


cxzl25:
cc @lightning-L

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -49,8 +49,7 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
 
   private val shutdown = new AtomicBoolean(false)
 
-  private val lifetimeTerminatingChecker =
-    ThreadUtils.newDaemonSingleThreadScheduledExecutor("spark-engine-lifetime-checker")
+  private var lifetimeTerminatingChecker: Option[ScheduledExecutorService] = None
```
pan3793, line 52: @volatile

pan3793 (approved):

cxzl25:
Thanks, merging to master/1.6.

PR #3315, opened 2022-08-23
Fast fail the kyuubi connection if the engine application has been terminated

turboFei:

We met the case that the LaunchEngine operation is succeed.

But the spark engine met exception and terminated with SUCCESS state.

And then the kyuubi connection stuck until reach the `kyuubi.session.engine.initialize.timeout`.

The engine application terminated in 1.5 minutes and the engine init timeout is 10m.

FYI:
image

So, we shall check the engine application state when waiting the engine initialization, and fast fail the connection if the application has been in terminated state.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -216,7 +216,7 @@ private[kyuubi] class EngineRef(
                 }
                 throw new KyuubiSQLException(
                   s"""
-                     |The engine application has been in terminate state. Please check the engine log.
+                     |The engine application has been terminated. Please check the engine log.
                      |FYI: ${appInfo.toMap.mkString("(\n", ",\n", "\n)")}
                      |""".stripMargin,
                   builder.getError)
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -205,6 +205,21 @@ private[kyuubi] class EngineRef(
             throw error
           }
         }
+
+        engineManager.getApplicationInfo(builder.clusterManager(), engineRefId).foreach { appInfo =>
+          if (ApplicationState.applicationTerminated(appInfo.state)) {
+            engineRef = discoveryClient.getEngineByRefId(engineSpace, engineRefId)
+            if (engineRef.isEmpty) {
+              throw new KyuubiSQLException(
+                s"""
+                   |The engine application has been in terminate state. Please check the engine log.
+                   |FYI: ${appInfo.toMap.mkString("(\n", ",\n", "\n)")}
+                   |""".stripMargin,
+                builder.getError)
+            }
+          }
+        }
+
         if (started + timeout <= System.currentTimeMillis()) {
           val killMessage = engineManager.killApplication(builder.clusterManager(), engineRefId)
           process.destroyForcibly()
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -208,6 +208,10 @@ private[kyuubi] class EngineRef(
 
         engineManager.getApplicationInfo(builder.clusterManager(), engineRefId).foreach { appInfo =>
           if (ApplicationState.isTerminated(appInfo.state)) {
+            MetricsSystem.tracing { ms =>
+              ms.incCount(MetricRegistry.name(ENGINE_FAIL, appUser))
+              ms.incCount(MetricRegistry.name(ENGINE_FAIL, "ENGINE_TERMINATE"))
+            }
             throw new KyuubiSQLException(
               s"""
                  |The engine application has been in terminate state. Please check the engine log.
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -206,19 +206,21 @@ private[kyuubi] class EngineRef(
           }
         }
 
-        Option(engineManager).foreach { engineMgr =>
-          engineMgr.getApplicationInfo(builder.clusterManager(), engineRefId).foreach { appInfo =>
-            if (ApplicationState.isTerminated(appInfo.state)) {
-              MetricsSystem.tracing { ms =>
-                ms.incCount(MetricRegistry.name(ENGINE_FAIL, appUser))
-                ms.incCount(MetricRegistry.name(ENGINE_FAIL, "ENGINE_TERMINATE"))
+        if (exitValue == Some(0)) {
+          Option(engineManager).foreach { engineMgr =>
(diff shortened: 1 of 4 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -339,20 +339,10 @@ class BatchJobSubmission(
...
-    }
+    applicationStatus.map(_.state).exists(ApplicationState.isFailed)
   }
 
   def applicationTerminated(applicationStatus: Option[ApplicationInfo]): Boolean = {
-    applicationStatus.map(_.state).exists {
-      case ApplicationState.FAILED => true
-      case ApplicationState.KILLED => true
-      case ApplicationState.FINISHED => true
-      case ApplicationState.NOT_FOUND => true
```
turboFei, line 354: Here I remove the `NOT_FOUND`, it should not be in terminate state(might has not been submitted to RM). cc @ulysses-you

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -205,6 +205,18 @@ private[kyuubi] class EngineRef(
...
+          if (ApplicationState.isTerminated(appInfo.state)) {
+            throw new KyuubiSQLException(
+              s"""
+                 |The engine application has been in terminate state. Please check the engine log.
+                 |FYI: ${appInfo.toMap.mkString("(\n", ",\n", "\n)")}
+                 |""".stripMargin,
+              builder.getError)
+          }
+        }
+
```
ulysses-you, line 219: I can not get it, what's wrong if we kill a terminated application ? And here should close the submit `process` ?

turboFei: If the engine is terminate，we shall not wait the engine available.

turboFei: Yes， we need close the submit process.

turboFei:
Will decrease the get application info frequency.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: Seems NOT_FOUND is need for batch, revert it.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
ulysses-you: the problem is: the submit process is available but the application is terminated ? can you describe how it happened

turboFei: FYI:
The engineId is `83bdadc9-88df-4c6a-b700-dd1e07dffb65`.

image

and the submit process finished at 1:46:19

image

But kyuubi close the session until the init timeout.
image

ulysses-you: I see this issue, it might be some problem between spark engine initialization and register itself to discovery service when the submit process does not affect the application status (i.e. cluster mode).

In this case, shall we put the application info status check code inside the `exitValue == 0` branch ? this issue should only happen if process exit successfully.

turboFei: we can not rely on the exitValue of spark-submit process.

Maybe we do not wait the app completion and the spark-submit process terminate once the app submitted to resource manager.

ulysses-you: > Maybe we do not wait the app completion and the spark-submit process terminate once the app submitted to resource manager.

in this case, the exitValue of submit process is 0 ..

turboFei:
I can not understand why the UT of EngineRef failed....
It is not expected.

turboFei:
also cc @ulysses-you

ulysses-you:
```
 test("start and get engine address with lock") {
    val id = UUID.randomUUID().toString
    conf.set(KyuubiConf.ENGINE_SHARE_LEVEL, USER.toString)
    conf.set(KyuubiConf.ENGINE_TYPE, SPARK_SQL.toString)
    conf.set(KyuubiConf.FRONTEND_THRIFT_BINARY_BIND_PORT, 0)
    conf.set(HighAvailabilityConf.HA_NAMESPACE, "engine_test")
    conf.set(HighAvailabilityConf.HA_ADDRESSES, getConnectString())
    val engine = new EngineRef(conf, user, id, null)
```

some tests do not have a usable applicationManager ..

turboFei:
> ```
>  test("start and get engine address with lock") {
>     val id = UUID.randomUUID().toString
>     conf.set(KyuubiConf.ENGINE_SHARE_LEVEL, USER.toString)
>     conf.set(KyuubiConf.ENGINE_TYPE, SPARK_SQL.toString)
>     conf.set(KyuubiConf.FRONTEND_THRIFT_BINARY_BIND_PORT, 0)
>     conf.set(HighAvailabilityConf.HA_NAMESPACE, "engine_test")
>     conf.set(HighAvailabilityConf.HA_ADDRESSES, getConnectString())
>     val engine = new EngineRef(conf, user, id, null)
> ```
> 
> some tests do not have a usable applicationManager ..

thanks, it is helpful

turboFei:
test passed, cc @ulysses-you  @yaooqinn @pan3793 @wForget

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
ulysses-you: @turboFei this comment has not resovled.

There are three kinds of status of submit process:
- not exit
- exit normally with code == 0
- exit with code != 0

Do we nned check the status from application manager with option 1 and 3 ?

For option 1, if the submit process is running how can the status from application manager become terminated ? if so, it should be a bug from engine (e.g. spark).
For option 3, we have already fast failed so it does not affect.

turboFei: Seems for 1, it is not needed.

So you mean that, only check the app terminate on case 2.

turboFei: @ulysses-you thanks, resolved.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -205,6 +205,26 @@ private[kyuubi] class EngineRef(
             throw error
           }
         }
+
+        if (exitValue == Some(0)) {
```
ulysses-you, line 209: can we add some comments why we do this

ulysses-you (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
turboFei: thanks, added.

pan3793 (approved):

turboFei:
thanks @ulysses-you  @pan3793 , merged to master

turboFei:
merge to branch-1.6 as well.

cc @pan3793

pan3793:
please change milestone as well

PR #3318, opened 2022-08-24
Transfer the TGetInfoReq to kyuubi engine side to check the connection valid and keep connection alive

turboFei:

Now the connection path is:

`client`-> `kyuubiServer` -> `kyuubiEngine`

Maybe the connection between `client -> kyuubiServer` is valid, but the connection between `kyuubiServer -> kyuubiEngine` is not.

So we need check the whole path.

BTW, before, if customer invoke KyuubiConnection::isValid, it can make the connection between client and kyuubiServer keep alive.

Now, it can make the whole connection path keep alive.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala
@@ -75,7 +75,7 @@ abstract class AbstractSession(
     }
   }
 
-  private def withAcquireRelease[T](userAccess: Boolean = true)(f: => T): T = {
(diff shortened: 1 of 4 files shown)


turboFei:
cc @yaooqinn @pan3793

yaooqinn (approved):

yaooqinn:
BTW, if we get info from engines, shall we respect engine's original info?

turboFei:
> shall we respect engine's original info?

Nice to have.

turboFei:
thanks for the review, merged to master

pan3793:
cherry pick to 1.6

PR #3348, opened 2022-08-26
Specifies none checks to be performed on Javadoc comments for maven-javadoc-plugin

SteNicholas:

Skip maven javadoc plugin for release publish.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2292,6 +2292,7 @@
                         <artifactId>maven-javadoc-plugin</artifactId>
                         <configuration>
                             <quiet>true</quiet>
+                            <doclint>none</doclint>
                         </configuration>
                         <executions>
                             <execution>


pan3793:
The nexus requires `<name>-javadoc.jar` and `<name>-source.jar` exist alongside `<name>.jar`, we should fix the Java Docs instead of skipping it.

externals/kyuubi-trino-engine/pom.xml
```diff
@@ -115,6 +115,23 @@
...
                 </executions>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                    <doclint>none</doclint>
+                    <quiet>true</quiet>
+                </configuration>
```
pan3793, line 125: how about configuring it in parent pom?

pan3793 (approved):

pan3793:
Thanks, merging to master/1.6

PR #3354, opened 2022-08-26
Support to get engine id, name, url in KyuubiConnection

turboFei:

Support to get engine id, name, url in KyuubiConnection.

It is helpful and friendly for BI.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
@@ -53,6 +53,10 @@ object KyuubiSparkUtil extends Logging {
   }
 
   def engineId: String = globalSparkContext.applicationId
+  def engineName: String = globalSparkContext.appName
+  def engineUrl: String = globalSparkContext.getConf.getOption(
+    "spark.org.apache.hadoop.yarn.server.webproxy.amfilter.AmIpFilter.param.PROXY_URI_BASES")
+    .orElse(globalSparkContext.uiWebUrl).getOrElse("")
 
   lazy val diagnostics: String = {
     val sc = globalSparkContext
@@ -62,9 +66,9 @@ object KyuubiSparkUtil extends Logging {
     // scalastyle:off line.size.limit
     // format: off
     s"""
-       |           Spark application name: ${sc.appName}
+       |           Spark application name: $engineName
        |                 application ID: $engineId
-       |                 application web UI: $webUrl
+       |                 application web UI: $engineUrl
        |                 master: ${sc.master}
        |                 deploy mode: ${sc.deployMode}
        |                 version: ${sc.version}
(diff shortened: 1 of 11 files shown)


turboFei:
cc @yaooqinn @wForget

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -62,9 +66,9 @@ object KyuubiSparkUtil extends Logging {
     // scalastyle:off line.size.limit
     // format: off
     s"""
-       |           Spark application name: ${sc.appName}
+       |           Spark application name: $engineName
        |                 application ID: $engineId
-       |                 application web UI: $webUrl
```
wForget, line 67: The webUrl(L63) is not used and needs to be removed.

wForget (commented):
LGTM,

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
turboFei: resolved, thanks

wForget (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
```diff
@@ -164,8 +169,14 @@ class KyuubiSyncThriftClient private (
...
-      .filter(_.containsKey("kyuubi.engine.id"))
-      .map(_.get("kyuubi.engine.id"))
+      .filter(_.containsKey(KYUUBI_ENGINE_ID))
+      .map(_.get(KYUUBI_ENGINE_ID))
+    _engineName = Option(resp.getConfiguration)
+      .filter(_.containsKey(KYUUBI_ENGINE_NAME))
+      .map(_.get(KYUUBI_ENGINE_NAME))
+    _engineUrl = Option(resp.getConfiguration)
+      .filter(_.containsKey(KYUUBI_ENGINE_URL))
+      .map(_.get(KYUUBI_ENGINE_URL))
```
cxzl25, line 179: Here filter and map seem a bit redundant, get should be enough.

cxzl25 (approved):

turboFei:
thanks @wForget @cxzl25 for the review, merging to master

kyuubi-server/src/main/scala/org/apache/kyuubi/client/KyuubiSyncThriftClient.scala
cxzl25: Sorry, I thought it was scala map.

```scala
scala> Option(Map("key"->"v")).map(_.get("k")).flatten
res0: Option[String] = None
```

turboFei:
I found that, if the user use the high version kyuubi jdbc including this pr to access low version kyuubi server.

It will throw `out of sequence` thrift exception.

pan3793:
> I found that, if the user use the high version kyuubi jdbc including this pr to access low version kyuubi server.
> 
> It will throw `out of sequence` thrift exception.

it's critical, we need to fix it

PR #3359, opened 2022-08-26
[KYUUBI #3359] Minor change LdapAuthenticationProviderImpl to adapt AD

jetliu1711:
Compatible with the Microsoft Active Directory LDAP.
Currently, Apache Kyuubi support open LDAP, cannot support Microsoft Active Directory LDAP,I try much more times, So
I try to debug LdapAuthenticationProviderImpl class,I find something wrong with use Active Directory LDAP to connect to kyuubi authorized.it can be connect ,but close connect have something wrong,so I clear the close method,it can be very well。support Active Directory LDAP。


pan3793:
> ... but close connect have something wrong ...

Could please provide more details about the error? e.g. the exception stacktrace

pan3793:
#4152 fully ported LDAP-related functionalities from Apache Hive, it supposes to support Active Directory, would you please help to verify it?

PR #3374, opened 2022-08-30
Support password authentication for Trino engine

hddong:

Support SSL for trino engine.


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -897,6 +897,55 @@ object KyuubiConf {
       .stringConf
       .createOptional
 
+  val ENGINE_TRINO_CONNECTION_PASSWORD: OptionalConfigEntry[String] =
+    buildConf("kyuubi.session.engine.trino.connection.password")
+      .doc("The password used for connecting to trino cluster")
```
pan3793, line 902: fill password in URL but not `beeline -p` is not intuitive i think

hddong: > fill password in URL but not `beeline -p` is not intuitive i think

Yes, but Kyuubi server authentication uses `-P`. And the passwords of server and engine are different at most time.

docs/deployment/settings.md
```diff
@@ -456,6 +456,13 @@ kyuubi.session.engine.startup.error.max.size|8192|During engine bootstrapping, i
 kyuubi.session.engine.startup.maxLogLines|10|The maximum number of engine log lines when errors occur during engine startup phase. Note that this max lines is for client-side to help track engine startup issue.|int|1.4.0
 kyuubi.session.engine.startup.waitCompletion|true|Whether to wait for completion after engine starts. If false, the startup process will be destroyed after the engine is started. Note that only use it when the driver is not running locally, such as yarn-cluster mode; Otherwise, the engine will be killed.|boolean|1.5.0
 kyuubi.session.engine.trino.connection.catalog|&lt;undefined&gt;|The default catalog that trino engine will connect to|string|1.5.0
+kyuubi.session.engine.trino.connection.keystore.password|&lt;undefined&gt;|The keystore password used for connecting to trino cluster|string|1.7.0
```
yaooqinn, line 459: remove session?

yaooqinn: Mostly,
- `kyuubi.session.engine` means how a session prepares or interacts with an engine.
- `kyuubi.engine` means how an engine itself works

hddong: Get it and modified.

pan3793: We should update the existing configurations to make them consistent, and maybe we can remove `connection.` as well?

```
    buildConf("kyuubi.engine.trino.url")
      .version("1.6.0")
      .withAlternative("kyuubi.session.engine.trino.connection.url")
      .stringConf
      .createOptional
```

pan3793 (approved):
Let's change the version to 1.8.0 and merge it

pan3793:
Merged to master

Issue #3376, opened 2022-08-30
[FEATURE] Support `Json description` for PlanOnlyStatement in Spark engine


Yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Support `Json description` for PlanOnlyStatement in Spark engine, Users can perform SQL scheduling based on the JSON description of TreeNode

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3377, opened 2022-08-30
[KYUUBI #3376] PlanOnly supports output in different styles

Yikf:

Fix https://github.com/apache/incubator-kyuubi/issues/3376

This pr aims support planOnly output in different styles (Users can perform SQL scheduling based on the JSON description of TreeNode)


yaooqinn:
how about add `kyuubi.operation.plan.only.output.style=plain,json,etc`

yaooqinn:
what if users want an analyzed json plan?

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/PlanOnlyMode.scala
```diff
@@ -0,0 +1,118 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+import java.util.Locale
+
+import org.apache.kyuubi.KyuubiException
+
+sealed trait PlanOnlyMode {
```
yaooqinn, line 24: can we keep using enum?

yaooqinn: revert the unnecessary change, minimize the diff. looks out of the scope of this pr to me

yaooqinn: ok

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
```diff
@@ -70,14 +71,18 @@ class SparkSQLOperationManager private (name: String) extends OperationManager(n
         case OperationLanguages.SQL =>
           val mode =
             OperationModes(spark.conf.get(OPERATION_PLAN_ONLY_MODE.key, operationModeDefault))
+          val style =
```
yaooqinn, line 74: we can move this inside PlanOnlyStatement, it is much simplier

yaooqinn (approved):

yaooqinn:
thanks, merged to master

PR #3400, opened 2022-09-04
[CI] Run install before checking style

pan3793:

To avoid unexpected errors on bumping version and adding new modules.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -27,15 +27,14 @@ concurrency:
   group: linter-${{ github.ref }}
   cancel-in-progress: true
 
-# no install runs before style check because it is expensive, we can skip build if style fails
-# Well, sometimes when we introduce a new module, it may 'fail' the style check for missing
-# dependency we just create for other module to inherit.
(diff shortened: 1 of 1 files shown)


cfmcgrady:
Do we have a lightweight approach to check if the snapshot is available instead install the project directly?

pan3793:
I think no

yaooqinn:
able to trigger via a new module?

cfmcgrady:
how about

```
mvn dependency:resolve -DincludeGroupIds="org.apache.kyuubi" -DincludeScope="compile" -DexcludeTransitive=true
```

[1] https://maven.apache.org/plugins/maven-dependency-plugin/usage.html#dependency:resolve
[2] https://maven.apache.org/plugins/maven-dependency-plugin/resolve-mojo.html

cfmcgrady:
> how about
> 
> ```
> mvn dependency:resolve -DincludeGroupIds="org.apache.kyuubi" -DincludeScope="compile" -DexcludeTransitive=true
> ```

raised a new PR based on this idea. #3422

pan3793:
Close and in favor #3422

PR #3408, opened 2022-09-05
[TEST] Remove the unstable flink process builder test case

pan3793:

Fix https://github.com/apache/incubator-kyuubi/issues/3405

The test case is unstable, the result depends on whether we pre-build `kyuubi-flink-sql-engine` first, so we just remove it to avoid confusing users.


SteNicholas (approved):
LGTM.

pan3793:
Thanks, merging to master/branch-1.6

Issue #3343, opened 2022-08-25
[Improvement] [AUTHZ] Skip privilege checks for CreateViewCommand of LocalTempView and GlobalTempView


bowenliang123:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently privliges for all persisted and non-persisted view in `CreateViewCommand` checks to Ranger will prevent general view usages. Creating temp views, as `LocalTempView` or `GlobalTempView`, should be excluded for Authz plugin privliges checking. 

As for permanent views, the create action still will be left for checking table privilege, and the query for permanent views was introduced in #3325.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3419, opened 2022-09-06
[MINOR][IMPROVE] Verify whether to skip the temporary view creation authentication, shoud use non-permission user instead of admin

Yikf:

After https://github.com/apache/incubator-kyuubi/issues/3343, we skip  checks for CreateViewCommand of LocalTempView and GlobalTempView, we should verify this by using the user who does not have permission to create the view instead of admin

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
@@ -447,6 +447,26 @@ abstract class RangerSparkExtensionSuite extends AnyFunSuite
         assert(sql(s"show table extended from $db like '$table*'").collect().length === 0))
     }
   }
+
+  test("[KYUUBI #3343] pass temporary view creation") {
+    val tempView = "temp_view"
+    val globalTempView = "global_temp_view"
+    doAs("denyuser", sql(s"CREATE TEMPORARY VIEW $tempView AS select * from values(1)"))
+
+    doAs(
+      "denyuser",
+      sql(s"CREATE OR REPLACE TEMPORARY VIEW $tempView" +
+        s" AS select * from values(1)"))
+
+    doAs(
+      "denyuser",
+      sql(s"CREATE GLOBAL TEMPORARY VIEW $globalTempView AS SELECT * FROM values(1)"))
+
+    doAs(
+      "denyuser",
+      sql(s"CREATE OR REPLACE GLOBAL TEMPORARY VIEW $globalTempView" +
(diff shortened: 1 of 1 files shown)


extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -505,27 +505,23 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
   }
 
   test("[KYUUBI #3343] pass temporary view creation") {
-    val table = "hive_src"
     val tempView = "temp_view"
     val globalTempView = "global_temp_view"
+    doAs("user_w/o_permission", sql(s"CREATE TEMPORARY VIEW $tempView AS select * from values(1)"))
```
yaooqinn, line 510: let's use alphabet only for usernames to avoid surprises?

bowenliang123:
Back in #3343 , these ut is intended to verify someone who has select privilege have the create temp view rights. So maybe we should keep the original ut and add ut for `user_w/o_permission`. They are varying cases.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -505,27 +505,23 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
   }
 
   test("[KYUUBI #3343] pass temporary view creation") {
```
yaooqinn, line 507: btw, we can move this test case to superclass ?

bowenliang123:
> > Back in #3343 , these ut were intended to verify someone who has select privilege on source tables have the create temp view rights. Plus Authz is checking output tables after input tables. With select from table, we could verify input table privilege checks pass and then output tables. So maybe we should keep the original ut and add ut for `user_w/o_permission`. They maybe vary in cases.
> 
> I think we just need to check to see if we skipped the temporary view check

Alright, I am neutral here.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
bowenliang123: If just checking view privileges on select values, consider move to PrivilegesBuilderSuite？
Put ut here is to check views with hive input tables.

yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #3426, opened 2022-09-06
[Improvement] Drop temp view should be skipped permission check


Yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Drop temp view should be skipped permission check

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3427, opened 2022-09-06
[KYUUBI #3426] Drop temp view should be skipped permission check

Yikf:

Fix https://github.com/apache/incubator-kyuubi/issues/3426

This pr aims to skip permission check for drop temp view.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
@@ -68,6 +68,19 @@ object PrivilegesBuilder {
     }
   }
 
+  private def isTempView(
+      tableIdent: TableIdentifier,
+      spark: SparkSession): Boolean = {
+    val parts = tableIdent.database match {
+      case Some(db) =>
+        Seq(db, tableIdent.table)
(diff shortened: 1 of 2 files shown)


yaooqinn:
we have CI jobs for different spark versions, right?

LGTM, if tests are green

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
```diff
@@ -68,6 +68,16 @@ object PrivilegesBuilder {
     }
   }
 
+  private def isTempView(
+      tableIdent: TableIdentifier,
+      spark: SparkSession): Boolean = {
+    val parts = tableIdent match {
+      case TableIdentifier(table, None) => Seq(table)
+      case TableIdentifier(tblName, Some(dbName)) => Seq(dbName, tblName)
```
ulysses-you, line 76: it's better to use such code to have a better compatibility:
```scala
tableIdent.database match {
  case Some(db) =>
    Seq(db, tableIdent.table)
  case _ =>
    Seq(tableIdent.table)
}
```

Since the Spark master branch (3.4) will be:
`TableIdentifier(table: String, database: Option[String], catalog: Option[String])`

yaooqinn (approved):

yaooqinn:
thanks, merged to master

Issue #3420, opened 2022-09-06
[FEATURE] Kyuubi proxy spark on kubernetes driver pod ui


zwangsheng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Help proxy spark on kubernetes cluster deploy-mode spark ui.

### Motivation

When Spark On Kubernetes with cluster deploy-mode, spark driver runs in the pods, using the Kubernetes network. Users cannot directly access the spark ui. 
Related to Spark On Kubernetes Doc[2], you need to `kubectl port-forward` for each spark application, which is very inconvenient. 
So related to Kyuubi Issue 2681[1], we design using kyuubi to help port forward spark ui.

### Describe the solution

See more detail with Google Doc https://docs.google.com/document/d/1FOMELnyAN9JSKeYi3qSgcVukvj8IUzYk7KF0eW5CNc8/edit#

### Additional context

#2681 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

zwangsheng:
Ping @pan3793 @yaooqinn @ulysses-you

minyk:
we're planning similar feature for our product, but using svc object of spark-driver pod is our approach. we think this is more reliable than ip address in the zookeeper.

zwangsheng:
> we're planning similar feature for our product, but using svc object of spark-driver pod is our approach. we think this is more reliable than ip address in the zookeeper.

Thanks for your reply. This issue aims to help Kyuubi to have the ability to expose Spark UI On Kubernetes, so as to more conveniently implement Spark On Kubernetes solutions without relying on external solutions. Of course, we also welcome other solutions that can help us to implement Access UI.

PR #3436, opened 2022-09-07
[SUBTASK] [KYUUBI #3420] Spark SQL Engine register zk with ui info

zwangsheng:

See more in #3420 
This pr aim to improve spark sql engine register zookeeper with spark ui info

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/ZookeeperSparkEngineRegisterSuite.scala
+++ b/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/ZookeeperSparkEngineRegisterSuite.scala
@@ -20,14 +20,14 @@ package org.apache.kyuubi.engine.spark
 import java.nio.charset.StandardCharsets
 import java.util.UUID
 
-import org.apache.kyuubi.config.KyuubiConf.PROXY_KUBERNETES_SPARK_UI_ENABLED
+import org.apache.kyuubi.config.KyuubiConf.PROXY_SPARK_UI_ENABLED
 import org.apache.kyuubi.ha.client.zookeeper.ZookeeperClientProvider
 
 class ZookeeperSparkEngineRegisterSuite extends WithDiscoverySparkSQLEngine
   with WithEmbeddedZookeeper {
   override def withKyuubiConf: Map[String, String] =
     super.withKyuubiConf ++ zookeeperConf ++ Map(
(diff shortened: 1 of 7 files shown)


zwangsheng:
cc @pan3793 @yaooqinn

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/ZookeeperSparkEngineRegisterSuite.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.kyuubi.ha.client.zookeeper.ZookeeperClientProvider
+
+class ZookeeperSparkEngineRegisterSuite extends WithDiscoverySparkSQLEngine
+  with WithEmbeddedZookeeper {
+  override def withKyuubiConf: Map[String, String] =
+    super.withKyuubiConf ++ zookeeperConf ++ Map(
+      PROXY_SPARK_UI_ENABLED.key -> "true",
+      "spark.ui.enabled" -> "true")
+
+  override val namespace: String = s"/kyuubi/deregister_test/${UUID.randomUUID().toString}"
```
pan3793, line 33:
```suggestion
  override val namespace: String = s"/kyuubi/deregister_test/${UUID.randomUUID}"
```

kyuubi-common/src/main/scala/org/apache/kyuubi/service/FrontendService.scala
```diff
@@ -41,4 +41,9 @@ trait FrontendService {
    * An optional `ServiceDiscovery` for [[FrontendService]] to expose itself
    */
   val discoveryService: Option[Service]
+
+  /**
+   * An optional `UI`for each engine expose it's ui
+   */
+  def ui: Option[String] = None
```
pan3793, line 48: How about changing it to 
```
def extraConfs: Map[String, String] = Map.empty
```
?
We may want to expose more information in the future, e.g. metrics

cc @turboFei @yaooqinn @cxzl25 @wForget

turboFei: +1 for 
```
def extraConfs: Map[String, String] = Map.empty
```

zwangsheng: Thanks for suggestion, it has been revised.

PR #3439, opened 2022-09-07
[KYUUBI #3435] Command should not execute when plan only mode is set to PHYSICAL or EXECUTION

### LINKED ISSUE #3435: [BUG] Command should not execute when plan only mode is set to PHYSICAL or EXECUTION
### Code of Conduct

### Search before asking

### Describe the bug

```scala
      case PhysicalMode =>
        val physical = spark.sql(statement).queryExecution.sparkPlan
        iter = new IterableFetchIterator(Seq(Row(physical.toString())))
      case ExecutionMode =>
        val executed = spark.sql(statement).queryExecution.executedPlan 
        iter = new IterableFetchIterator(Seq(Row(executed.toString())))
```
when `kyuubi.operation.plan.only.mode` is set to `PHYSICAL` or `EXECUTION`, we get plan from `queryExecution`, both `sparkPlan` and `executedPlan` will call `assertCommandExecuted()` ,  the  Command plan such as `DropTableCommand` will execute

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

lsm1:

close #3435 


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -107,10 +108,15 @@ class PlanOnlyStatement(
           SQLConf.get.maxToStringFields,
           printOperatorId = false))))
       case PhysicalMode =>
-        val physical = spark.sql(statement).queryExecution.sparkPlan
+        val analyzed = spark.sessionState.analyzer.execute(plan)
+        spark.sessionState.analyzer.checkAnalysis(analyzed)
+        val physical = spark.sessionState.executePlan(analyzed, CommandExecutionMode.SKIP).sparkPlan
```
cxzl25, line 113: Is it better to use optimizedPlan here?
```scala
val optimized = spark.sessionState.optimizer.execute(analyzed)
```

wForget:
This change doesn't seem to be compatible with Spark 3.1, can you make it compatible?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -137,9 +137,13 @@ class PlanOnlyStatement(
...
         iter = new IterableFetchIterator(Seq(Row(optimized.toJSON)))
       case PhysicalMode =>
-        val physical = spark.sql(statement).queryExecution.sparkPlan
+        val analyzed = spark.sessionState.analyzer.execute(plan)
+        spark.sessionState.analyzer.checkAnalysis(analyzed)
+        val optimized = spark.sessionState.optimizer.execute(analyzed)
+        val physical = spark.sessionState.planner.plan(ReturnAnswer(optimized)).next()
         iter = new IterableFetchIterator(Seq(Row(physical.toJSON)))
       case ExecutionMode =>
+        warn("Plan like Command will real execute")
```
lsm1, line 146: ExplainCommand return string, when use  json style ,the plan cannot be wrapped as ExplainCommand.Add warning for this case

yaooqinn:
shall we remove this two? they are dangerous and not very useful

lsm1:
[SPARK-38542](https://github.com/apache/spark/pull/35836) add `assertCommandExecuted` since spark 3.2.0,spark 3.1 is normal. we can fix this case except `execution mode` in json style, can we just remove `execution mode` in json style?

yaooqinn:
ok

yaooqinn:
this introduces unstable private class from spark, not a good solution to me

cxzl25 (approved):

cxzl25:
https://github.com/apache/kyuubi/pull/6575

PR #3454, opened 2022-09-08
[CI] Check Kyuubi modules available only when `**/pom.xml` is changed

cfmcgrady:

Check Kyuubi modules available only when `**/pom.xml` is changed

https://github.com/marketplace/actions/paths-changes-filter#conditional-execution

Pass CI.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -38,6 +38,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            is-pom-changed:
+              - '**/pom.xml'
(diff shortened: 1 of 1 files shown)


pan3793 (approved):

.github/workflows/style.yml
```diff
@@ -47,6 +53,7 @@ jobs:
           check-latest: false
       - name: Check kyuubi modules avaliable
         id: modules-check
+        if: steps.filter.outputs.is-pom-changed == 'true'
```
pan3793, line 56: the step does not match PR title

cfmcgrady: Updated the PR title

yaooqinn (approved):

pan3793:
Thanks, merging to master

PR #3459, opened 2022-09-09
[BUILD] Share jars to reduce binary release tarball size

pan3793:

The 1.6.0 binary tarball is nearly 300MB, we can reduce the size by sharing jars between engine and server, as beeline does.

```
-rw-r--r--   1 chengpan  staff   155M Sep  9 14:21 apache-kyuubi-1.7.0-SNAPSHOT-bin.tgz
```

### DIFF
--- a/build/dist
+++ b/build/dist
@@ -236,10 +236,9 @@ cp -r "$KYUUBI_HOME/kyuubi-assembly/target/scala-$SCALA_VERSION/jars/" "$DISTDIR
 cp "$KYUUBI_HOME"/kyuubi-hive-beeline/target/*.jar "$DISTDIR/beeline-jars/"
 
 # Share the jars w/ server to reduce binary size
-cd $DISTDIR/beeline-jars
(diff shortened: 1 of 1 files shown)


build/dist
```diff
@@ -314,7 +328,7 @@ if [[ "$MAKE_TGZ" == "true" ]]; then
   TARDIR="$KYUUBI_HOME/$TARDIR_NAME"
   rm -rf "$TARDIR"
   cp -R "$DISTDIR" "$TARDIR"
-  tar czf "$TARDIR_NAME.tgz" -C "$KYUUBI_HOME" "$TARDIR_NAME"
+  tar czf "$TARDIR.tgz" -C "$KYUUBI_HOME" "$TARDIR_NAME"
```
pan3793, line 331: it changes the output tgz always under project home, not current home.

hddong:
LGTM, add trino engine works well.

pan3793:
Thanks, merging to master/1.6

pan3793:
Thanks for verification

Issue #3464, opened 2022-09-09
[Subtask] Support for pooling external catalog of hive connector


Yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

Currently, Kyuubi supports hive connector for read/write hive table, it is implemented based on the [Apache Spark DataSource V2](https://www.databricks.com/session/apache-spark-data-source-v2), but there's a potential issue;

Kyuubi use `kyuubi.engine.single.spark.session`=[false](https://kyuubi.apache.org/docs/latest/deployment/settings.html#:~:text=kyuubi.engine.single.spark.session) to provide concurrency sql execution in context isolation, this cause spark.newSession invoked for each transaction, In spark v1 catalog, externalCatalog is shared in the mutiple session, but in catalog v2 architecture, it's big different with v1, v2 catalogs are managed by `CatalogManager` which is [session level](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/internal/SessionState.scala#L97), this means that each session will have a separate catalogManager, so in the case of multiple sessions, hivecatalog will be initialized multiple times. this causes two problems: 1  Multiple sessions may be wasted initializing multiple HiveExternalCatalog, which may cause the JVM namespace to swell. 2  Multiple HiveClient connections may be initialized;

This issue aims to pool externalCatalog to address the potential issues mentioned above.

### Parent issue

https://github.com/apache/incubator-kyuubi/issues/3365

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3465, opened 2022-09-09
[KYUUBI #3464] Support for pooling external catalog

Yikf:

Fix https://github.com/apache/incubator-kyuubi/issues/3464, currently, Kyuubi supports hive connector for read/write hive table, it is implemented based on the [Apache Spark DataSource V2](https://www.databricks.com/session/apache-spark-data-source-v2), but there's a potential issue;

Kyuubi use `kyuubi.engine.single.spark.session`=[false](https://kyuubi.apache.org/docs/latest/deployment/settings.html#:~:text=kyuubi.engine.single.spark.session) to provide concurrency sql execution in context isolation, this cause spark.newSession invoked for each transaction, in spark v1 catalog, externalCatalog is shared in the mutiple session, but in catalog v2 architecture, it's big different with v1, v2 catalogs are managed by `CatalogManager` which is [session level](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/internal/SessionState.scala#L97), this means that each session will have a separate catalogManager, so in the case of multiple sessions, hivecatalog will be initialized multiple times. this causes two problems: 1  multiple sessions may be wasted initializing multiple HiveExternalCatalog, which may cause the JVM namespace to swell. 2  multiple HiveClient connections may be initialized;

This issue aims to pool externalCatalog to address the potential issues mentioned above.

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogManager.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogManager.scala
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.spark.connector.hive
+
+import java.util.concurrent.ConcurrentHashMap
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.hive.kyuubi.connector.HiveBridgeHelper.HiveExternalCatalog
+
+import org.apache.kyuubi.spark.connector.hive.KyuubiHiveConnectorConf.EXTERNAL_CATALOG_SHARE_POLICY
+
+object ExternalCatalogManager {
+  private var manager: ExternalCatalogManager = _
+
+  def getOrCreate(sparkSession: SparkSession): ExternalCatalogManager = {
+    if (manager == null) {
+      synchronized {
+        if (manager == null) {
+          val conf = sparkSession.sessionState.conf
+          manager = ExternalCatalogSharePolicy.fromString(
+            conf.getConf(EXTERNAL_CATALOG_SHARE_POLICY)) match {
+            case OneForAllPolicy => new OneForAllPolicyManager()
+            case OneForOnePolicy => OneForOnePolicyManager
+          }
+        }
+      }
+    }
+    manager
+  }
+
+  private[kyuubi] def reset(): Unit = {
+    if (manager != null) {
+      manager.invalidateAll()
+      manager = null
+    }
+  }
+}
+
+abstract class ExternalCatalogManager {
+
+  def take(ticket: Ticket): HiveExternalCatalog
+
+  def invalidateAll(): Unit = {}
+}
+
+/**
+ * A [[OneForAllPolicy]] policy for the externalCatalog manager, which caches the externalCatalog
+ * according to the catalogName, aiming for only one of each catalog globally.
+ */
+class OneForAllPolicyManager() extends ExternalCatalogManager {
+  private val catalogCache = new ConcurrentHashMap[String, HiveExternalCatalog]()
+
+  override def take(ticket: Ticket): HiveExternalCatalog = {
+    catalogCache.computeIfAbsent(
+      ticket.catalogName,
+      _ => {
+        new HiveExternalCatalog(ticket.sparkConf, ticket.hadoopConf)
+      })
+  }
+}
+
+/**
+ * A [[OneForOnePolicy]] policy for the externalCatalog manager, It doesn't actually cache any
+ * externalCatalog, each session will have its own externalCatalog.
+ */
+object OneForOnePolicyManager extends ExternalCatalogManager {
+
+  override def take(ticket: Ticket): HiveExternalCatalog = {
+    new HiveExternalCatalog(ticket.sparkConf, ticket.hadoopConf)
+  }
+}
+
+case class Ticket(catalogName: String, sparkConf: SparkConf, hadoopConf: Configuration)
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
@@ -38,7 +38,7 @@ import org.apache.spark.sql.connector.catalog.NamespaceChange.RemoveProperty
 import org.apache.spark.sql.connector.expressions.Transform
 import org.apache.spark.sql.execution.datasources.DataSource
 import org.apache.spark.sql.hive.HiveUDFExpressionBuilder
-import org.apache.spark.sql.hive.kyuubi.connector.HiveBridgeHelper.{catalogV2Util, postExternalCatalogEvent, HiveExternalCatalog, HiveMetastoreCatalog, HiveSessionCatalog}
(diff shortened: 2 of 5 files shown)


extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogPool.scala
```diff
@@ -0,0 +1,116 @@
...
+
+import org.apache.kyuubi.spark.connector.hive.KyuubiHiveConnectorConf.{EXTERNAL_CATALOG_POOL_ENABLED, EXTERNAL_CATALOG_POOL_SIZE}
+
+object ExternalCatalogPool {
+  private var pool: ExternalCatalogPool = _
+
+  def getOrCreate(sparkSession: SparkSession): ExternalCatalogPool = {
+    if (pool == null) {
+      synchronized {
+        val conf = sparkSession.sessionState.conf
```
pan3793, line 35: should we check again `if (pool == null)`?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogPool.scala
```diff
@@ -0,0 +1,118 @@
...
+
+case class NoopPool() extends ExternalCatalogPool {
+
+  override def take(ticket: Ticket): HiveExternalCatalog = {
+    val externalCatalog = new HiveExternalCatalog(ticket.sparkConf, ticket.hadoopConf)
+    externalCatalog
+  }
+}
+
+case class Ticket(catalogName: String, sparkConf: SparkConf, hadoopConf: Configuration)
```
pan3793, line 118: Does `SparkConf`, `Configuration` implement hashCode and equals?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogPool.scala
```diff
@@ -0,0 +1,118 @@
...
+  override def invalidateAll(): Unit = {
+    catalogPool.clear()
+  }
+}
+
+case class CatalogQueue(threshold: Int, ticket: Ticket) {
+  private val catalogs = new ArrayBlockingQueue[HiveExternalCatalog](threshold)
+
+  def poll(): HiveExternalCatalog = synchronized {
+    if (canGrow) {
```
ulysses-you, line 85: Should we track the used external catalog ? otherwise this pool has less meaning. If the life of external catalog is same with Spark session, then the status mechine should like:
```
new Kyuubi Session -> new Spark Session -> getOrCreate external catalog

close Kyuubi Session -> noop -> add back the external catalog to the pool
```
If the used external catalog is less than threshold, we can grow. But if over threshold we can throw exception or do other policy, e.g. create a temporary external catalog which will be released when Kyuubi session closed and the behavior just like a thread pool.

ulysses-you: if only for cache, I think the we can make it simple that provides two policies:
1. one for one; which is the current behavior, we can add a new config to enable
2. one for all; which is same with v1 and it's better to make this as default

We can add new policy in future if there exists requirements.

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogCache.scala
```diff
@@ -0,0 +1,132 @@
...
+  def poll(): HiveExternalCatalog = synchronized {
+    if (canGrow) {
+      val catalog = new HiveExternalCatalog(ticket.sparkConf, ticket.hadoopConf)
+      assertExpectedAction(catalogs.offer(catalog))
+      catalog
+    } else {
+      val headCatalog = catalogs.remove()
+      assertExpectedAction(catalogs.offer(headCatalog))
+      headCatalog
+    }
```
ulysses-you, line 117: it's really hard to get this behavior and make it hard to debug if there is something wrong.. I prefer not add this cache policy.

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogCache.scala
```diff
@@ -0,0 +1,132 @@
...
+  override def take(ticket: Ticket): HiveExternalCatalog = {
+    catalogCache.computeIfAbsent(
+      ticket.catalogName,
+      _ => {
+        new HiveExternalCatalog(ticket.sparkConf, ticket.hadoopConf)
+      })
+  }
+}
+
+case class NoopCache() extends ExternalCatalogCache {
```
ulysses-you, line 95: the name looks unreadable, how about: `OneForOnePolicy` and `OneForAllPolicy` ?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogManager.scala
```diff
@@ -0,0 +1,91 @@
...
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.hive.kyuubi.connector.HiveBridgeHelper.HiveExternalCatalog
+
+import org.apache.kyuubi.spark.connector.hive.KyuubiHiveConnectorConf.EXTERNAL_CATALOG_SHARE_POLICY
+
+object ExternalCatalogManager {
+  private var pool: ExternalCatalogManager = _
```
ulysses-you, line 30: `pool` should be named

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogManager.scala
```diff
@@ -0,0 +1,91 @@
...
+        new HiveExternalCatalog(ticket.sparkConf, ticket.hadoopConf)
+      })
+  }
+}
+
+/**
+ * A [[OneForOnePolicy]] policy for the externalCatalog manager, It doesn't actually cache any
+ * externalCatalog, each session will have its own externalCatalog.
+ */
+case class OneForOnePolicyManager() extends ExternalCatalogManager {
```
ulysses-you, line 83: object

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/ExternalCatalogManager.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  override def take(ticket: Ticket): HiveExternalCatalog = {
+    val externalCatalog = new HiveExternalCatalog(ticket.sparkConf, ticket.hadoopConf)
+    externalCatalog
```
ulysses-you, line 87: unnecessary variable

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -53,6 +53,8 @@ class HiveTableCatalog(sparkSession: SparkSession)
 
   def this() = this(SparkSession.active)
 
+  private val externalCatalogPool = ExternalCatalogManager.getOrCreate(sparkSession)
```
ulysses-you, line 56: externalCatalogManager

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/KyuubiHiveConnectorConf.scala
```diff
@@ -0,0 +1,42 @@
...
+  import org.apache.spark.sql.internal.SQLConf.buildStaticConf
+
+  val EXTERNAL_CATALOG_SHARE_POLICY =
+    buildStaticConf("spark.sql.kyuubi.hive.connector.externalCatalog.share.policy")
+      .internal()
+      .doc(s"Indicates the share policy for the externalCatalog in the Kyuubi Connector, we use" +
+        "'all' by default. " +
+        "<li>ALL: Indicate to an external catalog is shared globally with the HiveCatalogs " +
+        "with the same catalogName. </li> " +
+        "<li>ONE: Indicate to an external catalog is used by only one HiveCatalog. </li> ")
```
ulysses-you, line 33: can we preserve the whole name `ONE_FOR_ONE` and `ONE_FOR_ALL` ?  it can help to recall the behavior

ulysses-you (approved):

ulysses-you:
thank you @Yikf , merging to master

PR #3473, opened 2022-09-11
Add Docker Compose based Kyuubi Playground

pan3793:

Add [Docker Compose](https://docs.docker.com/compose/)-based Kyuubi Playground, w/ basic docs.

- Only support Spark engine for now.

- The images are not part of the official apache binary release, so we don't have license issue.

- The images will be published to the docker hub under `nekyuubi` namespace, which is managed by one of the PPMC members, but not the whole PPMC.

- The images support both x86 & arm platforms.

image


pan3793:
cc @Yikf, I have published the docker image to docker hub, would you like to test this PR if you have time?

pan3793:
I think I misunderstand the "daemonic", you are talking about something like `docker compose up -d`, but I'm thinking about running the program inside the container as a daemon.

I will update the README to mention the former.

pan3793:
README has been updated as @Yikf suggested, I tested the playground on ARM Linux, and I'm going to merge this one and left other comments as follow-up.

PR #3486, opened 2022-09-14
Enable Hudi related unit tests

cfmcgrady:

Enable Hudi related unit tests

Pass CI.


pan3793:
-1 for this change.

The Hudi bundle jar is not hygienic, it collects lots of 3rd part jars w/o proper relocation which may cause potential class conflict. e.g. prometheus, curator, http client, parquet, rocksdb

pan3793:
How about separating the Hudi IT into an independent module? The the classpath of server and spark engine modules will not be affected by Hudi bundle jar

Issue #3490, opened 2022-09-15
[Improvement] Improve the KyuubiConf#kyuubiConfEntries code logic


Yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?
Currently, when Kyuubi get or set conf by ConfigEntry, it does not check whether ConfigEntry have been registered, so this pr aims to two points as follow:

- Require `ConfigEntry` has been registered;
- Using copy-on-write for Kyuubi#kyuubiConfEntries to reduce contention in concurrent workloads.

A few days ago we found a bunch of lock blocking problem in the Spark cluster, the stack is as follows, we maybe need improve kyuubiConfEntries to reduce contention in concurrent workloads.
<img width="1340" alt="image" src="https://user-images.githubusercontent.com/51110188/190308718-aeb24c67-9b87-4bf8-be46-86fbb9654eef.png">

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3491, opened 2022-09-15
[KYUUBI #3490] Improve the KyuubiConf#kyuubiConfEntries code logic

Yikf:

Fix https://github.com/apache/incubator-kyuubi/issues/3490
Currently, when Kyuubi get or set conf by `ConfigEntry`, it does not check whether `ConfigEntry` have been registered, so this pr aims to two points as follow:

- Require ConfigEntry has been registered;
- Using copy-on-write for Kyuubi#kyuubiConfEntries to reduce contention in concurrent workloads.


pan3793 (approved):

pan3793:
Thanks, merging to master

Issue #3493, opened 2022-09-15
[Subtask] Add error-classes.json to unify error info


yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

This pr aims to add error-classes.json to unify error info

### Parent issue

https://github.com/apache/incubator-kyuubi/issues/670

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #3494, opened 2022-09-15
[KYUUBI #3493] Add error-classes.json to unify error info

yikf:

This pr aims to add error-classes.json to unify error info.


yaooqinn:
I do not get the point of doing this, even in the spark codebase.

yaooqinn:
Most errors, a.k.a, execution errors, happen on the engine side, I can not image a proper way to recapture them and wrap them with sqlState/errorCode.

So, the errors that we produce and provide to end-users are few, e.g. connection errors and sys err.

PR #3498, opened 2022-09-16
Fix GetTables operation in Flink

cxzl25:
#1646

In Flink, the schema returned by `GetTables` is incorrect, causing some database management tools to not see the table.


cxzl25:
## Fix
image

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetTables.scala
```diff
@@ -21,50 +21,79 @@ import scala.collection.JavaConverters._
...
+                  flinkTable.getComment,
+                  null,
+                  null,
+                  null,
+                  null,
+                  null)
+              }
           }
-      }
+      }.toArray
```
pan3793, line 81: nit: change line 51 `toSeq` to `toArray` could avoid this transformation

pan3793 (approved):
Thanks for the investigation and fixing, LGTM, only minor comments

pan3793:
Thanks, merging to master/1.6

Issue #3482, opened 2022-09-13
[FEATURE] choose one group when using group share level


lianneli:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

When a user has several groups in ugi, user can choose a group needed when using group share level. 

### Motivation

Group level is more efficient than user level, with less time of resource apply and less parameters to optimize. For user, engines for different groups mean different queues and different resources. Now the group is chosen automatically by header group in ugi, which may not flexible for user, as well as the platform in upper level cannot provide a better suggestion for user.

### Describe the solution

em...may be I can solve this problem by using parameters transferred by jdbc connection  

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

bowenliang123:
I am afraid that choosing group name by connection param, without appropriate authentication or management, may cause privilige leaking, group level resource control leaking and etc.

pan3793:
Does Hadoop have the primary group concept?

bowenliang123:
Another user-group binding in my mind is #3308 , fetching UserStore from Ranger which contains user to group(s) relations. But authz is spark engine side plugin.

pan3793:
We can extract the `GroupProvider` traits and provide different mechanism

bowenliang123:
Yes , continue to allow choosing group name via params and cheking with different mechanisms with LDAP、Kerberos、JDBC 、Ranger UserStore and etc to restrict allowed groups.

lianneli:
I thought so once, but get groups from Kerberos or Ranger UserStore makes it more complexible for Kyuubi server. Maybe enhance the `org.apache.kyuubi.service.authentication.PasswdAuthenticationProvider` is more effective. For my scene, select one group name from UGI GroupNames through JDBC is enough.

HaoYang670:
I guess this [PR](https://github.com/apache/kyuubi/pull/4822) can solve the issue. cc @lianneli @pan3793

PR #3509, opened 2022-09-18
[KYUUBI #3482]support select group name from ugi's GroupNames for group share level

lianneli:

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -1348,6 +1348,16 @@ object KyuubiConf {
     .version("1.2.0")
     .fallbackConf(LEGACY_ENGINE_SHARE_LEVEL)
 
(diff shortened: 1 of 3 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1348,6 +1348,16 @@ object KyuubiConf {
     .version("1.2.0")
     .fallbackConf(LEGACY_ENGINE_SHARE_LEVEL)
 
+  val ENGINE_SHARE_LEVEL_GROUP_NAME: ConfigEntry[Option[String]] =
+    buildConf("kyuubi.engine.share.level.group.name")
```
pan3793, line 1352: How about "kyuubi.engine.share.level.group.primaries"(`ConfigEntry[Seq[String]]`)?
And we can define the prefered list of group names.

pan3793: cc @yaooqinn

Issue #3510, opened 2022-09-18
[Bug] Reading an uninitialized log should return empty rowSet


yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Reading an uninitialized log should return empty rowSet , So we can set `EmptyRowSet` for `TFetchResultsResp` to avoid client side NPE if FetchResults fail on server side.

Configure log4j for `ERROR` level, causing the engine side no log file output, so server side throw `java.nio.file. NoSuchFileException`, at this time the client get engine log, throws the NPE.

### Affects Version(s)

1.6.0

### Kyuubi Server Log Output

```logtalk
beeline> !connect jdbc:hive2://B000000335049P:10009/
Connecting to jdbc:hive2://B000000335049P:10009/
Enter username for jdbc:hive2://B000000335049P:10009/:
Enter password for jdbc:hive2://B000000335049P:10009/:
Exception in thread "Thread-3" java.lang.NullPointerException
	at org.apache.kyuubi.jdbc.hive.cli.ColumnBasedSet.<init>(ColumnBasedSet.java:43)
	at org.apache.kyuubi.jdbc.hive.cli.RowSetFactory.create(RowSetFactory.java:31)
	at org.apache.kyuubi.jdbc.hive.KyuubiConnection.getExecLog(KyuubiConnection.java:270)
	at org.apache.hive.beeline.KyuubiCommands.showRemainingLogsIfAny(KyuubiCommands.java:389)
	at org.apache.hive.beeline.KyuubiCommands.access$000(KyuubiCommands.java:30)
	at org.apache.hive.beeline.KyuubiCommands$KyuubiLogRunnable.run(KyuubiCommands.java:559)
	at java.lang.Thread.run(Thread.java:748)
21:56:12.086 [main] ERROR org.apache.kyuubi.jdbc.hive.KyuubiConnection - Error fetching launch engine result
java.lang.NullPointerException: null
	at org.apache.kyuubi.jdbc.hive.cli.ColumnBasedSet.<init>(ColumnBasedSet.java:43) ~[kyuubi-hive-jdbc-1.7.0-SNAPSHOT.jar:1.7.0-SNAPSHOT]
[...]

PR #3511, opened 2022-09-18
[KYUUBI #3510] Reading an uninitialized log should return empty rowSet

Yikf:

Fix https://github.com/apache/incubator-kyuubi/issues/3510

Reading an uninitialized log should return empty rowSet , So we can set EmptyRowSet for TFetchResultsResp to avoid client side NPE if FetchResults fail on server side.

For example: Configure log4j for `ERROR` level, causing the engine side no log file output, so server side throw `java.nio.file. NoSuchFileException`, at this time the client get engine log, throws the NPE.


kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -87,6 +88,8 @@ class OperationLog(path: Path) {
   private lazy val writer = Files.newBufferedWriter(path, StandardCharsets.UTF_8)
   private lazy val reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)
 
+  private var initialized: Boolean = false
```
pan3793, line 91: `@volatile`

pan3793 (approved):

Issue #3512, opened 2022-09-18
[Improvement] Use dedicated ExecutionContext for EventBus async execution


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, Kyuubi EventBus uses `scala.concurrent.ExecutionContext.Implicits.global` to execute async event handler. Generally, it's discouraged to use that global ec, instead, we should create dedicated ec for each workload.

### How should we improve?

Use dedicated ExecutionContext for EventBus async execution

### Are you willing to submit PR?

- [ ] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

pan3793:
In short, the global ec is out of control. It would be shared by the whole JVM, users or other components may use it to run heavy compute workloads, share the global ec may block others workloads or blocked by others. And, it's friendly for debugging if we separate the workloads into dedicated thread pools (ec) w/ unique thread names.

PR #3524, opened 2022-09-20
[KYUUBI #3512] Use dedicated ExecutionContext for EventBus async execution

Yikf:

Fix https://github.com/apache/incubator-kyuubi/issues/3512

Currently, Kyuubi EventBus uses `scala.concurrent.ExecutionContext.Implicits.global` to execute async event handler. Generally, it's discouraged to use that global ec, instead, we should create dedicated ec for each workload, this pr aims to use dedicated ExecutionContext for EventBus async execution

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -2120,4 +2120,26 @@ object KyuubiConf {
         _.toSet.subsetOf(Set("JSON", "JDBC", "CUSTOM")),
         "Unsupported event loggers")
       .createWithDefault(Seq("JSON"))
+
+  val ASYNC_EVENT_HANDLER_POLL_SIZE: ConfigEntry[Int] =
+    buildConf("kyuubi.event.async.pool.size")
+      .doc("Number of threads in the async event handler thread pool")
+      .version("1.7.0")
+      .intConf
+      .createWithDefault(8)
+
+  val ASYNC_EVENT_HANDLER_WAIT_QUEUE_SIZE: ConfigEntry[Int] =
+    buildConf("kyuubi.event.async.pool.wait.queue.size")
+      .doc("Size of the wait queue for the async event handler thread pool")
+      .version("1.7.0")
+      .intConf
+      .createWithDefault(100)
+
+  val ASYNC_EVENT_HANDLER_KEEPALIVE_TIME: ConfigEntry[Long] =
+    buildConf("kyuubi.event.async.pool.keepalive.time")
+      .doc("Time(ms) that an idle async thread of the async event handler thread pool will wait" +
+        " for a new task to arrive before terminating")
+      .version("1.7.0")
+      .timeConf
+      .createWithDefault(Duration.ofSeconds(60).toMillis)
 }
--- a/kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
+++ b/kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
@@ -18,13 +18,15 @@
 package org.apache.kyuubi.events
 
 import scala.collection.mutable
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutorService, Future}
 import scala.reflect.{classTag, ClassTag}
 import scala.util.Try
 
 import org.apache.kyuubi.Logging
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.config.KyuubiConf.{ASYNC_EVENT_HANDLER_KEEPALIVE_TIME, ASYNC_EVENT_HANDLER_POLL_SIZE, ASYNC_EVENT_HANDLER_WAIT_QUEUE_SIZE}
 import org.apache.kyuubi.events.handler.EventHandler
+import org.apache.kyuubi.util.ThreadUtils
 
 /**
  * The [[EventBus]] is responsible for triggering Kyuubi event, registering event handlers and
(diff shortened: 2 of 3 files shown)


kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -42,6 +45,19 @@ sealed trait EventBus {
...
   private val defaultEventBus = EventBusLive()
+  private val conf: KyuubiConf = KyuubiConf()
+  conf.loadFileDefaults()
+
+  private val poolSize = conf.get(ASYNC_EVENT_HANDLER_POLL_SIZE)
+  private val waitQueueSize = conf.get(ASYNC_EVENT_HANDLER_WAIT_QUEUE_SIZE)
+  private val keepAliveMs = conf.get(ASYNC_EVENT_HANDLER_KEEPALIVE_TIME)
+
+  private val syncEventExecutionContext =
+    ExecutionContext.fromExecutorService(ThreadUtils.newDaemonQueuedThreadPool(
```
iodone, line 56: The global `ExecutionContext` is `ForkJoinPool`, we are using `ThreadPoolExecutor` here, which one is suitable for us to use here? Or is there another better one?

iodone: I don't have any suggestions here, because I haven't used asynchronous mode in the actual scenario. If I want to handle the event with asynchronous mode,  prefer to implement it in the `EventHandler`. we can control the event cache in the EventHandler to reduce the number of threads.

iodone: However, it is user-friendly to provide the default asynchronous mode.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -42,6 +45,18 @@ sealed trait EventBus {
 
 object EventBus extends Logging {
   private val defaultEventBus = EventBusLive()
+  private val conf: KyuubiConf = KyuubiConf().loadFileDefaults()
+
+  private val poolSize = conf.get(ASYNC_EVENT_HANDLER_POLL_SIZE)
+  private val waitQueueSize = conf.get(ASYNC_EVENT_HANDLER_WAIT_QUEUE_SIZE)
+  private val keepAliveMs = conf.get(ASYNC_EVENT_HANDLER_KEEPALIVE_TIME)
+
+  private val syncEventExecutionContext =
```
pan3793, line 54: sync or async ?

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -18,13 +18,16 @@
 package org.apache.kyuubi.events
 
 import scala.collection.mutable
+import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.ExecutionContext.Implicits.global
```
pan3793, line 22: where is the global pool used after changing?

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
pan3793: BTW, let's simplify it by using `implicits`

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2120,4 +2120,26 @@ object KyuubiConf {
         _.toSet.subsetOf(Set("JSON", "JDBC", "CUSTOM")),
         "Unsupported event loggers")
       .createWithDefault(Seq("JSON"))
+
+  val ASYNC_EVENT_HANDLER_POLL_SIZE: ConfigEntry[Int] =
+    buildConf("kyuubi.async.event.handler.pool.size")
+      .doc("Number of threads in the async event handler thread pool")
+      .version("1.7.0")
+      .intConf
+      .createWithDefault(100)
```
pan3793, line 2129: 100 threads to handle events?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2120,4 +2120,26 @@ object KyuubiConf {
         _.toSet.subsetOf(Set("JSON", "JDBC", "CUSTOM")),
         "Unsupported event loggers")
       .createWithDefault(Seq("JSON"))
+
+  val ASYNC_EVENT_HANDLER_POLL_SIZE: ConfigEntry[Int] =
+    buildConf("kyuubi.async.event.handler.pool.size")
+      .doc("Number of threads in the async event handler thread pool")
```
pan3793, line 2126: let's use `kyuubi.event.async.` namespaces and short the keys

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -42,6 +44,18 @@ sealed trait EventBus {
...
+  private val poolSize = conf.get(ASYNC_EVENT_HANDLER_POLL_SIZE)
+  private val waitQueueSize = conf.get(ASYNC_EVENT_HANDLER_WAIT_QUEUE_SIZE)
+  private val keepAliveMs = conf.get(ASYNC_EVENT_HANDLER_KEEPALIVE_TIME)
+
+  implicit private lazy val asyncEventExecutionContext: ExecutionContextExecutorService =
+    ExecutionContext.fromExecutorService(ThreadUtils.newDaemonQueuedThreadPool(
+      poolSize,
+      waitQueueSize,
+      keepAliveMs,
+      "sync-event-handler-pool"))
```
pan3793, line 58: async

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
pan3793: handlerPool => pool

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2120,4 +2120,26 @@ object KyuubiConf {
         _.toSet.subsetOf(Set("JSON", "JDBC", "CUSTOM")),
         "Unsupported event loggers")
       .createWithDefault(Seq("JSON"))
+
+  val ASYNC_EVENT_HANDLER_POLL_SIZE: ConfigEntry[Int] =
+    buildConf("kyuubi.event.async.handlerPool.size")
+      .doc("Number of threads in the async event handler thread pool")
+      .version("1.7.0")
+      .intConf
+      .createWithDefault(10)
```
pan3793, line 2129: 8

pan3793 (approved):
Thanks, only minor comments

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
pan3793: the physical core may not be suitable for all workloads, for hybrid workloads like ECS, the machine may have hundreds of cores, but for K8s, typically 2~8 cores per pod.

pan3793:
Thanks, merging to master

Issue #3444, opened 2022-09-08
[FEATURE] The planOnly mode of kyuubi spark engine support SQL lineage output


iodone:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

 Executing sql with the `PlanOnly` mode will get the output of SQL's lineage

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

lianneli:
I'm so curious about the result with SQL's lineage~

iodone:
> I'm so curious about the result with SQL's lineage~

You can refer to this pr https://github.com/apache/incubator-kyuubi/pull/3185

PR #3537, opened 2022-09-21
Remove kyuubi dependency of the spark lineage plugin

wForget:

Remove kyuubi dependency of the spark lineage plugin.

image


wForget:
cc @pan3793 @iodone

iodone:
> cc @pan3793 @iodone

What is the reason for the removal? Is it possible to have the lineage plugin applied to the custom spark engine that is not Kyubi Engine?

wForget:
> Is it possible to have the lineage plugin applied to the custom spark engine that is not Kyubi Engine?

Yes, I would prefer it to remove kyuubi dependencies and be able to use it in any spark application.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkSQLEngineEventListener.scala
```diff
@@ -50,6 +51,13 @@ class SparkSQLEngineEventListener(
     event match {
       case e: SessionEvent => updateSessionStore(e)
       case e: SparkOperationEvent => updateStatementStore(e)
+      case e: SparkListenerEvent if SparkUtilsHelper.lineageClassesArePresent =>
+        e match {
+          case OperationLineageEventWrapper(lineageEvent) =>
+            updateStatementLineageStore(lineageEvent)
```
iodone, line 57: Here we can collect both `operationEvent` and `lineageEvent`, whether to consider adding `statementId` into the `operationEvent` with the same `executionId`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
```diff
@@ -61,6 +61,14 @@ class EngineEventsStore(store: KVStore) {
     }
   }
 
+  def getStatementLineage(executionId: String): Option[SparkOperationLineageEvent] = {
+    try {
```
iodone, line 65: This method has never been called?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkSQLEngineEventListener.scala
wForget: > Here we can collect both `operationEvent` and `lineageEvent`, whether to consider adding `statementId` into the `operationEvent` with the same `executionId`

If we want to get the statementId we may need to use SQLOperationListener and make larger change, should we do that?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/events/EngineEventsStore.scala
wForget: > This method has never been called?

Yes, it's just for convenience we use it later such as displaying lineage in the Spark UI. do i need to remove it?

wForget:
@iodone , thanks for your review, sorry for replying so late.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/kyuubi/SparkSQLEngineEventListener.scala
wForget: How about we get the `statementId` via `qe.sparkSession.sparkContext.getLocalProperty("kyuubi.statement.id")` in `SparkOperationLineageQueryExecutionListener#onSuccess`?

iodone: `SparkOperationEvent` was been stored in the same `kvstore` with `LineageEvent `, can we associate them here?

The listener is a separate thread from the execution thread of the spark session,  and I'm not sure if we can get it correctly.

wForget: The KVStore seems to require statementId to get SparkOperationEvent.

iodone: You are right, it seems to be difficult to correlate the two.

iodone:
@wForget #3444, In order to implement the `PlanOnly` mode of lineage,  imported `SQLParserHelper` with spark-lineage-plugin. This will cause the kyuubi spark engine to strongly depend on spark-lineage-plugin. 
Are you in favor of this usage, and if that holds true, your implementation here could just refer directly to spark-lineage-plugin. WDYT?

yaooqinn:
A general question, what a spark plugin has to rely on the kyuubi code path?

wForget:
> A general question, why does a spark plugin have to rely on the kyuubi code path?

The current lineage plugin is more like a kyuubi spark engine plugin, we rely on kyuubi-events module to produce kyuubi lineage event. This pr wants to change lineage to spark event, in order to remove kyuubi dependency.
However, I thought we might use lineage spark event in engine later such as show it on spark ui, so I introduced lineage plugin dependency in kyuubi engine again.

yaooqinn:
So we can not call it a plugin, it's a upstream package

yaooqinn:
oh, I mess this PR with @iodone's one

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SemanticVersion.scala
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.lineage.helper
+
+/**
+ * Encapsulate a component (Kyuubi/Spark/Hive/Flink etc.) version
+ * for the convenience of version checks.
+ */
+case class SemanticVersion(majorVersion: Int, minorVersion: Int) {
```
yaooqinn, line 24: hmm, before introducing this, the version comparison seems to be very simple. Now, plenty copies of this class..

pan3793:
Glancing at the comments, I think the author and reviewers reached a consensus the lineage plugin should be a pure Spark plugin and should not depend on `kyuubi-common`, `kyuubi-event` modules, but have some arguments on the relationship between `kyuubi-spark-sql-engine` and `kyuubi-spark-lineage`, so, can we narrow this PR scope into making `kyuubi-spark-lineage` independent w/ `kyuubi-event`? Some users want to use this plugin along w/ the latest release `kyuubi-spark-sql-engine` but got class conflict error since there were incompatible changes that occurred in `kyuubi-event`.

wForget:
> Glancing at the comments, I think the author and reviewers reached a consensus the lineage plugin should be a pure Spark plugin and should not depend on `kyuubi-common`, `kyuubi-event` modules, but have some arguments on the relationship between `kyuubi-spark-sql-engine` and `kyuubi-spark-lineage`, so, can we narrow this PR scope into making `kyuubi-spark-lineage` independent w/ `kyuubi-event`?

Good idea. I will narrow this PR scope.

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SemanticVersion.scala
wForget: > hmm, before introducing this, the version comparison seems to be very simple. Now, plenty of copies of this class.

In this plugin we only use isSparkVersionAtMost method, do I need to simply reimplement a method instead of copying `SemanticVersion`?

wForget:
seems unrelated failure:
```
AdminResourceSuite:
  List("serviceUri=localhost:32971;version=1.7.0-SNAPSHOT;refId=8bd0f165-37a6-49cf-af67-d82f7a0b3cbb;sequence=0000000000") had size 1 instead of expected size 0 (AdminResourceSuite.scala:103)
```

iodone:
> seems unrelated failure:
> 
> ```
> AdminResourceSuite:
> - delete engine - user share level *** FAILED ***
>   List("serviceUri=localhost:32971;version=1.7.0-SNAPSHOT;refId=8bd0f165-37a6-49cf-af67-d82f7a0b3cbb;sequence=0000000000") had size 1 instead of expected size 0 (AdminResourceSuite.scala:103)
> ```

I have the same problem in #3558 , but other unrelated changes do not trigger this failure case.

wForget:
> > seems unrelated failure:
> > ```
> > AdminResourceSuite:
> > - delete engine - user share level *** FAILED ***
> >   List("serviceUri=localhost:32971;version=1.7.0-SNAPSHOT;refId=8bd0f165-37a6-49cf-af67-d82f7a0b3cbb;sequence=0000000000") had size 1 instead of expected size 0 (AdminResourceSuite.scala:103)
> > ```
> 
> I have the same problem in #3558 , but other unrelated changes do not trigger this failure case.

This test case doesn't use the kyuubi-spark-lineage plugin, let me re-run the check.

iodone:
> > > seems unrelated failure:
> > > ```
> > > AdminResourceSuite:
> > > - delete engine - user share level *** FAILED ***
> > >   List("serviceUri=localhost:32971;version=1.7.0-SNAPSHOT;refId=8bd0f165-37a6-49cf-af67-d82f7a0b3cbb;sequence=0000000000") had size 1 instead of expected size 0 (AdminResourceSuite.scala:103)
> > > ```
> > 
> > 
> > I have the same problem in #3558 , but other unrelated changes do not trigger this failure case.
> 
> This test case doesn't use the kyuubi-spark-lineage plugin, let me re-run the check.

In my case, this failure case still exists with re-running

wForget:
> In my case, this failure case still exists with re-running

It is being followed up in #3753, let's try again later.

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #3543, opened 2022-09-23
Handle zookeeper watch events

cxzl25:

In #1785, the `stop()` call of `org.apache.kyuubi.ha.client.ServiceDiscovery#stopGracefully` was removed, causing the znode to be recreated after deletion.

Currently in `DeRegisterWatcher`, the `NodeDeleted` event is received, and the znode is not immediately closed, which causes `org.apache.curator.framework.recipes.nodes.PersistentNode#watcher` to receive the `NodeDeleted` event and asynchronously creates the same znode, which makes the service node unable to gracefully shutdown.

In addition, because the watcher is one-time, after receiving the `NodeDataChanged` event, we should re-register the watcher to avoid missing the next `NodeDeleted` event.


kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala
```diff
@@ -385,12 +381,26 @@ class ZookeeperDiscoveryClient(conf: KyuubiConf) extends DiscoveryClient {
     localServiceNode
   }
 
-  class DeRegisterWatcher(instance: String, serviceDiscovery: ServiceDiscovery) extends Watcher {
+  private def watchNode(): Unit = {
+    if (zkClient.checkExists
+        .usingWatcher(watcher.asInstanceOf[Watcher]).forPath(serviceNode.getActualPath) == null) {
+      // No node exists, throw exception
+      throw new KyuubiException(s"Unable to create znode for this Kyuubi " +
```
wForget, line 388: Will kyuubi exit when an exception is thrown when re-watching?

cxzl25: This is another thread (`org.apache.zookeeper.ClientCnxn.EventThread`) handling the watch event, so it won't cause kyuubi to exit.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala
```diff
@@ -385,12 +381,26 @@ class ZookeeperDiscoveryClient(conf: KyuubiConf) extends DiscoveryClient {
...
+    extends Watcher {
     override def process(event: WatchedEvent): Unit = {
       if (event.getType == Watcher.Event.EventType.NodeDeleted) {
-        warn(s"This Kyuubi instance ${instance} is now de-registered from" +
-          s" ZooKeeper. The server will be shut down after the last client session completes.")
+        warn(s"This Kyuubi instance $instance is now de-registered from" +
+          " ZooKeeper. The server will be shut down after the last client session completes.")
+        ZookeeperDiscoveryClient.this.deregisterService()
         serviceDiscovery.stopGracefully()
+      } else if (event.getType == Watcher.Event.EventType.NodeDataChanged) {
```
ulysses-you, line 401: after we fix the re-create node issue during stopGracefully, when will NodeDataChanged happen ?

cxzl25: Not sure in what scenario the `NodeDataChanged` event occurs.
I found that the znode mtime of some users in the production environment has changed, so I added logs and re-watched.
If we use `zkCli.sh` set znode data, a `NodeDataChanged` event will be sent,

cxzl25: Yes.

The `DeRegisterWatcher` implemented by `HiveServer2` only handles the `NodeDeleted` event and does not handle `NodeDataChanged`.

https://github.com/apache/hive/blob/f940a0963f68b790e468f6e3f77721cc4edb2161/service/src/java/org/apache/hive/service/server/HiveServer2.java#L635

If the watcher implemented by `HDFS` has a `NodeDataChanged` event, it will re-watch.

https://github.com/apache/hadoop/blob/d78b0b39a6360c7131662c818dd498fc644484d7/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/ha/ActiveStandbyElector.java#L688-L689

ulysses-you: looks reasonable

ulysses-you:
looks good to me, cc @yaooqinn @pan3793 @turboFei if you have other comments

turboFei (approved):

pan3793 (approved):

pan3793:
Thanks, merging to master/1.6

Issue #3544, opened 2022-09-23
[Improvement] bin/kyuubi stop should check pid after force kill


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

`kill -9 <pid>` can not guarantee process shutdown immediately, we should check and wait until pid disappeared before returning.

Otherwise, `bin/kyuubi restart` may fail w/ bind failed.

![image](https://user-images.githubusercontent.com/26535726/191907295-7eb1cb38-293b-4bb4-949d-eba0e5734f21.png)

### Are you willing to submit PR?

- [ ] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

asagjj:
@hddong Hello, I would like to contribute in this issue. can i have a try?

pan3793:
Of course, PR is welcome, anytime :)

hddong:
Welcome and assigned to you. @asagjj

PR #3546, opened 2022-09-23
[KYUUBI #3544] Fix bin/kyuubi stop check

asagjj:
1、Check and wait until process disappeared before remove pid file. 
2、If pid file was still not be removed，we should remind user to kill and remove manually .

### DIFF
--- a/bin/kyuubi
+++ b/bin/kyuubi
@@ -154,17 +154,19 @@ function stop_kyuubi() {
     TARGET_ID="$(cat "$pid")"
     if [[ $(ps -p "$TARGET_ID" -o comm=) =~ "java" ]]; then
       echo "Stopping $CLASS"
-      kill "$TARGET_ID" && rm -f "$pid"
+      kill "$TARGET_ID"
       for i in {1..20}
       do
         sleep 0.5
         if [[ ! $(ps -p "$TARGET_ID" -o comm=) =~ "java" ]]; then
+          rm -f "$pid"
           break
         fi
       done
 
-      if [[ $(ps -p "$TARGET_ID" -o comm=) =~ "java" ]]; then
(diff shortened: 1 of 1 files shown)


bin/kyuubi
```diff
@@ -154,17 +154,19 @@ function stop_kyuubi() {
...
+          rm -f "$pid"
           break
         fi
       done
 
-      if [[ $(ps -p "$TARGET_ID" -o comm=) =~ "java" ]]; then
-        echo "Failed to stop kyuubi after 10 seconds, try 'kill -9 ${TARGET_ID}' forcefully "
+      if [ -f ${pid} ]; then
+        echo "Failed to stop kyuubi after 10 seconds, try 'kill -9 ${TARGET_ID}' forcefully,
+        and remove pid file: ${pid}"
```
pan3793, line 169: how about run `kill -9 "$TARGET_ID"` here, then wait and check again?

pan3793, line 169:
```suggestion
      if [ -f ${pid} ]; then
        echo "Failed to stop kyuubi after 10 seconds, try 'kill -9 ${TARGET_ID}' forcefully"
        kill -9 ${TARGET_ID}
        for i in {1..20}
        do
          sleep 0.5
          if [[ ! $(ps -p "$TARGET_ID" -o comm=) =~ "java" ]]; then
            rm -f "$pid"
            break
          fi
        done
        if [ -f ${pid} ]; then
          echo "Failed to stop kyuubi forcefully after 10 seconds"
          exit 1
        else
          kyuubi_logo
          echo "Bye!"
        fi
```

pan3793 (approved):

cxzl25 (approved):

Issue #3549, opened 2022-09-24
[Improvement] Support query id in Flink engine


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Flink engine now doesn't support query id. 

### How should we improve?

We could support query id with Flink's own job id.

### Are you willing to submit PR?

- [X] Yes. I can submit a PR independently to improve.
- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #3550, opened 2022-09-24
[KYUUBI #3549] Support query id in Flink engine

link3280:

Flink engine now doesn't support query id. This PR fixes the problem.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -23,6 +23,7 @@ import java.util
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 
+import org.apache.flink.api.common.JobID
 import org.apache.flink.table.api.ResultKind
 import org.apache.flink.table.client.gateway.TypedResult
 import org.apache.flink.table.data.{GenericArrayData, GenericMapData, RowData}
@@ -53,6 +54,8 @@ class ExecuteStatement(
 
   private val operationLog: OperationLog =
     OperationLog.createOperationLog(session, getHandle)
+
+  var jobId: Option[JobID] = None
 
   override def getOperationLog: Option[OperationLog] = Option(operationLog)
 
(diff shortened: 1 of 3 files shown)


externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -152,6 +155,9 @@ class ExecuteStatement(
 
   private def runOperation(operation: Operation): Unit = {
     val result = executor.executeOperation(sessionId, operation)
+    if (result.getJobClient.isPresent) {
+      jobId = Some(result.getJobClient.get().getJobID)
+    }
```
pan3793, line 160: There is helper class `FlinkOperation#RichOptional` could simplify this code.

pan3793:
Does the feature has requirements for Flink version?

link3280:
> Does the feature has requirements for Flink version?

Nope, this should be compatible with Flink 1.12 above.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
link3280: I'm not sue if `FlinkOperation#RichOptional` helps, because it's not simply converting java `Optional<JobClient>` to scala `Option[JobClient]`,  but getting a `JobID` from `Optional<JobClient>` then making it a `Option[JobID]`. Please correct me if I'm wrong.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/FlinkSQLOperationManager.scala
```diff
@@ -181,6 +181,19 @@ class FlinkSQLOperationManager extends OperationManager("FlinkSQLOperationManage
...
+    // otherwise there would be TTransportException
+    operation match {
+      case exec: ExecuteStatement =>
+        if (exec.jobId.isEmpty) {
+          ""
+        } else {
+          exec.jobId.get.toHexString
+        }
+      case plan: PlanOnlyStatement =>
+        ""
```
pan3793, line 194:
```suggestion
    operation match {
      case exec: ExecuteStatement => exec.jobId.map(_.toHexString).getOrElse("")
      case _: PlanOnlyStatement => ""
```

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
pan3793, line 160:
```suggestion
    result.getJobClient.asScala.map(_.getJobID)
```

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #3555, opened 2022-09-26
[FEATURE] Helm install help set kyuubi pod global env

zwangsheng:

On kubernetes case, we may need global env.
We add examples in `docker/helm/values.yaml`


pan3793 (approved):

pan3793:
Thanks, merging to master/1.6

PR #3558, opened 2022-09-26
[KYUUBI #3444] Support the planOnly mode of kyuubi spark engine support SQL lineage

### LINKED ISSUE #3444: [FEATURE] The planOnly mode of kyuubi spark engine support SQL lineage output

iodone:
Closes #3444 


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -136,10 +142,25 @@ class PlanOnlyStatement(
...
       case UnknownMode => throw unknownModeError(mode)
       case _ =>
         throw KyuubiSQLException(s"The operation mode $mode" +
           " doesn't support in Spark SQL engine.")
     }
   }
+
+  private def parseLineage(spark: SparkSession, plan: LogicalPlan): String = {
+    val analyzed = spark.sessionState.analyzer.execute(plan)
+    spark.sessionState.analyzer.checkAnalysis(analyzed)
```
ulysses-you, line 157: `optimized` has done this check ?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -136,10 +142,25 @@ class PlanOnlyStatement(
...
+    val optimized = spark.sessionState.optimizer.execute(analyzed)
+    val lineageParser = new LineageParser { def sparkSession = spark }
+    val lineage = lineageParser.parse(optimized)
+
+    val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
+    val r = mapper.writeValueAsString(lineage)
+    r
```
ulysses-you, line 164: unnecessary variable

ulysses-you (approved):
looks a good feature, cc @yaooqinn @cfmcgrady

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -51,6 +51,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
```
yaooqinn, line 54: any idea about removing this from the engine？ It makes engine module hard for long-term maintenance and cross version support

iodone: @yaooqinn @ulysses-you @cfmcgrady 
https://github.com/apache/incubator-kyuubi/pull/3537#issuecomment-1260305095 Here just discussed the dependency with the plugin. @wForget suggested extracting the common module to reuse `SQLLineageParserHelper`, can you give some other advice?

ulysses-you: the lineage module does not shade the spark, so I think the worst case is the lineage can not work and the engine won't be broken ?

we can add try-catch for those code in engine and narrow the exception

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
iodone: Call the`executeAndCheck` method in the `analyzed` stage when query execution, this method includes `checkAnalysis`, but here just call the `spark.sessionState.analyzer.execute(plan)`, so `optimized` don't do this check

externals/kyuubi-spark-sql-engine/pom.xml
yaooqinn: A spark engine plugin can depend on engine lib(not suggested, but ok) and spark libs for the compilation phase, but the engine lib can not do the opposite.

yaooqinn: how about have a rule or logic inside the lineage plugin, the engine side can push down a config or something, then you determine the output

ulysses-you:
the github action seems get sick.. every commit need approve to run : (

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -93,4 +96,13 @@ object KyuubiSparkUtil extends Logging {
     val runtimeVersion = SemanticVersion(runtimeSparkVer)
     (runtimeVersion.majorVersion, runtimeVersion.minorVersion)
   }
+
+  def parsePlan(spark: SparkSession, plan: LogicalPlan): String = {
+    import org.apache.kyuubi.plugin.lineage.helper.LineageParser
+    val lineageParser =
+      new LineageParser { def sparkSession = spark }
```
ulysses-you, line 103: can we use reflect here ? then we can remove maven dependency

iodone: Using reflection is similar to setting the dependency of spark-lineage-plugin to `optional` and requires that this dependency logic be put into a separate file to pass all the unit tests.

externals/kyuubi-spark-sql-engine/pom.xml
iodone: @yaooqinn Can we use `optional` dependencies here? Refer to the definition of maven `optional`: 

> The idea is that some of the dependencies are only used for certain features in the project and will not be needed if that feature isn't used

yaooqinn: Can we just make the full functionality on the plugin side? IMHO, it is not necessary to include a dependency based on my experience to create more than one plugins

iodone: @yaooqinn I have removed the dependency with your advice

wForget:
We can define a PlanOnlyStatement mode interface in Engine, and then implement it in the lineage plugin.

iodone:
@yaooqinn @wForget This feature has been overlooked. A new version has recently been released. The engine has removed the strong dependency on the plugin by reflecting on the provider class of the plugin. Can you help me review it again?

pan3793:
cc @ulysses-you

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #3545, opened 2022-09-23
[Improvement] [AUTHZ] Rules related to authz may be disabled by the set statement.


cfmcgrady:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

```scala
val stmt = s"CREATE TABLE IF NOT EXISTS default.my_table (key int, value int) USING parquet"
try {
  spark.sql(stmt)
} catch {
  case e: Exception =>
    e.printStackTrace()
}
try {
  spark.sql("desc default.my_table").show(truncate = false)
} catch {
  case e: Exception =>
    e.printStackTrace()
}
spark.sql("set spark.sql.optimizer.excludedRules=org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization")
spark.sql(stmt)
spark.sql("desc default.my_table").show(truncate = false)
```

<details>
    <summary>OUTPUT</summary>

```
org.apache.kyuubi.plugin.spark.authz.AccessControlException: Permission denied: user [fchen] does not have [create] privilege on [default/my_table]
	at org.apache.kyuubi.plugin.spark.authz.ranger.SparkRangerAdminPlugin$.verify(SparkRangerAdminPlugin.scala:128)
	at org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization$.$anonfun$checkPrivileges$5(RuleAuthorization.scala:88)
	at org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization$.$anonfun$checkPrivileges$5$adapted(RuleAuthorization.scala:87)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization$.checkPrivileges(RuleAuthorization.scala:87)
	at org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization.apply(RuleAuthorization.scala:36)
[...]

PR #3559, opened 2022-09-26
[KYUUBI #3545] Forbid running the set statement to disable the AUTHZ rules

cfmcgrady:

The PR injected a new Spark parser to disable the set statement that the key equals `spark.sql.optimizer.excludedRules` and the value contains `org.apache.kyuubi.plugin.spark.authz.ranger.*`.

```
set spark.sql.optimizer.excludedRules=org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization
```

```
Authz rules cannot be disabled by a set statement.
org.apache.kyuubi.plugin.spark.authz.AccessControlException: Authz rules cannot be disabled by a set statement.
	at org.apache.kyuubi.plugin.spark.authz.ranger.DisabledExcludeAuthzParser.parsePlan(DisabledExcludeAuthzParser.scala:38)
```

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz/pom.xml
@@ -294,6 +294,26 @@
                 <directory>${project.basedir}/src/test/resources</directory>
             </testResource>
         </testResources>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-sources</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>src/main/spark-${spark.binary.version}</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
     </build>
 
 </project>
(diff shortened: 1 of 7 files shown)


cfmcgrady:
HI, @Yikf thank you for your input.

> There are 3 cases to configure this key for users:
>
>  1) spark.sessionState.conf.setConfString(OPTIMIZER_EXCLUDED_RULES.key, somevalues)
>  2) As a default configuration in spark-default.conf
>  3) Set command

I think you miss another case to specify this config

4) [Via JDBC Connection URL](https://kyuubi.apache.org/docs/r1.6.0-incubating/deployment/settings.html#via-jdbc-connection-url)

The Kyuubi end users can modify configuration values through methods 1, 3, and 4. This PR only deals with method 3 and I also think it is difficult to restrict users from method 1 (in the scala execution mode)

> logging a WARN only instead of an outright failure.

I don't think logging a WARN message is safe in a production environment

cfmcgrady:
Sorry for not explaining clearly before @Yikf 

> I think that it's better to add a analyzer rule FunnelExcludableRule to hanld these scenarios,

Since Spark does not provide a mechanism to prevent end-user exclusion rules, then the Kyuubi end users can exclude `FunnelExcludableRule` and AUTHZ-related rules. So I don't think adding new a rule is a safe way to intercept this configuration, we need to deal with it one by one.

```sql
set spark.sql.optimizer.excludedRules=org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization,FunnelExcludableRule
```

cfmcgrady:
thank you for your explanation, will raise a new PR based on your idea. @Yikf

Issue #3563, opened 2022-09-27
[FEATURE] [authZ/config] enhancement about authorization


lordk911:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

1、when user change language to scala , he/she can directly read a table's data dir , then the the table's column privilege will leak.
2、user also can use sql directly read a table's data file , for example : 
``select * from parquet.`/path/xxx`` 
even we can use config `spark.sql.runSQLOnFiles=false` to disable it , but it can be overwrite.

so, firstly we may need the "change language" operation can be  config to disable, or make only special user can do it. secondly some config item need to be protected to can not be overwriten.

### Are you willing to submit PR?

- [ ] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

yaooqinn:
https://github.com/apache/submarine/blob/release-0.6.0-RC0/submarine-security/spark-security/src/main/scala/org/apache/spark/sql/catalyst/optimizer/SubmarineConfigurationCheckExtension.scala


Oh, I haven't move the feature here.

lordk911:
> https://github.com/apache/submarine/blob/release-0.6.0-RC0/submarine-security/spark-security/src/main/scala/org/apache/spark/sql/catalyst/optimizer/SubmarineConfigurationCheckExtension.scala

I have a test about SubmarineConfigurationCheckExtension , it can banning SetCommand like `set spark.sql.runSQLOnFiles=false;`
but user can modify this kind of config on connection string.

yaooqinn:
> but user can modify this kind of config on connection string.

the server also have config ban list or ignore list for sessions

lordk911:
> the server also have config ban list or ignore list for sessions

![image](https://user-images.githubusercontent.com/19989300/192708457-04cf97cf-7330-4d39-af78-56ae6da5cad6.png)

PR #3564, opened 2022-09-27
[KYUUBI #3545][KYUUBI #3563] Support restrict spark configurations

cfmcgrady:

ban end-user from security settings

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
@@ -28,20 +28,18 @@ import org.apache.kyuubi.plugin.spark.authz.AccessControlException
  */
 case class AuthzConfigurationCheckExtension(spark: SparkSession) extends (LogicalPlan => Unit) {
 
-  final val RESTRICT_LIST_KEY = "spark.sql.kyuubi.conf.restricted.list"
+  final val RESTRICT_LIST_KEY = "spark.kyuubi.conf.restricted.list"
 
-  private val bannedList: Seq[String] =
-    RESTRICT_LIST_KEY ::
-      "spark.sql.runSQLOnFiles" ::
-      "spark.sql.extensions" ::
-      spark.conf.getOption(RESTRICT_LIST_KEY).map(_.split(',').toList).getOrElse(Nil)
+  private val restrictedConfList: Set[String] =
+    Set(RESTRICT_LIST_KEY, "spark.sql.runSQLOnFiles", "spark.sql.extensions") ++
+      spark.conf.getOption(RESTRICT_LIST_KEY).map(_.split(',').toSet).getOrElse(Set.empty)
 
   override def apply(plan: LogicalPlan): Unit = plan match {
     case SetCommand(Some((
           "spark.sql.optimizer.excludedRules" | "spark.sql.adaptive.optimizer.excludedRules",
           Some(v)))) if v.contains("org.apache.kyuubi.plugin.spark.authz.ranger") =>
       throw new AccessControlException("Excluding Authz security rules is not allowed")
-    case SetCommand(Some((k, Some(_)))) if bannedList.contains(k) =>
+    case SetCommand(Some((k, Some(_)))) if restrictedConfList.contains(k) =>
       throw new AccessControlException(s"Modifying config $k is not allowed")
     case _ =>
   }
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
@@ -42,7 +42,7 @@ case class AuthzConfigurationCheckExtension(spark: SparkSession) extends (Logica
           Some(v)))) if v.contains("org.apache.kyuubi.plugin.spark.authz.ranger") =>
       throw new AccessControlException("Excluding Authz security rules is not allowed")
     case SetCommand(Some((k, Some(_)))) if bannedList.contains(k) =>
-      throw new AccessControlException(s"Modifying $k is not allowed")
+      throw new AccessControlException(s"Modifying config $k is not allowed")
     case _ =>
   }
 }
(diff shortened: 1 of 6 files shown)


cfmcgrady:
This PR aims to restrict configurations from a set statement([Via SET Statements](https://kyuubi.apache.org/docs/r1.6.0-incubating/deployment/settings.html#via-set-statements)).
We need to also handle:

1). modify configurations in the scala execution mode. 

```
spark.sessionState.conf.setConfString(OPTIMIZER_EXCLUDED_RULES.key, somevalues)
```
2). set a configuration from the connection URL. [Via JDBC Connection URL](https://kyuubi.apache.org/docs/r1.6.0-incubating/deployment/settings.html#id2)

cfmcgrady:
@yaooqinn @Yikf @bowenliang123 @lordk911

cfmcgrady:
@Yikf Spark does not provide any public API to change the default behavior, preventing the optimizer phase rule be excluded by the conf `spark.sql.optimizer.excludedRules`, even though I agree that intercepting the set config at optimizer runtime is a better approach.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.spark.sql.execution.command.SetCommand
+
+import org.apache.kyuubi.plugin.spark.authz.AccessControlException
+
+/**
+ * For banning end-users from set restricted spark configurations
+ */
+case class AuthzConfigurationCheckExtension(spark: SparkSession) extends (LogicalPlan => Unit) {
+
+  final val RESTRICT_LIST_KEY = "spark.sql.kyuubi.conf.restricted.list"
```
yaooqinn, line 31: remove `sql`?

yaooqinn:
also update the doc? cc @bowenliang123

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  private val bannedList: Seq[String] =
```
bowenliang123, line 33: `bannedList` -> `restrictedConfList` ? to unify names for same concept.

And maybe better to use Set type rather than Seq.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
```diff
@@ -0,0 +1,48 @@
...
+      "spark.sql.extensions" ::
+      spark.conf.getOption(RESTRICT_LIST_KEY).map(_.split(',').toList).getOrElse(Nil)
+
+  override def apply(plan: LogicalPlan): Unit = plan match {
+    case SetCommand(Some((
+          "spark.sql.optimizer.excludedRules" | "spark.sql.adaptive.optimizer.excludedRules",
+          Some(v)))) if v.contains("org.apache.kyuubi.plugin.spark.authz.ranger") =>
+      throw new AccessControlException("Excluding Authz security rules is not allowed")
+    case SetCommand(Some((k, Some(_)))) if bannedList.contains(k) =>
+      throw new AccessControlException(s"Modifying $k is not allowed")
```
bowenliang123, line 45:
```suggestion
      throw new AccessControlException(s"Modifying config $k is not allowed")
```

yaooqinn:
explain for what, sorry.

yaooqinn:
I don't quite get the point after checking these comments. I can explain a bit about my idea, which may not the best solution anyway.

So,

1. What is the problem?
    - Configurations set by end-users may harm the whole system while the admin wants to ban them ahead.

2. Which way to set and what is the influence?

    - Configs though JDBC string
      - will change engine defaults if no live engine yet. End-users can do anything they want if we do not have a restricted list.
      - will change session defaults if the engine exists. End-users can change non-static SQL configurations only if we do not have a restricted list.
    - Configs though SET cmd. End-users can change non-static SQL configurations only if we do not have a restricted list.

3. So introduce ban lists?

  - For 2.1 and 2.2, we have kyuubi.session.conf.ignore.list and kyuubi.session.conf.restrict.list at serverside for banning session+ level changes.
  - For 2.3, like what I did in the submarine project at engineside for banning operation level changes.

4. For scala or any programming lang support in the feature?
  - If you let users run such code or program, it means that you give up sys security. otherwise, you shall ban it.
  - may we need another more complete ACL management framework for configs.

thanks.

Kent

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  private val bannedList: Seq[String] =
+    RESTRICT_LIST_KEY ::
+      "spark.sql.runSQLOnFiles" ::
+      "spark.sql.extensions" ::
+      spark.conf.getOption(RESTRICT_LIST_KEY).map(_.split(',').toList).getOrElse(Nil)
+
+  override def apply(plan: LogicalPlan): Unit = plan match {
+    case SetCommand(Some((
+          "spark.sql.optimizer.excludedRules" | "spark.sql.adaptive.optimizer.excludedRules",
```
yaooqinn, line 41: spark.sql.adaptive.optimizer.excludedRules seems not necessary.

cfmcgrady:
> also update the doc? cc @bowenliang123

截屏2022-09-30 上午11 55 47

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzConfigurationCheckExtension.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.execution.command.SetCommand
+
+import org.apache.kyuubi.plugin.spark.authz.AccessControlException
+
+/**
+ * For banning end-users from set restricted spark configurations
+ */
+case class AuthzConfigurationCheckExtension(spark: SparkSession) extends (LogicalPlan => Unit) {
```
yaooqinn, line 29: ConfigurationChecker

yaooqinn (approved):

cfmcgrady:
thanks all.

PR #3579, opened 2022-09-29
Use strict path in dist script to avoid unexpected cp behavior

pan3793:

Get different results when running `cp -r jars/ target` on macOS and Linux.

image
image


cxzl25 (approved):

pan3793:
Thanks, merging to master/1.6

Issue #3385, opened 2022-08-31
[Improvement] executor pod name has characters limited in spark on k8s


lianneli:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Executor pod name has characters [limited](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) in spark on k8s, but default executor pod name  with customed user name in Kyuubi may exceed the length. 

### How should we improve?

It will be better if Kyuubi can cut the name or set spark.kubernetes.executor.podNamePrefix automatically.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #3590, opened 2022-10-03
[KYUUBI #3385] Set spark.kubernetes.executor.podNamePrefix if missing

hanna-liashchuk:
In Spark, if Kubernetes executor pod is not conforming to Kubernetes naming conventions, a user gets the next error:
```
org.apache.kyuubi.KyuubiSQLException: org.apache.kyuubi.KyuubiSQLException: java.lang.IllegalArgumentException: 'kyuubi-connection-spark-sql-USER-8fc2847d-b65d-4a46-bb41-4a86385cb61c-988995839f29e15f' in spark.kubernetes.executor.podNamePrefix is invalid.
```
It can be fixed by manually setting 'spark.kubernetes.executor.podNamePrefix' in configurations, but since you cannot dynamically set a value for every user to contain their username, it will miss this essential information that helps to distinguish executors one from another. I suggest Kyuubi server adds this config automatically if it's not coming from user. 


hanna-liashchuk:
Hi @ulysses-you @yaooqinn,
could you please take a look at this and let me know wdyt? thanks

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -160,7 +160,9 @@ object SparkSQLEngine extends Logging {
       "20")
 
     val appName = s"kyuubi_${user}_spark_${Instant.now}"
+    val podNamePrefix = s"kyuubi_${user}_spark_sql"
```
ulysses-you, line 163: this name is not unique, I'm afraid user can not find one application if run some

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -160,7 +160,9 @@ object SparkSQLEngine extends Logging {
...
     _sparkConf.setIfMissing("spark.app.name", appName)
+    _sparkConf.setIfMissing("spark.kubernetes.executor.podNamePrefix", podNamePrefix)
```
zwangsheng, line 165: In kubernetes, same pod name is not supported. This may cause same user's another application can't scala up executor pod. How about add some uuid?

hanna-liashchuk:
hi @zwangsheng @ulysses-you, thanks for your input, indeed if there is more than one connection from a user - conflict would occur. I've added the Instant part and shortened the first part just to meet the limitations on the length. Leaving it as is plus adding a UUID would produce the same error as reported here. WDYT? Is Instant.now().toEpochMilli() a safe choice?

hanna-liashchuk:
btw, I tried setting spark.kubernetes.executor.podNamePrefix to something simple like 'kyuubi-sql' and connect as different users and there were no conflicts. Spark just increases the suffix '-exec-X' for every new pod. I.g. kyuubi-sql-exec-1 ... kyuubi-sql-exec-30 (if every user has 10 executors)

zwangsheng:
> btw, I tried setting spark.kubernetes.executor.podNamePrefix to something simple like 'kyuubi-sql' and connect as different users and there were no conflicts. Spark just increases the suffix '-exec-X' for every new pod. I.g. kyuubi-sql-exec-1 ... kyuubi-sql-exec-30 (if every user has 10 executors)

Hi @hanna-liashchuk , sry for reply late. Not quite sure what you mean. Let me explain a little, according to the purpose of this PR, we want Kyuubi to help each spark sql engine to set up a Kubernetes-Compliant executor pod name prefix, so that for a spark sql engine, whatever user comes to connect afterwards will be the name of the user created at the beginning, for example If the spark user creates the engine, then the executor pod prefix should be `kyuubi-spark-sql-XXX-exec-XXX`.

hanna-liashchuk:
@zwangsheng Sorry, my latter comment was confusing. Indeed, the goal as you said. I meant that if 'spark.kubernetes.executor.podNamePrefix' doesn't have a distinct value, it wouldn't cause any errors or unexpected behavior as Spark will just increment the exec number. 
I picked kyuubi-${user}-${Instant.now().toEpochMilli} cause in most cases it would be distinct and still conform to the Kubernetes naming conventions, adding UUID would cause the lengths of the pod name to be longer than allowed.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -159,7 +159,9 @@ object SparkSQLEngine extends Logging {
       "20")
 
     val appName = s"kyuubi_${user}_spark_${Instant.now}"
+    val podNamePrefix = s"kyuubi-${user}-${Instant.now().toEpochMilli}"
```
zwangsheng, line 162: It's fine to me using `Instant.now().toEpochMilli` for unique. 

BTW we should move this into kubernetes case (Using `Utils.isOnK8s` help determine)

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -159,6 +159,10 @@ object SparkSQLEngine extends Logging {
       "20")
 
     val appName = s"kyuubi_${user}_spark_${Instant.now}"
+    if (Utils.isOnK8s) {
```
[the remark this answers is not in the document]
hanna-liashchuk: Wouldn't hurt :) done

zwangsheng (approved):
LGTM, also cc @pan3793

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
pan3793: The epochMilli LGTM, but still has confliction potential, and we have a better solution for that, see https://github.com/apache/incubator-kyuubi/issues/3345

pan3793: To be clear, either way is acceptable to me in this period.

pan3793: wait, why not use engineRefId here? it was passed by `spark.kyuubi.ha.engine.ref.id` to engine

hanna-liashchuk: `spark.kyuubi.ha.engine.ref.id `is pretty long and adding it all up with an average username would result in the same error as reported here.

hanna-liashchuk: from the documentation: 

> The prefix will be used to generate executor pod names in the form of podNamePrefix−exec−id, where the `id` is a positive int value, so the length of the `podNamePrefix` needs to be less than or equal to 47(= 63 - 10 - 6). 

and 

> You may use spark.kubernetes.executor.podNamePrefix to fully control the executor pod names. When this property is set, it’s highly recommended to make it unique across all jobs in the same namespace.

pan3793 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
pan3793: Make sense, thanks.

pan3793:
Thanks, merging to master

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -168,6 +168,8 @@ object SparkSQLEngine extends Logging {
 
     if (Utils.isOnK8s) {
       kyuubiConf.setIfMissing(FRONTEND_CONNECTION_URL_USE_HOSTNAME, false)
+      val podNamePrefix = s"kyuubi-${user}-${Instant.now().toEpochMilli}"
```
pan3793, line 171: @hanna-liashchuk, it introduces a regression if `user` contains char which not allow in https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

```
java.lang.IllegalArgumentException: 'kyuubi-bdms_pancheng-1667361024618' in spark.kubernetes.executor.podNamePrefix is invalid. must conform https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names and the value length <= 237
	at org.apache.spark.internal.config.TypedConfigBuilder.$anonfun$checkValue$1(ConfigBuilder.scala:108) ~[spark-core_2.12-3.3.1.1.jar:3.3.1.1]
	at org.apache.spark.internal.config.TypedConfigBuilder.$anonfun$transform$1(ConfigBuilder.scala:101) ~[spark-core_2.12-3.3.1.1.jar:3.3.1.1]
	at scala.Option.map(Option.scala:230) ~[scala-library-2.12.15.jar:?]
	at org.apache.spark.internal.config.OptionalConfigEntry.readFrom(ConfigEntry.scala:239) ~[spark-core_2.12-3.3.1.1.jar:3.3.1.1]
	at org.apache.spark.internal.config.OptionalConfigEntry.readFrom(ConfigEntry.scala:214) ~[spark-core_2.12-3.3.1.1.jar:3.3.1.1]
	at org.apache.spark.SparkConf.get(SparkConf.scala:261) ~[spark-core_2.12-3.3.1.1.jar:3.3.1.1]
	at org.apache.spark.deploy.k8s.KubernetesConf.get(KubernetesConf.scala:70) ~[spark-kubernetes_2.12-3.3.1.1.jar:3.3.1.1]
	at org.apache.spark.deploy.k8s.KubernetesExecutorConf.<init>(KubernetesConf.scala:156) ~[spark-kubernetes_2.12-3.3.1.1.jar:3.3.1.1]
	at org.apache.spark.deploy.k8s.KubernetesConf$.createExecutorConf(KubernetesConf.scala:246) ~[spark-kubernetes_2.12-3.3.1.1.jar:3.3.1.1]
	at org.apache.spark.scheduler.cluster.k8s.ExecutorPodsAllocator.$anonfun$requestNewExecutors$1(ExecutorPodsAllocator.scala:394) ~[spark-kubernetes_2.12-3.3.1.1.jar:3.3.1.1]
	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:158) ~[scala-library-2.12.15.jar:?]
	at org.apache.spark.scheduler.cluster.k8s.ExecutorPodsAllocator.requestNewExecutors(ExecutorPodsAllocator.scala:387) ~[spark-kubernetes_2.12-3.3.1.1.jar:3.3.1.1]
```

pan3793: And it's invalid if the `user` is extremely long, we need to
1. replace invalid chars w/ "-", and strip if the length exceeds the threshold
2. add negative tests in K8s it modules

based on the regression, I'm going to revert this PR first, and please re-submit if you have time.

pan3793:
Reverted

- master https://github.com/apache/incubator-kyuubi/commit/223e54e9b32a422a6e486108cc4e463de4a5c17c
- branch-1.6 https://github.com/apache/incubator-kyuubi/commit/36121579e212abd41ecbbbb5aee62805488d94b8

cc @cxzl25, sorry I think this is a blocking issue because it affects users who use Kyuubi w/ Spark on K8s as long as the user name contains char not allowed in https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names, e.g. "_"

Issue #2266, opened 2022-04-01
[Bug] kyuubi 1.5.0 version ,beeline connection fails in kerberos environment


jiaoqingbo:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

kyuubi 1.5.0 version beeline connection fails in kerberos environment，
Using the same configuration, the 1.4.1 version of beeline connected successfully

### Affects Version(s)

1.5.0

### Kyuubi Server Log Output

```logtalk
15:35:55.085 ERROR org.apache.thrift.server.TThreadPoolServer: Error occurred during processing of message.
java.lang.RuntimeException: org.apache.thrift.transport.TTransportException: Peer indicated failure: GSS initiate failed
	at org.apache.thrift.transport.TSaslServerTransport$Factory.getTransport(TSaslServerTransport.java:219) ~[libthrift-0.9.3.jar:0.9.3]
	at org.apache.kyuubi.service.authentication.HadoopThriftAuthBridgeServer$TUGIAssumingTransportFactory$$anon$4.run(HadoopThriftAuthBridgeServer.scala:117) ~[kyuubi-common_2.12-1.5.0-incubating.jar:1.5.0-incubating]
	at org.apache.kyuubi.service.authentication.HadoopThriftAuthBridgeServer$TUGIAssumingTransportFactory$$anon$4.run(HadoopThriftAuthBridgeServer.scala:116) ~[kyuubi-common_2.12-1.5.0-incubating.jar:1.5.0-incubating]
	at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_271]
	at javax.security.auth.Subject.doAs(Subject.java:360) ~[?:1.8.0_271]
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1855) ~[hadoop-client-api-3.3.2.jar:?]
	at org.apache.kyuubi.service.authentication.HadoopThriftAuthBridgeServer$TUGIAssumingTransportFactory.getTransport(HadoopThriftAuthBridgeServer.scala:116) ~[kyuubi-common_2.12-1.5.0-incubating.jar:1.5.0-incubating]
[...]

PR #3597, opened 2022-10-09
Engine should prefer ip for registering on K8s cluster mode

### LINKED ISSUE #3578: [Improvement] Supply a function to detect Kyuubi is running in K8s Pod
### Code of Conduct

### Search before asking

As lots of users reported the default configuration does not work on K8s cluster mode, to improve the user experience, let's start from improve the error message if we detect Kyuubi is running in K8s Pod

### How should we improve?

_No response_

pan3793:

When Kyuubi runs outside of K8s, and w/o enhanced DNS infrastructure, Kyuubi can not access the Pod by using the hostname of Pod, it blocks the user to run Spark on K8s w/ cluster mode out-of-box. 

Kyuubi provided a configuration `kyuubi.frontend.connection.url.use.hostname`, turn it off could address this issue, but we can not change the default value globally because of https://github.com/apache/incubator-kyuubi/issues/2266

To improve user experience, we can detect if the Driver is running inside the Pod, and if yes, 
change `kyuubi.frontend.connection.url.use.hostname` default value to `false`.

Close #3578

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -164,8 +164,12 @@ object SparkSQLEngine extends Logging {
     val defaultCat = if (KyuubiSparkUtil.hiveClassesArePresent) "hive" else "in-memory"
     _sparkConf.setIfMissing("spark.sql.catalogImplementation", defaultCat)
 
-    kyuubiConf.setIfMissing(KyuubiConf.FRONTEND_THRIFT_BINARY_BIND_PORT, 0)
+    kyuubiConf.setIfMissing(FRONTEND_THRIFT_BINARY_BIND_PORT, 0)
     kyuubiConf.setIfMissing(HA_ZK_CONN_RETRY_POLICY, RetryPolicies.N_TIME.toString)
+
+    if (Utils.isOnK8s) {
+      kyuubiConf.setIfMissing(FRONTEND_CONNECTION_URL_USE_HOSTNAME, false)
+    }
 
     // Pass kyuubi config from spark with `spark.kyuubi`
     val sparkToKyuubiPrefix = "spark.kyuubi."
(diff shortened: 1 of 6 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -329,4 +329,6 @@ object Utils extends Logging {
    */
   def getContextOrKyuubiClassLoader: ClassLoader =
     Option(Thread.currentThread().getContextClassLoader).getOrElse(getKyuubiClassLoader)
+
+  def isOnK8s: Boolean = Files.exists(Paths.get("/var/run/secrets/kubernetes.io"))
```
pan3793, line 333: the solution is found at https://stackoverflow.com/a/49045575, and got confirmed by our internal K8s team. but I don't if there is an official recommended way to do that.

pan3793:
cc @zwangsheng @jiaoqingbo @cxzl25

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -164,9 +164,13 @@ object SparkSQLEngine extends Logging {
     val defaultCat = if (KyuubiSparkUtil.hiveClassesArePresent) "hive" else "in-memory"
     _sparkConf.setIfMissing("spark.sql.catalogImplementation", defaultCat)
 
-    kyuubiConf.setIfMissing(KyuubiConf.FRONTEND_THRIFT_BINARY_BIND_PORT, 0)
+    kyuubiConf.setIfMissing(FRONTEND_THRIFT_BINARY_BIND_PORT, 0)
     kyuubiConf.setIfMissing(HA_ZK_CONN_RETRY_POLICY, RetryPolicies.N_TIME.toString)
 
+    if (Utils.isOnK8s) {
+      kyuubiConf.setIfMissing(FRONTEND_CONNECTION_URL_USE_HOSTNAME, false)
```
cfmcgrady, line 171: Maybe we should point it out in the documentation?

pan3793: make sense, updated

cxzl25 (approved):

jiaoqingbo (approved):
LGTM

zwangsheng (approved):

pan3793:
> What happen if currently env is k8s cluster mode w/ kerberos?

#2266 occurs on the server side, this change only affects the engine side. that is why I say we can not change the default value GLOBALLY.

pan3793:
Thanks, merging to master/1.6

PR #3600, opened 2022-10-10
[SPARK][K8S] Respect default static port for web UI on Spark K8s cluster mode

zwangsheng:

Now kyuubi will help set web ui port to 0 while scalaing up spark sql engine, which is good for the general situation, help avoid port conflicts.

But on k8s case, kyuubi should respect the user's default settings, here means web ui should be `4040`.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -168,9 +168,11 @@ object SparkSQLEngine extends Logging {
 
     if (Utils.isOnK8s) {
       kyuubiConf.setIfMissing(FRONTEND_CONNECTION_URL_USE_HOSTNAME, false)
-    } else {
-      _sparkConf.setIfMissing("spark.ui.port", "0")
+      sys.env.get("SPARK_APPLICATION_ID").foreach(_ =>
+        _sparkConf.setIfMissing("spark.ui.port", "4040"))
     }
+    // set web ui port 0 when no deploy k8s cluster mode
+    _sparkConf.setIfMissing("spark.ui.port", "0")
 
     // Pass kyuubi config from spark with `spark.kyuubi`
     val sparkToKyuubiPrefix = "spark.kyuubi."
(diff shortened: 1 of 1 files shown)


zwangsheng:
cc @pan3793

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -169,6 +168,8 @@ object SparkSQLEngine extends Logging {
 
     if (Utils.isOnK8s) {
       kyuubiConf.setIfMissing(FRONTEND_CONNECTION_URL_USE_HOSTNAME, false)
+    } else {
+      _sparkConf.setIfMissing("spark.ui.port", "0")
```
pan3793, line 172: this may cause regression if the user deploys Kyuubi on K8s and run Spark in Client mode, but I guess this is rare and acceptable? WDYT @iodone @yaooqinn

pan3793: let's narrow the condition to "!k8s_cluster_mode"

zwangsheng:
On kubernetes, spark driver pod will start with `--deploy-mode client`, we can't use `spark.submit.deployMode` to find whether driver is in it's pod or in kyuubi's pod.
Every spark driver pod created with this env ' SPARK_APPLICATION_ID', so we would like using `SPARK_APPLICATION_ID` to figure out deploy mode on kubernetes.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -171,6 +170,11 @@ object SparkSQLEngine extends Logging {
       kyuubiConf.setIfMissing(FRONTEND_CONNECTION_URL_USE_HOSTNAME, false)
     }
 
+    // set web ui port 0 when no deploy k8s cluster mode
```
pan3793, line 173: The comment is redundant since the code is clear enough to explain what we are doing here, the really helpful comments should explain WHY we do that.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -309,4 +313,7 @@ object SparkSQLEngine extends Logging {
       },
       "CreateSparkTimeoutChecker").start()
   }
+
+  private def isOnKubernetesUsingClusterMode: Boolean =
```
pan3793, line 317: isOnK8sClusterMode

pan3793 (approved):

pan3793:
Thanks, merging to master/1.6

PR #3601, opened 2022-10-10
[SPARK] Support infer columns for rebalance and sort

ulysses-you:

Improve the rebalance before writing rule.

The rebalance before writing rule adds a rebalance at the top of query for data writing command, however the default partitioning of rebalance uses RoundRobinPartitioning which would break the original partitioning of data. It may cause the the output data size bigger than before.

This pr supports infer the columns from join and aggregate for rebalance and sort to improve the compression ratio.

Note that, this improvement only works for static partition writing.


ulysses-you:
cc @yaooqinn @cfmcgrady

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/InferRebalanceAndSortOrders.scala
```diff
@@ -0,0 +1,109 @@
...
+        case s: Sort => Some((s.order.map(_.child), s.order.map(_.child)))
+        case p: Project =>
+          val aliasMap = getAliasMap(p.projectList)
+          candidateKeys(p.child, p.references).map { case (partitioning, ordering) =>
+            (
+              partitioning.map(p => aliasMap.getOrElse(p.canonicalized, p)),
+              ordering.map(o => aliasMap.getOrElse(o.canonicalized, o)))
+          }
+        case f: Filter => candidateKeys(f.child, output)
+        case s: SubqueryAlias => candidateKeys(s.child, output)
```
cfmcgrady, line 97: Pass through the `View` node.
```scala
        case v: View => candidateKeys(v.child, output)
```

ulysses-you: addressed

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -160,4 +160,20 @@ object KyuubiSQLConf {
       .version("1.5.0")
       .booleanConf
       .createWithDefault(false)
+
+  val INFER_REBALANCE_AND_SORT_ORDERS =
```
yaooqinn, line 164: let's expose these to docs

yaooqinn (approved):
LGTM, expect one minor comment

cfmcgrady (approved):

ulysses-you:

PR #3604, opened 2022-10-10
Propagate HADOOP_PROXY_USER on launching Flink engine

dongkelun:

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
@@ -137,6 +137,9 @@ trait ProcBuilder {
 
     val envs = pb.environment()
     envs.putAll(env.asJava)
+    if (shortName.equals("flink")) {
+      envs.put("HADOOP_PROXY_USER", proxyUser)
+    }
     pb.directory(workingDir.toFile)
     pb.redirectError(engineLog)
     pb.redirectOutput(engineLog)
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
@@ -137,9 +137,6 @@ trait ProcBuilder {
 
     val envs = pb.environment()
     envs.putAll(env.asJava)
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala
```diff
@@ -137,6 +137,9 @@ trait ProcBuilder {
 
     val envs = pb.environment()
     envs.putAll(env.asJava)
+    if (shortName.equals("flink")) {
```
yaooqinn, line 140: we can move this to flink-proc-builder？

dongkelun: There is childProcEnv in the history, but it is removed from this PR [https://github.com/apache/incubator-kyuubi/pull/2572](https://github.com/apache/incubator-kyuubi/pull/2572). Can I add it again?

yaooqinn: maybe we can put it into `pb.environment()` directly

dongkelun: Would it be better to move HADOOP_PROXY_USER to conf of FlinkProcessBuilder?

yaooqinn (approved):

pan3793:
cc @yanghua @SteNicholas, do you think it's a bug fix or new feature? should we merge it into branch-1.6?

pan3793 (approved):

yanghua (approved):
+1

pan3793:
Thanks, merging to master

pan3793:
@dongkelun the `HADOOP_PROXY_USER` has side-effects and causes #4762 on non-kerberized env, any thoughts?

pan3793:
@dongkelun have done some investigation, seems Flink does not support proxy user, so there is no ideal solution IMO now.

I will simply revert this patch from branch-1.7(master already removed it in b315123a6b6dfa7b03a5ab7875856bdfd4e0eaed), as a workaround, users can pass the env vars to the engine by `kyuubi.engineEnv.HADOOP_PROXY_USER=<>` to achieve this capability

PR #3619, opened 2022-10-12
Enhance JAVA_HOME detection to support RPM

pan3793:

By default, on CentOS 7 w/ openJDK 8 installed via YUM/RPM, the `/usr/bin/java` links to `/usr/lib/jvm/java-1.8.0-openjdk/jre/bin/java`, the current detection logic will think the `JAVA_HOME` is `/usr/lib/jvm/java-1.8.0-openjdk/jre`, and then cause the issue.
 
```
[ERROR] Failed to execute goal net.alchim31.maven:scala-maven-plugin:4.6.1:compile (scala-compile-first) on project kyuubi-common_2.12: wrap: java.io.IOException: Cannot run program "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.322.b06-1.el7_9.x86_64/jre/bin/javac" (in directory "/home/pancheng/apache-kyuubi"): error=2, No such file or directory -> [Help 1]
```

w/ this patch, the `build/dist` works expected on CentOS w/ openJDK 8 installed via `yum`, the debug info shows it detected the right `JAVA_HOME`
```
+ [[ -z '' ]]
++ command -v rpm
+ '[' /usr/bin/rpm ']'
++ rpm -E %java_home
+ RPM_JAVA_HOME=/usr/lib/jvm/java
+ '[' /usr/lib/jvm/java '!=' %java_home ']'
+ JAVA_HOME=/usr/lib/jvm/java
+ echo 'No JAVA_HOME set, proceeding with '\''/usr/lib/jvm/java'\'' learned from rpm'
No JAVA_HOME set, proceeding with '/usr/lib/jvm/java' learned from rpm
+ '[' -z /usr/lib/jvm/java ']'
+ [[ -z /usr/lib/jvm/java ]]
+ echo 'JAVA_HOME is set to /usr/lib/jvm/java'
JAVA_HOME is set to /usr/lib/jvm/java
```

### DIFF
--- a/build/dist
+++ b/build/dist
@@ -103,6 +103,15 @@ while (( "$#" )); do
 done
 # Setup java
 if [[ -z "$JAVA_HOME" ]]; then
+  # Fallback on JAVA_HOME from rpm, if found
+  if [ $(command -v rpm) ]; then
+    RPM_JAVA_HOME="$(rpm -E %java_home 2>/dev/null)"
+    if [ "$RPM_JAVA_HOME" != "%java_home" ]; then
+      JAVA_HOME="$RPM_JAVA_HOME"
+      echo "No JAVA_HOME set, proceeding with '$JAVA_HOME' learned from rpm"
+    fi
+  fi
+
   if [[ `command -v java` ]]; then
     # If java is in /usr/bin/java, we want /usr
     JAVA_HOME="$(dirname $(dirname $(which java)))"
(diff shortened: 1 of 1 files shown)


build/dist
```diff
@@ -114,6 +126,7 @@ if [[ -z "$JAVA_HOME" ]]; then
   exit -1
 fi
 
+export JAVA_HOME="$JAVA_HOME"
```
pan3793, line 129: we should export it, otherwise mvn can not see it.

cxzl25 (approved):

pan3793:
Thanks, merging to master/1.6

PR #3658, opened 2022-10-17
Support SSL for Kyuubi thrift binary connection

turboFei:

Support SSL for kyuubi thrift binary conenction.

Will add screenshot for IT.


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -379,6 +379,57 @@ object KyuubiConf {
...
+  val FRONTEND_SSL_PROTOCOL_EXCLUDES: ConfigEntry[Seq[String]] =
+    buildConf("kyuubi.frontend.ssl.protocol.excludes")
+      .doc("SSL Versions to disable.")
+      .version("1.7.0")
+      .stringConf
+      .toSequence()
+      .createWithDefault(Seq("SSLv2", "SSLv3"))
+
+  val FRONTEND_THRIFT_BINARY_SSL_INCLUDE_CIPHER_SUITES: ConfigEntry[Seq[String]] =
+    buildConf("kyuubi.frontend.thrift.binary.ssl.include.ciphersuites")
```
pan3793, line 426: I would suggestion one of the following name style.
```
kyuubi.frontend.ssl.exclude.protocols
kyuubi.frontend.thrift.binary.ssl.include.ciphersuites
```
```
kyuubi.frontend.ssl.protocols.excluded
kyuubi.frontend.thrift.binary.ssl.ciphersuites.included
```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -584,22 +635,34 @@ object KyuubiConf {
...
+      .withAlternative("kyuubi.frontend.ssl.keystore.password")
       .stringConf
       .createOptional
 
-  val FRONTEND_THRIFT_HTTP_SSL_PROTOCOL_BLACKLIST: ConfigEntry[String] =
+  val FRONTEND_THRIFT_HTTP_SSL_PROTOCOL_BLACKLIST: ConfigEntry[Seq[String]] =
     buildConf("kyuubi.frontend.thrift.http.ssl.protocol.blacklist")
       .doc("SSL Versions to disable when using HTTP transport mode.")
       .version("1.6.0")
+      .withAlternative("kyuubi.frontend.ssl.protocol.excludes")
```
pan3793, line 654: kyuubi.frontend.ssl.exclude.protocols

pan3793 (approved):
LGTM, only one issue

docs/deployment/settings.md
```diff
@@ -296,9 +296,16 @@ kyuubi.frontend.protocols|THRIFT_BINARY|A comma separated list for all frontend
 kyuubi.frontend.proxy.http.client.ip.header|X-Real-IP|The http header to record the real client ip address. If your server is behind a load balancer or other proxy, the server will see this load balancer or proxy IP address as the client IP address, to get around this common issue, most load balancers or proxies offer the ability to record the real remote IP address in an HTTP header that will be added to the request for other devices to use. Note that, because the header value can be specified to any ip address, so it will not be used for authentication.|string|1.6.0
 kyuubi.frontend.rest.bind.host|&lt;undefined&gt;|Hostname or IP of the machine on which to run the REST frontend service.|string|1.4.0
 kyuubi.frontend.rest.bind.port|10099|Port of the machine on which to run the REST frontend service.|int|1.4.0
+kyuubi.frontend.ssl.exclude.protocols|SSLv2,SSLv3|SSL Versions to disable.|seq|1.7.0
```
yaooqinn, line 299: irrelevant with kyuubi.frontend.thrift.http.ssl.protocol.blacklist？

turboFei:
```
  val FRONTEND_THRIFT_HTTP_SSL_PROTOCOL_BLACKLIST: ConfigEntry[Seq[String]] =
    buildConf("kyuubi.frontend.thrift.http.ssl.protocol.blacklist")
      .doc("SSL Versions to disable when using HTTP transport mode.")
      .version("1.6.0")
      .withAlternative("kyuubi.frontend.ssl.exclude.protocols")
      .stringConf
      .createWithDefault("SSLv2,SSLv3")
      .toSequence()
      .createWithDefault(Seq("SSLv2", "SSLv3"))
```
```
      .withAlternative("kyuubi.frontend.ssl.exclude.protocols")
```

turboFei: shall we add a alternative keys column in settings.md？

turboFei: `kyuubi.frontend.thrift.http.ssl.protocol.blacklist` is a legacy parameter added before.

and  pan suggested that using `exclude/include` instead of `blacklist/whitelist`.

yaooqinn: `exclude` is not specific to me. it's too ambiguous, is it `not supported(you shall not use it)`/`disabled(you can turn it on if you want)`/`not needed` or something else?

pan3793: How about
```
kyuubi.frontend.ssl.protocols.denied
kyuubi.frontend.ssl.protocols.allowed
```

yaooqinn: these sound like boolean parameters as they all end with a verb.

yaooqinn: does kyuubi.frontend.ssl mean it works for other frontend protocals?

yaooqinn: another thing confuses me and maybe confuse other users is the word protocol here, it can point to kyuubi frontend protocol or SSL version

yaooqinn, line 299: kyuubi.frontend(.binary).ssl.disallowed(disabled)Protocols, the doc part can be either removed as it says nothing more than the config name, or we can elaborate it a bit more.

turboFei: > does kyuubi.frontend.ssl mean it works for other frontend protocals?

Yes, the ssl parameters should be common for all the protocols

yaooqinn (approved):

turboFei:
thanks, merging to master

PR #3659, opened 2022-10-18
Support alternative keys in ConfigBuilder

turboFei:

Refer Spark PR: https://github.com/apache/spark/pull/18110
ConfigBuilder builds ConfigEntry which can only read value with one key, if we wanna change the config name but still keep the old one, it's hard to do.

This PR introduce ConfigBuilder.withAlternative, to support reading config value with alternative keys.


pan3793 (approved):

turboFei:
thanks, merged to master

PR #3670, opened 2022-10-20
[SPARK][K8S] Fix Spark master match rule on tagging Spark application

zwangsheng:

Fix kyuubi application manager tag spark kubernetes application with yarn tag.

The reason for this is the need to determine whether resourceManager start with K8S instead of being equal to K8S.

Remove duplicate code from spark batch process builder.

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KyuubiApplicationManagerSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KyuubiApplicationManagerSuite.scala
@@ -73,9 +73,13 @@ class KyuubiApplicationManagerSuite extends KyuubiFunSuite {
   }
 
   test("Test kyuubi application Manager tag spark on kubernetes application") {
-    val conf: KyuubiConf = KyuubiConf().set("", "")
+    val conf: KyuubiConf = KyuubiConf()
     val tag = "kyuubi-test-tag"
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KyuubiApplicationManagerSuite.scala
```diff
@@ -70,4 +71,15 @@ class KyuubiApplicationManagerSuite extends KyuubiFunSuite {
         localDirLimitConf)
     }
   }
+
+  test("Test kyuubi application Manager tag spark on kubernetes application") {
+    val conf: KyuubiConf = KyuubiConf().set("", "")
```
pan3793, line 76: why `.set("", "")`?

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KyuubiApplicationManagerSuite.scala
```diff
@@ -70,4 +71,15 @@ class KyuubiApplicationManagerSuite extends KyuubiFunSuite {
...
+    val tag = "kyuubi-test-tag"
+    KyuubiApplicationManager.tagApplication(tag, "SPARK", Some("k8s://kyuubi-test:8443"), conf)
```
pan3793, line 78: the correct pattern should be `k8s://https://xxx`?

pan3793 (approved):

pan3793:
Thanks, merging to master/1.6

PR #3684, opened 2022-10-24
Expose the engine id when registering and show the engine details for list engine command

turboFei:

Expose the engine id when registering the service. 

So that we can get more details about the engine when using `list engine` command.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
@@ -93,7 +93,7 @@ class SparkTBinaryFrontendService(
     }
   }
 
-  override def extraServiceInfo: Map[String, String] = {
+  override def attributes: Map[String, String] = {
     Map(KYUUBI_ENGINE_ID -> KyuubiSparkUtil.engineId)
   }
 }
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
@@ -92,6 +92,10 @@ class SparkTBinaryFrontendService(
       None
     }
   }
+
+  override def extraServiceInfo: Map[String, String] = {
+    Map(KYUUBI_ENGINE_ID -> KyuubiSparkUtil.engineId)
+  }
 }
 
 object SparkTBinaryFrontendService extends Logging {
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala
@@ -219,8 +219,10 @@ class ZookeeperDiscoveryClient(conf: KyuubiConf) extends DiscoveryClient {
         val (host, port) = DiscoveryClient.parseInstanceHostPort(instance)
         val version = p.split(";").find(_.startsWith("version=")).map(_.stripPrefix("version="))
         val engineRefId = p.split(";").find(_.startsWith("refId=")).map(_.stripPrefix("refId="))
+        val extraInfo = p.split(";").map(_.split("=")).filter(_.size == 2).map(kv =>
+          (kv.head, kv.last)).toMap.filterNot(kv => Set("version", "refId").contains(kv._1))
         info(s"Get service instance:$instance and version:$version under $namespace")
-        ServiceNodeInfo(namespace, p, host, port, version, engineRefId)
+        ServiceNodeInfo(namespace, p, host, port, version, engineRefId, extraInfo)
       }
     } catch {
       case _: Exception if silent => Nil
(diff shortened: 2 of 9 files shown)


turboFei:
```
AdminResourceSuite:
- refresh Hadoop configuration of the kyuubi server
  List("serviceUri=localhost:40419;version=1.7.0-SNAPSHOT;kyuubi.engine.id=local-1666594567035;refId=823f7053-4ec6-4eaf-8521-3dc209393673;sequence=0000000000") had size 1 instead of expected size 0 (AdminResourceSuite.scala:103)
```

yaooqinn:
why a engine id means more information？

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala
```diff
@@ -342,9 +351,12 @@ class ZookeeperDiscoveryClient(conf: KyuubiConf) extends DiscoveryClient {
 
     val session = conf.get(HA_ENGINE_REF_ID)
       .map(refId => s"refId=$refId;").getOrElse("")
+    val extraInfo =
+      extraServiceInfo.map(kv => kv._1 + "=" + kv._2).mkString(";", ";", "").stripSuffix(";")
     val pathPrefix = ZKPaths.makePath(
       namespace,
-      s"serviceUri=$instance;version=${version.getOrElse(KYUUBI_VERSION)};${session}sequence=")
```
yaooqinn, line 347: what else will be added to to path?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
```diff
@@ -92,6 +92,10 @@ class SparkTBinaryFrontendService(
       None
     }
   }
+
+  override def extraServiceInfo: Map[String, String] = {
```
yaooqinn, line 96: why is spark special?

turboFei: image

Now I can only see the `KYUUBI_ENGINE_ID` is used in Spark.

We can extend other engines to support expose kyuubi.engine.id.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala
turboFei: now only the kyuubi.engine.id is added.

Maybe we can add more info.
such as kyuubi.engine.url and kyuubi.engine.name.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala
```diff
@@ -342,9 +351,12 @@ class ZookeeperDiscoveryClient(conf: KyuubiConf) extends DiscoveryClient {
...
+      s"serviceUri=$instance;version=${version.getOrElse(KYUUBI_VERSION)}" +
+        s"$extraInfo;${session}sequence=")
```
zwangsheng, line 359: we need `;` here to split version and extraInfo

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
yaooqinn: why extra is added? do we have an info method?

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/Engine.java
```diff
@@ -40,14 +43,16 @@ public Engine(
       String sharelevel,
       String subdomain,
       String instance,
-      String namespace) {
+      String namespace,
+      Map<String, String> details) {
```
yaooqinn, line 47: what does details stand for?

yaooqinn:
2 general questions
 - why do need this?
 - why do we exposing in such a way?

turboFei:
> why do need this?

We now support to list engines with kyuubi-admin command line.
And we need more insight for the engine information.

https://github.com/apache/incubator-kyuubi/pull/3680

> why do we expose it in this way?

turboFei:
> why do we expose it in this way?

It is compatible with before.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/Engine.java
turboFei: all the info saved in the engine discovery node.
version, sequence, engineId and so on.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
turboFei: > why extra is added? do we have an info method?

I did not get it.

turboFei:
cc @yaooqinn

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
pan3793: The extra is to distinguish w/ the service host/port information, if it's confusing, I suggest naming it attributes here.

pan3793 (approved):

turboFei:
thanks for the review, merging to master

PR #3718, opened 2022-10-28
[KYUUBI #3717] Support flink engine see primary keys

waywtdcc:

*Support flink engine see primary keys*


yaooqinn:
cc @yanghua @SteNicholas

waywtdcc:
This may not require regular expressions. I will modify it later.

pan3793:
Yea, schema and table name should not be patterns but the exact name

yanghua:
Will review it later.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala
```diff
@@ -0,0 +1,103 @@
...
+
+      val schemaName =
+        if (StringUtils.isEmpty(schemaNameOrEmpty)) catalog.getDefaultDatabase
+        else schemaNameOrEmpty
+
+      val flinkTable = tableEnv.from(s"`$catalogName`.`$schemaName`.`$tableName`")
+
+      val resolvedSchema = flinkTable.getResolvedSchema
+      val uniqueConstraint = resolvedSchema.getPrimaryKey.orElse(
+        UniqueConstraint.primaryKey("null_pri", Collections.emptyList()))
```
pan3793, line 58: why should we use this default value?

waywtdcc: Thanks for your reviews. I have modified it

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala
```diff
@@ -0,0 +1,104 @@
...
+          .map { case (column, pos) =>
+            toColumnResult(
+              catalogName,
+              schemaName,
+              tableName,
+              uniqueConstraint.getName,
+              column,
+              pos)
+          }
+      }
```
cxzl25, line 70:
```suggestion
      val columns = primaryKeySchema.asScala.map { uniqueConstraint =>
        uniqueConstraint
          .getColumns.asScala.toArray.zipWithIndex
          .map { case (column, pos) =>
            toColumnResult(
              catalogName,
              schemaName,
              tableName,
              uniqueConstraint.getName,
              column,
              pos)
          }
      }.getOrElse(Array.empty)
```

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala
```diff
@@ -0,0 +1,104 @@
...
+      pos: Int): Row = {
+    // format: off
+    Row.of(
+      catalogName, // TABLE_CAT
+      schemaName, // TABLE_SCHEM
+      tableName, // TABLE_NAME
+      columnName, // COLUMN_NAME
+      Integer.valueOf(pos + 1), // KEY_SEQ
+      pkName // PK_NAME
+    )
```
cxzl25, line 100:
```suggestion
    Row.of(
      catalogName,              // TABLE_CAT
      schemaName,               // TABLE_SCHEM
      tableName,                // TABLE_NAME
      columnName,               // COLUMN_NAME
      Integer.valueOf(pos + 1), // KEY_SEQ
      pkName                    // PK_NAME
    )
```

cxzl25 (commented):
Left some comment, LGTM overall.

pan3793:
please rebase master to solve the conflict, and it would be good if you can test w/ DBeaver to verify this functionality.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala
```diff
@@ -0,0 +1,102 @@
...
+      val tableEnv = sessionContext.getExecutionContext.getTableEnvironment
+
+      val catalogName =
+        if (StringUtils.isEmpty(catalogNameOrEmpty)) tableEnv.getCurrentCatalog
+        else catalogNameOrEmpty
+      val catalog = tableEnv.getCatalog(catalogName).get()
+
+      val schemaName =
+        if (StringUtils.isEmpty(schemaNameOrEmpty)) catalog.getDefaultDatabase
+        else schemaNameOrEmpty
```
jiaoqingbo, line 49: If catalogNameOrEmpty is empty, tableEnv.getCurrentCatalog will be selected. In this case, if schemaNameOrEmpty is empty, schemaName should choose tableEnv.getCurrentDatabase instead of catalog.getDefaultDatabase. WDYT @pan3793

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
```diff
@@ -163,6 +163,78 @@ class FlinkOperationSuite extends WithFlinkSQLEngine with HiveJDBCTestHelper {
...
+           | )
+    """.stripMargin)
+
+      val metaData = statement.getConnection.getMetaData
+
+      Seq(tableName1, tableName2, tableName3) foreach { tableName =>
+        val rowSet = metaData.getPrimaryKeys("", "", tableName)
+
+        if (tableName.equals(tableName3)) {
+          assert(rowSet.next() == false)
```
jiaoqingbo, line 220: assert(!rowSet.next())

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala
pan3793: good catch, sgtm, @waywtdcc would you please change as @jiaoqingbo suggested?

waywtdcc: OK, thank you for your review. I have fixed it

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala
```diff
@@ -0,0 +1,101 @@
...
+    } catch onError()
+  }
+
+  private def toColumnResult(
+      catalogName: String,
+      schemaName: String,
+      tableName: String,
+      pkName: String,
+      columnName: String,
+      pos: Int): Row = {
```
yanghua, line 88: minor comment: make the order of args equals to `Row.of(x, y, x)` and `.columns(...)` look more readable.

pan3793 (approved):
Overall LGTM, please address other reviewers' comments

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala
```diff
@@ -0,0 +1,101 @@
...
+    try {
+      val tableEnv = sessionContext.getExecutionContext.getTableEnvironment
+
+      val catalogName =
+        if (StringUtils.isEmpty(catalogNameOrEmpty)) tableEnv.getCurrentCatalog
+        else catalogNameOrEmpty
+
+      val schemaName =
+        if (StringUtils.isEmpty(schemaNameOrEmpty)) tableEnv.getCurrentDatabase
+        else schemaNameOrEmpty
```
jiaoqingbo, line 48: It seems that there is still a problem here. When the catalogName is not equal to tableEnv.getCurrentCatalog, the obtained schemaName is still the CurrentDatabase of the current catalog. Correctly modify the code should be similar to the following way.


```val schemaName =
        if (StringUtils.isEmpty(schemaNameOrEmpty)) {
          if (catalogName != tableEnv.getCurrentCatalog) {
            catalog.getDefaultDatabase
          } else {
            tableEnv.getCurrentDatabase
          }
        } else schemaNameOrEmpty

Comment (waywtdcc) [reply externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala:48]: val schemaName =
        if (StringUtils.isEmpty(schemaNameOrEmpty)) {
          if (tableEnv.getCurrentDatabase != null) {
            tableEnv.getCurrentDatabase
          } else {
            tableEnv.getCatalog(catalogName).get().getDefaultDatabase
          }
        } else schemaNameOrEmpty  Is that OK?

Comment (waywtdcc) [reply externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala]: Thank you for your reviews. I think it's good. I have revised the relevant content.

File: externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala (hunk 6 above)
Comment (waywtdcc) [reply externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala:93]: it is list type.

Comment (yanghua) [reply externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala:93]: > it is list type.

What's the meaning?

Actually, my suggestion is:

```
private def toColumnResult(
      catalogName: String,
      schemaName: String,
      tableName: String,
      columnName: String,
      pos: Int,
      pkName: String): Row = {
```

It's the same order in which you assemble the objects. Moreover, this method is only for assembling objects.

File: externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala (hunk 7 above)
Comment (yanghua) [reply externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala]: Can we add a test for this logic?

File: externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala (hunk 6 above)
Comment (waywtdcc) [reply externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetPrimaryKeys.scala]: OK, thanks for your reviews. I modified the order, please take a look.

File: externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala [hunk 8]
```diff
@@ -163,6 +163,78 @@ class FlinkOperationSuite extends WithFlinkSQLEngine with HiveJDBCTestHelper {
...
+           |  c3 integer
+           | )
+           | with (
+           |   'connector' = 'filesystem'
+           | )
+    """.stripMargin)
+
+      val metaData = statement.getConnection.getMetaData
+
+      Seq(tableName1, tableName2, tableName3) foreach { tableName =>
```
yanghua, line 216: I found the `tableName3` seems redundant? The definition of `tableName1 ` and `tableName3 ` are almost the same(no primary keys). If we want to verify if we can fetch primary keys, we can just use `tableName1`. WDYT?

pan3793: I think the original idea here is define single pk for `tableName1`

```
CONSTRAINT pk_con primary key(id1) NOT ENFORCED
```

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
```diff
@@ -163,6 +163,78 @@ class FlinkOperationSuite extends WithFlinkSQLEngine with HiveJDBCTestHelper {
...
+            assert(rowSet.getString(TABLE_CAT) === "default_catalog")
+            assert(rowSet.getString(TABLE_SCHEM) === "default_database")
+            assert(rowSet.getString(TABLE_NAME) === tableName)
+            assert(rowSet.getString(COLUMN_NAME) === s"id$pos")
+            assert(rowSet.getInt(KEY_SEQ) === pos)
+            assert(rowSet.getString(PK_NAME) === "pk_con")
+            pos += 1
+          }
+        }
+
```
pan3793, line 233: `assert pos > 1`?

yanghua (approved):
+1, LGTM.

pan3793:
Thanks, merging to master

PR #3720, opened 2022-10-29
Support to show remaining logs with timeout if the batch failed

turboFei:

If the batch failed, we shall try to show the remaining logs with timeout.

Existing UT.

### DIFF
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
@@ -45,25 +45,46 @@ class LogBatchCommand(
     withKyuubiRestClient(normalizedCliConfig, restConfigMap, conf) { kyuubiRestClient =>
       val batchRestApi: BatchRestApi = new BatchRestApi(kyuubiRestClient)
       val batchId = normalizedCliConfig.batchOpts.batchId
-      var from = math.max(normalizedCliConfig.batchOpts.from, 0)
       val size = normalizedCliConfig.batchOpts.size
 
+      @volatile
       var log: OperationLog = null
+      @volatile
       var done = false
       var batch = this.batch.getOrElse(batchRestApi.getBatchById(batchId))
       val kyuubiInstance = batch.getKyuubiInstance
 
       withKyuubiInstanceRestClient(kyuubiRestClient, kyuubiInstance) { kyuubiInstanceRestClient =>
         val kyuubiInstanceBatchRestApi: BatchRestApi = new BatchRestApi(kyuubiInstanceRestClient)
+
+        val logThread = new Thread("batch-log-thread") {
+          override def run(): Unit = {
+            var from = math.max(normalizedCliConfig.batchOpts.from, 0)
+            // if it has been done and the last fetched log is empty, break the log loop
+            while (!(done && log != null && log.getRowCount == 0)) {
+              try {
+                log = kyuubiInstanceBatchRestApi.getBatchLocalLog(
+                  batchId,
+                  from,
+                  size)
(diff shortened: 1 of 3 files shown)


turboFei:
cc @lightning-L  @pan3793

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
```diff
@@ -45,43 +45,81 @@ class LogBatchCommand(
...
+                  size)
+                from += log.getLogRowSet.size
+                log.getLogRowSet.asScala.foreach(x => info(x))
+              } catch {
+                case e: Exception => error(s"Error fetching batch logs: ${e.getMessage}")
+              }
+
+              // if it has been done and the last fetched logs is non empty, do not wait a interval
+              if (!done || (log != null && log.getRowCount > 0)) {
+                Thread.sleep(conf.get(CTL_BATCH_LOG_QUERY_INTERVAL))
```
pan3793, line 88: Should InterruptedException be caught here?

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
```diff
@@ -100,7 +138,7 @@ class LogBatchCommand(
     var batch: Batch = null
 
     if (!normalizedCliConfig.logOpts.forward) {
-      return (batch, true)
+      return (batch, log != null)
```
pan3793, line 141: would you please explain a little about why this change is required?

turboFei: before, the fetch log and check state in the same context.

Now the fetch log operation is in log thread.

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
```diff
@@ -79,7 +79,33 @@ class LogBatchCommand(
...
+          val timeout = conf.get(CTL_BATCH_LOG_ON_FAILURE_TIMEOUT)
+          var hasRemainingLogs = true
+          while (hasRemainingLogs && System.currentTimeMillis() - startTime < timeout) {
+            try {
+              val log = kyuubiInstanceBatchRestApi.getBatchLocalLog(
+                batchId,
+                from,
+                size)
+              from += log.getLogRowSet.size
+              log.getLogRowSet.asScala.foreach(x => info(x))
```
pan3793, line 99: how about extracting the duplicated code block?

pan3793 (approved):
LGTM except the code dup

turboFei:
> LGTM except the code dup

thanks, addressed.

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
```diff
@@ -55,15 +55,17 @@ class LogBatchCommand(
 
       withKyuubiInstanceRestClient(kyuubiRestClient, kyuubiInstance) { kyuubiInstanceRestClient =>
         val kyuubiInstanceBatchRestApi: BatchRestApi = new BatchRestApi(kyuubiInstanceRestClient)
+
+        def getOperationLogOnce(): OperationLog = {
```
pan3793, line 59: prefer to use "fetch" or "retrieve" to distinguish w/ "getter", and it not fetch but also print the log, the method name should mention that, or introduce a param to control the display behavior.

turboFei:
thanks, merging to master

PR #3722, opened 2022-10-29
Set default parameters for BatchesResource in server side

turboFei:

Set the same default values for BatchesResource in server side.


pan3793 (approved):

turboFei:
thanks, merged to master

PR #3739, opened 2022-11-01
[REST] Remove unused parameters in SessionOpenRequest

hddong:

We should put the authorization info in request header, user,password,ip in SessionOpenRequest are unnecessary.

- [] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request


pan3793:
Are you touching a released publish API? cc @turboFei

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -145,7 +145,7 @@ private[v1] class SessionsResource extends ApiRequestContext with Logging {
     val handle = fe.be.openSession(
       TProtocolVersion.findByValue(request.getProtocolVersion),
       userName,
-      request.getPassword,
+      null,
```
turboFei, line 148: prefer `""`

turboFei:
>  Are you touching a released public API?

I am ok with this, we can release it in 1.7.0.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
hddong: > prefer `""`

Done.

pan3793:
If it's an unreleased API, the change makes sense to me.

hddong:
@pan3793 @turboFei This API has released in 1.5.0-Incubating, but I think it worth revising. It is mislead users to pass authentication information through the HTTP POST body, which leads to authentication failure.

turboFei (approved):

turboFei:
could you help rebase the code @hddong ?

@pan3793 I think the change is fine, it was a mistake and now we are fixing it.

yaooqinn (approved):

turboFei:
> Should we mention it in the migration guide?

It is better to mention it in the migration guide.

pan3793:
> > Should we mention it in the migration guide?
> 
> It is better to mention it in the migration guide.

it's fine then.

hddong:
@turboFei  Have rebased, and added on migration guide.
cc @pan3793 @yaooqinn

turboFei (approved):

docs/deployment/migration-guide.md
```diff
@@ -24,6 +24,7 @@
 * Since Kyuubi 1.7, Kyuubi returns engine's information for `GetInfo` request instead of server. To restore the previous behavior, set `kyuubi.server.info.provider` to `SERVER`.
 * Since Kyuubi 1.7, Kyuubi session type `SQL` is refactored to `INTERACTIVE`, because Kyuubi supports not only `SQL` session, but also `SCALA` and `PYTHON` sessions.
   User need to use `INTERACTIVE` sessionType to look up the session event.
+* Since Kyuubi 1.7, the REST API of `Open(create) a session` will not contains parameters `user` `password` and `IpAddr`. We should set user and password in `Authorization` of http request when needed.
```
turboFei, line 27: nit: 

```
User and password should be set in `Authorization` of http request if needed.
```

PR #3746, opened 2022-11-02
Support to specify multiple config items with `--conf` in kyuubi command line

turboFei:

Using `unbounded()` for `--conf`.


kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/CommonCommandLine.scala
```diff
@@ -49,11 +49,13 @@ trait CommonCommandLine {
         .text("The value of hive.server2.proxy.user config."),
       opt[String]("conf")
```
ulysses-you, line 50: try Map[String, String] ?

turboFei: it works with `.unbounded()` now.

ulysses-you (approved):

pan3793 (approved):

PR #3747, opened 2022-11-03
Fix flaky test MetadataManagerSuite: metadata request metrics

turboFei:

Seems there is some delay for the finally block.
image
Reset the MetadataManager after each test to fix it.


turboFei:
alternative: https://github.com/apache/incubator-kyuubi/pull/3748

turboFei:
```
  List("serviceUri=localhost:42259;version=1.7.0-SNAPSHOT;refId=2395c94d-10ec-480c-a9a4-0d108a9aaf11;sequence=0000000000") had size 1 instead of expected size 0 (AdminResourceSuite.scala:103)
```

pan3793:
> alternative: #3748

I lean toward to this approach

Issue #3723, opened 2022-10-30
[Subtask] save the SQL session metadata into MetadataManager


lightning-L:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

save the SQL session metadata into MetadataManager

### Parent issue

https://github.com/apache/incubator-kyuubi/issues/3554

### Are you willing to submit PR?

- [ ] Yes. I can submit a PR independently to improve.
- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

lightning-L:
1. save the SQL session metadata into MetadataManager
2. list the live sessions from the MetadataManager
3. support to list sessions with filter condition, such as user

EricGao888:
@lightning-L Hi, this feature seems significant when using kyuubi restful sessions in HA deployment. May I ask why it is not merged into main branch? Thanks : )

z1131392774:
Hi @pan3793 @yaooqinn ,sorry to bother you.

Following up on my previous comment, I'd like to take over this inactive issue. I have some initial thoughts on the implementation and would appreciate your feedback.

Here are my preliminary ideas:

1. SessionState Definition

I propose creating a new SessionState enum, similar to OperationState. My initial thought is to include the following states:

- INITIALIZED: Session is initialized.
- RUNNING: Session is active and running operations.
- PENDING: Session is pending, waiting for the engine to start.
- CLOSED: Session is closed by a user request.
- TIMEOUT: Session is closed due to inactivity timeout.
- ERROR: Session terminated with an error.

Are these states sufficient, or do you have other suggestions?

2. Fields to Persist for Session Metadata

To store session metadata in the MetadataStore, I think we should persist the following core fields, which are essential for management and troubleshooting:

- Session Identity: identifier, sessionType
- User & Connection: realUser, username, ipAddress
- State & Timestamps: state, createTime, endTime
- Kyuubi Instance: kyuubiInstance
- Configuration: requestConf
- Engine Details: engineType, engineId, engineName, engineUrl, engineError
- Session/Job Name: requestName

Does this list seem complete? Is there anything to add or remove?

I will wait for your feedback before starting the work.

[...]

PR #3755, opened 2022-11-03
[KYUUBI #3723] save the SQL session metadata into MetadataManager

lightning-L:

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
@@ -36,14 +36,13 @@ import org.apache.kyuubi.config.KyuubiReservedKeys.{KYUUBI_CLIENT_IP_KEY, KYUUBI
 import org.apache.kyuubi.events.KyuubiEvent
 import org.apache.kyuubi.operation.OperationHandle
 import org.apache.kyuubi.server.api.ApiRequestContext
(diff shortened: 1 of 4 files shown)


lightning-L:
@turboFei 
- SessionData contains duration and idleTime, which is calculated by session lastAccessTime and lastIdleTime, but those two are not stored in the Metadata. How to get duration and idleTime after fetching the session metadata?
- We could get Resource/ClassName/ClusterManager for session from engine proc builder. But before engine is created, the prod builder is null. Currently I am not sure how to set those values when opening session.
- The batch session and sql session related operation are mixed in one class(KyuubiSessionManager), do we need a refactor? For example, separate this into KyuubiSessionManager and KyuubiBatchSessionManager.

turboFei:
>  SessionData contains duration and idleTime, which is calculated by session lastAccessTime and lastIdleTime, but those two are not stored in the Metadata. How to get duration and idleTime after fetching the session metadata?

Seems there is no appropriate solution now.

> We could get Resource/ClassName/ClusterManager for session from engine proc builder. But before engine is created, the prod builder is null. Currently I am not sure how to set those values when opening session.

How about leave them empty now?

> The batch session and sql session related operation are mixed in one class(KyuubiSessionManager), do we need a refactor? For example, separate this into KyuubiSessionManager and KyuubiBatchSessionManager.

I think it is not needed.


cc @ulysses-you  @pan3793

ulysses-you:
session rest api is less used.. is it possible to focus on batch api ?

PR #3756, opened 2022-11-03
[KYUUBI #3764] Introduce the base parser framework for kyuubi server

Yikf:

Close https://github.com/apache/incubator-kyuubi/issues/3764, This pr aims to add the base parser framework to kyuubi server for the following purposes:
- Add syntax that is executable on the server side to enhance capabilities
- Add the ability to translate SQL

### DIFF
--- a/dev/dependencyList
+++ b/dev/dependencyList
@@ -18,6 +18,7 @@
 HikariCP/4.0.3//HikariCP-4.0.3.jar
 animal-sniffer-annotations/1.21//animal-sniffer-annotations-1.21.jar
 annotations/4.1.1.4//annotations-4.1.1.4.jar
+antlr4-runtime/4.8//antlr4-runtime-4.8.jar
 aopalliance-repackaged/2.6.1//aopalliance-repackaged-2.6.1.jar
 automaton/1.11-8//automaton-1.11-8.jar
 classgraph/4.8.138//classgraph-4.8.138.jar
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/session/FlinkSQLSessionManager.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/session/FlinkSQLSessionManager.scala
@@ -21,6 +21,7 @@ import org.apache.flink.table.client.gateway.context.DefaultContext
 import org.apache.flink.table.client.gateway.local.LocalExecutor
 import org.apache.hive.service.rpc.thrift.TProtocolVersion
 
+import org.apache.kyuubi.config.KyuubiConf.FrontendProtocols.FrontendProtocol
 import org.apache.kyuubi.engine.flink.operation.FlinkSQLOperationManager
 import org.apache.kyuubi.session.{Session, SessionHandle, SessionManager}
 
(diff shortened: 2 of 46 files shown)


wForget:
I prefer syntax like iceberg spark procedures:

```
CALL  func(params)
```

FYI: #1301

kyuubi-server/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSQL.g4
```diff
@@ -0,0 +1,178 @@
+/*
```
yaooqinn, line 1: can we separate Lexer and Parser?

kyuubi-server/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSQL.g4
```diff
@@ -0,0 +1,178 @@
...
+    ;
+
+runnable
+    : KILL ENGINE                                                   #killEngine
+    | LAUNCH ENGINE                                                 #launchEngine
+    | ALTER SESSION SET propertyList                                #alterEngineConfig
+    ;
+
+translate
+    : RENAME TABLE identifier TO identifier                         #renameTable
```
pan3793, line 65: we should be careful about extending the syntax, currently, the generic Kyuubi SQL scope is narrowed to management. it says, the typical ANSI-like SQL is 

```
SHOW <RESOURCES>
CREATE <RESOURCE> ...
ALTER <RESOURCE> SET ...
DROP <RESOURCE> ...
```

In Kyuubi case, I think we can start w/ the resource `SESSION` or `ENGINE`.


The "translate" should be protocol-specific, e.g. we only need it for MySQL FE now.

kyuubi-server/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSQL.g4
Yikf: Thanks, will separate Lexer and Parser files

kyuubi-server/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSQL.g4
Yikf: Thanks for your thoughts, That's great; and thanks @wForget, prefer this style to that of `CALL  func(params)`, what do you think~

@pan3793 I like this style for extending the syntaxm, but and have a bit weird if we use `CREATE ENGINE` for launch engine, `DROP ENGINE` for kill/stop engine.

kyuubi-server/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSQL.g4
```diff
@@ -0,0 +1,178 @@
...
+statement
+    : runnable                                                      #logicalRunnable
+    | translate                                                     #logicalTranslate
+    | .*?                                                           #passThrough
+    ;
+
+runnable
+    : KILL ENGINE                                                   #killEngine
+    | LAUNCH ENGINE                                                 #launchEngine
+    | ALTER SESSION SET propertyList                                #alterEngineConfig
```
cfmcgrady, line 61: then the `ALTER` DDL still works? for instance:

```sql
ALTER TABLE loc_orc SET fileformat orc;
```

PR #3762, opened 2022-11-04
[KYUUBI #3782][PYSPARK] Initial support PySpark

### LINKED ISSUE #3758: [FEATURE] Can we mix python the same way we mix scala?
### Code of Conduct

### Search before asking

### Describe the feature

Compared with Scala, Python is more commonly used by data analysts and bi development engineers. Can we support mixing python, just like mixing scala?

cfmcgrady:

Close #3758 #3782

Limitations:
- only support kyuubi beeline

Examples:

截屏2022-11-04 下午5 16 11

截屏2022-11-04 下午5 16 32


turboFei:
fixed the ci issue, could you rebase this pr on master?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/api/python/KyuubiPythonGatewayServer.scala
```diff
@@ -0,0 +1,66 @@
...
+import java.io.{DataOutputStream, File, FileOutputStream}
+import java.nio.charset.StandardCharsets.UTF_8
+import java.nio.file.Files
+
+import org.apache.spark.SparkConf
+import org.apache.spark.internal.Logging
+
+object KyuubiPythonGatewayServer extends Logging {
+
+  val CONNECTION_FILE_PATH = "/tmp/connection.info"
```
pan3793, line 29:
```suggestion
  val CONNECTION_FILE_PATH = s"${sys.props("java.io.tmpdir")}/connection.info"
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -0,0 +1,238 @@
...
+    t2.interrupt()
+    workerProcess.destroy()
+  }
+}
+
+object ExecutePython extends Logging {
+  //  val pythonExec = "/usr/local/share/anaconda3/envs/pyspark-3.3.1-py36/bin/python"
+
+  // TODO:(fchen) get from conf
+  val pythonExec = sys.env.get("PYSPARK_PYTHON")
```
pan3793, line 103:
```suggestion
  val pythonExec = sys.env.getOrElse("PYSPARK_PYTHON", sys.env.getOrElse("PYSPARK_DRIVER_PYTHON", "python3"))
```

externals/kyuubi-spark-sql-engine/src/main/resources/python/kyuubi_util.py
```diff
@@ -0,0 +1,89 @@
...
+import shlex
+import shutil
+import socket
+import platform
+import tempfile
+import time
+from subprocess import Popen, PIPE
+
+if sys.version >= '3':
+    xrange = range
```
pan3793, line 31: where is `xrange` used?

externals/kyuubi-spark-sql-engine/src/main/resources/python/kyuubi_util.py
```diff
@@ -0,0 +1,89 @@
...
+
+from py4j.java_gateway import java_import, JavaGateway, JavaObject, GatewayParameters
+from py4j.clientserver import ClientServer, JavaParameters, PythonParameters
+from pyspark.context import SparkContext
+from pyspark.serializers import read_int, write_with_length, UTF8Deserializer
+from pyspark.sql import SparkSession
+
+
+def connect_to_exist_gateway():
+    conn_info_file = '/tmp/connection.info'
```
pan3793, line 41: how about passing the path by env?

externals/kyuubi-spark-sql-engine/src/main/resources/python/kyuubi_util.py
```diff
@@ -0,0 +1,89 @@
...
+    # gateway.proc = proc
+
+    # Import the classes used by PySpark
+    java_import(gateway.jvm, "org.apache.spark.SparkConf")
+    java_import(gateway.jvm, "org.apache.spark.api.java.*")
+    java_import(gateway.jvm, "org.apache.spark.api.python.*")
+    java_import(gateway.jvm, "org.apache.spark.ml.python.*")
+    java_import(gateway.jvm, "org.apache.spark.mllib.api.python.*")
+    java_import(gateway.jvm, "org.apache.spark.resource.*")
+    # TODO(davies): move into sql
```
pan3793, line 69: .

dik111:
will it suport kyuubi jdbc in the future？

cfmcgrady:
> will it suport kyuubi jdbc in the future？

what do you mean `kyuubi jdbc`?

pan3793:
> will it suport kyuubi jdbc in the future？

From my understanding, the current implementation supports both Hive Jdbc driver and Kyuubi Jdbc driver, but things are different in the Beeline case.

Beeline is responsible to split SQL(python/scala code) before sending each query through the JDBC statement, and in default, Beeline calls `trim` for each query before sending, it may break the python code since python is indenting sensitive. That's why this PR also modified some code in the Beeline module but does not touch the JDBC module.

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
```diff
@@ -560,4 +558,56 @@ public void run() {
       }
     }
   }
+
+  @Override
+  public String handleMultiLineCmd(String line) throws IOException {
```
pan3793, line 563: I believe you copy this method from the superclass w/ a little bit of modification, but it's hard to distinguish what's exactly modified here. How about making a prepared PR to copy the original method from the superclass? This could benefit reviewers.

dik111:
> > will it suport kyuubi jdbc in the future？
> 
> what do you mean `kyuubi jdbc`?

I mean can I run pyspark code through jdbc or rest api,  because we use Kyuubi as an online SQL gateway

bowenliang123:
I think it's a YES. The python script is sent on JDBC connection in screenshots, just exact same way as the SQL or scala scripts. But it seems like a drawback with a mandatory semicolon at the end of submission to inform SQL cli standard. @dik111

externals/kyuubi-spark-sql-engine/src/main/resources/python/kyuubi_util.py
cfmcgrady: It is used in L137 by function `execute_reply_error(exc_type, exc_value, tb)`

externals/kyuubi-spark-sql-engine/src/main/resources/python/kyuubi_util.py
cfmcgrady: updated.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/api/python/KyuubiPythonGatewayServer.scala
cfmcgrady: to avoid multiple Spark engine conflicts, create a temporary directory each time.

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
cfmcgrady: reverted changes by https://github.com/apache/incubator-kyuubi/pull/3762/commits/1c1cf31d2b5e296ef4df7307433880f484800c8a and filed a new issue #3770 for tracking this

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
```diff
@@ -683,6 +683,16 @@ class SparkOperationSuite extends WithSparkSQLEngine with HiveMetadataTests with
     }
   }
 
+  test("pyspark support") {
+    withMultipleConnectionJdbcStatement()({ statement =>
```
pan3793, line 687: we should assume `python3` is available before running this test

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -0,0 +1,238 @@
...
+        val candidates = Paths.get(cwd, "kyuubi-download", "target")
+          .toFile.listFiles(homeDirFilter)
+        if (candidates == null) None else candidates.map(_.toPath).headOption
+      }.find(Files.exists(_)).map(_.toAbsolutePath.toFile.getCanonicalPath)
+      .getOrElse {
+        throw new IllegalStateException("SPARK_HOME not found!")
+      }
+  }
+
+  private def start_stderr_steam_reader(process: Process): Thread = {
```
pan3793, line 155: why use such naming style?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -0,0 +1,238 @@
...
+    val traceback = response.map(_.content.getTraceback()).getOrElse(Array.empty)
+    iter =
+      new ArrayFetchIterator[Row](Array(Row(output, status, ename, evalue, Row(traceback: _*))))
+  }
+
+}
+
+case class SessionPythonWorker(
+    t1: Thread,
+    t2: Thread,
```
pan3793, line 69: could you give t1 t2 a meaningful name?

externals/kyuubi-spark-sql-engine/src/main/resources/python/kyuubi_util.py
pan3793: it calls `range` instead of `xrange`, so I think it's redundant. And we are not tend to support Python2, all `if sys.version >= '3'` could be removed.

cfmcgrady: updated.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
cfmcgrady: added.

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
cfmcgrady: addressed in https://github.com/apache/incubator-kyuubi/pull/3762/commits/3e4d6e3fc05c641e519ac41d2330bc05dfb61d7f

pan3793 (approved):

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
bowenliang123: Shall we add unspported multi-line scripts to limitation in this PR description?

cfmcgrady: it's supported now

pan3793:
I think it's already in a good shape to get into master, thanks @cfmcgrady and all reviewers.

bowenliang123 (approved):

bowenliang123:
LGTM. And locally tested.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -0,0 +1,238 @@
...
+object ExecutePython extends Logging {
+
+  // TODO:(fchen) get from conf
+  val pythonExec =
+    sys.env.getOrElse("PYSPARK_PYTHON", sys.env.getOrElse("PYSPARK_DRIVER_PYTHON", "python3"))
+  private val isPythonGatewayStart = new AtomicBoolean(false)
+  val kyuubiPythonPath = Files.createTempDirectory("")
+  def init(): Unit = {
+    if (!isPythonGatewayStart.get()) {
+      synchronized {
```
bowenliang123, line 108: Is it thread safe here? it's locking the code block itself rather than syncing the AtomicBoolean and with no double check.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -0,0 +1,238 @@
...
+        KyuubiPythonGatewayServer.start()
+        writeTempPyFile(kyuubiPythonPath, "execute_python.py")
+        writeTempPyFile(kyuubiPythonPath, "kyuubi_util.py")
+        isPythonGatewayStart.set(true)
+      }
+    }
+  }
```
bowenliang123, line 115:
```suggestion
private val kyuubiPythonPath = Files.createTempDirectory("")
  lazy val init: Unit = {
    KyuubiPythonGatewayServer.start()
    writeTempPyFile(kyuubiPythonPath, "execute_python.py")
    writeTempPyFile(kyuubiPythonPath, "kyuubi_util.py")
  }
```

simply to use `lazy val` to lock the object instance instead.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
cfmcgrady: added double-check.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
cfmcgrady: How does the `lazy val` expression lock the object instance?

bowenliang123: `lazy val` in Scala basically equivalent to auto added method in compile time as `synchronized (this) +  double check empty` in Java

cfmcgrady: > How does the `lazy val` expression lock the object instance?

FYI: https://www.baeldung.com/scala/lazy-val

cfmcgrady: > How about simply use lazy val to lock the object instance instead.

I think `double-check` style is more readable than `lazy val`, WDYT? cc @pan3793

yaooqinn (approved):

yaooqinn:
Let's create a ticket to track for such a big feature.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
bowenliang123: But more readness burden for one-time usage, and introduced more memory footprint for this single check.

pan3793: either is fine to me.

pan3793:
Thanks, merging to master

PR #3577, opened 2022-09-28
Transfer connection url when opening session

turboFei:
### _Why are the changes needed?_

Now we support multiple frontend services.

We need transfer the connection url into session to identify the connection type.

It can also fix #3575

### _How was this patch tested?_
Pass existing UT.

PR #3765, opened 2022-11-05
[KYUUBI #3577][FOLLOWUP] Set default connectionUrl to empty string to prevent the UT stuck for metadata request retry

turboFei:

For UT, we might open batch session directly and do not via rest frontend service.

So the connection url is null, and will stuck for metadata request retrying.


turboFei:
cc @pan3793

pan3793 (approved):

turboFei:
thanks, merging to master

PR #3766, opened 2022-11-05
Support real user for KyuubiSession

turboFei:

Support real user for kyuubi session

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
@@ -136,7 +136,13 @@ abstract class TFrontendService(name: String)
     }
   }
 
-  protected def getUserName(req: TOpenSessionReq): (String, String) = {
+  /**
+   * Get the real user and the proxy user.
+   * The real user is the user used for session authentication.
+   * The proxy user is the session owner, if proxy user is used, it is the proxy user,
+   * otherwise, it is the real user.
+   */
+  protected def getRealUserAndProxyUser(req: TOpenSessionReq): (String, String) = {
     val realUser: String =
       ServiceUtils.getShortName(authFactory.getRemoteUser.getOrElse(req.getUsername))
     if (req.getConfiguration == null) {
@@ -145,6 +151,7 @@ abstract class TFrontendService(name: String)
       realUser -> getProxyUser(req.getConfiguration, authFactory.getIpAddress.orNull, realUser)
     }
   }
+
   protected def getIpAddress: String = {
     authFactory.getIpAddress.orNull
   }
@@ -157,7 +164,7 @@ abstract class TFrontendService(name: String)
   protected def getSessionHandle(req: TOpenSessionReq, res: TOpenSessionResp): SessionHandle = {
     val protocol = getMinVersion(SERVER_VERSION, req.getClient_protocol)
     res.setServerProtocolVersion(protocol)
-    val (realUser, userName) = getUserName(req)
+    val (realUser, proxyUser) = getRealUserAndProxyUser(req)
     val ipAddress = getIpAddress
     val configuration =
       Map(KYUUBI_CLIENT_IP_KEY -> ipAddress) ++
@@ -167,7 +174,7 @@ abstract class TFrontendService(name: String)
           KYUUBI_SESSION_REAL_USER_KEY -> realUser)
     val sessionHandle = be.openSession(
       protocol,
-      userName,
+      proxyUser,
       req.getPassword,
       ipAddress,
       configuration)
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
@@ -32,7 +32,7 @@ import org.apache.thrift.transport.TTransport
 import org.apache.kyuubi.{KyuubiSQLException, Logging, Utils}
 import org.apache.kyuubi.Utils.stringifyException
 import org.apache.kyuubi.config.KyuubiConf.FRONTEND_CONNECTION_URL_USE_HOSTNAME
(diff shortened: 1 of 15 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -136,13 +136,13 @@ abstract class TFrontendService(name: String)
     }
   }
 
-  protected def getUserName(req: TOpenSessionReq): String = {
+  protected def getUserName(req: TOpenSessionReq): (String, String) = {
```
pan3793, line 139: please add doc to explain the return values

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -157,12 +157,14 @@ abstract class TFrontendService(name: String)
   protected def getSessionHandle(req: TOpenSessionReq, res: TOpenSessionResp): SessionHandle = {
     val protocol = getMinVersion(SERVER_VERSION, req.getClient_protocol)
     res.setServerProtocolVersion(protocol)
-    val userName = getUserName(req)
+    val (realUser, userName) = getUserName(req)
```
pan3793, line 160: `userName` is a little bit confusing then, can we give the clear definition somewhere about different `*user`s?

kyuubi-common/src/test/scala/org/apache/kyuubi/KerberizedTestHelper.scala
```diff
@@ -34,6 +34,7 @@ import org.ietf.jgss.{GSSContext, GSSException, GSSManager, GSSName}
 import org.scalatest.time.SpanSugar._
 
 trait KerberizedTestHelper extends KyuubiFunSuite {
+  val principalUserName = "client"
```
pan3793, line 37: better to name it `clientPrincipalUserName`, to distinguish w/ server principal

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
```diff
@@ -274,15 +274,15 @@ final class KyuubiTHttpFrontendService(
...
+  override protected def getUserName(req: TOpenSessionReq): (String, String) = {
     var userName: String = SessionManager.getUserName
     // using the remote ip address instead of that in proxy http header for authentication
     val ipAddress: String = SessionManager.getIpAddress
     if (userName == null) userName = req.getUsername
     userName = getShortName(userName)
     val effectiveClientUser: String = getProxyUser(req.getConfiguration, ipAddress, userName)
     debug("Client's username: " + effectiveClientUser)
-    effectiveClientUser
+    userName -> effectiveClientUser
```
pan3793, line 285: let's unify the terminologies

pan3793: Should these two users be reversed?
I think `effectiveClientUser` means `realUser` here, correct?

An alternative way is keeping this but reversing other places. (IMO, return `(super user, proxy user)` is more intuitive than `(proxy user, super user)`)

turboFei:
@pan3793 all comments addressed

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -136,15 +136,22 @@ abstract class TFrontendService(name: String)
...
   }
 
-  protected def getUserName(req: TOpenSessionReq): String = {
+  /**
+   * Get the real user and the proxy user.
+   * The real user is the user used for session authentication.
+   * The proxy user is the session user, if proxy user is used, it is the proxy user,
+   * otherwise, it is the real user.
+   */
+  protected def getRealUserAndProxyUser(req: TOpenSessionReq): (String, String) = {
```
pan3793, line 145:
```suggestion
  /**
   * Get the real user and the session user.
   * The real user is the user used for session authentication.
   * The session user is the proxy user if proxy user is provided, otherwise is the real user.
   */
  protected def getRealUserAndSessionUser(req: TOpenSessionReq): (String, String) = {
```

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -136,15 +136,22 @@ abstract class TFrontendService(name: String)
...
     val realUser: String =
       ServiceUtils.getShortName(authFactory.getRemoteUser.getOrElse(req.getUsername))
     if (req.getConfiguration == null) {
-      realUser
+      realUser -> realUser
```
pan3793, line 149: let's make it more clear by using the following pattern.
```
val realUser = ...
val sessionUser = if (...) ... else ...
(realUser -> sessionUser)
```

turboFei:
thanks @pan3793  addressed

pan3793 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
```diff
@@ -274,15 +274,13 @@ final class KyuubiTHttpFrontendService(
...
+    val realUser = getShortName(Option(SessionManager.getUserName).getOrElse(req.getUsername))
     // using the remote ip address instead of that in proxy http header for authentication
     val ipAddress: String = SessionManager.getIpAddress
-    if (userName == null) userName = req.getUsername
-    userName = getShortName(userName)
-    val effectiveClientUser: String = getProxyUser(req.getConfiguration, ipAddress, userName)
-    debug("Client's username: " + effectiveClientUser)
-    effectiveClientUser
+    val sessionUser: String = getProxyUser(req.getConfiguration, ipAddress, realUser)
+    debug(s"Client's real user: $realUser, session user $sessionUser")
```
pan3793, line 282:
```suggestion
    debug(s"Client's real user: $realUser, session user: $sessionUser")
```

PR #3769, opened 2022-11-07
[SUBTASK]Add additional engine ui url information to the ha

zwangsheng:

Subtask of #3420

Add engine ui url info.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
@@ -96,8 +96,11 @@ class SparkTBinaryFrontendService(
   override def attributes: Map[String, String] = {
     var map = Map(KYUUBI_ENGINE_ID -> KyuubiSparkUtil.engineId)
     sc.uiWebUrl.foreach { url =>
-      map ++= Map(KYUUBI_ENGINE_UI_URL -> url)
+      val str = url.split("//:")(1)
+      map ++= Map(KYUUBI_ENGINE_UI_URL -> str)
+      info(str)
     }
+    info(map)
     map
   }
 }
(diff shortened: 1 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
```diff
@@ -94,7 +94,11 @@ class SparkTBinaryFrontendService(
   }
 
   override def attributes: Map[String, String] = {
-    Map(KYUUBI_ENGINE_ID -> KyuubiSparkUtil.engineId)
+    var map = Map(KYUUBI_ENGINE_ID -> KyuubiSparkUtil.engineId)
+    sc.uiWebUrl.foreach { url =>
+      map ++= Map(KYUUBI_ENGINE_UI_URL -> url)
```
cxzl25, line 99: `WithKyuubiServer` disables spark ui by default (`conf.set("spark.ui.enabled", "false")`) , so UT appears NPE because there is no UI URL.

But after enable spark ui, we will find that some characters in spark ui url are illegal.

org.apache.curator.utils.PathUtils#validatePath
```
            } else if (c == '/' && lastc == '/') {
                reason = "empty node name specified @" + i;
                break;
```

```
Caused by: java.lang.IllegalArgumentException: Invalid path string "/kyuubi_test_1.7.0-SNAPSHOT_USER_SPARK_SQL/csy/default/serviceUri=localhost:56196;version=1.7.0-SNAPSHOT;kyuubi.engine.id=local-1667988857376-692873171;kyuubi.engine.ui.url=http://XXX:56175;refId=f7bae472-5e6e-42e9-8013-a8f72b3adfdd;sequence=" caused by empty node name specified @179
	at org.apache.kyuubi.shade.org.apache.curator.utils.PathUtils.validatePath(PathUtils.java:102)
	at org.apache.kyuubi.shade.org.apache.curator.framework.recipes.nodes.PersistentNode.<init>(PersistentNode.java:153)
	at org.apache.kyuubi.ha.client.zookeeper.ZookeeperDiscoveryClient.createPersistentNode(ZookeeperDiscoveryClient.scala:390)
```

zwangsheng:
Hi @cxzl25 . Thanks for your info! I will push fix commit later with enabled `spark.ui.enabled` and remove scheme `http://`

PR #3778, opened 2022-11-08
[KYUUBI #3658][FOLLOWUP] Do not use lazy val to prevent port BindException

turboFei:

If binary ssl enabled, we will bind the port to ssl server socket.
if we invoke the lazy val serverSocket, it will throw BindException.

image


pan3793 (approved):

turboFei:
thanks, merged to master

Issue #3781, opened 2022-11-08
[Bug] One session add repeated jar, other session DML sqls will failed


CHzxp:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

If the user repeatedly adds the jars in one session , the dml sql from other session will fail. this is because AddJarCommand will add jar to common sparkContext, other session task  failed when down dependent jars to container .

### Affects Version(s)

master/1.7

### Kyuubi Engine Log Output

```logtalk
Caused by: org.apache.spark.SparkException: File ./xxx-0.0.1.jar exists and does not match contents of hdfs://user_zzz/yyy/udf/xxx-0.0.1.jar
at org.apache.spark.util.Utils$.copyFile(Utils.scala:685)
at org.apache.spark.util.Utils$.fetchFile(Utils.scala:542)
at org.apache.spark.executor.Executor.$anonfun$updateDependencies$13(Executor.scala:963)
at org.apache.spark.executor.Executor.$anonfun$updateDependencies$13$adapted(Executor.scala:955)
```

### Are you willing to submit PR?

- [X] Yes. I can submit a PR independently to fix.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #3783, opened 2022-11-08
[KYUUBI #3781] add repeat jar will case other session DML failed

CHzxp:
Why are the changes needed?_
As described in issues https://github.com/apache/incubator-kyuubi/issues/3781: other session task will fail when other session add repeat jar

How was this patch tested?_


ulysses-you:
The reason is Spark does not support dependencies isolation, I'm fine with a workaround

extensions/spark/kyuubi-extension-spark-3-2/src/main/scala/org/apache/kyuubi/sql/AddRepeatJarCloseSession.scala
```diff
@@ -0,0 +1,53 @@
...
+
+case class AddRepeatJarCloseSession(session: SparkSession) extends Rule[LogicalPlan] {
+
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+
+    plan match {
+      case i @ AddJarsCommand(paths) =>
+        val checkExistsResult = findJarNameIfExists(session.sparkContext, paths)
+        if (checkExistsResult.nonEmpty) {
+          session.close()
```
ulysses-you, line 33: Why close session ? it will shutdown the Spark application

extensions/spark/kyuubi-extension-spark-3-2/src/main/scala/org/apache/kyuubi/sql/AddRepeatJarCloseSession.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.command.AddJarsCommand
+
+case class AddRepeatJarCloseSession(session: SparkSession) extends Rule[LogicalPlan] {
+
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+
+    plan match {
+      case i @ AddJarsCommand(paths) =>
```
ulysses-you, line 30: `AddFilesCommand` and `AddArchivesCommand` have the same issue ?

CHzxp:
> The reason is Spark does not support dependencies isolation, I'm fine with a workaround

how do you solve this problem?

extensions/spark/kyuubi-extension-spark-3-2/src/main/scala/org/apache/kyuubi/sql/AddRepeatJarCloseSession.scala
CHzxp: i will test it after , this problem is from  add jar command

ulysses-you:
> how do you solve this problem?

@CHzxp I mean this pr is a workaround

CHzxp:
> > how do you solve this problem?
> 
> @CHzxp I mean this pr is a workaround

fine, close sparkSession is not a elegant solution, let me think again, thank you !

CHzxp:
My original idea is to close the user's jdbc session, because  I think the current session is a dangerous connection unless user modify their sqls or set config spark.files.overwrite=true

ulysses-you:
@CHzxp we do not need close this session, the add jar command have not executed. A simple way is just throw exception to forbid.

+ 1 to add a new config for this rule

CHzxp:
> @CHzxp we do not need close this session, the add jar command have not executed. A simple way is just throw exception to forbid.
> 
> +1 to add a new config for this rule
when user execute multiple sqls, add repeat add will failed if throw exception, but other sqls will execute continue after this add repeat jar command. i think it is better to discard all left sql after repeat jar command instead of execute left sql continue. because spark can not decide to choose which jar is user want to use , it is a user operation.

ulysses-you:
> add repeat add will failed if throw exception, but other sqls will execute continue after this add repeat jar command

how can this happen ? In a sql script, if the previous sql failed then the whole script should fail. It's not about this pr, For example if user `create table t` failed then the follow `select * from t` should still be executed ? This behavior (close connection) should be corrected at client side.

CHzxp:
> > add repeat add will failed if throw exception, but other sqls will execute continue after this add repeat jar command
> 
> how can this happen ? In a sql script, if the previous sql failed then the whole script should fail. It's not about this pr, For example if user `create table t` failed then the follow `select * from t` should still be executed ? This behavior (close connection) should be corrected at client side.

@ulysses-you sqls from beeline -e or beeline -f or jdbc will split into multiple single sql and execute them one by one.
code location :
org.apache.hive.beeline.Commands#execute
line = line.trim();
            List<String> cmdList = this.getCmdList(line, entireLineAsCommand);

            for(int i = 0; i < cmdList.size(); ++i) {
                String sql = ((String)cmdList.get(i)).trim();
                if (sql.length() != 0 && !this.executeInternal(sql, call)) {
                    return false;
                }
            }

CHzxp:
> > add repeat add will failed if throw exception, but other sqls will execute continue after this add repeat jar command
> 
> how can this happen ? In a sql script, if the previous sql failed then the whole script should fail. It's not about this pr, For example if user `create table t` failed then the follow `select * from t` should still be executed ? This behavior (close connection) should be corrected at client side.
@ulysses-you ， I seem to understand

CHzxp:
> > add repeat add will failed if throw exception, but other sqls will execute continue after this add repeat jar command
> 
> how can this happen ? In a sql script, if the previous sql failed then the whole script should fail. It's not about this pr, For example if user `create table t` failed then the follow `select * from t` should still be executed ? This behavior (close connection) should be corrected at client side.

if user execute sqls by jdbc conn? We can't guarantee that every sql executed by the user has nothing to do with the jar, so we think the current session is a dangerous session， so close this session. namely we only execute normal sql before this repeat jar unless user config  spark.files.overwrite=true

PR #3774, opened 2022-11-07
Output OpenSessionCount when Engine is waiting for exit because it reaches the Max lifetime

lsm1:
### _Why are the changes needed?_

When the Engine reaches the maximum running time, it can output the OpenSessionCount while waiting to exit

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #3630, opened 2022-10-14
[KYUUBI #3627] Support vanilla Jetty for Spark packaged by sbt

pan3793:
### _Why are the changes needed?_

Fix #3627.

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [x] Add screenshots for manual tests if appropriate

manual test, and the class not found error is gone.

- [ ] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #3792, opened 2022-11-09
[SPARK] Engine UI support grace stop

cxzl25:
Now the kill operation of the Spark engine ui will directly shut down, and the SQL that is being executed by the engine will fail. 

We can support grace stop, first log off the engine from the znode, and then wait for the SQL execution to complete.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -50,6 +50,7 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
   private val shutdown = new AtomicBoolean(false)
 
   @volatile private var lifetimeTerminatingChecker: Option[ScheduledExecutorService] = None
+  @volatile private var stopEngineExec: Option[ScheduledExecutorService] = None
 
   override def initialize(conf: KyuubiConf): Unit = {
     val listener = new SparkSQLEngineListener(this)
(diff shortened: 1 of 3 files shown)


cxzl25:
## Current

image

## Fix

image

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineTab.scala
```diff
@@ -63,20 +63,34 @@ case class EngineTab(
   sparkUI.foreach { ui =>
     try {
       // Spark shade the jetty package so here we use reflection
-      val sparkServletContextHandlerClz = loadSparkServletContextHandler
```
wForget, line 66: seems to revert #3630

ulysses-you:
seems confused to end user, just the word for some softwares: `setup` / `auto setup`/ `one click setup`..

pan3793:
How about "Stop Immediately" / "Stop Gracefully"?
And we can explain more when mouse hover the button, or on the confirm dialog after clicking.

wForget (approved):

pan3793 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineTab.scala
```diff
@@ -107,4 +122,10 @@ case class EngineTab(
       engine.get.stop()
     }
   }
+
+  def handleGracefulKillRequest(request: HttpServletRequest): Unit = {
+    if (killEnabled && engine.isDefined && engine.get.getServiceState != ServiceState.STOPPED) {
+      engine.get.gracefulStop()
```
ulysses-you, line 128: what happens if user multiple clicks `gracefulStop` ?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -83,6 +84,33 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
...
+        if (deregistered.compareAndSet(false, true)) {
+          info(s"Spark engine is de-registering from engine discovery space.")
+          frontendServices.flatMap(_.discoveryService).foreach(_.stop())
+        }
+        val openSessionCount = backendService.sessionManager.getOpenSessionCount
+        if (openSessionCount > 0) {
+          info(s"$openSessionCount connection(s) are active, delay shutdown")
+        } else {
+          info(s"Spark engine has no open session now, terminating.")
+          stop()
```
ulysses-you, line 107: seems `DeRegisterWatcher` will do this, `frontendServices.flatMap(_.discoveryService).foreach(_.stop())` is enough ?

cxzl25: Yes. also remove #3774

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EngineTab.scala
cxzl25: Because the service can only be stopped once, when the service is in the stop state, multiple clicks will not affect.

## Engine log
### First click
```
INFO EngineServiceDiscovery: Service[EngineServiceDiscovery] is stopped.
```
## Second click
```
WARN EngineServiceDiscovery: Service[EngineServiceDiscovery] is not started(STOPPED) yet.
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -83,6 +84,31 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
...
+          frontendServices.flatMap(_.discoveryService).foreach(_.stop())
+        }
+
+        if (backendService.sessionManager.getOpenSessionCount <= 0) {
+          info(s"Spark engine has no open session now, terminating.")
+          stop()
+        }
+      }
+    }
+    stopEngineExec =
```
ulysses-you, line 109: shall we check `deregistered` is true here to avoid user multi-clicks ?

ulysses-you (approved):

cxzl25:
Thanks, merged to master.

PR #3800, opened 2022-11-11
provide flexible scene of tenant

### LINKED ISSUE #3809: [FEATURE]
### Code of Conduct

### Search before asking

### Describe the feature

1. In the kyuubi SeesionLimiter module, the connection's limit canbe a different value of each user
2. In the EnfineRef class of tenant module, the group name canbe fetching from database not only hadoop ugi
3. In addition, the two features are all with the help of modifing MetadataStore module slightly 

### Motivation

Using kyuubi session limit and engine share more flexible 

### Describe the solution

Feature 1 
A:   using respective value to limit connections in the SessionLimit class which is fetched from mysql
Feature 2 
A:  using respective value to representing the group name in the EngineRef class which is fetched from mysql
Feature 3.
A: extracting the datasouce in the JDBCMetadataStore to change the method of connection fetch

### Additional context

_No response_

qiaoting:
1. To close #3809

- [ ] Adding an accurate connection limit in SessionLimiter class. It will limit connection by using different number which configed in database
- [ ] At Group's share level, The EngineRef class providing a new method of fetching group's name
- [ ] Adjusting the MetadataStore module by extracting the datasource poll 
- [ ] the two fetures is not enabled by default in the KyuubiConf class

### According test
- [ ]  Test case of connection limit in SessionLimiterSuite class
- [ ]  Group's share level is small change and it was testing in the project of myself

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -2030,22 +2030,6 @@ object KyuubiConf {
       .intConf
       .createOptional
 
-  val SERVER_LIMIT_CONNECTIONS_MANUAL_ENABLE: ConfigEntry[Boolean] =
-    buildConf("kyuubi.server.limit.connections.manual.enable")
-      .doc("Whether to enable custom limit. If it is enabled, Kyuubi will limit the" +
-        " conn's num according to user")
-      .version("1.7.0")
-      .booleanConf
-      .createWithDefault(false)
-
-  val ENGINE_SHARE_LEVEL_GROUP_MANUAL_ENABLE: ConfigEntry[Boolean] =
-    buildConf("kyuubi.engine.share.level.group.manual.enable")
-      .doc("Whether to enable custom group share. If it is enabled, Kyuubi will get the" +
-        " user's group by using relation configured in mysql.")
-      .version("1.7.0")
-      .booleanConf
-      .createWithDefault(false)
-
   val SESSION_PROGRESS_ENABLE: ConfigEntry[Boolean] =
     buildConf("kyuubi.operation.progress.enabled")
       .doc("Whether to enable the operation progress. When true," +
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -2030,6 +2030,22 @@ object KyuubiConf {
       .intConf
       .createOptional
 
+  val SERVER_LIMIT_CONNECTIONS_CUSTOM_ENABLE: ConfigEntry[Boolean] =
+    buildConf("kyuubi.server.limit.connections.custom.enable")
+      .doc("Whether to enable custom limit. If it is enabled, Kyuubi will limit the" +
+        " conn's num according to user")
+      .version("1.6.1")
+      .booleanConf
+      .createWithDefault(false)
+
+  val ENGINE_SHARE_LEVEL_GROUP_CUSTOM_ENABLE: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.share.level.group.custom.enable")
+      .doc("Whether to enable custom group share. If it is enabled, Kyuubi will get the" +
+        " user's group by using relation configured in mysql.")
+      .version("1.6.1")
+      .booleanConf
+      .createWithDefault(false)
+
   val SESSION_PROGRESS_ENABLE: ConfigEntry[Boolean] =
     buildConf("kyuubi.operation.progress.enabled")
       .doc("Whether to enable the operation progress. When true," +
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -2030,19 +2030,19 @@ object KyuubiConf {
       .intConf
       .createOptional
 
-  val SERVER_LIMIT_CONNECTIONS_CUSTOM_ENABLE: ConfigEntry[Boolean] =
-    buildConf("kyuubi.server.limit.connections.custom.enable")
+  val SERVER_LIMIT_CONNECTIONS_MANUAL_ENABLE: ConfigEntry[Boolean] =
+    buildConf("kyuubi.server.limit.connections.manual.enable")
       .doc("Whether to enable custom limit. If it is enabled, Kyuubi will limit the" +
         " conn's num according to user")
-      .version("1.6.1")
+      .version("1.7.0")
       .booleanConf
       .createWithDefault(false)
 
-  val ENGINE_SHARE_LEVEL_GROUP_CUSTOM_ENABLE: ConfigEntry[Boolean] =
-    buildConf("kyuubi.engine.share.level.group.custom.enable")
+  val ENGINE_SHARE_LEVEL_GROUP_MANUAL_ENABLE: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.share.level.group.manual.enable")
       .doc("Whether to enable custom group share. If it is enabled, Kyuubi will get the" +
         " user's group by using relation configured in mysql.")
-      .version("1.6.1")
+      .version("1.7.0")
       .booleanConf
       .createWithDefault(false)
 
--- a/kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
+++ b/kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
@@ -29,3 +29,29 @@ CREATE TABLE metadata(
     INDEX user_name_index(user_name),
     INDEX engine_type_index(engine_type)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `t_sys_user` (
+                              `id` bigint(20) NOT NULL AUTO_INCREMENT,
+                              `group_id` bigint(20) DEFAULT NULL COMMENT '组ID',
+                              `name` varchar(255) NOT NULL COMMENT '用户名',
+                              `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
+                              `is_del` tinyint(4) DEFAULT '0' COMMENT '是否软删除',
+                              PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `t_limit_user` (
+                                `id` bigint(20) NOT NULL AUTO_INCREMENT,
+                                `user_id` bigint(20) NOT NULL COMMENT '用户ID',
+                                `num` int(20) NOT NULL COMMENT '最大连接数',
+                                `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
+                                `is_del` tinyint(4) DEFAULT '0' COMMENT '是否软删除',
+                                PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `t_sys_group` (
+                               `id` bigint(20) NOT NULL AUTO_INCREMENT,
+                               `name` varchar(255) NOT NULL COMMENT '组名，共享GROUP模式使用的名称',
+                               `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
+                               `is_del` tinyint(4) DEFAULT '0' COMMENT '是否删除',
+                               PRIMARY KEY (`id`)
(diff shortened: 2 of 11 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2030,6 +2030,22 @@ object KyuubiConf {
       .intConf
       .createOptional
 
+  val SERVER_LIMIT_CONNECTIONS_CUSTOM_ENABLE: ConfigEntry[Boolean] =
+    buildConf("kyuubi.server.limit.connections.custom.enable")
+      .doc("Whether to enable custom limit. If it is enabled, Kyuubi will limit the" +
```
pan3793, line 2035: how does custom mean?

pan3793:
@qiaoting thanks for contribution. please restore the PR template, read and replace the content based on your change.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2030,6 +2030,22 @@ object KyuubiConf {
...
+        " conn's num according to user")
+      .version("1.6.1")
```
pan3793, line 2037: the next feature version is 1.7.0

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
```diff
@@ -29,3 +29,29 @@ CREATE TABLE metadata(
...
+                              `is_del` tinyint(4) DEFAULT '0' COMMENT '是否软删除',
+                              PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `t_limit_user` (
+                                `id` bigint(20) NOT NULL AUTO_INCREMENT,
+                                `user_id` bigint(20) NOT NULL COMMENT '用户ID',
+                                `num` int(20) NOT NULL COMMENT '最大连接数',
+                                `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
+                                `is_del` tinyint(4) DEFAULT '0' COMMENT '是否软删除',
```
pan3793, line 47: kyuubi does not use soft del style

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
```diff
@@ -29,3 +29,29 @@ CREATE TABLE metadata(
...
+                              `group_id` bigint(20) DEFAULT NULL COMMENT '组ID',
+                              `name` varchar(255) NOT NULL COMMENT '用户名',
+                              `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT '创建时间',
+                              `is_del` tinyint(4) DEFAULT '0' COMMENT '是否软删除',
+                              PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `t_limit_user` (
+                                `id` bigint(20) NOT NULL AUTO_INCREMENT,
+                                `user_id` bigint(20) NOT NULL COMMENT '用户ID',
```
pan3793, line 44: please avoid using chinese

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
qiaoting: early，the config is kyuubi.server.limit.connections.userdefine.enable. After, I  thought userdefine is also unthinkable. therefore, using custom representing customized param which can be defiend by Users themselves

pan3793: if you want to put them into upstream, then custom is not suitable.

yanghua: What do you think of a more straightforward title? For example **throttle** or **downgrade**?

`kyuubi.server.throttle.enable`?

yanghua:
Hi @qiaoting It would be better to file a GitHub issue to track this PR. Please reference the [issue tracker](https://kyuubi.apache.org/issue_tracking.html) and [Pull Request](https://kyuubi.apache.org/pull_request.html) documentation.

qiaoting:
> Hi @qiaoting It would be better to file a GitHub issue to track this PR. Please reference the [issue tracker](https://kyuubi.apache.org/issue_tracking.html) and [Pull Request](https://kyuubi.apache.org/pull_request.html) documentation.

OK

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
qiaoting: I have makd some adjust and commited according your suggestion

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
qiaoting: I have makd some adjust and commited according your suggestion

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
qiaoting: I have makd some adjust and commited according your suggestion

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
qiaoting: I have maked some adjust and committed it according to your suggestion

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
qiaoting: I have maked some adjust and committed it according to your suggestion

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
qiaoting: I have maked some adjust and committed it according to your suggestion

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
```diff
@@ -29,3 +29,29 @@ CREATE TABLE metadata(
...
     INDEX engine_type_index(engine_type)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `t_sys_user` (
+                              `id` bigint(20) NOT NULL AUTO_INCREMENT,
+                              `group_id` bigint(20) DEFAULT NULL COMMENT 'group id',
+                              `name` varchar(255) NOT NULL COMMENT 'user name',
+                              `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'create time',
+                              `remark` varchar(500) NOT NULL COMMENT 'remark',
+                              PRIMARY KEY (`id`)
```
yanghua, line 39: Please notice the indent.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -133,6 +134,13 @@ class KyuubiServer(name: String) extends Serverable(name) {
     if (conf.get(MetricsConf.METRICS_ENABLED)) {
       addService(new MetricsSystem)
     }
+
+    // initialize meta source
+    if (conf.get(FRONTEND_PROTOCOLS).map(FrontendProtocols.withName)
+      .contains(FrontendProtocols.REST)) {
```
yanghua, line 140: Is it the only scenario that we should initialize the meta store?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/UserCustomStore.scala
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.server.metadata
+
+import java.io.Closeable
+
+trait UserCustomStore extends Closeable {
```
yanghua, line 22: What about `UserConfigurationStore` or `UserLevelConfStore` or something else? cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/UserCustomStore.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def getUserLimitList(): Seq[Tuple2[String, Int]]
```
yanghua, line 24: The `Limit` naming is still easily confusing.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCUserCustomStore.scala
```diff
@@ -0,0 +1,153 @@
...
+package org.apache.kyuubi.server.metadata.jdbc
+
+import java.sql.{Connection, PreparedStatement, ResultSet, SQLException}
+
+import scala.collection.mutable.ListBuffer
+
+import org.apache.kyuubi.{KyuubiException, Logging, Utils}
+import org.apache.kyuubi.server.metadata.UserCustomStore
+
+class JDBCUserCustomStore extends UserCustomStore with Logging {
```
yanghua, line 27: ditto, it depends on the naming of `UserCustomStore`.

yanghua (changes requested):
@qiaoting Left some comments.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
qiaoting: This is need changing by kyuubi official

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/UserCustomStore.scala
qiaoting: Ok, I get it

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-mysql.sql
yanghua: @qiaoting response?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -83,7 +85,9 @@ private[kyuubi] class EngineRef(
   // Launcher of the engine
   private[kyuubi] val appUser: String = shareLevel match {
     case SERVER => Utils.currentUser
-    case GROUP =>
+    case GROUP => if (userGroupMap.nonEmpty) {
+        userGroupMap.getOrElse(user, user)
+      } else {
```
pan3793, line 90: we actually need a `UserGroupProvider` here, and the implementation could be `HadoopUserGroupProvider`, `UnixUserGroupProvider`, `LDAPUserGroupProvider` `JbdcUserGroupProvider` etc. But TBH, we should be careful about providing the JDBC-based implementation too early, because once done, we have responsibilities to keep compatibility and database upgrading scripts.

pan3793: I try my best to recall some previous points on the "group".

- one user may have multi groups, so the question here are,
  - only use the primary group, or use the primary group in default but allow the user to use the specific one?
  - all groups should be passed to the engine, it's useful for cases like authz plugin.

Further, some admins want the user to use their personal account(e.g. LDAP) for authentication, but run Job in service account, it's also a kind of "user group mapping" case IMO.

pan3793:
This PR scope is quite large IMO, and the change does not generic and solid enough, I suggest you focus on one part at the beginning, e.g. the user group mapping. I left some comments about this part.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
pan3793: cc @leoluan2009 as you are working on LDAP enhancement, this is something we need to do in the next step.

qiaoting: Thank you for your valuable advice，let me soak it up.

pan3793: FYI, the pluggable `GroupProvider` is added in https://github.com/apache/incubator-kyuubi/pull/3897

qiaoting: Ok, thank you

PR #3803, opened 2022-11-14
Improve the data interchange performance via Arrow serialization/deserialization

### LINKED ISSUE #3794: [SubTask] Initial Implement Arrow-based results serialization
### Code of Conduct

### Search before asking

Improve the data interchange performance via Arrow serialization/deserialization

### How should we improve?

_No response_

cfmcgrady:

to close #3794 

This PR introduces the Apache Arrow library to improve the performance of data interchange.

data flow diagram with saprk engine

```
1. Spark Executor -> 2. Spark Driver -> 3. Kyuubi Server -> 4. Beeline/KyuubiHiveDriver
```

1. for simple types(Numeric/String/Binary/Boolean/Datetime/Interval), we serialize the columns on the Spark executor side and deserialize them on the client side.

2. for complex types(Array/Map/Struct), we first convert these columns to hive string and serialize the string columns on the executor side, and finally deserialize them as string type on the client side. 

### Input dataset

```scala
spark.sql(
  """
    |select * from tpcds.sf10.catalog_sales limit 2000000
    |""".stripMargin)
  .write
  .mode("overwrite")
  .save("/tmp/parquet/tpcds/sf10/catalog_sales_2000000")
```

### Query

cat /tmp/b.sql

```sql
select * from parquet.`/tmp/parquet/tpcds/sf10/catalog_sales_2000000`;
```

### Beeline command
```
bin/beeline -u 'jdbc:hive2://0.0.0.0:10009/;' --hiveconf kyuubi.session.result.codec=simple/arrow --hiveconf kyuubi.operation.incremental.collect=false/true -f /tmp/b.sql > /dev/null
```

### Spark

```
/Users/fchen/Software/spark-3.3.0-bin-hadoop3/bin/spark-submit \
    --class org.apache.kyuubi.engine.spark.SparkSQLEngine \
    --driver-memory 6g \
    --master local[*] \
    ......
```

### Result

| kyuubi.session.result.codec | kyuubi.operation.incremental.collect | time spent|
| --- | --- | --- |
| arrow | false | 272.178 seconds |
| arrow | true | 146.76 seconds |
| simple | true | 607.523 seconds |
| simple | false | 429.374 seconds |


yaooqinn:
👍

pan3793:
This performance improvement looks more reasonable to me

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkArrowbasedOperationSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.apache.kyuubi.engine.spark.operation
+
+import org.apache.kyuubi.engine.spark.WithSparkSQLEngine
+import org.apache.kyuubi.operation.SparkDataTypeTests
+
+class SparkArrowbasedOperationSuite extends WithSparkSQLEngine with SparkDataTypeTests {
+
+  override protected def jdbcUrl: String = getJdbcUrl + "#kyuubi.session.result.codec=arrow"
```
pan3793, line 25: why not put it into `withKyuubiConf` ?

pom.xml
```diff
@@ -1727,6 +1752,8 @@
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>${maven.plugin.compiler.version}</version>
                     <configuration>
+                        <skipMain>true</skipMain> <!-- skip compile -->
+                        <skip>true</skip> <!-- skip testCompile -->
```
pan3793, line 1756: why should add them?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiQueryResultSet.java
```diff
@@ -29,14 +29,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** KyuubiQueryResultSet. */
-public class KyuubiQueryResultSet extends KyuubiBaseResultSet {
+class KyuubiQueryResultSet extends KyuubiBaseResultSet {
```
pan3793, line 32: why remove public?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -199,15 +199,32 @@ private boolean executeWithConfOverlay(String sql, Map<String, String> confOverl
...
-            .setClient(client)
-            .setSessionHandle(sessHandle)
-            .setStmtHandle(stmtHandle)
-            .setMaxRows(maxRows)
-            .setFetchSize(fetchSize)
-            .setScrollable(isScrollableResultset)
-            .build();
+
+    String resultCodec = connection.getResultCodec().toLowerCase(Locale.ROOT);
+    System.out.println("kyuubi.session.result.codec:" + resultCodec);
```
pan3793, line 204: nit

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -231,15 +248,30 @@ public boolean executeAsync(String sql) throws SQLException {
...
-        new KyuubiQueryResultSet.Builder(this)
-            .setClient(client)
-            .setSessionHandle(sessHandle)
-            .setStmtHandle(stmtHandle)
-            .setMaxRows(maxRows)
-            .setFetchSize(fetchSize)
-            .setScrollable(isScrollableResultset)
-            .build();
+    String resultCodec = connection.getResultCodec().toLowerCase(Locale.ROOT);
+    System.out.println("kyuubi.session.result.codec:" + resultCodec);
```
pan3793, line 252: nit

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkArrowbasedOperationSuite.scala
cfmcgrady: `withKyuubiConf` only changes the engine behavior, we need a session-level setting.

put it into `jdbcVars`

kyuubi-hive-jdbc/pom.xml
```diff
@@ -37,6 +37,57 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-base</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-json-provider</artifactId>
```
pan3793, line 83: why should we include jackson extensions? I saw arrow-vector depends on jackson-core, but is it a hard dependency or not? I checked the kyuubi-hive-jdbc-shaded, it's changed from 10mb to 17mb after this patch, better to keep it small

pan3793:
please update NOTICE, shading rules for kyuubi-hive-jdbc-shaded, I don't have other major questions other than this one.

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkDataTypeTests.scala
```diff
@@ -0,0 +1,345 @@
...
+ */
+
+package org.apache.kyuubi.operation
+
+import java.sql.{Date, Timestamp}
+
+import org.apache.kyuubi.engine.SemanticVersion
+
+trait SparkDataTypeTests extends HiveJDBCTestHelper {
+  protected lazy val SPARK_ENGINE_MAJOR_MINOR_VERSION: (Int, Int) = sparkEngineMajorMinorVersion
```
pan3793, line 25: can be simplified by `SemanticVersion`

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiArrowBasedResultSet.java
```diff
@@ -0,0 +1,452 @@
...
+      throw new IllegalStateException("arrow schema is null");
+    }
+    allocator = ArrowUtils.rootAllocator.newChildAllocator("ArrowResultSet", 0, Long.MAX_VALUE);
+    root = VectorSchemaRoot.create(arrowSchema, allocator);
+  }
+
+  @Override
+  public int findColumn(String columnName) throws SQLException {
+    int columnIndex = 0;
+    boolean findColumn = false;
```
pan3793, line 66:
```suggestion
    boolean found= false;
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -190,4 +211,15 @@ class ExecuteStatement(
         SparkOperationEvent(this, operationListener.flatMap(_.getExecutionId)))
     }
   }
+
+  // TODO:(fchen) make this configurable
+  val kyuubiBeelineConvertToString = true
+
+  def convertComplexTypeToStringTypeOrNot(df: DataFrame): DataFrame = {
```
pan3793, line 218:
```suggestion
  def convertComplexType(df: DataFrame): DataFrame = {
```

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiArrowQueryResultSet.java
```diff
@@ -0,0 +1,498 @@
...
+        columnTypes.add(primitiveTypeEntry.getType());
+        columnAttributes.add(getColumnAttributes(primitiveTypeEntry));
+      }
+      arrowSchema =
+          ArrowUtils.toArrowSchema(
+              columnNames, convertComplexTypeToStringType(columnTypes), columnAttributes);
+    } catch (SQLException eS) {
+      throw eS; // rethrow the SQLException as is
+    } catch (Exception ex) {
+      ex.printStackTrace();
```
pan3793, line 276: nit

kyuubi-hive-jdbc/pom.xml
cfmcgrady: Reduced pom.xml to minimum requirements, the shaded jar still has 16m size.

```
>ll -h kyuubi-hive-jdbc-shaded/target/kyuubi-hive-jdbc-shaded-1.7.0-SNAPSHOT.jar
-rw-r--r--  1 fchen  staff    16M 11 25 10:23 kyuubi-hive-jdbc-shaded/target/kyuubi-hive-jdbc-shaded-1.7.0-SNAPSHOT.jar
```

pan3793: if we split out the optional deps out, e.g. zk, arrow, the core jar should be small enough. it's something we are plan to do in the future.

cfmcgrady:
```
mvn dependency:tree -pl kyuubi-hive-jdbc | grep -v test
```

the output before this PR:

```
[INFO] org.apache.kyuubi:kyuubi-hive-jdbc:jar:1.7.0-SNAPSHOT
[INFO] +- org.apache.thrift:libfb303:jar:0.9.3:compile
[INFO] +- org.apache.thrift:libthrift:jar:0.9.3:compile
[INFO] +- org.apache.hive:hive-service-rpc:jar:3.1.3:compile
[INFO] +- commons-codec:commons-codec:jar:1.15:compile
[INFO] +- org.apache.commons:commons-lang3:jar:3.10:compile
[INFO] +- com.google.guava:guava:jar:30.1-jre:compile
[INFO] |  \- com.google.guava:failureaccess:jar:1.0.1:compile
[INFO] +- com.google.code.findbugs:jsr305:jar:3.0.2:provided
[INFO] +- org.apache.curator:curator-framework:jar:2.12.0:compile
[INFO] +- org.apache.curator:curator-client:jar:2.12.0:compile
[INFO] +- org.apache.httpcomponents:httpclient:jar:4.5.13:compile
[INFO] |  \- org.apache.httpcomponents:httpcore:jar:4.4.15:compile
[INFO] +- org.apache.zookeeper:zookeeper:jar:3.4.14:compile
[INFO] +- org.slf4j:jcl-over-slf4j:jar:1.7.35:compile
[INFO] +- org.slf4j:slf4j-api:jar:1.7.35:compile
```

the output after this PR:

```diff
 [INFO] org.apache.kyuubi:kyuubi-hive-jdbc:jar:1.7.0-SNAPSHOT
+[INFO] +- org.apache.arrow:arrow-vector:jar:10.0.0:compile
+[INFO] |  +- org.apache.arrow:arrow-format:jar:10.0.0:compile
+[INFO] |  +- org.apache.arrow:arrow-memory-core:jar:10.0.0:compile
+[INFO] |  \- com.google.flatbuffers:flatbuffers-java:jar:1.12.0:compile
+[INFO] +- org.apache.arrow:arrow-memory-netty:jar:10.0.0:compile
+[INFO] |  +- io.netty:netty-buffer:jar:4.1.73.Final:compile
+[INFO] |  \- io.netty:netty-common:jar:4.1.73.Final:compile
+[INFO] +- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.4:compile
+[INFO] +- com.fasterxml.jackson.core:jackson-core:jar:2.13.4:compile
+[INFO] +- com.fasterxml.jackson.core:jackson-databind:jar:2.13.4:compile
+[INFO] +- com.fasterxml.jackson.datatype:jackson-datatype-jsr310:jar:2.13.4:compile
 [INFO] +- org.apache.thrift:libfb303:jar:0.9.3:compile
 [INFO] +- org.apache.thrift:libthrift:jar:0.9.3:compile
 [INFO] +- org.apache.hive:hive-service-rpc:jar:3.1.3:compile
 [INFO] +- commons-codec:commons-codec:jar:1.15:compile
 [INFO] +- org.apache.commons:commons-lang3:jar:3.10:compile
 [INFO] +- com.google.guava:guava:jar:30.1-jre:compile
 [INFO] |  \- com.google.guava:failureaccess:jar:1.0.1:compile
 [INFO] +- com.google.code.findbugs:jsr305:jar:3.0.2:provided
 [INFO] +- org.apache.curator:curator-framework:jar:2.12.0:compile
 [INFO] +- org.apache.curator:curator-client:jar:2.12.0:compile
 [INFO] +- org.apache.httpcomponents:httpclient:jar:4.5.13:compile
 [INFO] |  \- org.apache.httpcomponents:httpcore:jar:4.4.15:compile
 [INFO] +- org.apache.zookeeper:zookeeper:jar:3.4.14:compile
 [INFO] +- org.slf4j:jcl-over-slf4j:jar:1.7.35:compile
 [INFO] +- org.slf4j:slf4j-api:jar:1.7.35:compile
```

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/SparkDataTypeTests.scala
cfmcgrady: updated.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiArrowBasedResultSet.java
cfmcgrady: since the method is forked from `KyuubiBaseResultSet::findColumn`, I prefer to leave it as the current name.

kyuubi-hive-jdbc-shaded/pom.xml
```diff
@@ -118,6 +131,10 @@
                             <pattern>org.apache.zookeeper</pattern>
                             <shadedPattern>${kyuubi.shade.packageName}.org.apache.zookeeper</shadedPattern>
                         </relocation>
+                        <relocation>
+                            <pattern>org.slf4j</pattern>
```
pan3793, line 135: I'm afraid we can not relocate classes under `org.slf4j`, because the relocation break the log framework binding.

kyuubi-hive-jdbc-shaded/pom.xml
```diff
@@ -74,10 +74,15 @@
                                 <exclude>LICENSE.txt</exclude>
                                 <exclude>NOTICE.txt</exclude>
                                 <exclude>mozilla/**</exclude>
+                                <exclude>arrow-git.properties</exclude>
```
pan3793, line 77: `META-INF/native-image/**` should be excluded as well, because the shade plugin can not handle it properly.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -160,16 +161,38 @@ abstract class SparkOperation(session: Session)
...
 
       resultRowSet
     }
 
   override def shouldRunAsync: Boolean = false
+
+  protected def arrowEnabled(): Boolean = {
+    // normalized config is required, to pass unit test
+    session.asInstanceOf[AbstractSession].normalizedConf
+      .getOrElse("kyuubi.session.result.codec", "simple")
```
pan3793, line 191: `kyuubi.operation.result.codec` because
- there are some configurations under namespace `kyuubi.operation.result.`, e.g. `kyuubi.operation.incremental.collect`, `kyuubi.operation.result.max.rows`, we should make the configuration name consistent
- instead of `connection.getResultCodec()`, the result codec can be detected from `TGetResultSetMetadataResp`, e.g. the server can append `__kyuubi_arrow_format__` to to comment of each columns in the schema to indicate that the result is encoded w/ arrow.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -1366,4 +1380,8 @@ public String getEngineName() {
   public String getEngineUrl() {
     return engineUrl;
   }
+
+  public String getResultCodec() {
```
pan3793, line 1384: the result codec should be operation level instead of connection, if you want this to a temporary solution, please avoid using public.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiArrowBasedResultSet.java
pan3793: so we have the opportunity to reuse more code?

kyuubi-hive-jdbc-shaded/pom.xml
cfmcgrady: good catch, reverted.

kyuubi-hive-jdbc-shaded/pom.xml
cfmcgrady: updated.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiArrowBasedResultSet.java
cfmcgrady: yep

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
cfmcgrady: let me try to append the hint to `status#infoMessages` and do the refactor

cfmcgrady: This PR is big enough already, I think we'll rename it to `kyuubi.operation.result.codec` first and refactor it in a separate PR, what do you think?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
cfmcgrady: the same with https://github.com/apache/incubator-kyuubi/pull/3803#discussion_r1033112667

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
pan3793: > rename it to `kyuubi.operation.result.codec` first ...

sgtm

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
pan3793: it's fine, but can we use package instead of public here?

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #3838, opened 2022-11-23
[KYUUBI #3839][Authz] Introduce signature mechanism to protect session variable on engine side

### LINKED ISSUE #3839: [FEATURE] [Authz] Introduce signature mechanism to protect session variable on engine side
### Code of Conduct

### Search before asking

### Describe the feature

To introduce asymmetric encryption signing to prevent manipulation on user identity, which could cause privilege leaking in scripts.

- Authz plugin relies on `kyuubi.session.user` local property of the spark session to identify end-user for authorization checking
- `kyuubi.session.user` could be easily changed in scala or python script
- use SHA256withECDSA to sign and verify the integrity of `kyuubi.session.user`, by generating a signature with the private key on the server side and verifying sign with a public key on the Spark engine side.
The session signing feature is introduced with asymmetric encryption to prevent manipulation of session user identity or other key session elements, which could cause privilege leaking in scripts.

1. Server: Server creates and holds the singleton `keypair` for session signing if feature enabled
2. Server -> Engine: Server passes the `public key` to Engine when launching
3. Server -> Engine: Server generates session signing on `session user` when opening Kyuubi Session to Engine
4. Session -> Statement: Kyuubi session create Statement with context bringing `publickey` and `session user signature`
5. Engine: Engine verify `session user signature` with signature wherever necessary, e.g. in Authz for session name authentication.

"ECDSA", with a shorter key length and better performance than RSA, is the supported asymmetric encryption in the initial implementation, which is widely supported on JDK7+ in HotSpot or OpenJDK. The session signature is generated with `SHA256withECDSA`.

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

bowenliang123:

to close #3839 .

The session signing feature is introduced with asymmetric encryption to prevent manipulation of session user identity or other key session elements, which could cause privilege leaking in scripts.

1. Server: Server creates and holds the singleton `keypair` for session signing if feature enabled
2. Server -> Engine: Server passes the `public key` to Engine when launching
3. Server -> Engine: Server generates session signing on `session user` when opening Kyuubi Session to Engine
4. Session -> Statement: Kyuubi session create Statement with context bringing `publickey` and `session user signature`
5. Engine: Engine verify `session user signature` with signature wherever necessary, e.g. in Authz for session name authentication.

ECDSA, with a shorter key length and better performance than RSA, is the supported asymmetric encryption in the initial implementation, which is widely supported on JDK7+ in HotSpot or OpenJDK. The session signature is generated with `SHA256withECDSA`.


bowenliang123:
Linked to issue #3839 .

bowenliang123:
cc @yaooqinn @turboFei @pan3793

bowenliang123:
also cc @cfmcgrady

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
```diff
@@ -74,6 +85,13 @@ private[authz] object AuthZUtils {
   def getAuthzUgi(spark: SparkContext): UserGroupInformation = {
     // kyuubi.session.user is only used by kyuubi
     val user = spark.getLocalProperty("kyuubi.session.user")
+    if (StringUtils.isNotBlank(user)) {
+      if (!verifyKyuubiSessionUser(spark, user)) {
+        throw new AccessControlException(
```
pan3793, line 90: we can move the `throw new AccessControlException` into `verifyKyuubiSessionUser` to make the caller side clearer.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -70,10 +72,20 @@ class ExecutePython(
       worker.runCode(s"spark.sparkContext.setLocalProperty('$key', '$value')")
     }
     try {
+
```
pan3793, line 75: nit: unnecessary new line

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -73,11 +76,24 @@ abstract class SparkOperation(session: Session)
...
       spark.sparkContext.setJobGroup(statementId, redactedStatement, forceCancel)
       spark.sparkContext.setLocalProperty(KYUUBI_SESSION_USER_KEY, session.user)
       spark.sparkContext.setLocalProperty(KYUUBI_STATEMENT_ID_KEY, statementId)
+
+      if (isSessionUserVerifyEnabled) {
+        val (publicKey, privateKey) = SignUtils.generateKeyPair
+        val signed = SignUtils.signWithECDSA(session.user, privateKey)
+        val publicKeyStr = Base64.getEncoder.encodeToString(publicKey.getEncoded)
+        spark.sparkContext.setLocalProperty(KYUUBI_SESSION_USER_PUBIC_KEY, publicKeyStr)
+        spark.sparkContext.setLocalProperty(KYUUBI_SESSION_USER_SIGN, signed)
```
pan3793, line 94: seems you forget to clear the local properties

kyuubi-common/src/main/scala/org/apache/kyuubi/util/SignUtils.scala
```diff
@@ -0,0 +1,45 @@
...
+import java.nio.charset.StandardCharsets
+import java.security.{KeyPairGenerator, PrivateKey, PublicKey, SecureRandom, Signature}
+import java.security.spec.ECGenParameterSpec
+import java.util.Base64
+
+object SignUtils {
+
+  private lazy val ecKeyPairGenerator = {
+    val g = KeyPairGenerator.getInstance("EC")
+    g.initialize(new ECGenParameterSpec("secp256k1"), new SecureRandom())
```
pan3793, line 28: tip, using `Random` w/ a random seed generates indeterminate results, it's secure and should be used for production, but makes the test not repeatable, better to allow setting the seed in UT.

pan3793 (commented):
Thanks for your effort on security, left some comments. and cc @turboFei as you do lots of security work on Kyuubi in the past.

turboFei:
Could you add some ut？
Shall we redact the key for en/decryption？

bowenliang123:
> Could you add some ut？ Shall we redact the key for en/decryption？

OK, let me mark this as WIP. Will add some ut soon.

docs/deployment/settings.md
```diff
@@ -497,6 +497,7 @@ kyuubi.session.idle.timeout|PT6H|session idle timeout, it will be closed when it
 kyuubi.session.local.dir.allow.list||The local dir list that are allowed to access by the kyuubi session application. User might set some parameters such as `spark.files` and it will upload some local files when launching the kyuubi engine, if the local dir allow list is defined, kyuubi will check whether the path to upload is in the allow list. Note that, if it is empty, there is no limitation for that and please use absolute path list.|seq|1.6.0
 kyuubi.session.name|&lt;undefined&gt;|A human readable name of session and we use empty string by default. This name will be recorded in event. Note that, we only apply this value from session conf.|string|1.4.0
 kyuubi.session.timeout|PT6H|(deprecated)session timeout, it will be closed when it's not accessed for this duration|duration|1.0.0
+kyuubi.session.user.verify.enabled|true|Whether to verify the integrity of session user name in Spark engine within Authz plugin.|boolean|1.7.0
```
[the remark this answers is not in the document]
bowenliang123: OK, changed the config name to `kyuubi.session.user.sign.enabled`.

As for the rest configs suggested, they are not good to put in configs this time maybe.
- algorithm , 
        - not suitable config in one option, itis related to different details including the KeyPair generation algorithm (like "EC"), and the signing algorithm (like "SHA256withECDSA")
        - used by both the server and engine side plugin Authz
- public.key / private.key
        - set to spark conf , may be show on Spark UI, even being redacted
        - if set to spark conf via `SparkProcBuilder`, they will be easily leaking to using scripts for regenearting fake  session user which passing the use sign verifying in Authz.
 ```
"""
privatekey = spark.context.getconf.get("privatekey")

newsign = sign(fakeuser, privatekey)
spark.context.setLocalProp(fakeuser)

# invalid access with fake session user
df = spark.sql("select * from confidential_table")
 ```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
```diff
@@ -20,6 +20,8 @@ package org.apache.kyuubi.config
 object KyuubiReservedKeys {
   final val KYUUBI_CLIENT_IP_KEY = "kyuubi.client.ipAddress"
   final val KYUUBI_SESSION_USER_KEY = "kyuubi.session.user"
+  final val KYUUBI_SESSION_USER_SIGN = "kyuubi.session.user.sign"
+  final val KYUUBI_SESSION_USER_PUBIC_KEY = "kyuubi.session.user.public.key"
```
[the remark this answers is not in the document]
bowenliang123: fixed.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
bowenliang123: Good point, thx.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
bowenliang123: It's alright to leave these local properties set, as they are generated and refreshed in every statement execution.
But it's okay to clear them as well. Changes updated.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -74,6 +74,16 @@ class ExecutePython(
...
       setSparkLocalProperties(KYUUBI_SESSION_USER_KEY, session.user)
       setSparkLocalProperties(KYUUBI_STATEMENT_ID_KEY, statementId)
+
+      if (isSessionUserVerifyEnabled) {
+        setSparkLocalProperties(
+          KYUUBI_SESSION_SIGN_PUBLICKEY,
+          session.conf.get(KYUUBI_SESSION_SIGN_PUBLICKEY).orNull)
+        setSparkLocalProperties(
+          KYUUBI_SESSION_USER_SIGN,
+          session.conf.get(KYUUBI_SESSION_USER_SIGN).orNull)
```
pan3793, line 84: If `isSessionUserVerifyEnabled`, `KYUUBI_SESSION_SIGN_PUBLICKEY` `KYUUBI_SESSION_USER_SIGN` must be present?

pan3793:
Overall this LGTM, could you please add test cases on server side? I think it should include
- positive case
- sign enabled can not disable by session configuration
- unmatched user and sign
- sign enabled but no sign in configuration

kyuubi-common/src/main/scala/org/apache/kyuubi/util/SignUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+    g
+  }
+
+  def generateKeyPair(algorithm: String = "EC"): KeyPair = {
```
pan3793, line 32: It's fine to return `(PublicKey, PrivateKey)`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
bowenliang123: Yes, these two configs are set to session via `configOverlay`  in `OpenSession` on server side , when session sign enabled.

kyuubi-common/src/main/scala/org/apache/kyuubi/util/SignUtils.scala
bowenliang123: OK, it's better.

bowenliang123:
With the session user `bowen2`, querying with pyspark.
In Authz plugin on engine side, the session user `bowen2` are verified with public key generated on server side and checked for table privileges with session user.

- session user verified, no access privileges to the table:
image

- session user disguised as `superadmin` directly in SparkContext in user code (multi-line scripts for single execution), session user `superadmin`  is not verified:
image

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
```diff
@@ -74,6 +85,10 @@ private[authz] object AuthZUtils {
   def getAuthzUgi(spark: SparkContext): UserGroupInformation = {
     // kyuubi.session.user is only used by kyuubi
     val user = spark.getLocalProperty("kyuubi.session.user")
+    if (StringUtils.isNotBlank(user)) {
```
pan3793, line 88: if `kyuubi.session.user.sign.enabled` is true and `kyuubi.session.user` is absent, we should fail w/ `AccessControlException`

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
```diff
@@ -175,4 +190,38 @@ private[authz] object AuthZUtils {
...
+      if (!isVerified) {
+        throw new AccessControlException(s"Permission denied: user [$user] is not verified")
+      }
+    }
+  }
+
+  private def verifySignWithECDSA(
+      plainText: String,
+      signature: String,
+      publicKeyStr: String): Boolean = {
```
pan3793, line 217: publicKeyBase64

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
```diff
@@ -175,4 +190,38 @@ private[authz] object AuthZUtils {
...
+
+  private def verifyKyuubiSessionUser(spark: SparkContext, user: String): Unit = {
+    val isSessionUserVerifyEnabled =
+      spark.getConf.getBoolean(s"spark.kyuubi.session.user.sign.enabled", defaultValue = false)
+    if (isSessionUserVerifyEnabled) {
+      val isVerified = {
+        try {
+          val userPubKeyStr = spark.getLocalProperty("kyuubi.session.sign.publickey")
+          val userSign = spark.getLocalProperty("kyuubi.session.user.sign")
+          verifySignWithECDSA(user, userSign, userPubKeyStr)
```
pan3793, line 202: `userSign` and `userPubKeyStr` may be null

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
pan3793: my point is should we fail fast instead of `.orNull`?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
bowenliang123: Yes, let's check them not null.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
bowenliang123: OK. If `kyuubi.session.user.sign.enabled` is enabled, force to check session user.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/AuthZUtils.scala
```diff
@@ -175,4 +194,38 @@ private[authz] object AuthZUtils {
...
+        } else {
+          verifySignWithECDSA(user, userSign, userPubKeyBase64)
+        }
+      } catch {
+        case _: Exception =>
+          false
+      }
+    }
+    if (!isVerified) {
+      throw new AccessControlException(s"Permission denied: user [$user] is not verified")
```
pan3793, line 214: "Permission denied" is not suitable here, it's some kind of "illegal identifier" or "invalid user identifier"

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -89,6 +104,10 @@ abstract class SparkOperation(session: Session)
       spark.sparkContext.setLocalProperty(SPARK_SCHEDULER_POOL_KEY, null)
       spark.sparkContext.setLocalProperty(KYUUBI_SESSION_USER_KEY, null)
       spark.sparkContext.setLocalProperty(KYUUBI_STATEMENT_ID_KEY, null)
+      if (isSessionUserSignEnabled) {
+        spark.sparkContext.setLocalProperty(KYUUBI_SESSION_SIGN_PUBLICKEY, null)
+        spark.sparkContext.setLocalProperty(KYUUBI_SESSION_USER_SIGN, null)
+      }
```
pan3793, line 110:
```suggestion
      spark.sparkContext.setLocalProperty(KYUUBI_SESSION_SIGN_PUBLICKEY, null)
      spark.sparkContext.setLocalProperty(KYUUBI_SESSION_USER_SIGN, null)
```

pan3793 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -193,6 +208,25 @@ abstract class SparkOperation(session: Session)
     // TODO: (fchen) make all operation support arrow
     getClass.getCanonicalName == classOf[ExecuteStatement].getCanonicalName
   }
+
+  protected def setSessionUserSign(): Unit = {
+    (
+      session.conf.get(KYUUBI_SESSION_SIGN_PUBLICKEY),
+      session.conf.get(KYUUBI_SESSION_USER_SIGN)) match {
+      case (Some(pubKey), Some(userSign)) =>
+        setSparkLocalProperty(KYUUBI_SESSION_SIGN_PUBLICKEY, pubKey)
```
cfmcgrady, line 217: >  Server -> Engine: Server generates session signing on session user when opening Kyuubi Session to Engine

Since each session user has its own `KYUUBI_SESSION_USER_SIGN`, but all the user sessions within one application share one `SparkContext`, this means the verification mechanism only works for CONNECTION/USER mode?

bowenliang123: It should work for SERVER share level as well.
1. setting properties to SparkContext is to put value in a ThreadLocal prop holders, not sharing with other threads.
2. `runInternal` is running  the code in synchronized, which means  `setSessionUserSign` and `claeaSessionUserSign` properly set/clear the properties in the same thread  right before/after the code excuted

cfmcgrady (approved):

turboFei (approved):

pan3793:
Thanks @bowenliang123 great work! Merged to master

bowenliang123:
> Thanks @bowenliang123 great work! Merged to master

Many thanks for the support, help, and discussion from all the reviewers!
Hope this PR will bring general security improvement in current and future expected scenarios, especially for dynamic scripting like Scala, Python and etc, which is greatly in demand for dynamic flexibility while privileges are guarded at the front line.

bowenliang123:
This is also applied to pyspark support. https://github.com/apache/kyuubi/issues/3780

PR #3847, opened 2022-11-23
Add jdbc-shaded profile to support IDE debug

cxzl25:
1. `mvn antrun:run@build-info -pl kyuubi-common`
2. debug engine suite


pan3793:
How about using `kyuubi-hive-jdbc` in default and switch to `kyuubi-hive-jdbc-shaded` in CI?

kyuubi-common/src/main/scala/org/apache/kyuubi/package.scala
```diff
@@ -30,7 +30,8 @@ package object kyuubi {
     if (buildFileStream == null) {
       throw new KyuubiException(s"Can not load the core build file: $buildFile, if you meet " +
         s"this exception when running unit tests " +
-        s"please make sure you have run the `mvn package` or `mvn generate-resources` command.")
+        s"please make sure you have run the `mvn package` or " +
+        s"`mvn antrun:run@build-info -pl kyuubi-common` command.")
```
cxzl25, line 34:
```
Caused by: org.apache.kyuubi.KyuubiException: Can not load the core build file: kyuubi-version-info.properties, if you meet this exception when running unit tests please make sure you have run the `mvn package` or `mvn antrun:run@build-info -pl kyuubi-common` command.
	at org.apache.kyuubi.package$BuildInfo$.<init>(package.scala:31)
	at org.apache.kyuubi.package$BuildInfo$.<clinit>(package.scala)
```

cxzl25: It seems that no matter what phase the `maven-antrun-plugin` is configured in, IntelliJ IDEA will not run when building the project, so the `kyuubi-version-info.properties` file will not be generated.

pom.xml
```diff
@@ -2332,6 +2334,13 @@
             </properties>
         </profile>
 
+        <profile>
+            <id>jdbc-use-shaded</id>
```
pan3793, line 2338: jdbc-shaded

pan3793 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/package.scala
cxzl25: If you specify the execution id to run ( `mvn antrun:run@build-info -pl kyuubi-common` ) , you need maven version 3.3.1 or higher.

> MNG-5768 specify execution-id for direct plugin goal invocation from command line
https://issues.apache.org/jira/browse/MNG-5768

Issue #3805, opened 2022-11-15
[GA] `Kyuubi Spark Nightly` failed with PySpark-related tests.


cfmcgrady:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What kind of test improve?

- [ ] Fix flaky tests.
- [ ] Fix bug in tests.
- [ ] Increase test coverage.
- [X] Other kinds of test improve.

### Describe the issues of the existing tests or improvements for new tests

https://github.com/apache/incubator-kyuubi/actions/runs/3458458256/jobs/5772898116

```
- pyspark support *** FAILED ***
  org.apache.kyuubi.jdbc.hive.KyuubiSQLException: SPARK_HOME not found!
  at org.apache.kyuubi.jdbc.hive.Utils.verifySuccess(Utils.java:85)
  at org.apache.kyuubi.jdbc.hive.Utils.verifySuccessWithInfo(Utils.java:71)
  at org.apache.kyuubi.jdbc.hive.KyuubiStatement.runAsyncOnServer(KyuubiStatement.java:272)
  at org.apache.kyuubi.jdbc.hive.KyuubiStatement.executeWithConfOverlay(KyuubiStatement.java:195)
  at org.apache.kyuubi.jdbc.hive.KyuubiStatement.execute(KyuubiStatement.java:190)
  at org.apache.kyuubi.jdbc.hive.KyuubiStatement.executeQuery(KyuubiStatement.java:397)
  at org.apache.kyuubi.engine.spark.operation.PySparkTests.$anonfun$runPySparkTest$1(PySparkTests.scala:66)
  at org.apache.kyuubi.engine.spark.operation.PySparkTests.$anonfun$runPySparkTest$1$adapted(PySparkTests.scala:64)
  at org.apache.kyuubi.operation.JDBCTestHelper.$anonfun$withMultipleConnectionJdbcStatement$3(JDBCTestHelper.scala:60)
  at org.apache.kyuubi.operation.JDBCTestHelper.$anonfun$withMultipleConnectionJdbcStatement$3$adapted(JDBCTestHelper.scala:60)
  ...
  Cause: java.lang.IllegalStateException: SPARK_HOME not found!
[...]

PR #3850, opened 2022-11-24
[KYUUBI #3805][FOLLOWUP] Skipping PySpark tests for nightly build

cfmcgrady:

correct the wrong PySpark tag setting, and make it take effect for `spark-master` only.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2257,7 +2257,7 @@
             <id>spark-master</id>
             <properties>
                 <spark.version>3.4.0-SNAPSHOT</spark.version>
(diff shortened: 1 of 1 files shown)


cfmcgrady:
cc @pan3793 @bowenliang123

bowenliang123:
`PySparkTest` is still left in global setting , what you think whether we should keep this? @cfmcgrady 
https://github.com/apache/incubator-kyuubi/blob/master/pom.xml#L221

bowenliang123 (approved):

cfmcgrady:
> `PySparkTest` is still left in global setting , what you think whether we should keep this? @cfmcgrady https://github.com/apache/incubator-kyuubi/blob/master/pom.xml#L221

it should take effect for `spark-master` only

bowenliang123:
LGTM.
Thanks for the fix. And sorry for my mistakes in last PRs of https://github.com/apache/incubator-kyuubi/pull/3846 and https://github.com/apache/incubator-kyuubi/pull/3832 which failed to correct the whole problem.

pan3793 (approved):

PR #3877, opened 2022-11-29
[KYUUBI #3865] Arrow-based result compression support

### LINKED ISSUE #3865: [Subtask] Arrow-based result compression support
### Code of Conduct

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

Arrow-based result compression support

### Parent issue

#3863 

cfmcgrady:

to close #3865 

Arrow-based result compression support

### Input dataset

```
spark.sql(
  """
    |select * from tpcds.sf10.catalog_sales limit 2000000
    |""".stripMargin)
  .write
  .mode("overwrite")
  .save("/tmp/parquet/tpcds/sf10/catalog_sales_2000000")
```

### Query

cat /tmp/b.sql

```sql
select * from parquet.`/tmp/parquet/tpcds/sf10/catalog_sales_2000000`;
```

### Beeline command

```
bin/beeline -u 'jdbc:hive2://${remote-ip}:10009/;' --hiveconf kyuubi.operation.result.codec=arrow --hiveconf kyuubi.operation.incremental.collect=true --hiveconf kyuubi.operation.result.compression.codec=gzip/lz4/zstd/snappy  -f /tmp/b.sql > /dev/null
```

### Result

| kyuubi.operation.result.compression.codec | time spent (round 1) | time spent (round 2) |
| --- | --- | --- |
| lz4 | 116.563 seconds | 114.017 seconds |
| snappy | 117.992 seconds | 122.061 seconds |
| zstd | 107.774 seconds | 108.832 seconds |
| gzip | 143.361 seconds | 141.129 seconds |
| none | 183.002 seconds| 169.164 seconds |


pan3793:
I suggest remove the snappy support, comparing to other format, I don't see advantage of snappy.

- gzip: high compression ratio, used widely and many languages and platforms have the built-in support.
- lz4: the fastest one, also has pure java implementation.
- zstd: the future-proof one, excellence in both speed and compression ratio.

cfmcgrady:
> I suggest removing the snappy support, comparing to other formats, I don't see advantage of snappy.
> 
> * gzip: high compression ratio, used widely, many languages and platforms have the built-in support.
> * lz4: the fastest one, also has pure java implementation.
> * zstd: the future-proof one, excellence in both speed and compression ratio.

SGTM, and removed `snappy` from the build-in support. End users can easily add `snappy` support as needed.

kyuubi-hive-jdbc/pom.xml
```diff
@@ -168,6 +168,16 @@
...
         </dependency>
+
+        <dependency>
+            <groupId>org.lz4</groupId>
+            <artifactId>lz4-java</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.luben</groupId>
+            <artifactId>zstd-jni</artifactId>
```
pan3793, line 179: shading lz4 and zstd-jni may cause some problems, have you check if they support relocating native libs like netty does?

docs/deployment/settings.md
```diff
@@ -438,6 +438,7 @@ kyuubi.operation.plan.only.output.style|plain|Configures the planOnly output sty
...
 kyuubi.operation.query.timeout|&lt;undefined&gt;|Timeout for query executions at server-side, take affect with client-side timeout(`java.sql.Statement.setQueryTimeout`) together, a running query will be cancelled automatically if timeout. It's off by default, which means only client-side take fully control whether the query should timeout or not. If set, client-side timeout capped at this point. To cancel the queries right away without waiting task to finish, consider enabling kyuubi.operation.interrupt.on.cancel together.|duration|1.2.0
 kyuubi.operation.result.codec|simple|Specify the result codec, available configs are: <ul> <li>SIMPLE: the result will convert to TRow at the engine driver side. </li> <li>ARROW: the result will be encoded as Arrow at the executor side before collecting by the driver, and deserialized at the client side. note that it only takes effect for kyuubi-hive-jdbc clients now.</li></ul>|string|1.7.0
+kyuubi.operation.result.compression.codec|lz4|The codec used to compress arrow-based result. By default, Kyuubi provides three codecs: `lz4`, `zstd`, `gzip`|string|1.7.0
```
bowenliang123, line 441: As `no compression` is also supported in this PR, the `none` value can be added to doc in a way to disable compression.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1457,6 +1457,16 @@ object KyuubiConf {
       .transform(_.toLowerCase(Locale.ROOT))
       .createWithDefault("simple")
 
+  val OPERATION_RESULT_COMPRESSION_CODEC: ConfigEntry[String] =
+    buildConf("kyuubi.operation.result.compression.codec")
```
bowenliang123, line 1461: As the config is used for Arrow-based results, shall we change it to `kyuubi.operation.result.arrow.compression`?

kyuubi-hive-jdbc/pom.xml
bowenliang123: Refering to zstd doc, (http://facebook.github.io/zstd/)
[aircompressor](https://github.com/airlift/aircompressor) implements zstd compression purely in java. https://github.com/airlift/aircompressor/tree/master/src/main/java/io/airlift/compress/zstd

bowenliang123:
> I suggest removing the snappy support, comparing to other formats, I don't see advantage of snappy.
> 
> * gzip: high compression ratio, used widely, many languages and platforms have the built-in support.
> * lz4: the fastest one, also has pure java implementation.
> * zstd: the future-proof one, excellence in both speed and compression ratio.

+1 for these 3 supported compression list, and `lz4` is the best to be the default compression type.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -73,3 +79,53 @@ object SparkDatasetHelper {
...
+      case "zstd" => ZstdCompressionCodec
+      case "gzip" => GZIPCompressionCodec
+      case _ => NoCompressionCodec
+    }
+  }
+}
+
+object NoCompressionCodec extends CompressionCodec {
+  override def outputStream(baos: ByteArrayOutputStream): OutputStream = {
+    throw new UnsupportedOperationException()
```
bowenliang123, line 113: If `no-compression` is not supported, why not just throw this exception in `CompressionCodecFactory`.`createCodec`?

pan3793:
Forwarding informations from the offline discussion:

1. we plan to defer this feature after Spark Connect implemented arrow compression, because there are some different ways(e.g. arrow native, stream compression, block compression) to implement compression, we'd better to keep consistent w/ upstream.
2. [aircompressor](https://github.com/airlift/aircompressor) is good option for client side, the roughly testing shows it has same decompression performance as lz4-java and zstd-jni, but there are some memory allocation issue on lz4 stream decompression.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1708,6 +1708,16 @@ object KyuubiConf {
       .transform(_.toLowerCase(Locale.ROOT))
       .createWithDefault("thrift")
 
+  val OPERATION_RESULT_COMPRESSION_CODEC: ConfigEntry[String] =
+    buildConf("kyuubi.operation.result.compression.codec")
+      .doc("The codec used to compress arrow-based result. By default, Kyuubi provides three " +
+        "codecs: `lz4`, `zstd`, `gzip`")
```
bowenliang123, line 1714: How about listing `none` as an option for disabling compression?

Issue #3875, opened 2022-11-29
[Improvement] Filter out kyuubi.metadata.store.* configurations in engine process builder


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

```
2022-11-29 02:11:48 [INFO] [KyuubiSessionManager-exec-pool: Thread-310] org.apache.kyuubi.engine.EngineRef#59 - Launching engine:
/opt/spark/bin/spark-submit \
	--class org.apache.kyuubi.engine.spark.SparkSQLEngine \
	--conf spark.hive.server2.thrift.resultset.default.fetch.size=1000 \
	--conf spark.kyuubi.client.ipAddress=*.*.*.* \
	--conf spark.kyuubi.engine.credentials=*.*.*.* \
	--conf spark.kyuubi.engine.submit.time=1669659108462 \
	--conf spark.kyuubi.frontend.connection.url.use.hostname=false \
	--conf spark.kyuubi.frontend.protocols=THRIFT_BINARY,REST \
	--conf spark.kyuubi.ha.engine.ref.id=d1bd8b7e-aa08-481e-baf7-c05eddbb81b3 \
	--conf spark.kyuubi.ha.namespace=/kyuubi-k8s-test_1.7.0-SNAPSHOT_CONNECTION_SPARK_SQL/hive/d1bd8b7e-aa08-481e-baf7-c05eddbb81b3 \
	--conf spark.kyuubi.ha.zookeeper.acl.enabled=false \
	--conf spark.kyuubi.ha.zookeeper.namespace=kyuubi-k8s-test \
	--conf spark.kyuubi.ha.zookeeper.publish.configs=false \
	--conf spark.kyuubi.ha.zookeeper.quorum=*.*.*:2182 \
	--conf spark.kyuubi.kubernetes.client.build.preferFromPodEnv=true \
	--conf spark.kyuubi.kubernetes.namespace=spark \
	--conf spark.kyuubi.metadata.store.jdbc.database.schema.init=true \
	--conf spark.kyuubi.metadata.store.jdbc.database.type=mysql \
	--conf spark.kyuubi.metadata.store.jdbc.driver=com.mysql.jdbc.Driver \
[...]

PR #3879, opened 2022-11-29
[KYUUBI #3875] Filter out kyuubi.metadata.store.* configurations in e…

lightning-L:
…ngine process builder

Before modification:
snipaste_2022-11-29_21-16-14

After modification:
Snipaste_2022-11-29_22-20-11

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -213,6 +213,9 @@ object KyuubiConf {
       val updatedMap = new java.util.HashMap[String, ConfigEntry[_]](kyuubiConfEntries)
       updatedMap.put(entry.key, entry)
       kyuubiConfEntries = updatedMap
+      if (entry.serverOnly) {
+        serverOnlyConfEntries += entry
+      }
     }
 
   // For testing only
@@ -2153,7 +2156,7 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
-  private val serverOnlyConfEntries: Set[ConfigEntry[_]] = Set(
+  private var serverOnlyConfEntries: Set[ConfigEntry[_]] = Set(
     FRONTEND_BIND_HOST,
     FRONTEND_BIND_PORT,
     FRONTEND_THRIFT_BINARY_BIND_HOST,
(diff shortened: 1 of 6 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -182,6 +182,10 @@ private[kyuubi] class EngineRef(
 
     conf.set(HA_NAMESPACE, engineSpace)
     conf.set(HA_ENGINE_REF_ID, engineRefId)
+
+    val metadataStoreConf = conf.getAll.filter(x => x._1.startsWith(KYUUBI_METADATA_STORE_PREFIX))
```
pan3793, line 186: I suppose we can add those configurations to `KyuubiConf#serverOnlyConfEntries` to solve this problems.

lightning-L: The metadata sore related configs are defined in `org.apache.kyuubi.server.metadata.jdbc.JDBCMetadataStoreConf._`, which is in kyuubi-server. KyuubConf is in kyuubi-common and has no access to JDBCMetadataStoreConf. Any solution for this?

pan3793: Emm... how about
1. making `serverOnlyConfEntries` mutable
2. adding method `KyuubiConf#registerServerOnlyConfEntry`
3. invoking `KyuubiConf#registerServerOnlyConfEntry` in `JDBCMetadataStoreConf`

lightning-L: what about making `serverOnly` a property for `ConfigEntry` and `ConfigBuilder`? just like `internal`

pan3793: sgtm

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2153,7 +2156,7 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
-  private val serverOnlyConfEntries: Set[ConfigEntry[_]] = Set(
+  private var serverOnlyConfEntries: Set[ConfigEntry[_]] = Set(
     FRONTEND_BIND_HOST,
```
pan3793, line 2160: can we update these config entries as well?

pan3793 (approved):

turboFei (approved):

pan3793:
Thanks, merging to master

PR #3884, opened 2022-12-01
Execute scala code supports asynchronous and query timeout

cxzl25:
Execute scala code now does not support async and query timeout.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteScala.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteScala.scala
@@ -18,6 +18,7 @@
 package org.apache.kyuubi.engine.spark.operation
 
 import java.io.File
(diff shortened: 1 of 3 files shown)


yaooqinn (approved):

turboFei (approved):

turboFei:
cc @cfmcgrady 
Does ExecutePython support query timeout?

cfmcgrady:
> cc @cfmcgrady Does ExecutePython support query timeout?

No, it's not supported right now.

Meanwhile, I have a question, if it is not a spark job, can it be canceled correctly if the operation times out?

cxzl25:
> Meanwhile, I have a question, if it is not a spark job, can it be canceled correctly if the operation times out?

If the Scala code submits a Spark job, the query times out at this time, the thread will be interrupted, and the corresponding job will be canceled through the `JobGroupId`.

If Scala code does not submit the Spark job, it is just some scala. When the code does not respond to interrupt, it will not be really cancelled. Just like Spark executor kill task, it just tries to interrupt thread.

turboFei:
thanks, merged to master

PR #3886, opened 2022-12-02
Bump hive service rpc version to 4.0.0-alpha-2

turboFei:

It supports
- UploadData and DownloadData API.
- TGetOperationStatusResp::getNumModifiedRows


pan3793:
I think it will be part of the final Hive 4.0 thrift protocol, but "alpha" looks too early, we may want to wait for a "beta" or "rc" version then get it in, so I suggest pending it a well. Could you provide the implementation as well?

PR #3891, opened 2022-12-02
[KYUUBI #3890] [Improvement] Prefer `binPack` style for importSelectors in scalafmt styling

### LINKED ISSUE #3890: [Improvement] Prefer `binPack` style for importSelectors in scalafmt styling
### Code of Conduct

### Search before asking

Update `importSelectors = singleLine` to `importSelectors = binPack` in `.scalafmt.conf`.

With about 114 occurrences, the currently used `singleLine` for `importSelectors` is violated with another config `maxColumn=100` by forcing all other import from one source formated in single line no matter how long it is.

For example, in  `v2Commands.scala` line 31 has the length with over 202 chars in one line.
And in its case, it has to importing in static way one by one , or it will break if import all the elements and methods of `import org.apache.kyuubi.plugin.spark.authz.v2Commands.CommandType.CommandType` 

(https://github.com/apache/incubator-kyuubi/blob/master/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/v2Commands.scala#L31)

Scalafmt doc for importselectors:
https://scalameta.org/scalafmt/docs/configuration.html#importselectors

### How should we improve?

_No response_

bowenliang123:

to close #3890 .

Change `importSelectors = singleLine` to `importSelectors = binPack` in `.scalafmt.conf`.


turboFei:
emmm, but it looks ugly for scala code. sorry for the -1

bowenliang123:
Let me try to explain this PR in some more reasons,
1. `singleLine` style for imports does force the line of long imports to break `maxColumn` rule, which is not self-explicable and brings worse readability in practical use
2. long imports are unavoidable in some cases, or using `import any._` with wildcard will cause cyclic dependencies (e.g. v2commands.scala as the example the issue mentioned)
3. `binPack` style won't impact the multi-imports which is not over-sized
4. `binPack` sytle is the original capability from `scalafmt`, and wont introduced extra mannual workload for auto linting and reformat using IDE or tools.

bowenliang123:
What do you think. @pan3793 @yaooqinn 
Which is more important self-explicable rules and readability in single screen VS single line for minimum line breaks?

turboFei:
> 1. `singleLine` style for imports does force the line of long imports to break `maxColumn` rule, which is not self-explicable and brings worse readability in practical use

1. For the scala code style setting, it allows to ignore the import part.
```
  <check level="error" class="org.scalastyle.file.FileLineLengthChecker" enabled="true">
    <parameters>
      <parameter name="maxLineLength"><![CDATA[100]]></parameter>
      <parameter name="tabSize"><![CDATA[2]]></parameter>
      <parameter name="ignoreImports">true</parameter>
    </parameters>
  </check>
```
2. `import` is not code, so it is reasonable to limit the import length.
3. for a corner case, if my code package depth is big, how do you prevent it to exceed the maxLineLength?

turboFei:
> 2.long imports are unavoidable in some cases, or using import any._ with wildcard will cause cyclic dependencies (e.g. v2commands.scala as the example the issue mentioned)

I wonder that you can split the multiple imports into two lines manually?

FYI:

Before:
```
import org.apache.kyuubi.plugin.spark.authz.v2Commands.CommandType.{CommandType, HasChildAsIdentifier, HasQueryAsLogicalPlan, HasTableAsIdentifier, HasTableAsIdentifierOption, HasTableNameAsIdentifier}
```
After:
```
import org.apache.kyuubi.plugin.spark.authz.v2Commands.CommandType.{CommandType, HasChildAsIdentifier, HasQueryAsLogicalPlan}
import org.apache.kyuubi.plugin.spark.authz.v2Commands.CommandType.{HasTableAsIdentifier, HasTableAsIdentifierOption, HasTableNameAsIdentifier}
```


image

turboFei:
> 3. binPack style won't impact the multi-imports which is not over-sized
It looks not good for this case, at least, I think we should not limit the import length with the same size of code.
image

pan3793:
I prefer to keep it as-is.

bowenliang123:
OK, closing this PR. Based on the discussion, the advantage achieved by binpack style for important is not significant and not part of consensu.
Thanks for the reviews.

PR #3897, opened 2022-12-03
Supplying pluggable GroupProvider

pan3793:

Kyuubi supports GROUP engine share level, currently, it just simply delegates the group provider to Hadoop UserGroupInformation, which is not flexible enough for users who want to use other group mapping mechanisms, e.g. LDAP, JDBC.

This PR supplies a pluggable plugin interface `GroupProvider` and provides a built-in `HadoopGroupProvider` which has the same behavior w/ the current implementation.

W/ this change, users can easily implement `LDAPGroupProvider`, `JDBCGroupProvider`, `FileGroupProvider`, `CustomGroupProvider`, etc. then the GROUP engine share level will be more powerful and flexible.

The alternative option is to guide users to learn and extend the Hadoop group mapping system[1].

[1] https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/GroupsMapping.html


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1977,6 +1977,21 @@ object KyuubiConf {
...
+    buildConf("kyuubi.session.group.provider")
+      .doc("A group provider plugin for Kyuubi Server. This plugin can provide primary group " +
+        "and groups information for different user or session configs. This config value " +
+        "should be a class which is a child of 'org.apache.kyuubi.plugin.GroupProvider' which " +
+        "has zero-arg constructor. Kyuubi provides the following built-in implementations: " +
+        "<li>hadoop: delegate the user group mapping to hadoop UserGroupInformation.</li>")
+      .version("1.7.0")
+      .stringConf
+      .transform {
+        case "hadoop" => "org.apache.kyuubi.session.HadoopGroupProvider"
```
pan3793, line 1990: give the built-in a shortcut

pan3793:
cc @yanghua @jiaoqingbo @cxzl25

yanghua (approved):
+1

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
jiaoqingbo: Whenever a built -in implementation is added, you need to change the code here and re -generate the document. It seems that it is not very friendly to users. Can't you define it like session_conf_advisor?

cxzl25 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
pan3793: We should provide the built-in implementation list in the description, and provide a shortcut like "hadoop", "ldap", "unix" for convenient, but allow users to set the full class name for third party implementation.
I'm planning to change to the `SESSION_CONF_ADVISOR`'s documentation to align w/ this way, because it does not clear which built-in implementations are available.

pan3793:
Thanks all for review, merging to master

PR #3898, opened 2022-12-04
Init operation log for ExecutePython to prevent error when fetching log

turboFei:

To prevent the error as below:
image


cfmcgrady (approved):

turboFei:
thanks, merging to master

bowenliang123:
Good, it's also another known bug. Thanks for the effort.

PR #3900, opened 2022-12-05
Fallback to SQL language mode if fail to init python environment and fix log

turboFei:

Now, if failed to init python environment, we have to re-create the connection.

In this pr, it will fallback to SQL language mode for above case.

image

image


cfmcgrady (approved):
LGTM

turboFei:
thanks, merged to master

PR #3913, opened 2022-12-06
[KYUUBI #3385] Set executor pod name prefix if missing in spark on k8s case

zwangsheng:

See more in #3590
For #3590 this PR is reverted, and author not reply for long time.
Fix this issue #3385 here.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -18,6 +18,7 @@
 package org.apache.kyuubi.engine.spark
 
 import java.time.Instant
+import java.util.UUID
 import java.util.concurrent.{CountDownLatch, ScheduledExecutorService, ThreadPoolExecutor, TimeUnit}
 import java.util.concurrent.atomic.AtomicBoolean
 
@@ -192,6 +193,10 @@ object SparkSQLEngine extends Logging {
 
     if (Utils.isOnK8s) {
       kyuubiConf.setIfMissing(FRONTEND_CONNECTION_URL_USE_HOSTNAME, false)
+
+      _sparkConf.setIfMissing(
+        "spark.kubernetes.executor.podNamePrefix",
+        generateExecutorPodNamePrefixForK8s(user))
     }
 
     // Set web ui port 0 to avoid port conflicts during non-k8s cluster mode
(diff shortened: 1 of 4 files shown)


externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -71,6 +71,12 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
```
yaooqinn, line 74: Hmm?for what?

zwangsheng: Still in progress 😢

ulysses-you:
shall we do this in `EngineRef` ? I remember `SparkProcessBuilder` also do validation for pod name.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/SparkSQLEngineSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.spark
+
+import org.apache.kyuubi.KyuubiFunSuite
+
+class SparkSQLEngineSuite extends KyuubiFunSuite {
+  private val EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH = 237
```
pan3793, line 23: why define again?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -342,4 +351,23 @@ object SparkSQLEngine extends Logging {
...
+
+  @VisibleForTesting
+  def generateExecutorPodNamePrefixForK8s(userName: String): String = {
+    val resolvedUserName =
+      userName.trim.toLowerCase(Locale.ROOT)
+        .replaceAll("[^a-z0-9\\-]", "-")
+        .replaceAll("-+", "-")
+        .replaceAll("^-", "")
+    val podNamePrefixWithUser = s"kyuubi-$resolvedUserName-${Instant.now().toEpochMilli}"
+    if (podNamePrefixWithUser.length >= EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH) {
```
pan3793, line 363: = is safe

pan3793 (approved):
lgtm, only minor comments

zwangsheng:
> shall we do this in `EngineRef` ? I remember `SparkProcessBuilder` also do validation for pod name.

For validation, it looks more like to help fail fast on the Kyuubi side and is not supposed to help configure spark conf.

We do a lot of operations on `SparkSQLEngine` to help users optimize their configuration, so it makes sense to put this here.

pan3793:
Thanks, merging to master

PR #3916, opened 2022-12-06
[KYUUBI #3915] Client support detecting ResultSet codec

### LINKED ISSUE #3915: [Subtask] Client support detecting ResultSet codec
### Code of Conduct

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

Client support detecting ResultSet codec

### Parent issue

#3863 

cfmcgrady:

to close #3915

This pr adds support for jdbc client detecting result set codec

1. in this PR, hints are added in the `TStatus.getInfoMessages()` to return, and the hints were added when the client retrieves the result set schema from the server
2. the hints mechanism is a general extension when we need to change the client behavior, e.g. adding support for  result set compression

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
@@ -26,7 +26,6 @@ import org.apache.kyuubi.jdbc.hive.adapter.SQLStatement;
 import org.apache.kyuubi.jdbc.hive.cli.FetchType;
 import org.apache.kyuubi.jdbc.hive.cli.RowSet;
 import org.apache.kyuubi.jdbc.hive.cli.RowSetFactory;
-import org.apache.kyuubi.jdbc.hive.common.HiveDecimal;
 import org.apache.kyuubi.jdbc.hive.logs.InPlaceUpdateStream;
 import org.apache.kyuubi.jdbc.hive.logs.KyuubiLoggable;
 import org.apache.thrift.TException;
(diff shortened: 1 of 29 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -743,4 +763,57 @@ public String getQueryId() throws SQLException {
...
+    List<String> infoMessages = metadataResp.getStatus().getInfoMessages();
+    if (infoMessages != null) {
+      metadataResp
+          .getStatus()
+          .getInfoMessages()
+          .forEach(
+              line -> {
+                String[] keyValue = line.split("=");
+                String key = keyValue[0];
+                String value = keyValue[1];
```
pan3793, line 803: will cause NPE if server return some message does not contains '='

pan3793 (approved):
LGTM, only minor comments

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
cfmcgrady: Log a warning message if the input does not contain "="

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -796,15 +796,15 @@ private void parseMetadata(
       metadataResp
           .getStatus()
           .getInfoMessages()
+          .stream()
+          .filter(hint -> Utils.isKyuubiOperationHint(hint))
           .forEach(
               line -> {
                 String[] keyValue = line.split("=");
                 if (keyValue.length == 2) {
                   String key = keyValue[0];
```
pan3793, line 805: since the pattern is CASE_INSENSITIVE, we need to normalize the key here, e.g. to lower/upper case, and the split should always reserve 2 parts, then assert is more suitable here.

pan3793 (approved):

pan3793:
Thanks, merging to master

PR #3923, opened 2022-12-07
Add dedicated batch session idle timeout and prevent the batch submission operation timeout

turboFei:

Prevent batch session & operation closed by kyuubi because of timeout

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
@@ -296,9 +296,8 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
         val current = System.currentTimeMillis
         if (!shutdown) {
           for (session <- handleToSession.values().asScala) {
-            val sessionIdleTimeout = session.sessionIdleTimeoutThreshold
-            if (session.lastAccessTime + sessionIdleTimeout <= current &&
-              session.getNoOperationTime > sessionIdleTimeout) {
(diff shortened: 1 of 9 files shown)


ulysses-you:
would batch session leak if we disable timeout ?

turboFei:
> would batch session leak if we disable timeout ?
It will timeout after the operation terminated.

kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -288,6 +288,19 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
...
+          warn(s"Error closing idle session ${session.handle}", e)
+      }
+    } else {
+      session.closeExpiredOperations()
+    }
+  }
+
   private def startTimeoutChecker(): Unit = {
     val interval = conf.get(SESSION_CHECK_INTERVAL)
     val timeout = conf.get(SESSION_IDLE_TIMEOUT)
```
ulysses-you, line 306: How about add a new filed `sessionTimeout` in `Session`. Use  `SESSION_IDLE_TIMEOUT ` by default in `AbstractSession`, and use `BATCH_SESSION_IDLE_TIMEOUT` in `KyuubiBatchSessionImpl`.

Then we can make each session use it's own timeout threshold.

turboFei: thanks, will address it

ulysses-you (approved):

turboFei:
thanks, merged to master and 1.6.2

PR #3949, opened 2022-12-08
Print the simple batch report if no log fetch to prevent misunderstand

turboFei:

With yarn-cluster mode, if `spark.yarn.submit.waitAppCompletion` is false and `kyuubi-ctl` wait completion is true, we need still print some simple batch report to prevent misunderstand.

The batch report format refer spark-submit log.
```
Application report for <ApplicationID> (state: RUNNING)
```

### DIFF
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
@@ -83,8 +83,8 @@ class LogBatchCommand(
           if (!done) {
             if (!Option(log).exists(_.getRowCount > 0)) {
               Option(batch).foreach { batch =>
-                info(s"Batch report for ${batch.getId} (appId: ${batch.getAppId}," +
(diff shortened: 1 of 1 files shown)


kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala
```diff
@@ -81,6 +81,11 @@ class LogBatchCommand(
           }
 
           if (!done) {
+            if (!Option(log).exists(_.getRowCount > 0)) {
+              Option(batch).foreach { batch =>
+                info(s"Batch report for ${batch.getId} (state: ${batch.getState})")
```
pan3793, line 86: How about print applicationId as well?

turboFei: done

pan3793 (approved):

pan3793 (approved):

turboFei:
thanks, merged to master and 1.6.2

PR #3950, opened 2022-12-08
Fix the batch metadata in-consistent issue on open batch session failure

turboFei:

I found that, there is some confused batch metadata.

The metadata state in mysql is PENDING, but the batch session state is `ERROR`.

I think the RC is that:
- the metadata insert is in KyuubiBatchSessionImpl
- If open batch session failed, the BatchJobSubmission opHandle has not been added into batch session operation set.
- then the close session will not update the batch metadata

In this pr, I re-order the code when opening batch session.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -285,6 +285,12 @@ class BatchJobSubmission(
       }
 
       MetricsSystem.tracing(_.decCount(MetricRegistry.name(OPERATION_OPEN, opType)))
+
+      if (!builder.processLaunched) {
+        setState(OperationState.CANCELED)
+        updateBatchMetadata()
+      }
+
       // fast fail
       if (isTerminalState(state)) {
         killMessage = (false, s"batch $batchId is already terminal so can not kill it.")
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -155,27 +155,6 @@ class BatchJobSubmission(
   }
 
   override protected def runInternal(): Unit = session.handleSessionException {
-    if (session.recoveryMetadata.isEmpty) {
-      val metaData = Metadata(
-        identifier = batchId,
-        sessionType = session.sessionType,
-        realUser = session.realUser,
-        username = session.user,
-        ipAddress = session.ipAddress,
-        kyuubiInstance = session.connectionUrl,
-        state = OperationState.PENDING.toString,
-        resource = resource,
-        className = className,
-        requestName = batchName,
-        requestConf = session.normalizedConf,
-        requestArgs = batchArgs,
-        createTime = session.createTime,
-        engineType = batchType,
-        clusterManager = builder.clusterManager())
-
-      session.sessionManager.insertMetadata(metaData)
-    }
-
     val asyncOperation: Runnable = () => {
       try {
         if (recoveryMetadata.exists(_.peerInstanceClosed)) {
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -287,8 +287,13 @@ class BatchJobSubmission(
       MetricsSystem.tracing(_.decCount(MetricRegistry.name(OPERATION_OPEN, opType)))
 
       if (!builder.processLaunched) {
+        builder.close()
+        if (recoveryMetadata.isDefined) {
+          killMessage = killBatchApplication()
+        }
         setState(OperationState.CANCELED)
         updateBatchMetadata()
+        return
       }
 
       // fast fail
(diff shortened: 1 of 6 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -80,7 +80,7 @@ class KyuubiSessionImpl(
     sessionManager.groupProvider.primaryGroup(user, optimizedConf.asJava),
     handle.identifier.toString,
     sessionManager.applicationManager)
-  private[kyuubi] val launchEngineOp = sessionManager.operationManager
+  private[kyuubi] lazy val launchEngineOp = sessionManager.operationManager
```
turboFei, line 83: prevent the OP_CONN metrics increase before openOpration(launchEngineOp)

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -113,6 +113,11 @@ class KyuubiBatchSessionImpl(
       ms.incCount(MetricRegistry.name(CONN_OPEN, user))
     }
 
+    // have to add batchJobSubmissionOp into session opHandleSet here, to update the batch metadata
+    // on openSession failure(may be open session for recovery), otherwise, the batch metadata state
+    // might lose control.
+    opHandleSet.add(batchJobSubmissionOp.getHandle)
```
ulysses-you, line 119: does `runOperation(batchJobSubmissionOp)` not work ?

turboFei: it might fail before `runOperation(batchJobSubmissionOp)`

turboFei: let me try to reorder them

turboFei: it should be fine now.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -113,6 +113,11 @@ class KyuubiBatchSessionImpl(
       ms.incCount(MetricRegistry.name(CONN_OPEN, user))
     }
 
+    checkSessionAccessPathURIs()
+
+    // we should call super.open before running batch job submission operation
```
pan3793, line 118: the comment explained what the code does, but not WHY should we do that, it's more important.

turboFei: updated

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -113,6 +114,17 @@ class KyuubiBatchSessionImpl(
...
     }
 
+    try {
+      checkSessionAccessPathURIs()
+      // create the operation root directory before running batch job submission operation
+      super.open()
+    } catch {
+      case e: Throwable if recoveryMetadata.isDefined =>
+        error(s"Error opening recovered batch session $handle", e)
+        Utils.tryLogNonFatalError(batchJobSubmissionOp.close())
```
ulysses-you, line 124: shall we close `batchJobSubmissionOp` inside `KyuubiBatchSessionImpl.close` ?

turboFei: sounds good

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -285,6 +285,12 @@ class BatchJobSubmission(
       }
 
       MetricsSystem.tracing(_.decCount(MetricRegistry.name(OPERATION_OPEN, opType)))
+
+      if (!builder.processLaunched) {
+        setState(OperationState.CANCELED)
+        updateBatchMetadata()
```
ulysses-you, line 291: close and return

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -285,6 +285,17 @@ class BatchJobSubmission(
       }
 
       MetricsSystem.tracing(_.decCount(MetricRegistry.name(OPERATION_OPEN, opType)))
+
+      if (!builder.processLaunched) {
+        builder.close()
+        if (recoveryMetadata.isDefined) {
+          killMessage = killBatchApplication()
```
ulysses-you, line 292: what' the difference in recovery mode ? there can not exist application, no ?

turboFei: to keep the state align, we have to kill the running application for recovery mode if did exist.

ulysses-you: how can the application live if builder is not processLaunched ?

turboFei: before recovery:
- 1. the batch is in PENDING state and process is running
- 2. kyuubi server crash, but the process has submitted the app but the batch state is still PENDING
recovery:
- 1. the batch builder is not launched
- 2. but the batch application is running

ulysses-you (approved):

PR #3958, opened 2022-12-10
Fix Spark session timezone format

cxzl25:
The Spark session supports setting the time zone through `spark.sql.session.timeZone` and formatting according to the time zone, but `timestamp` does not use timezone, resulting in some incorrect results.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
@@ -261,7 +261,7 @@ object RowSet {
         formatLocalDate(ld)
 
       case (t: Timestamp, TimestampType) =>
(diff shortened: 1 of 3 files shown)


cxzl25:
```sql
set spark.sql.session.timeZone=UTC;
select from_utc_timestamp(from_unixtime(1670404535000/1000,'yyyy-MM-dd HH:mm:ss'),'GMT+08:00') as time_utc8;
set spark.sql.session.timeZone=GMT+8;
select from_utc_timestamp(from_unixtime(1670404535000/1000,'yyyy-MM-dd HH:mm:ss'),'GMT+08:00') as time_utc8;
```

## Current
```
2022-12-08 01:15:35.0
2022-12-08 01:15:35.0
```
## Fix

```
2022-12-07 17:15:35.0
2022-12-08 01:15:35.0
```

turboFei (approved):

cfmcgrady (approved):
Arrow encoding also has the same problem, we can fix it in a separate PR

PR #3964, opened 2022-12-12
check state

turboFei:

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -298,20 +298,20 @@ class BatchJobSubmission(
 
       MetricsSystem.tracing(_.decCount(MetricRegistry.name(OPERATION_OPEN, opType)))
 
-      if (!builder.processLaunched) {
-        builder.close()
-        if (recoveryMetadata.isDefined) {
-          killMessage = killBatchApplication()
-        }
-        setState(OperationState.CANCELED)
-        updateBatchMetadata()
-        return
-      }
-
       // fast fail
       if (isTerminalState(state)) {
         killMessage = (false, s"batch $batchId is already terminal so can not kill it.")
         builder.close()
+        return
+      }
+
+      // the builder might not be launched for recovery mode
+      if (recoveryMetadata.isEmpty && !builder.processLaunched) {
+        builder.close()
+        if (isTerminalState(state)) {
+          setState(OperationState.CANCELED)
+        }
+        updateBatchMetadata()
         return
       }
 


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -298,20 +298,20 @@ class BatchJobSubmission(
...
-        setState(OperationState.CANCELED)
-        updateBatchMetadata()
         return
       }
 
-      // fast fail
-      if (isTerminalState(state)) {
-        killMessage = (false, s"batch $batchId is already terminal so can not kill it.")
+      // the builder might not be launched for recovery mode
+      if (recoveryMetadata.isEmpty && !builder.processLaunched) {
```
ulysses-you, line 309: recoveryMetadata.nonEmpty  ?

turboFei: how about checking `getBackgroundHandle == null`?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -298,20 +298,22 @@ class BatchJobSubmission(
...
-        }
-        setState(OperationState.CANCELED)
-        updateBatchMetadata()
         return
       }
 
-      // fast fail
-      if (isTerminalState(state)) {
-        killMessage = (false, s"batch $batchId is already terminal so can not kill it.")
+      if (getBackgroundHandle == null) {
```
ulysses-you, line 308: we should trust the existed state machine to avoid use other flag to see the state. better to check with state `running` and `pending`

turboFei: But `BatchJobSubmission` is special, it will only change the state to `RUNNING` if get valid applicationInfo.

ulysses-you: shall we check `currentApplicationInfo` here and only kill if app id is not null ? I think it's no meaning to track the local state.

turboFei: seems we can check `if (state == OperationState.INITIALIZED) `

turboFei: have updated the code

turboFei:
thanks for the suggestion, updated

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -319,16 +309,21 @@ class BatchJobSubmission(
         killMessage = killBatchApplication()
         builder.close()
       } finally {
-        if (killMessage._1 && !isTerminalState(state)) {
-          // kill success and we can change state safely
-          // note that, the batch operation state should never be closed
+        if (state == OperationState.INITIALIZED) {
```
ulysses-you, line 312: can we add some comments to explain why we do not need to check whether kill succeeded

turboFei: thanks, updated

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #2887, opened 2022-06-15
[Improvement] Kyuubi pool-share-level  provides random policies may cause task-hot-issues or  production accident 


FishManHome:
### What would you like to be improved?

Provide more balanced policies based on spark engine operation  soon！！！

PR #3966, opened 2022-12-12
[KYUUBI #2887][FOLLOWUP] Rename configuration to kyuubi.engine.pool.selectPolicy

pan3793:

After some thought and some offline discussions, I propose to change the configuration key from `kyuubi.engine.pool.balance.policy` to `kyuubi.engine.pool.selectPolicy`.

And in the future if we introduce properties for specific policy, e.g. supposing we have a engine selection policy named `LOAD`, the properties would be

```
kyuubi.engine.pool.selectPolicy.load.abc=xxxx
kyuubi.engine.pool.selectPolicy.load.xyz=xxxx
```

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -1646,10 +1646,12 @@ object KyuubiConf {
     .createWithDefault(-1)
 
   val ENGINE_POOL_BALANCE_POLICY: ConfigEntry[String] =
-    buildConf("kyuubi.engine.pool.balance.policy")
-      .doc("The balance policy of queries in engine pool.<ul>" +
-        " <li>RANDOM - Randomly use the engine in the pool</li>" +
-        " <li>POLLING - Polling use the engine in the pool</li> </ul>")
+    buildConf("kyuubi.engine.pool.selectPolicy")
+      .doc("The policy of selecting engine in a pool engine for session." +
+        "<ul>" +
+        "<li>RANDOM - Randomly use the engine in the pool</li>" +
+        "<li>POLLING - Polling use the engine in the pool</li>" +
+        "</ul>")
       .version("1.7.0")
       .stringConf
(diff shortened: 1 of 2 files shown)


pan3793:
cc @ychris78 as well

turboFei (approved):

cxzl25 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1646,10 +1646,12 @@ object KyuubiConf {
     .createWithDefault(-1)
 
   val ENGINE_POOL_BALANCE_POLICY: ConfigEntry[String] =
-    buildConf("kyuubi.engine.pool.balance.policy")
-      .doc("The balance policy of queries in engine pool.<ul>" +
-        " <li>RANDOM - Randomly use the engine in the pool</li>" +
-        " <li>POLLING - Polling use the engine in the pool</li> </ul>")
+    buildConf("kyuubi.engine.pool.selectPolicy")
```
ulysses-you, line 1649: seems kyuubi use `a.b.c` instead of camelbak, so just `kyuubi.engine.pool.policy` ?

pan3793: Yea, this introduces the camel style to make the configuration key more clear.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1646,10 +1646,12 @@ object KyuubiConf {
...
+      .doc("The policy of selecting engine in a pool engine for session." +
```
yaooqinn, line 1650: The select policy of an engine from the corresponding engine pool engine for a session

pan3793: updated

pan3793:
Thanks, merging to master

PR #3975, opened 2022-12-13
Support to post batch session/operation event

turboFei:

Support to post batch session/operation event.

Besides, make some improvements.
- wrap `session.open` with `handleSessionException`
- add sessionType for KyuubiOperationEvent


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -186,7 +186,6 @@ class KyuubiSessionImpl(
   override protected def runOperation(operation: Operation): OperationHandle = {
     if (operation != launchEngineOp) {
       waitForEngineLaunched()
-      sessionEvent.totalOperations += 1
```
turboFei, line 189: should we treat launchEngineOp as an operation?
cc @wForget

wForget: > should we treat launchEngineOp as an operation? cc @wForget

This means whether we should treat `openSession` as an operation, i tend not to treat as an operation.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -143,6 +145,13 @@ class BatchJobSubmission(
   private def setStateIfNotCanceled(newState: OperationState): Unit = state.synchronized {
     if (state != CANCELED) {
       setState(newState)
+      applicationInfo.filter(_.id != null).foreach { ai =>
+        session.getSessionEvent.foreach(_.engineId = ai.id)
+      }
+      if (newState == RUNNING) {
+        session.onEngineOpened()
```
wForget, line 152: Will this be called multiple times?

turboFei: One time is expected

turboFei: check the session opened time to confirm that

kyuubi-server/src/main/scala/org/apache/kyuubi/events/KyuubiOperationEvent.scala
```diff
@@ -81,6 +84,7 @@ object KyuubiOperationEvent {
       status.completed,
       status.exception,
       session.handle.identifier.toString,
-      session.user)
+      session.user,
+      session.sessionType.toString)
```
turboFei, line 88: now we have two kind of sessions, it is better to post the session type, so that we can group by session type to check the sql/batch operations trend.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -77,6 +77,8 @@ class KyuubiBatchSessionImpl(
       sessionManager.validateBatchConf(batchRequest.getConf.asScala.toMap)
   }
 
+  override lazy val name: Option[String] = Option(batchRequest.getName)
```
turboFei, line 80: I think it is not needed to respect `KyuubiConf.SESSION_NAME.key` here
@ulysses-you

turboFei: please ignore

turboFei:
cc @ulysses-you  @wForget

wForget (approved):
LGTM

wForget:
By the way, I think the `SessionType.SQL` type is confusing, because the `SessionType.SQL` type of Session can also execute scala, python, etc.

turboFei:
> By the way, I think the SessionType.SQL type is confusing, because the SessionType.SQL type of Session can also execute scala, python, etc.


Yes, maybe `INTERACTIVE` is better.

turboFei:
thanks, merged to master

PR #3983, opened 2022-12-15
[KYUUBI  #3982] [FEATURE] introduce refreshing user defaults configs

### LINKED ISSUE #3982: [FEATURE] introduce refreshing user defaults configs
### Code of Conduct

### Search before asking

### Describe the feature

Introduce feature of refresh user defaults config (which starts with three continuous underscores "___") from config file via `kyuubi-admin` cli and `refresh/user_defaults_conf` Rest API.

1. add `refreshUserDefaultsConf` methond in KyuubiServer to read user defautls configs from property file and apply config changes to server's KyuubiConf
3. add `refresh/user_defaults_conf` api to AdminRestApi calling `refreshUserDefaultsConf` of KyuubiServer
3. add config type `usersConfigs` in kyuubi-admin cli

This feature will
- help to apply user defaults conf without restarting server or losing connections
- load latest config for engine launch, e.g. spark related config `spark.*`

It won't
- affect the components already started and using the clone of server conf
- affect configs for launched engine instance

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

bowenliang123:

to close #3982 .

Introduce feature of refresh user defaults config (as `___${user}___.*` which starts with three continuous underscores "___") from config file via `kyuubi-admin` cli and `refresh/user_defaults_conf` Rest API.

1. add `refreshUserDefaultsConf` methond in KyuubiServer to read user defautls configs from property file and apply config changes to server's KyuubiConf
3. add `refresh/user_defaults_conf` api to AdminRestApi calling `refreshUserDefaultsConf` of KyuubiServer
3. add config type `userDefautls` in kyuubi-admin cli refresh command

This feature will
- help to apply user defaults conf without restarting server or losing connections
- load latest config for engine launch, e.g. spark related config `spark.*`

It won't
- affect the components already started and using the clone of server conf
- affect configs for launched engine instance

### DIFF
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
@@ -102,6 +102,6 @@ object AdminCommandLine extends CommonCommandLine {
           .optional()
           .action((v, c) => c.copy(adminConfigOpts = c.adminConfigOpts.copy(configType = v)))
           .text("The valid config type can be one of the following: " +
-            s"$HADOOP_CONF, $TYPE_USER_DEFAULTS_CONF."))
+            s"$HADOOP_CONF, $USER_DEFAULTS_CONF."))
   }
 }
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
@@ -100,6 +100,6 @@ object AdminCommandLine extends CommonCommandLine {
         arg[String]("<configType>")
           .optional()
           .action((v, c) => c.copy(adminConfigOpts = c.adminConfigOpts.copy(configType = v)))
-          .text("The valid config type can be one of the following: hadoopConf, serverConf."))
+          .text("The valid config type can be one of the following: hadoopConf, userDefaults."))
   }
 }
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
@@ -100,6 +100,6 @@ object AdminCommandLine extends CommonCommandLine {
         arg[String]("<configType>")
           .optional()
           .action((v, c) => c.copy(adminConfigOpts = c.adminConfigOpts.copy(configType = v)))
-          .text("The valid config type can be one of the following: hadoopConf."))
+          .text("The valid config type can be one of the following: hadoopConf, serverConf."))
   }
 }
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
@@ -20,6 +20,7 @@ package org.apache.kyuubi.ctl.opt
 import scopt.{OParser, OParserBuilder}
 
 import org.apache.kyuubi.KYUUBI_VERSION
+import org.apache.kyuubi.ctl.cmd.refresh.RefreshConfigCommandConfigType._
 
 object AdminCommandLine extends CommonCommandLine {
 
@@ -100,6 +101,7 @@ object AdminCommandLine extends CommonCommandLine {
         arg[String]("<configType>")
           .optional()
           .action((v, c) => c.copy(adminConfigOpts = c.adminConfigOpts.copy(configType = v)))
-          .text("The valid config type can be one of the following: hadoopConf, userDefaults."))
+          .text("The valid config type can be one of the following: " +
+            s"$HADOOP_CONF, $TYPE_USER_DEFAULTS_CONF."))
   }
 }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
@@ -105,6 +105,10 @@ object KyuubiServer extends Logging {
     val _hadoopConf = KyuubiHadoopUtils.newHadoopConf(new KyuubiConf().loadFileDefaults())
     hadoopConf = _hadoopConf
   }
+
+  private[kyuubi] def reloadServerConf(): Unit = synchronized {
+    kyuubiServer.conf.loadFileDefaults()
+  }
 }
 
 class KyuubiServer(name: String) extends Serverable(name) {
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
@@ -108,13 +108,11 @@ object KyuubiServer extends Logging {
 
   private[kyuubi] def refreshUserDefaultsConf(): Unit = kyuubiServer.conf.synchronized {
     val existedUserDefaults = kyuubiServer.conf.getAllUserDefaults
-    val newLoadedUserDefaults = new KyuubiConf(true).getAllUserDefaults
-    for ((k, _) <- existedUserDefaults) {
-      if (!newLoadedUserDefaults.contains(k)) {
-        kyuubiServer.conf.unset(k)
-      }
+    val refreshedUserDefaults = KyuubiConf().loadFileDefaults().getAllUserDefaults
+    for ((k, _) <- existedUserDefaults if !refreshedUserDefaults.contains(k)) {
+      kyuubiServer.conf.unset(k)
     }
-    for ((k, v) <- newLoadedUserDefaults) {
+    for ((k, v) <- refreshedUserDefaults) {
       kyuubiServer.conf.set(k, v)
     }
   }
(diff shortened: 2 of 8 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -105,6 +105,10 @@ object KyuubiServer extends Logging {
     val _hadoopConf = KyuubiHadoopUtils.newHadoopConf(new KyuubiConf().loadFileDefaults())
     hadoopConf = _hadoopConf
   }
+
+  private[kyuubi] def reloadServerConf(): Unit = synchronized {
+    kyuubiServer.conf.loadFileDefaults()
```
bowenliang123, line 110: @turboFei @pan3793 
Have a look at this PR if you have time.
It's reusing the `loadFileDefaults` method to read the config file  `kyuubi-defaults.conf` and overriding server configs. And if any configs not in the config file are set during runtime, they won't be affected.

pan3793:
An alternative option is wrapper `settings` and `reader` of `KyuubiConf` by Guava cache, WDYT? cc @yaooqinn @ulysses-you @turboFei

bowenliang123:
I am open to refreshing server conf manually or automatically. Auto-refreshed cache with intervals provided is also good to me.
Considered replacing `settings` from map to Guava cache. There's one thing that bothered me when getting evicted key from the cache loader, it will cause extra repeated IO load of reading the whole config file for each key.

ulysses-you:
I think the first things is: we should make sure what configs can be hot updated. e.g. what happens if user modify the Kyuubi server port or host ?

turboFei:
how about referring hadoop implementation
https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Reconfigurable.java

bowenliang123:
I would like to change this PR for reloading users' default configs (with `"___${user}___"` prefix) this time rather than all the server configs, to limit the impact for engine instance launching.

For `Reconfigurable` support in each config, it could be introduced later with a detailed design and related mechanism for event or restarting components.

pan3793:
> I would like to change this PR for reloading users' default configs (with `"___${user}___"` prefix) ...

SGTM, I suppose this way does not introduce side effects.

bowenliang123:
> I would like to change this PR for reloading users' default configs (with `"___${user}___"` prefix) this time rather than all the server configs, to limit the impact for engine instance launching.
> 
> For `Reconfigurable` support in each config, it could be introduced later with a detailed design and related mechanism for event or restarting components.

cc @pan3793 @ulysses-you @turboFei 
I have changed this PR to refreshing user defaults configs, please have a check.

ulysses-you:
I'm not against this feature. But a question: can this be covered by `SessionConfAdvisor` ?

bowenliang123:
> I'm not against this feature. But a question: can this be covered by `SessionConfAdvisor` ?

`SessionConfAdvisor` was considered and may not be the best solution for this scenario.
1. user defaults configs are prepared and set in `kyuubi-defaults.conf`. From the  perspective of ops, changing it as where it is is more straight, and this refreshing feature only helps applying changes without restarting server instance.
2. general usage of `SessionConfAdvisor` is not limit to applying specific default values to conf but also contains some rules for proper conf settings. e.g. fix max mem usage over 4g to 4g at maximum.
3. only one `SessionConfAdvisor` supported right now and it may conflict to other possible usage of `SessionConfAdvisor`
4. user defaults configs may cover larger area than session confs

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
```diff
@@ -100,6 +100,6 @@ object AdminCommandLine extends CommonCommandLine {
         arg[String]("<configType>")
           .optional()
           .action((v, c) => c.copy(adminConfigOpts = c.adminConfigOpts.copy(configType = v)))
-          .text("The valid config type can be one of the following: hadoopConf."))
+          .text("The valid config type can be one of the following: hadoopConf, userDefaults."))
```
pan3793, line 103: userDefaults => userDefaultsConf

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -64,6 +64,26 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
   }
 
+  @ApiResponse(
+    responseCode = "200",
+    content = Array(new Content(
+      mediaType = MediaType.APPLICATION_JSON)),
+    description = "refresh the user defaults configs")
+  @POST
+  @Path("refresh/user_defaults_conf")
+  def refreshUserDefaultsConf(): Response = {
```
ulysses-you, line 74: shall we lock this method ?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -105,6 +105,23 @@ object KyuubiServer extends Logging {
     val _hadoopConf = KyuubiHadoopUtils.newHadoopConf(new KyuubiConf().loadFileDefaults())
     hadoopConf = _hadoopConf
   }
+
+  private[kyuubi] def refreshUserDefaultsConf(): Unit = kyuubiServer.conf.synchronized {
+    kyuubiServer.conf.loadFileDefaults()
```
ulysses-you, line 110: what does this do ?

kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala
bowenliang123: OK, let me fix it.
Thus the cli command will become "kyuubi-admin refresh config userDefaultsConf". `conf` occurs twice.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -105,6 +105,23 @@ object KyuubiServer extends Logging {
...
+    val newLoadedUserDefaults =
+      new KyuubiConf(true).getAllWithPrefix(USER_DEFAULTS_CONF_QUOTE, "")
```
pan3793, line 112: Emm... is this implementation correct?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -105,6 +105,23 @@ object KyuubiServer extends Logging {
...
+    val existedUserDefaults = kyuubiServer.conf.getAllWithPrefix(USER_DEFAULTS_CONF_QUOTE, "")
+
+    for ((k, _) <- existedUserDefaults) {
+      if (!newLoadedUserDefaults.contains(k)) {
+        kyuubiServer.conf.unset(k)
+      }
+    }
+
+    for ((k, v) <- newLoadedUserDefaults) {
+      kyuubiServer.conf.set(k, v)
```
pan3793, line 122: Emm... is this implementation correct?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
bowenliang123: Sorry for this mistake. legacy code from the initial version for refreshing all server configs.
Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
bowenliang123: `refreshUserDefaultsConf` of KyuubiServer is synchronized on kyuubiServer.conf to prevent concurrent racing.

bowenliang123:
Also, consider backporting this feature to 1.6 ?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala
```diff
@@ -105,6 +105,19 @@ object KyuubiServer extends Logging {
     val _hadoopConf = KyuubiHadoopUtils.newHadoopConf(new KyuubiConf().loadFileDefaults())
     hadoopConf = _hadoopConf
   }
+
+  private[kyuubi] def refreshUserDefaultsConf(): Unit = kyuubiServer.conf.synchronized {
+    val existedUserDefaults = kyuubiServer.conf.getAllUserDefaults
+    val newLoadedUserDefaults = new KyuubiConf(true).getAllUserDefaults
```
ulysses-you, line 111: you should call `new KyuubiConf(true).loadFileDefaults`, otherwise nothing would happen

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -64,6 +64,26 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
+    val userName = fe.getSessionUser(Map.empty[String, String])
+    val ipAddress = fe.getIpAddress
+    info(s"Receive refresh Kyuubi server conf request from $userName/$ipAddress")
+    if (!userName.equals(administrator)) {
+      throw new NotAllowedException(
+        s"$userName is not allowed to refresh the user defaults conf")
+    }
+    info(s"Reloading the Kyuubi server conf")
+    KyuubiServer.refreshUserDefaultsConf()
+    Response.ok(s"Refresh the server conf successfully.").build()
```
ulysses-you, line 84: nit: `Refresh the user defaults conf successfully`

ulysses-you (commented):
lgtm except one minor comment

ulysses-you:
thanks, merging to master

PR #3988, opened 2022-12-15
Final stage config isolation support write only

ulysses-you:

Detect and inject a tag if plan is for writing, then skip doing final stage isolation at query preparation phase.

To make final stage config more flexible with complex Spark application.


yaooqinn:
It surprises me that it did not only support for the write side

ulysses-you:
@yaooqinn It is a long time issue but I have not been talked about it something wrong..

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiQueryStagePreparation.scala
```diff
@@ -58,6 +58,16 @@ case class FinalStageConfigIsolation(session: SparkSession) extends Rule[SparkPl
...
 
     if (isFinalStage(plan)) {
+      // We can not get the whole plan at query preparation phase to detect if current plan is
+      // for writing, so we depend on a tag which is been injected at post resolution phase.
+      // Note: we should still do clean up previous config for non-final stage to avoid such case:
+      // the first statement is write, but the second statement is query.
+      if (conf.getConf(FINAL_STAGE_CONFIG_ISOLATION_WRITE_ONLY) &&
+        conf.getConf(KyuubiSQLConf.MARK_NUM_OUTPUT_COLUMNS) &&
+        !MarkNumOutputColumnsRule.isWrite(session, plan)) {
+        return plan
```
ulysses-you, line 68: It's real hard to detect if current plan is inside a subquery or a cached plan, so I just check output columns for writing. Most of case should be covered.

ulysses-you:
cc @yaooqinn @cfmcgrady

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiQueryStagePreparation.scala
```diff
@@ -58,6 +58,16 @@ case class FinalStageConfigIsolation(session: SparkSession) extends Rule[SparkPl
     }
 
     if (isFinalStage(plan)) {
+      // We can not get the whole plan at query preparation phase to detect if current plan is
+      // for writing, so we depend on a tag which is been injected at post resolution phase.
+      // Note: we should still do clean up previous config for non-final stage to avoid such case:
+      // the first statement is write, but the second statement is query.
+      if (conf.getConf(FINAL_STAGE_CONFIG_ISOLATION_WRITE_ONLY) &&
+        conf.getConf(KyuubiSQLConf.MARK_NUM_OUTPUT_COLUMNS) &&
```
cfmcgrady, line 66: Rule `MarkNumOutputColumnsRule` doesn't make any sense when condition `FINAL_STAGE_CONFIG_ISOLATION_WRITE_ONLY` is false, maybe we can simplify these two config?

ulysses-you: make sense, removed config `MARK_NUM_OUTPUT_COLUMNS`

ulysses-you:
@yaooqinn @cfmcgrady any comments ?

yaooqinn (approved):

ulysses-you:

Issue #3990, opened 2022-12-16
[Subtask] Support for translating `SELECT TABLE_CAT FROM system.jdbc.catalogs` to `Getcatalogs` operation in Trino front protocol


Yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the subtask

Kyuubi introduce Trino front protocol in https://github.com/apache/incubator-kyuubi/issues/3901, this issue as a subtask to support for translating `SELECT TABLE_CAT FROM system.jdbc.catalogs` to `Getcatalogs` operation.

### Parent issue

https://github.com/apache/incubator-kyuubi/issues/3901

### Are you willing to submit PR?

- [X] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #3991, opened 2022-12-16
[KYUUBI #3990] Support for translating `SELECT TABLE_CAT FROM system.jdbc.catalogs` to `Getcatalogs` operation in Trino front protocol

Yikf:

close https://github.com/apache/incubator-kyuubi/issues/3990, Kyuubi introduce Trino front protocol in https://github.com/apache/incubator-kyuubi/issues/3901, this issue as a subtask to support for translating `SELECT TABLE_CAT FROM system.jdbc.catalogs` to `Getcatalogs` operation.


Yikf:
How about we integrate the Trino front-end protocol with the parser module this way? @ulysses-you

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -265,8 +268,12 @@ class KyuubiSessionImpl(
           runAsync,
           command)
         runOperation(operation)
+      case translateNode: TrinoTranslatorNode if frontendProtocol == TRINO =>
```
yaooqinn, line 271: Things get crazy

Yikf:
After offline discussion, it will be implemented in the trino fe side

Issue #3973, opened 2022-12-13
[Regression] kyuubi-ctl should recognize kyuubi.ha.zookeeper.quorum after upgrading to 1.6.0


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

https://github.com/apache/incubator-kyuubi/pull/2766 introduced a regression, assume user upgrading from older version, and use `kyuubi.ha.zookeeper.quorum` instead of `kyuubi.ha.addresses`, the `kyuubi-ctl list` will fail w/ 
```
hadoop@spark5:~/kyuubi_hz_cluster_8/kyuubi$ bin/kyuubi-ctl list
2022-11-30 11:59:45.556 INFO org.apache.kyuubi.Utils: Loading Kyuubi properties from /home/hadoop/kyuubi_hz_cluster_8/kyuubi/conf/kyuubi-defaults.conf
2022-11-30 11:59:45.720 WARN org.apache.hadoop.util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Exception in thread "main" org.apache.kyuubi.KyuubiException: Zookeeper quorum is not specified and no default value to load
	at org.apache.kyuubi.ctl.util.Validator$.fail(Validator.scala:81)
	at org.apache.kyuubi.ctl.util.Validator$.validateZkArguments(Validator.scala:31)
	at org.apache.kyuubi.ctl.cmd.list.ListCommand.validate(ListCommand.scala:27)
	at org.apache.kyuubi.ctl.ControlCliArguments.parse(ControlCliArguments.scala:57)
	at org.apache.kyuubi.ctl.ControlCliArguments.<init>(ControlCliArguments.scala:39)
	at org.apache.kyuubi.ctl.ControlCli$$anon$1$$anon$2.<init>(ControlCli.scala:58)
	at org.apache.kyuubi.ctl.ControlCli$$anon$1.parseArguments(ControlCli.scala:58)
	at org.apache.kyuubi.ctl.ControlCli.doAction(ControlCli.scala:33)
	at org.apache.kyuubi.ctl.ControlCli$$anon$1.doAction(ControlCli.scala:78)
[...]

PR #3997, opened 2022-12-16
[KYUUBI #3973] Fix a conf fallback for `kyuubi-ctl list server` when kyuubi-defaults.conf use deprecated zk key

CavemanIV:

To fix https://github.com/apache/incubator-kyuubi/issues/3973, we found similar problem during upgrade from 1.5 to 1.6

Manual test on internal cluster, UT will be added later on

### DIFF
--- a/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/Command.scala
+++ b/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/Command.scala
@@ -57,8 +57,8 @@ abstract class Command[T](cliConfig: CliConfig) extends Logging {
   private def useDefaultPropertyValueIfMissing(): CliConfig = {
     var arguments: CliConfig = cliConfig.copy()
     if (cliConfig.zkOpts.zkQuorum == null) {
-      val quorum = conf.getOption(HA_ADDRESSES.key).
-        getOrElse(conf.getOption(HA_ZK_QUORUM.key).orNull)
+      val quorum =
+        conf.getOption(HA_ADDRESSES.key).getOrElse(conf.getOption(HA_ZK_QUORUM.key).orNull)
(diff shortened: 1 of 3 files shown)


kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/Command.scala
```diff
@@ -57,11 +57,13 @@ abstract class Command[T](cliConfig: CliConfig) extends Logging {
   private def useDefaultPropertyValueIfMissing(): CliConfig = {
     var arguments: CliConfig = cliConfig.copy()
     if (cliConfig.zkOpts.zkQuorum == null) {
-      conf.getOption(HA_ADDRESSES.key).foreach { v =>
```
pan3793, line 60: maybe it can be simply fixed by `conf.get(HA_ADDRESSES)`

CavemanIV: coding style consistent, done

pan3793: Have you try my suggestion? The `KyuubiConf` system should support fallback mechanism

pan3793: for code style, it can be quickly fixed by `dev/reformat`

CavemanIV: right `conf.get` is simpler and easy. I didn't notice `.get` and `getOption` have different implementation, thought fallback not working

cxzl25 (approved):

pan3793 (approved):

pan3793:
Thanks, merged to master/1.6

PR #4002, opened 2022-12-18
[FEATURE] Add Kyuubi Web UI

zhaohehuhu:


zwangsheng:
It looks like this is a huge PR with a lot of features. Wonder if you'd like to break it into more purposeful PR.

PR #4004, opened 2022-12-18
[K8S][HELM] Add configuration files support to helm chart

dnskr:
The changes allow to:
1. set Kyuubi configuration files:
    - `kyuubi-env.sh`
    - `kyuubi-defaults.conf`
    - `log4j2.xml`
2. restart (recreate) Kyuubi server pods if configuration files have changed
    - this probably should be revisited after https://github.com/apache/incubator-kyuubi/pull/3983

### DIFF
--- a/docker/helm/templates/kyuubi-configmap.yaml
+++ b/docker/helm/templates/kyuubi-configmap.yaml
@@ -18,7 +18,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ .Release.Name }}-kyuubi-defaults
+  name: {{ .Release.Name }}
   labels:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     app.kubernetes.io/name: {{ .Chart.Name }}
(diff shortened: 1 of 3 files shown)


dnskr:
I'm not sure about default values for `kyuubi-defaults.conf`:
1. I can't find `kyuubi.kubernetes.namespace` config both documentation and source code, so I deleted it.
2. I found that `kyuubi.frontend.bind.host` and `kyuubi.frontend.bind.port` marked as deprecated in the [doc](https://kyuubi.apache.org/docs/latest/deployment/settings.html#frontend), but I can't find what should be used instead.

pan3793:
Thanks @dnskr 

> 1. I can't find `kyuubi.kubernetes.namespace` config both documentation and source code, so I deleted it.

You can find it in the [master branch doc(unreleased)](https://kyuubi.readthedocs.io/en/latest/index.html), the [latest doc](https://kyuubi.apache.org/docs/latest) represents the latest stable release doc.

> 2. I found that `kyuubi.frontend.bind.host` and `kyuubi.frontend.bind.port` marked as deprecated in the [doc](https://kyuubi.apache.org/docs/latest/deployment/settings.html#frontend), but I can't find what should be used instead.

Sorry for confusing, I opened https://github.com/apache/incubator-kyuubi/pull/4005 to explain and fix it.

docker/helm/templates/kyuubi-configmap.yaml
```diff
@@ -18,39 +18,30 @@
...
-
-    #
-    # kyuubi.authentication           NONE
-    #
+    ## Helm chart provided Kyuubi configurations
     kyuubi.frontend.bind.host={{ .Values.server.bind.host }}
     kyuubi.frontend.bind.port={{ .Values.server.bind.port }}
     kyuubi.kubernetes.namespace={{ .Release.Namespace }}
 
-    # Details in https://kyuubi.apache.org/docs/latest/deployment/settings.html
```
pan3793, line 56: can we reserve this link so that users can find all available configurations?

docker/helm/templates/kyuubi-deployment.yaml
```diff
@@ -36,6 +36,8 @@ spec:
       labels:
         app.kubernetes.io/name: {{ .Chart.Name }}
         app.kubernetes.io/instance: {{ .Release.Name }}
+      annotations:
+        checksum/conf: {{ include (print $.Template.BasePath "/kyuubi-configmap.yaml") . | sha256sum }}
```
pan3793, line 40: I'm not really familiar w/ helm, is it recommended? where can I find docs on that?

dnskr: Yes, it's official helm recommended [Chart Development Tips and Tricks](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) which actually sounds funny.
It's used in Apache Airflow [helm chart](https://github.com/apache/airflow/blob/672264b0af4874274bd130ba42a78e8e72f3d3ff/chart/templates/webserver/webserver-deployment.yaml#L79-L86) for example.

pan3793 (approved):

pan3793:
@hddong @dnskr some thoughts about the schema of `values.yaml`

```
server:
  bind:
    host: 0.0.0.0
    port: 10009
  confDir:
  conf:
    kyuubiEnv:
    kyuubiDefaults:
    log4j2:
```

How about change `server` to `kyuubi`? And we may need a similar part of `spark` and `hadoop`

For "bind" part, the port is not sufficient since Kyuubi supports multi protocols, not only thrift binary now.

dnskr:
> How about change `server` to `kyuubi`? And we may need a similar part of `spark` and `hadoop`
> 
> For "bind" part, the port is not sufficient since Kyuubi supports multi protocols, not only thrift binary now.

Agree with both suggestions. I left it as is just because I'm quite new in Kyuubi and don't fully understand all moving pieces. But I have enthusiasm to continue exploring it and creating PRs.

pan3793:
Merged to master, thanks @dnskr for your contribution and passion. The team is suffering from COVID-19, and may not be able to reply in time.

bowenliang123:
@dnskr Hi, https://github.com/apache/kyuubi/issues/3982 is closed, while I am afraid that it won't help in your case since it was refactored to refresh users' default configs only as dissucssed in PR 3983.

But you can still consider follow the similar pattern to achieve it as the following as 3982 does,
- add REST API to apply config changes and implement related method in server
- add command or config type for refreshing configs in `kyuubi-admin`

PR #4005, opened 2022-12-19
Remove deprecated from description of kyuubi.frontend.bind.host

pan3793:

Unlike `kyuubi.frontend.bind.port`, which is deprecated and replaced by `kyuubi.frontend.thrift.binary.bind.port`, `kyuubi.frontend.bind.host` is still encouraged to use, and it can be overwritten by `kyuubi.frontend.<protocol>.bind.host`, e.g. `kyuubi.frontend.thrift.binary.bind.host`, `kyuubi.frontend.thrift.http.bind.host`, `kyuubi.frontend.rest.bind.host`, `kyuubi.frontend.mysql.bind.host`

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -385,8 +385,7 @@ object KyuubiConf {
       .createWithDefault(Seq(FrontendProtocols.THRIFT_BINARY.toString))
 
   val FRONTEND_BIND_HOST: OptionalConfigEntry[String] = buildConf("kyuubi.frontend.bind.host")
-    .doc("Hostname or IP of the machine on which to run the thrift frontend service " +
-      "via binary protocol.")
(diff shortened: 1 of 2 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -385,7 +385,7 @@ object KyuubiConf {
       .createWithDefault(Seq(FrontendProtocols.THRIFT_BINARY.toString))
 
   val FRONTEND_BIND_HOST: OptionalConfigEntry[String] = buildConf("kyuubi.frontend.bind.host")
-    .doc("(deprecated) Hostname or IP of the machine on which to run the thrift frontend service " +
+    .doc("Hostname or IP of the machine on which to run the thrift frontend service " +
```
yaooqinn, line 388: Not thrift only？

pan3793: Yes, updated.

yaooqinn (approved):

PR #4007, opened 2022-12-19
[KYUUBI #4006] Fix the incorrect execution of the source command

### LINKED ISSUE #4006: [Bug] The source command of beeline cannot be executed correctly
### Code of Conduct

### Search before asking

### Describe the bug

Executing the source command dose not any response.

image

due to extra space:

image

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

- [X] Yes. I can submit a PR independently to fix.

wForget:

close #4006

refer https://github.com/apache/hive/blob/7c343471aa68d7d06209694d9d6b181bd58e0793/beeline/src/java/org/apache/hive/beeline/Commands.java#L856

Successfully executed after fix:
image

### DIFF
--- a/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
+++ b/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
@@ -45,7 +45,7 @@ public class KyuubiCommands extends Commands {
 
   /** Extract and clean up the first command in the input. */
   private String getFirstCmd(String cmd, int length) {
-    return cmd.substring(length);
+    return cmd.substring(length).trim();
   }
 
   private String[] tokenizeCmd(String cmd) {
--- a/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
+++ b/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
@@ -82,9 +82,8 @@ public class KyuubiCommands extends Commands {
     BufferedReader reader = null;
     try {
       reader = new BufferedReader(new FileReader(sourceFile));
-      String extra = reader.readLine();
-      String lines = null;
-      while (extra != null) {
+      String lines = null, extra;
+      while ((extra = reader.readLine()) != null) {
         if (beeLine.isComment(extra)) {
           continue;
         }
@@ -93,7 +92,6 @@ public class KyuubiCommands extends Commands {
         } else {
           lines += "\n" + extra;
         }
-        extra = reader.readLine();
       }
       String[] cmds = lines.split(";");
       for (String c : cmds) {


kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
```diff
@@ -45,7 +45,7 @@ public boolean sql(String line) {
 
   /** Extract and clean up the first command in the input. */
   private String getFirstCmd(String cmd, int length) {
-    return cmd.substring(length);
+    return cmd.substring(length).trim();
```
cxzl25, line 48: This should be to avoid submitting Python code to remove indentation and cause syntax failure.  
https://github.com/apache/incubator-kyuubi/pull/3762

Wouldn't it be better if we could match different commands to determine trim?

cc @cfmcgrady  @pan3793

wForget: > This should be to avoid submitting Python code to remove indentation and cause syntax failure.

The `getFirstCmd` method is called only in `sourceFile`, I don't understand why it is affected?

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
```diff
@@ -82,9 +82,8 @@ private boolean sourceFileInternal(File sourceFile) throws IOException {
     BufferedReader reader = null;
     try {
       reader = new BufferedReader(new FileReader(sourceFile));
-      String extra = reader.readLine();
-      String lines = null;
-      while (extra != null) {
+      String lines = null, extra;
+      while ((extra = reader.readLine()) != null) {
         if (beeLine.isComment(extra)) {
```
wForget, line 87: comment line will cause infinite loop.

wForget: cc @turboFei

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
cfmcgrady: > > This should be to avoid submitting Python code to remove indentation and cause syntax failure.
> 
> The `getFirstCmd` method is called only in `sourceFile`, I don't understand why it is affected?

seems you are reasonable

also cc @bowenliang123, can you help verify this change?

cfmcgrady: tested locally, this change does not affect submitting python code.

cfmcgrady (approved):

turboFei (approved):

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
cxzl25: [HIVE-16693](https://issues.apache.org/jira/browse/HIVE-16693): beeline "source" command freezes if you have a comment in it

Fix Version: 3.0.0

cxzl25 (approved):

turboFei:
thanks, merging to master

turboFei:
thanks, merged to master and branch-1.6

PR #4010, opened 2022-12-19
Print stack trace in `KyuubiOperation#onError` method

cxzl25:

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
@@ -57,7 +57,7 @@ abstract class KyuubiOperation(session: Session) extends AbstractOperation(sessi
     case e: Throwable =>
       state.synchronized {
         if (isTerminalState(state)) {
-          warn(s"Ignore exception in terminal state with $statementId: $e")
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
```diff
@@ -57,7 +57,7 @@ abstract class KyuubiOperation(session: Session) extends AbstractOperation(sessi
     case e: Throwable =>
       state.synchronized {
         if (isTerminalState(state)) {
-          warn(s"Ignore exception in terminal state with $statementId: $e")
```
cxzl25, line 60: "$e" only get message

```
22/12/19 10:31:24,012 [KyuubiSessionManager-exec-pool: Thread-91361] WARN ExecuteStatement: Ignore exception in terminal state with a3b609f4-1654-4f5a-87d8-8864036df21a: org.apache.thrift.transport.TTransportException: java.net.SocketException: Socket closed
```

pan3793: I'm not sure if it's by design, because we just ignore the exception here. but i'm fine w/ this change.

pan3793:
Merged to master

PR #4019, opened 2022-12-21
Binding python/sql spark session

cfmcgrady:

Bind python and SQL spark session, then the variables we set on the python side can be visited on the SQL side

After this PR, we can change the execution mode from python to sql by running

```python
spark.sql("SET kyuubi.operation.language=SQL").show()
```

5091671606960_ pic_hd

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
@@ -18,6 +18,7 @@
 package org.apache.kyuubi.engine.spark.session
 
 import java.util.concurrent.{ScheduledExecutorService, TimeUnit}
+import java.util.concurrent.atomic.AtomicReference
 
 import org.apache.hive.service.rpc.thrift.TProtocolVersion
 import org.apache.spark.sql.SparkSession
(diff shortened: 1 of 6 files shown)


cfmcgrady:
cc @bowenliang123 @turboFei

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -177,3 +181,18 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
...
+
+object SparkSQLSessionManager {
+  private val sessionManager = new AtomicReference[SparkSQLSessionManager]
+
+  private def setSessionManager(manager: SparkSQLSessionManager): Unit = {
+    sessionManager.set(manager)
+  }
+
+  def getSparkSession(uuid: String): SparkSession = {
+    sessionManager.get()
```
turboFei, line 193: how about get from object `SparkSQLEngine`.

cfmcgrady: we need a static method here, and call the method and return sql spark session on the python side. how can we get the sql spark session from `SparkSQLEngine`?

bowenliang123 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
turboFei:
```
  SparkSQLEngine.currentEngine.map(_.backendService.sessionManager)

```

turboFei: with that, SparkSQLSessionManager::setSessionManager is not needed.

cxzl25 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
cfmcgrady: addressed.

turboFei (approved):

cfmcgrady:
thanks all. merging to master(v1.7.0)

Issue #3586, opened 2022-09-30
[Umbrella] [Authz] Support multiple Ranger services for different Spark catalogs


bowenliang123:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the proposal

Currently, Kyuubi Authz plugin inits a singleton plugin instance that only supports single Ranger policy service for all Spark catalogs. This limits the use range of the plugin with one default catalog or it will introduce chaos when querying in different catalogs.

To solve this problem, we will introduce multiple service support for different Spark V2 catalogs. 
Key points as listed,

- reserve default Ranger plugin instance as default fallback handler for all access privilege checking, row-filter and data masking for any default catalog or no explicit catalog identified.
- continuing to use Hive service def for building privilege types and operation types
- parsing catalog names from `DataSourceV2Relation` LogicalPlans in `Privilegebuilder`
- configuring and initialing RangerPlugin with the mapping of the catalog name to each Ranger service name
- support catalog in `AccessRequest`, row-filter and datamasking and diverting requests  to the corresponding RangerPlugin instance

### Task list

- [ ] (abandoned)#3594
- [x] #3605
- [x] #3984
- [ ] #4015 
- [ ] #4032 

1. PrivilegeObject -> AccessResource 加catalog字段并传递
2. Desciptor -> PrivilegeObject 传递 catalog，Desciptor新增Database类，拓展Function类、Table类 加catalog字段
3. spec -> Desciptor 提取catlog并填入，加入catalog extractor （for dsv2）, 及在tablespec和dbspec加catalog spec

### Are you willing to submit PR?

- [X] Yes. I can submit a PR independently to improve.
- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4032, opened 2022-12-27
[Authz] Extract catalog for DataSourceRelationV2 in buildQuery and  pass catalog of queries and commands to PrivilegeObject

bowenliang123:

subtask of #3586 .

1. passing catalog field from `Table` and `Database` to PrivilegeObject for privilege checks. As for`Function` related commands,  no catalog supported other than spark_catalog or v1 catalog in Spark 3.1-3.3.
2. extract catalog from `DataSourceRelationV2` in buildQuery with added `CatalogPluginOptionCatalogExtractor`
3. add ut for v2commands SetNamespaceProperties/CreateNamespace/SetNamespaceLocation/DescribeNamespace/DropNameSpace to verify related PrivilegeObject extraction with catalog in spark 3.3
4. in future PR,  SparkRangerAdminPlugin will use catalog-related instances for privilege checks using catalog info from PrivilegeObject 


bowenliang123:
cc @yaooqinn

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
```diff
@@ -155,15 +161,19 @@ abstract class PrivilegesBuilderSuite extends AnyFunSuite
   }
 
   test("CreateDatabaseCommand") {
+    assume(isSparkVersionAtMost("3.2"))
```
yaooqinn, line 164: what's wrong with these test after such a change?

bowenliang123: We add catalog assertion in thi ut.

In spark 3.0-3.2,  `CREATE DATABASE` sql is parsed into `CreateDatabaseCommand`, which is v1 command and contains no catalog info . And this ut asserts `catalog` in  parsed PviilegeObject is None.

 In spark 3.3, the same ddl will is pared into `CreateNamespace` of v2commands instead, and the asserted catalog is extracted by another related ResolvedDBObjectNameDatabaseExtractor with real value from CatalogPlugin. New ut for this case is also added to `V2CommandsPrivilegesSuite.scala` in this PR.

yaooqinn: in this case, we shall not put a uncommon test case in the base trait.

bowenliang123: I think they are actually testing different command plans and they are good enough to reflect spark version variation impact for this ddl.

yaooqinn (approved):

yaooqinn:
thanks, merged to master

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
bowenliang123: Plus, the migration from v1 to v2 is generally refactoring command detail itself affecting same commands to all and not related to v2 plans.

yaooqinn: thanks for the explanation. I want to eliminate spark version and v1/v2 table testing, maybe it's too early to do this.

PR #4044, opened 2022-12-30
[Authz] Serde for queries

yaooqinn:

Add serde layer for queries, which extracts table information from a scan relation, such as HiveTableRelation, logical relation, etc.


bowenliang123:
We should also consider ScanSpec for building function privileges of perm function usages in queries, as in pending PR 3633 (https://github.com/apache/kyuubi/pull/3633). The core logic to this is scanning logical plan by `transformAllExpressions` and build function privileges for UDFs.

yaooqinn:
> We should also consider ScanSpec for building function privileges of perm function usages in queries

How can ScanSpec configure functions?

bowenliang123:
> How can ScanSpec configure functions?
If `ScanSpec` is used for extracting all related resources, functions could be considered one type among them.
Or if it is limited to scanned resources, the used function support in queries can be introduced later.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/serde/CommandSpec.scala
```diff
@@ -71,4 +79,33 @@ case class TableCommandSpec(
...
+          None
+      }
+    }
+  }
+}
+
+case class ScanSpec(
+    classname: String,
+    scanDescs: Seq[ScanDesc]) extends CommandSpec {
+  override def opType: String = "QUERY"
```
bowenliang123, line 99: maybe extract this repeated opType "QUERY" to the object for preventing magic value? or use `OperationType.QUERY.toString`.

extensions/spark/kyuubi-spark-authz/src/main/resources/META-INF/services/org.apache.kyuubi.plugin.spark.authz.serde.TableExtractor
```diff
@@ -16,6 +16,7 @@
 #
 
 org.apache.kyuubi.plugin.spark.authz.serde.CatalogTableTableExtractor
+org.apache.kyuubi.plugin.spark.authz.serde.CatalogTableOptionTableExtractor
```
bowenliang123, line 19:
```suggestion
org.apache.kyuubi.plugin.spark.authz.serde.CatalogTableOptionTableExtractor
org.apache.kyuubi.plugin.spark.authz.serde.CatalogTableTableExtractor
```
sort the extractors in alphabet order.

bowenliang123:
LGTM. Just some minor comments. Let's merge it after passing the CI.

Glad to see this PR reaching a milestone for Serde refactoring, by completely removing hardcoded rules and extraction of `v2Commands` and `IcebergCommands` and using spec generation instead. 

Big applause to @yaooqinn for your elegant redesign and all the hard work. 👏🏻 🍻

yaooqinn:
thanks @bowenliang123 for the check, merged to master

PR #4067, opened 2023-01-03
Reset the operation log before fetching new one

turboFei:

https://github.com/apache/kyuubi/blob/9342a29284234e21c73f96024a740c6e93e7cb33/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/cmd/log/LogBatchCommand.scala#L134-L141

We need to reset the operation log before fetching new operation log, otherwise, we might meet corner case.

- The last operation log is not empty
- we meet log not found exception
- we can not terminate the loop because the `shouldFinishLog` is always false


ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #4057, opened 2022-12-31
[Bug] shell script about the var wrong


xiaoyuandajian:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

description: I'm trying to solve the issue about #2062 , the job is to add shell style check util to the GitHub action, I try to add that in my fork project, when test the shellcheck, I found some script syntax errors
1. it's about /bin/stop-application.sh , the reason is $# is a int type, if we want compare, we should us -lt\-gt, but the script us < or >
2. it's about /bin/kyuubi, we use exit -1, I know if we not use the code 0, it always report failed, but in shell, the right number is 0-255
3. it's about bin/docker-image-tool.sh, we should use #* replace #@

```
In bin/docker-image-tool.sh line 230:
if [[ "$@" = *--help ]] || [[ "$@" = *-h ]]; then
      ^--^ SC2199: Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @).
                              ^--^ SC2199: Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @).
```

```
In bin/kyuubi line 33:
if [[ "$@" = *--help ]] || [[ "$@" = *-h ]]; then
      ^--^ SC2199: Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @).
                              ^--^ SC2199: Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @)
In bin/kyuubi line 51:
    exit -1
         ^-- SC2242: Can only exit with status 0-255. Other data should be written to stdout/stderr.
```

```
In bin/stop-application.sh line 19:
if [[ $# < 1 ]] ; then
         ^-- SC2071: < is for string comparisons. Use -lt instead.
```

I think we should fix the bug in fix #206

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

xiaoyuandajian:
I don't know the issue is right or wrong, should I fix that?

xiaoyuandajian:
[...]

PR #4078, opened 2023-01-04
[INFRA] Enable shellcheck on CI

### LINKED ISSUE #2062: [Improvement] Add a job or step for shell script style in Github Action
### Code of Conduct

### Search before asking

shell script style consistency

### How should we improve?

Maybe adding https://github.com/marketplace/actions/shell-linter  in sytle.yaml is a good choice, or something else

xiaoyuandajian:

Fix #2062


.github/workflows/style.yml
```diff
@@ -100,3 +100,12 @@ jobs:
           echo "If there is a problem that cannot be fixed by the command, "
           echo "you need to manually fix it by following the information told by the command above"
           echo "---------------------------------------------------------------------------------"
+  shellcheck:
```
pan3793, line 103: please add a new line before it.

bin/docker-image-tool.sh
```diff
@@ -227,6 +227,7 @@ Examples:
 EOF
 }
 
+# shellcheck disable=SC2199
```
pan3793, line 230: this is an acceptable approach, we can disable the check first, and fix them one by one in the future. (if the fix is simple, it's fine to fix in place in this PR as well)

bin/kyuubi
```diff
@@ -48,6 +49,7 @@ function kyuubi_rotate_log() {
     num=${KYUUBI_MAX_LOG_FILES}
   else
     echo "Error: KYUUBI_MAX_LOG_FILES must be a positive number, but got ${KYUUBI_MAX_LOG_FILES}"
+    # shellcheck disable=SC2242
     exit -1
```
pan3793, line 53: this can be fixed in place, we just need to return a no-zero value here, 1 should be good.

xiaoyuandajian: actually I try this in my fork project, the shell script has many grammar mistake, so I use #shellcheck jump that, I want fix these bug in issue #4057 before do this shellcheck

xiaoyuandajian: what do you think about this idea

pan3793: > the shell script has many grammar mistake

the shell script grammar is flexible; some are style issues, and should not be treated as bug/mistake, because they have worked as expected.

> ... fix these bug in issue https://github.com/apache/kyuubi/issues/4057 before do this shellcheck

w/o shellcheck, how to verify the fix?

xiaoyuandajian: en, I think I can fix some grammar mistake, then other problems we can use # to ignore them

pan3793: sgtm

.github/workflows/style.yml
pan3793, line 103:
```suggestion

  shellcheck:
```

pan3793 (approved):

pan3793:
Thanks, merged to master

xiaoyuandajian:
I'm pleasure that

xiaoyuandajian:
> @xiaoyuandajian I just noticed that[1]
> 
> > ludeeus/action-shellcheck@master is not allowed to be used in apache/kyuubi.
> 
> I opened the https://issues.apache.org/jira/browse/INFRA-24064 to ask the ASF INFRA team to evaluate this plugin and hope it can be added to the whitelist.
> 
> [1] https://github.com/apache/kyuubi/actions/runs/3856320011

yes, I hope so

PR #4084, opened 2023-01-04
[Authz] Implicitly extracting name of Extractors and OperationType in spec file generation preventing use class name String directly

bowenliang123:

- add implicit methods `classSimpleName` and `operationTypeStr` to package object `org.apache.kyuubi.plugin.spark.authz` in test sources
- increase compilation alignment to extractor itself, and easier use for future third party adaptation avoiding calling wrong `getName` of the calss
- preventing use String or class name as magic value directly which easily causes mistakes
- no changes with spec JSON file regeneration after this PR 

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/gen/package.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/gen/package.scala
@@ -18,14 +18,11 @@
 package org.apache.kyuubi.plugin.spark.authz
 
 import scala.language.implicitConversions
-import scala.reflect.ClassTag
 
 import org.apache.kyuubi.plugin.spark.authz.OperationType.OperationType
 
 package object gen {
-  implicit def classSimpleName[T](implicit ct: ClassTag[T]): String = {
-    ct.runtimeClass.getSimpleName
-  }
+  implicit def classSimpleName[T](clz: Class[_]): String = clz.getSimpleName
 
   implicit def operationTypeStr(t: OperationType): String = t.toString
 }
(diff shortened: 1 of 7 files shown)


bowenliang123:
cc @yaooqinn

yaooqinn:
The extractor name has not necessarily to be simple classname

bowenliang123:
> The extractor name has not necessarily to be simple classname

Or let's change the method name to `classSimpleName` in AuthzUtils ?

yaooqinn:
Let's define an `implicit def` in the test package instead of `AuthZUtils`. Maybe a package object in `o.a.k.p.s.a.gen` or `JsonSpecFileGenerator`  

```scala  
implicit def classSimpleName[T](clz: Class[_]): String = clz.getSimpleName
```

And then, we can use it on the caller side like 
```scala
DatabaseDesc("databaseName", classOf[StringDatabaseExtractor])
```

Similarly, we can also have an `implicit def` for OperationType

```scala
implicit def operationTypeStr(t: OperationType): String = t.toString
```

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/gen/package.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz
+
+import scala.reflect.ClassTag
+
+package object gen {
+  implicit def classSimpleName[T](implicit ct: ClassTag[T]): String = {
```
yaooqinn, line 23: If you define it in such a way and call it explicitly, the `implicit` here is not necessary

bowenliang123: OK, let's use it in an implicit way as you suggested example above.

yaooqinn (approved):
LGTM

yaooqinn:
thanks. merged to master

PR #4098, opened 2023-01-05
Separate Trino lexer file

ulysses-you:

There is something difference between Spark and Trino with string pattern.
Spark:
```
STRING
    : '\'' ( ~'\'' | '\'\'' )* '\''
    | 'R\'' (~'\'')* '\''
    | 'R"'(~'"')* '"'
    ;
```

Trino:
```
STRING
    : '\'' ( ~'\'' | '\'\'' )* '\''
```

Since this is for Trino only, so we should follow Trino behavior.

Pass CI


ulysses-you:
cc @Yikf

kyuubi-server/src/main/antlr4/org/apache/kyuubi/sql/KyuubiSqlBaseLexer.g4
```diff
@@ -128,10 +127,9 @@ IDENTIFIER
     : [A-Za-z_$0-9\u0080-\uFFFF]*?[A-Za-z_$\u0080-\uFFFF]+?[A-Za-z_$0-9\u0080-\uFFFF]*
     ;
 
+// Follow Trino STRING
 STRING
     : '\'' ( ~'\'' | '\'\'' )* '\''
-    | 'R\'' (~'\'')* '\''
-    | 'R"'(~'"')* '"'
```
Yikf, line 134: It is also used in the kyuubi server parser

ulysses-you: Good point. Then we also can not modify it which may break server parser. Server parser should follow Spark. how about adding a `TRINO_STRING` so they are isolated.

Yikf: That's a good idea, but one of my concerns is that if have both `STRING` and `TRINO_STRING` in a lexer file, have a conflict problem, and you're going to have to think about the order of the two, but whatever the order, you're going to have a problem;

For `STRING`, Kyuubi has no strong dependency and needs to be consistent with Spark, in the following ways:

1. Keep the status quo, `STRING` is the `STRING` defined by Kyuubi;
2. Keep it consistent with Spark `STRING`. We use the following method to avoid conflicts with `ESCAPE`
```
fragment SEARCH_STRING_ESCAPE: '\'' '\\' '\'';

STRING_ESCAPE
    : SEARCH_STRING_ESCAPE
    ;

STRING
    : '\'' ( ~'\'' | '\'\'' )* '\''
    | 'R\'' (~'\'')* '\''
    | 'R"'(~'"')* '"'
    ;
```
But either way, there's a problem. here's an example, Both of the following SQL are parsed as `GetSchemas`:
```SQL
SELECT TABLE_SCHEM COMMA TABLE_CATALOG FROM SYSTEM_JDBC_SCHEMAS TABLE_CATALOG = 'test_catalog'

SELECT TABLE_SCHEM COMMA TABLE_CATALOG FROM SYSTEM_JDBC_SCHEMAS TABLE_CATALOG = "test_catalog"
```
But I think this is acceptable and users of `TRINO FE` will follow this. If we are worried about having a custom table with the same schema as the built-in table and querying the same SQL using Spark, only the stirng format is not consistent, we can cut the lexer file in the future. But I don't think that's the case.

@ulysses-you FYI, WDYT

ulysses-you: after offline discussion, we are going to separate Trino lexer file

kyuubi-server/src/main/antlr4/org/apache/kyuubi/sql/KyuubiTrinoFeBaseParser.g4
```diff
@@ -17,50 +17,46 @@
...
-    | TABLE_CAT EQ catalog=STRING+                                                                  #catalogFilter
+    : TABLE_CAT IS NULL                                                                                 #nullCatalog
+    | TABLE_CAT EQ catalog=STRING                                                                       #catalogFilter
     ;
 
 tableSchemaFilter
-    : TABLE_SCHEM IS NULL                                                                           #nulTableSchema
-    | TABLE_SCHEM LIKE schemaPattern=STRING+ ESCAPE SEARCH_STRING_ESCAPE                            #schemaFilter
+    : TABLE_SCHEM IS NULL                                                                               #nulTableSchema
+    | TABLE_SCHEM LIKE schemaPattern=STRING ESCAPE STRING                                               #schemaFilter
```
Yikf, line 52: emm, It should be `ESCAPE '\'` instead of `ESCAPE STRING`, `ESCAPE '\'` is fixed.

Like `... ESCAPE '\'` can be parsed to `GetSchemas`, but `... ESCAPE 'aaa'` can not.

kyuubi-server/src/main/scala/org/apache/kyuubi/sql/parser/trino/KyuubiTrinoFeAstBuilder.scala
```diff
@@ -112,6 +112,6 @@ class KyuubiTrinoFeAstBuilder extends KyuubiTrinoFeBaseParserBaseVisitor[AnyRef]
   }
 
   override def visitTypesFilter(ctx: TypesFilterContext): List[String] = {
-    ctx.stirngInValue().asScala.map(v => unescapeSQLString(v.getText)).toList
+    ctx.STRING().asScala.map(v => unescapeSQLString(v.getText)).toList
```
Yikf, line 115: maybe `stirngInValue ` is better, `STRING` is more like a singular string than a list, But I'ok for now change

Yikf:
Thanks! @ulysses-you  👍,  LGTM, left a few comment

Yikf (approved):

ulysses-you:
@Yikf  thanks for review, merging to master

PR #4105, opened 2023-01-06
[INFRA] Show  timestamp for GA test suite

turboFei:

Provide more insights if the GA unit test stuck for long time.


turboFei:
> does this break the static name for IDE debugger？

Thanks for the reminder, it did break the IDE debugger.

How about just show timestamp for the suite name?

yaooqinn:
Maybe we can have long enough global time limits for tests https://www.scalatest.org/scaladoc/3.0.6/org/scalatest/concurrent/TimeLimits.html

PR #4109, opened 2023-01-06
Enforce max Java bytecode version in Maven dependencies with `enforceBytecodeVersion` rule 

bowenliang123:

- Enforce max Java bytecode version in dependencies with `enforceBytecodeVersion` rule from MojoHaus‘s Extra Enforcer Rules (https://www.mojohaus.org/extra-enforcer-rules/enforceBytecodeVersion.html)
- Prevent accidentally bumping or introducing maven dependencies a Java version over baseline (currently Java 8)

Currently, some dependencies remain on older versions for Java 8 compatibility. This enforceBytecodeVersion  rule will help to enforce them aligning to Java 8. 

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -212,7 +212,7 @@
 
         <maven.plugin.build.helper.version>3.2.0</maven.plugin.build.helper.version>
         <maven.plugin.download.version>1.6.6</maven.plugin.download.version>
-        <maven.plugin.enforcer.mojo.extra.rules.version>1.6.1</maven.plugin.enforcer.mojo.extra.rules.version>
+        <maven.plugin.enforcer.mojo.rules.version>1.6.1</maven.plugin.enforcer.mojo.rules.version>
         <maven.plugin.scala.version>4.6.1</maven.plugin.scala.version>
         <maven.plugin.surefire.version>3.0.0-M7</maven.plugin.surefire.version>
         <maven.plugin.scalatest.version>2.0.2</maven.plugin.scalatest.version>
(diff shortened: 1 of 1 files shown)


bowenliang123:
For example, the lastest release version of Hikari is 5.0.1 required Java 11. If we bump it to 5.0.1, the enforcer will check `enforceBytecodeVersion` rule and fails with the error message.

```
[INFO] Restricted to JDK 8 yet com.zaxxer:HikariCP:jar:5.0.1:compile contains com/zaxxer/hikari/HikariConfig.class targeted to JDK 11
[ERROR] Rule 1: org.apache.maven.plugins.enforcer.EnforceBytecodeVersion failed with message:
Found Banned Dependency: com.zaxxer:HikariCP:jar:5.0.1
```

pom.xml
```diff
@@ -212,6 +212,7 @@
 
         <maven.plugin.build.helper.version>3.2.0</maven.plugin.build.helper.version>
         <maven.plugin.download.version>1.6.6</maven.plugin.download.version>
+        <maven.plugin.enforcer.mojo.extra.rules.version>1.6.1</maven.plugin.enforcer.mojo.extra.rules.version>
```
pan3793, line 215: maven.plugin.enforcer.extra.rules.version

pan3793 (approved):
LGTM, except to the property name

pom.xml
bowenliang123: `mojo` should be kept in the prop name, as `org.codehaus.mojo:extra-enforcer-rules` is third-party rules provider for mavn enforcer plugin.

pan3793: Under the premise of being identifiable, we prefer the short name.

bowenliang123: What about `maven.plugin.enforcer.mojo.rules.version` ?

pan3793:
Thanks, merged to master

Issue #4111, opened 2023-01-06
[Bug] An exception encounter "failed to generate operation log" if operation don't implement getOperationLog


zhaomin1423:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

If the operation don't implement the method of org.apache.kyuubi.operation.Operation#getOperationLog, will have a exception in the log file.

### Affects Version(s)

1.6.0/master

### Kyuubi Engine Log Output

```logtalk
2023-01-06 16:07:41.969 ERROR org.apache.kyuubi.engine.jdbc.JdbcTBinaryFrontendService: Error fetching results:
org.apache.kyuubi.KyuubiSQLException: OperationHandle [b354e589-0ad6-4537-944b-61f65715a24e] failed to generate operation log
	at org.apache.kyuubi.KyuubiSQLException$.apply(KyuubiSQLException.scala:69) ~[kyuubi-common_2.12-1.7.0-SNAPSHOT.jar:1.7.0-SNAPSHOT]
	at org.apache.kyuubi.operation.OperationManager.$anonfun$getOperationLogRowSet$2(OperationManager.scala:146) ~[kyuubi-common_2.12-1.7.0-SNAPSHOT.jar:1.7.0-SNAPSHOT]
	at scala.Option.getOrElse(Option.scala:189) ~[scala-library-2.12.17.jar:?]
	at org.apache.kyuubi.operation.OperationManager.getOperationLogRowSet(OperationManager.scala:146) ~[kyuubi-common_2.12-1.7.0-SNAPSHOT.jar:1.7.0-SNAPSHOT]
	at org.apache.kyuubi.session.AbstractSession.fetchResults(AbstractSession.scala:236) ~[kyuubi-common_2.12-1.7.0-SNAPSHOT.jar:1.7.0-SNAPSHOT]
	at org.apache.kyuubi.service.AbstractBackendService.fetchResults(AbstractBackendService.scala:204) ~[kyuubi-common_2.12-1.7.0-SNAPSHOT.jar:1.7.0-SNAPSHOT]
[...]

PR #4113, opened 2023-01-06
[KYUUBI #4111] JDBC ExecuteStatement operation should contain operationLog

Yikf:

Close https://github.com/apache/kyuubi/issues/4111, JDBC ExecuteStatement operation should contain operationLog

### DIFF
--- a/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteStatement.scala
@@ -23,6 +23,7 @@ import org.apache.kyuubi.engine.jdbc.schema.{Column, Row, Schema}
 import org.apache.kyuubi.engine.jdbc.session.JdbcSessionImpl
 import org.apache.kyuubi.engine.jdbc.util.ResultSetWrapper
 import org.apache.kyuubi.operation.{ArrayFetchIterator, IterableFetchIterator, OperationState}
+import org.apache.kyuubi.operation.log.OperationLog
 import org.apache.kyuubi.session.Session
 
 class ExecuteStatement(
@@ -32,6 +33,9 @@ class ExecuteStatement(
     queryTimeout: Long,
     incrementalCollect: Boolean)
   extends JdbcOperation(session) with Logging {
+
+  private val operationLog: OperationLog = OperationLog.createOperationLog(session, getHandle)
+  override def getOperationLog: Option[OperationLog] = Option(operationLog)
 
   override protected def runInternal(): Unit = {
     addTimeoutMonitor(queryTimeout)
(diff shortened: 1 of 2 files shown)


Yikf:
cc @zhaomin1423

zhaomin1423 (approved):
LGTM

turboFei (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala
```diff
@@ -142,8 +143,12 @@ abstract class OperationManager(name: String) extends AbstractService(name) {
       order: FetchOrientation,
       maxRows: Int): TRowSet = {
     val operationLog = getOperation(opHandle).getOperationLog
-    operationLog.map(_.read(maxRows)).getOrElse {
-      throw KyuubiSQLException(s"$opHandle failed to generate operation log")
+
```
turboFei, line 146: nit: not needed line

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala
```diff
@@ -142,8 +143,12 @@ abstract class OperationManager(name: String) extends AbstractService(name) {
...
+    operationLog match {
+      case Some(_) => operationLog.map(_.read(maxRows)).getOrElse {
```
turboFei, line 148:
```
case Some(log) => log.read(maxRows)
```

turboFei (approved):

yaooqinn:
Hmm, isn't the error handled when an operation log is created?

yaooqinn:
the original intentions are:
1. the operations that need operation logs shall implement getOperationLog
2. the client will only call fetching log for those operations which implement getOperationLog
3. the error raise only when the operation implements getOperationLog but fails to create the operation log file

zhaomin1423:
> the original intentions are:
> 
> 1. the operations that need operation logs shall implement getOperationLog
> 2. the client will only call fetching log for those operations which implement getOperationLog
> 3. the error raise only when the operation implements getOperationLog but fails to create the operation log file

2 can't work well now.

Yikf:
Yeah, 2 can not work well；

End-user can fetch logs through `OperationResource` as well, The client will only can fetch log for those operations which implement `getOperationLog` seems impossible to limit?

yaooqinn:
After this change, we can not determine whether the log is failed or just not implemented

yaooqinn:
> 2 can't work well now.

**This is by contract**, these OperationResouce's clients shall be able to benefit from the current implementation too.

Yikf:
> After this change, we can not determine whether the log is failed or just not implemented

Good point, Have a question is which of the following designs is more reasonable:
1. Fetch empty log if the operation don't implements getOperationLog
2. Have the contract, client will only call fetching log for those operations which implement getOperationLog

If 1, we can handle whether the log is failed or just not implemented; 
if 2, we can handle this case according to this contract at client/jdbc engine side;

> shall be able to benefit from the current implementation too.

What are the benefits? (sorry for I didn't know the previous design ) : )

yaooqinn:
> What are the benefits? (sorry for I didn't know the previous design ) : )

Well, the thrift and rest calls have no differences on both the log API and the backend operations. It's up to much higher level APIs(such as JDBC), or end-users, to do the error handling.

Currently, the meta operations are no operation logs and are executed synchronously. The rest call can also invoke these operations synchronously, which makes it has no chance to call log fetching

Yikf:
I fix this issue at jdbc engine side, please take a look if you find a moment, thanks @yaooqinn @turboFei @zhaomin1423

yaooqinn (approved):

turboFei (approved):

zhaomin1423 (approved):

turboFei:
thanks, merged to master

PR #4022, opened 2022-12-23
[KYUUBI #3968] Upgrading and migration script for Jdbc

lightning-L:
### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #4119, opened 2023-01-07
Return app submission time for batch

turboFei:

Return the batch app submission time, it is useful to provide the app elapse time.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -158,7 +158,7 @@ class BatchJobSubmission(
         session.getSessionEvent.foreach(_.engineId = ai.id)
       }
       if (newState == RUNNING) {
-        session.onEngineOpened(_appSubmissionTime)
+        session.onEngineOpened()
       }
     }
   }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -74,6 +74,9 @@ class BatchJobSubmission(
   private var killMessage: KillResponse = (false, "UNKNOWN")
   def getKillMessage: KillResponse = killMessage
 
+  @volatile private var _applicationSubmissionTime: Long = _
+  def applicationSubmissionTime: Long = _applicationSubmissionTime
+
   @VisibleForTesting
   private[kyuubi] val builder: ProcBuilder = {
     Option(batchType).map(_.toUpperCase(Locale.ROOT)) match {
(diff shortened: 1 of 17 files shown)


turboFei:
also cc @lightning-L 
This pr has metadata schema change, it is blocked by #4022

pan3793 (approved):

turboFei:
thanks, merged to master

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -96,7 +99,14 @@ class BatchJobSubmission(
 
   override private[kyuubi] def currentApplicationInfo: Option[ApplicationInfo] = {
     // only the ApplicationInfo with non-empty id is valid for the operation
-    applicationManager.getApplicationInfo(builder.clusterManager(), batchId).filter(_.id != null)
+    val applicationInfo =
+      applicationManager.getApplicationInfo(builder.clusterManager(), batchId).filter(_.id != null)
+    applicationInfo.foreach { _ =>
+      if (_appSubmissionTime <= 0) {
+        _appSubmissionTime = System.currentTimeMillis()
```
ulysses-you, line 106: Seems it's the `appStartTime` ? Submisstion time is the time before submiting succeed.

ulysses-you: why we donot set it in `submitAndMonitorBatchJob` method ?

turboFei: > why we donot set it in submitAndMonitorBatchJob method ?

Because in the `submitAndMonitorBatchJob` method, there is interval to track the app state.
And for the `getBatch` api, it will also call currentApplicationState to build the batch report, to keep the state align, it is better to update the app submission time in `currentApplicationState`.

ulysses-you: I think we should make state machine and the time trace consistent.

If the submitssion time means when the state changes to running, then we should update the time around this code
```
setStateIfNotCanceled(OperationState.RUNNING)
```

Otherwise the client may get the batch info with `submisstion time = 0 but state is running` or `submisstion time != 0 but state is pending`.

turboFei: for batch report, it is consistent with the appId/appState/appState now.

turboFei: We can treat it as a filed of batch application info.

turboFei: here is the pr for followup https://github.com/apache/kyuubi/pull/4129

PR #4144, opened 2023-01-11
[KYUUBI #4106] Introduce resource file uploading in batch creation via REST API

### LINKED ISSUE #4106: [FEATURE] Support upload app jar file in POST /batches
### Code of Conduct

### Search before asking

### Describe the feature

Currently In REST Api, `POST /batches` (https://kyuubi.readthedocs.io/en/latest/client/rest/rest_api.html#post-batches)  only 
 supports executing a path to the resource in external storage (eg. hdfs, s3 and etc).

Supporting uploading the jar file in this API will help users avoid uploading the jar file to other external storage first.

- support upload jar file in POST request (but it may violate the JSON mediaType, need to add a new API maybe)
- put jar file in temporary folder on server
- submit job using the local path on server to jar file in above
- clean up temp jar file after submission or finishing execution

### Motivation

_No response_

### Describe the solution

_No response_

### Additional context

_No response_

bowenliang123:

to close #4106 .

1. add `POST /batches`API in `BatchesResource` of REST API, which consumes `multipart/form-data` media type to support uploading `resourcefile`. And a `batchRequest` form data part in the JSON format string of `BatchRequest`, as in the required request body of `POST /batches`
2. the uploaded `resourceFile` is saved to a temp local file which will be cleaned up after job execution at the end of  `submitAndMonitorBatchJob`
3. the local temp copy of `resourceFile` will be used as `resource` in `BatchJobSubmission`, eg. as <application-jar> for spark-submit

Todos in follow-up:
1. add a related description in Rest API doc
4. add `multipart` media type support to `RestClient` and implement `createBatchWithUploadingResource` in `BatchRestApi`


dev/dependencyList
```diff
@@ -100,6 +102,12 @@ jetty-util-ajax/9.4.50.v20221201//jetty-util-ajax-9.4.50.v20221201.jar
 jetty-util/9.4.50.v20221201//jetty-util-9.4.50.v20221201.jar
 jline/0.9.94//jline-0.9.94.jar
 jul-to-slf4j/1.7.36//jul-to-slf4j-1.7.36.jar
+junit-jupiter-api/5.9.1//junit-jupiter-api-5.9.1.jar
```
pan3793, line 105: I think it does not make sense to ship testing jar

bowenliang123: Agree.  This is a draft version only, to show the possible change or impact in API and dependencies.
Will try to refine the dependency list later.

bowenliang123: Addressed. Excluded `junit-jupiter` in `jersey-media-multipart` mannually.
This scope bug exists in ~2.38 and 3.x , till this fix in unreleased 2.39 (https://github.com/eclipse-ee4j/jersey/commit/3b182ebdf35ce471e63bdcae1090cce842b933f1).

bowenliang123:
cc @turboFei @pan3793

turboFei:
some concern:
1. how to recover the batch job for k8s env if the pod recreate
   - save the uploaded file on hdfs to persistent?

bowenliang123:
In that case, uploading to storage should be done before executing it.
And executing resource in external storage is already supported with `POST /batches`.

This main conern of this PR with to provide both uploading and cleaning up support for temporary resources files.
Cleaning up after execution is good enough in this initial support without introducing complex in uploading to external storages(eg. auth,method,client,protocol and etc.), I think.

Or any suggestion for this ? @pan3793

bowenliang123:
Tests pased on my local machine, but getting error in ci runs.

https://github.com/apache/kyuubi/actions/runs/3908113144/jobs/6677980088
```
Cause: com.fasterxml.jackson.core.JsonParseException: Unexpected character ('h' (code 104)): was expecting either '*' or '/' for a comment
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 3]
```

bowenliang123:
This PR should be ready for review., please have check if you have time. Thx. @turboFei  @pan3793

 Spent some time to fix the exception in CI test run.
This PR only introduces a new Rest API on the server side, and the client support could be done in the followup  since multipart POST is not supported in the current implementation of http client wrapping.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -135,6 +135,19 @@ object Utils extends Logging {
...
+   * Create a temporary file inside the given parent directory. The file will be
+   * automatically deleted when the VM shuts down.
+   */
+  def createTempFile(
+      dir: Path,
+      prefix: String,
+      suffix: String): File = {
+    val f = Files.createTempFile(dir, prefix, suffix).toFile
+    f.deleteOnExit()
+    f
```
turboFei, line 148: seems this code is not used.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
```diff
@@ -32,6 +32,9 @@ public class BatchRequest {
   private Map<String, String> conf;
   private List<String> args;
 
+  /** whether clean up temporary uploaded resource file in local path after execution */
```
turboFei, line 35: whether to clean up

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
```diff
@@ -32,6 +32,9 @@ public class BatchRequest {
...
+  private Boolean cleanupResourceFile = false;
```
turboFei, line 36: Seems we do not need expose this field in API, just always remove uploaded files?

turboFei: it can be an internal option for upload case.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -125,7 +132,7 @@ object KyuubiApplicationManager {
             s"Relative path ${uri.getPath} is not allowed, please use absolute path.")
         }
 
-        if (!localDirAllowList.exists(uri.getPath.startsWith(_))) {
+        if (!(localDirAllowList :+ tempDirForUpload.toString).exists(uri.getPath.startsWith(_))) {
```
turboFei, line 135: seems not needed, we can check whether the resource is uploaded in `BatchJobSubmission`.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -57,7 +60,8 @@ class BatchJobSubmission(
     className: String,
     batchConf: Map[String, String],
     batchArgs: Seq[String],
-    recoveryMetadata: Option[Metadata])
+    recoveryMetadata: Option[Metadata],
+    isCleanupResourceFile: Boolean)
```
turboFei, line 64: rename to a variable to present the resource is a temporary uploaded one? 
1. cleanup it after batch terminated
2. do not check the resource for this case, see `KyuubiBatchSessionImpl::checkSessionAccessPathURIs`

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -300,6 +308,20 @@ class BatchJobSubmission(
     }
   }
 
+  private def deleteTempLocalResourceFile(filePath: String): Unit = {
+    try {
+      val uri = new URI(filePath)
+      uri.getScheme match {
```
turboFei, line 314: seems it can be only local file, because the uploaded resource is internal and full control.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -300,6 +308,20 @@ class BatchJobSubmission(
...
+        case null | "file" =>
+          FileUtils.deleteQuietly(Paths.get(filePath).toFile)
+        case _ =>
+          warn(s"only local path supported, skip deleting file $filePath")
+      }
+    } catch {
+      case e: Exception => warn(e)
```
turboFei, line 321: log more helpful information, including the resource path and action.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
```diff
@@ -81,7 +81,8 @@ class KyuubiOperationManager private (name: String) extends OperationManager(nam
       className: String,
       batchConf: Map[String, String],
       batchArgs: Seq[String],
-      recoveryMetadata: Option[Metadata]): BatchJobSubmission = {
+      recoveryMetadata: Option[Metadata],
+      isCleanupResourceFile: Boolean): BatchJobSubmission = {
```
turboFei, line 85: how about get it from `KyuubiBatchSessionImpl`, because we need to check the option to determine whether check the resource path in `KyuubiBatchSessionImpl` as well.

turboFei: how about use a reserved key to present whether the resource file is uploaded.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -187,6 +191,29 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+      mediaType = MediaType.APPLICATION_JSON,
+      schema = new Schema(implementation = classOf[Batch]))),
+    description = "create and open a batch session with uploading resource file")
+  @POST
+  @Path("/upload")
+  @Consumes(Array(MediaType.MULTIPART_FORM_DATA))
+  def openBatchSessionWithUpload(
+      @FormDataParam("batchRequest") batchRequestStr: String,
+      @FormDataParam("resourceFile") resourceFileInputStream: InputStream,
+      @FormDataParam("resourceFile") resourceFileMetadata: FormDataContentDisposition): Batch = {
```
turboFei, line 206: is it duplicated？

turboFei: I mean there are two FormDataParams with the same name `resourceFile `.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -80,6 +80,7 @@ class KyuubiBatchSessionImpl(
   override lazy val name: Option[String] = Option(batchRequest.getName).orElse(
     normalizedConf.get(KyuubiConf.SESSION_NAME.key))
 
+  // todo: support uploaded app jar
   private[kyuubi] lazy val batchJobSubmissionOp = sessionManager.operationManager
```
turboFei, line 84: how about moving this TODO to `BatchResource` or just remove it?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -89,7 +90,8 @@ class KyuubiBatchSessionImpl(
       batchRequest.getClassName,
       normalizedConf,
       batchRequest.getArgs.asScala,
-      recoveryMetadata)
+      recoveryMetadata,
+      batchRequest.isCleanupResourceFile)
```
turboFei, line 94: refer the `sessionName` method of KyuubiBatchSessionImpl.

kyuubi-server/src/main/scala/org/apache/kyuubi/util/FileUtil.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.hadoop.shaded.org.apache.commons.io.FileUtils
+
+object FileUtil {
+
+  /**
+   * write InputStream to temp file in directory
+   * @param is
+   * @param dir
+   * @param fileName
+   * @return File
```
turboFei, line 34: if there is no meaning description for the params, you can remove it.

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/BatchesResourceSuite.scala
```diff
@@ -199,6 +203,31 @@ class BatchesResourceSuite extends KyuubiFunSuite with RestFrontendTestHelper wi
     assert(!deleteBatchResponse.readEntity(classOf[CloseBatchResponse]).isSuccess)
   }
 
+  test("open batch session with uploading resource") {
+    val requestObj = newSparkBatchRequest(Map("spark.master" -> "local"))
+    val requestStr = mapper.writeValueAsString(requestObj)
+    val exampleJarFile = {
+      val sparkProcessBuilder = new SparkProcessBuilder("kyuubi", KyuubiConf())
```
turboFei, line 210: using `sparkBatchTestResource` directly. 

See `BatchTestHelper`.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: Seems both are required, as instruction from Jerysey's doc here(https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/media.html#d0e9649).
One is for getting input stream and the other for uploaded file's metadata.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: Removed. Thx for the catch.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
bowenliang123: Addressed.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
bowenliang123: Agree and addressed. Extracted a dedicated method `openBatch` for the use in both `POST /bathes` and `POST /bathes/upload`  with internal control `isCleanupResourceFile` option.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
bowenliang123: Addressed. Reverted this method by removing tempDirForUpload.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
bowenliang123: Agree. And changed to use `FileUtils.deleteQuietly` directly, with no exception need to be catched.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala
bowenliang123: Changed to get the `isCleanupResourceFile` option from `KyuubiBatchSessionImpl`.

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/BatchesResourceSuite.scala
bowenliang123: This's a helpful hint. Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
bowenliang123: `isCleanupResourceFile` is no longer needed here. Removed.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -19,14 +19,18 @@ package org.apache.kyuubi.engine
 
 import java.io.File
 import java.net.{URI, URISyntaxException}
+import java.nio.file.Path
 import java.util.{Locale, ServiceLoader}
 
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
 
+import org.apache.hadoop.shaded.org.apache.commons.io.FileUtils
```
turboFei, line 28:
```
import org.apache.commons.io.FileUtils
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -58,6 +62,7 @@ class KyuubiApplicationManager extends AbstractService("KyuubiApplicationManager
         case NonFatal(e) => warn(s"Error stopping ${op.getClass.getSimpleName}: ${e.getMessage}")
       }
     }
+    FileUtils.deleteQuietly(tempDirForUpload.toFile)
```
turboFei, line 65:
```
    FileUtils.deleteQuietly(KyuubiApplicationManager.tempDirForUpload.toFile)
```

and remove
```
import org.apache.kyuubi.engine.KyuubiApplicationManager.tempDirForUpload
```

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -18,11 +18,13 @@
 package org.apache.kyuubi.operation
 
 import java.io.IOException
+import java.nio.file.Paths
 import java.util.Locale
 import java.util.concurrent.TimeUnit
 
 import com.codahale.metrics.MetricRegistry
 import com.google.common.annotations.VisibleForTesting
+import org.apache.hadoop.shaded.org.apache.commons.io.FileUtils
```
turboFei, line 27: ditto

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -267,6 +269,10 @@ class BatchJobSubmission(
       }
     } finally {
       builder.close()
+      if (session.isResourceFromUpload) {
+        info(s"cleanup uploaded resource file after execution: $resource")
+        FileUtils.deleteQuietly(Paths.get(resource).toFile)
```
turboFei, line 274: It is better to provide more info.
```
if (!(FileUtils.deleteQuietly(Paths.get(resource).toFile)) {
  error(...)
}
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
turboFei:
```
if (!(FileUtils.deleteQuietly(Paths.get(resource).toFile)) {
  error(...)
}
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
```diff
@@ -22,8 +22,12 @@ import javax.ws.rs.ext.ContextResolver
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 
-class KyuubiScalaObjectMapper extends ContextResolver[ObjectMapper] {
-  private val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
+import org.apache.kyuubi.server.api.KyuubiScalaObjectMapper.mapper
```
turboFei, line 25: ditto, you can import `KyuubiScalaObjectMapper._` in the class

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
bowenliang123: `commons-io` is not available in the server module, shall we add this to dependency?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -43,6 +43,7 @@ class KyuubiBatchSessionImpl(
     override val sessionManager: KyuubiSessionManager,
     val sessionConf: KyuubiConf,
     batchRequest: BatchRequest,
+    val isResourceFromUpload: Boolean,
```
turboFei, line 46: private[kyuubi] val isResourceFromUpload = batchRequest.getConf.asScala.get(KyuubiReservedKeys.....).map(_.toBoolean).getOrElse(false)

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -183,7 +222,8 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       "anonymous",
       ipAddress,
       request.getConf.asScala.toMap,
-      request)
+      request,
+      isResourceFromUpload)
```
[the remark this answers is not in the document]
turboFei: see KyuubiReservedKeys

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -134,7 +134,8 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       ipAddress: String,
       conf: Map[String, String],
       batchRequest: BatchRequest,
-      recoveryMetadata: Option[Metadata] = None): KyuubiBatchSessionImpl = {
+      recoveryMetadata: Option[Metadata] = None,
+      isResourceFromUpload: Boolean = false): KyuubiBatchSessionImpl = {
```
turboFei, line 138: with reserved key, you do not need change the constructor per time.

kyuubi-server/src/main/scala/org/apache/kyuubi/util/FileUtil.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.apache.kyuubi.util
+
+import java.io.{File, InputStream}
+import java.nio.file.Path
+
+import scala.util.Random
+
+import org.apache.hadoop.shaded.org.apache.commons.io.FileUtils
```
turboFei, line 25: ditto

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
bowenliang123: Addressed.
1. prefer the direct way `FileUtils.deleteQuietly(KyuubiApplicationManager.tempDirForUpload.toFile)` with less imports
2. use `FileUtils.deleteDirectory` and log exception instead of `FileUtils.deleteQuietly` which doesn't throw exceptions.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -94,6 +94,18 @@ class KyuubiApplicationManager extends AbstractService("KyuubiApplicationManager
       case None => None
     }
   }
+
+  private def deleteTempDirForUpload(): Unit = {
+    try {
+      FileUtils.deleteDirectory(KyuubiApplicationManager.tempDirForUpload.toFile)
```
turboFei, line 100: why not use Utils.deleteDirectoryRecursively

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
turboFei: If that, you can use Utils.deleteDirectoryRecursively instead.

bowenliang123: Added `commons-io` dependency to server module, and use it in FileUtils replacing `import org.apache.hadoop.shaded.org.apache.commons.io.FileUtils`.
.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
bowenliang123: Addressed. Delete file and logging error with added `Utils.deleteFile` method.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
bowenliang123: Addressed. Thx.

kyuubi-server/src/main/scala/org/apache/kyuubi/util/FileUtil.scala
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
bowenliang123: Yes, it's now using `Utils.deleteDirectoryRecursively` here.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
bowenliang123: Agree, it's a very good suggestion. Changed to setting a new reserved key in session configs instead.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
```diff
@@ -22,8 +22,12 @@ import javax.ws.rs.ext.ContextResolver
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 
-class KyuubiScalaObjectMapper extends ContextResolver[ObjectMapper] {
-  private val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
+import org.apache.kyuubi.server.api.KyuubiScalaObjectMapper._
 
+class KyuubiScalaObjectMapper extends ContextResolver[ObjectMapper] {
   override def getContext(aClass: Class[_]): ObjectMapper = mapper
```
turboFei, line 28: = KyuubiScalaObjectMapper.mapper

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
```diff
@@ -22,8 +22,12 @@ import javax.ws.rs.ext.ContextResolver
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 
-class KyuubiScalaObjectMapper extends ContextResolver[ObjectMapper] {
-  private val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
+import org.apache.kyuubi.server.api.KyuubiScalaObjectMapper._
```
turboFei, line 25: remove it

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -182,7 +182,13 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       ipAddress: String,
       conf: Map[String, String],
       batchRequest: BatchRequest): SessionHandle = {
-    val batchSession = createBatchSession(user, password, ipAddress, conf, batchRequest)
+    val batchSession =
```
turboFei, line 185: unneeded change

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -80,6 +80,12 @@ class KyuubiBatchSessionImpl(
   override lazy val name: Option[String] = Option(batchRequest.getName).orElse(
     normalizedConf.get(KyuubiConf.SESSION_NAME.key))
 
+  /**
+   * whether the resource file is from uploading
+   */
+  lazy val isResourceUploaded: Boolean =
```
turboFei, line 86: prefer single line comment here . 
```
// 
```

turboFei: it should be fine to remove lazy

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
```diff
@@ -83,8 +83,9 @@ class KyuubiBatchSessionImpl(
   /**
    * whether the resource file is from uploading
    */
-  lazy val isResourceUploaded: Boolean =
-    conf.getOrElse(KyuubiReservedKeys.KYUUBI_SESSION_BATCH_RESOURCE_UPLOADED_KEY, "false").toBoolean
+  lazy val isResourceUploaded: Boolean = batchRequest.getConf
+    .getOrDefault(KyuubiReservedKeys.KYUUBI_SESSION_BATCH_RESOURCE_UPLOADED_KEY,
+      "false").toBoolean
```
turboFei: seems the line length is sufficient.

```
  lazy val isResourceUploaded: Boolean = batchRequest.getConf
    .getOrDefault(KyuubiReservedKeys.KYUUBI_SESSION_BATCH_RESOURCE_UPLOADED_KEY, "false").toBoolean
```

bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/util/FileUtil.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * write InputStream to temp file in directory
+   */
+  def writeToTempFile(is: InputStream, dir: Path, fileName: String): File = {
+    val prefix = GFiles.getNameWithoutExtension(fileName)
+    val suffix = GFiles.getFileExtension(fileName)
+    val file = new File(dir.toFile, s"$prefix${Random.nextLong()}.$suffix")
+    file.deleteOnExit()
+    JFiles.copy(is, file.toPath, StandardCopyOption.REPLACE_EXISTING);
```
bowenliang123, line 38: Changed to use Java `Files` here, without using `commons-io`'s `FileUtils`. @turboFei

kyuubi-server/src/main/scala/org/apache/kyuubi/util/FileUtil.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * write InputStream to temp file in directory
+   */
+  def writeToTempFile(is: InputStream, dir: Path, fileName: String): File = {
+    val prefix = GFiles.getNameWithoutExtension(fileName)
+    val suffix = GFiles.getFileExtension(fileName)
+    val file = new File(dir.toFile, s"$prefix${Random.nextLong()}.$suffix")
+    file.deleteOnExit()
+    JFiles.copy(is, file.toPath);
```
bowenliang123, line 38: Changed to use Java's `Files` here, without using `commons-io`'s `FileUtils`. @turboFei

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSessionImpl.scala
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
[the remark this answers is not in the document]
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
bowenliang123: Addressed.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -147,6 +163,36 @@ object Utils extends Logging {
...
   }
 
+  /**
+   * write InputStream to temp file in directory
+   * @param is inputstream for file
+   * @param dir dir path for temp file creation
+   * @param fileName original file name with suffix
+   * @return created new temp file in dir in file name with random path
+   */
+  def writeToTempFile(is: InputStream, dir: Path, fileName: String): File = {
```
turboFei, line 173: is the inputStream closed finally?

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -147,6 +163,36 @@ object Utils extends Logging {
     dir
   }
 
+  /**
+   * write InputStream to temp file in directory
+   * @param is inputstream for file
+   * @param dir dir path for temp file creation
+   * @param fileName original file name with suffix
+   * @return created new temp file in dir in file name with random path
```
turboFei, line 171: > in dir in file name

seems typo

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -147,6 +163,36 @@ object Utils extends Logging {
...
+      }
+      val filePath = Paths.get(dir.toString, s"$prefix-${abs(Random.nextInt())}$suffix")
+      Files.copy(is, filePath, StandardCopyOption.REPLACE_EXISTING)
+      val file = filePath.toFile
+      file.deleteOnExit()
+      file
+    } catch {
+      case e: Exception =>
+        error(
+          s"failed to write to temp file in path ${dir.toString}, original file name: $fileName",
```
turboFei, line 190: ${dir.toString} -> $dir

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
pan3793, line 173: Please clearly document WHO is responsible to close the `is`, the caller, or this method? and what will happen if the thing goes wrong?

pan3793, line 173: please document WHO is responsible to close the `is`, caller, or this method? what will happen if things go wrong?

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -147,6 +163,36 @@ object Utils extends Logging {
...
+  def writeToTempFile(is: InputStream, dir: Path, fileName: String): File = {
+    try {
+      if (!dir.toFile.exists()) {
+        dir.toFile.mkdirs()
+      }
+      val (prefix, suffix) = fileName.lastIndexOf(".") match {
+        case i if i > 0 => (fileName.substring(0, i), fileName.substring(i))
+        case _ => (fileName, "")
+      }
+      val filePath = Paths.get(dir.toString, s"$prefix-${abs(Random.nextInt())}$suffix")
```
turboFei, line 182: I recommend to use the below pattern.

```
rootDir/sessionId/fileName
```

turboFei: otherwise, it is difficult to scale out if we want to support upload more files not only resource file.

turboFei: BTW, now the pattern with random, I wonder there might be duplicate issue.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -135,6 +137,20 @@ object Utils extends Logging {
...
+   * delete file in path with logging
+   * @param filePath path to file for deletion
+   * @param errorMessage message as prefix logging with error exception
+   */
+  def deleteFile(filePath: String, errorMessage: String): Unit = {
+    try {
+      Files.delete(Paths.get(filePath))
+    } catch {
+      case e: Exception =>
+        error(s"$errorMessage: $filePath ", e)
```
turboFei, line 150: I think that you can remove the errorMessage here.

```
error(s"Failed to delete file: $file", e)
```

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
turboFei: and delete the `rootDir/sessionId` dir finally instead of delete the single file.

turboFei: seems you can get the sessionId here.

Then I recommend to use UUID instead of random int.

turboFei: cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -161,6 +164,41 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+    responseCode = "200",
+    content = Array(new Content(
+      mediaType = MediaType.APPLICATION_JSON,
+      schema = new Schema(implementation = classOf[Batch]))),
+    description = "create and open a batch session with uploading resource file")
+  @POST
+  @Path("/upload")
+  @Consumes(Array(MediaType.MULTIPART_FORM_DATA))
+  def openBatchSessionWithUpload(
+      @FormDataParam("batchRequest") batchRequestStr: String,
```
pan3793, line 180: why use `text/plain` to accept the JSON data instead of `application/json`?

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: Sorry, let fix this problem with properly closing the InputStream.

bowenliang123: OK, I will make sure it's closed in this problem and add this description to the doc.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: Addressed and change to get deserialized BatchRequest within JSON part in multipart request.
That would be much better. Thx for the hint.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: Addressed.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: Addressed.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: UUID might be too long for the random identifier here, considering the original filename is preserved in new file name.
To avoid dumplication issue and increase readiblity, 
how about `rootDir/${original_preffix}-${datetime_YYYYMMDDHHMMSS}-${randomInt}.${original_suffix}`?
@pan3793 @turboFei

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: Docs updated.

bowenliang123: Implementation updated with closing input stream after copy.

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/BatchesResourceSuite.scala
```diff
@@ -207,7 +207,7 @@ class BatchesResourceSuite extends KyuubiFunSuite with RestFrontendTestHelper wi
     val requestStr = mapper.writeValueAsString(requestObj)
     val exampleJarFile = Paths.get(sparkBatchTestResource.get).toFile
     val multipart = new FormDataMultiPart()
-      .field("batchRequest", requestStr)
+      .field("batchRequest", requestStr, MediaType.APPLICATION_JSON_TYPE)
```
pan3793, line 210: "requestStr" => "requestObj"?

bowenliang123: Good, fixed.

bowenliang123:
The latest commits cover all the points from the discussion. Thanks for all the help and detailed hints or suggestions.
Happy Chinese New Year and best wishes to you guys!

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -161,6 +163,40 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+  }
+
+  @ApiResponse(
+    responseCode = "200",
+    content = Array(new Content(
+      mediaType = MediaType.APPLICATION_JSON,
+      schema = new Schema(implementation = classOf[Batch]))),
+    description = "create and open a batch session with uploading resource file")
+  @POST
+  @Path("/upload")
```
pan3793, line 176: I'm a little bit concerned w/ this path since it's not only uploading the JAR but also creating the batch job. Comparing the existing API, the differences are
1) the accepted media type is multipart/form-data
2) it allows(or requires) the JAR as a parameter

So, would it be good to use the same path w/ opensession? Since they have different consume media types, the framework should correctly dispatch the request to the method.

cc @turboFei

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
```diff
@@ -23,7 +23,9 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 
 class KyuubiScalaObjectMapper extends ContextResolver[ObjectMapper] {
-  private val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
+  override def getContext(aClass: Class[_]): ObjectMapper = KyuubiScalaObjectMapper.mapper
+}
 
-  override def getContext(aClass: Class[_]): ObjectMapper = mapper
+object KyuubiScalaObjectMapper {
+  val mapper: ObjectMapper = new ObjectMapper().registerModule(DefaultScalaModule)
```
turboFei, line 30: you can revert the change here now.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
turboFei: I recommend you reserve an accumulator instead of randomInt.

Likes yarn applicationId(application_$clusterTimestamp_$accId).

turboFei: randomInt is meaningless.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
turboFei: nice catch, the `upload` api looks does not make senses.

could you try to use `POST /sessions` and check whether it works?

bowenliang123: Yes, different accepting media type with @Consumes with same combination  HTTP method and path can be handled in different actions. 

Let's make it in both handlers with same path:
- `POST /batches` for `application/json` media type: accepting BatchRequest in JSON format from HTTP request body
- `POST /batches` for `multipart/form-data` media type`:  accepting two parts from body, including the `batchRequest` form data parm for BatchRequest in JSON format, and the other form part `resourceFile` of uploaded resource file.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
bowenliang123: OK, good catch, reverted all changes in KyuubiScalaObjectMapper.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: The accumulator is a better option, while it also could be cleaned up after server's shutdown.

turboFei: It does not matter， there is a timestamp.

bowenliang123: Addressed. Turned to using an atomic counter for file name identifier generation instead.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: Applied changes as discussed above.

bowenliang123:
OK, all the discussions are now resolved.

pan3793 (approved):
Thanks for the great work, LGTM on my side

turboFei (approved):
LGTM

bowenliang123:
Thanks for all the detailed reviews and productive suggestions.
Merged to master.

turboFei:
After think twice:

1. do not cleanup the uploader dir when service stop, because the uploaded file will be deleted after batch completion, otherwise, it can not be recovered
2. do not change the file name when uploading if you want to upload file/jars later
   - `uploadRootDir/timestamp-${count}/${file}

turboFei:
cc @bowenliang123 @pan3793  please check above comments

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
```diff
@@ -24,6 +24,7 @@ object KyuubiReservedKeys {
   final val KYUUBI_SESSION_SIGN_PUBLICKEY = "kyuubi.session.sign.publickey"
   final val KYUUBI_SESSION_USER_SIGN = "kyuubi.session.user.sign"
   final val KYUUBI_SESSION_REAL_USER_KEY = "kyuubi.session.real.user"
+  final val KYUUBI_SESSION_BATCH_RESOURCE_UPLOADED_KEY = "kyuubi.session.batch.resource.uploaded"
```
pan3793, line 27: Should it be `kyuubi.session.batch` instead of `kyuubi.batch.session`? As we already have `kyuubi.batch.session.idle.timeout`, they should be consistent.

cc @bowenliang123 @turboFei

pan3793: Or just remove "session", and use the "kyuubi.batch." namespace.

turboFei: SGTM @bowenliang123 @pan3793

PR #4150, opened 2023-01-12
Lock scala

turboFei:

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala
@@ -101,7 +101,9 @@ private[spark] case class KyuubiSparkILoop private (
 
   def clearResult(statementId: String): Unit = result.unset(statementId)
 
-  def interpretWithRedirectOutError(statement: String): IR.Result = withLockRequired {
+  def interpretWithRedirectOutError(
+      statement: String,
+      lockRequired: Boolean): IR.Result = withLockRequired(lockRequired) {
     Console.withOut(output) {
       Console.withErr(output) {
         this.interpret(statement)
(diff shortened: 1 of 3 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2255,6 +2255,13 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val OPERATION_SPARK_SCALA_LOCK_ENABLED: ConfigEntry[Boolean] =
```
yaooqinn, line 2258: synchronized directly

turboFei:
CC @cxzl25

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala
```diff
@@ -120,4 +123,16 @@ private[spark] object KyuubiSparkILoop {
     iLoop.initialize()
     iLoop
   }
+
+  private val lock = new ReentrantLock()
+  private[spark] def withLockRequired[T](lockRequired: Boolean)(block: => T): T = {
```
cxzl25, line 128: Also add a lock here?

https://github.com/apache/kyuubi/blob/3f8189994b9bd3d36b050f3c4334302a21b0b603/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala#L84-L87

yaooqinn:
The configuration is not needed, we shall always interpret code sequentially

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/KyuubiSparkILoop.scala
turboFei: nice catch, resolved.

turboFei:
> The configuration is not needed, we shall always interpret code sequentially

done.

cxzl25 (approved):

Yikf (approved):

yaooqinn (approved):

turboFei:
thanks, merged to master and branch-1.6

PR #4151, opened 2023-01-12
Support to ignore subdomain when engine pool conditions are met

turboFei:

Support to ignore subdomain when engine pool conditions are met.
It is difficult to understand the engine routing rule for customers, especially for the engine pool and subdomain.
Maybe a user use subdomain before, and then they want to use engine pool.
Then specify the engine pool name, engine pool size, but the result is not expected.
They need to remove the subdomain specified before, otherwise, the engine pool will not work.
I need tell customer that, the engine subdomain config has higher priority and tell them the routing rule.
The understanding cost is high.
With this change, it is clear for engine pool conditions.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -1652,11 +1652,13 @@ object KyuubiConf {
     .checkValues(EngineType.values.map(_.toString))
     .createWithDefault(EngineType.SPARK_SQL.toString)
 
-  val ENGINE_POOL_FIRST: ConfigEntry[Boolean] = buildConf("kyuubi.engine.pool.first")
-    .doc("Used for arbitration when engine pool conditions are met and subdomain is specified.")
-    .version("1.7.0")
-    .booleanConf
-    .createWithDefault(false)
+  val ENGINE_POOL_IGNORE_SUBDOMAIN: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.pool.ignoreSubdomain")
+      .doc("Whether to ignore subdomain when engine pool conditions are met.")
+      .internal
+      .version("1.7.0")
+      .booleanConf
+      .createWithDefault(false)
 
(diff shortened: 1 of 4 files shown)


turboFei:
how do you think about? @yaooqinn  @pan3793

pan3793:
what's the background of this change?

turboFei:
> what's the background of this change?


It is difficult to understand the engine routing rule for customers, especially for the engine pool and subdomain.


Maybe a user use subdomain before, and then they want to use engine pool.

Then specify the engine pool name, engine pool size, but the result is not expected.

They need to remove the subdomain specified before, otherwise, the engine pool will not work.

I need tell customer that, the engine subdomain config has higher priority.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1652,6 +1652,12 @@ object KyuubiConf {
     .checkValues(EngineType.values.map(_.toString))
     .createWithDefault(EngineType.SPARK_SQL.toString)
 
+  val ENGINE_POOL_FIRST: ConfigEntry[Boolean] = buildConf("kyuubi.engine.pool.first")
```
pan3793, line 1655: we shall change it to `kyuubi.engine.pool.ignoreSubdomain` and mark it as internal, so the primary users will not care about what's the relationship between engine pool and subdomain

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1652,6 +1652,14 @@ object KyuubiConf {
     .checkValues(EngineType.values.map(_.toString))
     .createWithDefault(EngineType.SPARK_SQL.toString)
 
+  val ENGINE_POOL_IGNORE_SUBDOMAIN: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.pool.ignoreSubdomain")
+      .doc("Whether to ignore subdomain when engine pool conditions are met.")
```
pan3793, line 1657: please replace "subdomain" w/ the full conf key

pan3793 (approved):

turboFei:
thanks, merged to master

Issue #4149, opened 2023-01-12
[Improvement] Refactor FilteredShowObjectCommands based on Authz Plan Serde Layer


zhouyifan279:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Refactor FilteredShowObjectCommands based on Authz Plan Serde Layer

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

Issue #3687, opened 2022-10-24
[Umbrella] Make table owner defaults to kyuubi.session.user when create table


zhouyifan279:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the proposal

When run Kyuubi Spark SQL Engine in GROUP or SERVER mode, newly-created table's owner defaults to engine launch user.
However, some people may want to set table owner to Kyuubi session user.

We should support this behavior in KyuubiSparkExtension and provide a conf to enable/disable it.

### Task list

- [] #Set table owner to Kyuubi session user for table-creation Command having CatalogTable field
- [] #Set table owner to Kyuubi session user for table-creation Command not having CatalogTable field

### Are you willing to submit PR?

- [X] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4154, opened 2023-01-12
[KYUUBI #4149] Refactor FilteredShowObjectCommands based on Authz Pla…

zhouyifan279:
…n Serde Layer

1. Refactor FilteredShowObjectCommands based on Authz Plan Serde Layer
2. #3687 can reuse the framework implemented in this PR


bowenliang123:
I agree with this PR overall, but can you summarize the changes and refactoring design in the description for clarification ? 
Including but not limited to, what new optimizer rule and planner strategy introduced or refactored and their relationship, what privilege changed of which type of source. And most importantly, the key reason or consideration to these changes.
@zhouyifan279

PR #4155, opened 2023-01-12
Reduce the application info call for batch

turboFei:

Reduce the application info call for batch.
- If terminated and applicationInfo is defined, return applicationInfo directly.
- For batch report, return the existing applicationInfo directly. 

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -70,11 +70,9 @@ class BatchJobSubmission(
   private[kyuubi] val batchId: String = session.handle.identifier.toString
 
   @volatile private var _applicationInfo: Option[ApplicationInfo] = None
-  def getOrFetchCurrentApplicationInfo: Option[ApplicationInfo] = {
-    _applicationInfo match {
-      case Some(_) => _applicationInfo
-      case None => currentApplicationInfo
-    }
+  def getOrFetchCurrentApplicationInfo: Option[ApplicationInfo] = _applicationInfo match {
+    case Some(_) => _applicationInfo
+    case None => currentApplicationInfo
   }
 
   private var killMessage: KillResponse = (false, "UNKNOWN")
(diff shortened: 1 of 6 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -68,7 +68,7 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
   private def buildBatch(session: KyuubiBatchSessionImpl): Batch = {
     val batchOp = session.batchJobSubmissionOp
     val batchOpStatus = batchOp.getStatus
-    val batchAppStatus = batchOp.currentApplicationInfo
+    val batchAppStatus = batchOp.getApplicationInfo
```
ulysses-you, line 71: I'm not sure it can alway return a correct info, shall we make this method as `getOrFetchCurrentApplicationInfo` ? Then we in that method, we can call `currentApplicationInfo` if applicationInfo is empty.

turboFei: thanks, done

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -69,7 +69,13 @@ class BatchJobSubmission(
 
   private[kyuubi] val batchId: String = session.handle.identifier.toString
 
-  private var applicationInfo: Option[ApplicationInfo] = None
+  @volatile private var _applicationInfo: Option[ApplicationInfo] = None
+  def getOrFetchCurrentApplicationInfo: Option[ApplicationInfo] = {
```
bowenliang123, line 73: minor: the empty brace can be removed.

bowenliang123 (approved):
LGTM.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: done

ulysses-you (approved):

PR #4156, opened 2023-01-13
Issue engine token when opening engine session to prevent token expiration

turboFei:

Issue engine token when opening engine session to prevent token expiration.

```
  def authToken(tokenStr: String): Unit = {
    val identifier =
      try {
        KyuubiInternalAccessIdentifier.fromJson(decrypt(tokenStr))
      } catch {
        case _: Exception =>
          throw KyuubiSQLException("Invalid engine access token")
      }
    if (identifier.issueDate + identifier.maxDate < System.currentTimeMillis()) {
      throw KyuubiSQLException("The engine access token is expired")
    }
  }
```


cfmcgrady (approved):

turboFei:
thanks, merged to master

PR #4157, opened 2023-01-13
补充python连接kyuubi server信息

w517424787:
补充pyhive连接kyuubi server时，当hive没做用户验证和权限控制时，不用添加auth='NOSASL'。

另外，补充通过dbapi连接kyuubi server，当hive没做用户验证和权限控制时，需设置auth_mechanism='PLAIN'。

### DIFF
--- a/docs/client/python/pyhive.md
+++ b/docs/client/python/pyhive.md
@@ -24,7 +24,7 @@
 PyHive works with Python 2.7 / Python 3. Install PyHive via pip for the Hive interface.
 
 ```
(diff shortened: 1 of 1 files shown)


bowenliang123:
Seems like the suggested `impyla` python package is not part of PyHive and it is a counterpart to PyHive. The necessity should be discussed.

w517424787:
if set `kyuubi.authentication=NOSASL` in kyuubi-default.conf，use `pyhive` connect kyuubi server，the auth need set 'NOSASL'

PR #4159, opened 2023-01-13
[DOCS] Refine doc for kyuubi.authentication

pan3793:

Add catalog tree for `kyuubi.authentication` options

image


pan3793:
cc @bowenliang123

bowenliang123 (approved):

cxzl25 (approved):

turboFei (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -726,19 +726,37 @@ object KyuubiConf {
...
-      " For SASL authentication, KERBEROS and PLAIN auth type are supported at the same time," +
-      " and only the first specified PLAIN auth type is valid.")
+      "The following tree describes the catalog of each option." +
+      "<ul>" +
+      "  <li><code>NOSASL</code></li>" +
+      "  <li>SASL" +
+      "    <ul>" +
+      "      <li>SASL/PLAIN</li>" +
+      "        <ul>" +
+      "          <li><code>NONE</code></li>" +
```
yaooqinn, line 745: how about keeping the desc?

pan3793: the main purpose of the tree is to make users clear the catalog of each option, and it was explained above, duplication seems too redundant.

yaooqinn: my mistake, I didn't notice that the descriptions are still there

PR #4164, opened 2023-01-13
Accumulate the operation terminal state

turboFei:

For the KyuubiOperation, its final state should be `CLOSED`, and now the `FINISHED`, `ERROR` states, which are important for kyuubi administer, are not accumulated.

In this pr, we check whether the old state is terminal state, if it is, do not decrease it.


pan3793 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
```diff
@@ -176,7 +176,9 @@ abstract class KyuubiOperation(session: Session) extends AbstractOperation(sessi
 
   override def setState(newState: OperationState): Unit = {
     MetricsSystem.tracing { ms =>
-      ms.markMeter(MetricRegistry.name(OPERATION_STATE, opType, state.toString.toLowerCase), -1)
+      if (!isTerminal(state)) {
```
Yikf, line 179: nit: isTerminal -> isTerminalState

turboFei: refactor to `if (!OperationState.isTerminal(state))`.

bowenliang123 (approved):
LGTM

turboFei:
thanks, merged to master

PR #4172, opened 2023-01-15
[WIP] Feat aws provider

### LINKED ISSUE #4118: [FEATURE] Support aws iam for authentiction / authorization
### Code of Conduct

### Search before asking

### Describe the feature

Provide a way for users to :
- get access to data based on their users IAM roles
- login thought access token, id/secret or other IAM stuff

### Motivation

currently AFAIK, only user/password and kerberos are provided

### Describe the solution

kyuubi itself could have a role which could assume other role. For a given user, kyuubi could check if the role it provided as jdbc parameter can be assumed and passed to the engine (spark supports IAM roles)

### Additional context

_No response_

parisni:

This allows to deal with aws iam author. The idea is to make the kyuubi jdbc driver able to inject iam stuff. The aws sdk dependencies have been added. Multiple awsProvider will be implemented:

-  [ ] default chain provider
- [x] sso (integration tests made manually)
- [ ] basic auth (access/secret)
- [ ] assumed roles
 
Fixes #4118 

### DIFF
--- a/kyuubi-hive-jdbc/pom.xml
+++ b/kyuubi-hive-jdbc/pom.xml
@@ -104,6 +104,12 @@
         </dependency>
 
         <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>sso</artifactId>
+            <version>2.19.16</version>
+        </dependency>
+
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcConnectionParams.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcConnectionParams.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 
 public class JdbcConnectionParams {
+
   // Note on client side parameter naming convention:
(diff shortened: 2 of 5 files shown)


yaooqinn:
Hey @parisni, can we switch the target branch to the default?

parisni:
> Why is there such a big gap between 1.6 and 1.7? I noticed that your new code change has little relevance to the old ones.
well when I compile the jdbc-driver 1.7 I am not able to deal with 1.6 kyuubi server. 

> FYI, the client libs are about to be transferred to https://github.com/apache/kyuubi-client repo
Then this PR will be transfered to kyuubi-client since the implementation is only on client side, which convert jdbc variables/config into relevant spark conf

parisni:
in the current proposal, all takes place on the jdbc-drivern side. 
Since only the client has the aws iam knowledge to access the aws
ressources. 
Then the driver only acts as an intermediary layer, with no action but
passing
the configs to the engine.

We may move the configuration builder (spark.hadoop.s3a...) on the
server side, and make it possible 
to support multiple engine. Thought ?

yaooqinn:
> well when I compile the jdbc-driver 1.7 I am not able to deal with 1.6 kyuubi server. 

Can you take a look at this? @pan3793 @turboFei 

Sorry for being late @parisni, it was Spring Festival in China for the past 2 weeks

turboFei:
> well when I compile the jdbc-driver 1.7 I am not able to deal with 1.6 kyuubi server.

Could you provide the error message? @parisni

PR #4185, opened 2023-01-18
Handle session exception for KyuubiSessionImpl::openEngineSession

turboFei:

It is helpful for restful use case. 

If the kyuubi session failed, customer can get the session info and check the exception. 

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/SessionsResourceSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/SessionsResourceSuite.scala
@@ -31,6 +31,7 @@ import org.apache.kyuubi.{KyuubiFunSuite, RestFrontendTestHelper}
 import org.apache.kyuubi.client.api.v1.dto._
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiReservedKeys.KYUUBI_SESSION_CONNECTION_URL_KEY
+import org.apache.kyuubi.engine.ShareLevel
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/SessionsResourceSuite.scala
```diff
@@ -277,4 +277,22 @@ class SessionsResourceSuite extends KyuubiFunSuite with RestFrontendTestHelper {
...
+
+    var response = webTarget.path("api/v1/sessions")
+      .request(MediaType.APPLICATION_JSON_TYPE)
+      .post(Entity.entity(requestObj, MediaType.APPLICATION_JSON_TYPE))
+
+    val sessionHandle = response.readEntity(classOf[SessionHandle]).getIdentifier
+
+    eventually(timeout(1.minutes), interval(200.milliseconds)) {
+      response = webTarget.path(s"api/v1/sessions/$sessionHandle").request().get()
+      val sessionEvent = response.readEntity(classOf[String])
```
turboFei, line 294: will meet json parse exception with `response.readEntity(classOf[KyuubiSessionEvent])`

bowenliang123 (approved):

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/SessionsResourceSuite.scala
bowenliang123: I think this comment could be added to ut which is helpful to understand.

turboFei:
thanks, merging to master

PR #4187, opened 2023-01-19
[Improvement] Print revision commit time on server startup

bowenliang123:

- commit date is more readable and helpful to determine earlier/later version, especially for snapshot branch
- adding revision commit date from git to `kyuubi-version-info.properties`
- Print revision date on server startup 

image

### DIFF
--- a/build/kyuubi-build-info
+++ b/build/kyuubi-build-info
@@ -32,6 +32,7 @@ echo_build_properties() {
   echo kyuubi_trino_version="$9"
   echo user="$USER"
   echo revision=$(git rev-parse HEAD)
+  echo revision_time=$(git show -s --format=%ci HEAD)
   echo branch=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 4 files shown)


build/kyuubi-build-info
```diff
@@ -32,6 +32,7 @@ echo_build_properties() {
   echo kyuubi_trino_version="$9"
   echo user="$USER"
   echo revision=$(git rev-parse HEAD)
+  echo revision_date=$(git show -s --format=%ci HEAD)
```
pan3793, line 35: "date" is confusing since that JSR 310 uses "date" to represent `year-month-day` and "time" to represent `hour:minute:sec[.]`, how about "revision_time" or "revision_timestamp"?

bowenliang123: Addressed. Let's use `revision_time` for the name. Thanks for the detailed hints.

bowenliang123:
thanks, merged to master.

PR #4188, opened 2023-01-19
[Improvement] Introduce code formatting for `*.g4` Antlr4 grammar files

bowenliang123:

- introduce code formatting for Antlr4 files (eg. for KyuubiSql grammar) using `antlr4Formatter` from `spotless-maven-plugin` (doc: https://github.com/diffplug/spotless/tree/main/plugin-maven#antlr4)


ulysses-you:
To be honest, I do not like this new style. Here are my thoughts:
1.  the new style just use more blank line which make it un-readable
2. the end on line `#xxx` is not aligned
3. the current style follows Spark

bowenliang123:
> To be honest, I do not like this new style. Here are my thoughts:
> 
> 1. the new style just use more blank line which make it un-readable
> 2. the end on line `#xxx` is not aligned
> 3. the current style follows Spark

Thanks for reviewing, and I agree with the points you listed.
The rough idea of this PR is to bring auto-formatting to Antlr files, but currently no more options supported to align with the style we expected.

Shall we just close this PR ?

ulysses-you:
yeah, I think so. We can re-open it when we find a better way.

bowenliang123:
OK, closing this PR. Thx.
Btw, we are not expecting a better solution for Antlr styling from the future version of `spotless-maven-plugin`, since from `2.31.0` it requires Java 11+ we are now pinning it to `2.30.0`.

PR #4189, opened 2023-01-19
Scala repl output log level adjusted to debug

imperio-wxm:
Motivation: When the running result of scala code is very long, all the information will be printed to info, and it is not very elegant to print the result set to the info level.


pan3793 (approved):

pan3793:
Happy new year! The change makes sense, thanks for your contribution.

PR #4191, opened 2023-01-19
[Improvement] Skip fetching Maven dependency snapshot update when building `dependencyList`

bowenliang123:

- Skip Maven dependency snapshot update fetching to building `dependencyList` faster without waiting for fetching daily snapshots checking (only snapshots of submodules used).

### DIFF
--- a/build/dependency.sh
+++ b/build/dependency.sh
@@ -32,7 +32,7 @@ DEP="${PWD}/dev/dependencyList"
 
 
 function build_classpath() {
-  $MVN dependency:build-classpath -pl kyuubi-ctl,kyuubi-server,kyuubi-assembly |\
+  $MVN dependency:build-classpath --no-snapshot-updates -pl kyuubi-ctl,kyuubi-server,kyuubi-assembly |\
     grep -v "INFO\|WARN" | \
     tail -1 | \
     tr ":" "\n" | \


build/dependency.sh
```diff
@@ -81,6 +81,8 @@ cat >"${DEP_PR}"<<EOF
 
 EOF
 
+$MVN dependency:resolve
```
pan3793, line 84: is it possible to combine it w/ line 35? just like 

```
$MVN dependency:resolve dependency:build-classpath -pl ...
```

bowenliang123: Several reasons to do it separately,
1.  better not to do in one place as not sure whether the change will impact or contaminate the outputs for dependencyList.
2. solely resolving maven dependency and the output will be more clear to read and to know which process we are waiting for (eg. resolving deps , building_classpath)

pan3793:
cc @cfmcgrady

cfmcgrady:
> thus many daily updated snapshot releases will be downloaded again and this process is invisible and make it seems jammed with no console output.

do we have any daily updated snapshot dependencies?

bowenliang123:
> > thus many daily updated snapshot releases will be downloaded again and this process is invisible and make it seems jammed with no console output.
> 
> do we have any daily updated snapshot dependencies?

Maven is redownloading the submodules from `apache.snapshots` every day after daily snapshots are published.
image

cfmcgrady:
Since all dependencies are stable, and only project submodules are snapshot versions, why don't we skip snapshot updates when updating the dependencies list?

We can skip snapshot update by running

```
mvn goal --no-snapshot-updates
```
or
```
mvn goal -nsu
```

FYI: https://stackoverflow.com/questions/1348603/how-can-i-get-maven-to-stop-attempting-to-check-for-updates-for-artifacts-from-a

cfmcgrady:
Since all dependencies are stable, and only project submodules are snapshot versions, why don't we skip snapshot updates when updating the dependencies list?

We can skip snapshot update by running

```
mvn goal --no-snapshot-updates
```
or
```
mvn goal -nsu
```

FYI: https://stackoverflow.com/questions/1348603/how-can-i-get-maven-to-stop-attempting-to-check-for-updates-for-artifacts-from-a

cfmcgrady (approved):

bowenliang123:
> Since all dependencies are stable, and only project submodules are snapshot versions, why don't we skip snapshot updates when updating the dependencies list?
> 
> We can skip snapshot update by running
> 
> ```
> mvn goal --no-snapshot-updates
> ```
> 
> or
> 
> ```
> mvn goal -nsu
> ```
> 
> FYI: https://stackoverflow.com/questions/1348603/how-can-i-get-maven-to-stop-attempting-to-check-for-updates-for-artifacts-from-a

SGTM. Applied the suggestion of using `--no-snapshot-updates` option and also changed the title and description of this PR.

bowenliang123:
Thanks, merged to master.

PR #4192, opened 2023-01-19
[INFRA] Make all CI jobs run on `ubuntu-22.04`

bowenliang123:

In 21 CI workflow jobs, 18 run on `ubuntu-22.04` and 3 run on `ubuntu-latest` (the same as `ubuntu-22.04` currently),

Making all CI jobs run on `ubuntu-22.04` to ensure alignment in the same operation system version, and it benefits from future version updates.

CI actions pass.

### DIFF
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -22,7 +22,7 @@ on: issues
 
 jobs:
   greeting:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     permissions:
       issues: write
       pull-requests: write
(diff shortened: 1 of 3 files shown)


pan3793:
Thanks for finding this issue, but we prefer to use the explicit version, to avoid breaking changes on the "latest" tag changing to a new major version.

bowenliang123:
> Thanks for finding this issue, but we prefer to use the explicit version, to avoid breaking changes on the "latest" tag changing to a new major version.

Good suggestion, make all CI jobs running on `ubuntu-22.04` instead.
`ubuntu-22.04` is still the latest available runner for actions according to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners .

pan3793 (approved):

bowenliang123:
thanks, merged to master.

PR #4195, opened 2023-01-20
[INFRA] Separate the steps for build and test for Spark CI jobs

bowenliang123:

For fast-failing with the errors in compilation and builds, separate the steps for build and testing for Spark, the same way as in `Flink Test` and `Hive Test`. Separated steps also help to find exception messages in very long logging output, which could be up to around 3k lines, in sections.

image

CI actions pass.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -87,7 +87,7 @@ jobs:
       - name: Test Kyuubi and Spark with maven w/o linters
         run: |
           TEST_MODULES="dev/kyuubi-codecov"
-          ./build/mvn test ${MVN_OPT} -P!jdbc-shaded -pl ${TEST_MODULES} -am \
+          ./build/mvn test ${MVN_OPT} -pl ${TEST_MODULES} -am \
           -Pspark-${{ matrix.spark }} ${{ matrix.spark-archive }} ${{ matrix.exclude-tags }}
       - name: Code coverage
         if: |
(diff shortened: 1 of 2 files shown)


bowenliang123:
Do you have any idea for the failures in the testing step after separation? @pan3793

pan3793:
I guess the separation will extend the whole time for the positive cases, the compiling error should be easy to catch locally.

bowenliang123:
The whole time should not be expanded in much added time.
As `mvn install` should have properly compiled and installed the latest jar's in the local repo, `mvn test` just use all the compiled source and tests from local.

bowenliang123:
The current obstacle is why `mvn test` just failed to lookup compiled classes after `mvn install -DskipTests`?
https://github.com/apache/kyuubi/actions/runs/3963918607/jobs/6792291190#step:7:547
```
Error: ] /home/runner/work/kyuubi/kyuubi/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/PySparkTests.scala:28: object KyuubiHiveDriver is not a member of package org.apache.kyuubi.jdbc
Error: ] /home/runner/work/kyuubi/kyuubi/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/PySparkTests.scala:29: object hive is not a member of package org.apache.kyuubi.jdbc
Error: ] /home/runner/work/kyuubi/kyuubi/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/PySparkTests.scala:70: not found: type KyuubiHiveDriver
```

PR #4201, opened 2023-01-23
[Improvement] Remove Gson as direct dependency by using Jackson instead in Authz tests

bowenliang123:

- use `Jackson` instead of `Gson` in Authz tests, where it is the only place `Gson` is used in project
- remove `Gson` as the dependency
- Notice: Gson project is in maintenance mode, (see https://github.com/google/gson/pull/2085)
- Gson is still on `dependencyList`, as depended in kyuubi-ha
image


pan3793 (approved):

bowenliang123:
thanks, merged to master.

PR #4205, opened 2023-01-28
Fix typo in TFrontendService

turboFei:

Duplicate with
https://github.com/apache/kyuubi/blob/5357a0211bd50e3789a5cffb43e104587586a287/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala#L265-L280

### DIFF
--- a/build/dist
+++ b/build/dist
@@ -18,7 +18,7 @@
 #
 
 #
-# Script to create a binary distribution for easy deploys of Spark.
+# Script to create a binary distribution for easy deploys of Kyuubi.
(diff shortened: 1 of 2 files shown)


turboFei:
https://github.com/apache/kyuubi/blob/master/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala#L265-L280

pan3793 (approved):

turboFei:
thanks, merged to master

PR #4219, opened 2023-01-31
[KYUUBI #4218] Using DB and table name when checking Delta table schema.

### LINKED ISSUE #4218: [Bug] Delta Lake columns only show in DBeaver/Power BI/etc. in the "default" database
### Code of Conduct

### Search before asking

### Describe the bug

When trying to list columns in a Delta Lake table using a tool like DBeaver or Power BI the operation only works if the table is in the "default" database. If the table is in any other database, the operation will fail with a "table or view not found" error.

### Affects Version(s)

master/1.6.1

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

nousot-cloud-guy:

To close #4218 .
This change ensures BI tools can list columns on Delta Lake tables in all schemas.

image

image

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v2_4.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v2_4.scala
@@ -141,7 +141,7 @@ class CatalogShim_v2_4 extends SparkCatalogShim {
       catalog.getTablesByName(identifiers).flatMap { t =>
         val tableSchema =
           if (t.provider.getOrElse("").equalsIgnoreCase("delta")) {
-            spark.table(t.identifier.table).schema
+            spark.table(f"${db}.${t.identifier.table}").schema
           } else {
             t.schema
           }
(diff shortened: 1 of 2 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v2_4.scala
```diff
@@ -141,7 +141,7 @@ class CatalogShim_v2_4 extends SparkCatalogShim {
       catalog.getTablesByName(identifiers).flatMap { t =>
         val tableSchema =
           if (t.provider.getOrElse("").equalsIgnoreCase("delta")) {
-            spark.table(t.identifier.table).schema
+            spark.table(f"${db}.${t.identifier.table}").schema
```
pan3793, line 144: can we use `t.identifier.toString` here?

pan3793:
Thanks for investigating and making this PR, could you please verify the case when the db or table name contains dots?

pan3793:
After second thought, I think we can totally remove the current workaround, and just invert the pattern match' branch order in `CatalogShim_v3_0#getColumnsByCatalog`

```patch
- catalog match {
-   case builtin if builtin.name() == SESSION_CATALOG =>
-     super.getColumnsByCatalog(...)
-   case tc: TableCatalog =>
-     ...
- }

+ catalog match {
+   case tc: TableCatalog =>
+     ...
+   case builtin if builtin.name() == SESSION_CATALOG =>
+     super.getColumnsByCatalog(...)
+ }
```

Would you please try this way? @nousot-cloud-guy

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v2_4.scala
pan3793, line 144:
```suggestion
            spark.table(t.identifier.toString).schema
```

nousot-cloud-guy:
> 

That seems to be working. I've updated the PR with the reverts and new commit.

pan3793 (approved):

pan3793:
Merged to master/1.6, thanks @nousot-cloud-guy

PR #4224, opened 2023-02-01
Enlarge REST client default timeout

pan3793:

The current default socket timeout is set to 3s, which is too short, when Kyuubi Server is busy, e.g. during GC, the client will fail w/ socket read timeout after 3s if no bytes are returned in the socket.

This PR proposes to enlarge the REST client default timeout

- connect timeout from 3s to 30s
- socket timeout from 3s to 2min
- attempt interval keep 3s

They keep consistent w/ `org.apache.kyuubi.ctl.CtlConf`


turboFei (approved):

bowenliang123 (approved):

pan3793:
Thanks, merged to master

PR #4228, opened 2023-02-01
Migrate `httpclient` to `httpclient5` in REST client

bowenliang123:

- Migrate `httpclient4`(org.apache.httpcomponents:httpclient) to `httpclient5`(org.apache.httpcomponents.client5:httpclient5) for REST API client
- `httpclient4` is still used in `kyuubi-hive-jdbc` as required by `libthrift`, and shipped in `beeline-jars` folder of distribution tarball while `httpclient5` jars in `jars` folder.

Reasons:
- `httpclient4` is now in maintenance mode and `httpclient5` is recommended (https://hc.apache.org/httpcomponents-client-5.2.x/)
- More modern features and expansibility integrated into httpclient5, like the async client, flexible pooling, HTTP/2
- make good use of the newer Java version, as `httpclient5` requires Java 8 while `httpclient4` requires Java 6

Changes:
- no major change in higher level to REST API client end-users


kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/HttpClientFactory.java
```diff
@@ -35,23 +42,35 @@ public class HttpClientFactory {
...
+    ConnectionConfig connConfig =
+        ConnectionConfig.custom()
+            .setConnectTimeout(conf.getConnectTimeout(), TimeUnit.MILLISECONDS)
+            .build();
+    cm.setConnectionConfig(connConfig);
 
     return HttpClientBuilder.create()
         .setDefaultRequestConfig(requestConfig)
-        .setSSLSocketFactory(sslSocketFactory)
+        .setConnectionManager(cm)
```
bowenliang123, line 73: note: `setSSLSocketFactory` method is removed in httpclient5, migrated to set socketFactoryRegistry via the ConnectionManager.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/HttpClientFactory.java
```diff
@@ -35,23 +42,35 @@ public class HttpClientFactory {
   public static CloseableHttpClient createHttpClient(RestClientConf conf) {
     RequestConfig requestConfig =
         RequestConfig.custom()
-            .setSocketTimeout(conf.getSocketTimeout())
-            .setConnectTimeout(conf.getConnectTimeout())
+            .setResponseTimeout(conf.getSocketTimeout(), TimeUnit.MILLISECONDS)
```
bowenliang123, line 45: note: `setResponseTimeout` in httpclient5 replaced `setSocketTimeout`.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/HttpClientFactory.java
```diff
@@ -35,23 +42,35 @@ public class HttpClientFactory {
   public static CloseableHttpClient createHttpClient(RestClientConf conf) {
     RequestConfig requestConfig =
         RequestConfig.custom()
-            .setSocketTimeout(conf.getSocketTimeout())
-            .setConnectTimeout(conf.getConnectTimeout())
```
bowenliang123, line 39: note: `setConnectTimeout` is removed in `httpclient5`, and changed to use ConnectionConfig to set connect timeout.
```
ConnectionConfig.custom()
            .setConnectTimeout(conf.getConnectTimeout(), TimeUnit.MILLISECONDS)
```

pan3793:
Do you know which project adopts HC5? IMO, I would keep the latest HC4 for a while, until HC5 is adopted widely.

bowenliang123:
I would regard `httpclient` as a underlying lib inside the REST client. With this migration, we are able to provide more dynamic features, efficiency, and performance (like aysnc, reactor, pooling, http2 via proxy, fluent) without changing the usage of high-level API. `httpclient5` has remained most of the common concepts of `httpclient4` as in `classic` packages, therefore, no major burden to evaluate or improve it in details.

bowenliang123:
> Do you know which project adopts HC5? IMO, I would keep the latest HC4 for a while, until HC5 is adopted widely.

The latest version of `com.clickhouse:clickhouse-jdbc` 0.4.0 is dependent on `org.apache.httpcomponents.client5:httpclient5` 5.2.1.

pom.xml
```diff
@@ -153,8 +153,10 @@
         <hive.archive.name>apache-hive-${hive.engine.hive.version}-bin.tar.gz</hive.archive.name>
         <hive.archive.mirror>${apache.archive.dist}/hive/hive-${hive.engine.hive.version}</hive.archive.mirror>
         <hive.archive.download.skip>false</hive.archive.download.skip>
+        <httpcore.version>4.4.15</httpcore.version>
         <httpclient.version>4.5.14</httpclient.version>
```
zwangsheng, line 157: Why do we reuse 4.4.15 for http core , based on the fact that we already use 4.4.16

zwangsheng, line 157: Since it is not used, we should delete `httpclient.version`.

pom.xml
```diff
@@ -153,8 +153,10 @@
...
         <httpcore.version>4.4.16</httpcore.version>
```
zwangsheng, line 158: Duplicate propertt

pom.xml
bowenliang123: Just as based on the previous commit. Easy to be fixed before community agree with HC5 migration.

bowenliang123: Search again. jdbc driver is still using HC4, as mentioned in PR description.

pom.xml
bowenliang123: Nope.

bowenliang123: Addressed.

bowenliang123:
Closing this PR.
If this PR merged, we still have to maintain dependencies on both HTTP clients 4 and 5, although from separate modules.

PR #4231, opened 2023-02-02
[KYUUBI #4144][FOLLOWUP] Do not cleanup upload dir because of batch recovery and fix temp file leak

turboFei:

Address comments: https://github.com/apache/kyuubi/pull/4144#issuecomment-1412078077

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -17,7 +17,8 @@
 
 package org.apache.kyuubi.operation
 
-import java.io.{File, IOException}
+import java.io.IOException
+import java.nio.file.{Files, Paths}
 import java.util.Locale
 import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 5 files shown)


turboFei:
Seems need to cleanup when closing batch op to prevent temp file leak. Will address it later.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -363,6 +363,12 @@ class BatchJobSubmission(
   override def isTimedOut: Boolean = false
 
   override protected def eventEnabled: Boolean = true
+
+  private def cleanupUploadedResourceIfNeeded(): Unit = {
+    if (session.isResourceUploaded && new File(resource).exists()) {
```
bowenliang123, line 368: How about skip check the existence of the file, and then make `Utils.deleteFile` use `Files.deleteIfExists` instead?
```suggestion
    if (session.isResourceUploaded) {
```

bowenliang123 (approved):
LGTM. Minor suggestion only.

pan3793 (approved):

turboFei:
thanks, merged to master

PR #4241, opened 2023-02-03
Only force close engine ref when open session failed

ulysses-you:

We do not need force close engine builder and application if open session succeded. The engine itself will shutdown in connection level.

This pr tries to fix regression of https://github.com/apache/kyuubi/pull/2482

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -250,7 +250,7 @@ class KyuubiSessionImpl(
     try {
       if (_client != null) _client.closeSession()
     } finally {
-      openSessionError.foreach {_ => if (engine != null) engine.close()}
+      openSessionError.foreach { _ => if (engine != null) engine.close() }
       sessionEvent.endTime = System.currentTimeMillis()
       EventBus.post(sessionEvent)
       traceMetricsOnClose()
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -105,6 +105,8 @@ class KyuubiSessionImpl(
 
   private var _engineSessionHandle: SessionHandle = _
 
+  private var openSessionError: Option[Throwable] = None
+
   override def open(): Unit = handleSessionException {
     traceMetricsOnOpen()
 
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -247,7 +250,7 @@ class KyuubiSessionImpl(
     try {
       if (_client != null) _client.closeSession()
     } finally {
-      if (engine != null) engine.close()
+      openSessionError.foreach { _ => if (engine != null) engine.close() }
```
Yikf, line 253: The egine will not be reused if the sharelevel is a CONNECTION and the session is closed, so should close engine whether open session fails or not?

ulysses-you: With connection share level, engine should close gracefully by itself. Engine exit code will not be 0 if we close engine by Kyuubi Server.

wForget: Can we add a timeout to wait for the engine status to complete, and then kill the application.

ulysses-you: We can, but it's a bit complex. I think this pr is enough to fix regression.

I see your requreiements that to avoid the leaked engine. We can improve that in a global view rather than wait and kill one by one. e.g. Use a thread to detect if there is a leaked engine.

Note, it's the key to define what is the `leaked engine`. In general, A `leaked engine` can be:
- status is finished in resource manager but engine build process is alive (for all share level)
- session is closed but status is not fiinshed or engine build process is alive (for connect share level and batch mode)

We can have a new pr if you are interested in.

wForget: > We can, but it's a bit complex. I think this pr is enough to fix regression.
> 
> I see your requreiements that to avoid the leaked engine. We can improve that in a global view rather than wait and kill one by one. e.g. Use a thread to detect if there is a leaked engine.
> 
> Note, it's the key to define what is the `leaked engine`. In general, A `leaked engine` can be:
> 
> * status is finished in resource manager but engine build process is alive (for all share level)
> * session is closed but status is not fiinshed or engine build process is alive (for connection share level and batch mode)
> 
> We can have a new pr if you are interested in.

I agree with you, this pr is enough to fix the previous issue.

wForget (approved):

ulysses-you:

PR #4242, opened 2023-02-03
Support thrift binary non-ssl and ssl connections at the same time

turboFei:

Before we only support thrift binary non-ssl, since #3658, we support thrift binary ssl connection.

If customer want to migrate from non-ssl to ssl thrift binary connections, we need support non-ssl and ssl connections at the same time.

This PR will help for the thrift binary ssl migration.

For example:

before: 
```
kyuubi.frontend.protocols=THRIFT_BINARY
kyuubi.frontend.thrift.binary.bind.port=10009
```

after:
```
kyuubi.frontend.protocols=THRIFT_BINARY,THRIFT_BINARY_SSL
kyuubi.frontend.thrift.binary.bind.port=10009
kyuubi.frontend.thrift.binary.ssl.bind.port=10010
```

image


turboFei:
how do you think about? @yaooqinn  @pan3793 

I think it should be common case if they want to make SSL migration.

docs/deployment/settings.md
```diff
@@ -311,6 +311,7 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
+| kyuubi.frontend.thrift.binary.ssl.bind.port            | 10009             | Port of the machine on which to run the thrift frontend service via the binary protocol with SSL enabled.                                                                                                                                                                                                                                                                                                                                                                                                                   | int      | 1.7.0 |
```
pan3793, line 314: we should give it a different value

pan3793:
I think it's a common case, many services support to listen plain/ssl port at the same time.

bowenliang123 (approved):
LGTM

pan3793 (approved):

yaooqinn:
A bit weird to me for the naming part

yaooqinn:
This does not support HA and introduces bugs when the SSL on/off together w/ the same namespace.

docs/deployment/settings.md
```diff
@@ -311,6 +311,7 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
+| kyuubi.frontend.thrift.binary.ssl.bind.port            | 10011             | Port of the machine on which to run the thrift frontend service via the binary protocol with SSL enabled.                                                                                                                                                                                                                                                                                                                                                                                                                   | int      | 1.7.0 |
```
yaooqinn, line 314: kyuubi.frontend.thrift.binary.bind.port also can be used for ssl, how do we suppose users to understand?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -371,7 +371,7 @@ object KyuubiConf {
 
   object FrontendProtocols extends Enumeration {
     type FrontendProtocol = Value
-    val THRIFT_BINARY, THRIFT_HTTP, REST, MYSQL, TRINO = Value
+    val THRIFT_BINARY, THRIFT_BINARY_SSL, THRIFT_HTTP, REST, MYSQL, TRINO = Value
```
yaooqinn, line 374: THRIFT_BINARY, THRIFT_BINARY_SSL seems not discrete mathematically

pan3793: I think the idea here is:

1. THRIFT_BINARY, THRIFT_BINARY_SSL should be like HTTP, HTTPS eventually, there is discrete and can be enabled at the same time.
2. currently we keep SSL option of THRIFT_BINARY for compatibility, the deprecating message has been added and this feature will be removed in the 2.0

Issue #4245, opened 2023-02-04
[Improvement] Support UserDefinedType of spark since Spark3.2.0


xuefeimiaoao:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

As user-defined type (UDT) API is public since Spark 3.2.0, kyuubi should support udt schema converter.

reference:
 [SPARK-7768](https://issues.apache.org/jira/browse/SPARK-7768)

### How should we improve?

Add udt to SchemaHelper.

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4246, opened 2023-02-04
[KYUUBI #4245][SPARK] Support UserDefinedType for Spark 3.2 and above

xuefeimiaoao:

to close issue #4245
As user-defined type (UDT) API is public since Spark 3.2.0, kyuubi should support udt schema converter.

reference:
[SPARK-7768](https://issues.apache.org/jira/browse/SPARK-7768)

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
@@ -42,21 +42,21 @@ object SchemaHelper {
   final val YEAR_MONTH_INTERVAL = "YearMonthIntervalType"
 
   /**
-   * Spark 3.2.0 DataType UserDefinedType's class name.
-   * And UDT API has been public since Spark 3.2.0.
-   * Return true if dt is subClass of UserDefinedType, which has been set to public
-   * since Spark 3.2.0
+   * SPARK-7768: UDT API has been public since Spark 3.2.0.
+   * Here we distinguish the UDT if the datatype is subClass of UserDefinedType
+   * that is available in public since Spark 3.2.0
    */
   final private val isUserDefinedType = (dt: DataType) => {
-    val dataType = "DataType"
-    val userDefinedType = "UserDefinedType"
+    val dataType = "org.apache.spark.sql.types.DataType"
+    // Canonical name of Spark 3.2.0 DataType UserDefinedType.
+    val userDefinedType = "org.apache.spark.sql.types.UserDefinedType"
 
     var curClass: Class[_] = dt.getClass
-    while (!curClass.getSimpleName.equals(userDefinedType)
-      && !curClass.getSimpleName.equals(dataType)) {
+    while (!userDefinedType.equals(curClass.getCanonicalName)
+      && !dataType.equals(curClass.getCanonicalName)) {
       curClass = curClass.getSuperclass
     }
-    curClass.getSimpleName.equals(userDefinedType)
+    userDefinedType.equals(curClass.getCanonicalName)
   }
 
   def toTTypeId(typ: DataType): TTypeId = typ match {
@@ -82,7 +82,7 @@ object SchemaHelper {
     case _: ArrayType => TTypeId.ARRAY_TYPE
     case _: MapType => TTypeId.MAP_TYPE
     case _: StructType => TTypeId.STRUCT_TYPE
-    case dt: DateType if isUserDefinedType(dt) => TTypeId.USER_DEFINED_TYPE
+    case udt: DataType if isUserDefinedType(udt) => TTypeId.USER_DEFINED_TYPE
     case other =>
       throw new IllegalArgumentException(s"Unrecognized type name: ${other.catalogString}")
   }
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
@@ -41,6 +41,24 @@ object SchemaHelper {
    */
   final val YEAR_MONTH_INTERVAL = "YearMonthIntervalType"
 
+  /**
+   * Spark 3.2.0 DataType UserDefinedType's class name.
+   * And UDT API has been public since Spark 3.2.0.
+   * Return true if dt is subClass of UserDefinedType, which has been set to public
+   * since Spark 3.2.0
+   */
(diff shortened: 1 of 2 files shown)


bowenliang123:
Can you add some unit tests (including positive and negative) to ensure this won't break in lower spark version and converting correctly?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -41,6 +41,24 @@ object SchemaHelper {
...
 
+  /**
+   * Spark 3.2.0 DataType UserDefinedType's class name.
+   * And UDT API has been public since Spark 3.2.0.
+   * Return true if dt is subClass of UserDefinedType, which has been set to public
+   * since Spark 3.2.0
+   */
+  final private val isUserDefinedType = (dt: DataType) => {
+    val dataType = "DataType"
+    val userDefinedType = "UserDefinedType"
```
bowenliang123, line 52: Comparing with the simple class name of the class could be confused if `DataType` or `UserDefinedType ` IS exactly used as in user's defined type. 
It would be better to use the class name. `org.apache.spark.sql.types` `org.apache.spark.sql.types.UserDefinedType`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -41,6 +41,24 @@ object SchemaHelper {
...
+
+    var curClass: Class[_] = dt.getClass
+    while (!curClass.getSimpleName.equals(userDefinedType)
+      && !curClass.getSimpleName.equals(dataType)) {
```
bowenliang123, line 56: Simlify to use `StringUtils.equalsAnyIgnoreCase`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -41,6 +41,24 @@ object SchemaHelper {
    */
   final val YEAR_MONTH_INTERVAL = "YearMonthIntervalType"
 
+  /**
+   * Spark 3.2.0 DataType UserDefinedType's class name.
+   * And UDT API has been public since Spark 3.2.0.
+   * Return true if dt is subClass of UserDefinedType, which has been set to public
+   * since Spark 3.2.0
```
bowenliang123, line 48:
```suggestion
   * Return true if the datatype is subClass of UserDefinedType, 
   * which is available in public since Spark 3.2.0
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -41,6 +41,24 @@ object SchemaHelper {
    */
   final val YEAR_MONTH_INTERVAL = "YearMonthIntervalType"
 
+  /**
+   * Spark 3.2.0 DataType UserDefinedType's class name.
+   * And UDT API has been public since Spark 3.2.0.
```
pan3793, line 46: let's mention the jira ticket SPARK-7768 as well

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
xuefeimiaoao: Sure, it's best to do it.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
xuefeimiaoao: Sure.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -64,7 +82,7 @@ object SchemaHelper {
     case _: ArrayType => TTypeId.ARRAY_TYPE
     case _: MapType => TTypeId.MAP_TYPE
     case _: StructType => TTypeId.STRUCT_TYPE
-    // TODO: it is private now, case udt: UserDefinedType => TTypeId.USER_DEFINED_TYPE
+    case dt: DateType if isUserDefinedType(dt) => TTypeId.USER_DEFINED_TYPE
```
yaooqinn, line 85: Is case udt: UserDefinedType => TTypeId.USER_DEFINED_TYPE good to go？

BTW, you shall not use `DateType` but `DataType` here.

yaooqinn (changes requested):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
xuefeimiaoao: Thanks, i will revise the dataType. Besides, UserDefinedType is still private in spark 3.1, so we may avoid importing this class.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -41,6 +41,24 @@ object SchemaHelper {
...
+   * SPARK-7768: UDT API has been public since Spark 3.2.0.
+   * Here we distinguish the UDT if the datatype is subClass of UserDefinedType
+   * that is available in public since Spark 3.2.0
+   */
+  final private val isUserDefinedType = (dt: DataType) => {
+    val dataType = "org.apache.spark.sql.types.DataType"
+    // Canonical name of Spark 3.2.0 DataType UserDefinedType.
+    val userDefinedType = "org.apache.spark.sql.types.UserDefinedType"
+
+    var curClass: Class[_] = dt.getClass
```
bowenliang123, line 54:
```suggestion
    var curClass: Class[_] = dt.getClass
```
`currentClass` is more prefered as `cur` is not common abbreviation for `current`

xuefeimiaoao: Sure, thank you.

Issue #3780, opened 2022-11-08
[Umbrella] Add support for executing Python/PySpark scripts


cfmcgrady:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues) and found no similar issues.

### Describe the proposal

Add support for executing Python/PySpark code snippet

### Task list

- #3775
- #3782 
- #3797
- #3807 
- #3896 
- #3900 
- #3908 
- #4018 
- #4019 
- #4028 
- #4026 
- #4031 

### Are you willing to submit PR?

- [ ] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

yaooqinn:
cc @pan3793 @cfmcgrady, shall we close this and target it to v1.7?

cfmcgrady:
+1, SGTM

pan3793:
+1, and thank @cfmcgrady for driving this.

PR #4256, opened 2023-02-06
[DOCS] Add PYTHON option to `kyuubi.operation.language` config

bowenliang123:

- as pyspark is supported in #3780, exposing PYTHON option for `kyuubi.operation.language` as an experimental feature 


bowenliang123:
cc @cfmcgrady @pan3793

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2160,8 +2160,11 @@ object KyuubiConf {
   val OPERATION_LANGUAGE: ConfigEntry[String] =
     buildConf("kyuubi.operation.language")
       .doc("Choose a programing language for the following inputs" +
-        " <ul><li>SQL: (Default) Run all following statements as SQL queries.</li>" +
-        " <li>SCALA: Run all following input a scala codes</li></ul>")
+        "<ul>" +
+        " <li>SQL: (Default) Run all following statements as SQL queries.</li>" +
+        " <li>SCALA: Run all following input as scala codes</li>" +
+        " <li>PYTHON: (Experimental) Run all following input as PySpark codes</li>" +
```
bowenliang123, line 2166: which is better `as PySpark cods`, `as Python codes` or `as Python codes with Spark engine`?

pan3793 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
cfmcgrady: +1 for `as Python codes with Spark engine`

bowenliang123: Good, updated.

cfmcgrady (approved):

bowenliang123:
Thanks, merged to master/1.7 .

PR #4247, opened 2023-02-06
Bump maven from 3.8.6 to 3.6.7

zwangsheng:
### _Why are the changes needed?_

For `maven version` in CI is 3.8.7(get info by adding `/usr/bin/mvn --version` in `.github/workflows/dep.yaml`), 
popo_2023-02-06  09-34-07

We should bump maven version property in parent pom.yaml 3.8.6 => 3.8.7.
### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.readthedocs.io/en/master/develop_tools/testing.html#running-tests) locally before make a pull request

PR #4261, opened 2023-02-07
[KYUUBI #4247] `build/mvn` should use the exact version defined in `pom.xml`

edddddy:

As mentioned in #4247, it seems to be better that using the same maven version with master. Thus, modify the version check logic in `build/mvn` from equal or greater to equal.


pan3793:
Hi @edddddy, thanks for your contribution, would you please fill out the PR template, and supply more background?

pan3793 (approved):

PR #4266, opened 2023-02-07
[INFRA] Add Dependency Review step to prevent introducing vulnerable dependencies

bowenliang123:

- Add actions/dependency-review-action (https://github.com/actions/dependency-review-action) as new step in the dependency workflow, which fails when future PR is bringing vulnerable dependencies.
- configuring `fail-on-severity` to `moderate` level

image

- [x] Pass the Dependency Check CI job

### DIFF
--- a/.github/workflows/dep.yml
+++ b/.github/workflows/dep.yml
@@ -23,8 +23,9 @@ on:
       - master
       - branch-*
     paths:
-      # dependency check happens only pom changes
+      # when pom or dependency workflow changes
       - '**/pom.xml'
+      - '.github/workflows/dep.yml'
 
 concurrency:
   group: dep-${{ github.ref }}
@@ -57,3 +58,7 @@ jobs:
           -pl kyuubi-ctl,kyuubi-server,kyuubi-assembly -am
       - name: Check dependency list
         run: build/dependency.sh
+      - name: Dependency Review
+        uses: actions/dependency-review-action@v3
+        with:
+          fail-on-severity: moderate


.github/workflows/dep.yml
```diff
@@ -57,3 +58,7 @@ jobs:
           -pl kyuubi-ctl,kyuubi-server,kyuubi-assembly -am
       - name: Check dependency list
         run: build/dependency.sh
+      - name: Dependency Review
+        uses: actions/dependency-review-action@v3
+        with:
+          fail-on-severity: low
```
yaooqinn, line 64: Is `low` too strict? Use moderate?

.github/workflows/dep.yml
```diff
@@ -25,6 +25,7 @@ on:
     paths:
       # dependency check happens only pom changes
       - '**/pom.xml'
+      - '.github/workflows/dep.yml'
```
yaooqinn, line 28: Is this necessary?

.github/workflows/dep.yml
bowenliang123: The docs of action does not clarify how it classify the level of severity, so I prefer to keep it in `low` the same as the default value. We could lower this level if necessary in future, as it only affects future PRs.

.github/workflows/dep.yml
bowenliang123: `dep.yml` is designed to check and verify dependencies although doesn't bring changes in dependencies, I think it's good to leave it in paths. As for this PR, this is add to trigger the Dependency Check job.

.github/workflows/dep.yml
yaooqinn: The risk for vulnerability is determined by the CVSS score. If you go to check the mvn central or NVD, it's common to see an artifact have some low-level vulnerabilities. I guess it is not practical for us to add such a critical rule that blocks PRs frequently

bowenliang123: OK, changed to `moderate`.

yaooqinn (approved):

bowenliang123:
Thanks, merged to master.

PR #4271, opened 2023-02-08
Clarify incremental collect mode ignore `kyuubi.operation.result.max.rows`

pan3793:

`kyuubi.operation.result.max.rows` does not take effect on incremental collect mode because of performance concerns, this PR updates the configuration docs to mention that.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -73,18 +73,22 @@ class ExecuteStatement(
       val resultMaxRows = spark.conf.getOption(OPERATION_RESULT_MAX_ROWS.key).map(_.toInt)
         .getOrElse(session.sessionManager.getConf.get(OPERATION_RESULT_MAX_ROWS))
       iter = if (incrementalCollect) {
-        if (resultMaxRows <= 0) {
+        val internalIterator = if (resultMaxRows <= 0) {
           info("Execute in incremental collect mode")
+          if (arrowEnabled) {
+            SparkDatasetHelper.toArrowBatchRdd(convertComplexType(result)).toLocalIterator
+          } else {
+            result.toLocalIterator().asScala
+          }
         } else {
           info(s"Execute in incremental collect mode max result rows[$resultMaxRows]")
-        }
-        var internalIterator = if (arrowEnabled) {
-          SparkDatasetHelper.toArrowBatchRdd(convertComplexType(result)).toLocalIterator
-        } else {
-          result.toLocalIterator().asScala
-        }
-        if (resultMaxRows <= 0) {
-          internalIterator = internalIterator.take(resultMaxRows)
+          if (arrowEnabled) {
+            // this will introduce shuffle and hurt performance
+            val limitedResult = result.limit(resultMaxRows)
+            SparkDatasetHelper.toArrowBatchRdd(convertComplexType(limitedResult)).toLocalIterator
+          } else {
+            result.toLocalIterator().asScala.take(resultMaxRows)
+          }
         }
         new IterableFetchIterator[Any](new Iterable[Any] {
           override def iterator: Iterator[Any] = internalIterator
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -74,8 +74,9 @@ class ExecuteStatement(
         .getOrElse(session.sessionManager.getConf.get(OPERATION_RESULT_MAX_ROWS))
       iter = if (incrementalCollect) {
         if (resultMaxRows > 0) {
-          logger.warn(s"Ignore ${OPERATION_RESULT_MAX_ROWS.key} on incremental collect mode.")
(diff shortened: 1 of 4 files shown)


cfmcgrady (approved):
Looks like branch-1.6 also has the same issue

pan3793:
> In fact, It is by design before.

@turboFei, I see your intention, incremental collection w/ limit seems not a good combination, I would keep the existing behavior and update the comments

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -70,43 +70,40 @@ class ExecuteStatement(
...
       result = spark.sql(statement)
 
-      iter =
-        if (incrementalCollect) {
-          info("Execute in incremental collect mode")
+      val resultMaxRows = spark.conf.getOption(OPERATION_RESULT_MAX_ROWS.key).map(_.toInt)
+        .getOrElse(session.sessionManager.getConf.get(OPERATION_RESULT_MAX_ROWS))
+      iter = if (incrementalCollect) {
+        if (resultMaxRows > 0) {
+          logger.warn(s"Ignore ${OPERATION_RESULT_MAX_ROWS.key} on incremental collect mode.")
```
pan3793, line 77: @turboFei How do you think this way?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -259,13 +259,13 @@ abstract class SparkOperation(session: Session)
 
   override def shouldRunAsync: Boolean = false
 
-  protected def arrowEnabled(): Boolean = {
-    resultFormat().equalsIgnoreCase("arrow") &&
+  protected def arrowEnabled: Boolean = {
```
pan3793, line 262: `()` is removed since it's a pure function w/o side effects, this change addresses the IDEA warnings.

pan3793: @cfmcgrady

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -70,43 +70,40 @@ class ExecuteStatement(
...
+        }
+        val internalIterator = if (arrowEnabled) {
+          SparkDatasetHelper.toArrowBatchRdd(convertComplexType(result)).toLocalIterator
+        } else {
+          result.toLocalIterator().asScala
+        }
+        new IterableFetchIterator[Any](new Iterable[Any] {
+          override def iterator: Iterator[Any] = internalIterator
```
turboFei, line 85: It changes the behavior.

before:
```
override def iterator: Iterator[Row] = result.toLocalIterator().asScala
```

now:

```
val internalIterator = result.toLocalIterator().asScala
override def iterator: Iterator[Any] = internalIterator
```

turboFei: this pr is too complex and  is more possible to hide bug

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
turboFei: it looks good.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
pan3793: I see your point, that's why ut `test fetch orientation with incremental collect mode` failed.

pan3793: changed `val internalIterator` to def `internalIterator()` to resolve the issue

bowenliang123 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -70,43 +70,41 @@ class ExecuteStatement(
...
+          warn(s"Ignore ${OPERATION_RESULT_MAX_ROWS.key} on incremental collect mode.")
+        }
+        info("Execute in incremental collect mode")
+        def internalIterator(): Iterator[Any] = if (arrowEnabled) {
+          SparkDatasetHelper.toArrowBatchRdd(convertComplexType(result)).toLocalIterator
+        } else {
+          result.toLocalIterator().asScala
+        }
+        new IterableFetchIterator[Any](new Iterable[Any] {
+          override def iterator: Iterator[Any] = internalIterator()
```
turboFei, line 86: maybe:
```
() => Iterator[Any]
```

turboFei: maybe:
```
val internalIterator: () => Iterator[Any] = () => {
result.toLocalIterator().asScala
}

override def iterator: Iterator[Any] = internalIterator


```

turboFei (approved):

PR #4285, opened 2023-02-09
Do not cancel CI jobs on stable branch

pan3793:

Since https://github.com/apache/kyuubi/pull/2175, the previous running CI jobs will be canceled on the same branch or PR to save CI resources, this PR keeps the behavior for PR but doesn't cancel the running CI jobs when commit to the stable branch.

Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value

I pushed an empty commit to verify the behavior does not change in PR, and I need to merge this to master first, then merge another PR immediately to verify the rest behaviors.

### DIFF
--- a/.github/workflows/dep.yml
+++ b/.github/workflows/dep.yml
@@ -27,7 +27,7 @@ on:
       - '**/pom.xml'
 
 concurrency:
-  group: dep-${{ github.ref }}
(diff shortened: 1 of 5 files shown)


cfmcgrady (approved):
+1

pan3793:
Thanks, merged to master

PR #4288, opened 2023-02-09
Use `eclipse-temurin:8-jdk-focal` as default base image

pan3793:

eclipse-temurin is the successor for openjdk, see https://github.com/apache/spark/pull/37705

"The core change is: the OS of base image changes debian-bullseye to ubuntu-focal (based on debian bullseye)."

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -29,9 +29,9 @@
 
 # Declare the BASE_IMAGE argument in the first line, for more detail
 # see: https://github.com/moby/moby/issues/38379
-ARG BASE_IMAGE=openjdk:8-jdk
+ARG BASE_IMAGE=eclipse-temurin:8-jdk-focal
 
-FROM maven:3.6-jdk-8 as builder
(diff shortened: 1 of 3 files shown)


build/Dockerfile
```diff
@@ -29,9 +29,9 @@
 
 # Declare the BASE_IMAGE argument in the first line, for more detail
 # see: https://github.com/moby/moby/issues/38379
-ARG BASE_IMAGE=openjdk:8-jdk
+ARG BASE_IMAGE=eclipse-temurin:8-jdk-focal
 
-FROM maven:3.6-jdk-8 as builder
+FROM eclipse-temurin:8-jdk-focal as builder
```
pan3793, line 34: it's hard to maintain the maven version in multi places, so it's better to use `build/mvn` to auto-download the expected version.

ulysses-you (approved):

pan3793:
Thanks, merged to master

PR #4290, opened 2023-02-09
Bump maven version to 3.9.0

bowenliang123:

- maven 3.9.0 release note: https://maven.apache.org/docs/3.9.0/release-notes.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,7 @@
 
     <properties>
         <java.version>1.8</java.version>
-        <maven.version>3.8.7</maven.version>
+        <maven.version>3.9.0</maven.version>
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
         <scala.version>2.12.17</scala.version>


pan3793:
There is a perf regression, I prefer to wait the next patched version

Observed roughly 10% slow-down on large builds when compared to Maven 3.8.7, tracked on [MNG-7677](https://issues.apache.org/jira/browse/MNG-7677).

bowenliang123:
> There is a perf regression, I prefer to wait the next patched version
> 
> Observed roughly 10% slow-down on large builds when compared to Maven 3.8.7, tracked on [MNG-7677](https://issues.apache.org/jira/browse/MNG-7677).

Agree. Closing this PR.

pan3793:
It may be a valuable direction.
https://github.com/apache/maven-build-cache-extension

bowenliang123:
> It may be a valuable direction. https://github.com/apache/maven-build-cache-extension

Seems it requires 3.9.x + , as described on cache extension: (https://github.com/apache/maven-build-cache-extension#status)
`The code currently relies on un-released modifications in the core Maven project, available both for latest 3.9.x and for 4.0-alpha-x SNAPSHOTS described in [MNG-7391](https://issues.apache.org/jira/browse/MNG-7391).`

This cache extension relies on two key points, 1. cache the modules by unified cache key generation, 2. share the caches between building jobs, in remote storage or `target` folder in local project path.

This extension may not be able to bring too much gain in CI builds, as we do not share caches between jobs and only cache the dependencies rather than build classed in  local repo of `~/.m2` by `cache` Github action right now.

PR #4293, opened 2023-02-09
[DRAFT] increase compiler heap size for ci jobs

bowenliang123:

- increase compiler heap size for ci jobs, to see whether it helps to speed up build time

### DIFF
--- a/.github/workflows/dep.yml
+++ b/.github/workflows/dep.yml
@@ -46,7 +46,7 @@ jobs:
           check-latest: false
       - name: build
         env:
-          MAVEN_OPTS: -Dorg.slf4j.simpleLogger.defaultLogLevel=error
+          MAVEN_OPTS: -Dorg.slf4j.simpleLogger.defaultLogLevel=error -Dmaven.compiler.heap.size=4096m
         run: >-
           build/mvn clean install
           -Pflink-provided,spark-provided,hive-provided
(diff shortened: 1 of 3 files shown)


bowenliang123:
Increasing memory size (to 4GB) contributes to build time reduction , but not much.
Build time seems more rely on the utility rate of the host machine.
But I think still good to do this, what do you think?
cc @cfmcgrady @pan3793

pan3793:
I prefer to keep as-is if there is no meaningful gain.

bowenliang123:
OK. Closing this PR.

cfmcgrady:
I doubt that this configuration really works? at least, it does not work on my local build. We need to change the max heap size in the script `build/mvn` instead of `pom.xml`.

bowenliang123:
I changed the MAVEN_OPTS in ci jobs which will be used in `build/mvn` in this line.

```
export MAVEN_OPTS=${MAVEN_OPTS:-"$_COMPILE_JVM_OPTS"}
```

cfmcgrady:
I mean change `-Xmx2g` to `-Xmx4g`, not the configuration of the plugin `maven-compiler-plugin`, it does not work on my local build.

https://github.com/apache/kyuubi/blob/7cd206642c962b233a86edfe17f0bb0c99e34c70/build/mvnd#L25

PR #4315, opened 2023-02-13
[INFRA] Check modules available before build step in dependency workflow

bowenliang123:

- Check whether modules are available by running `mvnd dependency:resolve`. if true, skip the `build` step of `mvnd install`.
- Reducing the time cost of dependency workflow from  ~ 10 min to ~4 min ([see log](https://github.com/apache/kyuubi/actions/runs/4162071562/jobs/7200786535)).

- [x] Pass CI workflows


cfmcgrady (approved):
LGTM

bowenliang123:
Thanks, merged to master.

PR #4323, opened 2023-02-14
Improve trino session context

ulysses-you:

This pr improves the trino session context:
1. always reuse the kyuubi session if session id exists, so we can restore the session context for next query
2. transform trino client information to kyuubi session, e.g. trino request source (trino-cli)

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/trino/api/TrinoContext.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/trino/api/TrinoContext.scala
@@ -30,7 +30,9 @@ import org.apache.hive.service.rpc.thrift.{TGetResultSetMetadataResp, TRowSet, T
 
 import org.apache.kyuubi.operation.OperationState.FINISHED
 import org.apache.kyuubi.operation.OperationStatus
+import org.apache.kyuubi.server.trino.api.Query.KYUUBI_SESSION_ID
 
+// TODO: Support replace `preparedStatement` for Trino-jdbc
 /**
  * The description and functionality of trino request
  * and response's context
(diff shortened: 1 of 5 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/trino/api/TrinoContext.scala
```diff
@@ -140,15 +142,17 @@ object TrinoContext {
...
-    trinoContext.catalog.foreach(
-      responseBuilder.header(TRINO_HEADERS.responseSetCatalog, _))
-    trinoContext.schema.foreach(
-      responseBuilder.header(TRINO_HEADERS.responseSetSchema, _))
+    // Note, We have injected kyuubi session id to session context so that the next query can find
+    // the previous session to restore the query context.
+    // It's hard to follow the Trino style that set all context to http headers.
+    // Because we do not know the context at server side. e.g. `set k=v`, `use database`.
+    // We also can not inject other session context into header before we supporting to map
+    // query result to session context.
```
ulysses-you, line 150: cc @iodone  
We can not inject session context using the current request header. e.g. the current query is `set k=v2` but he exists request header is `k=v`. So we only inject kyuubi session id to help next query find session context.

iodone: Instead of returning all session information to the client with HTTP response header, set it to sessionConfig and use kyuubi session to save the session information in server side?

ulysses-you: yea, I think it's not the trino way but more suitable for Kyuubi

iodone: Nice, elegant solutions

ulysses-you:
thanks, merging to master/1.7

PR #4328, opened 2023-02-14
Make Trino jdbc driver work

ulysses-you:

according to `io.trino.jdbc.ColumnInfo`, there are some type requring signature parameter.
- varchar(n)
- char(n)
- decimal(precision, scale)

It failed with trino jdbc now
image


ulysses-you:
cc @Yikf @iodone @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/server/trino/api/TrinoScalaObjectMapper.scala
```diff
@@ -19,14 +19,54 @@ package org.apache.kyuubi.server.trino.api
...
     .registerModule(new Jdk8Module)
 
+  /**
+   * Not sure why jackson annotations do not work for some Trino dto, e.g. `ClientTypeSignature`.
+   * Without this change, it will throw exception during serializing when return back response.
+   * This is a workaround to be compatible with Trino json serialization.
+   * TODO: Remove this when find root cause
+   */
+  private def serializeTrinoClientTypeModule(): Module = {
+    val module = new SimpleModule("trinoSerializer")
```
iodone, line 42: image
Can we refer the implementaion of `QueryResults` codec in trino client ?

ulysses-you: ah, good point. I have validated it works

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/VersionInfo.java
```diff
@@ -26,10 +29,13 @@ public class VersionInfo {
 
   public VersionInfo() {}
 
-  public VersionInfo(String version) {
+  // Explicitly specifies JsonProperty to be compatible if disable auto detect feature
+  @JsonCreator
+  public VersionInfo(@JsonProperty("version") String version) {
```
ulysses-you, line 34: This is the only one dto which is used by trino and kyuubi. Since we disable auto detect feature for Trino, Explicitly specifies JsonProperty here to be compatible both of them.

PR #4334, opened 2023-02-15
[REST] Rest client should catch `NoHttpResponse Exception` and retry

zwangsheng:

[NoHttpResponseException](https://hc.apache.org/httpclient-legacy/exception-handling.html)

> In some circumstances, usually when under heavy load, the web server may be able to receive requests but unable to process them. A lack of sufficient resources like worker threads is a good example. This may cause the server to drop the connection to the client without giving any response. HttpClient throws NoHttpResponseException when it encounters such a condition. In most cases it is safe to retry a method that failed with NoHttpResponseException.

In case of Kyuubi Server is overloaded and does not have enough resources to allocate threads to handle the corresponding request, the request will be dropped and no response will be returned, Kyuubi Rest Client should catch this exception and retry.


zwangsheng:
cc @yaooqinn @ulysses-you @pan3793 @turboFei

bowenliang123 (approved):

ulysses-you (approved):

pan3793:
Thanks, merged to master/1.7/1.6

PR #4344, opened 2023-02-16
Expose exec pool work queue size metrics

turboFei:

It can help to know the backend pressure if the exec pool is full.


pan3793 (approved):

bowenliang123:
Good to have the queue size metrics exposed. But where do we have the `work queue` concept?
As the metrics come from the queue size in `execPool`, is it good enough to have `exec.pool.queue.size` as metric name instead?

turboFei:
I think work queue should be fine, it is a normal concept in `ThreadPoolExecutor`.

```
    /**
     * The queue used for holding tasks and handing off to worker
     * threads.  We do not require that workQueue.poll() returning
     * null necessarily means that workQueue.isEmpty(), so rely
     * solely on isEmpty to see if the queue is empty (which we must
     * do for example when deciding whether to transition from
     * SHUTDOWN to TIDYING).  This accommodates special-purpose
     * queues such as DelayQueues for which poll() is allowed to
     * return null even if it may later return non-null when delays
     * expire.
     */
    private final BlockingQueue<Runnable> workQueue;
```

bowenliang123:
OK. Just mention it as we don't have work queue in kyuubi level, and bring it from lower level.

bowenliang123 (approved):

Yikf (approved):

cxzl25 (approved):

turboFei:
thanks all, merging to master

PR #4348, opened 2023-02-16
[INFRA] Cache engine archives in CI jobs for maven download plugin

bowenliang123:

- to avoid violation in Apache archives website's daily total download size quote, prevent repeatedly downloading engine archives by manual cache in actions

- [ ] Pass CI jobs and check cached engine archives


pan3793 (approved):

pan3793:
Thanks, merged to master, we need a period to evaluate the improvement.

bowenliang123:
> Thanks, merged to master, we need a period to evaluate the improvement.

Agree. And we could extract a separate job that downloads all engine achives prior to all the integration tests for better improvement.

PR #4354, opened 2023-02-17
[KYUUBI#4343] Add support initializationSQLs from file

### LINKED ISSUE #4343: [Improvement] Support loading initializationSQLs from file
### Code of Conduct

### Search before asking

 Support loading initializationSQLs from file

### How should we improve?

Initialize the sql by configuring an hdfs file path and then reading the file contents

thomasg19930417:

close #4343


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -52,7 +59,49 @@ object KyuubiSparkUtil extends Logging {
...
+          val hadoopConf = sparkSession.sparkContext.hadoopConfiguration
+          val fs: FileSystem = engineInitSqlFilePath.getFileSystem(hadoopConf)
+          if (fs.exists(engineInitSqlFilePath)) {
+            val sqlInputStream: FSDataInputStream = fs.open(engineInitSqlFilePath)
+            reader = new BufferedReader(new InputStreamReader(sqlInputStream))
+            // Filter out the sql using '--' comment
+            val sqlText = ListBuffer[String]()
+            Stream.continually(reader.readLine())
+              .takeWhile(null != _).filter(item => !item.startsWith("--") && item.length > 0)
+              .map(item => item.replace(";", "")).foreach(item => sqlText.append(item))
```
pan3793, line 79: The implementation looks fragile ... have you considered 
```
select regexp_replace(html, '\\\\n|\\\\\t|\\\\\r', ';')
```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1391,6 +1393,30 @@ object KyuubiConf {
       .toSequence(";")
       .createWithDefault(Nil)
 
+  val ENGINE_INITIALIZE_SQL_TYPE: ConfigEntry[String] =
+    buildConf("kyuubi.engine.initialize.sql.type")
+      .doc(s"When set kyuubi.engine.initialize.sql.type=${InitSqlType.PATH}" +
+        s",you can use the hdfs path  set ${ENGINE_SESSION_INITIALIZE_SQL} " +
+        s"and ${ENGINE_INITIALIZE_SQL}")
+      .version("1.6.2")
```
pan3793, line 1401: it should be 1.8.0

pan3793:
@thomasg19930417 Thanks for your contribution, some tips for new contributors, please follow the [developer guide](https://kyuubi.readthedocs.io/en/master/develop_tools/developer.html#format-all-code) to fix code style and regenerate docs since you add new configurations.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -52,7 +59,49 @@ object KyuubiSparkUtil extends Logging {
     }
   }
 
+  def getInitializeSql(sparkSession: SparkSession, config: Seq[String]): Seq[String] = {
+
+    def getSqlTextFromFile(path: Option[String]): Seq[String] = {
+      path.map(filePath => {
+        var reader: BufferedReader = null
+        try {
+          info(s"========>load init sql from file: ${filePath}")
```
pan3793, line 68: please remove the leading `========>`, and keep log clear and simple

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
thomasg19930417: OK, I will modify it

bowenliang123:
@thomasg19930417 Can you show more detailed concerns and add more description to this PR,  including and not limited to the purpose, the changes, the config it provides, the range this PR covering and the consequence and sequence of them. 

As for me, I would  suggest,
1. provide a config to decide the source(s) type of initial SQLs, rather than just chaining them in preset oreder
2. please clarify the source type you suggested to support, local file, any remote filesystem or else. HDFS from the issue description might not be urgent for first choice to introduced.
3. the limitation to the path(s), either one or more paths supported, any other limit to this.
4. what's the init sql file meant or treated to non-SQL language type, as Scala, Python and etc also supported now in Kyuubi.

thomasg19930417:
> @thomasg19930417 Can you show more detailed concerns and add more description to this PR, including and not limited to the purpose, the changes, the config it provides, the range this PR covering and the consequence and sequence of them.
> 
> As for me, I would suggest,
> 
> 1. provide a config to decide the source(s) type of initial SQLs, rather than just chaining them in preset oreder
> 2. please clarify the source type you suggested to support, local file, any remote filesystem or else. HDFS from the issue description might not be urgent for first choice to introduced.
> 3. the limitation to the path(s), either one or more paths supported, any other limit to this.
> 4. what's the init sql file meant or treated to non-SQL language type, as Scala, Python and etc also supported now in Kyuubi.



> @thomasg19930417 Can you show more detailed concerns and add more description to this PR, including and not limited to the purpose, the changes, the config it provides, the range this PR covering and the consequence and sequence of them.
> 
> As for me, I would suggest,
> 
> 1. provide a config to decide the source(s) type of initial SQLs, rather than just chaining them in preset oreder
> 2. please clarify the source type you suggested to support, local file, any remote filesystem or else. HDFS from the issue description might not be urgent for first choice to introduced.
> 3. the limitation to the path(s), either one or more paths supported, any other limit to this.
> 4. what's the init sql file meant or treated to non-SQL language type, as Scala, Python and etc also supported now in Kyuubi.

At present, initialization sql is controlled by two parameters, kyuubi.engine.initialize.sql and kyuubi.engine.session.initialize.sql. These two configurations are configured in kyuubi-defult. Assuming that there are many initial udf functions to be created, I think it would be better to configure them in a separate file, Therefore, I added a parameter kyuubi.engine.initialize.sql.type here to control whether to initialize or read the initialized sql from the file
About initSql, I think the current implementation only supports sql.

thomasg19930417:
@bowenliang123 
At present, initialization sql is controlled by two parameters, kyuubi.engine.initialize.sql and kyuubi.engine.session.initialize.sql. These two configurations are configured in kyuubi-defult. Assuming that there are many initial udf functions to be created, I think it would be better to configure them in a separate file, Therefore, I added a parameter kyuubi.engine.initialize.sql.type here to control whether to initialize or read the initialized sql from the file
About initSql, I think the current implementation only supports sql.

thomasg19930417:
@pan3793 
I resubmitted the code. Thank you for your help under review

PR #4304, opened 2023-02-10
[KYUUBI #4202] [AUTHZ] [HELP-WANTED] fix data masking for DataSourceV2Relation

bowenliang123:
### _Why are the changes needed?_

To close #4202 .

Errors when applying data masking rules to DatasourceRelationV2:
image

- [x] add unit test for simulation
- [ ] align exprId in Project in higher level to the lower one  of the logical plan

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] [Run test](https://kyuubi.readthedocs.io/en/master/develop_tools/testing.html#running-tests) locally before make a pull request

Issue #4202, opened 2023-01-25
[Bug] Issue when applying Masking policies on Iceberg Tables


praveenkumarb1207:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Following the instructions in https://github.com/apache/kyuubi/blob/master/docs/security/authorization/spark/install.md , I have installed Kyuubi plugin by copying all the jars and required configuration files as mentioned in the link to $SPARK_HOME .

**Following is the version Information :** 

Ranger Version - 2.3.0
Spark Version   - Spark 3.2.3  with Hadoop 3.3.4
Apache Hive     - 3.1.2

I have created an Iceberg table in Hive using spark .

**spark-shell command :** 

```
spark-shell --packages "org.apache.spark:spark-hive_2.12:3.2.3,org.apache.hadoop:hadoop-aws:3.3.4,org.apache.iceberg:iceberg-spark-runtime-3.2_2.12:1.1.0" \
    --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,org.apache.kyuubi.plugin.spark.authz.ranger.RangerSparkExtension \
    --conf spark.sql.catalog.spark_catalog=org.apache.iceberg.spark.SparkSessionCatalog \
    --conf spark.sql.catalog.spark_catalog.type=hive \
    --conf spark.sql.catalog.demo=org.apache.iceberg.spark.SparkCatalog \
    --conf spark.sql.catalog.demo.type=hadoop \
    --conf spark.sql.catalog.demo.warehouse=$PWD/warehouse --proxy-user pb1207	
```
**Code :** 

```
import spark.implicits._
  
  val df = Seq(
  
    (1, "America"),
  
    (2, "India"),
  
    (3, "London")
  
  ).toDF("id", "country")
  
df.write.format("iceberg").mode("overwrite").option("path","s3a:/bucket/test/test_iceberg.parquet").saveAsTable("test.test_iceberg_table")
```

Iceberg table got successfully created in Hive . 

**code :** 

`spark.sql("show create table test.test_iceberg_table").show(false)`

**output :** 

![image](https://user-images.githubusercontent.com/59603987/214647412-f15969ba-3dca-4913-a02c-038cd3ab944b.png)

[...]

PR #4358, opened 2023-02-17
[KYUUBI #4202] Fix reference resolution when data masking enabled for V2 relations

yaooqinn:

Follow up to close #4304 with a complete solution.

Before this PR, we did row filtering, and data masking in the same analyzer rule as both of them added patterns upon the scans, which are leaf nodes.

After this, we separate them into two different rules. Because the filtering still adds filter patterns upon the leaves, but the masking now adds masker expressions upon the leaves first and later on the root of a query.

The reason why we move the maskers from leaves to the root is to fix attribute resolution errors. The errors #4202 occur when we try to insert a node with resolved expressions between the root and its child, while if the parents are resolved ahead, the newly added expressions may become missing attributes. The new approach can avoid the bug.

The full data masking rule contains two separate stages.

 * Step1: RuleApplyDataMaskingStage0
    - lookup the full plan for supported scans
    - once found, get masker configuration from external column by column
    - use spark sql parser to generate an unresolved expression for each masker
    - add a projection with new output on the right top of the original scan if the output has      changed
    - Add DataMaskingStage0Marker to track the original expression and its masker expression.
 
 * Step2: Spark native rules will resolve our newly added maskers
 
 * Step3: RuleApplyDataMaskingStage1
   - It will fulfill the missing attributes with a related masker expression buffered by DataMaskingStage0Marker.


yaooqinn:
cc @praveenkumarb1207 @bowenliang123 @pan3793

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/RuleEliminateMarker.scala
```diff
@@ -22,6 +22,9 @@ import org.apache.spark.sql.catalyst.rules.Rule
 
 class RuleEliminateMarker extends Rule[LogicalPlan] {
   override def apply(plan: LogicalPlan): LogicalPlan = {
-    plan.transformUp { case rf: RowFilterAndDataMaskingMarker => rf.child }
+    plan.transformUp {
+      case rf: DataMaskingMarker => rf.child
+      case rf: RowFilterMarker => rf.child
```
bowenliang123, line 27:
```suggestion
      case marker: DataMaskingMarker | RowFilterMarker => marker.child
```

bowenliang123, line 27:
```suggestion
      case DataMaskingMarker(child, _) => child
      case RowFilterMarker(child) => child
```

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleApplyDataMasking.scala
```diff
@@ -0,0 +1,107 @@
...
+        val queries = tableCommandSpec.queries(cmd)
+        cmd.mapChildren {
+          case marker: DataMaskingMarker => marker
+          case query if queries.contains(query) && query.resolved =>
+            applyDataMasking(query, ugi, opType)
+          case o => o
+        }
+      case cmd: Command if cmd.childrenResolved =>
+        cmd.mapChildren(applyDataMasking(_, ugi, opType))
+      case cmd: Command => cmd
```
bowenliang123, line 53: Is there ut for commands with masking rules applied in V2?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleApplyRowFilter.scala
```diff
@@ -45,46 +46,21 @@ class RuleApplyRowFilterAndDataMasking(spark: SparkSession) extends Rule[Logical
...
-    }
-
-    if (filterExprStr.isEmpty) {
-      Project(newOutput, RowFilterAndDataMaskingMarker(plan))
-    } else {
-      val filterExpr = parse(filterExprStr.get)
-      Project(newOutput, Filter(filterExpr, RowFilterAndDataMaskingMarker(plan)))
-    }
+    val filterExpr = SparkRangerAdminPlugin.getFilterExpr(art).map(parse)
+    filterExpr.foldLeft(plan)((p, e) => Filter(e, RowFilterMarker(p)))
```
bowenliang123, line 64:
```suggestion
    filterExpr.foldLeft(plan)((p, expr) => Filter(expr, RowFilterMarker(p)))
```
Maybe it's better for understanding with the var name `expr` ?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/DataMaskingMarker.scala
```diff
@@ -29,3 +29,11 @@ case class RowFilterAndDataMaskingMarker(child: LogicalPlan) extends UnaryNode
     copy(child = newChild)
 
 }
+case class RowFilterMarker(child: LogicalPlan) extends UnaryNode with WithInternalChild {
+
+  override def output: Seq[Attribute] = child.output
+
+  override def withNewChildInternal(newChild: LogicalPlan): LogicalPlan =
+    copy(child = newChild)
```
bowenliang123, line 37:
```suggestion
  override def withNewChildInternal(newChild: LogicalPlan): LogicalPlan = copy(child = newChild)
```

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleApplyDataMasking.scala
yaooqinn: yes

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/RuleEliminateMarker.scala
bowenliang123: duplicated `rf` is more like the short form of `rowfilter` and not suitable for both DataMaskingMarker and RowFilterMarker. And in suggested style we could just skip naming it .

bowenliang123 (approved):
LGTM overall, as CI passes.
I would suggest summarizing all the key changes and refactoring designs to the description as hints for better understanding.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/RuleEliminateMarker.scala
yaooqinn: The suggested style is not a good habit. it makes refactoring more difficult and the code more error-prone.

yaooqinn:
@bowenliang123 does the tests need to be refined? It looks less rigorous to me

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/util/RuleEliminateMarker.scala
bowenliang123:
```
      case dm: DataMaskingMarker => dm.child
      case rf: RowFilterMarker => rf.child
```
OK, back to the existing style. How about this?

bowenliang123:
> @bowenliang123 does the tests need to be refined? It looks less rigorous to me

Good to have more tests on `create table as select` or select with joins ?

yaooqinn:
> > @bowenliang123 does the tests need to be refined? It looks less rigorous to me
> 
> Good to have more tests on `create table as select` or select with joins ?

Can you make a follow-up to make the original `data masking` tests work for v2 relations, there is a seq for us to add new SQL patterns, we do not need new tests

Jackhjf:
This item previously existed in two issues：

1. The fields of the two tables are the same
    create two tables: create table test(id string,name string),create table test_copy(id string,name string),Among them, the test1 table configures the masking strategy，use this sql:insert into test select *from test_copy， The data found in the test_copy is not masking data
2. The fields of the two tables are not the same
      create two tables: create table test(id1 string,name1 string),create table test_copy(id string,name string),Among them, the test1 table configures the masking strategy，use this sql:insert into test select *from test_copy， The data found in the test_copy is not masking data

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/RuleApplyDataMaskingStage0.scala
```diff
@@ -0,0 +1,74 @@
...
+ *   - lookup the full plan for supported scans
+ *   - once found, get masker configuration for external column by column
+ *   - use spark sql parser to generate a unresolved expression for each masker
+ *   - add a projection with new output on the right top of the original scan if the output has
+ *     changed
+ *   - Add DataMaskingStage0Marker to track the original expression and its masker expression.
+ *
+ * Step2: Spark native rules will resolved our newly added maskers
+ *
+ * Step3: [[RuleApplyDataMaskingStage1]]
```
bowenliang123, line 42: These comments are key designs for the refactor in this PR. Better to copy them to the PR description as well.

bowenliang123 (approved):
LGTM. Thanks for the hard work with the complete solution and test cases for refactoring data masking. 
Consider changing the PR name to `Refactoring data masking rules for reference resolution in V2 relations` in order to highlight the importance.

yaooqinn:
thanks, merged to master and 1.7. also cc @pan3793, the RM of v1.7.0

PR #4369, opened 2023-02-19
[KYUUBI #4338][FOLLOWUP] Fix K8s integration tests

pan3793:

Correct the Spark image tag to recover "Kyuubi Server On Kubernetes Integration Test"

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -434,7 +434,6 @@ jobs:
         run: >-
           ./build/mvn ${MVN_OPT} clean install
           -Pflink-provided,hive-provided
-          -Pspark-3.2
           -Pkubernetes-it
           -Dtest=none -DwildcardSuites=org.apache.kyuubi.kubernetes.test.spark
       - name: Print Driver Pod logs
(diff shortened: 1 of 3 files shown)


integration-tests/kyuubi-kubernetes-it/src/test/scala/org/apache/kyuubi/kubernetes/test/deployment/KyuubiOnKubernetesTestsSuite.scala
```diff
@@ -54,7 +54,7 @@ class KyuubiOnKubernetesWithSparkTestsBase extends WithKyuubiServerOnKubernetes
     super.connectionConf ++
       Map(
         "spark.master" -> s"k8s://$miniKubeApiMaster",
-        "spark.kubernetes.container.image" -> "apache/spark:3.3.2",
+        "spark.kubernetes.container.image" -> "apache/spark:v3.3.2",
```
pan3793, line 57: https://hub.docker.com/r/apache/spark/tags

.github/workflows/master.yml
```diff
@@ -434,7 +434,6 @@ jobs:
         run: >-
           ./build/mvn ${MVN_OPT} clean install
           -Pflink-provided,hive-provided
-          -Pspark-3.2
```
Yikf, line 437: It's better to specify `spark-3.3`.

Apache Spark version of the kubernetes-it is currently definitive and unique, It should be consist with [kubernetes-it](https://github.com/apache/kyuubi/blob/master/integration-tests/kyuubi-kubernetes-it/src/test/scala/org/apache/kyuubi/kubernetes/test/spark/SparkOnKubernetesTestsSuite.scala#L48) to to prevent serialization problems in k8s-client mode.

Before we finish this [TODO](https://github.com/apache/kyuubi/blob/master/integration-tests/kyuubi-kubernetes-it/src/test/scala/org/apache/kyuubi/kubernetes/test/spark/SparkOnKubernetesTestsSuite.scala#L45), it's a good idea to specify a profile in case we change the default spark version of kyuubi and spark docker doesn't have mirroring issues with that version

Yikf (approved):

.github/workflows/master.yml
pan3793: Yes, exactly.

pan3793:
Thanks, merged to master/1.7

PR #4372, opened 2023-02-20
Support to return null value for OperationsResource rowset

turboFei:

With restful api, for query `select cast(null as int) c1`

#### Before
the result is `0`

#### After
the result is `null`.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
@@ -222,7 +222,7 @@ private[v1] class OperationsResource extends ApiRequestContext with Logging {
                 if (i.getI64Val.isSetValue) {
                   i.getI64Val.getFieldValue(TI64Value._Fields.VALUE)
                 } else {
-                  null;
+                  null
                 }
             })
         }).asJava)
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
@@ -183,43 +183,43 @@ private[v1] class OperationsResource extends ApiRequestContext with Logging {
             i.getSetField.name(),
             i.getSetField match {
               case TColumnValue._Fields.STRING_VAL =>
-                if (i.isSetStringVal) {
+                if (i.getStringVal.isSetValue) {
                   i.getStringVal.getFieldValue(TStringValue._Fields.VALUE)
                 } else {
                   null
                 }
               case TColumnValue._Fields.BOOL_VAL =>
-                if (i.isSetBoolVal) {
+                if (i.getBoolVal.isSetValue) {
                   i.getBoolVal.getFieldValue(TBoolValue._Fields.VALUE)
                 } else {
                   null
                 }
               case TColumnValue._Fields.BYTE_VAL =>
-                if (i.isSetByteVal) {
+                if (i.getByteVal.isSetValue) {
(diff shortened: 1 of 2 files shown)


turboFei:
cc @pan3793

turboFei:
cc @simon824

bowenliang123 (approved):
LGTM. Good catch.

And can we tract a general method `getFieldValueOrNull`for this ?

pan3793 (approved):

bowenliang123:
> LGTM. Good catch.
> 
> And can we tract a general method `getFieldValueOrNull`for this ?

turboFei:
> And can we tract a general method getFieldValueOrNullfor this ?

@bowenliang123  I thought about that, but it only be used for one time.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
```diff
@@ -183,19 +183,47 @@ private[v1] class OperationsResource extends ApiRequestContext with Logging {
             i.getSetField.name(),
             i.getSetField match {
               case TColumnValue._Fields.STRING_VAL =>
-                i.getStringVal.getFieldValue(TStringValue._Fields.VALUE)
+                if (i.getStringVal.isSetValue) {
```
bowenliang123, line 186: A possible risk is that `getStringVal` on the unset field will throw RuntimeException here.
Maybe use `i.isSetStringVal` for the case condition is better? And make null as default return.

turboFei: I tried to use that, but it does not work. you can have a try.

turboFei: With 
```
case TColumnValue._Fields.STRING_VAL =>
                if (i.isSetStringVal) {
                  i.getStringVal.getFieldValue(TStringValue._Fields.VALUE)
                } else {
                  null
                }
              case TColumnValue._Fields.BOOL_VAL =>
                if (i.isSetBoolVal) {
                  i.getBoolVal.getFieldValue(TBoolValue._Fields.VALUE)
                } else {
                  null
                }
              case TColumnValue._Fields.BYTE_VAL =>
                if (i.isSetByteVal) {
                  i.getByteVal.getFieldValue(TByteValue._Fields.VALUE)
                } else {
                  null
                }
              case TColumnValue._Fields.DOUBLE_VAL =>
                if (i.isSetDoubleVal) {
                  i.getDoubleVal.getFieldValue(TDoubleValue._Fields.VALUE)
                } else {
                  null
                }
              case TColumnValue._Fields.I16_VAL =>
                if (i.isSetI16Val) {
                  i.getI16Val.getFieldValue(TI16Value._Fields.VALUE)
                } else {
                  null
                }
              case TColumnValue._Fields.I32_VAL =>
                if (i.isSetI32Val) {
                  i.getI32Val.getFieldValue(TI32Value._Fields.VALUE)
                } else {
                  null
                }
              case TColumnValue._Fields.I64_VAL =>
                if (i.isSetI64Val) {
                  i.getI64Val.getFieldValue(TI64Value._Fields.VALUE)
                } else {
                  null
                }
```

The result is:

```
ArrayBuffer(null, false, 0, 0.0, 0, 0, 0)
```

turboFei: so, to keep the style align.

I decided to use `i.get*Val.isSetValue` finally.

bowenliang123: Yeah, I have tried but failed before raising the original comment. Let's keep it as it is.

turboFei:
thanks, merging to master and branch-1.7

Issue #4379, opened 2023-02-20
[Improvement] Exclude python envirnoment in spoltless maven plugin


a49a:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, the spotless maven plugin checks python virtual environment paths like `docs/kyuubi/lib/python*/*md`. We should exclude them.

> [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.30.0:check (default) on project kyuubi-parent: The following files had format violations:
[ERROR]     docs/kyuubi/lib/python3.8/site-packages/sphinx_book_theme/assets/translations/README.md
[ERROR]         @@ -41,7 +41,6 @@
[ERROR]          ···```bash
[ERROR]          ···prettier·sphinx_book_theme/translations/jsons/<message·name>.json
[ERROR]          ···```
[ERROR]         -
[ERROR]          8.·Run·`python·src/sphinx_book_theme/_compile_translations.py`
[ERROR]          9.·This·will·generate·the·locale·files·(`.mo`)·that·Sphinx·uses·in·its·translation·machinery,·and·put·them·in·`locales/<language-code>/LC_MESSAGES/<msg>.mo`.
[ERROR]          
[ERROR]     docs/kyuubi/lib/python3.9/site-packages/sphinx_book_theme/assets/translations/README.md
[ERROR]         @@ -41,7 +41,6 @@
[ERROR]          ···```bash
[ERROR]          ···prettier·sphinx_book_theme/translations/jsons/<message·name>.json
[ERROR]          ···```
[ERROR]         -
[ERROR]          8.·Run·`python·src/sphinx_book_theme/_compile_translations.py`
[ERROR]          9.·This·will·generate·the·locale·files·(`.mo`)·that·Sphinx·uses·in·its·translation·machinery,·and·put·them·in·`locales/<language-code>/LC_MESSAGES/<msg>.mo`.
[ERROR]      

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4384, opened 2023-02-20
[KYUUBI #4379][BUILD] Exclude python virtual environments in the spotless maven plugin

a49a:

Exclude python virtual environments in the spotless maven plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2074,6 +2074,9 @@
                                 <include>*/README.md</include>
                                 <include>docker/playground/README.md</include>
                             </includes>
+                            <excludes>
+                                <exclude>docs/kyuubi/**/*.md</exclude>
+                            </excludes>
                             <flexmark>
                                 <version>${flexmark.version}</version>
                             </flexmark>
--- a/pom.xml
+++ b/pom.xml
@@ -2075,7 +2075,7 @@
                                 <include>docker/playground/README.md</include>
                             </includes>
                             <excludes>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -2074,6 +2074,9 @@
                                 <include>*/README.md</include>
                                 <include>docker/playground/README.md</include>
                             </includes>
+                            <excludes>
+                                <exclude>docs/kyuubi/**/*.md</exclude>
```
pan3793, line 2078: are you supposing that the virtual env is named "kyuubi"?

a49a: I have changed the path to be more general.

pom.xml
```diff
@@ -2074,6 +2074,9 @@
                                 <include>*/README.md</include>
                                 <include>docker/playground/README.md</include>
                             </includes>
+                            <excludes>
+                                <exclude>docs/*/lib/python*/**/*.md</exclude>
+                            </excludes>
```
pan3793, line 2079: what will happen if both `includes` and `excludes` is configured? cc @bowenliang123

bowenliang123: https://github.com/diffplug/spotless/blob/maven/2.30.0/plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java#L284

bowenliang123 (approved):
LGTM

PR #4390, opened 2023-02-21
Allow user to provide batch id on submitting batch job

pan3793:

This PR proposes to allow the user to provide a batch id on submitting a batch job. If the batch id already existed in metastore, Kyuubi ignores this submission and just returns the existing one, w/ a marker in response, this could avoid duplicated batch job submission.

Talking about the implementation, the key things are

How does the user set the custom batch id?

- User can optionally set the `kyuubi.batch.id` in `conf: Map[String, String]`, and the value must be a UUID, for Java users, it can be generated by `UUID.randomUUID().toString()`

How does the Kyuubi Server detect the duplication?

- It's simple in single Kyuubi Server instance case, Kyuubi just needs to look up the metastore before creating a batch job
- In HA mode, suppose the user requests to create the batch jobs w/ the same batch id concurrently, multiple Kyuubi Servers may process the request and try to insert to metastore DB at the same time, but only the first insertion success, others will fail w/ "duplicated key", Kyuubi Server needs to catch this error and return the existing batch job information instead of creating a new one.

How does the user know if the returned batch job is new created or duplicated?

- a new field `batchInfo: Map[String, String]` is added to the response, and for duplicated batch job, `"kyuubi.batch.duplicated": "true"` will be contained.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -19,13 +19,14 @@ package org.apache.kyuubi.server.api.v1
 
 import java.io.InputStream
 import java.util
-import java.util.{Collections, Locale}
+import java.util.{Collections, Locale, UUID}
 import java.util.concurrent.ConcurrentHashMap
 import javax.ws.rs._
 import javax.ws.rs.core.MediaType
 import javax.ws.rs.core.Response.Status
 
 import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
 import scala.util.control.NonFatal
 
 import io.swagger.v3.oas.annotations.media.{Content, Schema}
@@ -45,6 +46,7 @@ import org.apache.kyuubi.server.api.v1.BatchesResource._
 import org.apache.kyuubi.server.metadata.MetadataManager
 import org.apache.kyuubi.server.metadata.api.Metadata
 import org.apache.kyuubi.session.{KyuubiBatchSessionImpl, KyuubiSessionManager, SessionHandle}
+import org.apache.kyuubi.util.JdbcUtils
 
 @Tag(name = "Batch")
 @Produces(Array(MediaType.APPLICATION_JSON))
@@ -105,7 +107,8 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       session.connectionUrl,
       batchOpStatus.state.toString,
       session.createTime,
-      batchOpStatus.completed)
+      batchOpStatus.completed,
+      Map.empty[String, String].asJava)
   }
 
   private def buildBatch(
@@ -142,7 +145,8 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
         metadata.kyuubiInstance,
         currentBatchState,
         metadata.createTime,
-        metadata.endTime)
+        metadata.endTime,
+        Map.empty[String, String].asJava)
     }.getOrElse(MetadataManager.buildBatch(metadata))
   }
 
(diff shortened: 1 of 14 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -210,22 +210,35 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
-    val sessionHandle = sessionManager.openBatchSession(
-      userName,
-      "anonymous",
-      ipAddress,
-      request.getConf.asScala.toMap,
-      request)
-    buildBatch(sessionManager.getBatchSessionImpl(sessionHandle))
+    val userProvidedBatchId = request.getConf.asScala.get(KYUUBI_SESSION_BATCH_ID_KEY)
+      // the user provides batch_id must be an UUID
+      .map { batchId => UUID.fromString(batchId); batchId }
```
turboFei, line 215: should be provide the more detail error message if the specified id is not an UUID.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -210,22 +210,35 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+
+    userProvidedBatchId.flatMap { batchId =>
+      Option(sessionManager.getBatchFromMetadataStore(batchId))
+    } match {
+      case Some(batch) =>
+        warn(s"duplicated submission: ${batch.getId}, ignore and return the existing batch.")
+        batch
+      case None =>
```
turboFei, line 223: if the existing batch is None.

we need remove the `KYUUBI_SESSION_BATCH_ID_KEY` from batchRequest.getConf

turboFei (changes requested):
left some comments

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -210,22 +210,35 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
-        KYUUBI_SESSION_BATCH_RESOURCE_UPLOADED_KEY -> isResourceFromUpload.toString,
-        KYUUBI_SESSION_REAL_USER_KEY -> fe.getRealUser())).asJava)
-    val sessionHandle = sessionManager.openBatchSession(
-      userName,
-      "anonymous",
-      ipAddress,
-      request.getConf.asScala.toMap,
-      request)
-    buildBatch(sessionManager.getBatchSessionImpl(sessionHandle))
+    val userProvidedBatchId = request.getConf.asScala.get(KYUUBI_SESSION_BATCH_ID_KEY)
```
turboFei, line 213: I recommend to new a java conf map, and use `map.remove()`.

turboFei:
seems mistake, let me think it twice.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
turboFei: mistake.

If batchId is specified,
- if the batch is existing, return the batch.
- if the batch does exist, use the specified batchId


If no batchId specified, generate a new one

ulysses-you:
How about add a `traceToken` in rest client ? Trino support that and Kyuubi server should return traceToken back to client. Then developer can deduplicate by itself.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
pan3793: updated

pan3793:
> How about add a `traceToken` in rest client ? Trino support that and Kyuubi server should return traceToken back to client. Then developer can deduplicate by itself.

I think it makes things more complex, IMO let the user provide a custom batch id does not has many risks, we already restrict the batch id MUST be a UUID.

pan3793:
cc @turboFei @ulysses-you it's ready for the next review.

ulysses-you (approved):

turboFei (approved):

turboFei:
TODO: render the batchInfo in kyuubi-ctl

pan3793:
> render the batchInfo in kyuubi-ctl

Tracking in https://github.com/apache/kyuubi/issues/4438

pan3793:
Thanks, merged to master/1.7

Issue #4388, opened 2023-02-21
[Improvement] Limit the max rows to get nextRowSet


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

It is necessary to protect the kyuubi gateway.

### How should we improve?

If the max rows specified by client side is larger than the limitation, fail the request directly.

We can not choose the minimum of max rows specified by client side and server side limitation, it might break the logical that client determine whether the result set fetch is finished.

For example, for rest api:
<img width="648" alt="image" src="https://user-images.githubusercontent.com/6757692/220239537-d545e89d-3af3-4dbe-8509-21539c5bf355.png">

Customer might check whether `rowCount < maxRows`, if we choose the minimum, it will cause customer mistake to think that all result set has been fetched.

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

turboFei:
cc @lightning-L

PR #4395, opened 2023-02-21
[KYUUBI #4388] Limit the max rows for get nextRowSet api

lightning-L:


lightning-L:
@turboFei

docs/deployment/settings.md
```diff
@@ -440,6 +440,7 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
+| kyuubi.operation.rest.fetch.max.rows            | 5000                                                                            | Max rows limit for getting result row set api. If the max rows specified by client-side is larger than the limit, request will fail directly.                                                                                                                                                                                                                                                                                                                                                            | int      | 1.7.0 |
```
turboFei, line 443: not only for rest call, we need also limit the jdbc

docs/deployment/settings.md
```diff
@@ -431,6 +431,7 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
+| kyuubi.operation.client.fetch.max.rows          | 5000                                                                            | Max rows limit for getting result row set operation. If the max rows specified by client-side is larger than the limit, request will fail directly.                                                                                                                                                                                                                                                                                                                                                      | int      | 1.7.0 |
```
turboFei, line 434: Refer the style liks:
```
kyuubi.server.limit.connections.per.user
``


how about:

```
kyuubi.server.limit.client.fetch.max.rows
```

Comment (turboFei) [reply docs/deployment/settings.md:434]: refer the config:

  val SERVER_LIMIT_CONNECTIONS_PER_USER_IPADDRESS: OptionalConfigEntry[Int] =
    buildConf("kyuubi.server.limit.connections.per.user.ipaddress")
      .doc("Maximum kyuubi server connections per user:ipaddress combination." +
        " Any user-ipaddress exceeding this limit will not be allowed to connect.")
      .version("1.6.0")
      .serverOnly
      .intConf
      .createOptional

File: docs/deployment/settings.md [hunk 3]
```diff
@@ -455,6 +455,7 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
 | kyuubi.server.batch.limit.connections.per.user.ipaddress | &lt;undefined&gt; | Maximum kyuubi server batch connections per user:ipaddress combination. Any user-ipaddress exceeding this limit will not be allowed to connect.                                                                                                | int      | 1.7.0 |
 | kyuubi.server.info.provider                              | ENGINE            | The server information provider name, some clients may rely on this information to check the server compatibilities and functionalities. <li>SERVER: Return Kyuubi server information.</li> <li>ENGINE: Return Kyuubi engine information.</li> | string   | 1.6.1 |
+| kyuubi.server.limit.client.fetch.max.rows                | 5000              | Max rows limit for getting result row set operation. If the max rows specified by client-side is larger than the limit, request will fail directly.                                                                                            | int      | 1.7.0 |
```
turboFei, line 458:
```
kyuubi.server.limit.connections.per.user.ipaddress
kyuubi.server.batch.limit.connections.per.user.ipaddress
kyuubi.server.limit.client.fetch.max.rows   
```


should we rename the config `kyuubi.server.batch.limit.connections.per.user.ipaddress` to `kyuubi.server.limit.batch.connections.per.user.ipaddress` ? @pan3793

turboFei: raised #4398

kyuubi-common/src/main/scala/org/apache/kyuubi/service/AbstractBackendService.scala
```diff
@@ -201,6 +201,11 @@ abstract class AbstractBackendService(name: String)
       orientation: FetchOrientation,
       maxRows: Int,
       fetchLog: Boolean): TRowSet = {
+    val maxRowsLimit = conf.get(KyuubiConf.SERVER_LIMIT_CLIENT_FETCH_MAX_ROWS)
```
turboFei, line 204: how about using a lazy val.

```
private lazy val maxRowsLimit = conf.get(KyuubiConf.SERVER_LIMIT_CLIENT_FETCH_MAX_ROWS)
```

turboFei:
Af think twice， I think that we can set the default value to optional， to do not break the original behavior.

turboFei (approved):

turboFei:
thanks, merging to master

PR #4396, opened 2023-02-22
[KYUUBI #3876][FOLLOWUP] Update the rest api docs for open session response

turboFei:

### DIFF
--- a/docs/client/rest/rest_api.md
+++ b/docs/client/rest/rest_api.md
@@ -99,10 +99,10 @@ Create a session
 
 #### Response Body
 
-| Name           | Description                                                                                         | Type   |
-|:---------------|:----------------------------------------------------------------------------------------------------|:-------|
-| identifier     | The session handle identifier                                                                       | String |
-| kyuubiInstance | The Kyuubi instance that holds the session and to call for the following operations in the session. | String |
+| Name           | Description                                                                                        | Type   |
+|:---------------|:---------------------------------------------------------------------------------------------------|:-------|
+| identifier     | The session handle identifier                                                                      | String |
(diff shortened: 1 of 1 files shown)


docs/client/rest/rest_api.md
```diff
@@ -99,9 +99,10 @@ Create a session
 
 #### Response Body
 
-| Name       | Description                   | Type   |
-|:-----------|:------------------------------|:-------|
-| identifier | The session handle identifier | String |
+| Name           | Description                                                                                         | Type   |
+|:---------------|:----------------------------------------------------------------------------------------------------|:-------|
+| identifier     | The session handle identifier                                                                       | String |
+| kyuubiInstance | The Kyuubi instance that holds the session and to call for the following operations in the session. | String |
```
cxzl25, line 105: Look at the end of most of the Description without a dot`.`, can it be removed here?

cxzl25 (approved):

docs/client/rest/rest_api.md
turboFei: thanks, fixed

turboFei:
> What's [HADP-3876]?

Fixed

bowenliang123:
The `The Kyuubi instance` explaining `kyuubiInstance` here may be not that clear, this description doesn't show a specific form or entity concept to expect. Shall we should describe it in FE's `connectionUrl ` just as in #3876.

pan3793 (approved):

turboFei:
> The The Kyuubi instance explaining kyuubiInstance here may be not that clear, this description doesn't show a specific form or entity concept to expect. Shall we should describe it in FE's connectionUrl  just as in https://github.com/apache/kyuubi/pull/3876.

It is easy to known once they take action.

PR #4410, opened 2023-02-24
Improve `kyuubi-defaults.conf.template`

pan3793:

This PR changes the `kyuubi-defaults.conf.template` because

- remove deprecated conf `kyuubi.frontend.bind.port`
- add some properties which were frequently asked by users

image

### DIFF
--- a/conf/kyuubi-defaults.conf.template
+++ b/conf/kyuubi-defaults.conf.template
@@ -17,10 +17,19 @@
 
 ## Kyuubi Configurations
 
-#
-# kyuubi.authentication           NONE
-# kyuubi.frontend.bind.host       localhost
-# kyuubi.frontend.bind.port       10009
-#
-
 # Details in https://kyuubi.readthedocs.io/en/master/deployment/settings.html
+
+#
+# kyuubi.authentication                    NONE
+#
+# kyuubi.frontend.bind.host                <undefined>
+# kyuubi.frontend.protocols                THRIFT_BINARY
+# kyuubi.frontend.thrift.binary.bind.port  10009
+#
+# kyuubi.engine.type                       SPARK_SQL
+# kyuubi.engine.share.level                USER
+# kyuubi.session.engine.initialize.timeout PT3M
+#
+# kyuubi.ha.addresses                      <undefined>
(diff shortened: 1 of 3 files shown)


turboFei (approved):

cxzl25 (approved):

conf/kyuubi-defaults.conf.template
```diff
@@ -17,10 +17,19 @@
...
 
+# Details in https://kyuubi.readthedocs.io/en/master/deployment/settings.html
+
 #
-# kyuubi.authentication           NONE
-# kyuubi.frontend.bind.host       localhost
-# kyuubi.frontend.bind.port       10009
+# kyuubi.authentication                    NONE
+#
+# kyuubi.frontend.bind.host                <undefined>
```
yaooqinn, line 25: why use `<undefined>`?

pan3793: the default value is undefined, and Kyuubi will bind the first non-localhost network card.

yaooqinn:
> all properties should be the default value to make it intuitive for users

I do not get it

pan3793:
```
kyuubi.frontend.bind.host localhost
```
This is an example, the user may think Kyuubi binds `localhost` in default when looking `kyuubi-defaults.conf.template`, but actually not.

docs/deployment/settings.md
```diff
@@ -122,9 +122,18 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
 ## Kyuubi Configurations
 
 #
-# kyuubi.authentication           NONE
```
yaooqinn, line 125: We may give a much more practical example.

The current example is from the template directly, where all examples for individual configs are commented out, while we shall give meaningful ones on the contrary.

Or, for legible, we could delete this entire block here. For example, `You can configure the Kyuubi properties in $KYUUBI_HOME/conf/kyuubi-defaults.conf, see $KYUUBI_HOME/conf/kyuubi-defaults.conf.template as an example,. `

yaooqinn (approved):
LGTM

pan3793:
Thanks, merged to master/1.7

PR #4415, opened 2023-02-25
Align the server/engine ExecuteStatement operation handle for Spark engine

turboFei:

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -40,7 +40,7 @@ class ExecuteStatement(
     override val shouldRunAsync: Boolean,
     queryTimeout: Long,
     incrementalCollect: Boolean,
-    override val handle: OperationHandle)
+    override protected val handle: OperationHandle)
   extends SparkOperation(session) with Logging {
 
   private val operationLog: OperationLog = OperationLog.createOperationLog(session, getHandle)
@@ -177,7 +177,7 @@ class ArrowBasedExecuteStatement(
     override val shouldRunAsync: Boolean,
     queryTimeout: Long,
     incrementalCollect: Boolean,
-    handle: OperationHandle)
+    override protected val handle: OperationHandle)
   extends ExecuteStatement(
     session,
     statement,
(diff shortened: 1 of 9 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -123,7 +123,7 @@ abstract class SparkOperation(session: Session)
 
   protected def eventEnabled: Boolean = true
 
-  if (eventEnabled) EventBus.post(SparkOperationEvent(this))
+  if (handle != null && eventEnabled) EventBus.post(SparkOperationEvent(this))
```
pan3793, line 126: in which case that `handle == null` is true?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -39,13 +41,18 @@ class ExecuteStatement(
...
+    opHandleOption: Option[OperationHandle])
   extends SparkOperation(session) with Logging {
 
+  override protected val handle: OperationHandle = opHandleOption.getOrElse(OperationHandle())
+
   private val operationLog: OperationLog = OperationLog.createOperationLog(session, getHandle)
   override def getOperationLog: Option[OperationLog] = Option(operationLog)
   override protected def supportProgress: Boolean = true
 
+  EventBus.post(SparkOperationEvent(this))
```
pan3793, line 54: it's quite tricky behavior, we'd better refactor this to avoid confusing developers, if not easy, add comments to explain what happened as least.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
turboFei: For execute statement， it override the handle.

turboFei: Otherwise，npe thrown.

turboFei: fixed

pan3793 (approved):

turboFei:
thanks, merged to master

PR #4418, opened 2023-02-26
Allow disable metadata operation async retry and fail fast on unrecoverable DB error

pan3793:

The key changes are

1. allow disabling metadata operation retry
2. always fail fast on "duplicated key on unique index" error

Currently, when metadata operations failed, we always do async retry, to tolerate long-time metadata store outages w/o blocking the submission request, but it can not guarantee metadata consistency eventually, e.g. when inserting data violates the unique key restriction, it will never succeed, and block any following update request for the batch job, in such cases, the client gets succeed response but the metadata can not be updated correctly.

We should distinguish between recoverable and unrecoverable errors, for unrecoverable errors, we should fail fast, but the fact is it's hard to enumerate all recoverable nor unrecoverable errors, in this PR, we just enumerate the "duplicated key" as unrecoverable errors, and provide a switch to disable async retry so that the error can propagate to client correctly.

Some configurations are renamed w/ the `async.` prefix(the original key still takes effect) because we may introduce the sync retry logic in the future.


docs/deployment/settings.md
```diff
@@ -403,9 +403,11 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
+| kyuubi.metadata.request.retry.async.enabled     | false                                                    | Whether to retry in async when metadata request failed. If false, when metadata request failed, block the request thread and retry synchronously until success or reach the threshold; if true, return immediately even the metadata request failed, and schedule it in background until success. In async mode, we can not guarantee metadata consistency eventually, but it's still valuable since we can tolerate long-time metadata store outages w/o blocking the submission request.                                              | boolean  | 1.7.0 |
```
turboFei, line 406: looks like behavior change

docs/deployment/settings.md
```diff
@@ -403,9 +403,11 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
+| kyuubi.metadata.request.retry.maxAttempts       | 3                                                        | The max attempts for metadata request retry.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            | int      | 1.7.0 |
```
turboFei, line 410: behavior change for async mode.

turboFei: is it possible to lose metadata for async mode if the metastore does not recover in-time.

pan3793: case 1: 
- t1: metadata store outage
- t2: client opens batch, kyuubi inserts metadata failed, but returns a successful response
- t3: server done
- t4: metadata store recovered.

case 2: kyuubi inserts metadata record but encounters "duplicated unique key" error, which means the insert will never succeed, it blocks all following update operations of this batch job.

docs/deployment/settings.md
pan3793: mentioned in migration guide

ulysses-you:
retry synchronously may introduce state consistent issue. say, the first action udates state to running and second action (different thread) upadates state to cancel, and the first action fails but retry successfully.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
```diff
@@ -37,44 +37,55 @@ class MetadataManager extends AbstractService("MetadataManager") {
...
     new ConcurrentHashMap[String, AtomicInteger]()
 
-  private val requestsRetryTrigger =
-    ThreadUtils.newDaemonSingleThreadScheduledExecutor("metadata-requests-retry-trigger")
+  private lazy val requestsRetryInterval =
+    conf.get(KyuubiConf.METADATA_REQUEST_RETRY_INTERVAL)
 
-  private var requestsRetryExecutor: ThreadPoolExecutor = _
+  private lazy val requestsAsyncRetryEnabled =
+    conf.get(KyuubiConf.METADATA_REQUEST_ASYNC_RETRY_ENABLED)
```
ulysses-you, line 51: so this is the key change that introduce a config to control if we should retry ?

ulysses-you:
can you update the outdate pr description ?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
pan3793: yea

pan3793:
> can you update the outdate pr description ?

updated

turboFei (approved):

pan3793:
Thanks, merged to master/1.7

PR #4419, opened 2023-02-26
Implement simple EngineSecuritySecretProvider

pan3793:

This PR implements a simple `EngineSecuritySecretProvider` beside the existing zookeeper implementation, which simplifies the user threshold to use RESTful API w/ HA mode, and this PR also allows the user set `kyuubi.engine.security.secret.provider` using short name 'simple' or 'zookeeper' as well as the full class name.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -2068,8 +2068,23 @@ object KyuubiConf {
         "subclass of `EngineSecuritySecretProvider`.")
       .version("1.5.0")
       .stringConf
-      .createWithDefault(
-        "org.apache.kyuubi.service.authentication.ZooKeeperEngineSecuritySecretProviderImpl")
+      .transform {
+        case "simple" =>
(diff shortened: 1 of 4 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2068,8 +2068,23 @@ object KyuubiConf {
...
-      .createWithDefault(
-        "org.apache.kyuubi.service.authentication.ZooKeeperEngineSecuritySecretProviderImpl")
+      .transform {
+        case "simple" =>
+          "org.apache.kyuubi.service.authentication.SimpleEngineSecuritySecretProviderImpl"
+        case "zookeeper" =>
+          "org.apache.kyuubi.service.authentication.ZooKeeperEngineSecuritySecretProviderImpl"
+        case other => other
+      }
+      .createWithDefault("simple")
```
pan3793, line 2078:
```suggestion
      .createWithDefault("zookeeper")
```

turboFei (approved):
LGTM, but simple provider is not recommended

pan3793:
It's a tradeoff between convenience and security, let the user make decisions by themselves.

pan3793:
Thanks, merged to master/1.7

PR #4427, opened 2023-02-27
Simplify assertions in Authz tests in declarative style

bowenliang123:

Simplify assertions in Authz tests, by 
- add `doAs(user) { ... }`
- add `runSqlAs` for wrapping SQL running and asserting a successful `Try`
- add `runSqlAsWithAccessException` to intercept `AccessControlException` and assert error message

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/SparkSessionProvider.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/SparkSessionProvider.scala
@@ -22,6 +22,7 @@ import java.security.PrivilegedExceptionAction
 
 import scala.util.Try
 
+import org.apache.commons.lang3.StringUtils
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{DataFrame, SparkSession, SparkSessionExtensions}
@@ -102,15 +103,15 @@ trait SparkSessionProvider {
   }
 
   protected def runSqlAsWithAccessException(user: String = "someone")(
-      sqlStr: String,
+      sqlText: String,
       isCollect: Boolean = false,
       isExplain: Boolean = false,
-      contains: String = null): Unit = {
-    withClue(sqlStr) {
+      contains: String = ""): Unit = {
+    withClue(sqlText) {
       val e = intercept[AccessControlException](
         doAs(
           user, {
-            val df = sql(sqlStr)
+            val df = sql(sqlText)
             if (isExplain) {
               df.explain()
             }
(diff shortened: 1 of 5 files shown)


extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/SparkSessionProvider.scala
```diff
@@ -71,23 +75,78 @@ trait SparkSessionProvider {
...
 
-  protected def doAs[T](user: String, f: => T): T = {
+  protected def doAs[T](user: String)(f: => T): T = {
     UserGroupInformation.createRemoteUser(user).doAs[T](
       new PrivilegedExceptionAction[T] {
         override def run(): T = f
       })
   }
+
+  protected def runSqlAs(user: String)(
```
pan3793, line 85: I think it's kind of over-engineering, which increases the onboarding costs for new developer

yaooqinn: +1

bowenliang123: It's not forcing user/developer to use `runSqlAs`. `runSqlAs` and `runSqlAsWithAccessException` are together providing a concentrative and declarative way in most general scenarios to decide running what SQL in which user with asserting what messages. `doAs` is still good to be used with them,  and remained as you see.

yaooqinn: I have added a `checkAnswer` method and it looks better than this one. Meanwhile, there would be great if we could have a similar one for expected failures, for example.

```scala
def checkError[T <: Throwable : ClassTag](
  user: String,
  query: String,
  expectedErrMsg: String): Unit = {
...
  }
```

bowenliang123: It's not sure `checkAnswer` is better in some way.

As for this PR, the key points for changes in signatures of `doAs` and `runAs` includes:
- extract user as first param group for `As`, to focus in the user idendity and also to eliminate unhelpfule line break every time
- handy to config the decision points in assertion and and the execution modes
- wrapping repeated exectution template

After all, these methods are used in test cases only. It's good enough to provide well polished signatures for simplication.

yaooqinn: Please use simple functions for tests, especially in test traits/helpers, etc.

- Complex helpers make the debug experience worse.
- Playing tricks with if-else, etc., make the function itself hard to understand. while you wrote once, thousands of reads may follow.
- Currying is not so that welcome in tests
- etc

yaooqinn:
Discussions for PR are welcome, people are free to express their agreements/disagreements.

BTW, better to comment with a reason while closing a PR

bowenliang123:
Closing this PR as it's in the immature status of early stage. Not too many common benefits.

PR #4437, opened 2023-03-02
[Authz] Fix dependencies conflict by replacing `jersey-bundle` with `jersey-client`

bowenliang123:

- `jerysey-bundle` 1.x , which is depended by `ranger-plugins-common`, contains `javax.ws.rs.*` (like `javax.ws.rs.core.*`) packages that conflicted to `jarkarta.ws.rs-api-2.1.6.jar` in Spark 3.0+ jars
- exclude `jerysey-bundle` 1.x and include `jersey-client` 1.19.4 to align with `ranger-plugins-common`

Diff in Authz dependency jar list:
```diff
gethostname4j-1.0.0.jar
jackson-jaxrs-1.9.13.jar
- jersey-bundle-1.19.3.jar
+ jersey-client-1.19.4.jar
+ jersey-core-1.19.4.jar 
jetty-client-9.4.50.v20221201.jar
jetty-http-9.4.50.v20221201.jar
jetty-io-9.4.50.v20221201.jar
jetty-util-9.4.50.v20221201.jar
jna-5.7.0.jar
jna-platform-5.7.0.jar
ranger-plugins-audit-2.3.0.jar
ranger-plugins-common-2.3.0.jar
ranger-plugins-cred-2.3.0.jar
```

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz/pom.xml
@@ -32,7 +32,9 @@
     <url>https://kyuubi.apache.org/</url>
 
     <properties>
+        <!-- the following components' version may need to tune to align w/ the ranger.version-->
         <gethostname4j.version>1.0.0</gethostname4j.version>
+        <jersey.client.version>1.19.4</jersey.client.version>
         <jna.version>5.7.0</jna.version>
     </properties>
 
@@ -42,6 +44,10 @@
             <artifactId>ranger-plugins-common</artifactId>
             <version>${ranger.version}</version>
             <exclusions>
+                <exclusion>
+                    <groupId>com.sun.jersey</groupId>
+                    <artifactId>jersey-bundle</artifactId>
+                </exclusion>
                 <exclusion>
                     <groupId>org.apache.ranger</groupId>
                     <artifactId>ranger-plugin-classloader</artifactId>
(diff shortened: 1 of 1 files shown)


bowenliang123:
cc @yaooqinn @pan3793 @zhouyifan279

pan3793:
Does `jersey-client` has the transitive deps? Can you list the final jars that users should be copied?

bowenliang123:
> Does `jersey-client` have transitive deps? Can you list the final jars that users should copy?

OK, diff in jar list added to the description. See above.

yaooqinn (approved):

pan3793:
The conflict classes `javax.ws.rs.*` you mentioned in description still existing in `jsr311-api-1.1.1.jar`

pan3793:
Maybe we should try to exclude `jsr311-api-1.1.1.jar` from `jersey-client`?

bowenliang123:
> Maybe we should try to exclude `jsr311-api-1.1.1.jar` from `jersey-client`?

Agree. Let's do this.

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -33,6 +33,7 @@
 
     <properties>
         <gethostname4j.version>1.0.0</gethostname4j.version>
+        <jersey.client.version>1.19.4</jersey.client.version>
         <jna.version>5.7.0</jna.version>
     </properties>
```
pan3793, line 38: let's move the `ranger.version` from the root pom to this place, and add comments to tips users may need to tune them align w/ `ranger.version`

bowenliang123: Addressed.

pan3793:
Thanks, merged to master/1.7

ASiegeLion:
The spark web ui  can be displayed normally
91d298f3600add9ec1d0ef00ef03f1f

ASiegeLion:
The spark web ui  can be displayed normally
image

bowenliang123:
Thank you for the validation. @ASiegeLion

Issue #4171, opened 2023-01-15
[Improvement]  GetTables is too slow


parisni:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

The GetTables operation (triggered for eg by dbeaver when listing the tables) is slow when the number of tables is not small. The getTables process takes 1 minute for 200 tables/views, on various catalog (jdbc, hive...).

### How should we improve?

This operation might be batched

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

pan3793:
Do you have a test on Hive/Trino? What's their performance?

parisni:
I didn't test trino, only hive and jdbc. Both are surprisingly slow as compared to catalog listing (show tables * desc table )

PR #4444, opened 2023-03-04
[KYUUBI #4171] Support skip retrieving table's properties to speed up GetTables operation

liaoyt:

`GetTables` operation is too slow because it queries table details info one by one, but then only a table comment is used to construct a result row, which i think could be optional.
This PR add an optional config which can control this operation. By default, `GetTables` operation queries all message. Otherwise, `GetTables` operation just return table identifiers.


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2713,4 +2713,11 @@ object KyuubiConf {
       .version("1.7.0")
       .timeConf
       .createWithDefault(Duration.ofSeconds(60).toMillis)
+
+  val ENGINE_SPARK_LIST_TABLES: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.spark.list.tables")
```
pan3793, line 2718: I suggest using `kyuubi.operation.getTables.ignoreTableProperties` here, because

1. technically, this optimization can be applied to all engines.
2. back to the purpose of this PR, it aims to avoid the call of retrieving each table's properties to speed up the GetTables operation, for Spark, only "comment" is taken here, but it's better to make the configuration more generic, right?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v3_0.scala
```diff
@@ -166,10 +167,12 @@ class CatalogShim_v3_0 extends CatalogShim_v2_4 {
         val identifiers = namespaces.flatMap { ns =>
           tc.listTables(ns).filter(i => tp.matcher(quoteIfNeeded(i.name())).matches())
         }
+        val listTablesOnly = spark.conf.getOption(KyuubiConf.ENGINE_SPARK_LIST_TABLES.key)
+          .map(_.toBoolean).getOrElse(KyuubiConf.ENGINE_SPARK_LIST_TABLES.defaultVal.get)
```
pan3793, line 171: it does not consider the kyuubi session conf, I suggest evaluating the flag in the `GetTables`, you can refer to `SparkOperation#operationSparkListenerEnabled`

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2713,4 +2713,11 @@ object KyuubiConf {
...
+      .doc("Only query table identifiers when set to true. Work on Spark 3.x only.")
```
pan3793, line 2719: please update the description to match the configuration name, and we don't need to mention Spark 3.x since Kyuubi only support Spark 3.1 and above

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
liaoyt: Good suggestion. This can be a generic configuration and `kyuubi.operation.getTables.ignoreTableProperties` looks more reasonable, but the relational implement maybe different in other engine(Flink, Trino, eg), which can open another issue to follow up.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/shim/CatalogShim_v3_0.scala
liaoyt: Thanks, I'll have a look.

pan3793 (approved):

pan3793:
Thanks, merged to master

Issue #4452, opened 2023-03-06
[Improvement] Strip the redundant leading and tailing slashes of getZooKeeperNamespace.


merrily01:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

As described in the discussion: <https://github.com/apache/kyuubi/discussions/4436>

When the zk namespace contains leading and trailing slashes, the exception will cause the client connection to fail, and the exception prompt information is not clear.

### How should we improve?

Strip the redundant leading and tailing slashs of getZooKeeperNamespace and add exception stack information at the same time.

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4454, opened 2023-03-06
[KYUUBI #4452] Strip the redundant leading and tailing slash of getZooKeeperNamespace.

merrily01:

As described in the discussion:  <https://github.com/apache/kyuubi/discussions/4436>

When the zk namespace contains leading and trailing slashes, the exception will cause the client connection to fail, and the exception prompt information is not clear.

So we strip the redundant leading and tailing slashes of getZooKeeperNamespace and add exception stack information at the same time.

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
@@ -17,6 +17,7 @@
 
 package org.apache.kyuubi.jdbc.hive;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
@@ -32,7 +33,8 @@ class ZooKeeperHiveClientHelper {
   // Pattern for key1=value1;key2=value2
   private static final Pattern kvPattern = Pattern.compile("([^=;]*)=([^;]*);?");
 
-  public static String getZooKeeperNamespace(JdbcConnectionParams connParams) {
+  @VisibleForTesting
+  protected static String getZooKeeperNamespace(JdbcConnectionParams connParams) {
     String zooKeeperNamespace =
         connParams.getSessionVars().get(JdbcConnectionParams.ZOOKEEPER_NAMESPACE);
     if ((zooKeeperNamespace == null) || (zooKeeperNamespace.isEmpty())) {
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
@@ -32,12 +32,13 @@ class ZooKeeperHiveClientHelper {
   // Pattern for key1=value1;key2=value2
   private static final Pattern kvPattern = Pattern.compile("([^=;]*)=([^;]*);?");
 
-  private static String getZooKeeperNamespace(JdbcConnectionParams connParams) {
+  public static String getZooKeeperNamespace(JdbcConnectionParams connParams) {
     String zooKeeperNamespace =
         connParams.getSessionVars().get(JdbcConnectionParams.ZOOKEEPER_NAMESPACE);
     if ((zooKeeperNamespace == null) || (zooKeeperNamespace.isEmpty())) {
       zooKeeperNamespace = JdbcConnectionParams.ZOOKEEPER_DEFAULT_NAMESPACE;
     }
+    zooKeeperNamespace = zooKeeperNamespace.replaceAll("^/+", "").replaceAll("/+$", "");
     return zooKeeperNamespace;
   }
 
--- a/kyuubi-hive-jdbc/src/test/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelperTest.java
+++ b/kyuubi-hive-jdbc/src/test/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelperTest.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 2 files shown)


merrily01:
cc @pan3793 PTAL

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
```diff
@@ -112,7 +115,7 @@ static void configureConnParams(JdbcConnectionParams connParams)
       updateParamsWithZKServerNode(connParams, zooKeeperClient, serverNode);
     } catch (Exception e) {
       throw new ZooKeeperHiveClientException(
-          "Unable to read HiveServer2 configs from ZooKeeper", e);
+          "Unable to read HiveServer2 configs from ZooKeeper: " + e.getMessage());
```
pan3793, line 118: why change it? we should propagate the inner exception otherwise we will lost the stacktrace

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
```diff
@@ -38,6 +38,9 @@ private static String getZooKeeperNamespace(JdbcConnectionParams connParams) {
     if ((zooKeeperNamespace == null) || (zooKeeperNamespace.isEmpty())) {
       zooKeeperNamespace = JdbcConnectionParams.ZOOKEEPER_DEFAULT_NAMESPACE;
     }
+    if (zooKeeperNamespace.startsWith("/") || zooKeeperNamespace.endsWith("/")) {
+      zooKeeperNamespace.replaceAll("^/+", "").replaceAll("/+$", "");
```
pan3793, line 42: the call is cheap, we can remove if to simplify the code, you should re-assign the value to variable `zooKeeperNamespace`, and please add unit tests which should contains both positive and negative cases to verify your change

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
```diff
@@ -111,6 +112,7 @@ static void configureConnParams(JdbcConnectionParams connParams)
       String serverNode = serverHosts.get(new Random().nextInt(serverHosts.size()));
       updateParamsWithZKServerNode(connParams, zooKeeperClient, serverNode);
     } catch (Exception e) {
+      e.printStackTrace();
```
pan3793, line 115: we don't need it, the existing code is sufficient to propagate underlying exception stacktrace to user.

kyuubi-hive-jdbc/src/test/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelperTest.java
```diff
@@ -0,0 +1,69 @@
...
+    this.uri = uri;
+  }
+
+  @Test
+  public void testGetZooKeeperNamespace()
+      throws JdbcUriParseException, NoSuchMethodException, InvocationTargetException,
+          IllegalAccessException {
+    JdbcConnectionParams jdbcConnectionParams = extractURLComponents(uri, new Properties());
+    ZooKeeperHiveClientHelper zooKeeperHiveClientHelper = new ZooKeeperHiveClientHelper();
+    Method getZooKeeperNamespace =
```
pan3793, line 60: we can simply open the method to public for testing.

merrily01: done

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
merrily01: done

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
merrily01: Thanks for your suggestion.

On the basis of the current code, I tried several methods and failed to output the exception stacktrace.

I refer to the following: https://github.com/apache/kyuubi/blob/master/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiBaseResultSet.java#L300

So what is the specific way you say?

pan3793: Actually we should remove the `e.printStackTrace();` it's totally a bad practice.

> ...  I tried several methods and failed to output the exception stacktrace.

The caller is responsible to catch and handle the exception. i.e. the caller may want to catch and print the stacktrace using log system. 
```
Logger log = LoggerFactory.getLogger(xxx);
try {
  ...
} catch (Exception cause) {
  log.error(cause.getMessage(), cause)
}
```

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
```diff
@@ -32,12 +32,13 @@ class ZooKeeperHiveClientHelper {
   // Pattern for key1=value1;key2=value2
   private static final Pattern kvPattern = Pattern.compile("([^=;]*)=([^;]*);?");
 
-  private static String getZooKeeperNamespace(JdbcConnectionParams connParams) {
+  public static String getZooKeeperNamespace(JdbcConnectionParams connParams) {
```
zwangsheng, line 35: Can we use smaller range modifier and add `@VisibleForTesting`?

pan3793 (approved):

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
merrily01: done

zwangsheng (approved):

pan3793:
Thanks, merged to master/1.7

merrily01:
Thanks for your help.  cc/ @pan3793

pan3793:
@merrily01 thanks for your patience and contribution :)

PR #4464, opened 2023-03-06
Simplify and improve log for JDBCMetadataStore

pan3793:

1. enhance the log to include parameters for the prepared statement
2. remove redundant spaces from SQL
3. simplify code by using `JdbcUtils`


kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -479,22 +485,6 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
     }
   }
 
-  private def withConnection[T](autoCommit: Boolean = true)(f: Connection => T): T = {
-    var connection: Connection = null
-    try {
-      connection = hikariDataSource.getConnection
-      connection.setAutoCommit(autoCommit)
```
pan3793, line 486: `autoCommit` always is `true` in whole codebase

PR #4465, opened 2023-03-06
[KYUUBI #4219]invert the pattern match branch order

edddddy:

It does work after  several tests, that inverting the pattern match branch order in `CatalogShim_v3_0#getColumnsByCatalog`, which ensures BI tools can list columns on Delta Lake tables in all schemas.


yaooqinn:
can you check the failed tests?

Issue #4439, opened 2023-03-02
[Subtask] add get/close operation method for AdminResouce


lightning-L:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues) and found no similar issues.

### Describe the subtask

get/close operation for AdminResouce

### Parent issue

https://github.com/apache/kyuubi/issues/4375

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4468, opened 2023-03-07
[KYUUBI #4439][FOLLOWUP] Add dto class for operation data

turboFei:

- add dto for operation data, which is programming friendly
- show exception for session data

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -387,4 +387,19 @@ object Utils extends Logging {
     Option(Thread.currentThread().getContextClassLoader).getOrElse(getKyuubiClassLoader)
 
   def isOnK8s: Boolean = Files.exists(Paths.get("/var/run/secrets/kubernetes.io"))
+
+  /**
+   * Return a nice string representation of the exception. It will call "printStackTrace" to
+   * recursively generate the stack trace including the exception and its causes.
+   */
+  def exceptionString(e: Throwable): String = {
+    if (e == null) {
+      ""
+    } else {
+      // Use e.printStackTrace here because e.getStackTrace doesn't include the cause
+      val stringWriter = new StringWriter()
+      e.printStackTrace(new PrintWriter(stringWriter))
+      stringWriter.toString
+    }
+  }
 }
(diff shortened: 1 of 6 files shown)


turboFei:
also cc @lightning-L

turboFei:
BTW, we can add some filter conditions for the list sessions/operations api.

Such as filter the sessions/operations with user list.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/OperationData.java
```diff
@@ -0,0 +1,158 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.client.api.v1.dto;
+
+import java.util.Objects;
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+public class OperationData {
```
pan3793, line 24: should we add a conf field since that `ExecuteStatement` supports `confOverlay`?

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -387,4 +387,19 @@ object Utils extends Logging {
     Option(Thread.currentThread().getContextClassLoader).getOrElse(getKyuubiClassLoader)
 
   def isOnK8s: Boolean = Files.exists(Paths.get("/var/run/secrets/kubernetes.io"))
+
+  /**
+   * Return a nice string representation of the exception. It will call "printStackTrace" to
+   * recursively generate the stack trace including the exception and its causes.
+   */
+  def exceptionString(e: Throwable): String = {
```
pan3793, line 395: prettyPrint

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/OperationData.java
turboFei: difficult to add

pan3793 (approved):

pan3793:
Is https://kyuubi.readthedocs.io/en/master/client/rest/rest_api.html outdated?

turboFei:
thanks, merged to master

Issue #4467, opened 2023-03-07
[Bug] KubernetesApplicationOperation may return both pid and spark app Id


zwangsheng:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When using restful api submit batch job on kubenretes + cluster mode, kyuubi server will get batch info before kubernetes created spark driver pods.

So, kyuubi server will think this's may in kubernetes + client mode, using `jpsApplicationOperation` get spark-submit process.

It may cause in restful client log with both pid and appId.

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk
23/03/06 20:48:05 INFO KyuubiBatchJobSubmit: batchId: 6a6b76a6-2d8e-412b-b6c9-aa9862a4bb04, appId: 287
23/03/06 20:48:10 INFO KyuubiBatchJobSubmit: batchId: 6a6b76a6-2d8e-412b-b6c9-aa9862a4bb04, appId: 287, appStatus: RUNNING
23/03/06 20:48:15 INFO KyuubiBatchJobSubmit: batchId: 6a6b76a6-2d8e-412b-b6c9-aa9862a4bb04, appId: spark-cbe712c1573c42b1ac6ce322c612f9f9, appStatus: PENDING
23/03/06 20:48:20 INFO KyuubiBatchJobSubmit: batchId: 6a6b76a6-2d8e-412b-b6c9-aa9862a4bb04, appId: 287, appStatus: RUNNING
23/03/06 20:48:25 INFO KyuubiBatchJobSubmit: batchId: 6a6b76a6-2d8e-412b-b6c9-aa9862a4bb04, appId: spark-cbe712c1573c42b1ac6ce322c612f9f9, appStatus: PENDING
```

### Additional context

- [ ] Remove JpsApplicationOperation from KubernetesApplicationOperation
- [ ] Add `deployMode` arg in getAppInfo and killApplication function

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #4469, opened 2023-03-07
[KYUUBI #4467][K8S][BATCH] Tolerate Driver Pod ephemerally invisible after submitting

zwangsheng:

The following discussion assumes using Spark cluster mode w/ `waitCompletion=false`.

In Spark on Yarn, the application is visible immediately after `spark-submit` is returned, but things are different in Spark on K8s, Driver Pod is ephemerally invisible after submitting, so NOT_FOUND is returned instead of UNKNOWN or PENDING.

To tolerate the above case, `kyuubi.engine.submit.timeout` is introduced, ApplicationManager will report UNKNOWN instead of NOT_FOUND during the Driver Pod scheduling period.

More detail in #4467 
1. Remove `KubernetesApplicationOperation`'s `JpsApplicationOperation` for handle Client Deploy Mode(`YarnApplicationOperation` doesn't handle this either)
2. Add engine submit timeout for `KubernetesApplicationOperation` to return Unknown status when not found driver pod in time range.
3. GetApplicationInfo with it's submit time


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2531,6 +2531,13 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val ENGINE_SUBMIT_TIMEOUT: ConfigEntry[Long] =
+    buildConf("kyuubi.engine.submit.timeout")
+      .doc("The timeout for submitting engine")
+      .version("1.8.0")
```
pan3793, line 2537: it's kind of bug fix, should be 1.7.1

ulysses-you:
do you mean after this pr, we can not support get app info with spark on k8s with client mode ? can we check the deploy mode instead ? if it is cluster mode, get app info from k8s. If it is client client mode, get app info from jps.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -101,8 +102,23 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
...
-          // client mode
-          jpsOperation.getApplicationInfoByTag(tag)
+          // Kyuubi should wait second if pod is not be created
+          submitTime match {
+            case Some(time) =>
+              val elapsedTime = System.currentTimeMillis() - time
+              if (elapsedTime > submitTimeout) {
+                error(s"Can't find target driver pod with tag: $tag, " +
+                  s"has elapsed time: ${elapsedTime}ms (limit timeout: ${submitTimeout}ms), " +
+                  s"to submit time: ${submitTimeout}ms")
```
ulysses-you, line 112: this line seems unnecessary ?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
```diff
@@ -58,7 +58,7 @@ trait ApplicationOperation {
    * @param tag the unique application tag for engine instance.
    * @return [[ApplicationInfo]]
    */
-  def getApplicationInfoByTag(tag: String): ApplicationInfo
+  def getApplicationInfoByTag(tag: String, submitTime: Option[Long] = None): ApplicationInfo
```
ulysses-you, line 61: can we add a new method with the default implementation ?
```scala
def getApplicationInfoByTag(tag: String, submitTime: Long) = getApplicationInfoByTag(tag)
```

zwangsheng: I don't quite understand what you mean by using
```scala
def getApplicationInfoByTag(tag: String): ApplicationInfo = getApplicationInfoByTag(tag, None)
```
to support default implementation?

ulysses-you: you can have two methods:
```scala
def getApplicationInfoByTag(tag: String, submitTime: Long): ApplicationInfo =
  getApplicationInfoByTag(tag)

def getApplicationInfoByTag(tag: String): ApplicationInfo
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -101,8 +102,22 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
...
+            case Some(time) =>
+              val elapsedTime = System.currentTimeMillis() - time
+              if (elapsedTime > submitTimeout) {
+                error(s"Can't find target driver pod with tag: $tag, " +
+                  s"has elapsed time: ${elapsedTime}ms (limit timeout: ${submitTimeout}ms)")
+                ApplicationInfo(id = null, name = null, ApplicationState.NOT_FOUND)
+              } else {
+                warn("Wait for driver pod to be created, " +
+                  s"elapsed time: ${elapsedTime}ms, return UNKNOWN status")
+                ApplicationInfo(id = null, name = null, ApplicationState.UNKNOWN)
```
pan3793, line 116: We should use PENDING here, since UNKNOWN is kind of Terminated status, see `ApplicationState.isTerminated`

pan3793, line 116: better to use PENDING here, to un-surprise users

zwangsheng: Given that we are already using PENDING to indicate the status of Spark Driver Pod created on Kubernetes, does it give the user the illusion that the driver pod has been submitted successfully?

integration-tests/kyuubi-kubernetes-it/src/test/scala/org/apache/kyuubi/kubernetes/test/spark/SparkOnKubernetesTestsSuite.scala
```diff
@@ -147,23 +147,13 @@ class KyuubiOperationKubernetesClusterClientModeSuite
       batchRequest.getConf.asScala.toMap,
       batchRequest)
 
-    eventually(timeout(3.minutes), interval(50.milliseconds)) {
+    // [KYUUBI #4467] KubernetesApplicationOperator doesn't support client mode
+    eventually(timeout(1.minutes), interval(50.milliseconds)) {
```
pan3793, line 151: As we discussed before, let's just ignore this test case, and restore it after fixing K8s client mode.

zwangsheng: Sure

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
pan3793: compared two approaches, this one is clean and does not touch many things

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -49,6 +48,7 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
   }
 
   override def isSupported(clusterManager: Option[String]): Boolean = {
+    // TODO add deploy mode to check whether is supported
```
pan3793, line 51: IMO we should deny client mode, although this is a kind of temporary regression

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -101,8 +102,22 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
...
         } else {
-          // client mode
-          jpsOperation.getApplicationInfoByTag(tag)
+          // Kyuubi should wait second if pod is not be created
+          submitTime match {
+            case Some(time) =>
+              val elapsedTime = System.currentTimeMillis() - time
+              if (elapsedTime > submitTimeout) {
+                error(s"Can't find target driver pod with tag: $tag, " +
+                  s"has elapsed time: ${elapsedTime}ms (limit timeout: ${submitTimeout}ms)")
```
pan3793, line 111:
```suggestion
                error(s"Can't find target driver pod by tag: $tag, " +
                  s"elapsed time: ${elapsedTime}ms exceeds ${submitTimeout}ms.")
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -101,8 +102,22 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
           debug(s"Successfully got application info by $tag: $info")
           info
         } else {
-          // client mode
-          jpsOperation.getApplicationInfoByTag(tag)
+          // Kyuubi should wait second if pod is not be created
+          submitTime match {
```
pan3793, line 106: for complex logical, we can just `return` in `if` branch, and save an `else` block.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
zwangsheng: Here just a TODO notice, put comments here for notify developers.
I'll make a follow up PR to let `JpsApplicationOperator` support client mode.

pan3793: A clear rejection is better than uncontrollable behavior

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
zwangsheng: Warn log should also be required by users

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2531,6 +2531,13 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val ENGINE_SUBMIT_TIMEOUT: ConfigEntry[Long] =
+    buildConf("kyuubi.engine.submit.timeout")
+      .doc("The timeout for submitting engine")
```
turboFei, line 2536: could you add the config scope in the description, seems it is only for k8s.

zwangsheng: Sure

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2541,6 +2541,13 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(true)
 
+  val ENGINE_SUBMIT_TIMEOUT: ConfigEntry[Long] =
+    buildConf("kyuubi.engine.submit.timeout")
+      .doc("The timeout for submitting engine to kubernetes cluster")
```
pan3793, line 2546: Period to tolerant Driver Pod ephemerally invisible after submitting. In some Resource Managers, e.g. K8s, the Driver Pod is not invisible immediately after `spark-submit` is returned.

zwangsheng: Modified.

pan3793:
Thanks, merged to master/1.7

PR #4479, opened 2023-03-08
Restore JDBC Kerberos authentication behavior for UGI.doAs

pan3793:

A typical use case of Hadoop UGI w/ Kyuubi Hive JDBC is
```
UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
ugi.doAs(() -> {
  Connection conn = DriverManager.getConnection(
        "jdbc:kyuubi://host:10009/default;principal=kyuubi@_HOST/ABC.ORG");
  ...
});
```

After https://github.com/apache/kyuubi/pull/3023, Kyuubi Hive JDBC implements the Kerberos authentication by using JDK directly instead of Hadoop `UserGroupInformation`, but it also introduce a breaking change for Hadoop users, including the above case. As workaround, user should add `kerberosAuthType=fromSubject` alongside w/ `principal=kyuubi@_HOST/ABC.ORG` to make it work.

This PR propose to restore the behavior before https://github.com/apache/kyuubi/pull/3023 by handling UGI.doAs explicitly.

And this PR makes the `clientPrincipal` `clientKeytab` as the highest priority, so in below cases, `clientPrincipal` `clientKeytab` take effects instead of UGI.

```
UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
ugi.doAs(() -> {
  Connection conn = DriverManager.getConnection(
        "jdbc:kyuubi://host:10009/default;principal=kyuubi@_HOST/ABC.ORG;" + 
        "clientPrincipal=tom@_HOST/ABC.ORG;clientKeytab=/path/xxx.keytab");
  ...
});
```

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -832,11 +832,11 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
 
   private boolean isFromSubjectAuthMode() {
     return isSaslAuthMode()
-            && hasSessionValue(AUTH_PRINCIPAL)
-            && !hasSessionValue(AUTH_KYUUBI_CLIENT_PRINCIPAL)
-            && !hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB)
-            && (AUTH_KERBEROS_AUTH_TYPE_FROM_SUBJECT.equalsIgnoreCase(
-            sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE))
+        && hasSessionValue(AUTH_PRINCIPAL)
+        && !hasSessionValue(AUTH_KYUUBI_CLIENT_PRINCIPAL)
+        && !hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB)
+        && (AUTH_KERBEROS_AUTH_TYPE_FROM_SUBJECT.equalsIgnoreCase(
+                sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE))
             || isHadoopUserGroupInformationDoAs());
   }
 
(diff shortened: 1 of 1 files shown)


cxzl25 (approved):

pan3793:
Have a offline talk w/ @gabrywu, this PR fixed the original issue but also brings a new one, considering the following case, the `clientPrincipal` and `clientKeytab` will be totally ignored and always respect UGI, personally I think it's an acceptable result, but would like to know other guys thoughts.

cc @cxzl25 @zhouyifan279 @turboFei 

```
UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
ugi.doAs(() -> {
  Connection conn = DriverManager.getConnection(
        "jdbc:kyuubi://host:10009/default;principal=kyuubi@_HOST/ABC.ORG;" + 
        "clientPrincipal=tom@_HOST/ABC.ORG;clientKeytab=/path/xxx.keytab");
  ...
});
```

turboFei (approved):

cxzl25 (approved):
LGTM

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -30,10 +30,7 @@
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
-import java.security.AccessControlContext;
-import java.security.AccessController;
-import java.security.KeyStore;
-import java.security.SecureRandom;
+import java.security.*;
```
merrily01, line 33: Why not introduce a specific reference, but modify it to * ?

pan3793: it's the result of `dev/reformat`, part of Google Java Format standard

merrily01: ok, thx

turboFei:
thanks, merged to master and 1.7

PR #4482, opened 2023-03-08
Support to disable batch resource upload function

turboFei:

Support to disable batch resource upload in server side for service stability. 


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1511,6 +1511,13 @@ object KyuubiConf {
       .timeConf
       .createWithDefault(Duration.ofSeconds(5).toMillis)
 
+  val BATCH_RESOURCE_UPLOAD_ENABLED: ConfigEntry[Boolean] =
+    buildConf("kyuubi.batch.resource.upload.enabled")
+      .internal
+      .doc("Whether to enable Kyuubi batch resource upload function.")
```
pan3793, line 1517: nit: let's add a version for it, even it's internal

pan3793 (approved):

bowenliang123 (approved):

cxzl25 (approved):

PR #4483, opened 2023-03-08
Enable REST frontend protocol by default

pan3793:

REST frontend protocol was introduced in Kyuubi 1.4.0 https://github.com/apache/kyuubi/pull/1349, more and more users are using REST in recent days, and to simplify the planned Web UI developments, I think we should enable the REST by default for master(1.8.0-SNAPSHOT).

I do not remove the "experiment" because we don't have strong confident to ensure the API stability in this time, I would like to defer it until the community has confident to mark it as stable.


bowenliang123 (approved):
LGTM. Great to see REST protocol enabled by default. Other features relying to REST like `kyuubi-admin` also benefit from this.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -174,14 +174,14 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
...
         isStarted.set(true)
-        info(s"$getName has started at ${server.getServerUri}")
         startBatchChecker()
         startInternal()
       } catch {
         case e: Exception => throw new KyuubiException(s"Cannot start $getName", e)
       }
     }
     super.start()
+    info(s"Exposing REST endpoint at: ${server.getServerUri}")
```
pan3793, line 184: Make output consistent
```
KyuubiTBinaryFrontendService: Service[KyuubiTBinaryFrontend] is started.
KyuubiTBinaryFrontendService: Starting and exposing JDBC connection at: jdbc:hive2://10.221.96.184:10009/
...
KyuubiRestFrontendService: Service[KyuubiRestFrontendService] is started.
KyuubiRestFrontendService: Exposing REST endpoint at: 10.221.96.184:10099
```

bowenliang123: Possible to output the protocol as well?

pan3793: OK, currently it does not support SSL, so simply added `http://`

turboFei (approved):

cxzl25 (approved):

pan3793:
Thanks all, merged to master

PR #4484, opened 2023-03-08
[K8S][HELM] Set specific Kyuubi version to the chart appVersion

dnskr:
The change is needed to use stable released version of Kyuubi instead of night/dev builds by default.

### DIFF
--- a/charts/kyuubi/Chart.yaml
+++ b/charts/kyuubi/Chart.yaml
@@ -20,7 +20,7 @@ name: kyuubi
 description: A Helm chart for Kyuubi server
 type: application
 version: 0.1.0
-appVersion: "master-snapshot"
+appVersion: 1.7.0
 home: https://kyuubi.apache.org
 icon: https://raw.githubusercontent.com/apache/kyuubi/master/docs/imgs/logo.png
 sources:


charts/kyuubi/Chart.yaml
```diff
@@ -20,7 +20,7 @@ name: kyuubi
 description: A Helm chart for Kyuubi server
 type: application
 version: 0.1.0
```
pan3793, line 22: question: is there any suggestion/guidance for this version?

pan3793 (approved):

charts/kyuubi/Chart.yaml
dnskr: Nothing in particular. Chart version needs to be changed on each chart release. By release I mean [creating chart archive](https://helm.sh/docs/helm/helm_package/#helm-package) and adding it to [chart repository](https://helm.sh/docs/topics/chart_repository/).
I would prefer to keep `0.1.0` for the initial dev release, which is much closer now because of the official Kyuubi docker image. Not sure how often the chart should be released. For example MinIO, Argo and many others release it quite frequently (for one or few changes), but Airflow releases helm chart only once for each minor (x.Y.z) application version.

pan3793: Educated, thanks @dnskr

pan3793:
Thanks, merged to master/1.7, and also cc @yanghua as you are investigating the kubevela related stuffs.

yanghua:
> Thanks, merged to master/1.7, and also cc @yanghua as you are investigating the kubevela related stuffs.

sorry for the late reply, thanks!

PR #4494, opened 2023-03-10
`bin/kyuubi` should use `exec` to run Kyuubi server

wangmiao1002:

The difference between w/ and w/o `exec` in shell script

https://www.baeldung.com/linux/exec-command-in-shell-script

> When using exec, however, the command following exec replaces the current shell. This means no subshell is created and the current process is replaced with this new command.

We must do this change to make `bin/kyuubi` suitable for Cloudera Manager.

### DIFF
--- a/bin/kyuubi
+++ b/bin/kyuubi
@@ -155,7 +155,7 @@ function start_kyuubi() {
 
 function run_kyuubi() {
   echo "Starting $CLASS"
-  nice -n "${KYUUBI_NICENESS:-0}" ${cmd}
+  exec nice -n "${KYUUBI_NICENESS:-0}" ${cmd}
 }
 
 function stop_kyuubi() {


pan3793:
@wangmiao1002 thanks for contributing, I modified your PR title from "Modify script run_kyuubi mode" to "bin/kyuubi should use exec to run Kyuubi server", the naming principle is try your best to make it short and precise, would you mind update the PR description to supply more details and background about this change? To un-surprise users who revisit this PR in the future.

pan3793:
Thanks, merged to master/1.7

pan3793:
@wangmiao1002 your git commit username and mail is `miaowang <miaowang@gaojihealth.com>`, seems the mail does not linked your GitHub account, thus your contribution is not counted by GitHub, would you like to update your GitHub profile to link the mail?

pan3793:
After linking the mail account to your GitHub account, your avatar will present in the commit history https://github.com/apache/kyuubi/commit/f420238f4f171bab9a04860c5cf7f6b2181621ec

wangmiao1002:
Thank you for your modification. It looks great

PR #4501, opened 2023-03-13
Allow administrator to specify the user for engine admin

turboFei:

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
@@ -196,7 +196,11 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
       @QueryParam("sharelevel") shareLevel: String,
       @QueryParam("subdomain") subdomain: String,
       @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String): Response = {
-    val userName = fe.getSessionUser(hs2ProxyUser)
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -196,7 +196,11 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
       @QueryParam("sharelevel") shareLevel: String,
       @QueryParam("subdomain") subdomain: String,
       @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String): Response = {
-    val userName = fe.getSessionUser(hs2ProxyUser)
+    val userName = if (isAdministrator(fe.getRealUser())) {
+      Option(hs2ProxyUser).getOrElse(fe.getRealUser())
```
turboFei, line 200: it is tricky but I do not want to add a new param for this.

turboFei: cc @pan3793

pan3793:
The basic idea is that administrator can proxy any user even it's not allowed by `core-site.xml`?

turboFei:
> The basic idea is that administrator can proxy any user even it's not allowed by core-site.xml?

yes.

pan3793 (approved):
LGTM, we can centralize all logic about proxy user in the future.

turboFei:
thanks, merging to master

PR #4502, opened 2023-03-13
Reduce build concurency mvnd.minThreads in CI builds

bowenliang123:

- to fix mvnd failure in `Install` step of `Sytle Check` job in (https://github.com/apache/kyuubi/actions/runs/4403008223/jobs/7711886893#step:7:1012)
- by lowering `mvnd.minThreads` to reduce memory pressure in concurrent builds

### DIFF
--- a/build/mvnd
+++ b/build/mvnd
@@ -25,7 +25,7 @@ _CALLING_DIR="$(pwd)"
 _COMPILE_JVM_OPTS="-Xms2g -Xmx2g -XX:ReservedCodeCacheSize=1g -Xss128m"
 
 if [ "$CI" ]; then
-  export MAVEN_CLI_OPTS="-Dmvnd.minThreads=8 --no-transfer-progress --errors --fail-fast -Dstyle.color=always"
+  export MAVEN_CLI_OPTS="-Dmvnd.minThreads=4 --no-transfer-progress --errors --fail-fast -Dstyle.color=always"
 fi
 
 # Installs any application tarball given a URL, the expected tarball name,


build/mvnd
```diff
@@ -25,7 +25,7 @@ _CALLING_DIR="$(pwd)"
 _COMPILE_JVM_OPTS="-Xms2g -Xmx2g -XX:ReservedCodeCacheSize=1g -Xss128m"
 
 if [ "$CI" ]; then
-  export MAVEN_CLI_OPTS="-Dmvnd.minThreads=8 --no-transfer-progress --errors --fail-fast -Dstyle.color=always"
+  export MAVEN_CLI_OPTS="-Dmvnd.minThreads=4 --no-transfer-progress --errors --fail-fast -Dstyle.color=always"
```
yaooqinn, line 28: can we just use nproc to get the num of available processors

bowenliang123: `--threads` supports getting by the num of available processors, but `mvnd.minThreads` does not. The config `mvnd.minThreads` is used on purpose as the fallback, and if `--threads` set later in command, the `mvnd.minThreads` will be ignored.

yaooqinn: I mean

export MAVEN_CLI_OPTS="-Dmvnd.minThreads=$(nproc) --no-transfer-progress --errors --fail-fast -Dstyle.color=always"

bowenliang123: Good suggestion. Will do it in a follow-up if necessary in future.
This PR mainly concerns lower the memory pressure on build concurrency. Let's see whether it helps this time.

Issue #4508, opened 2023-03-14
[Bug] [Kyuubi-hive-beeline] Using Beeline with -i run sql faild


thomasg19930417:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

when using beeline  with -i , Enter will trigger the execution of sql.
I added some logs to the code and found that the content read in the handleMultiLineCmd is empty at this time. According to the context code, I suspect that the consoleReader refers to FileInputStream when the - i parameter is used. Therefore, I added the following changes here and found that it can solve the current problem

```patch
+ if(beeLine.getOpts().getInitFiles() !=  null) {
+   beeLine.setConsoleReader(
+     new ConsoleReader(beeLine.getInputStream(), beeLine.getErrorStream()));
+ }
```
![image](https://user-images.githubusercontent.com/20243868/224956881-9e7f891f-078a-45fd-ab88-482aef48c3e5.png)

### Affects Version(s)

1.7.0 以及以下版本

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

pan3793:
@thomasg19930417 in this case, better to send PR directly, so that reviewers can see your changes and the CI will check if your change breaks something.

thomasg19930417:
> @thomasg19930417 in this case, better to send PR directly, so that reviewers can see your changes and the CI will check if your change breaks something.

Ok, I will submit PR later

PR #4510, opened 2023-03-14
Fix Beeline with -i run sql faild

thomasg19930417:

### DIFF
--- a/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
+++ b/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
@@ -22,6 +22,8 @@ import static org.apache.kyuubi.jdbc.hive.JdbcConnectionParams.*;
 import java.io.*;
 import java.sql.*;
 import java.util.*;
+
+import jline.console.ConsoleReader;
 import org.apache.hive.beeline.logs.KyuubiBeelineInPlaceUpdateStream;
 import org.apache.kyuubi.jdbc.hive.KyuubiStatement;
 import org.apache.kyuubi.jdbc.hive.Utils;
@@ -527,6 +529,9 @@ public class KyuubiCommands extends Commands {
         extra = beeLine.getConsoleReader().readLine(prompt.toString());
       }
 
+      if(beeLine.getOpts().getInitFiles() != null) {
+        beeLine.setConsoleReader(new ConsoleReader(beeLine.getInputStream(), beeLine.getErrorStream()));
+      }
       if (extra == null) { // it happens when using -f and the line of cmds does not end with ;
         break;
       }


thomasg19930417:
#4508

pan3793:
cc @cfmcgrady @cxzl25, would you please take a look?

cxzl25:
Thanks for your contribution. @thomasg19930417 
I haven't reproduced the problem you mentioned locally, can you provide a reproduced script?

thomasg19930417:
> Thanks for your contribution. @thomasg19930417 I haven't reproduced the problem you mentioned locally, can you provide a reproduced script?
@cxzl25 
Just beeline -i init.sql, the sql content can be arbitrary, such as set spark.sql.ansi.enabled=false, wait for the session to start, enter any code and press Enter to trigger sql execution instead of the default semicolon

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
```diff
@@ -521,6 +522,13 @@ public String handleMultiLineCmd(String line) throws IOException {
             "Console reader not initialized. This could happen when there "
                 + "is a multi-line command using -e option and which requires further reading from console");
       }
+
+      if (beeLine.getOpts().getInitFiles() != null) {
+        beeLine.getConsoleReader().shutdown();
+        beeLine.setConsoleReader(
```
cxzl25, line 528: Is it better to call `BeeLine#initializeConsoleReader` directly, this method has set some properties of `ConsoleReader`?

thomasg19930417: Thanks for the review, I have modified the code

cxzl25:
How about this?
org.apache.hive.beeline.KyuubiCommands#connect
```java
      beeLine.runInit();
      beeLine.initializeConsoleReader(null);
```

See Hive master branch has similar problems, you can also submit a PR to Hive.

thomasg19930417:
@cxzl25 
Thanks for your review, I agree that this is a good place to put it, the handleMultiLineCmd call in this method should cause the instance to be re-initialized with each input, and if (beeLine.getOpts().getinitfiles ()! = null) or reinitialize in any case

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
```diff
@@ -484,7 +484,11 @@ public boolean connect(Properties props) throws IOException {
       if (!beeLine.isBeeLine()) {
         beeLine.updateOptsForCli();
       }
+
```
cxzl25, line 487: Rmove empty line

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
```diff
@@ -521,6 +525,7 @@ public String handleMultiLineCmd(String line) throws IOException {
             "Console reader not initialized. This could happen when there "
                 + "is a multi-line command using -e option and which requires further reading from console");
       }
+
```
cxzl25, line 528: ditto

cxzl25 (approved):

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
pan3793, line 528:
```suggestion
```

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
pan3793, line 487:
```suggestion
```

cxzl25 (approved):

pan3793:
Thanks, merged to master/1.7

PR #4516, opened 2023-03-15
Return kyuubi instance in SessionData and OperationData

turboFei:

It is helpful, because now kyuubi gateway does not support multiple nodes ha.


pan3793 (approved):

yaooqinn:
I am wondering whether this should be a map - `fe -> url`?

yaooqinn (approved):

turboFei:
> I am wondering whether this should be a map - fe -> url?

It has been in the SessionData conf field.

But not in the OperationData.

turboFei:
thanks, merged to master

PR #4521, opened 2023-03-15
[KYUUBI #3739][FOLLOWUP][REST] Remove unused parameters in SessionsResource rest docs

turboFei:

as title.

### DIFF
--- a/docs/deployment/migration-guide.md
+++ b/docs/deployment/migration-guide.md
@@ -19,7 +19,7 @@
 
 ## Upgrading from Kyuubi 1.7.0 to 1.7.1
 
-* Since Kyuubi 1.7.1, the REST API of `Open(create) a session` will not contain parameters `protocolVersion`.
+* Since Kyuubi 1.7.1, `protocolVersion` is removed from the request parameters of the REST API `Open(create) a session`. All removed or unknown parameters will be silently ignored and affects nothing.
 
 ## Upgrading from Kyuubi 1.6 to 1.7
 
--- a/docs/deployment/migration-guide.md
+++ b/docs/deployment/migration-guide.md
@@ -17,6 +17,10 @@
 
 # Kyuubi Migration Guide
 
+## Upgrading from Kyuubi 1.7.0 to 1.7.1
+
+* Since Kyuubi 1.7.1, the REST API of `Open(create) a session` will not contain parameters `protocolVersion`.
+
 ## Upgrading from Kyuubi 1.6 to 1.7
 
 * In Kyuubi 1.7, `kyuubi.ha.zookeeper.engine.auth.type` does not fallback to `kyuubi.ha.zookeeper.auth.type`.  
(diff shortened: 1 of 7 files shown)


docs/deployment/migration-guide.md
```diff
@@ -24,7 +24,7 @@
 * Since Kyuubi 1.7, Kyuubi returns engine's information for `GetInfo` request instead of server. To restore the previous behavior, set `kyuubi.server.info.provider` to `SERVER`.
 * Since Kyuubi 1.7, Kyuubi session type `SQL` is refactored to `INTERACTIVE`, because Kyuubi supports not only `SQL` session, but also `SCALA` and `PYTHON` sessions.
   User need to use `INTERACTIVE` sessionType to look up the session event.
-* Since Kyuubi 1.7, the REST API of `Open(create) a session` will not contains parameters `user` `password` and `IpAddr`. User and password should be set in `Authorization` of http request if needed.
+* Since Kyuubi 1.7, the REST API of `Open(create) a session` will not contain parameters `protocolVersion`, `user` `password` and `IpAddr`. User and password should be set in `Authorization` of http request if needed.
```
pan3793, line 27: question: what will happen if a user uses Kyuubi 1.7.0 client access Kyuubi Server 1.7.1

turboFei: Nothing happen. By default，protocol is 0， V1

turboFei: The unknown field should be ignore.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -137,7 +137,7 @@ private[v1] class SessionsResource extends ApiRequestContext with Logging {
     val userName = fe.getSessionUser(request.getConfigs.asScala.toMap)
     val ipAddress = fe.getIpAddress
     val handle = fe.be.openSession(
-      TProtocolVersion.findByValue(request.getProtocolVersion),
+      TProtocolVersion.values().max,
```
yaooqinn, line 140: nit: make it a final val

docs/deployment/migration-guide.md
yaooqinn: We need doc it separately with patch-numbered releases, 1.7 looks aimless

yaooqinn (approved):

docs/deployment/migration-guide.md
```diff
@@ -17,14 +17,18 @@
 
 # Kyuubi Migration Guide
 
+## Upgrading from Kyuubi 1.7.0 to 1.7.1
+
+* Since Kyuubi 1.7.1, the REST API of `Open(create) a session` will not contain parameters `protocolVersion`.
```
pan3793, line 22: I think we should clear document it does not break user to use old REST client to connect Kyuubi Server.
```
Since Kyuubi 1.7.1, `protocolVersion` is removed from the request parameters of the REST API `Open(create) a session`. All removed or unknown parameters will be silently ignored and affects nothing.
```

pan3793 (approved):

Issue #4526, opened 2023-03-15
[Improvement] Ensure that the feature of `Supplying pluggable GroupProvider` is only valid for the GROUP engine share level.


merrily01:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Ensure that the feature of `Supplying pluggable GroupProvider` is only valid for the GROUP engine share level.

### How should we improve?

Fix by moving the sessionMgr.groupProvider to the inside of EngineRef.appUser.

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4527, opened 2023-03-15
[KYUUBI #4526][FOLLOWUP] Defer evaluation for GroupProvider

merrily01:

As described in the discussion: https://github.com/apache/kyuubi/discussions/4512 

Since the pr https://github.com/apache/kyuubi/pull/3897 , we introduced the feature of `Supplying pluggable GroupProvider`.

But now when we create a new EngineRef, no matter which engine share level is used, getGroupNames will always be called.

This may cause users to encounter some confusion stack trace when they do not use the GROUP engine share level.

So we solve the problem through moving the `sessionManager.groupProvider.primaryGroup` to the inside of `EngineRef.appUser`, to ensure that the feature of `Supplying pluggable GroupProvider` is only valid for the GROUP engine share level.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -19,6 +19,7 @@ package org.apache.kyuubi.engine
 
 import java.util.concurrent.TimeUnit
 
+import scala.collection.JavaConverters._
 import scala.util.Random
 
 import com.codahale.metrics.MetricRegistry
@@ -40,6 +41,7 @@ import org.apache.kyuubi.ha.client.{DiscoveryClient, DiscoveryClientProvider, Di
 import org.apache.kyuubi.metrics.MetricsConstants.{ENGINE_FAIL, ENGINE_TIMEOUT, ENGINE_TOTAL}
 import org.apache.kyuubi.metrics.MetricsSystem
 import org.apache.kyuubi.operation.log.OperationLog
+import org.apache.kyuubi.plugin.GroupProvider
 
 /**
  * The description and functionality of an engine at server side
@@ -51,7 +53,7 @@ import org.apache.kyuubi.operation.log.OperationLog
 private[kyuubi] class EngineRef(
     conf: KyuubiConf,
     user: String,
-    primaryGroup: String,
+    groupProvider: GroupProvider,
     engineRefId: String,
     engineManager: KyuubiApplicationManager)
   extends Logging {
@@ -85,7 +87,7 @@ private[kyuubi] class EngineRef(
   // Launcher of the engine
   private[kyuubi] val appUser: String = shareLevel match {
     case SERVER => Utils.currentUser
-    case GROUP => primaryGroup
+    case GROUP => groupProvider.primaryGroup(user, Map.empty[String, String].asJava)
     case _ => user
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -87,7 +87,7 @@ private[kyuubi] class EngineRef(
   // Launcher of the engine
   private[kyuubi] val appUser: String = shareLevel match {
     case SERVER => Utils.currentUser
-    case GROUP => groupProvider.primaryGroup(user, Map.empty[String, String].asJava)
+    case GROUP => groupProvider.primaryGroup(user, conf.getAll.asJava)
     case _ => user
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -77,7 +77,7 @@ class KyuubiSessionImpl(
   lazy val engine: EngineRef = new EngineRef(
     sessionConf,
     user,
-    sessionManager.groupProvider.primaryGroup(user, optimizedConf.asJava),
+    sessionManager.groupProvider,
     handle.identifier.toString,
     sessionManager.applicationManager)
   private[kyuubi] val launchEngineOp = sessionManager.operationManager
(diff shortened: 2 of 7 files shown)


pan3793:
@merrily01, thanks for your contribution, would you mind updating your GitHub profile to link the email <maruilei@58.com> to your GitHub account? Otherwise, your contributions will not be counted by GitHub.

merrily01:
> @merrily01, thanks for your contribution, would you mind updating your GitHub profile to link the email [maruilei@58.com](mailto:maruilei@58.com) to your GitHub account? Otherwise, your contributions will not be counted by GitHub.

Sorry，already updated.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -51,7 +53,7 @@ import org.apache.kyuubi.operation.log.OperationLog
 private[kyuubi] class EngineRef(
     conf: KyuubiConf,
     user: String,
-    primaryGroup: String,
+    groupProvider: GroupProvider,
```
yaooqinn, line 56: It seems also can be lazily initialized, not needed in the constructor?

merrily01: Thanks for your reply.

It seems that the groupProvider needs to obtain from KyuubiSessionManager. Whether it is passing KyuubiSessionManager through the constructor or trying to initialize and create one, the chaotic changes introduced and the cost paid may be greater than the current transformation.

yaooqinn: the cost is mostly test code refactoring, but I'm OK to keep the PR AS-IS.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -77,7 +77,7 @@ class KyuubiSessionImpl(
   lazy val engine: EngineRef = new EngineRef(
     sessionConf,
     user,
-    sessionManager.groupProvider.primaryGroup(user, optimizedConf.asJava),
```
yaooqinn, line 80: this is also a bugfix as using optimized conf is wrong here

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -85,7 +87,7 @@ private[kyuubi] class EngineRef(
   // Launcher of the engine
   private[kyuubi] val appUser: String = shareLevel match {
     case SERVER => Utils.currentUser
-    case GROUP => primaryGroup
+    case GROUP => groupProvider.primaryGroup(user, Map.empty[String, String].asJava)
```
yaooqinn, line 90: sessionConf.getAll

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/EngineRefTests.scala
```diff
@@ -68,7 +71,9 @@ trait EngineRefTests extends KyuubiFunSuite {
     Seq(None, Some("suffix")).foreach { domain =>
       conf.set(KyuubiConf.ENGINE_SHARE_LEVEL, CONNECTION.toString)
       domain.foreach(conf.set(KyuubiConf.ENGINE_SHARE_LEVEL_SUBDOMAIN.key, _))
-      val engine = new EngineRef(conf, user, "grp", id, null)
+      conf.set(KyuubiConf.GROUP_PROVIDER, "hadoop")
+
+      val engine = new EngineRef(conf, user, PluginLoader.loadGroupProvider(conf), id, null)
```
yaooqinn, line 76: The group provider is unnecessary here, but OK to hold down the LOC

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
merrily01: Ok, thank you

yaooqinn (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
merrily01: Yeah, actually it seems that conf is not used in subsequent calls ? 
```
# HadoopGroupProvider.scala

  override def groups(user: String, sessionConf: JMap[String, String]): Array[String] =
    UserGroupInformation.createRemoteUser(user).getGroupNames match {
      case Array() =>
        warn(s"There is no group for $user, use the client user name as group directly")
        Array(user)
      case groups => groups
    }

```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
merrily01: Done

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/EngineRefTests.scala
merrily01: Ok,  thank you

merrily01:
Hi~ @yaooqinn @pan3793  sorry to bother you guys,

I found that some checks in the workflow failed, such as:

```
Error: Exception in thread "pool-182-thread-3" java.lang.Error: org.apache.kyuubi.KyuubiSQLException: Timeout to lock on path [/engine_test2_1.8.0-SNAPSHOT_USER_SPARK_SQL_lock/runner/default] after 4500 ms. There would be some problem that other session may create engine timeout.

```

And I don't know whether it is related to code change or test environment. 

So, mabey I want to know how to re-trigger the pipline?

In addition, by the way, in our project specification, when I modify or reply to the reviewer's comments, 

who should close the `Resolve conversation`? the contributor or reviewer?

pan3793:
> ... whether it is related to code change or test environment.

@merrily01 it looks like a flaky test, @ulysses-you I remember you have fixed it but seems flaky again.

> who should close the Resolve conversation? the contributor or reviewer?

we don't have strict rules for that, normally, the committer will glance the conversations and evaluate if the PR can get merged.

Issue #4525, opened 2023-03-15
[Bug] partition filed can  not filter data   when querying  hive_catalog with hive connector 


silvermissile:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

partition filed not effective in where contition   

``` sql 
CREATE TABLE hive_store.tmp.students (name VARCHAR(64), address VARCHAR(64), student_id INT)
USING PARQUET PARTITIONED BY (student_id);

INSERT INTO hive_store.tmp.students VALUES
('Amy Smith', '123 Park Ave, San Jose', 111111),
('1Bush Brown', '456 Taylor St, Cupertino', 111111),
('Bob Brown', '456 Taylor St, Cupertino', 222222),
('2Bob Brown', '456 Taylor St, Cupertino', 222222),
('Bob Brown', '456 Taylor St, Cupertino', 333333),
('3Woods Brown', '456 Taylor St, Cupertino', 333333),
('Cathy Johnson', '789 Race Ave, Palo Alto', 333333);


select * from hive_store.tmp.students ;
+----------------+---------------------------+-------------+
|      name      |          address          | student_id  |
+----------------+---------------------------+-------------+
| 1Bush Brown    | 456 Taylor St, Cupertino  | 111111      |
| Amy Smith      | 123 Park Ave, San Jose    | 111111      |
| 2Bob Brown     | 456 Taylor St, Cupertino  | 222222      |
| Bob Brown      | 456 Taylor St, Cupertino  | 222222      |
| Cathy Johnson  | 789 Race Ave, Palo Alto   | 333333      |
| 3Woods Brown   | 456 Taylor St, Cupertino  | 333333      |
| Bob Brown      | 456 Taylor St, Cupertino  | 333333      |
+----------------+---------------------------+-------------+



select student_id,  count(1)   from hive_store.tmp.students where group by  student_id;
+-------------+-----------+
| student_id  | count(1)  |
+-------------+-----------+
| 111111      | 2         |
| 222222      | 2         |
| 333333      | 3         |
+-------------+-----------+

select * from hive_store.tmp.students where student_id= 333333;
+----------------+---------------------------+-------------+
|      name      |          address          | student_id  |
+----------------+---------------------------+-------------+
[...]

PR #4528, opened 2023-03-15
[KYUUBI #4525][KSHC] Partitioning predicates should take effect to filter data

Yikf:

This PR aims to close https://github.com/apache/kyuubi/issues/4525.

The root cause of this problem is that Apache Spark does predicate push-down in `V2ScanRelationPushDown`, but the spark-hive-connector does not apply push-down predicates for data filtering.

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveScanBuilder.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveScanBuilder.scala
@@ -37,6 +37,8 @@ case class HiveScanBuilder(
       catalogTable = table,
       dataSchema = dataSchema,
       readDataSchema = readDataSchema(),
-      readPartitionSchema = readPartitionSchema())
+      readPartitionSchema = readPartitionSchema(),
+      partitionFilters = partitionFilters,
+      dataFilters = dataFilters)
   }
 }
--- a/extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveQuerySuite.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveQuerySuite.scala
@@ -107,6 +107,30 @@ class HiveQuerySuite extends KyuubiHiveTest {
     }
   }
 
+  test("[KYUUBI #4525] Partitioning predicates should take effect to filter data") {
+    withSparkSession(Map("hive.exec.dynamic.partition.mode" -> "nonstrict")) { spark =>
+      val table = "hive.default.employee"
+      withTempPartitionedTable(spark, table) {
+        spark.sql(
+          s"""
+             | INSERT OVERWRITE
+             | $table
+             | VALUES("yi", "2022", "0808"),("yi", "2023", "0316")
+             |""".stripMargin).collect()
+
+        checkQueryResult(
+          s"select * from $table where year = '2022'",
+          spark,
+          Array(Row.apply("yi", "2022", "0808")))
+
+        checkQueryResult(
+          s"select * from $table where year = '2023'",
+          spark,
+          Array(Row.apply("yi", "2023", "0316")))
+      }
+    }
+  }
+
   test("Partitioned table insert and all static insert") {
     withSparkSession() { spark =>
       val table = "hive.default.employee"


Yikf:
Please take a look if you find time, thanks @pan3793.

BTW, The reason I missed this is that predicate pushdown is implemented `withFilter` in Apache Spark 3.2, but Apache Spark3.3 changes this.

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveScanBuilder.scala
```diff
@@ -37,6 +37,8 @@ case class HiveScanBuilder(
       catalogTable = table,
       dataSchema = dataSchema,
       readDataSchema = readDataSchema(),
-      readPartitionSchema = readPartitionSchema())
+      readPartitionSchema = readPartitionSchema(),
+      partitionFilters = partitionFilters,
+      dataFilters = dataFilters)
```
pan3793, line 42: Does `dataFilters` affects correctness?

ulysses-you: It is designed as a kind of file pruning, so it only affect performance. For this hive datasource v2, I think it has no effect.

ulysses-you:
> BTW, The reason I missed this is that predicate pushdown is implemented withFilter in Apache Spark 3.2, but Apache Spark3.3 changes this.

@Yikf do you mean after this pr, this connector only works with Spark3.3 ?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveScanBuilder.scala
Yikf: Yes, `dataFilters` has no effect on correctness. 

KSHC does not support pushdown filters to file sources, so, in fact, both `dataFilters` and `pushedFilters` are useless.

pan3793 (approved):

ulysses-you:
I did not see any limitation that it only works with Spark3.3. Shall we pin Spark version to 3.3 in pom.xml at hive connector module ?  Otherwise it follows the parent pom.xml which can be changed by profile.

pan3793:
@ulysses-you it's already done, FYI
https://github.com/apache/kyuubi/blob/0f45f26a354490c763eec186dacf0615b783b9dd/pom.xml#L2332-L2343

ulysses-you:
ah, then it's fine

pan3793:
Thanks, merged to master/1.7

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveScanBuilder.scala
flaming-archer: > Yes, `dataFilters` has no effect on correctness.
> 
> KSHC does not support pushdown filters to file sources, so, in fact, both `dataFilters` and `pushedFilters` are useless.

I understand that this PR cannot solve pushedfilters, only partitionFilters are used, dataFilters are still not used.Is that right?  `   DataFilters are never used ,see override def listFiles(
      partitionFilters: Seq[Expression],
      dataFilters: Seq[Expression]): Seq[PartitionDirectory] = {
    def isNonEmptyFile(f: FileStatus): Boolean = {
      isDataPath(f.getPath) && f.getLen > 0
    }
    val selectedPartitions =
      if (partitionSpec().partitionColumns.isEmpty) {
        HiveConnectorUtils.createPartitionDirectory(
          InternalRow.empty,
          allFiles().filter(isNonEmptyFile)) :: Nil
      } else {
        if (recursiveFileLookup) {
          throw new IllegalArgumentException(
            "Datasource with partition do not allow recursive file loading.")
        }
        prunePartitions(partitionFilters, partitionSpec()).map {
          case partPath @ PartitionPath(values, path) =>
            val files: Seq[FileStatus] = leafDirToChildrenFiles.get(path) match {
              case Some(existingDir) =>
                // Directory has children files in it, return them
                existingDir.filter(f => matchPathPattern(f) && isNonEmptyFile(f))

              case None =>
                // Directory does not exist, or has no children files
                Nil
            }
            val partDir = HiveConnectorUtils.createPartitionDirectory(values, files)
            // Update Partition Directory -> binding Hive part map
            updatePartDirHivePartitionMapping(partDir, partPath)

            partDir
        }
      }
    logTrace("Selected files after partition pruning:\n\t" + selectedPartitions.mkString("\n\t"))
    selectedPartitions
  }`

flaming-archer: > Yes, `dataFilters` has no effect on correctness.
> 
> KSHC does not support pushdown filters to file sources, so, in fact, both `dataFilters` and `pushedFilters` are useless.

PushedFilters don't  used either.

PR #4531, opened 2023-03-15
[KYUUBI#4530] fix data masking  does not support CJK for strategies such as MASK_SHOW_FIRST_4 or MASK_SHOW_FIRST_4

Jackhjf:
…es such as MASK_SHOW_FIRST_4 or MASK_SHOW_FIRST_4

The data masking  does not support Chinese for strategies such as MASK_SHOW_FIRST_4 or MASK_SHOW_FIRST_4
image

create table iceberg_spark.test_copy(id string,name string);
insert into  iceberg_spark.test_copy values(“111111”，“中国人人人儿呢热");
insert into  iceberg_spark.test_copy values(“222223”，“rrrrxxxxxx");

Configure the ranger strategy
image
image

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
@@ -135,7 +135,8 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     val pos = if (hasLen) ", 5" else ""
     val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
-    val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
+    val chinese = s"regexp_replace($lower, '[\u4e00-\u9fa5]', 'x'$pos)"
+    val digits = s"regexp_replace($chinese, '[0-9]', 'n'$pos)"
     digits
   }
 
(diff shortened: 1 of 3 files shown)


extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
```diff
@@ -135,7 +135,8 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     val pos = if (hasLen) ", 5" else ""
     val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
-    val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
+    val chinese = s"regexp_replace($lower, '[\u4e00-\u9fa5]', 'x'$pos)"
```
pan3793, line 138: What's the behavior on other popular databases/OLAP systems?
And I remember that UTF-8 reserved range for CJK, can we extend it to cover CJK?

Jackhjf: > What's the behavior on other popular databases/OLAP systems? And I remember that UTF-8 reserved range for CJK, can we extend it to cover CJK?
Spark needs to access the metastore to query data, and the metadata information in the metastore is stored in mysql.In mysql, the character set is also utf-8
image

CJK can be covered by the following regular expression
image

yaooqinn:
Do we need a custom masker here instead? the hive masker udfs only support alphabets&digits

yaooqinn:
closed via #4643

PR #4532, opened 2023-03-16
[AUTHZ] Displays the columns involved in extracting the aggregation operator

Yikf:

This PR aims to display the columns involved in extracting the aggregation operator.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
@@ -1645,6 +1645,48 @@ class HiveCatalogPrivilegeBuilderSuite extends PrivilegesBuilderSuite {
     val accessType = ranger.AccessType(po, operationType, isInput = false)
     assert(accessType === AccessType.CREATE)
   }
+
+  test("KYUUBI #4532: Displays the columns involved in extracting the aggregation operator") {
+    // case1: There is no project operator involving all columns.
+    val plan1 = sql(s"SELECT COUNT(key), MAX(value) FROM $reusedPartTable GROUP BY pid")
+      .queryExecution.optimizedPlan
+    val (in1, out1, _) = PrivilegesBuilder.build(plan1, spark)
+    assert(in1.size === 1)
+    assert(out1.isEmpty)
+    val pi1 = in1.head
+    assert(pi1.columns.size === 3)
+    assert(pi1.columns === Seq("key", "value", "pid"))
+
+    // case2: Some columns are involved, and the group column is not selected.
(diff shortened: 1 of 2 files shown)


Yikf:
Could you please take a look if you find time, thanks @yaooqinn @bowenliang123 

Although the `Aggregate` involved columns will sink into the `Project`, is it necessary to display the extracted agg columns to prevent Spark Optimizer Rule changes?

yaooqinn:
Make sense to me. Can we also add some tests with having/grouping cases?

bowenliang123:
I suggest providing relevant tests for evaluation with/without extracting aggregation columns.

bowenliang123:
As new tests for aggregation pass successfully without extracting columns from Aggregate, I prefer to keep the tests but not the extraction.

Yikf:
> As new tests for aggregation pass successfully without extracting columns from Aggregate, I prefer to keep the tests but not the extraction.

Do you mean just add test coverage instead of changing the code?

bowenliang123:
> > As new tests for aggregation pass successfully without extracting columns from Aggregate, I prefer to keep the tests but not the extraction.
> 
> Do you mean just add test coverage instead of changing the code?

Yes, I think so.

Yikf:
> > > As new tests for aggregation pass successfully without extracting columns from Aggregate, I prefer to keep the tests but not the extraction.
> > 
> > 
> > Do you mean just add test coverage instead of changing the code?
> 
> Yes, I think so.

In fact, this PR is not a bug fix, existing logic can also extract agg columns, this is because Spark pushes agg columns down to the project operator, while kyuubi overrides the project operator.

Adding this pr is just an optimization in case the agg column is not pushed down to project after the spark optimizer rule is changed

bowenliang123:
Well noticed, it's good to add ut for extending coverage in aggregation with and without aggregation columns used in output projections.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
```diff
@@ -1645,6 +1645,26 @@ class HiveCatalogPrivilegeBuilderSuite extends PrivilegesBuilderSuite {
     val accessType = ranger.AccessType(po, operationType, isInput = false)
     assert(accessType === AccessType.CREATE)
   }
+
+  test("Displays the columns involved in extracting the aggregation operator") {
```
yaooqinn, line 1649: can we also test the having-clause

Yikf: done

yaooqinn: KYUUBI #4532:

yaooqinn (approved):
LGTM

Yikf:
thanks all, merged to master/1.7

Issue #4533, opened 2023-03-16
[Improvement] [AUTHZ] Cols extracted by authZ should resolve caseSensitive


Yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Cols extracted by authZ should resolve caseSensitive.

Let's say that have a case like as following:
```sql
CREATE TABLE t (pid STRING, key INT) USING PARQUET
```

```sql
SELECT key, pid FROM t WHERE PID = '1'
```
This query should be authenticated against columns key and pid, not key, pid, PID

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4534, opened 2023-03-16
[KYUUBI #4533][AUTHZ] Cols extracted by authZ should resolve caseSensitive

Yikf:

This PR aims to close https://github.com/apache/kyuubi/issues/4533

Cols extracted by authZ should resolve caseSensitive.

Let's say that have a case like as following:
```sql
CREATE TABLE t (pid STRING, key INT) USING PARQUET
```

```sql
SELECT key, pid FROM t WHERE PID = '1'
```
This query should be authenticated against columns key and pid, not key, pid, PID

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
@@ -65,8 +65,12 @@ object PrivilegesBuilder {
       if (projectionList.isEmpty) {
         privilegeObjects += PrivilegeObject(table, plan.output.map(_.name))
       } else {
+        val planOutputMapping = plan.output.map(e => e.exprId -> e.name).toMap
         val cols = (projectionList ++ conditionList).flatMap(collectLeaves)
-          .filter(plan.outputSet.contains).map(_.name).distinct
+          .filter(e => plan.outputSet.contains(e))
+          .map(_.exprId)
+          .distinct
+          .map(planOutputMapping(_))
         privilegeObjects += PrivilegeObject(table, cols)
       }
     }
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
@@ -1687,6 +1687,26 @@ class HiveCatalogPrivilegeBuilderSuite extends PrivilegesBuilderSuite {
     assert(pi4.columns.size === 2)
     assert(pi4.columns === Seq("key", "pid"))
   }
+
+  test("KYUUBI #4533: Cols extracted by authZ should resolve caseSensitive") {
+    val plan1 = sql(s"SELECT key, pid FROM $reusedPartTable WHERE PID = '1'")
+      .queryExecution.optimizedPlan
+    val (in1, out1, _) = PrivilegesBuilder.build(plan1, spark)
+    assert(in1.size === 1)
+    assert(out1.isEmpty)
+    val pi1 = in1.head
+    assert(pi1.columns.size === 2)
+    assert(pi1.columns === Seq("key", "pid"))
+
+    val plan2 = sql(s"SELECT key FROM $reusedPartTable WHERE PID = '1'")
+      .queryExecution.optimizedPlan
+    val (in2, out2, _) = PrivilegesBuilder.build(plan2, spark)
+    assert(in2.size === 1)
+    assert(out2.isEmpty)
+    val pi2 = in2.head
+    assert(pi2.columns.size === 2)
+    assert(pi2.columns === Seq("key", "pid"))
+  }
 }
 
 case class SimpleInsert(userSpecifiedSchema: StructType)(@transient val sparkSession: SparkSession)


Yikf:
Please take a look if you have time, thanks @yaooqinn @bowenliang123

yaooqinn:
can we add and an e2e test case?

Yikf:
> can we add an e2e test case?

done

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -437,6 +437,28 @@ abstract class RangerSparkExtensionSuite extends AnyFunSuite
     }
     doAs("admin", assert(sql("show tables from global_temp").collect().length == 0))
   }
+
+  test("KYUUBI #4533: Cols extracted by authZ should resolve caseSensitive") {
```
yaooqinn, line 441: what is the purpose of this test?

yaooqinn:
can you also check how apache/ranger deal with the case sensitivity

yaooqinn:
The current approach is to ignore the case sensitivity of both the spark and ranger sides, and we send all attributes blindly/transparently to apache/ranger and let it judge.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
```diff
@@ -1687,6 +1687,26 @@ class HiveCatalogPrivilegeBuilderSuite extends PrivilegesBuilderSuite {
...
+
+  test("KYUUBI #4533: Cols extracted by authZ should resolve caseSensitive") {
+    val plan1 = sql(s"SELECT key, pid FROM $reusedPartTable WHERE PID = '1'")
+      .queryExecution.optimizedPlan
+    val (in1, out1, _) = PrivilegesBuilder.build(plan1, spark)
+    assert(in1.size === 1)
+    assert(out1.isEmpty)
+    val pi1 = in1.head
+    assert(pi1.columns.size === 2)
+    assert(pi1.columns === Seq("key", "PID"))
```
bowenliang123, line 1699: What is the exact rules of extracting the column name, either conform to runtime declared in SQL or to the original name in source table.

Based on my plain understanding to the word `caseSensitive`, it that ["key", "pid", "PID"] should be the the correct output columns of the query "SELECT key, pid FROM $reusedPartTable WHERE PID = '1'" ? Here the output with  ["key", "PID"] reflects "pid" is overritten by "PID".

But curiously the original column name in source table of the test is `pid` , all in lower case.

And if you are following the declared case sensitive way in SQL, please also other cases should be discussed and tested, including the mixed combinationof of uppercase and lowercase letters.

bowenliang123: As for me, I would prefer to use the original name of the source table instead of case-sensitive declared column in runtime SQL.

Yikf:
> can you also check how apache/ranger deal with the case sensitivity

I tested it with an internal plugin and found it to be insensitive. But, I think that this is most accurate if it matches the case of the relation output.

The current approach may duplicate authentication for the same column. This PR fix this issue.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
Yikf: Yes, we should keep consistent with raw table.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
```diff
@@ -65,8 +65,12 @@ object PrivilegesBuilder {
       if (projectionList.isEmpty) {
         privilegeObjects += PrivilegeObject(table, plan.output.map(_.name))
       } else {
+        val planOutputMapping = plan.output.map(e => e.exprId -> e.name).toMap
```
Yikf, line 68: 1. Construct a standard dictionary
2. Dereferenced `NamedExpression` by `exprId`
3. Get the column name of the original table by querying the dictionary with `exprId`

yaooqinn:
> I tested it with an internal plugin and found it to be insensitive. 

Well, the case sensitivity of the spark resolver itself is a mass, and now it supports multi-catalogs which makes it even worse in such a case. I don't think we shall add a transformer here to make thing much more complicated

yaooqinn:
> The current approach may duplicate authentication for the same column

This is not a big deal. It is a defense one

Yikf:
> > The current approach may duplicate authentication for the same column
> 
> This is not a big deal. It is a defense one

Agree with you.

BTW, We always refer to the original table columns of conditions. What's wrong with that? 

An example is as follows:  -> Have a table `t1`, it has a col named `nAme`; 

Conditions contain `NAME` and `name`, If `Name`'s exprId, `name`'s exprId and `nAme`'s exprId are consistent, they always point to the original table's `nAme`,  Whether Spark is case-sensitive or not, only colA should be authenticated, right?

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilderSuite.scala
bowenliang123: SGTM.

yaooqinn:
> Whether Spark is case-sensitive or not, only `nAme` should be authenticated, right?

As an open system, we can not assure that by ourselves. While users can build a 'closed system' based on spark, ranger, and our plugin, they can determine whether the data/meta store, and the policy shares the same case sensitivity rule.

PR #4541, opened 2023-03-16
Support to customize the attributes to expose for Spark engine

turboFei:

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -2764,6 +2764,15 @@ object KyuubiConf {
       .stringConf
       .createWithDefault("bin/python")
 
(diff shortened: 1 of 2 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2677,6 +2677,15 @@ object KyuubiConf {
       .stringConf
       .createWithDefault("bin/python")
 
+  val ENGINE_SPARK_REGISTER_ATTRIBUTES: ConfigEntry[Seq[String]] =
+    buildConf("kyuubi.engine.spark.register.attributes")
+      .internal
+      .doc("The attributes to expose when registering for Spark engine.")
```
pan3793, line 2683: We have some built-in attributes, so let's add extra: "attributes" -> "extra attributes".

pan3793 (approved):

zwangsheng (approved):

pan3793:
cc @cxzl25 as well, since you were done similar things

turboFei:
thanks, merged to master

Issue #4523, opened 2023-03-15
[Subtask] support close engine session gracefully


lightning-L:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues) and found no similar issues.

### Describe the subtask

in engine side, do not terminate the session life cycle once receive the close session request, support fireoff, and wait the current statements finished(configurable)

### Parent issue

https://github.com/apache/kyuubi/issues/4426

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4545, opened 2023-03-17
[KYUUBI #4523] support close engine session gracefully

lightning-L:


lightning-L:
@turboFei

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -157,6 +157,10 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
   }
 
   override def closeSession(sessionHandle: SessionHandle): Unit = {
+    val closeGracefully = conf.get(ENGINE_SPARK_SESSION_CLOSE_GRACEFULLY)
+    if (closeGracefully) {
+      operationManager.waitForOperationsToFinish()
```
turboFei, line 162: Is it possible to interrupt this if the connection back？

turboFei: We can concern about this after supporting reconnection.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -157,6 +157,10 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
   }
 
   override def closeSession(sessionHandle: SessionHandle): Unit = {
+    val closeGracefully = conf.get(ENGINE_SPARK_SESSION_CLOSE_GRACEFULLY)
```
ulysses-you, line 160: is it possible to make this feature work on all engines ? i.e., move code to `SessionManager.closeSession`

turboFei: +1, we also need to do not close engine operation in kyuubi gateway side.

lightning-L: done

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala
```diff
@@ -189,4 +192,30 @@ abstract class OperationManager(name: String) extends AbstractService(name) {
...
+      val timedOut = diff > stopTimeoutMs
+      if (timedOut) {
+        warn("Timed out while waiting the spark operations to finish " +
+          "(timeout = " + stopTimeoutMs + ")")
+      }
+      timedOut
+    }
+
+    info("Waiting for all the operations to finish.")
+    while (!hasTimedOut && !allOperations().filter(op =>
```
ulysses-you, line 213: `allOperations().exists(...)` to avoid transform tail operations if has active operation.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -157,19 +157,20 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
...
-      val session = getSession(sessionHandle)
-      if (session != null) {
-        userIsolatedCacheLock.synchronized {
-          if (userIsolatedCacheCount.containsKey(session.user)) {
-            val (count, _) = userIsolatedCacheCount.get(session.user)
-            userIsolatedCacheCount.put(session.user, (count - 1, System.currentTimeMillis()))
+    try {
+      super.closeSession(sessionHandle)
+
+      if (!userIsolatedSparkSession) {
```
ulysses-you, line 163: it reminders me that, shall we only wait for the operations which belong to user ? e.g., for group level sharing, it has no meaning to wait for all operations finish if one user want to close a session.

yaooqinn:
what is the purpose of this PR?

yaooqinn:
Why do we need to block a session from closing to support reconnecting? @turboFei

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
lightning-L: yes, it is useful for CONNECTION share level (spark engine will not be stopped and app could finish even though client connection dies for some reasons). For GROUP share level, no need to wait

lightning-L:
We have some uses cases that use CONNECTION share level and may be running some heavy jobs which take hours to finish. 
Sometimes, client connection to gateway dies for unexpected reasons (network jitter, pod recreate, etc). According to current implementation, closeSession will be called, and spark engine will be stopped and running app will be killed.
And we need to enable recoverability for such use cases.

So we need to wait for active operations to finish when closing, and also support reconnect.

@yaooqinn 
@turboFei  correct me if I am wrong

yaooqinn:
> `closeSession` will be called

Shall we bypass `closeSession` directly? 

Another thing is that we need not check running operations to close or remain. The client can crash at any point, and reconnect if it tries.

lightning-L:
bypass be.closeSession directly is also feasible and is easier to implement.
And we will have timeout checker for session. So if client crashes and never comes back, the session could also be closed eventually.

yaooqinn:
Let's implement this in `FeTServerEventHandler#deleteContext` for problematic cases first. The current way has a narrow scope on the spark engine but a broad impact on different kinds of closing calls.

lightning-L:
I have found similar ticket in hive:
https://issues.apache.org/jira/browse/HIVE-13415
https://github.com/apache/hive/blob/master/service/src/java/org/apache/hive/service/cli/thrift/ThriftBinaryCLIService.java#L153

yaooqinn:
The hive one looks reasonable to me. Let's port it from Hive based on our scenario

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkSQLOperationManager.scala
```diff
@@ -27,7 +27,7 @@ import org.apache.kyuubi.config.KyuubiReservedKeys.KYUUBI_OPERATION_HANDLE_KEY
 import org.apache.kyuubi.engine.spark.repl.KyuubiSparkILoop
 import org.apache.kyuubi.engine.spark.session.SparkSessionImpl
 import org.apache.kyuubi.engine.spark.shim.SparkCatalogShim
-import org.apache.kyuubi.operation.{NoneMode, Operation, OperationHandle, OperationManager, PlanOnlyMode}
+import org.apache.kyuubi.operation._
 import org.apache.kyuubi.session.{Session, SessionHandle}
```
turboFei, line 31: Need to revert this

docs/deployment/settings.md
```diff
@@ -372,6 +372,7 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
 | kyuubi.session.check.interval                        | PT5M                    | The check interval for session timeout.                                                                                                                                                                                                                                                                                                                                                                                  | duration | 1.0.0 |
+| kyuubi.session.close.on.disconnect                   | true                    | Session will be closed when connection is closed. Set this to false to have session outlive its parent connection.                                                                                                                                                                                                                                                                                                       | boolean  | 1.8.0 |
```
turboFei, line 375: For the description，'closed' is not accurate.

turboFei (approved):

turboFei:
BTW, kyuubi is not same with hive.

I think that we shall close the session from kyuubi gateway to engine.

turboFei:
> BTW, kyuubi is not same with hive.
> 
> I think that we shall close the session from kyuubi gateway to engine.


We can do this in another pr that to support re-connect.

yaooqinn (approved):

turboFei:
Thanks @lightning-L ，merged to master.

turboFei:
Seems it is easy to touch the limitation `kyuubi.server.limit.connections.per.user` if enable this config.

how do you think about? @yaooqinn

turboFei:
how about just reserve the sessions in engine side?

turboFei:
cc @lightning-L  @yaooqinn

yaooqinn:
seems OK to me

PR #4554, opened 2023-03-17
[CHAT] Code improvement in ChatGPTProvider

bowenliang123:

- set authentication as default header in client construction instead of  request construction
- handle response's status code in scala style 
- transforming config's long value to int with `.intValue` instead of `asInstanceOf` casting
- fix var name to `response`


pan3793 (approved):

pan3793:
Thanks, merged to master

PR #4558, opened 2023-03-19
[CHAT] Make ChatGPT model ID configurable

bowenliang123:

- Make ChatGPT model ID configurable


pan3793 (approved):

bowenliang123:
According to OpenAI's API reference(https://platform.openai.com/docs/api-reference/completions/create), many detailed settings of chat completion request, other than model id, like `max_tokens`, `topP`,  can be configured.
Do you guys think we should expose them one by one in separate config in Kyuubi , or just expose a config in JSON style, that consumes dynamic settings for ChatGPT? @pan3793 @cxzl25

pan3793:
I would suggest just exposing the most frequently usage small set of the whole parameters.

pan3793:
Please add this PR into the Umbrella task list

bowenliang123:
> I would suggest just exposing the most frequently usage small set of the whole parameters.

Agree. And let's expose parameters on specific purpose.

bowenliang123:
> Please add this PR into the Umbrella task list

Added. Thx for reminder.

bowenliang123:
Thanks, merged to master.

PR #4560, opened 2023-03-19
[KSHC] Support Kerberized HMS in cluster mode w/o keytab

pan3793:

This PR aims to make Kyuubi Spark Hive Connector(KSHC) support kerberized HMS in cluster mode w/o keytab(which is the typical use case in Kyuubi) by implementing a `HadoopDelegationTokenProvider`.

To enable access to an kerberized HMS using KSHC, the minimal configurations are
```
spark.sql.catalog.warm=org.apache.kyuubi.spark.connector.hive.HiveTableCatalog
spark.sql.catalog.warm.hive.metastore.uris=<thrift-uris>
```
then it's able to run federation query across metastores 
```
SELECT * FROM spark_catalog.db1.tbl1 JOIN warm.db2.tbl2 ON ...
```

In addition, it allows disabling token renewal for each catalog explicitly
```
spark.sql.catalog.warm.delegation.token.renewal.enabled=false
```

The current implementation has some limitations:

the catalog configuration must be present on the Spark application bootstrap, which means the catalog configurations should be set in `spark-defaults.conf` or append as `--conf` like:
```
spark-[sql|shell|submit] \
  --conf spark.sql.catalog.xxx=org.apache.kyuubi.spark.connector.hive.HiveTableCatalog
  --conf spark.sql.catalog.xxx.hive.abc=xyz
```

but does not work for dynamic registering through SET statement, e.g. `SET spark.sql.catalog.xxx=`

```
> (select count(*) from hive_2.mammut.test_7) union ( select count(*) from spark_catalog.test.test01 limit 1);
+-----------+
| count(1)  |
+-----------+
| 4         |
| 1         |
+-----------+
2 rows selected (8.378 seconds)
```

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
@@ -73,9 +73,12 @@ class HiveTableCatalog(sparkSession: SparkSession)
 
   private lazy val hadoopConf: Configuration = {
     val conf = sparkSession.sessionState.newHadoopConf()
-    catalogOptions.asScala.foreach {
-      case (k, v) => conf.set(k, v)
-      case _ =>
+    catalogOptions.asScala.foreach { case (k, v) => conf.set(k, v) }
+    // fallback to `hive.metastore.uris` if token signature is absent, the logic is
+    // consistent w/ KyuubiHiveConnectorDelegationTokenProvider
+    if (!catalogOptions.containsKey("hive.metastore.token.signature")
+      && catalogOptions.containsKey("hive.metastore.uris")) {
+      conf.set("hive.metastore.token.signature", catalogOptions.get("hive.metastore.uris"))
     }
     conf
   }
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
@@ -44,6 +44,7 @@ import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
 import org.apache.kyuubi.spark.connector.hive.HiveTableCatalog.{toCatalogDatabase, CatalogDatabaseHelper, IdentifierHelper, NamespaceHelper}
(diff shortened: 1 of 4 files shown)


pan3793:
cc @zhouyifan279

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/KyuubiHiveConnectorDelegationTokenProvider.scala
```diff
@@ -0,0 +1,184 @@
...
+  logInfo(s"Hadoop Delegation Token provider for service [$serviceName] is initialized.")
+
+  override def serviceName: String = "kyuubi-hive-connector"
+
+  private val classNotFoundErrorStr = "You are attempting to use the " +
+    s"${getClass.getCanonicalName}, but your Spark distribution is not built with Hive libraries."
+
+  private def hiveConf(
+      sparkConf: SparkConf,
+      hadoopConf: Configuration,
```
yaooqinn, line 51: The Hadoop configuration here seems to be shared by all hive clients, which means you share the same security key to open all the doors of different HMS-s?

pan3793: The catalog `hive2` could overwrite some dedicated Hadoop configurations by `spark.sql.catalog.hive2.<hadoop-conf-key>=<value>`.

What do you mean "security key"?

pan3793:
@yaooqinn main changes after your last review, please take a look again

- switch from `RetryingMetaStoreClient` to `HiveMetaStoreClient`
- one token request failure does not block the subsequent token obtaining
- allows to disable token renewal for catalog explicitly

yaooqinn (approved):

zhouyifan279:
Most user may not be familiar with `hive.metastore.token.signature`'s usage.
Can we generate a `hive.metastore.token.signature` for each catalog by default?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -73,9 +73,12 @@ class HiveTableCatalog(sparkSession: SparkSession)
 
   private lazy val hadoopConf: Configuration = {
     val conf = sparkSession.sessionState.newHadoopConf()
-    catalogOptions.asScala.foreach {
-      case (k, v) => conf.set(k, v)
-      case _ =>
```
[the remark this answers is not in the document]
Yikf: useless branch

Yikf (approved):

pan3793:
Thanks all, merged to master

PR #4569, opened 2023-03-21
Extract JsonUtils in common module

bowenliang123:


yaooqinn:
It's a terrible artifact for sharing across modules

PR #4574, opened 2023-03-21
Support stage level schedule for final write stage

ulysses-you:

This pr adds a new rule `FinalStageResourceManager` to inject custom reousrce profile for the final write stage. 

It  nowprovide two features:

- Kill redundant executors
We first get the final stage partition which is the actually required cores, then kill the redundant executors. The priority of kill executors follow:
  1. kill executor who is younger than other (The older the JIT works better)
  2. kill executor who produces less shuffle data first 

- Custom resource profile
We can specify the custom executor resource for final write stage. It now supports:
  ```
  executor core
  executor memory
  executor memory overhead
  executor off heap memory
  ```

The reason why add this feature is that, if the previous stage contains lots executors but final stage has less, then the tasks of final stage would be scheduled randomly in all exists executors which may cause resource waste. e.g., each executor only run 1 or 2 tasks but holds 4 or 5 cores.

test manually

- test for the kill executor
image

- test for custom resource profile
image

image


docs/extensions/engines/spark/rules.md
```diff
@@ -63,7 +63,7 @@ Now, you can enjoy the Kyuubi SQL Extension.
 
 Kyuubi provides some configs to make these feature easy to use.
 
-|                                Name                                 | Default Value |                                                                                                                                                                     Description                                                                                                                                                                      | Since |
+| Name                                                                | Default Value |                                                                                                                                                                     Description                                                                                                                                                                      | Since |
```
merrily01, line 66: Nit: no changes required

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -190,4 +190,43 @@ object KyuubiSQLConf {
       .version("1.7.0")
       .booleanConf
       .createWithDefault(true)
+
+  val FINAL_WRITE_STAGE_RESOURCE_ISOLATION_ENABLE =
```
merrily01, line 194: Nit:

Is it possible to standardize the style of variables, such as the end of _ENABLE`D` or simply remove _ENABLE?

e.g:

`FINAL_STAGE_CONFIG_ISOLATION_WRITE_ONLY` or `ZORDER_USING_ORIGINAL_ORDERING_ENABLED `

ulysses-you: thank you for the nit, it should be `enabled`

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/spark/CustomResourceProfileExec.scala
```diff
@@ -0,0 +1,103 @@
...
+      .getOrElse(sparkContext.getConf.get("spark.executor.memoryOverhead", "1G"))
+  private val executorOffHeapMemory = conf.getConf(FINAL_WRITE_STAGE_EXECUTOR_OFF_HEAP_MEMORY)
+
+  override lazy val metrics: Map[String, SQLMetric] = {
+    val base = Map(
+      "executorCores" -> SQLMetrics.createMetric(sparkContext, "executor cores"),
+      "executorMemory" -> SQLMetrics.createMetric(sparkContext, "executor memory (GB)"),
+      "executorMemoryOverhead" -> SQLMetrics.createMetric(
+        sparkContext,
+        "executor memory overhead (GB)"))
```
pan3793, line 62: For consistent, maybe better to use `MiB`

```
  private[spark] val EXECUTOR_MEMORY = ConfigBuilder(SparkLauncher.EXECUTOR_MEMORY)
    .doc("Amount of memory to use per executor process, in MiB unless otherwise specified.")
    .version("0.7.0")
    .bytesConf(ByteUnit.MiB)
    .createWithDefaultString("1g")
```

bowenliang123:
Is there any description available for the manual test results other than the screenshots? I don't quite understand the screenshots showing consequence related to the purposed features.

ulysses-you:
@bowenliang123 I have added some tags, ideally, the screenshots show that:
1. kill executor
2. inject custom resource profile

bowenliang123:
LGTM. Thx for the supplement.

bowenliang123:
LGTM. Thx for the supplement.

yaooqinn:
please separate this PR to small ones

ulysses-you:
I seperate thie pr into two parts: 1. kill executor, 2. inject custom resource profile.

here is the pr https://github.com/apache/kyuubi/pull/4592 for kill executors

PR #4575, opened 2023-03-21
Fix the empty last line may causes the session to exit directly

Kiss736921:
If the last line of the file is a blank line, there may be a cmd whose content is empty but the length is not 0 after split by ";". When running to this cmd, it will return false, causing the session to exit directly.

vim test.sql and write :
image
then start beeline and source this file successfully and session won't exit：
image
image


cxzl25 (approved):

turboFei (approved):

turboFei:
Thanks，merged to master and 1.7

PR #4582, opened 2023-03-23
[FLINK]Under the select statement, the results can be read in a stream

waywtdcc:

Now the select statement waits for the maximum value configured by kyuubi.session.engine.flink.max.rows, and then returns uniformly. This is not necessary, the resulting data can be streamed back. Under the select statement, the results can be read in a stream


waywtdcc:
> Have you verified w/ `beeline --incremental=true`?

Now the flink engine will stop the task and return data only when the number of results reaches kyuubi.session.engine.flink.max.rows for reading unbounded streams, but this may take a long time. The user wants to query a piece of data and return the data, and stop the task when kyuubi.session.engine.flink.max.rows is reached

waywtdcc:
> Have you verified w/ `beeline --incremental=true`?

hello?

PR #4585, opened 2023-03-23
Authz policy file generation

bowenliang123:

- generate Authz policy file for testing to focus on  manage all the policies in the generator
- auto-increased `id` and related `guid` for each policy
- list and reused users, resources, for evaluation impacts when policy changes
- add `policies_base.json` as base template file of authz template including decoration details and service def

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz/pom.xml
@@ -321,7 +321,6 @@
     </dependencies>
 
     <build>
-
         <testResources>
             <testResource>
                 <directory>${project.basedir}/src/test/resources</directory>
@@ -331,4 +330,38 @@
         <testOutputDirectory>target/scala-${scala.binary.version}/test-classes</testOutputDirectory>
     </build>
 
+    <profiles>
+        <profile>
+            <id>genpolicy</id>
+            <activation>
+                <!-- activated when Ranger version is identical to required-->
+                <property>
+                    <name>ranger.version</name>
+                    <value>2.3.0</value>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>build-helper-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>add-test-source</id>
+                                <goals>
+                                    <goal>add-test-source</goal>
+                                </goals>
+                                <phase>generate-sources</phase>
+                                <configuration>
+                                    <sources>
+                                        <source>src/test/gen/scala</source>
+                                    </sources>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
 </project>
--- a/extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/PolicyJsonFileGenerator.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/PolicyJsonFileGenerator.scala
@@ -1,0 +1,357 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.gen
+
+import java.nio.file.Paths
+import java.util.UUID
+import java.util.concurrent.atomic.AtomicLong
+
+import scala.language.implicitConversions
+
+import com.fasterxml.jackson.annotation.JsonInclude.Include
+import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
+import com.fasterxml.jackson.databind.json.JsonMapper
+import com.fasterxml.jackson.databind.node.ObjectNode
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import org.apache.ranger.plugin.model.RangerPolicy
+
+import org.apache.kyuubi.plugin.spark.authz.gen.KRangerPolicyItemAccess.allowTypes
+import org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator.RangerAccessType.{all, alter, create, drop, index, lock, read, select, update, use, write, RangerAccessType}
+import org.apache.kyuubi.plugin.spark.authz.gen.RangerClassConversions.getRangerObject
+
+/**
+ * Generates the policy file to test/main/resources dir.
+ *
+ * Usage:
+ * build/mvn scala:run -pl :kyuubi-spark-authz_2.12
+ * -DmainClass=org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator
+ */
+private object PolicyJsonFileGenerator {
+  def main(args: Array[String]): Unit = {
+    writeRangerServicePolicesJson()
+  }
+
+  final private val mapper: ObjectMapper = JsonMapper.builder()
+    .addModule(DefaultScalaModule)
+    .serializationInclusion(Include.NON_NULL)
+    .build()
+
+  def writeRangerServicePolicesJson(): Unit = {
+    val pluginHome = getClass.getProtectionDomain.getCodeSource.getLocation.getPath
+      .split("target").head
+    val policyFileName = "sparkSql_hive_jenkins.json"
+    val policyFile = Paths.get(pluginHome, "src", "test", "resources", policyFileName).toFile
+    // scalastyle:off println
+    println(s"Writing ranger policies to $policyFileName.")
+    // scalastyle:on println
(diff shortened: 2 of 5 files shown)


bowenliang123:
> Can we use ranger classes?

I suggest not doing so. Classes and their attributes of Ragner differ in different versions.
By using customed classes in the generator, we could guarantee the stable output of policy files.

bowenliang123:
The tests are still running on the JSON file, not the classes introduced here. The customed classes are decoupled with testing and not the context for it. And the default value and template method greatly reduced the redundant code and help us to focus in key details of policies, like resource, access, users.

Yikf:
I also prefer ranger classes, Since these classes are only used by gen policies for tests, Should we use `ranger.version ` classes defined in pom.xml? 

And can verify whether policies generated using classes of the default ranger version will cause instability in the test of [other ranger versions supported by kyuubi](https://kyuubi.readthedocs.io/en/master/security/authorization/spark/build.html#build-against-different-apache-spark-versions).

bowenliang123:
My concern is that to make policy file conform to this generator, we will introduce tests in builds to validate the output from generator an the file content.

But if we do use the Ranger policies class, it fails when it's crossing Ranger versions. (It does happen.)
And also some temporary fields are irrelevant to JSON serialization, we will be hard to control the serialization details to Ranger class itself or exclude them.

yaooqinn:
we can divide them to non test source paths, and add them via `add-test-source` in the runtime

bowenliang123:
Changed to use Ranger classes for policy generation. @yaooqinn @Yikf 

Policy generation with different Ranger versions:
- 0.6.0/0.7.0: 8 errors in compilation
- 1.1.0/1.2.0: 1 errors in compilation
- 2.0.0/2.1.0/2.2.0/2.3.0: passed ✅

bowenliang123:
Closing this PR as no concrete agreement in variation of Ranger version for generation and testing.

bowenliang123:
Reopening this PR as the generator moved away from the default test source and no more compilation errors.

The Command for regenerating policy files with `genpolicy` policy enabled to add `PolicyJsonFileGenerator` as test sources:
```
mvn scala:run -Pgenpolicy \
  -DmainClass=org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator \
  -pl :kyuubi-spark-authz_2.12
```


Testing results with all other different Ranger versions:
- 0.7.0/1.1.0/1.2.0/2.0.0/2.1.0/2.2.0/2.3.0: passed ✅
- 0.6.0 : 6 tests failed (tests of `show functions` and `Support {OWNER} variable for queries` , but they are regression failures from master master, and will be reported to issue later)


cc @yaooqinn @Yikf

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -331,4 +331,38 @@
         <testOutputDirectory>target/scala-${scala.binary.version}/test-classes</testOutputDirectory>
     </build>
 
+    <profiles>
+        <profile>
+            <id>genpolicy</id>
+            <activation>
+                <property>
+                    <name>ranger.version</name>
```
Yikf, line 339: Why do we need this property?

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -331,4 +331,38 @@
...
+                    <value>2.3.0</value>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>build-helper-maven-plugin</artifactId>
```
Yikf, line 347: Just out of curiosity, Why do we need this plugin? 

`src/test/gen/scala` should place the golden file instead of the source.

extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/PolicyJsonFileGenerator.scala
```diff
@@ -0,0 +1,358 @@
...
+
+import org.apache.kyuubi.plugin.spark.authz.gen.KRangerPolicyItemAccess.allowTypes
+import org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator.RangerAccessType.{all, alter, create, drop, index, lock, read, select, update, use, write, RangerAccessType}
+import org.apache.kyuubi.plugin.spark.authz.gen.RangerClassConversions._
+
+/**
+ * Generates the policy file to test/main/resources dir.
+ *
+ * Usage:
+ * mvn scala:run -Pgenpolicy
```
Yikf, line 41: To re-generate policy golden files, run:
{{{
  mvn scala:run -Pgenpolicy -pl :kyuubi-spark-authz_2.12 -DmainClass=org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator
}}}

extensions/spark/kyuubi-spark-authz/pom.xml
bowenliang123: Here adds no property, it acts as the trigger condition to ensure generating policy file is aligned with the targeted ranger version.

extensions/spark/kyuubi-spark-authz/pom.xml
bowenliang123: The generator class is put in `src/test/gen/scala` rather than in default test source path `src/test/gen/scala` , in order to prevent build and test failed on older Ranger versions.

extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/PolicyJsonFileGenerator.scala
bowenliang123: Which part it is suggested, the comment description or the command?

yaooqinn:
> add policies_base.json as base template file of authz template including decoration details and service def

this file contains 1600+ lines, why can not be generated?

extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/RangerGenWrapper.scala
```diff
@@ -0,0 +1,204 @@
...
+    denyPolicyItems: List[RangerPolicyItem] = List.empty,
+    allowExceptions: List[RangerPolicyItem] = List.empty,
+    denyExceptions: List[RangerPolicyItem] = List.empty,
+    dataMaskPolicyItems: List[RangerDataMaskPolicyItem] = List.empty,
+    rowFilterPolicyItems: List[RangerRowFilterPolicyItem] = List.empty,
+    id: Int = 0,
+    guid: String = "",
+    isEnabled: Boolean = true,
+    version: Int = 1) {
+  def get: RangerPolicy = {
```
yaooqinn, line 43: instead of defining a case class and a get method, we can simply define a method with parameters

bowenliang123: Negative, the maximum number of parameters in Scala / Scala fmt is 10.

bowenliang123:
> > add policies_base.json as base template file of authz template including decoration details and service def
> 
> this file contains 1600+ lines, why can not be generated?

Sure can be done in followups, as it requires more handwork to do so.
This PR mainly focuses in policy generation.

extensions/spark/kyuubi-spark-authz/pom.xml
Yikf: > in order to prevent build and test failed on older Ranger versions.

Golden files should be generated manually by the developer, so the profile should be useless, we can have a ENV, like:
``` shell
KYUUBI_AUTHZ_GENERATE_POLICY_GOLDEN_FILES=1 ./build/mvn scala:run -Pgenpolicy -pl :kyuubi-spark-authz_2.12 -DmainClass=org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator -Dranger.version=2.3.0
```

1. KYUUBI_AUTHZ_GENERATE_POLICY_GOLDEN_FILES is used to prevent golden files from being generated for each test, so a can be removed plugin `build-helper-maven-plugin`;
2. -Dranger.version=2.3.0 activates the ranger version;


BTW: Does this generator have to depend on Apache Ranger 2.3.0? So, do we need \<activation\>\/\<property\>\/\<ranger.version\> to enable the version condition?

extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/PolicyJsonFileGenerator.scala
Yikf: DOC

yaooqinn (approved):

extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/RangerGenWrapper.scala
yaooqinn: If this is a formatter issue, it's easy to bypass

extensions/spark/kyuubi-spark-authz/pom.xml
bowenliang123: The profile `genpolicy` is used for including `PolicyJsonFileGenerator` as additional test source via `build-helper-maven-plugin` plugin.

And here used the `<activation>/<property>/<ranger.version>` to auto activated the `genpolicy` profile to ensure same ranger version used as required in `genpolicy` itself.

For the ENV suggestion, it's not that helpful as runtime switch to control whether generated the policy file as the generator is not runned in tests. Will introduce this switch if we do check the policy file as the same as generator outputs.

Therefore, the command here in method comments is updated as `build/mvn scala:run -pl :kyuubi-spark-authz_2.12 -DmainClass=org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator`.

Yikf (approved):

bowenliang123:
Thanks for the detailed review. Merged to master / 1.7.1 .

Merging to 1.7 is for the possible fix and testing related to policy file.

PR #3742, opened 2022-11-01
Add FileSessionConfAdvisor to manage session level configuration

ychris78:
### _Why are the changes needed?_

When a kyuubi service connections to multiple engine clusters, we can only explicitly specify relevant parameters after the URL, for example: 
`beeline -u 'jdbc:hive2://127.0.0.1:10009/?kyuubi.engineEnv.FLINK_HOME=/opt/flink;kyuubi.ha.namespace=kyuubi-ns-c;kyuubi.engine.type=FLINK_SQL'`
This method is extremely unfriendly to users.

I implemented a SessionConfAdvisor implementation class `org.apache.kyuubi.session.FileSessionConfAdvisor` based on the conf file that to manage session level configurations, so that we can access this cluster only by specifying a unique engine name, for example:
access cluster-a:
`beeline -u 'jdbc:hive2://127.0.0.1:10009/?kyuubi.session.conf.profile=cluster-a'`

access cluster-c:
`beeline -u 'jdbc:hive2://127.0.0.1:10009/?kyuubi.session.conf.profile=cluster-c'`

`kyuubi-session-cluster-a.conf` configuration is as follows:
```
    kyuubi.ha.namespace kyuubi-ns-a
    kyuubi.engine.type SPARK_SQL
    kyuubi.engine.pool.balance.policy POLLING

    kyuubi.engineEnv.SPARK_HOME /opt/spark30
    kyuubi.engineEnv.HADOOP_CONF_DIR /opt/hadoop_conf_dir
```

`kyuubi-session-cluster-c.conf` configuration is as follows:
```
    kyuubi.ha.namespace kyuubi-ns-a
    kyuubi.engine.type SPARK_SQL

    kyuubi.engineEnv.SPARK_HOME /opt/spark32
    kyuubi.engineEnv.HADOOP_CONF_DIR /opt/hadoop_conf_dir
```

### _How was this patch tested?_
- [x] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #4589, opened 2023-03-23
[HELM] Update template on port-foward usage guide NOTES.txt

phyyou:
- Fix Stale Docs URL in Helm Chart
- ~~Update kyuubiDefaults structure~~
- Update NOTES.txt
  - When kyuubi.kyuubiConf.kyuubiDefaults."kyuubi.frontend.bind.host" set localhost, render port-forward example  in NOTES.txt for development usage guide
- ~~Add extraFiles for config mount~~
  - ~~extraFiles for another file mount in kyuubi conf directory.~~

Based on https://github.com/apache/kyuubi/issues/4561#issuecomment-1481026954_

> I think Helm chart notes had been wrong from this commit.

>https://github.com/apache/kyuubi/blob/dbb741fc5b0d2fd4b0640ff70ea12aca75864166/charts/kyuubi/templates/NOTES.txt#L33-L34

> Can we delete this note or update this note  if `--address` option is work like this? :
> ```suggestion
>  - To access {{ $.Release.Name }}-{{ $name | kebabcase }} service from outside the cluster for debugging, run the following > > 
> command: 
>     kubectl port-forward svc/{{ $.Release.Name }}-{{ $name | kebabcase }} {{ tpl $frontend.service.port $ }}:{{ tpl >?> >$frontend.service.port $ }} -n {{ $.Release.Namespace }} --address <your-ip>
> ```

~~and more changes:~~
- ~~Update `.Values.kyuubiConf.kyuubiDefaults` to dict for more variant usage in templates. (like NOTES.txt changes)~~
- ~~Add extraFiles in configmap for like mounting hive-site-xml on kyuubi config directory.~~ 


zwangsheng:
Hi @phyyou, thanks for your PR.

I install your changes in local minikube cluster, i find kyuubi register self to zookeeper with `localhost:10009`, which may cause engine can't find kyuubi instance.

```log
2023-03-24 02:56:27.968 INFO org.apache.kyuubi.ha.client.zookeeper.ZookeeperDiscoveryClient: Created a /kyuubi/serviceUri=localhost:10009;version=1.7.0;sequence=0000000000 on ZooKeeper for KyuubiServer uri: localhost:10009
```

You can find the corresponding logic in `TFrontendService.scala#connectionUrl`.

pan3793:
> Update `.Values.kyuubiConf.kyuubiDefaults` to dict for more variant usage in templates. (like NOTES.txt changes)

I'm neutral on this change, @dnskr WDYT?

dnskr:
> > Update `.Values.kyuubiConf.kyuubiDefaults` to dict for more variant usage in templates. (like NOTES.txt changes)
> 
> I'm neutral on this change, @dnskr WDYT?

It's against recommended best practices: https://helm.sh/docs/chart_best_practices/values/#naming-conventions
It looks natural in this case, but it brings many inconveniences with setting values, for example:
1. Copy/Paste `a.b.c` to IntelliJ IDEA becomes
```yaml
  a:
    b:
      c:
```
2. Set property value in cli: https://helm.sh/docs/chart_best_practices/values/#consider-how-users-will-use-your-values
```bash
helm install kyuubi --set kyuubiConf.kyuubiDefaults."kyuubi\.frontend\.bind\.host"=localhost
```
3. Overwriting many default values.

I know only one mature chart which still uses similar property names, but looks like it was a mistake they did at the beginning: https://github.com/argoproj/argo-helm/blob/ccef4448748601bb253a89fa86ed036b0e620cc1/charts/argo-cd/values.yaml#L136

phyyou:
> It looks natural in this case, but it brings many inconveniences with setting values,

Thanks @dnskr. I definitely agree.
When I first thought of and created it, it seemed like a good idea, but when I actually tried Template Rendering with helm, and I realized it was inconvenient

I'd like to revert this, but I would like to think of a way to conveniently specify the configuration values in the Helm Chart.

Do you have any other good ideas for kyuubiConf in Helm Chart? @dnskr @pan3793

phyyou:
I was just inspired by the Helm Chart of Apache Airflow (https://github.com/apache/airflow/blob/main/chart/values.yaml#L1834-L1909).

What do you think about creating kyuubiDefaults with a similar structure? @dnskr

charts/kyuubi/values.yaml
```diff
@@ -101,17 +101,22 @@ server:
...
+  # See https://kyuubi.readthedocs.io/en/master/deployment/settings.html#kyuubi-configurations for more details
+  kyuubiDefaults: |
+
 
   # The value (templated string) is used for log4j2.xml file
-  # See https://kyuubi.apache.org/docs/latest/deployment/settings.html#logging for more details
+  # See https://kyuubi.readthedocs.io/en/master/deployment/settings.html#logging for more details
   log4j2: ~
 
+  extraFiles: {}
```
pan3793, line 116: it's valuable to change in a separate PR. let's focus on one thing in each PR, to make the future visitor easy to understand what we were doing.

phyyou: Thanks! got it!

pan3793:
> https://airflow.apache.org/docs/helm-chart/stable/parameters-ref.html

I like the idea of "examples", actually, I was asked, "I got no idea w/ a placeholder `~`, what should I assign to `kyuubiDefaults`?"

> ~
> ```
> Examples:
> 
> extraEnvFrom: |-
>   - secretRef:
>       name: '{{ .Release.Name }}-airflow-connections'
> extraEnvFrom: |-
>   - configMapRef:
>       name: '{{ .Release.Name }}-airflow-variables'
> ```

charts/kyuubi/values.yaml
dnskr: what is the use case for this property?

charts/kyuubi/values.yaml
```diff
@@ -101,17 +101,19 @@ server:
...
   # The value (templated string) is used for kyuubi-env.sh file
-  # See https://kyuubi.apache.org/docs/latest/deployment/settings.html#environments for more details
+  # See https://kyuubi.readthedocs.io/en/master/deployment/settings.html#environments for more details
   kyuubiEnv: ~
 
   # The value (templated string) is used for kyuubi-defaults.conf file
-  # See https://kyuubi.apache.org/docs/latest/deployment/settings.html#kyuubi-configurations for more details
-  kyuubiDefaults: ~
+  # See https://kyuubi.readthedocs.io/en/master/deployment/settings.html#kyuubi-configurations for more details
+  kyuubiDefaults: |
```
pan3793, line 109: it should be good to keep `~`, but let's add more comments, including "examples"

charts/kyuubi/values.yaml
```diff
@@ -101,17 +101,19 @@ server:
...
+
 
   # The value (templated string) is used for log4j2.xml file
-  # See https://kyuubi.apache.org/docs/latest/deployment/settings.html#logging for more details
+  # See https://kyuubi.readthedocs.io/en/master/deployment/settings.html#logging for more details
   log4j2: ~
 
+
```
pan3793, line 116: nit: unnecessary change

charts/kyuubi/values.yaml
pan3793: Users may use Kyuubi/Spark w/ Hive MetaStore, then `hive-site.xml` is required

charts/kyuubi/values.yaml
phyyou: Thanks!

charts/kyuubi/values.yaml
pan3793: And another use case is `FileSessionConfAdvisor` introduced in #3742, and emm... it isn't mentioned in docs, and we appreciate contributors who can help us to improvement the docs 😄

pan3793:
Do you have Kerberos use case? There are some extra work, like `keytab`, `krb5.conf`, to make the Helm chart work out-of-box

charts/kyuubi/values.yaml
dnskr: @pan3793  Would it be better to add specific properties instead? I.e. `hiveSite` etc

charts/kyuubi/values.yaml
phyyou: Thanks! I add Example usecases in kyuubiConf.

charts/kyuubi/values.yaml
pan3793: > @pan3793 Would it be better to add specific properties instead? I.e. `hiveSite` etc

I agree to use specific properties for specific files, but there are some non-enumerable files, e.g. in `FileSessionConfAdvisor` use case, there may have

- `kyuubi-session-etl.conf`
- `kyuubi-session-ad-hoc.conf`
- `kyuubi-session-admin.conf`
- `kyuubi-session-<...>.conf`

charts/kyuubi/values.yaml
```diff
@@ -101,15 +101,33 @@ server:
...
+  #
+  # kyuubi.frontend.bind.host=10.0.0.1
+  #
+  # kyuubi.engine.type=SPARK_SQL
+  # kyuubi.engine.share.level=USER
+  # kyuubi.session.engine.initialize.timeout=PT3M
+  #
+  # kyuubi.ha.addresses=zk1:2181,zk2:2181,zk3:2181
+  # kyuubi.ha.namespace=kyuubi
+  #
```
pan3793, line 125: I would suggest to including the "key" as well, user who not familiar w/ yaml will confusing how to handle `~`

```suggestion
  # Example:
  #
  #   kyuubiDefaults: |
  #     kyuubi.authentication=NONE
  #     kyuubi.frontend.bind.host=10.0.0.1
  #     kyuubi.engine.type=SPARK_SQL
  #     kyuubi.engine.share.level=USER
  #     kyuubi.session.engine.initialize.timeout=PT3M
  #     kyuubi.ha.addresses=zk1:2181,zk2:2181,zk3:2181
  #     kyuubi.ha.namespace=kyuubi
  #
```

phyyou: Got it! Thanks for suggestion.

dnskr:
> I just got a lot of inspiration Chart of Apache Airflow (https://github.com/apache/airflow/blob/main/chart/values.yaml#L1834-L1909).
> What do you think about kyuubiDefaults values with a similar structure? @dnskr

It might be a good idea, but managing many properties also might be exhausting. Also from my POV this requires stable configuration mechanism.

phyyou:
> It might be a good idea, but managing many properties also might be exhausting. Also from my POV this requires stable configuration mechanism.

I agree with the previous comment that managing a large number of properties can be quite challenging and exhausting, especially if there is no stable configuration mechanism in place.

I think It meens while it could be a good long-term goal to move towards a more structured approach. I do not think that it is a good solution at the moment. 

So, I think that while a structured approach like the Apache Airflow chart could be useful in the future, it is not a good solution at the moment, and we should focus on creating a stable configuration mechanism first.

phyyou:
> Do you have Kerberos use case? There are some extra work, like keytab, krb5.conf, to make the Helm chart work out-of-box

@pan3793

I don't have a complete understanding of the Kerberos use case, but I believe that we can refer to the configuration used in the Apache Airflow chart (https://github.com/apache/airflow/blob/main/chart/values.yaml#L413-L446) as a starting point.

So, I'd like to submit a new PR for krb5.conf with kyuubiConf.extraFiles.

charts/kyuubi/templates/NOTES.txt
```diff
@@ -30,9 +30,13 @@ In order to check the release status, use:
 {{ $name | snakecase | upper }}:
 - To access {{ $.Release.Name }}-{{ $name | kebabcase }} service within the cluster, use the following URL:
     {{ $.Release.Name }}-{{ $name | kebabcase }}.{{ $.Release.Namespace }}.svc.cluster.local
+{{- if $.Values.kyuubiConf.kyuubiDefaults }}
+{{- if or (contains "kyuubi.frontend.bind.host=localhost" $.Values.kyuubiConf.kyuubiDefaults ) (contains "kyuubi.frontend.bind.host=127.0.0.1" $.Values.kyuubiConf.kyuubiDefaults) }}
```
pan3793, line 34: should be ok, even it allows
- `kyuubi.frontend.bind.host <any spaces> localhost`
- `kyuubi.frontend.bind.host <any spaces> = <any spaces> localhost`

pan3793 (approved):
LGTM except for minor comments, thanks

pan3793:
@phyyou thanks for finding an example for kerberos, we can defer it if you don't have kerberos use case

charts/kyuubi/values.yaml
```diff
@@ -101,15 +101,30 @@ server:
...
+  # export JAVA_HOME=/usr/jdk64/jdk1.8.0_152
+  # export SPARK_HOME=/opt/spark
+  # export FLINK_HOME=/opt/flink
+  # export HIVE_HOME=/opt/hive
+  # See example at conf/kyuubi-env.sh.template and https://kyuubi.readthedocs.io/en/master/deployment/settings.html#environments for more details
   kyuubiEnv: ~
 
   # The value (templated string) is used for kyuubi-defaults.conf file
-  # See https://kyuubi.apache.org/docs/latest/deployment/settings.html#kyuubi-configurations for more details
+  # Example:
```
pan3793, line 113: nit: new line before `Example:`

charts/kyuubi/values.yaml
```diff
@@ -101,15 +101,30 @@ server:
 kyuubiConfDir: /opt/kyuubi/conf
 kyuubiConf:
   # The value (templated string) is used for kyuubi-env.sh file
-  # See https://kyuubi.apache.org/docs/latest/deployment/settings.html#environments for more details
+  # ## Examples ##
```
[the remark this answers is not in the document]
phyyou: Thanks!

charts/kyuubi/templates/NOTES.txt
phyyou: Got it.
I updated the Helm template to use a single regexMatch function to match localhost or 127.0.0.1 with any number of spaces.

charts/kyuubi/values.yaml
phyyou: Thanks

pan3793 (approved):

phyyou:
@pan3793

Sorry! It has erorr with regex.

I'm fixing it now.

phyyou:
@pan3793 

I'm sorry for the frequent mentions,

I fixed the when there was no equal sign now.

zwangsheng:
> Sorry for the late response. I have also experienced this error before. Specifically, I encountered this error when set `kyuubi.frontend.connection.url.use.hostname=false.`, but, it may require a deeper investigation to identify the root cause.

Well, seems we revert  set `kyuubi.frontend.bind.host=localhost,` which cause the problem.

LGTM, thanks.

zwangsheng (approved):

pan3793:
Thanks all, merged to master/1.7

phyyou:
@zwangsheng @pan3793 

Since [this commit](github.com/apache/kyuubi/commit/dbb741fc5b0d2fd4b0640ff70ea12aca75864166), 
`kyuubi.frontend.connection.url.use.hostname=false` is default value for helm chart.

In my opinion, it would be better not to use local Zookeeper in the Helm chart. For development environments.
I think it seems like a good idea to have an extra deployment option to turn on and off the Zookeeper deployment in the Helm chart.

pan3793:
> For development environments, I think it seems like a good idea to have an extra deployment option to turn on and off the Zookeeper deployment in the Helm chart.

SGTM.

charts/kyuubi/README.md
```diff
@@ -0,0 +1,43 @@
...
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ -->
+
+# Helm Chart for Apache Kyuubi
+
+[Apache Kyuubi](https://airflow.apache.org/) is a distributed and multi-tenant gateway to provide serverless SQL on Data Warehouses and Lakehouses.
```
yaooqinn, line 22: There is a typo here https://airflow.apache.org/ cc @phyyou @zwangsheng @pan3793 @dnskr

phyyou: Thanks

PR #4592, opened 2023-03-24
Support eagerly kill redundant executors

ulysses-you:

This pr adds a new rule `FinalStageResourceManager` to eagerly kill redundant executors 

We first get the final stage partition which is the actually required cores, then kill the redundant executors. The priority of kill executors follow:
  1. kill executor who is younger than other (The older the JIT works better)
  2. kill executor who produces less shuffle data first 

The reason why add this feature is that, if the previous stage contains lots executors but final stage has less, then the tasks of final stage would be scheduled randomly in all exists executors which may cause resource waste. e.g., each executor only run 1 or 2 tasks but holds 4 or 5 cores.

test manually

- test for the kill executor
image


ulysses-you:
cc @yaooqinn @pan3793 @bowenliang123 @cfmcgrady

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -190,4 +190,20 @@ object KyuubiSQLConf {
       .version("1.7.0")
       .booleanConf
       .createWithDefault(true)
+
+  val FINAL_WRITE_STAGE_EAGERLY_KILL_EXECUTORS_ENABLED =
+    buildConf("spark.sql.finalWriteStageEagerlyKillExecutors.enabled")
```
pan3793, line 195: it's valuable to introduce a new namespace `spark.sql.finalWriteStage.`
```suggestion
    buildConf("spark.sql.finalWriteStage.eagerlyKillExecutors.enabled")
```

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/spark/FinalStageResourceManager.scala
```diff
@@ -0,0 +1,199 @@
...
+      case (latest, rule) => rule.apply(latest)
+    }
+    shuffleRead match {
+      case AQEShuffleReadExec(stage: ShuffleQueryStageExec, partitionSpecs) =>
+        // The condition whether inject custom resource profile:
+        // - target executors < active executors
+        // - target executors > min executors
+        val numActiveExecutors = sc.getExecutorIds().length
+        val expectedCores = partitionSpecs.length
+        val targetExecutors = (((expectedCores / executorCores) + 1) * factor).toInt
```
pan3793, line 80: `(expectedCores / executorCores) + 1` =>  `math.ceil(expectedCores.toFloat / executorCores)`

yaooqinn, line 80: Keep at least 1 exec?

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/spark/FinalStageResourceManager.scala
```diff
@@ -0,0 +1,199 @@
...
+    executorToBlockSize.toSeq.sortBy(_._2).foreach { case (id, _) =>
+      if (executorIdsToKill.length < expectedNumExecutorToKill) {
+        executorIdsToKill.append(id)
+      }
+    }
+
+    executorIdsToKill.toSeq
+  }
+
+  private def killExecutors(
```
yaooqinn, line 159: org.apache.spark.SparkContext#killExecutors?

yaooqinn:
> kill executor who is younger than other (The older the JIT works better)

This is not always true, 
- the generated code may not JITed in the final stage.
- The old ones are very like to overloaded in some cases
- Potential Problems happen with shuffle tracking?

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/spark/FinalStageResourceManager.scala
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * This rule assumes the final write stage has less cores requirement than previous, otherwise
+ * this rule would take no effect.
+ *
+ * It provide a feature:
+ * 1. Kill redundant executors before running final write stage
+ */
+case class FinalStageResourceManager(session: SparkSession) extends Rule[SparkPlan] {
+  override def apply(plan: SparkPlan): SparkPlan = {
+    if (!conf.getConf(KyuubiSQLConf.FINAL_WRITE_STAGE_EAGERLY_KILL_EXECUTORS_ENABLED)) {
```
yaooqinn, line 42: dynamicAllocation enabled?

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/spark/FinalStageResourceManager.scala
ulysses-you: There is a story about DRA. Since https://github.com/apache/spark/pull/20604, `org.apache.spark.SparkContext#killExecutors` does not allow with DRA ON, so this pr hack the internal interface to kill executors. I think that pr is not very reaonable, it should be ok to kill executors with DRA ON if the min executor is less than the target executor.

ulysses-you:
@yaooqinn 

> the generated code may not JITed in the final stage.

yea, but at least, the JIT should work better for shuffle read and scheduler related code with the older one 

> the old ones are very like to overloaded in some cases

it really depend on the machine .. I'm not sure how to find the bad machine in a easy way

> Potential Problems happen with shuffle tracking?

We first kill executor according to the alive time and exclude the one who has shuffle data, so it should be safe for this case.

yaooqinn (approved):

ulysses-you:
thanks for review, merging to master !

PR #4594, opened 2023-03-24
Support PUT method in REST client

turboFei:

We have api that need `PUT` method.
image


ulysses-you:
should the docs you reference be `delete` ?

turboFei:
> should the docs you reference be delete ?

now the supported actions are `close` or `cancel`.

But I think that maybe we can support more actions. so, `put` should be better.

bowenliang123 (approved):
LGTM. PUT requests are idempotent and can be used for different purposes.

ulysses-you:
I'm fine with this, as it's a reserved client api to be compatible with future server feature.

pan3793 (approved):

turboFei:
thanks, merged to master

PR #4605, opened 2023-03-25
[K8S][HELM] Set IfNotPresent pullPolicy by default

dnskr:
The change is needed to avoid pulling immutable image `apache/kyuubi:1.7.0` from image registry each time pod created.
Current `pullPolicy: Always` was important when mutable image `apache/kyuubi:master-snapshot` was used.


pan3793 (approved):

pan3793:
Thanks, merged to master/1.7

PR #4606, opened 2023-03-25
[K8S][HELM] Add command and args configuration support

dnskr:
The change allows to change default command to launch Kyuubi and pass additional command line arguments.
It's needed if the chart user wants to print message, create files or similar in main container (`kyuubi-server`) when pod starts.

### DIFF
--- a/charts/kyuubi/templates/kyuubi-deployment.yaml
+++ b/charts/kyuubi/templates/kyuubi-deployment.yaml
@@ -50,6 +50,12 @@ spec:
         - name: kyuubi-server
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- with .Values.command }}
+          command: {{- tpl (toYaml .) $ | nindent 12 }}
+          {{- end }}
+          {{- with .Values.args }}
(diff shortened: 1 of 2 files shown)


pan3793:
I'm not sure it's a good practice to allow user custom the start command, any examples?

dnskr:
> I'm not sure it's a good practice to allow user customize the start command, any examples?

I think it is not a good practice for production usage, but very helpful for investigation, development and workarounds.

For example
- Airflow allows to specify custom command and args [airflow/chart/values.yaml#L455-L463](https://github.com/apache/airflow/blob/1f8bf783b89d440ecb3e6db536c63ff324d9fc62/chart/values.yaml#L455-L463). I used it to create task pools on startup and to install additional python packages without building custom docker image.
- Superset also has possibility to override command: [superset/helm/superset/values.yaml#L254-L257](https://github.com/apache/superset/blob/0fa421271e874e456ae71f7aca4c71130176332d/helm/superset/values.yaml#L254-L257)
- One more use case https://github.com/apache/kyuubi/issues/4515#issuecomment-1484080124

So it is not critical change, however gives more flexibility to the helm chart.

pan3793 (approved):

pan3793:
Thanks @dnskr for listing usage examples, it makes sense then.

PR #4613, opened 2023-03-26
Updated kyuubi-configmap.yaml

PasunuriSrinidhi:

In this example, I've added the following fields to the metadata section:

namespace: I've added the namespace where the ConfigMap should be created.

labels: I've added a custom label my-custom-label with the value my-custom-value.

annotations: I've added two custom annotations my-annotation and my-other-annotation with their respective values.

### DIFF
--- a/charts/kyuubi/templates/kyuubi-configmap.yaml
+++ b/charts/kyuubi/templates/kyuubi-configmap.yaml
@@ -19,12 +19,26 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
   labels:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     app.kubernetes.io/name: {{ .Chart.Name }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | quote }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+    my-custom-label: my-custom-value   # adding a custom label
+  annotations:
+    my-annotation: some-value         # adding a custom annotation
+    my-other-annotation: another-value
+  ownerReferences:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: {{ .Release.Name }}-deploy
+      uid: 12345678-90ab-cdef-0123-456789abcdef
+  finalizers:
+    - my-finalizer.example.com
+  resourceVersion: "123456"
+  selfLink: /api/v1/namespaces/{{ .Release.Namespace }}/configmaps/{{ .Release.Name }}
 data:
   {{- with .Values.kyuubiConf.kyuubiEnv }}
   kyuubi-env.sh: |


charts/kyuubi/templates/kyuubi-configmap.yaml
```diff
@@ -19,12 +19,26 @@ apiVersion: v1
...
 metadata:
   name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
   labels:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     app.kubernetes.io/name: {{ .Chart.Name }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | quote }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+    my-custom-label: my-custom-value   # adding a custom label
```
pan3793, line 29: the example stuffs should not be added directly, if you want, please define variables in values.yaml, and use template to inject them.

charts/kyuubi/templates/kyuubi-configmap.yaml
```diff
@@ -19,12 +19,26 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
```
pan3793, line 22: it looks correct to me, cc @dnskr

dnskr: It is correct, but not required because the namespace of helm release is used by default. I mean I don't see any reasons or values by adding `namespace: {{ .Release.Namespace }}`.

dnskr:
@PasunuriSrinidhi I'm not sure I understand the changes. Could you please explain it?

charts/kyuubi/templates/kyuubi-configmap.yaml
PasunuriSrinidhi: Okay I will define this variable, and  then use it in the  templates by referencing it .

charts/kyuubi/templates/kyuubi-configmap.yaml
pan3793: So, it's redundant, let's just keep it as-is to make thing simple.

PasunuriSrinidhi:
> @PasunuriSrinidhi I'm not sure I understand the changes. Could you please explain it?
The use of the namespace {{ .Release.Namespace }} in a Helm chart template file allows you to dynamically reference the Kubernetes namespace in which the chart is installed.

When a Helm chart is installed, it is deployed to a Kubernetes namespace, which is a logical boundary for resources in a Kubernetes cluster. The namespace isolates resources within it from other namespaces, allowing multiple teams to work on the same Kubernetes cluster without interfering with each other's resources.

By using {{ .Release.Namespace }} in a template file, you can reference the namespace in which the chart is deployed without hardcoding the namespace name in the template. This makes the chart more portable and reusable across different Kubernetes clusters with different namespace names.

For example, you might use {{ .Release.Namespace }} in a Kubernetes deployment template to specify the namespace in which to deploy a container:

charts/kyuubi/templates/kyuubi-configmap.yaml
PasunuriSrinidhi: okay then we will keep it simple

dnskr:
@PasunuriSrinidhi The following command deploys the chart to desired namespace:
```shell
helm install kyuubi charts/kyuubi -n $desired_namespace
```

dnskr:
Thanks for the PR! This PR is being closed due to inactivity. This isn't a judgement on the merit of the PR in any way. If this is still an issue with the latest version of Kyuubi, please reopen it and ask a committer to remove the Stale tag!

Thank you for using Kyuubi!

PR #4619, opened 2023-03-27
Fix beeline with -e When there are other SQL statements before the source statement, the source statement cannot be executed normally

thomasg19930417:

When using beeline -e, if there is a newline or space between the source statement and the previous sql, the source statement will be misjudged as a sql statement and sent to the server for execution
image

image


kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiCommands.java
```diff
@@ -54,12 +54,14 @@ private String[] tokenizeCmd(String cmd) {
...
 
   private boolean isSourceCMD(String cmd) {
+    cmd = cmd.trim();
     if (cmd == null || cmd.isEmpty()) return false;
     String[] tokens = tokenizeCmd(cmd);
     return tokens[0].equalsIgnoreCase("source");
   }
 
   private boolean sourceFile(String cmd) {
+    cmd = cmd.trim();
```
turboFei, line 64: Does it impact Python code？

@cfmcgrady

turboFei:
Cc @cfmcgrady

turboFei:
thanks for your contribution @thomasg19930417 , merged to master and 1.7.1

Issue #3872, opened 2022-11-29
[Bug] SparkBatchProcessBuilder should not set proxy user when keytab is used


wangyepeng2:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug
<img width="930" alt="企业微信截图_0b50c8f3-bccd-4c4a-b233-a62646341cb0" src="https://user-images.githubusercontent.com/44561824/204458501-32e4919b-ecc5-48ae-b4b1-0680899c88b1.png">

The above code is in class SparkBatchProcessBuilder .spark batch process will always set proxy user , even if i set the spark.kerberos.principal. it will cause error:
```
Only one of --proxy-user or --principal can be provided.
        at org.apache.spark.deploy.SparkSubmitArguments.error(SparkSubmitArguments.scala:634)
        at org.apache.spark.deploy.SparkSubmitArguments.validateSubmitArguments(SparkSubmitArguments.scala:279)
        at org.apache.spark.deploy.SparkSubmitArguments.validateArguments(SparkSubmitArguments.scala:234)
        at org.apache.spark.deploy.SparkSubmitArguments.<init>(SparkSubmitArguments.scala:119)
        at org.apache.spark.deploy.SparkSubmit$$anon$2$$anon$3.<init>(SparkSubmit.scala:1029)
        at org.apache.spark.deploy.SparkSubmit$$anon$2.parseArguments(SparkSubmit.scala:1029)
        at org.apache.spark.deploy.SparkSubmit.doSubmit(SparkSubmit.scala:85)
        at org.apache.spark.deploy.SparkSubmit$$anon$2.doSubmit(SparkSubmit.scala:1046)
        at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:1055)
        at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
```
### Affects Version(s)

master

### Kyuubi Engine Log Output

```logtalk
Only one of --proxy-user or --principal can be provided.
        at org.apache.spark.deploy.SparkSubmitArguments.error(SparkSubmitArguments.scala:634)
        at org.apache.spark.deploy.SparkSubmitArguments.validateSubmitArguments(SparkSubmitArguments.scala:279)
[...]

PR #4640, opened 2023-03-30
[KYUUBI #3872] fix:spark batch submit should not set proxy user when keytab is used

wangyepeng2:
see detail description in issue [KYUUBI #3872]


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkBatchProcessBuilder.scala
```diff
@@ -56,9 +56,7 @@ class SparkBatchProcessBuilder(
       buffer += s"${convertConfigKey(k)}=$v"
     }
 
-    setSparkUserName(proxyUser, buffer)
-    buffer += PROXY_USER
-    buffer += proxyUser
+    setUser(buffer)
```
pan3793, line 59:
```suggestion
    setupKerberos(buffer)
```

pan3793:
Thanks, merged to master/1.7

PR #4642, opened 2023-03-31
[KYUUBI #4641] Add MaxFileSizeStrategy to limit max scan file size

### LINKED ISSUE #4641: [FEATURE] Support limit maximum scan file size
### Code of Conduct

### Search before asking

### Describe the feature

I want to reject large queries by limiting the size of scanned files.

### Motivation

When using a shared engine, large queries take up more resources and block other queries.

### Describe the solution

Refer to the implementation of `maxPartitions` in watchdog

### Additional context

_No response_

wForget:

Add MaxFileSizeStrategy to limit max scan file size.
close #4641 


wForget:
> I'm wondering if we can unify the `MaxPartitionStrategy` and `MaxFileSizeStrategy`. They looks like depend on the similar plan pattern.

Sounds good, I will revise it later.

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxScanStrategy.scala
```diff
@@ -0,0 +1,293 @@
...
+              _)) =>
+        if (fsRelation.partitionSchema.nonEmpty) {
+          val (partitionKeyFilters, dataFilter) =
+            getPartitionKeyFiltersAndDataFilters(
+              fsRelation.sparkSession,
+              relation,
+              partitionSchema,
+              filters,
+              relation.output)
+          val prunedPartitions = fileIndex.listFiles(
```
ulysses-you, line 139: we should do list files when `maxScanPartitions > 0`, otherwise it would be the overhead

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxScanStrategy.scala
```diff
@@ -0,0 +1,293 @@
...
+ * This Strategy Add Planner Strategy after LogicalOptimizer
+ * @param session
+ */
+case class MaxScanStrategy(session: SparkSession)
+  extends Strategy
+  with SQLConfHelper
+  with PruneFileSourcePartitionHelper {
+  override def apply(plan: LogicalPlan): Seq[SparkPlan] = {
+    val maxScanPartitions = conf.getConf(KyuubiSQLConf.WATCHDOG_MAX_PARTITIONS).getOrElse(0)
+    val maxFileSize = conf.getConf(KyuubiSQLConf.WATCHDOG_MAX_FILE_SIZE).getOrElse(0L)
```
ulysses-you, line 45: it seems a optional value is better, and call `expectValue.exists(_ <= actualValue)`

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxScanStrategy.scala
```diff
@@ -0,0 +1,293 @@
...
+                     |Table: ${relation.tableMeta.qualifiedName}
+                     |Owner: ${relation.tableMeta.owner}
+                     |Partition Structure: ${relation.partitionCols.map(_.name).mkString(", ")}
+                     |""".stripMargin)
+              }
+              lazy val scanFileSize = totalPartitions.flatMap(_.stats).map(_.sizeInBytes).sum
+              if (maxFileSize > 0 && scanFileSize > maxFileSize) {
+                throw new MaxFileSizeExceedException(
+                  s"""
+                     |Your SQL job scan a whole huge table without any partition filter,
```
ulysses-you, line 99: shall we change the hint of exception ? user may have a valid partition filter, but the file size is still large enough

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxScanStrategy.scala
```diff
@@ -0,0 +1,293 @@
...
+                     |SQL job scan hive partition: ${prunedPartitions.size}
+                     |exceed restrict of hive scan maxPartition $maxScanPartitions
+                     |You should optimize your SQL logical according partition structure
+                     |or shorten query scope such as p_date, detail as below:
+                     |Table: ${relation.tableMeta.qualifiedName}
+                     |Owner: ${relation.tableMeta.owner}
+                     |Partition Structure: ${relation.partitionCols.map(_.name).mkString(", ")}
+                     |""".stripMargin)
+              }
+              lazy val scanFileSize = prunedPartitions.flatMap(_.stats).map(_.sizeInBytes).sum
```
pan3793, line 70: is the stats always accurate?

wForget: > is the stats always accurate?

It is not always accurate. 

The partition `totalSize` will be automatically calculated during metastore operations. But, for partitioned table or change without metadata operation, the statistic may not be accurate.

refer: https://cwiki.apache.org/confluence/display/hive/statsdev#StatsDev-Quickoverview

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxScanStrategy.scala
wForget: > user may have a valid partition filter, but the file size is still large enough

Here matches the table without partition filtering.

```
relation.prunedPartitions match {
  ...
  case _ >  // This means the prunedPartitions is None
```

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxScanStrategy.scala
```diff
@@ -0,0 +1,296 @@
...
+                throw partTableMaxFileExceedError(
+                  scanFileSize,
+                  maxFileSizeOpt.get,
+                  Some(relation.tableMeta),
+                  prunedPartitions.flatMap(_.storage.locationUri).map(_.toString),
+                  relation.partitionCols.map(_.name))
+              }
+            case _ =>
+              val totalPartitions = session
+                .sessionState.catalog.externalCatalog.listPartitions(
```
ulysses-you, line 84: I'm afraid it would introduce overhead, to be safe, how about using `listPartitionNames` ? If the acutal num partitions match maxNumPartition, then we can use `listPartitions` to check maxFileSize.

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/watchdog/MaxScanStrategy.scala
```diff
@@ -0,0 +1,309 @@
...
+
+              def scanFileSize: BigInt = {
+                val partitions = session
+                  .sessionState.catalog.externalCatalog.listPartitions(
+                    relation.tableMeta.database,
+                    relation.tableMeta.identifier.table)
+                partitionSize = Some(partitions.size)
+                partitions.flatMap(_.stats).map(_.sizeInBytes).sum
+              }
+              if (maxFileSizeOpt.exists(_ < scanFileSize)) {
```
ulysses-you, line 93: shall we check `maxScanPartitions` first , then `maxFileSize` ?

wForget: > shall we check `maxScanPartitions` first , then `maxFileSize` ?

If `scanPartitions` is calculated first and the limit is not exceeded, `listPartitions` method needs to be called again. I think the limit will not be exceeded most of the time, so calculating `scanFileSize` first will save one `listPartitions` call.

ulysses-you: we check scanPartitions by `listPartitionNames`, and to be safe we should use `listPartitionNames` first..

wForget: > we check scanPartitions by `listPartitionNames`, and to be safe we should use `listPartitionNames` first..

Got it, I will adjust it.

wForget:
This failure seems to be caused by #4714.  cc @huangzhir 

https://github.com/apache/kyuubi/actions/runs/4754456524/jobs/8447299902?pr=4642

```
SchedulerPoolSuite:
- Scheduler pool *** FAILED ***
  1681996257893 was not less than 0 (SchedulerPoolSuite.scala:108)
```

huangzhir:
> https://github.com/apache/kyuubi/actions/runs/4754456524/jobs/8447299902?pr=4642

sorry for causing trouble. The reason for this problem is that the listener's monitoring of the job is asynchronous. Job 1 has ended, but the listener has not yet received the message that Job 1 has ended, so the judgment was made prematurely.

@pan3793 Should I create a new pull request for the changes, or modify the existing one that has already been merged?

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Issue #4530, opened 2023-03-15
The kyuubi data masking module does not support Chinese for strategies such as MASK_SHOW_FIRST_4 or MASK_SHOW_FIRST_4


Jackhjf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

kyuubi authz version : master/1.8.0
ranger version 2.2.0
spark version 3.1.2
iceberg version 0.12.0

I configure the following data masking strategy on ranger

<img width="1574" alt="image" src="https://user-images.githubusercontent.com/33471639/225354868-d751ab49-275b-4691-a635-b73cd58c33b0.png">

it does not work for Chinese
<img width="1102" alt="image" src="https://user-images.githubusercontent.com/33471639/225355111-0a40ff86-5203-408a-b403-ec3fa3eea297.png">

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #4643, opened 2023-03-31
[KYUUBI #4530]  [AUTHZ] Support non-English chars for MASK, MASK_SHOW_FIRST_4, and MASK_SHOW_FIRST_4

huangzhir:
To fix https://github.com/apache/kyuubi/issues/4530.
1. The reason for issue https://github.com/apache/kyuubi/issues/4530  is that MASK_SHOW_FIRST_4 and MASK_SHOW_LAST_4 mask types are currently implemented using the regexp_replace method, which only replaces English letters and digits, but ignores other languages, such as Chinese.
2. To fix this issue, I modified the regexp_replace method to replace no-english characters to 'U' letters, so they will also be masked properly.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
@@ -133,7 +133,7 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
 
   private def regexp_replace(expr: String, hasLen: Boolean = false): String = {
     val pos = if (hasLen) ", 5" else ""
-    val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
+    val upper = s"regexp_replace($expr, '[A-Z\u4e00-\u9fa5]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
     val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
     digits
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
@@ -131,12 +131,14 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     }
   }
 
-  private def regexp_replace(expr: String, hasLen: Boolean = false): String = {
+  // public for test
+  def regexp_replace(expr: String, hasLen: Boolean = false): String = {
     val pos = if (hasLen) ", 5" else ""
-    val upper = s"regexp_replace($expr, '[A-Z\u4e00-\u9fa5]', 'X'$pos)"
+    val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
     val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
-    digits
+    val chinese = s"regexp_replace($digits, '[\u4e00-\u9fff]', '\u5bc6'$pos)"
+    chinese
   }
 
   /**
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
@@ -136,7 +136,7 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
     val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
-    val other = s"regexp_replace($digits, '[^A-Za-z0-9 ]', 'U'$pos)"
+    val other = s"regexp_replace($digits, '[^A-Za-z0-9]', 'U'$pos)"
     other
   }
 
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
@@ -109,8 +109,14 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
       } else if (result.getMaskTypeDef != null) {
         result.getMaskTypeDef.getName match {
           case "MASK" => regexp_replace(col)
-          case "MASK_SHOW_FIRST_4" => maskShowFirst4(col, isSparkVersionAtLeast("3.1"))
-          case "MASK_SHOW_LAST_4" => maskShowLast4(col)
+          case "MASK_SHOW_FIRST_4" if isSparkVersionAtLeast("3.1") =>
+            regexp_replace(col, hasLen = true)
+          case "MASK_SHOW_FIRST_4" =>
+            val right = regexp_replace(s"substr($col, 5)")
+            s"concat(substr($col, 0, 4), $right)"
+          case "MASK_SHOW_LAST_4" =>
+            val left = regexp_replace(s"left($col, length($col) - 4)")
+            s"concat($left, right($col, 4))"
           case "MASK_HASH" => s"md5(cast($col as string))"
           case "MASK_DATE_SHOW_YEAR" => s"date_trunc('YEAR', $col)"
           case _ => result.getMaskTypeDef.getTransformer match {
@@ -125,28 +131,13 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     }
   }
 
-  // public for test
-  def regexp_replace(expr: String, hasLen: Boolean = false): String = {
+  private def regexp_replace(expr: String, hasLen: Boolean = false): String = {
     val pos = if (hasLen) ", 5" else ""
     val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
     val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
     val other = s"regexp_replace($digits, '[^A-Za-z0-9 ]', 'U'$pos)"
     other
-  }
-
-  def maskShowFirst4(col: String, isSparkV31OrGreater: Boolean): String = {
-    if (isSparkV31OrGreater) {
-      regexp_replace(col, hasLen = true)
-    } else {
-      val right = regexp_replace(s"substr($col, 5)")
-      s"concat(substr($col, 0, 4), $right)"
-    }
-  }
-
-  def maskShowLast4(col: String): String = {
-    val left = regexp_replace(s"left($col, length($col) - 4)")
-    s"concat($left, right($col, 4))"
   }
 
   /**
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
@@ -109,14 +109,8 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
       } else if (result.getMaskTypeDef != null) {
         result.getMaskTypeDef.getName match {
           case "MASK" => regexp_replace(col)
-          case "MASK_SHOW_FIRST_4" if isSparkVersionAtLeast("3.1") =>
-            regexp_replace(col, hasLen = true)
-          case "MASK_SHOW_FIRST_4" =>
-            val right = regexp_replace(s"substr($col, 5)")
-            s"concat(substr($col, 0, 4), $right)"
-          case "MASK_SHOW_LAST_4" =>
-            val left = regexp_replace(s"left($col, length($col) - 4)")
-            s"concat($left, right($col, 4))"
+          case "MASK_SHOW_FIRST_4" => maskShowFirst4(col, isSparkVersionAtLeast("3.1"))
+          case "MASK_SHOW_LAST_4" => maskShowLast4(col)
           case "MASK_HASH" => s"md5(cast($col as string))"
           case "MASK_DATE_SHOW_YEAR" => s"date_trunc('YEAR', $col)"
           case _ => result.getMaskTypeDef.getTransformer match {
@@ -137,8 +131,22 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
     val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
-    val chinese = s"regexp_replace($digits, '[\u4e00-\u9fff]', '\u5bc6'$pos)"
-    chinese
+    val other = s"regexp_replace($digits, '[^A-Za-z0-9 ]', 'U'$pos)"
+    other
+  }
+
+  def maskShowFirst4(col: String, isSparkV31OrGreater: Boolean): String = {
+    if (isSparkV31OrGreater) {
+      regexp_replace(col, hasLen = true)
+    } else {
+      val right = regexp_replace(s"substr($col, 5)")
+      s"concat(substr($col, 0, 4), $right)"
+    }
+  }
+
+  def maskShowLast4(col: String): String = {
+    val left = regexp_replace(s"left($col, length($col) - 4)")
+    s"concat($left, right($col, 4))"
   }
 
   /**
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
@@ -50,11 +50,15 @@ class SparkRangerAdminPluginSuite extends AnyFunSuite {
     }
     assert(getMaskingExpr(buildAccessRequest(bob, "value1")).get === "md5(cast(value1 as string))")
     assert(getMaskingExpr(buildAccessRequest(bob, "value2")).get ===
-      "regexp_replace(regexp_replace(regexp_replace(value2, '[A-Z]', 'X'), '[a-z]', 'x')," +
-      " '[0-9]', 'n')")
+      "regexp_replace(regexp_replace(regexp_replace(regexp_replace(value2, '[A-Z]', 'X')," +
+        " '[a-z]', 'x'), '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U')")
     assert(getMaskingExpr(buildAccessRequest(bob, "value3")).get contains "regexp_replace")
     assert(getMaskingExpr(buildAccessRequest(bob, "value4")).get === "date_trunc('YEAR', value4)")
-    assert(getMaskingExpr(buildAccessRequest(bob, "value5")).get contains "regexp_replace")
+    assert(getMaskingExpr(buildAccessRequest(bob, "value5")).get ===
+      "concat(regexp_replace(regexp_replace(regexp_replace(regexp_replace(" +
+        "left(value5, length(value5) - 4), '[A-Z]', 'X'), '[a-z]', 'x')" +
+        ", '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U'), right(value5, 4))"
+    )
 
     Seq("admin", "alice").foreach { user =>
       val ugi = UserGroupInformation.createRemoteUser(user)
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
@@ -51,14 +51,13 @@ class SparkRangerAdminPluginSuite extends AnyFunSuite {
     assert(getMaskingExpr(buildAccessRequest(bob, "value1")).get === "md5(cast(value1 as string))")
     assert(getMaskingExpr(buildAccessRequest(bob, "value2")).get ===
       "regexp_replace(regexp_replace(regexp_replace(regexp_replace(value2, '[A-Z]', 'X')," +
-      " '[a-z]', 'x'), '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U')")
+      " '[a-z]', 'x'), '[0-9]', 'n'), '[^A-Za-z0-9]', 'U')")
     assert(getMaskingExpr(buildAccessRequest(bob, "value3")).get contains "regexp_replace")
     assert(getMaskingExpr(buildAccessRequest(bob, "value4")).get === "date_trunc('YEAR', value4)")
     assert(getMaskingExpr(buildAccessRequest(bob, "value5")).get ===
       "concat(regexp_replace(regexp_replace(regexp_replace(regexp_replace(" +
       "left(value5, length(value5) - 4), '[A-Z]', 'X'), '[a-z]', 'x')," +
-      " '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U'), right(value5, 4))"
-    )
+      " '[0-9]', 'n'), '[^A-Za-z0-9]', 'U'), right(value5, 4))")
 
     Seq("admin", "alice").foreach { user =>
       val ugi = UserGroupInformation.createRemoteUser(user)
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
@@ -51,13 +51,13 @@ class SparkRangerAdminPluginSuite extends AnyFunSuite {
     assert(getMaskingExpr(buildAccessRequest(bob, "value1")).get === "md5(cast(value1 as string))")
     assert(getMaskingExpr(buildAccessRequest(bob, "value2")).get ===
       "regexp_replace(regexp_replace(regexp_replace(regexp_replace(value2, '[A-Z]', 'X')," +
-        " '[a-z]', 'x'), '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U')")
+      " '[a-z]', 'x'), '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U')")
     assert(getMaskingExpr(buildAccessRequest(bob, "value3")).get contains "regexp_replace")
     assert(getMaskingExpr(buildAccessRequest(bob, "value4")).get === "date_trunc('YEAR', value4)")
     assert(getMaskingExpr(buildAccessRequest(bob, "value5")).get ===
       "concat(regexp_replace(regexp_replace(regexp_replace(regexp_replace(" +
-        "left(value5, length(value5) - 4), '[A-Z]', 'X'), '[a-z]', 'x')" +
-        ", '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U'), right(value5, 4))"
+      "left(value5, length(value5) - 4), '[A-Z]', 'X'), '[a-z]', 'x')," +
+      " '[0-9]', 'n'), '[^A-Za-z0-9 ]', 'U'), right(value5, 4))"
     )
 
     Seq("admin", "alice").foreach { user =>
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
@@ -128,7 +128,7 @@ trait DataMaskingTestBase extends AnyFunSuite with SparkSessionProvider with Bef
       "bob",
       "SELECT coalesce(max(value1), 1), coalesce(max(value2), 1), coalesce(max(value3), 1), " +
         "coalesce(max(value4), timestamp '2018-01-01 22:33:44'), coalesce(max(value5), 1) " +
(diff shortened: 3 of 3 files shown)


yaooqinn:
cc @bowenliang123 @pan3793

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
```diff
@@ -90,7 +90,7 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
 
   private def regexp_replace(expr: String, hasLen: Boolean = false): String = {
     val pos = if (hasLen) ", 5" else ""
-    val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
+    val upper = s"regexp_replace($expr, '[A-Z\u4e00-\u9fa5]', 'X'$pos)"
```
bowenliang123, line 93: Any reference for the starting and ending unicode for Chinese chars?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
```diff
@@ -90,7 +90,7 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
 
   private def regexp_replace(expr: String, hasLen: Boolean = false): String = {
     val pos = if (hasLen) ", 5" else ""
-    val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
```
bowenliang123, line 93: Chinese chars should not be part of upper chars. Use a separate replacer for it.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
```diff
@@ -90,7 +90,7 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
 
   private def regexp_replace(expr: String, hasLen: Boolean = false): String = {
```
bowenliang123, line 91: Use a separate ut cases for testing this `regexp_replace` method.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -388,6 +389,19 @@ abstract class RangerSparkExtensionSuite extends AnyFunSuite
           assert(sql(s"SELECT value1 FROM $db.${table}2").collect() ===
             Seq(Row(DigestUtils.md5Hex("1"))))
         })
+
+      doAs(
+        "admin",
+        sql(
+          s"INSERT INTO $db.$table SELECT 10, 1, 'hello', " +
+            s"'\u6d4b\u8bd5\u4e2d\u6587\u524d\u56db\u4e2aabcABC', " +
```
bowenliang123, line 397: Add comments to explain what and why to use this unicodes in tests. Source code is in UTF-8 encoding and good to have Chinese chars.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -342,6 +342,7 @@ abstract class RangerSparkExtensionSuite extends AnyFunSuite
         sql(
           s"INSERT INTO $db.$table SELECT 1, 1, 'hello', 'world', " +
             s"timestamp'2018-11-17 12:34:56', 'World'"))
+
```
bowenliang123, line 345: Remove the unessary blank line.

bowenliang123 (changes requested):
I have concerns in what to fix in this PR.
- What to do if expand the use range from original English chars to other Unicode chars, not limited to Chinese
- Use separate ut cases for testing `regexp_replace` method solely , for all the supposed supporting use ranges  
- Fix the conflicts

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
huangzhir: This is because of Scala code style checks, which do not allow Chinese characters. Using Chinese characters may cause code style check errors, even though the code can still run.
image

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
bowenliang123: so why `\u9fa5` used here?

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
```diff
@@ -76,10 +77,12 @@ trait DataMaskingTestBase extends AnyFunSuite with SparkSessionProvider with Bef
   test("simple query with a user doesn't have mask rules") {
     checkAnswer("kent", "SELECT key FROM default.src order by key", Seq(Row(1), Row(20), Row(30)))
   }
-
+  // SELECT 1, 1, 'hello', 'world', timestamp'2018-11-17 12:34:56', 'World'
   test("simple query with a user has mask rules") {
     val result =
-      Seq(Row(md5Hex("1"), "xxxxx", "worlx", Timestamp.valueOf("2018-01-01 00:00:00"), "Xorld"))
+      Seq(Row(md5Hex("1"), "xxxxx", "\u6d4b\u8bd5woxxx",
+        Timestamp.valueOf("2018-01-01 00:00:00"), "Xxxld\u6d4b\u8bd5"))
```
bowenliang123, line 84: I would say it's worse for understanding now.

bowenliang123:
Overall, it is not in my favor to fix "Chinese" chars.
It would suggest we move one step back to discuss what range of the chars to fix (what about other languages or even other unicode chars) , what is the relation between `regrex_replace` to the purpose, and what possible solution in Spark or in Ranger (like how they do in Ranger's Hive plugin).

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
bowenliang123: No. Source code do support utf8 chars including Chinese. Also, you would prefer reformatting tool for lining and consider inline-method comments. Comments are meant to make the case readable and easy to understand, especially you don't have exact Chinese words in the sql.

bowenliang123:
I think we should follow the same making rule implementation jas in Ranger's DataMasker. And just mask all the chars into `"x"` without classifying in upper/lower/digits or any type. @yaooqinn @huangzhir 

（https://github.com/apache/ranger/blob/000db1cbeb0f8fed1e02e88edbdc80386023b515/plugin-nestedstructure/src/main/java/org/apache/ranger/authorization/nestedstructure/authorizer/DataMasker.java#L250）, 

```
    private static String showFirstFour(String value){
        int length = StringUtils.length(value);

        return length <= 4 ? value : value.substring(0, 4) + StringUtils.repeat("x", length - 4);
    }
```

huangzhir:
Let me summarize how this issue came about and how Hive, Spark, and Trino handle it.

Hive's data masking is implemented using the functions mask({col}), mask_show_last_n({col}, 4, 'x', 'x', 'x', -1, '1'), and mask_show_first_n({col}, 4, 'x', 'x', 'x', -1, '1') (see https://github.com/apache/ranger/blob/7f5b82bff2df72f20f5c41ba095406d354f8acf0/agents-common/src/main/resources/service-defs/ranger-servicedef-hive.json#L387). 
```json 
                     {
				"itemId": 3,
				"name": "MASK_SHOW_FIRST_4",
				"label": "Partial mask: show first 4",
				"description": "Show first 4 characters; replace rest with 'x'",
				"transformer": "mask_show_first_n({col}, 4, 'x', 'x', 'x', -1, '1')"
			}
```
However, the implementation in the code ignores non-English character sets and simply returns the original data (see https://github.com/apache/hive/blob/7b3ecf617a6d46f48a3b6f77e0339fd4ad95a420/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFMask.java#L262).
```java 
      default:
        if(maskedOtherChar != UNMASKED_VAL) {
          return maskedOtherChar;
        }
        break;
    }
```

Regarding the related mask functions in Spark, there was a JIRA ticket (https://issues.apache.org/jira/browse/SPARK-23901) to add mask-related functions, but it was later decided that this was not a universal method, so the code was rolled back, and Spark does not currently have an implementation of mask-related functions.

Trino also does not implement these mask-related functions, but instead uses the regexp_replace function for data masking. However, Trino's regexp_replace function supports lambda expressions (see https://github.com/apache/ranger/blob/a0224b2fdef999b3e23e2374080df94bf38557a4/agents-common/src/main/resources/service-defs/ranger-servicedef-trino.json#L393).
```json
{
        "itemId": 2,
        "name": "MASK_SHOW_LAST_4",
        "label": "Partial mask: show last 4",
        "description": "Show last 4 characters; replace rest with 'X'",
        "transformer": "cast(regexp_replace({col}, '(.*)(.{4}$)', x -> regexp_replace(x[1], '.', 'X') || x[2]) as {type})"
      }
```

huangzhir:
> I think we should follow the same making rule implementation jas in Ranger's DataMasker. And just mask all the chars into `"x"` without classifying in upper/lower/digits or any other language type. @yaooqinn @huangzhir
> 
> （[https://github.com/apache/ranger/blob/000db1cbeb0f8fed1e02e88edbdc80386023b515/plugin-nestedstructure/src/main/java/org/apache/ranger/authorization/nestedstructure/authorizer/DataMasker.java#L250）](https://github.com/apache/ranger/blob/000db1cbeb0f8fed1e02e88edbdc80386023b515/plugin-nestedstructure/src/main/java/org/apache/ranger/authorization/nestedstructure/authorizer/DataMasker.java#L250%EF%BC%89),
> 
> ```
>     private static String showFirstFour(String value){
>         int length = StringUtils.length(value);
> 
>         return length <= 4 ? value : value.substring(0, 4) + StringUtils.repeat("x", length - 4);
>     }
> ```

If we don't refer to the implementations of Hive and Trino and there are no legacy issues, the natural way to implement data masking is to directly replace characters with "x". This PR is just a patch that fixes Chinese data masking and maintains some compatibility with Hive.

bowenliang123:
This PR is not quite persuasive to me for now.
I would like to see,
1. an explanation and implementation of different kinds of unicode chars. AndAgain Chinese is just one type of them, so PR 
2. not sure relation between unicode chars and utf8 chars, whether in stored data itself or inside Spark engine
2. short of e2e test, to masking data in data in table 
3. short of convincing test to test each case separately , all the ut mixed together , and part of tests limited to 3.2+

If this PR keeps to fix Chinese chars only (and only within the basic unicode range), I think I would not to pass it.
And I would like to hear any idea from  @yaooqinn  @pan3793 @Yikf @zhouyifan279

huangzhir:
I think we have the following solutions::
1. Make no changes and keep the behavior the same as Hive. (Convert uppercase to 'X', lowercase to 'x', and number to 'n')
2. Add a solution to address the issue of Chinese character obfuscation. (Convert Chinese characters to '密')
3. Replace each character set in Unicode with a different character.
4. Replace all characters except English letters and numbers with a single character. (Add replacement for other characters, e.g., convert all other characters to 'U')
5. Treat all characters in the same way. (Convert all characters to 'X')
6. Other solutions are welcome.

I am willing to work towards everyone's ideas to solve this problem.

yaooqinn:
> Add replacement for other characters, e.g., convert all other characters to 'U'

+1

huangzhir:
hello ,I'd like to request your help to review my PR. Based on @bowenliang123 @yaooqinn  your feedback, I've made some changes to the code and re-submitted the PR. The changes include:

Replacing all non-English letters, numbers, and spaces with the letter 'U'. This will ensure that all Unicode character sets for all languages are properly masked.
Adding unit tests specifically for regexp_replace, maskShowFirst4, and maskShowLast4.

@bowenliang123 Could you please review my changes and let me know if there's anything else I can improve on? Thank you

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
```diff
@@ -131,12 +125,28 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     }
   }
 
-  private def regexp_replace(expr: String, hasLen: Boolean = false): String = {
+  // public for test
```
yaooqinn, line 128: not necessary, you can add tests using sql queries

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
```diff
@@ -131,12 +125,28 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
...
+      regexp_replace(col, hasLen = true)
+    } else {
+      val right = regexp_replace(s"substr($col, 5)")
+      s"concat(substr($col, 0, 4), $right)"
+    }
+  }
+
+  def maskShowLast4(col: String): String = {
+    val left = regexp_replace(s"left($col, length($col) - 4)")
+    s"concat($left, right($col, 4))"
```
yaooqinn, line 149: seems not necessary changes

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
huangzhir: Thank you for your feedback. I have made the changes you suggested. Please let me know if there is anything else I can do to improve this PR.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
```diff
@@ -260,4 +260,50 @@ trait DataMaskingTestBase extends AnyFunSuite with SparkSessionProvider with Bef
         Seq(Row(md5Hex("1"), "hello")))
     }
   }
+
+  test("test MASK rule with non-English character set") {
```
yaooqinn, line 264: hi @huangzhir, these tests are not verifying any code we changed. please use a normal query like `select col0 from t1`, where `t1.col0` has a masker configuration with show_xxx_4. It shall be done by the masking rule automatically instead of what we do here manually

yaooqinn: you can refer to other tests in this test base

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
```diff
@@ -136,7 +136,8 @@ object SparkRangerAdminPlugin extends RangerBasePlugin("spark", "sparkSql")
     val upper = s"regexp_replace($expr, '[A-Z]', 'X'$pos)"
     val lower = s"regexp_replace($upper, '[a-z]', 'x'$pos)"
     val digits = s"regexp_replace($lower, '[0-9]', 'n'$pos)"
-    digits
+    val other = s"regexp_replace($digits, '[^A-Za-z0-9 ]', 'U'$pos)"
```
bowenliang123, line 139: The blank space should be treated as the common chars.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
bowenliang123: Maybe update the PR title too ?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPlugin.scala
huangzhir: Thank you for your feedback. I have made the changes you suggested.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
```diff
@@ -260,4 +260,51 @@ trait DataMaskingTestBase extends AnyFunSuite with SparkSessionProvider with Bef
...
+     */
+    // scalastyle:off
+    // AßþΔЙקم๗ቐあア叶葉엽 reference https://zh.wikipedia.org/zh-cn/Unicode#XML.E5.92.8CUnicode
+    val value1 = "hello WORD 123 ~!@# AßþΔЙקم๗ቐあア叶葉엽"
+    val value2 = "AßþΔЙקم๗ቐあア叶葉엽 hello WORD 123 ~!@#"
+    // scalastyle:on
+    doAs(
+      "admin",
+      sql(s"INSERT INTO default.src " +
+        s"SELECT 10, 4, '$value1', '$value1', timestamp'2018-11-17 12:34:56', '$value1'"))
```
yaooqinn, line 279: Can we move the data preparation to `setup`? Because it may become flaky when we add new tests in the future

yaooqinn:
The overall change looks good to me, except for minor revisions.

huangzhir:
there  are 2 checks were not successful. 
```
Error:  Plugin org.antlr:antlr4-maven-plugin:4.9.3 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.antlr:antlr4-maven-plugin:jar:4.9.3: The following artifacts could not be resolved: org.antlr:antlr4-maven-plugin:pom:4.9.3 (absent): Could not transfer artifact org.antlr:antlr4-maven-plugin:pom:4.9.3 from/to gcs-maven-central-mirror (https://maven-central-asia.storage-download.googleapis.com/maven2/): Connection timed out (Read failed) -> [Help 1]
```
```
Error:  Failed to execute goal on project kyuubi-spark-sql-engine_2.12: Could not resolve dependencies for project org.apache.kyuubi:kyuubi-spark-sql-engine_2.12:jar:1.8.0-SNAPSHOT: The following artifacts could not be resolved: org.apache.spark:spark-streaming_2.12:jar:3.2.3 (absent): Could not transfer artifact org.apache.spark:spark-streaming_2.12:jar:3.2.3 from/to gcs-maven-central-mirror (https://maven-central-asia.storage-download.googleapis.com/maven2/): Connection timed out (Read failed) -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project kyuubi-spark-sql-engine_2.12: Could not resolve dependencies for project org.apache.kyuubi:kyuubi-spark-sql-engine_2.12:jar:1.8.0-SNAPSHOT: The following artifacts could not be resolved: org.apache.spark:spark-streaming_2.12:jar:3.2.3 (absent): Could not transfer artifact org.apache.spark:spark-streaming_2.12:jar:3.2.3 from/to gcs-maven-central-mirror (https://maven-central-asia.storage-download.googleapis.com/maven2/): Connection timed out (Read failed)
```
These two errors seem to be caused by a network problem with connecting to https://maven-central-asia.storage-download.googleapis.com/maven2. After looking at other PRs, it seems that this issue is not unique.

pan3793:
Yea, the Google Maven Central mirror is not stable these days, I re-trigger the failed Job.

huangzhir:
```
Error:  Plugin org.antlr:antlr4-maven-plugin:4.9.3 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.antlr:antlr4-maven-plugin:jar:4.9.3: The following artifacts could not be resolved: org.antlr:antlr4-maven-plugin:pom:4.9.3 (absent): Could not transfer artifact org.antlr:antlr4-maven-plugin:pom:4.9.3 from/to gcs-maven-central-mirror (https://maven-central-asia.storage-download.googleapis.com/maven2/): Connection timed out (Read failed) -> [Help 1]
org.apache.maven.plugin.PluginResolutionException: Plugin org.antlr:antlr4-maven-plugin:4.9.3 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.antlr:antlr4-maven-plugin:jar:4.9.3
```
HI  @pan3793 ,There is still a check  not successful, and it seems to be a network issue. Could you please help to re-trigger the failed job again?

pan3793:
@huangzhir seems there is an un-addressed comment, and you can simply ignore the error caused by the network, the committer will check it before merging

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
huangzhir: I originally wanted to put it in the setup method, but putting it there caused several old test cases to fail. We need to modify the old test cases, such as adding a where clause in the old test cases. Do you mind this implementation approach?

yaooqinn: revising the old ones is OK with me.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
```diff
@@ -50,11 +50,14 @@ class SparkRangerAdminPluginSuite extends AnyFunSuite {
...
-      " '[0-9]', 'n')")
+      "regexp_replace(regexp_replace(regexp_replace(regexp_replace(value2, '[A-Z]', 'X')," +
+      " '[a-z]', 'x'), '[0-9]', 'n'), '[^A-Za-z0-9]', 'U')")
     assert(getMaskingExpr(buildAccessRequest(bob, "value3")).get contains "regexp_replace")
     assert(getMaskingExpr(buildAccessRequest(bob, "value4")).get === "date_trunc('YEAR', value4)")
-    assert(getMaskingExpr(buildAccessRequest(bob, "value5")).get contains "regexp_replace")
+    assert(getMaskingExpr(buildAccessRequest(bob, "value5")).get ===
+      "concat(regexp_replace(regexp_replace(regexp_replace(regexp_replace(" +
+      "left(value5, length(value5) - 4), '[A-Z]', 'X'), '[a-z]', 'x')," +
+      " '[0-9]', 'n'), '[^A-Za-z0-9]', 'U'), right(value5, 4))")
```
bowenliang123, line 60: I'm not sure about what's tested here. can we use plain value for assertion ?

bowenliang123 (approved):
LGTM overall. Minor comments.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
huangzhir: This is a test of the correctness of the getMaskingExpr method for the MASK_SHOW_LAST_4 rule, referring to the assert test MASK rule

huangzhir: > can we use plain value for assertion

I'm sorry, I didn't understand the meaning of this sentence.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
huangzhir: Thank you for your feedback. I have made the changes you suggested. Move the data preparation to setup，some tests were modified due to changes in the data.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
bowenliang123: I don't understand this test assertion.

bowenliang123 (changes requested):
Why is the filter condition `where key=1` added everywhere? Prefer not to break the existing regression tests, as it may change the behavior and plan for the SQL. And if you are determine to do so, explain the reasons and use separate tests.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
huangzhir: > Why is the filter condition `where key=1` added everywhere

 Based on @yaooqinn  suggestion, the data  preparation was moved to the setup method, which caused some changes in the data. The original test cases need to be modified by adding a where condition to keep searching for the original data.

huangzhir: > Why is the filter condition where key=1 added everywhere

Based on @yaooqinn suggestion, the data preparation was moved to the setup method, which caused some changes in the data. The original test cases need to be modified by adding a where condition to keep searching for the original data.

huangzhir:
> Why is the filter condition where key=1 added everywhere

Based on @yaooqinn suggestion, the data preparation was moved to the setup method, which caused some changes in the data. The original test cases need to be modified by adding a where condition to keep searching for the original data. Maintain the integrity of the old test results.

yaooqinn (approved):

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/SparkRangerAdminPluginSuite.scala
huangzhir: Originally, we used a simple 'contains' method to check. I think this can be improved by using '===' directly, which can more accurately determine whether the method of generating MASK_SHOW_LAST_4 based on this rule is correct or not.

bowenliang123: From my view, I would like to see a clear assertion with the pair of a concrete expected value  and a code to run. In current implementation , both sides on the assertion contain evaluation.
WDYT ? @yaooqinn

yaooqinn: I'm OK with the current implementation.

These tests are unimportant as we have e2e test cases. They are introduced when the data masking is not fully implemented. And at that time, I want to ensure the `getMaskingExpr` works well.

> both sides on the assertion contain evaluation.

I do not get your point. The author only uses '+' for newline.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
```diff
@@ -109,7 +128,7 @@ trait DataMaskingTestBase extends AnyFunSuite with SparkSessionProvider with Bef
       "bob",
       "SELECT coalesce(max(value1), 1), coalesce(max(value2), 1), coalesce(max(value3), 1), " +
         "coalesce(max(value4), timestamp '2018-01-01 22:33:44'), coalesce(max(value5), 1) " +
-        "FROM default.src",
+        "FROM default.src  where key = 1",
```
yaooqinn, line 131: nit: 2 spaces

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
```diff
@@ -119,26 +138,41 @@ trait DataMaskingTestBase extends AnyFunSuite with SparkSessionProvider with Bef
     checkAnswer(
       "bob",
       "SELECT value1, value2, value3, value4, value5 FROM default.src WHERE value2 in " +
-        "(SELECT value2 as key FROM default.src)",
+        "(SELECT value2 as key FROM default.src  where key = 1)",
```
yaooqinn, line 141: ditto

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
```diff
@@ -152,7 +186,7 @@ trait DataMaskingTestBase extends AnyFunSuite with SparkSessionProvider with Bef
 
   test("self join on a masked table") {
     val s = "SELECT a.value1, b.value1 FROM default.src a" +
-      " join default.src b on a.value1=b.value1"
+      " join default.src b on a.value1=b.value1  where a.key = 1 and b.key = 1 "
```
yaooqinn, line 189: ditto

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
huangzhir: Fixed, thanks!

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
huangzhir: Fixed, thanks!

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/datamasking/DataMaskingTestBase.scala
huangzhir: Fixed, thanks!

bowenliang123 (approved):

yaooqinn:
Thanks, @huangzhir, for the hard-working and patience all the way thru.

Merged to master.

huangzhir:
You're welcome! It was my pleasure to contribute to this project.

PR #4657, opened 2023-04-04
Building rest client to kyuubi instance including original host urls

turboFei:

Usually, the host url to create a batch is the load balancer uri.

To reduce the internal rest redirection when fetching log, we support to call the kyuubi instance that created the batch directly.

It is better to add original host urls when building the rest client to kyuubi instance, in case that there is firewall to the kyuubi instance directly.


turboFei:
cc @pan3793

pan3793 (approved):

turboFei:
thanks, merging to master

PR #4665, opened 2023-04-04
[KYUUBI #4656] client open session with original engineRefId

lightning-L:


lightning-L:
@turboFei

turboFei:
this comment need to be addressed.

https://github.com/apache/kyuubi/pull/4545#issuecomment-1478931310

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -78,7 +78,7 @@ class KyuubiSessionImpl(
     sessionConf,
     user,
     sessionManager.groupProvider,
-    handle.identifier.toString,
+    optimizedConf.get(KYUUBI_ENGINE_REFID).getOrElse(handle.identifier.toString),
```
cxzl25, line 81:
```suggestion
    optimizedConf.getOrElse(KYUUBI_ENGINE_REFID, handle.identifier.toString),
```

PR #4667, opened 2023-04-05
Set subdomainDefault to fault in list engine

yuruguo:
Set `subdomainDefault` to `default` in list engine restful


pan3793:
@yuruguo, thanks for your contribution, and please explain **WHY** are the changes needed, not **WHAT** you changed.

yuruguo:
> @yuruguo, thanks for your contribution, and please explain **WHY** are the changes needed, not **WHAT** you changed.

Kyuubi could use default value `default` to delete engine If the `subdomain` is not specified in the **delete interface**. 
https://github.com/apache/kyuubi/blob/5faebb1e75b57491ca655e0ba6fceacde5ef9459/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala#L221
In this case, the engine nodes under the `default` path will be deleted, and the corresponding **list interface** should also be consistent. Otherwise, no engine nodes will be returned.

yaooqinn:
cc @turboFei

turboFei:
seems that this is not needed for me.

turboFei:
It seems that, now it will list all engines, not only `default` subdomain. 

So, I think this pr is not needed.

PR #4670, opened 2023-04-06
[KYUUBI #3653][REST] AdminResource add list kyuubi server api

zwangsheng:

Add List Kyuubi Server Api for `AdminResource`

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
@@ -31,7 +31,7 @@ import org.apache.commons.lang3.StringUtils
 import org.apache.zookeeper.KeeperException.NoNodeException
 
 import org.apache.kyuubi.{KYUUBI_VERSION, Logging, Utils}
-import org.apache.kyuubi.client.api.v1.dto.{Engine, OperationData, SessionData}
+import org.apache.kyuubi.client.api.v1.dto.{Engine, OperationData, ServerData, SessionData}
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf._
 import org.apache.kyuubi.ha.HighAvailabilityConf.HA_NAMESPACE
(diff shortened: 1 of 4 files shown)


zwangsheng:
cc @turboFei @pan3793

turboFei:
some comments:
1. we can support `discoveryService` for KyuubiRestFrontendService
2. we shall return the rest url for the API in this pr

zwangsheng:
> some comments:
> 
> 1. we can support `discoveryService` for KyuubiRestFrontendService
> 2. we shall return the rest fe urls for the API in this pr

@turboFei 

Sure, made changes.

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/AdminResourceSuite.scala
```diff
@@ -488,4 +491,41 @@ class AdminResourceSuite extends KyuubiFunSuite with RestFrontendTestHelper {
...
+    lenient.when(serverDiscovery.fe).thenReturn(fe)
+    val namespace = conf.get(HighAvailabilityConf.HA_NAMESPACE)
+    withDiscoveryClient(conf) { client =>
+      client.registerService(conf, namespace, serverDiscovery)
+
+      val adminUser = Utils.currentUser
+      val encodeAuthorization = new String(
+        Base64.getEncoder.encode(
+          s"$adminUser:".getBytes()),
+        "UTF-8")
```
bowenliang123, line 507:
```suggestion
        StandardCharsets.UTF_8)
```

zwangsheng: Thanks for point this.

By the way, other similar places in this document have been modified and duplicate code has been reduced.

zwangsheng:
Also cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -296,6 +296,35 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
+    val ipAddress = fe.getIpAddress
+    info(s"Received list all live kyuubi servers request from $userName/$ipAddress")
+    if (!isAdministrator(userName)) {
+      throw new NotAllowedException(
+        s"$userName is not allowed to list all live kyuubi servers")
+    }
+    val kyuubiConf = fe.getConf
+    val servers = ListBuffer[ServerData]()
+    val serverSpec = DiscoveryPaths.makePath(null, kyuubiConf.get(HA_NAMESPACE))
+    withDiscoveryClient(kyuubiConf) { discoveryClient =>
```
pan3793, line 320: what will happen if HA is disabled

zwangsheng: When disabled HA, `Kyuubi Server` will register self to `EmbeddedZookeeper`.

It's predictably to get only itself when disabled HA.

pan3793 (approved):

turboFei (approved):
LGTM,

BTW, I recommend to support discovery service for KyuubiRestFrontendService.

And the URLs return in the api added in this PR should be the connection urls of `KyuubiRestFrontendService`s.

zwangsheng:
> BTW, I recommend to support discovery service for KyuubiRestFrontendService.
> And the URLs return in the api added in this PR should be the connection urls of `KyuubiRestFrontendService`s.

I agree with support discovery service for `KyuubiRestFrontendService` and  expose rest connection url in HA node.

Seems this is a different goal feature, i'd like make a new PR to implement.

zwangsheng:
ping @pan3793 @turboFei, we can merge this.

turboFei:
thanks, merging to master

PR #4677, opened 2023-04-07
[AUTHZ] Check generated policy file in test suite

bowenliang123:

- add ut to check generated Ranger policy file in #4585 
- manually activated `genpolicy` profile in CI builds, as the property based activation not auto-triggered  as expectedly with property `ranger.version=2.4.0` set in project parent pom
- Support regenerated policy file within the same test suite, by running
`KYUUBI_UPDATE=1 build/mvn clean test -pl :kyuubi-spark-authz_2.12 -Dtest=none -DwildcardSuites=org.apache.kyuubi.plugin.spark.authz.gen.PolicyJsonFileGenerator -Pgenpolicy`


yaooqinn (approved):

extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/PolicyJsonFileGenerator.scala
```diff
@@ -17,49 +17,59 @@
...
+      // scalastyle:on println
+      Files.write(
+        policyFilePath,
+        generatedStr.getBytes(StandardCharsets.UTF_8),
+        StandardOpenOption.CREATE,
+        StandardOpenOption.TRUNCATE_EXISTING)
+    } else {
+      val existedFileContent = Files.readAllLines(policyFilePath).asScala.mkString("\n")
+      withClue("Please regenerate the ranger policy file by running"
+        + "`KYUUBI_UPDATE=1 build/mvn clean test -Pgenpolicy"
```
pan3793, line 67: maybe `gen-policy` is better, align the style w/ others

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -332,14 +332,7 @@
 
     <profiles>
         <profile>
-            <id>genpolicy</id>
-            <activation>
-                <!-- activated when Ranger version is identical to required-->
-                <property>
-                    <name>ranger.version</name>
-                    <value>2.4.0</value>
-                </property>
```
bowenliang123, line 341: Here removes the property-based activation introduced previously in #4585, since it's not activated with `ranger. version` set in project parent pom in value `2.4.0` (which requires more investigation in profile mechanism).

cc @Yikf

extensions/spark/kyuubi-spark-authz/src/test/gen/scala/org/apache/kyuubi/plugin/spark/authz/gen/PolicyJsonFileGenerator.scala
bowenliang123: Good point, changed the profile name to `gen-policy`. Also updated usages in comments and ci workflows.

pan3793 (approved):

extensions/spark/kyuubi-spark-authz/pom.xml
Yikf: Thanks for ping me, So do we have to specify the profile manually now?

bowenliang123: Yes, and the profile name changed to `gen-policy`.

.github/workflows/style.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     strategy:
       matrix:
         profiles:
-          - '-Pflink-provided,hive-provided,spark-provided,spark-block-cleaner,spark-3.3,spark-3.2,spark-3.1,tpcds,kubernetes-it'
+          - '-Pflink-provided,hive-provided,spark-provided,spark-block-cleaner,spark-3.3,spark-3.2,spark-3.1,tpcds,kubernetes-it,gen-policy'
```
yaooqinn, line 37: This is not necessary

bowenliang123: If without adding this profile, the generator won't be checked for styling.

yaooqinn: I don't understand why

bowenliang123:
Thanks, merged to master.

PR #4678, opened 2023-04-10
Improve FinalStageResourceManager kill executors

ulysses-you:

This pr change two things:
1. add a config to kill executors if the plan contains table caches. It's not always safe to kill executors if the cache is referenced by two write-like plan.
2. force adjustTargetNumExecutors when killing executors. YarnAllocator` might re-request original target executors if DRA has not updated target executors yet. Note, DRA would re-adjust executors if there are more tasks to be executed, so we are safe. It's better to adjuest target num executor once we kill executors.

These issues are found during my POC


yaooqinn (approved):

ulysses-you:
thanks, merging to master

PR #4680, opened 2023-04-10
Skip to get the batch application status when creating batch

turboFei:

To prevent the `POST /batches` operation stuck when getting batch application status.

For example: the yarn client always failover because of resource manager issue.

And it is also not needed that get the batch app status info when creating batch.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -71,10 +71,10 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
 
   private def sessionManager = fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]
 
-  private def buildBatch(session: KyuubiBatchSessionImpl): Batch = {
+  private def buildBatch(session: KyuubiBatchSessionImpl, needAppStatus: Boolean = true): Batch = {
     val batchOp = session.batchJobSubmissionOp
     val batchOpStatus = batchOp.getStatus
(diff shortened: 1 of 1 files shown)


ulysses-you:
How can we return app id if we do not get application status when creating batch ? Do you mean we do not need to return app related info when creating batch ?

turboFei:
> How can we return app id if we do not get application status when creating batch ? Do you mean we do not need to return app related info when creating batch ?


Yes, the key information is batchId.

ulysses-you:
cc @pan3793 not sure if it breaks your customer ..

turboFei:
I think that usually the response for create batch request should be returned in milliseconds.

So, it is impossible to get the application status in most cases.

pan3793:
> it is impossible to get the application status in most cases.

What's the current result in your production case? UNKNOWN?

turboFei:
image

It is none, I think.

pan3793:
We have UNKNOWN state, why should return null/None instead?

pan3793:
I think we need to re-evaluate https://github.com/apache/kyuubi/pull/3922, it's similar to https://github.com/apache/kyuubi/pull/4469

turboFei:
cc @zwangsheng as well

pan3793:
My idea is, revert https://github.com/apache/kyuubi/pull/3922, and always propagate app status to the client.

During submitting phase, ApplicationManager returns UNKNOWN instead of NOT_FOUND when elapsed time < `kyuubi.engine.submit.timeout`, same as we did in https://github.com/apache/kyuubi/pull/4469 for K8s

zwangsheng:
I'm + 1 for get `Unknown` state, after spark submit but we not find application from yarn cluster.

And using timeout to check whether  submit success or not.

Unknown sounds more reasonable, when we submit but can fount app.

PR #4688, opened 2023-04-10
Fix the failure to read the operation log after executing Catalog and database operation

cxzl25:

Now `GetCurrentCatalog`/`GetCurrentDatabase`/`SetCurrentCatalog`/`SetCurrentDatabase`is executed through the statement, and the jdbc client will try to obtain the operation log corresponding to the statement. 
At present, these operations do not generate operation logs, so the engine log will be throw exception(`failed to generate operation log`).

```java
23/04/10 20:25:23 INFO GetCurrentCatalog: Processing anonymous's query[8218e7ed-b4a4-41ad-a1cc-6f82bf3d55bb]: INITIALIZED_STATE -> RUNNING_STATE, statement:
GetCurrentCatalog
23/04/10 20:25:23 INFO GetCurrentCatalog: Processing anonymous's query[8218e7ed-b4a4-41ad-a1cc-6f82bf3d55bb]: RUNNING_STATE -> FINISHED_STATE, time taken: 0.002 seconds
23/04/10 20:25:23 ERROR SparkTBinaryFrontendService: Error fetching results:
org.apache.kyuubi.KyuubiSQLException: OperationHandle [8218e7ed-b4a4-41ad-a1cc-6f82bf3d55bb] failed to generate operation log
        at org.apache.kyuubi.KyuubiSQLException$.apply(KyuubiSQLException.scala:69)
        at org.apache.kyuubi.operation.OperationManager.$anonfun$getOperationLogRowSet$2(OperationManager.scala:146)
        at scala.Option.getOrElse(Option.scala:189)
```

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
@@ -195,6 +195,8 @@ class OperationLog(path: Path) {
   }
 
   def close(): Unit = synchronized {
+    if (!initialized) return
+
     closeExtraReaders()
 
     trySafely {
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetCurrentCatalog.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/GetCurrentCatalog.scala
@@ -18,10 +18,16 @@
 package org.apache.kyuubi.engine.flink.operation
 
 import org.apache.kyuubi.engine.flink.result.ResultSetUtil
+import org.apache.kyuubi.operation.log.OperationLog
 import org.apache.kyuubi.operation.meta.ResultSetSchemaConstant.TABLE_CAT
 import org.apache.kyuubi.session.Session
 
 class GetCurrentCatalog(session: Session) extends FlinkOperation(session) {
+
+  private val operationLog: OperationLog =
+    OperationLog.createOperationLog(session, getHandle)
+
+  override def getOperationLog: Option[OperationLog] = Option(operationLog)
 
   override protected def runInternal(): Unit = {
     try {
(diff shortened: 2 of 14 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -195,6 +195,8 @@ class OperationLog(path: Path) {
   }
 
   def close(): Unit = synchronized {
+    if (!initialized) return
```
cxzl25, line 198: When the operation log is not initialized, the close method will throw an exception.

```java
Caused by: java.nio.file.NoSuchFileException:

        at org.apache.kyuubi.operation.log.OperationLog.reader$lzycompute(OperationLog.scala:89)
        at org.apache.kyuubi.operation.log.OperationLog.reader(OperationLog.scala:89)
        at org.apache.kyuubi.operation.log.OperationLog.$anonfun$close$1(OperationLog.scala:201)
        at org.apache.kyuubi.operation.log.OperationLog.trySafely(OperationLog.scala:221)
```

pan3793 (approved):

yaooqinn:
Why do we create operation logs for these lightweight operations?

yaooqinn:
Or why do we fetch logs from the client side?

cxzl25:
> Why do we create operation logs for these lightweight operations?

Because `OperationManager#getOperationLogRowSet` will check whether the operation log has a value, or allow `operationLog` to be none here instead of throwing an exception.

https://github.com/apache/kyuubi/blob/f0615a9aab0a5b755c16ee0b966a5ea59b98bd10/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala#L144-L152


> Or why do we fetch logs from the client side?

Because these operations are executed using `KyuubiStatement`, `isLogBeingGenerated` is true by default and will try to fetch query log, which is the default behavior of client.

https://github.com/apache/kyuubi/blob/f0615a9aab0a5b755c16ee0b966a5ea59b98bd10/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java#L649-L660

yaooqinn:
> Because these operations are executed using KyuubiStatement, isLogBeingGenerated is true by default and will try to fetch query log, which is the default behavior of client.

Can we update the caller side? Our meta operations do not have operation logs and they work well

cxzl25:
> Can we update the caller side? 

If it is to update the operations of the kyuubi jdbc client, the cost of these operations is a bit high.

> Our meta operations do not have operation logs and they work well

Other metadata operations are obtained through the `KyuubiDatabaseMetaData` method, which are directly called through RPC, so there is no need to call `KyuubiStatement` to obtain.

The execution cost of `OperationLog.createOperationLog` is very small. 
It just declares a file, it is not actually created, and it does not need to be deleted.

pan3793:
Thanks, merged to master/1.7

PR #4691, opened 2023-04-11
[REST] Configure FAIL_ON_UNKNOWN_PROPERTIES to false for KyuubiScalaObjectMapper

turboFei:

Do not failed on unknown properties in server side.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
@@ -19,11 +19,13 @@ package org.apache.kyuubi.server.api
 
 import javax.ws.rs.ext.ContextResolver
 
-import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 
 class KyuubiScalaObjectMapper extends ContextResolver[ObjectMapper] {
-  private val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
+  private val mapper = new ObjectMapper()
+    .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+    .registerModule(DefaultScalaModule)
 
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
```diff
@@ -19,11 +19,13 @@ package org.apache.kyuubi.server.api
...
 
-import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 
 class KyuubiScalaObjectMapper extends ContextResolver[ObjectMapper] {
-  private val mapper = new ObjectMapper().registerModule(DefaultScalaModule)
+  private val mapper = new ObjectMapper()
+    .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+    .registerModule(DefaultScalaModule)
```
bowenliang123, line 28: How about using a chaining and fluent style builder with JsonMapper (since 2.10) ?
`JsonMapper` itself extends `ObjectMapper`in JSON-format specific.

```suggestion
  private val mapper = JsonMapper.builder()
    .addModules(DefaultScalaModule)
    .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
    .build()
```

bowenliang123 (approved):
LGTM. Minor suggestion.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
turboFei: thanks for the suggestion.

Now there are multiple parts of code likes this. So, I think that we can just keep this as it is now.

Maybe you can raise a dedicated pr to refactor them all.

turboFei:
thanks, merged to master and 1.7.1

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/KyuubiScalaObjectMapper.scala
bowenliang123: SGTM.

PR #4692, opened 2023-04-11
[INFRA] Increase Maven Resolver Transport connection timeout

bowenliang123:

To resolve connection timeout when resolving dependency from Maven central's mirror.
The decreased connection timeout by switching the default transporter in Maven 3.9 may cause the problem.

- “the default transport in Maven 3.9.0+ is NOT Wagon anymore. ” (https://maven.apache.org/guides/mini/guide-http-settings.html)
- "Given the priority of native is higher than wagon, and that starting with Maven 3.9.0 both are present on resolver classpath, native is chosen by default. " (https://maven.apache.org/guides/mini/guide-resolver-transport.html#switching-between-transports)
- "The Resolver contains configuration for many aspects, including transport (see the keys prefixed with aether.connector." (https://maven.apache.org/guides/mini/guide-resolver-transport.html)
- "To set one of the configuration options from above just use system variables. "(https://maven.apache.org/resolver/configuration.html#set-configuration-from-apache-maven)
- https://github.com/apache/maven-wagon/blob/master/wagon-provider-api/src/main/java/org/apache/maven/wagon/Wagon.java#L44
Wagon's default connect timeout is 60000ms (1min)

- https://maven.apache.org/resolver/configuration.html
Resolver Transport's default connection timeout: 10000ms (10sec)

### DIFF
--- a/.github/maven/settings.xml
+++ b/.github/maven/settings.xml
@@ -5,7 +5,7 @@
         <server>
             <id>gcs-maven-central-mirror</id>
             <configuration>
-                <aether.connector.connectTimeout>60000</aether.connector.connectTimeout>
+                <connectTimeout>10000</connectTimeout>
             </configuration>
         </server>
     </servers>
--- a/.github/maven/settings.xml
+++ b/.github/maven/settings.xml
@@ -5,7 +5,7 @@
         <server>
             <id>gcs-maven-central-mirror</id>
             <configuration>
-                <connectTimeout>10000</connectTimeout>
+                <connectTimeout>60000</connectTimeout>
             </configuration>
         </server>
     </servers>
--- a/.github/maven/settings.xml
+++ b/.github/maven/settings.xml
@@ -1,0 +1,12 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
+    <servers>
+        <server>
+            <id>gcs-maven-central-mirror</id>
+            <configuration>
+                <aether.connector.connectTimeout>60000</aether.connector.connectTimeout>
+            </configuration>
+        </server>
+    </servers>
+</settings>
--- a/build/mvn
+++ b/build/mvn
@@ -25,7 +25,7 @@ _CALLING_DIR="$(pwd)"
 _COMPILE_JVM_OPTS="-Xms2g -Xmx2g -XX:ReservedCodeCacheSize=1g -Xss128m"
 
 if [ "$CI" ]; then
-  export MAVEN_CLI_OPTS="--no-transfer-progress --errors --fail-fast"
+  export MAVEN_CLI_OPTS="--no-transfer-progress --errors --fail-fast --settings ${GITHUB_WORKSPACE}/.github/maven/settings.xml"
 fi
 
 # Installs any application tarball given a URL, the expected tarball name,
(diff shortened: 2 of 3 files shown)


bowenliang123:
cc @yaooqinn @pan3793

.github/workflows/master.yml
```diff
@@ -35,6 +35,7 @@ env:
   MVN_OPT: -Dmaven.javadoc.skip=true -Drat.skip=true -Dscalastyle.skip=true -Dspotless.check.skip -Dorg.slf4j.simpleLogger.defaultLogLevel=warn -Pjdbc-shaded -Dmaven.plugin.download.cache.path=/tmp/engine-archives
   KUBERNETES_VERSION: v1.26.1
   MINIKUBE_VERSION: v1.29.0
+  "aether.connector.connectTimeout": 60000
```
LuciferYang, line 38: I am not familiar with this property, I want to know if it can be placed in `MVN_OPT` as `-Daether.connector.connectTimeout=60000` or can it be configured in `pom.xml`? thanks  ~

bowenliang123: Not sure about this. All I found is "To set one of the configuration options from above just use system variables. " mentioned in the description of this PR. But as you see some tests still failed when resolving dependencies, it seems that setting it here in the system env variable does not help. I would like to see what happens when you are doing setting it via property in pom in Spark.

.github/maven/settings.xml
```diff
@@ -0,0 +1,12 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
+    <servers>
+        <server>
+            <id>gcs-maven-central-mirror</id>
+            <configuration>
+                <aether.connector.connectTimeout>60000</aether.connector.connectTimeout>
```
cxzl25, line 8:
```xml
<connectTimeout>60000</connectTimeout>
```

https://maven.apache.org/guides/mini/guide-resolver-transport.html#timeouts

build/mvn
```diff
@@ -25,7 +25,7 @@ _CALLING_DIR="$(pwd)"
 _COMPILE_JVM_OPTS="-Xms2g -Xmx2g -XX:ReservedCodeCacheSize=1g -Xss128m"
 
 if [ "$CI" ]; then
-  export MAVEN_CLI_OPTS="--no-transfer-progress --errors --fail-fast"
+  export MAVEN_CLI_OPTS="--no-transfer-progress --errors --fail-fast --settings ${GITHUB_WORKSPACE}/.github/maven/settings.xml"
```
cxzl25, line 28:
```
-Dmaven.resolver.transport=wagon
```
https://maven.apache.org/guides/mini/guide-resolver-transport.html#how-to-upgrade-from-wagon-or-%E2%80%9Cnative-transport-does-not-work%E2%80%9D

.github/maven/settings.xml
bowenliang123: Let's try this. Thanks.

.github/workflows/master.yml
pan3793: introduce a `settings.xml` looks too complex for new developers

build/mvn
bowenliang123: It seems Wagon also requires settings.xml if we need to set connection timeout.
https://maven.apache.org/guides/mini/guide-http-settings.html#connection-timeouts

LuciferYang: The default timeout of `wagon` also changed?

LuciferYang: image

@bowenliang123 As you said, `Wagon's default connect timeout is 60000ms (1min)`, Do we still need to increase it?

bowenliang123: Not sure about this. @LuciferYang 
And according to @cxzl25 's advice, we are trying to switch to use Wagon in #4699 .

PR #4694, opened 2023-04-11
[KYUUBI #4693] Enhanced the table lineage for input tables

### LINKED ISSUE #4693: [Improvement] Enhanced the table lineage for input tables
### Code of Conduct

### Search before asking

```
      spark.sql("CREATE TABLE t1 (a string, b string) USING hive")
      spark.sql("CREATE TABLE t2 (a string, b string) USING hive")
      val ret0 = exectractLineage("select t1.a from t1 where t1.b in (select b from t2)")
      assert(ret0 == Lineage(
        List("default.t1", "default.t2"),
        List(),
        List(("a", Set("default.t1.a")))))

      val ret1 = exectractLineage("select t1.a from t1 join t2")
      assert(ret1 == Lineage(
        List("default.t1", "default.t2"),
        List(),
        List(("a", Set("default.t1.a")))))
```
In actual scenarios, it is necessary to display all input tables, even if a table may not contribute to the output columns.

### How should we improve?

_No response_

iodone:
close #4693 

### DIFF
--- a/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
+++ b/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
@@ -19,6 +19,7 @@ package org.apache.kyuubi.plugin.lineage.helper
 
 import scala.collection.immutable.ListMap
 import scala.util.{Failure, Success, Try}
+
 import org.apache.spark.internal.Logging
 import org.apache.spark.kyuubi.lineage.{LineageConf, SparkContextHelper}
 import org.apache.spark.sql.SparkSession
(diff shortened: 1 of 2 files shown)


ulysses-you:
I think we'd better to define lineage clearly. The current lineage is easy to follow that we only considers plan's output, which means if a column is used to do filter, sort or something do not affect schema, then we ignore it.

So if we want to consider those columns, how about adding a new mode to fully extract column and table lineage? For example:
```sql
INSERT INTO TABLE t
SELECT c1 FROM t1 WHERE c2 > 0 ORDER BY c3

-- The lineage should be:

ColumnUsage(to: String, from: String, usage: String)

Lineage(
  List("default.t1"),
  List("default.t"),
  List(
     ColumnUsage("c1", "default.t.c1", "OUTPUT"),
     ColumnUsage("N/A", "default.t.c2", "PREDICATE"),
     ColumnUsage("N/A", "default.t.c3", "ORDERING")
  )
)

```

PR #4699, opened 2023-04-12
[INFRA] Switch Maven's resolver transport to legacy Wagon

bowenliang123:

- since Mavn 3.9.x, default resolver transport changed to `native` and the default connection timeout shrinked from 60s to 10s.
- Switch back to legacy Wagon to avoid causing frequent connection timeout when resolving dependencies
https://maven.apache.org/guides/mini/guide-resolver-transport.html#switching-between-transports


bowenliang123:
Closing this PR. Leave as it is, since Maven reverted to 3.8.x.

PR #4733, opened 2023-04-19
Introduce Kafka event logger for server events

bowenliang123:

- introduce new event logger type `KAFKA`
- using Kafka 2.3.2 for testing , as the latest version supporting zookeeper 2.4.x. 


bowenliang123:
cc @lxbalex @jeanlyn

pan3793:
We can use Testcontianers to address the Kafka server compatibility issues

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2007,12 +2007,26 @@ object KyuubiConf {
...
+      .createOptional
+
   val SERVER_EVENT_LOGGERS: ConfigEntry[Seq[String]] =
     buildConf("kyuubi.backend.server.event.loggers")
       .doc("A comma-separated list of server history loggers, where session/operation etc" +
         " events go.<ul>" +
         s" <li>JSON: the events will be written to the location of" +
         s" ${SERVER_EVENT_JSON_LOG_PATH.key}</li>" +
+        s" <li>" +
+        s"KAFKA: the events sent to topic of `${SERVER_EVENT_KAFKA_TOPIC.key}`, " +
```
pan3793, line 2025: unify the format

dev/dependencyList
```diff
@@ -181,6 +183,7 @@ simpleclient_tracer_otel/0.16.0//simpleclient_tracer_otel-0.16.0.jar
 simpleclient_tracer_otel_agent/0.16.0//simpleclient_tracer_otel_agent-0.16.0.jar
 slf4j-api/1.7.36//slf4j-api-1.7.36.jar
 snakeyaml/1.33//snakeyaml-1.33.jar
+snappy-java/1.1.7.3//snappy-java-1.1.7.3.jar
```
pan3793, line 186: does it support arm?

pan3793: when introduce new artifacts in release tarball, don't forget to update LICENSE/NOTICE

kyuubi-server/src/main/scala/org/apache/kyuubi/events/ServerEventHandlerRegister.scala
```diff
@@ -36,6 +38,17 @@ object ServerEventHandlerRegister extends EventHandlerRegister {
       kyuubiConf)
   }
 
+  override def createKafkaEventHandler(kyuubiConf: KyuubiConf): EventHandler[KyuubiEvent] = {
+    val topic: String = kyuubiConf.get(SERVER_EVENT_KAFKA_TOPIC).getOrElse({
+      throw new IllegalArgumentException(s"${SERVER_EVENT_KAFKA_TOPIC.key} must be configured")
+    })
```
pan3793, line 44: Eliminate unnecessary `()`
```suggestion
    val topic: String = kyuubiConf.get(SERVER_EVENT_KAFKA_TOPIC).getOrElse {
      throw new IllegalArgumentException(s"${SERVER_EVENT_KAFKA_TOPIC.key} must be configured")
    }
```

kyuubi-server/src/main/scala/org/apache/kyuubi/events/ServerEventHandlerRegister.scala
```diff
@@ -36,6 +38,17 @@ object ServerEventHandlerRegister extends EventHandlerRegister {
...
+    val producerConf = new Properties()
+    kyuubiConf.getAllWithPrefix(KAFKA_SERVER_EVENT_HANDLER_PREFIX, "").foreach({
+      case (k, v) => producerConf.setProperty(k, v)
+    })
```
pan3793, line 48: ditto, for `()`

kyuubi-server/src/test/scala/org/apache/kyuubi/events/handler/ServerKafkaLoggingEventHandlerSuite.scala
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.apache.kafka.common.utils.Time
+
+import org.apache.kyuubi._
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.events.handler.ServerKafkaLoggingEventHandler.KAFKA_SERVER_EVENT_HANDLER_PREFIX
+import org.apache.kyuubi.operation.HiveJDBCTestHelper
+import org.apache.kyuubi.zookeeper.{EmbeddedZookeeper, ZookeeperConf}
+
+class ServerKafkaLoggingEventHandlerSuite extends WithKyuubiServer with HiveJDBCTestHelper
```
pan3793, line 39: suggest to use Testcontianers, and move it to a dedicated it module

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
```diff
@@ -0,0 +1,58 @@
...
+    if (!producerConf.containsKey("value.serializer")) {
+      producerConf.put("value.serializer", "org.apache.kafka.common.serialization.StringSerializer")
+    }
+
+    new KafkaProducer[String, String](producerConf)
+  }
+
+  override def apply(event: KyuubiEvent): Unit = {
+    try {
+      val record = new ProducerRecord[String, String](topic, event.eventType, event.toJson)
```
pan3793, line 51: I remember that Kafka allows setting event timestamp on constructing `ProducerRecord`, it may be useful in some cases.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * This event logger logs Kyuubi engine events to Kafka.
+ */
+class KafkaLoggingEventHandler(
+    topic: String,
+    producerConf: Properties,
+    kyuubiConf: KyuubiConf) extends EventHandler[KyuubiEvent]
+  with Logging {
+
+  private val kafkaProducer: KafkaProducer[String, String] = {
```
pan3793, line 37: when to close?

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.kyuubi.events.KyuubiEvent
+
+/**
+ * This event logger logs Kyuubi engine events to Kafka.
+ */
+class KafkaLoggingEventHandler(
+    topic: String,
+    producerConf: Properties,
+    kyuubiConf: KyuubiConf) extends EventHandler[KyuubiEvent]
+  with Logging {
```
pan3793, line 35: does the scalafmt allow merging it to the previous line?

kyuubi-server/src/test/scala/org/apache/kyuubi/events/handler/ServerKafkaLoggingEventHandlerSuite.scala
pan3793: Also, there is `Mock*` in Kafka client, which is much light that launch real Kafka broker

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: Sure. Fixed.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: Good suggestion. Setting timestamp with `System.currentTimeMillis`. Is that good enough? Or use Java8's `Instant.now.toEpochMilli` (with UTC clock)?

kyuubi-server/src/test/scala/org/apache/kyuubi/events/handler/ServerKafkaLoggingEventHandlerSuite.scala
bowenliang123: I may need further investigation for mocking a Kafka broker. I also considered the TestingServer in `curator-test`, as the way does in Ranger, but it's not working since the crossed versions among maven dependencies.

bowenliang123: mark: https://www.baeldung.com/kafka-mockproducer

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
bowenliang123: OK, change to mention the format with `KAFKA: the events serialized in JSON format`

dev/dependencyList
bowenliang123: > snappy-java supports Windows, Mac OS X, Linux (x86, x86_64, arm, etc...)

Acorrding to `snappy-java`'s doc here(https://github.com/xerial/snappy-java/blob/master/BUILD.md), arm architecture should be supported.

kyuubi-server/src/test/scala/org/apache/kyuubi/events/handler/ServerKafkaLoggingEventHandlerSuite.scala
bowenliang123: MockProducer of `Mock*` is not enough for testing which requires simulating a kafka server with bootstrap server URL. Turn to use TestContainers for staring Kafka instance with Docker image.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: While timestamp is not set , the producer will assign the timestamp using `System.currentTimeMillis()`.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
cxzl25: `KafkaLoggingEventHandler` doesn't seem to implement the `AutoCloseable` interface. Did I miss something?

bowenliang123: Oh, thanks for the catch. That's my mistake as I added and removed the closing mechanism several times for repeating unstable test results in random ut cases and I missed it the last time. Made `KafkaLoggingEventHandler` implement `AutoCloseable` again now.

bowenliang123: Don't know why when I close the handler in `KafkaLoggingEventHandler`. It fails the other ut,
```
- get operation log *** FAILED ***
  scala.collection.JavaConverters.asScalaBufferConverter[String](logRowSet.getLogRowSet()).asScala.exists(((x$1: String) => x$1.contains("show tables"))) was false (OperationsResourceSuite.scala:103)
```

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
pan3793: Usually, it should be event time, but I just notice that KyuubiEvent does not have event time ...

pom.xml
```diff
@@ -1524,6 +1531,22 @@
                 <artifactId>service</artifactId>
                 <version>${openai.java.version}</version>
             </dependency>
+
+            <dependency>
+                <groupId>org.apache.kafka</groupId>
+                <artifactId>kafka-clients</artifactId>
+                <version>${kafka.version}</version>
+                <exclusions>
```
pan3793, line 1539: it seems only has slf4j-api deps

bowenliang123:
cc @iodone @wForget 

I am looking for proper timing to close event handlers for Kyuubi Server. But repeated errors comes in the irreverent places.

As I tried to close event handlers via `EventBus.close()` in KyuubiSever's `stop()` or `stopServer`,
the `get operation log` test case failed, which is completely unrelated to Kafka logger introduced in this PR, and it also never uses the Kafka logger in context. 

If you guys have any idea for this, feel free to ping me any time. Thanks.

```
java.lang.IllegalStateException: Cannot perform operation after producer has been closed
	at org.apache.kafka.clients.producer.KafkaProducer.throwIfProducerClosed(KafkaProducer.java:892) ~[kafka-clients-2.8.1.jar:?]
	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java:901) ~[kafka-clients-2.8.1.jar:?]
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:885) ~[kafka-clients-2.8.1.jar:?]
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:773) ~[kafka-clients-2.8.1.jar:?]
	at org.apache.kyuubi.events.handler.KafkaLoggingEventHandler.apply(KafkaLoggingEventHandler.scala:50) ~[kyuubi-events_2.12-1.8.0-SNAPSHOT.jar:?]
	at org.apache.kyuubi.events.handler.KafkaLoggingEventHandler.apply(KafkaLoggingEventHandler.scala:31) ~[kyuubi-events_2.12-1.8.0-SNAPSHOT.jar:?]
	at org.apache.kyuubi.events.EventBus$EventBusLive.$anonfun$post$4(EventBus.scala:87) ~[kyuubi-events_2.12-1.8.0-SNAPSHOT.jar:?]
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23) ~[scala-library-2.12.17.jar:?]
- get operation log *** FAILED ***
  scala.collection.JavaConverters.asScalaBufferConverter[String](logRowSet.getLogRowSet()).asScala.exists(((x$1: String) => x$1.contains("show tables"))) was false (OperationsResourceSuite.scala:107)
```

kyuubi-server/src/test/scala/org/apache/kyuubi/events/handler/ServerJsonLoggingEventHandlerSuite.scala
```diff
@@ -197,6 +198,7 @@ class ServerJsonLoggingEventHandlerSuite extends WithKyuubiServer with HiveJDBCT
     server.initialize(conf)
     server.start()
     server.stop()
+    ServerEventHandlerRegister.registerEventLoggers(conf)
```
bowenliang123, line 201: According to @wForget 's advice, stopping the server in line 200 will also clear the event handlers in the singleton of EventBus. Re-register the event loggers here helps for continuing sending to existed event logger in server for the whole testsuite.

bowenliang123: Let's separate the closing event handler to this PR https://github.com/apache/kyuubi/pull/4777.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: Resolve this conversation as https://github.com/apache/kyuubi/pull/4777 close event handlers when stopping server.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: Yes, no available unified event time in KyuubiEvent right now.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
```diff
@@ -0,0 +1,69 @@
...
+      kafkaProducer.send(record)
+    } catch {
+      case e: Exception =>
+        error("Failed to send event in KafkaEventHandler", e)
+    }
+  }
+
+  override def close(): Unit = {
+    kafkaProducer.flush()
+    kafkaProducer.close()
```
cxzl25, line 63: I suggest to use close method with timeout.
```
org.apache.kafka.clients.producer.Producer#close(long, java.util.concurrent.TimeUnit)
```

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventLoggerType.scala
```diff
@@ -22,5 +22,5 @@ object EventLoggerType extends Enumeration {
   type EventLoggerType = Value
 
   // TODO: Only SPARK is done now
-  val SPARK, JSON, JDBC, CUSTOM = Value
+  val SPARK, JSON, JDBC, KAFKA, CUSTOM = Value
```
cxzl25, line 25: Should we add enumeration values instead of modifying the order?

bowenliang123: `KAFKA` value is added before `CUSTOM` here, in order to 
1. follow the same order as in docs
2. `CUSTOM` is an extra option, which could be listed behind the options with embedded support

I'm open to your suggestion. WDYT @pan3793 ?

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: Do you have any suggested conf for the timeout here? As for loggers, for servers, for server event loggers?

Or let's introduce proper config next time ?

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventLoggerType.scala
pan3793: Let's add the new one at the tail, to suppress the potential serializable issue. And the TODO comment is outdated

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
pan3793: let's leave a TODO here, we should add an `event_time` in `KyuubiEvent` and propagate it to Kafka, but it may break the backward compatibility, so better to do it later separately

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
pan3793: It makes sense to introduce a new configuration for it. For server's Kafka logger

kyuubi-server/pom.xml
```diff
@@ -486,6 +503,27 @@
             <artifactId>flexmark-all</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_${scala.binary.version}</artifactId>
```
pan3793, line 509: I think we are using Testcontainers for integration testing, why should we depend on the Kafka server?

And after second thought, I prefer to use `MockProducer` rather than launch a real Kafka broker, but the latter is also fine for me.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventLoggerType.scala
bowenliang123: SGTM. Let me move it to the tail, as you mentioned serializable problem.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: OK. Will do it in two steps.
1. introduce a new config ``, with default value 60 secs.
2. pass the config in `ServerEventHandlerRegister#createKafkaEventHandler`  to construct `KafkaLoggingEventHandler` instance.

pom.xml
```diff
@@ -1261,6 +1268,18 @@
...
+                    <exclusion>
+                        <groupId>org.slf4j</groupId>
+                        <artifactId>slf4j-api</artifactId>
+                    </exclusion>
+                </exclusions>
```
pan3793, line 1280: unnecessary exclusion

kyuubi-server/pom.xml
```diff
@@ -395,6 +395,23 @@
             <artifactId>swagger-ui</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
```
pan3793, line 400: should be provided or optional, avoid to  propagate to the engine moudles

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
pan3793: 60s is too long, the will blocks other shutdown procedures, likely 3~5s

kyuubi-server/pom.xml
bowenliang123: `MockProducer` is lightweight to simulate `producer-consumer` situation, but we need a broker bootstrap server from `producer-borker-consumer` here. And I don't know how to get a bootstrap server from a `MockProducer`.

pom.xml
bowenliang123: why we should skip this exclusion? `kafka-clients` is depending on `slf4j-api`.

kyuubi-server/pom.xml
bowenliang123: SGTM.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: OK, let's make it in 5s by default.

kyuubi-server/pom.xml
pan3793: the purpose of the consumer is to verify the events have been successfully delivered to Kafka, we can simply check the buffer of MockProducer instead

pom.xml
pan3793: We should respect the Maven transitive dependencies in most cases, just excluding the deps we truly don't need, and overriding versions in `dependencyManagement`(just overriding, without excluding)

But it's not suitable for components which is disastrous for dependencies management, like Hive, Hudi.

bowenliang123: SGTM. Removed this exclusion.

NOTICE-binary
```diff
@@ -1236,7 +1236,7 @@ This product optionally depends on 'zstd-jni', a zstd-jni Java compression
 and decompression library, which can be obtained at:
 
   * LICENSE:
-    * license/LICENSE.zstd-jni.txt (Apache License 2.0)
```
pan3793, line 1239: Thanks for fixing this license issue.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2006,12 +2006,34 @@ object KyuubiConf {
...
+      .stringConf
+      .createOptional
+
+  val SERVER_EVENT_KAFKA_CLOSE_TIMEOUT: ConfigEntry[Long] =
+    buildConf("kyuubi.backend.server.event.kafka.close.timeout")
+      .doc("Period to wait for Kafka producer of server event handlers to close.")
+      .version("1.8.0")
+      .serverOnly
+      .timeConf
+      .createWithDefault(Duration.ofMinutes(5000).toMillis)
```
pan3793, line 2023: 5000 minutes? what a crazy duration!

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventLoggerType.scala
```diff
@@ -21,6 +21,5 @@ object EventLoggerType extends Enumeration {
 
   type EventLoggerType = Value
 
-  // TODO: Only SPARK is done now
-  val SPARK, JSON, JDBC, CUSTOM = Value
+  val SPARK, JSON, JDBC, CUSTOM, KAFKA = Value
```
pan3793, line 24: all others is supported now?

bowenliang123: Seems JDBC is not supported.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
bowenliang123: Oops... Shaking hands on the keyboard. 😂 Fixed the time unit to milliseconds.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventLoggerType.scala
bowenliang123: I think this TODO can be removed. Not all the event logger types in both servers or the engine. It's a legacy todo hints.

NOTICE-binary
bowenliang123: 🍻 No problem. And we could make this change to a separate PR if you want.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2006,12 +2006,34 @@ object KyuubiConf {
...
     buildConf("kyuubi.backend.server.event.loggers")
       .doc("A comma-separated list of server history loggers, where session/operation etc" +
         " events go.<ul>" +
         s" <li>JSON: the events will be written to the location of" +
         s" ${SERVER_EVENT_JSON_LOG_PATH.key}</li>" +
+        s" <li>KAFKA: the events serialized in JSON format" +
+        s" and sent to topic of `${SERVER_EVENT_KAFKA_TOPIC.key}`. " +
+        s" Note: For the configs of Kafka producer," +
+        s" please specify them with the prefix: `kyuubi.backend.server.event.kafka.`." +
+        s" For example, kyuubi.backend.server.event.kafka.bootstrap.servers=127.0.0.1:9092" +
```
pan3793, line 2035:
```suggestion
        s" <li>KAFKA: the events will be serialized in JSON format" +
        s" and sent to topic of `${SERVER_EVENT_KAFKA_TOPIC.key}`." +
        s" Note: For the configs of Kafka producer," +
        s" please specify them with the prefix: `kyuubi.backend.server.event.kafka.`." +
        s" For example, `kyuubi.backend.server.event.kafka.bootstrap.servers=127.0.0.1:9092`" +
```

pan3793 (approved):
Thanks, LGTM, only minor things of docs

pan3793 (approved):
Thanks, LGTM, only minor things of docs

bowenliang123:
> Thanks, LGTM, only minor things of docs

Thanks for the detailed reviews. Wait for the CI builds and it will be merged. @pan3793 @cxzl25

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/KafkaLoggingEventHandler.scala
bowenliang123: Btw, it's removed for flushing the producer, for the reasons, the closing producer wakes up the internal sender to send pending requests (exact way in flushing), and the flushing method does not accept timeout.

cxzl25 (approved):
LGTM

bowenliang123:
Thanks, merged to master

PR #4739, opened 2023-04-20
Add operation lock instead of locking state Enumeration

turboFei:

We meet an issue that cause all the operation stuck when closing operation.

Because now all the operations try to lock a Scala Enumeration val.

And if one of them stuck, all the others will be keep stuck.

In this pr, I add a lock for each operation.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -24,6 +24,7 @@ import java.nio.file.{Files, Path, Paths, StandardCopyOption}
 import java.text.SimpleDateFormat
 import java.util.{Date, Properties, TimeZone, UUID}
 import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.locks.Lock
 
 import scala.collection.JavaConverters._
 import scala.sys.process._
@@ -405,6 +406,15 @@ object Utils extends Logging {
       val stringWriter = new StringWriter()
       e.printStackTrace(new PrintWriter(stringWriter))
       stringWriter.toString
+    }
+  }
+
+  def withLockRequired[T](lock: Lock)(block: => T): T = {
+    try {
+      lock.lock()
+      block
+    } finally {
+      lock.unlock()
     }
   }
 }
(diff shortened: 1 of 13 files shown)


turboFei:
cc @yaooqinn @pan3793 @ulysses-you

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
```diff
@@ -45,7 +46,16 @@ abstract class AbstractOperation(session: Session) extends Operation with Loggin
...
-  protected def cleanup(targetState: OperationState): Unit = state.synchronized {
+  private val lock: ReentrantLock = new ReentrantLock()
+
+  protected def withLockRequired[T](block: => T): T = {
+    try {
+      lock.lock()
+      block
+    } finally lock.unlock()
+  }
+
```
yaooqinn, line 57: Can we place it in Utils?

turboFei: done

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -407,4 +408,15 @@ object Utils extends Logging {
       stringWriter.toString
     }
   }
+
+  def withLockRequired[T](lock: Lock)(block: => T): T = {
+    var locked: Boolean = false
+    try {
+      lock.lockInterruptibly()
```
turboFei, line 415: how do you think about? cc @yaooqinn

turboFei: Also cc @pan3793

yaooqinn: The state changing can be safely used `lock`, I do not know other places. BTW, how do we handle InteruptExcetion?

turboFei: Reverted

yaooqinn (approved):
LGTM

pan3793:
`ResourceLock` is definitely a future-proof option

https://github.com/trinodb/trino/issues/17045

turboFei:
thanks, merged to master

there is conflicts with branch-1.7,

pan3793:
Cherry-picked to 1.7

PR #4749, opened 2023-04-21
 Fix flaky test issues in SchedulerPoolSuite

huangzhir:

To fix issue https://github.com/apache/kyuubi/issues/4713, a PR  https://github.com/apache/kyuubi/pull/4714 was submitted, but it had Flaky test issues. After 50 local tests, it succeeded 38 times and failed 12 times.
This PR addresses the issue of flaky tests.


huangzhir:
Here are the test results for 50 local tests:

Before the modification:
total_time: 823, avg_time: 16
total: 50, success: 38, fail: 12

After the modification:
total_time: 916, avg_time: 18
total: 50, success: 50, fail: 0.

pan3793 (approved):

cxzl25 (approved):

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/SchedulerPoolSuite.scala
```diff
@@ -101,6 +101,10 @@ class SchedulerPoolSuite extends WithSparkSQLEngine with HiveJDBCTestHelper {
       })
       threads.shutdown()
       threads.awaitTermination(20, SECONDS)
+      // make sure the SparkListener has received the finished events for job1 and job2.
```
cfmcgrady, line 104: if the root cause is the event not being received, let's call `KyuubiSparkContextHelper.waitListenerBus()` to make sure there are no more events in the spark event queue.

huangzhir: Thank you for pointing out, KyuubiSparkContextHelper.waitListenerBus is indeed a better idea.The average execution time is reduced by 1 second.

total_time: 882 ,avg_time: 17
total: 50, success: 50, fail: 0

pan3793:
Thanks, merged to master/1.7

cfmcgrady:
Good catch! thank you for the fix. @huangzhir

Issue #4745, opened 2023-04-20
[FEATURE]  Support Flink's LocalZonedTimestamp DataType


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

LocalZonedTimestamp is used in the result set of the `show job` statements, which is crucial for Flink job operations.

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #4751, opened 2023-04-22
[KYUUBI #4745] Support Flink's LocalZonedTimestamp DataType

link3280:

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
@@ -21,7 +21,9 @@ import java.{lang, util}
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import java.sql.{Date, Timestamp}
-import java.time.{LocalDate, LocalDateTime}
+import java.time.{Instant, LocalDate, LocalDateTime, ZonedDateTime, ZoneId}
+import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder}
+import java.time.temporal.ChronoField
 import java.util.Collections
 
 import scala.collection.JavaConverters._
@@ -42,15 +44,16 @@ object RowSet {
   def resultSetToTRowSet(
       rows: Seq[Row],
       resultSet: ResultSet,
+      zoneId: ZoneId,
       protocolVersion: TProtocolVersion): TRowSet = {
     if (protocolVersion.getValue < TProtocolVersion.HIVE_CLI_SERVICE_PROTOCOL_V6.getValue) {
-      toRowBaseSet(rows, resultSet)
+      toRowBaseSet(rows, resultSet, zoneId)
     } else {
-      toColumnBasedSet(rows, resultSet)
+      toColumnBasedSet(rows, resultSet, zoneId)
     }
   }
 
-  def toRowBaseSet(rows: Seq[Row], resultSet: ResultSet): TRowSet = {
+  def toRowBaseSet(rows: Seq[Row], resultSet: ResultSet, zoneId: ZoneId): TRowSet = {
     val rowSize = rows.size
     val tRows = new util.ArrayList[TRow](rowSize)
     var i = 0
@@ -60,7 +63,7 @@ object RowSet {
       val columnSize = row.getArity
       var j = 0
       while (j < columnSize) {
-        val columnValue = toTColumnValue(j, row, resultSet)
+        val columnValue = toTColumnValue(j, row, resultSet, zoneId)
         tRow.addToColVals(columnValue)
         j += 1
       }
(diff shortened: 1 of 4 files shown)


externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
```diff
@@ -739,6 +739,23 @@ abstract class FlinkOperationSuite extends HiveJDBCTestHelper with WithFlinkTest
...
 
+  test("execute statement - select timestamp with local time zone") {
+    withJdbcStatement() { statement =>
+      statement.executeQuery("CREATE VIEW T1 AS SELECT TO_TIMESTAMP_LTZ(4001, 3)")
+      statement.executeQuery("SET 'table.local-time-zone' = 'UTC'")
+      val resultSetUTC = statement.executeQuery("SELECT * FROM T1")
+      val metaData = resultSetUTC.getMetaData
+      assert(metaData.getColumnType(1) === java.sql.Types.OTHER)
+      assert(resultSetUTC.next())
+      assert(resultSetUTC.getString(1) === "1970-01-01 00:00:04.001 UTC")
```
link3280, line 750: `TIMESTAMP WITH LOCAL ZONE` should not show the zone part. Something is wrong.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
```diff
@@ -377,4 +398,23 @@ object RowSet {
         other.toString
     }
   }
+
+  /** should stay in sync with [[org.apache.hadoop.hive.common.type.TimestampTZUtil]] */
+  var TIMESTAMP_LZT_FORMATTER: DateTimeFormatter = {
+    val builder = new DateTimeFormatterBuilder()
+    // Date part
+    builder.append(DateTimeFormatter.ofPattern("uuuu-MM-dd"))
```
pan3793, line 406: what does `uuuu-MM-dd` mean?

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
```diff
@@ -377,4 +398,23 @@ object RowSet {
...
+    // Time part
+    builder.optionalStart().appendLiteral(" ").append(
+      DateTimeFormatter.ofPattern("HH:mm:ss")).optionalStart().appendFraction(
+      ChronoField.NANO_OF_SECOND,
+      1,
+      9,
+      true).optionalEnd().optionalEnd()
+    // Zone part
+    builder.optionalStart().appendLiteral(" ").optionalEnd()
+    builder.optionalStart().appendZoneOrOffsetId().optionalEnd()
```
pan3793, line 416: if it's not intended, why add this?

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/result/ResultSetSuite.scala
```diff
@@ -44,9 +46,9 @@ class ResultSetSuite extends KyuubiFunSuite {
       .data(rowsNew)
       .build
 
-    assert(RowSet.toRowBaseSet(rowsNew, resultSetNew)
-      === RowSet.toRowBaseSet(rowsOld, resultSetOld))
-    assert(RowSet.toColumnBasedSet(rowsNew, resultSetNew)
-      === RowSet.toColumnBasedSet(rowsOld, resultSetOld))
+    assert(RowSet.toRowBaseSet(rowsNew, resultSetNew, ZoneId.systemDefault())
```
pan3793, line 49: `ZoneId.systemDefault()` may cause different results in different environments, let's use a specific one, e.g. `America/Los_Angeles`(PST) `Asia/Shanghai`(CST)

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
link3280: Because parsing at JDBC requires the zone part, otherwise we would ran into this exception: https://github.com/apache/kyuubi/blob/06dd7d3754ee210f7744e46030965a6f5df1953b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/common/TimestampTZUtil.java#L84

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/schema/RowSet.scala
link3280: The format is copied from TimestampTZUtil in `kyuubi-hive-jdbc` module. IIUC, `uuuu` has better compatibility with years before `1970`.

link3280:
@pan3793 I started a thread to discuss the timestamp format issue https://www.mail-archive.com/dev@flink.apache.org/msg66257.html. But I don't think it's a blocker for this PR.

pan3793 (approved):
Thanks @link3280, the implementation makes sense to me, and I agree the Flink different behavior should not block us to merge this one.

pan3793:
Thanks, merged to master

Issue #3887, opened 2022-12-02
[Improvement] Provide kyuubiServerPrincipal as alias for principal in JDBC parameter


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Kyuubi JDBC Driver introduced new parameters `kyuubiClientPrincipal` and `kyuubiClientKeytab`in https://github.com/apache/incubator-kyuubi/pull/3023, and it had a parameter `principal` which is inherited from Hive JDBC Driver, it's quite confusing, so we'd better to add an alias  `kyuubiServerPrincipal` for `principal` to make it clear and consistent.

### How should we improve?

Provide `kyuubiServerPrincipal` as alias for `principal` in JDBC parameter

### Are you willing to submit PR?

- [ ] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

lightning-L:
I am interested in this task. Please assign it to me.

pan3793:
Thanks @lightning-L

PR #4764, opened 2023-04-24
[KYUUBI #3887][FOLLOWUP] Fix kyuubiServerPrincipal logic in KyuubiCommands

lightning-L:


lightning-L:
the logic should be && instead of ||.
When either `principal` or `kyuubiServerPrincipal` is not provided, should prompt for username and password.

pan3793:
Thanks, merged to master/1.7

PR #4767, opened 2023-04-25
Correct the submit time for BatchJobSubmission and check applicationInfo if submitted application

turboFei:

- if the kyuubi instance is unreachable, we should not transfer the  batch metadata create time as batch submit time
  - we should always wait the kyuubi instance recovery 
  - here using a fake submit time to prevent that the batch be marked  as terminated if application state is NOT_FOUND
- Inside the BatchJobSubmission, using the first get application info time as batch submit time.

In this pr, I also record whether the batch operation submit batch.

If it did submit batch application and the app is not started, we need to mark the batch state as ERROR.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -80,6 +80,7 @@ class BatchJobSubmission(
   def appStartTime: Long = _appStartTime
 
   private lazy val _submitTime = if (_appStartTime > 0) _appStartTime else System.currentTimeMillis
+  @volatile private var _submitApplication: Boolean = false
 
   @VisibleForTesting
   private[kyuubi] val builder: ProcBuilder = {
@@ -205,7 +206,12 @@ class BatchJobSubmission(
           }.getOrElse {
             submitAndMonitorBatchJob()
           }
-          setStateIfNotCanceled(OperationState.FINISHED)
+
+          if (_submitApplication && _applicationInfo == Some(ApplicationInfo.NOT_FOUND)) {
+            setStateIfNotCanceled(OperationState.ERROR)
+          } else {
+            setStateIfNotCanceled(OperationState.FINISHED)
+          }
         }
       } catch {
         onError()
@@ -227,6 +233,7 @@ class BatchJobSubmission(
   }
 
   private def submitAndMonitorBatchJob(): Unit = {
+    _submitApplication = true
     var appStatusFirstUpdated = false
     var lastStarvationCheckTime = createTime
     try {
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -139,21 +139,20 @@ class BatchJobSubmission(
 
     if (isTerminalState(state)) {
       if (_applicationInfo.isEmpty) {
-        _applicationInfo =
-          Option(ApplicationInfo(id = null, name = null, state = ApplicationState.NOT_FOUND))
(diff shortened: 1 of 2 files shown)


zwangsheng (approved):

turboFei:
thanks @zwangsheng , also cc @pan3793

cxzl25 (approved):

turboFei:
WIP

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -297,7 +297,8 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
               val batchAppStatus = sessionManager.applicationManager.getApplicationInfo(
                 metadata.clusterManager,
                 batchId,
-                Some(metadata.createTime))
+                // prevent that the batch be marked as terminated if application state is NOT_FOUND
+                Some(System.currentTimeMillis()))
```
zwangsheng, line 301: Why not using `createTime`?

turboFei: createTime is the session create time not batch submit time.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -78,6 +78,9 @@ class BatchJobSubmission(
 
   @volatile private var _appStartTime = recoveryMetadata.map(_.engineOpenTime).getOrElse(0L)
   def appStartTime: Long = _appStartTime
+  def appStarted: Boolean = _appStartTime > 0
+
+  private lazy val _submitTime = if (_appStartTime > 0) _appStartTime else System.currentTimeMillis
```
zwangsheng, line 83: if (appStarted)

turboFei: done

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
pan3793: `metadata.engineOpenTime`?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -265,7 +262,19 @@ class BatchJobSubmission(
           throw new KyuubiException(s"Process exit with value ${process.exitValue()}")
         }
 
-        applicationId(_applicationInfo).foreach(monitorBatchJob)
+        while (!appStarted && applicationId(_applicationInfo).isEmpty && !applicationTerminated(
+            _applicationInfo)) {
```
pan3793, line 266:
```suggestion
        while (!appStarted && applicationId(_applicationInfo).isEmpty &&
               !applicationTerminated(_applicationInfo)) {
```

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -265,7 +262,19 @@ class BatchJobSubmission(
...
+          Thread.sleep(applicationCheckInterval)
+          updateApplicationInfoMetadataIfNeeded()
+        }
+
+        applicationId(_applicationInfo) match {
+          case Some(appId) => monitorBatchJob(appId)
+          case None if !appStarted =>
+            throw new RuntimeException(
```
pan3793, line 274: I suppose this is a reachable branch, then use KyuubiException instead of RuntimeException.

turboFei:
all comments addressed. @pan3793

pan3793 (approved):
Looks reasonable to me

turboFei:
thanks, merged to 1.7.1/master

PR #4777, opened 2023-04-27
Deregister event handlers when stopping server with event handler made auto-closeable

bowenliang123:

- deregister event handlers when stopping Kyuubiserver, by deregister them from EventBus's handler Registry
- change `EventHandler` from `type` to `trait` and make it extending  `AutoCloseable`
- implement `close` method in `JsonLoggingEventHandler` for closing writers and streams

### DIFF
--- a/kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
+++ b/kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
@@ -40,6 +40,8 @@ sealed trait EventBus {
   def register[T <: KyuubiEvent: ClassTag](eventHandler: EventHandler[T]): EventBus
 
   def registerAsync[T <: KyuubiEvent: ClassTag](eventHandler: EventHandler[T]): EventBus
+
+  def deregisterAll(): Unit = {}
 }
 
 object EventBus extends Logging {
@@ -67,6 +69,10 @@ object EventBus extends Logging {
 
   def registerAsync[T <: KyuubiEvent: ClassTag](et: EventHandler[T]): EventBus =
     defaultEventBus.registerAsync[T](et)
+
+  def deregisterAll(): Unit = {
+    defaultEventBus.deregisterAll()
+  }
 
   private case class EventBusLive() extends EventBus {
     private[this] lazy val eventHandlerRegistry = new Registry
@@ -96,6 +102,11 @@ object EventBus extends Logging {
       asyncEventHandlerRegistry.register(et)
       this
     }
+
+    override def deregisterAll(): Unit = {
+      eventHandlerRegistry.deregisterAll()
+      asyncEventHandlerRegistry.deregisterAll()
+    }
   }
 
   private class Registry {
@@ -121,6 +132,11 @@ object EventBus extends Logging {
         parent <- getAllParentsClass(cls)
       } yield parent
       clazz :: parents
+    }
+
+    def deregisterAll(): Unit = {
+      eventHandlers.values.flatten.foreach(_.close())
+      eventHandlers.clear()
     }
   }
 }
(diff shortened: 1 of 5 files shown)


kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -122,5 +133,10 @@ object EventBus extends Logging {
       } yield parent
       clazz :: parents
     }
+
+    def deregisterAll(): Unit = {
```
pan3793, line 137: synchronized

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -96,6 +102,11 @@ object EventBus extends Logging {
       asyncEventHandlerRegistry.register(et)
       this
     }
+
+    override def deregisterAll(): Unit = {
```
pan3793, line 106: synchronized

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
```diff
@@ -68,6 +70,10 @@ object EventBus extends Logging {
   def registerAsync[T <: KyuubiEvent: ClassTag](et: EventHandler[T]): EventBus =
     defaultEventBus.registerAsync[T](et)
 
+  def deregisterAll(): Unit = {
+    defaultEventBus.deregisterAll()
```
pan3793, line 74: synchronized

bowenliang123: Which one are you suggesting? Synchronized on the object of EventBus instance or on the defaultEventBus instance?

```
def deregisterAll(): Unit = synchronized {
```
or
```
def deregisterAll(): Unit = defaultEventBus.synchronized {
```

pan3793 (approved):

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/JsonLoggingEventHandler.scala
```diff
@@ -108,6 +119,7 @@ class JsonLoggingEventHandler(
 }
 
 object JsonLoggingEventHandler {
-  val JSON_LOG_DIR_PERM: FsPermission = new FsPermission(Integer.parseInt("770", 8).toShort)
-  val JSON_LOG_FILE_PERM: FsPermission = new FsPermission(Integer.parseInt("660", 8).toShort)
+  private val JSON_LOG_DIR_PERM: FsPermission = new FsPermission(Integer.parseInt("770", 8).toShort)
+  private val JSON_LOG_FILE_PERM: FsPermission =
```
yaooqinn, line 123: why changing these？

yaooqinn:
Can we add some tests?

kyuubi-events/src/main/scala/org/apache/kyuubi/events/handler/JsonLoggingEventHandler.scala
bowenliang123: They could be private only and the changes are optional. This pr is closing the inside intermediateness of the json event handlers, so I think good to retract these two variables in private accessibility.

kyuubi-events/src/main/scala/org/apache/kyuubi/events/EventBus.scala
pan3793: the former

bowenliang123:
> Can we add some tests?

I think the changes in `[ServerJsonLoggingEventHandlerSuite.scala](https://github.com/apache/kyuubi/pull/4777/files#diff-91972dfd64cc4b9aa688957a9b4e2a78f328a1cdc26beae3048a8a62fb9b9a10R201)` reflects the situations. It is equivalent to testing both no messages after clearing up the event handlers and receiving messages by re-registering event handlers again.

wForget (approved):

bowenliang123:
Thanks, merged to master.

PR #4779, opened 2023-04-27
Fix kyuubi-server package missing mysql dependency jar

imperio-wxm:

1. kyuubi-server package lost mysql dependency, causing metadata to use mysql as backend to report an error.
2. After mysql-connector-j 8.x,  `com.mysql.cj.jdbc.Driver` should be used instead of  `com.mysql.jdbc.Driver` .


bowenliang123:
-1, as https://github.com/apache/kyuubi/issues/4778#issuecomment-1526861585

PR #4787, opened 2023-05-05
Log error when handling session exception

turboFei:

Log the session exception when handling it, otherwise, we can only lookup it from events.

image

For example, if the exception thrown by `engine.getOrCreate(discoveryClient, extraEngineLog)`, now there is no log for that.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
@@ -49,6 +49,7 @@ abstract class KyuubiSession(
       f
     } catch {
       case t: Throwable =>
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala
```diff
@@ -49,6 +49,7 @@ abstract class KyuubiSession(
       f
     } catch {
       case t: Throwable =>
+        error(s"Handling $user session[$handle] exception:", t)
```
pan3793, line 52: for exceptions that will be re-thrown, seems not need to print stacktrace here

turboFei: thanks, I see

Issue #2526, opened 2022-04-30
[Improvement] Add grafana dashboard template for kyuubi metrics


yangrong688:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Add grafana dashboard template for kyuubi metrics

### How should we improve?

Add a grafana dashboard (kyuubi_dashboard.json) and related docs.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

Narcasserun:
+1

silvermissile:
any progress ?

kongtrio:
+1

Swarvenstein:
Hello. Any progress on this task? I'm really interested in this one. Thanks in advance for it.

Pandas886:
any progress on this task?

smlHao:
> > any progress on this task?
> 
> I finished the dashboard

@Kyofin  Hi，Could you  share the grafana dashboard json  file ?

dnskr:
Closing this as completed.
This was addressed and resolved in PR https://github.com/apache/kyuubi/pull/5147, see https://github.com/apache/kyuubi/tree/master/grafana.

PR #4799, opened 2023-05-06
[Improvement] Add Grafana dashboard of Kyuubi

Kyofin:

Fininsh the issue : #2526 
Add Grafana dashboard of Kyuubi

### DIFF
--- a/kyuubi-metrics/dashboard.json
+++ b/kyuubi-metrics/dashboard.json
@@ -1,0 +1,884 @@
+{
+    "annotations": {
+      "list": [
+        {
+          "$$hashKey": "object:411",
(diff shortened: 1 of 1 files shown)


pan3793:
How was it tested? And what should they do if other developers want to improve it in the future?

zwangsheng (changes requested):
Thank you for your contributions, but for now, fixing datasource doesn't fit with the idea that this file should be a generic template file.

HaoYang670:
Thanks @Kyofin . I tried your template on Grafana today, it could work but I have to manually update the uid of datasource firstly. like @zwangsheng says, it should be more general to make it really a template.

PR #4719, opened 2023-04-17
Support submission timeout for yarn application manager and get the applicationInfo in-memory

turboFei:
### _Why are the changes needed?_

To prevent the create batch operation stuck.

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] [Run test](https://kyuubi.readthedocs.io/en/master/develop_tools/testing.html#running-tests) locally before make a pull request

PR #4807, opened 2023-05-08
[TEST] Retry more times to fix flaky test "spnego batch rest client"

bowenliang123:

- retry more times for `spnego batch rest client` ut of `BatchRestApiSuite`


pan3793:
The retry should help. #4719 makes the create batch return early, which increases the chance of this test failure

bowenliang123:
Yes, the tests pass with retries.

bowenliang123:
Let me rebase it and run CI again.

pan3793:
The failures caused by other tests, I think the fix works, is it ready to go?

bowenliang123:
Agree. It's good to go.

pan3793 (approved):

PR #4809, opened 2023-05-09
[CHAT] Set session user in ChatGPT request

bowenliang123:

- set session user when opening instance of ChatProvider
- use session user in ChatGPT request, to identify user of message and better monitor and abuse detection by OpenAI report( https://platform.openai.com/docs/api-reference/chat/create#chat/create-user)


pan3793 (approved):

bowenliang123:
Thanks, merged to master .

externals/kyuubi-chat-engine/src/main/scala/org/apache/kyuubi/engine/chat/provider/ChatGPTProvider.scala
```diff
@@ -60,6 +60,8 @@ class ChatGPTProvider(conf: KyuubiConf) extends ChatProvider {
     new OpenAiService(api)
   }
 
+  private var sessionUser: Option[String] = None
```
bowenliang123, line 63: Seems not right to have a shared `sessionUser` crossing sessions in ChatProvider instance.

PR #4813, opened 2023-05-10
Add the fetchorientation parameter to the /v1/operations/:operationId/log interface

huangzhir:

to close https://github.com/apache/kyuubi/issues/4732


huangzhir:
The original OperationLog class had a read(maxRows: Int) method without the FetchOrientation parameter, supporting only the FETCH_NEXT mode for reading data. However, when retrieving logs from the AbstractSession, the result data has the FetchOrientation parameter.
``` scala 
override def fetchResults(
    operationHandle: OperationHandle,
    orientation: FetchOrientation,
    maxRows: Int,
    fetchLog: Boolean): TRowSet = {
  if (fetchLog) {
    sessionManager.operationManager.getOperationLogRowSet(operationHandle, orientation, maxRows)
  } else {
    sessionManager.operationManager.getOperationNextRowSet(operationHandle, orientation, maxRows)
  }
}
``` 
In the original OperationManager code, due to the lack of FetchOrientation support in the OperationLog's read method, the implementation ignored the FetchOrientation parameter.
```scala
def getOperationLogRowSet(
    opHandle: OperationHandle,
    order: FetchOrientation,
    maxRows: Int): TRowSet = {
  val operationLog = getOperation(opHandle).getOperationLog
  operationLog.map(_.read(maxRows)).getOrElse {
    throw KyuubiSQLException(s"$opHandle failed to generate operation log")
  }
}
```
As a result, the operation logs provided to the external interface do not support the FetchOrientation parameter, and it is not possible to review the log data for an operation.

The mentioned pull request adds the FetchOrientation parameter to the OperationLog.read method and introduces support for the FetchOrientation parameter in the /v1/operations/:operationId/log interface.

ulysses-you:
@huangzhir thank you for the work. It seems the performance is poor for `FETCH_PRIOR`, for each request, we will reopen the log file and seek to taget position. Is there a real world use case for it ? If so, we might need to rewrite the code using `RandomAccessFile`. `FETCH_FIRST` looks more useful to me.

huangzhir:
@ulysses-you Thank you for your insightful comments. I concur with you on the significance of both FETCH_FIRST and FETCH_PRIOR. FETCH_FIRST is indeed crucial when it comes to re-reading logs.  Regarding FETCH_PRIOR, although the performance may not be ideal at the moment, its usage cases may be relatively few. Therefore, I propose we retain it for now. If there arises a need in the future, we can look into optimizing it, possibly with your suggestion of using a RandomAccess File. Your feedback is highly appreciated.

huangzhir:
> @huangzhir thank you for the work. It seems the performance is poor for `FETCH_PRIOR`, for each request, we will reopen the log file and seek to taget position. Is there a real world use case for it ? If so, we might need to rewrite the code using `RandomAccessFile`. `FETCH_FIRST` looks more useful to me.

 I looked at the Hive code (link: https://github.com/apache/hive/blob/a5d7f72dfe51fe8119d209bf61c64b8e94f2f42b/ql/src/java/org/apache/hadoop/hive/ql/session/OperationLog.java#LL156C31-L156C35) and noticed that FETCH_PRIOR is not implemented. Therefore, I have removed FETCH_PRIOR and referenced the Hive OperationLog code to rewrite the code. 
@ulysses-you Can you please review it?"

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
```diff
@@ -140,11 +140,13 @@ private[v1] class OperationsResource extends ApiRequestContext with Logging {
...
       @PathParam("operationHandle") operationHandleStr: String,
-      @QueryParam("maxrows") maxRows: Int): OperationLog = {
+      @QueryParam("maxrows") @DefaultValue("100") maxRows: Int,
+      @QueryParam("fetchorientation") @DefaultValue("FETCH_NEXT")
+      fetchOrientation: String): OperationLog = {
     try {
       val rowSet = fe.be.sessionManager.operationManager.getOperationLogRowSet(
         OperationHandle(operationHandleStr),
-        FetchOrientation.FETCH_NEXT,
+        FetchOrientation.withName(fetchOrientation),
```
ulysses-you, line 149: can we fast fail there that throw exception if user requests `FETCH_PRIOR` ?

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -154,14 +167,25 @@ class OperationLog(path: Path) {
...
    */
-  def read(maxRows: Int): TRowSet = synchronized {
+  def read(order: FetchOrientation = FETCH_NEXT, maxRows: Int): TRowSet = synchronized {
     if (!initialized) return ThriftUtils.newEmptyRowSet
-    val (logs, lines) = readLogs(reader, maxRows, maxRows)
+    if (order != FETCH_NEXT && order != FETCH_FIRST) {
+      throw KyuubiSQLException(s"$order in operation log is not supported")
+    }
+    if (order == FETCH_FIRST) {
+      resetReader()
```
ulysses-you, line 186: how about just close seekableReader and create a new seekableReader?

ulysses-you (commented):
thank you @huangzhir , support `FETCH_FIRST` sounds good to me.

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
huangzhir: I think just closing the ‘seekableReader’ and creating a new ‘seekableReader’ may not work because ‘seekableReader’ and ‘reader’ are different objects. Unless all instances of ‘reader’ are also replaced with ‘seekableReader’,

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
huangzhir: done

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -172,6 +196,21 @@ class OperationLog(path: Path) {
     toRowSet(logs)
   }
 
+  private def resetReader(): Unit = {
+    lock.synchronized {
```
ulysses-you, line 200: why not just use `synchronized { ... }` ?

huangzhir: I think that REST log requests can easily cause concurrent operations, so there is a high likelihood of concurrent access to resetReader(). If no lock is used, two scenarios may occur:

1. Data race: Multiple threads simultaneously operate on shared resources such as reader and extraReaders, which can lead to data inconsistency. For example, one thread may be closing the reader and setting it to null, while another thread might attempt to close the same reader, resulting in unexpected exceptions.

3. Inconsistent state: Due to the concurrent access of extraReaders by multiple threads, errors may occur during the execution of extraPaths.foreach(...). For instance, one thread may clear extraReaders after another thread has already created a new BufferedReader, resulting in an inconsistent state.


Therefore, I have decided to add a lock mechanism in that section.

ulysses-you: does not `synchronized { ... }` have the same effect ? other methods are protected by `synchronized {...}`, then if we introduce a new object to lock, I'm afraid there would be deadlock issue.

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -172,6 +196,21 @@ class OperationLog(path: Path) {
...
+      }
+      reader = null
+      closeExtraReaders()
+      extraReaders.clear()
+      extraPaths.foreach(path =>
+        extraReaders += Files.newBufferedReader(path, StandardCharsets.UTF_8))
+    }
+  }
+
   def read(from: Int, size: Int): TRowSet = synchronized {
```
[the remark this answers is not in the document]
huangzhir: I had some additional considerations. Here, there is high concurrency and the probability of issues occurring is very low. I agree with your point that introducing a new object as a lock would add overhead and potentially lead to deadlock issues. I will remove the object lock.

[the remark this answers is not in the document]
huangzhir: done

kyuubi-common/src/test/scala/org/apache/kyuubi/log/SeekableBufferedReaderSuite.scala
```diff
@@ -94,6 +95,80 @@ class SeekableBufferedReaderSuite extends KyuubiFunSuite {
...
+    assert(fetchIter.getPosition == 2)
+
+    fetchIter.fetchNext()
+    assert(fetchIter.getFetchStart == 2)
+    assert(fetchIter.getPosition == 2)
+    assertResult(2 until 3)(getRows(fetchIter, 1))
+    assert(fetchIter.getFetchStart == 2)
+    assert(fetchIter.getPosition == 3)
+
+    fetchIter.fetchPrior(2)
```
ulysses-you, line 114: does the test relate to this pr ? it seems we do not touch `FetchIterator`.

ulysses-you (approved):

ulysses-you:
thank you @huangzhir , merging to master

PR #4818, opened 2023-05-10
[K8S][HELM] Update default Kyuubi version to 1.7.1

dnskr:
The change is needed to use latest stable released version of Kyuubi in the Helm chart by default.


turboFei (approved):

pan3793:
Thanks @dnskr. We need to update docs to note where we should update when bumping the project version. (Currently only pom.xml)

Issue #4821, opened 2023-05-11
[Bug] Possible socket leak in KyuubiSessionImpl


dgzdot:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When openning a session in _KyuubiSessionImpl.open_ , _LaunchEngine_ operation will submit a backgroup task to launch an engine in the cluster, and _KyuubiSessionImpl._client_ will be assigned a value in the future.
Suppose a cluster with limited resources, user A  execute a sql via beeline, then launch engine task pending for resource for a long time, at the same time, beeline was killed for some reason. Kyuubi server will close the session, but the launch engine task still running in then backgroud, once needs met and engine start sucessfully,   _KyuubiSessionImpl._client_  will be assigned and hold the socket with engine, it will not release befor next gc.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #4826, opened 2023-05-11
[KYUUBI #4821] Avoid socket leak in KyuubiSessionImpl

dgzdot:
See [KYUUBI #4821] 

image
Picture indicates that backgroup engine launch task was canceled


turboFei:
how about add the `getBackgroundHandle.cacel` into AbstractOperation::close?

dgzdot:
> how about adding the `getBackgroundHandle.cancel` into AbstractOperation::close or KyuubiOperation::close?
> 
> Or just for LaunchEngine::close.

Adding cancel logic in `LaunchEngine::close` seems more reasonable, since this socket leak promblem occures only in startup. 
If it has no side effect, adding in  AbstractOperation::close is also a good choice. Any suggestions?

bowenliang123:
+1 for AbstractOperation::close.

PR #4829, opened 2023-05-12
Support to expose operation metrics

turboFei:

We need operation metrics to know the operation details.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
@@ -53,6 +53,17 @@ abstract class KyuubiOperation(session: Session) extends AbstractOperation(sessi
 
   def remoteOpHandle(): TOperationHandle = _remoteOpHandle
 
+  @volatile protected var _fetchLogCount = 0L
+  @volatile protected var _fetchResultsCount = 0L
+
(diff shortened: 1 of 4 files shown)


turboFei:
cc @cfmcgrady @wForget @cxzl25

turboFei:
cc @pan3793 have fixed the UT.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
```diff
@@ -53,6 +53,17 @@ abstract class KyuubiOperation(session: Session) extends AbstractOperation(sessi
 
   def remoteOpHandle(): TOperationHandle = _remoteOpHandle
 
+  @volatile protected var _fetchLogCount = 0L
+  @volatile protected var _fetchResultsCount = 0L
+
+  protected[kyuubi] def increaseFetchResultsCount(count: Int, fetchLog: Boolean): Unit = {
```
pan3793, line 59: prefer to break it up, instead of distinguishing it by a flag

pan3793 (approved):
The feature makes sense to me

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
turboFei: done

turboFei:
thanks, merged to master

PR #4836, opened 2023-05-13
Set UncaughtExceptionHandler for thread to log exception

turboFei:

To provide more insight, if there is uncaught exception.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiUncaughtExceptionHandler.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiUncaughtExceptionHandler.scala
@@ -23,10 +23,6 @@ import org.apache.kyuubi.Logging
 
 class KyuubiUncaughtExceptionHandler extends UncaughtExceptionHandler with Logging {
   override def uncaughtException(t: Thread, e: Throwable): Unit = {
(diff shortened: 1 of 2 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/util/KyuubiUncaughtExceptionHandler.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package org.apache.kyuubi.util
+
+import java.lang.Thread.UncaughtExceptionHandler
+
+import org.apache.kyuubi.Logging
+
+class KyuubiUncaughtExceptionHandler extends UncaughtExceptionHandler with Logging {
+  override def uncaughtException(t: Thread, e: Throwable): Unit = {
+    error(s"Exception in thread ${t.getName}", e)
```
pan3793, line 26: Not sure if we need to handle OOM Exception like `SparkUncaughtExceptionHandler` does, but at least let's emphasize "Uncaught exception" in the message

turboFei: added `Uncaught exception`

pan3793 (approved):

turboFei:
thanks, merged to 1.7.2

PR #4837, opened 2023-05-13
Fix operation exception leak issue

turboFei:

For now, if some exception in the operation is not caught, it will cause the operation stuck in pending state.
For example, for `ExecutePython` operation.
https://github.com/apache/kyuubi/blob/474f0972a443c62c3d77a9977d61b59a18a0f0a6/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala#L139-L172

If exception thrown in `withLocalProperties` method, the operation will be stuck in pending state forever.
image

To prevent the operation exception leak, I raised this pr.

In this pr:
- catch the exception in beforeRun
- catch the exception in runInternal
- catch the exception in operation backgroup thread 

BTW: for spark operation `getNextRestultSet` method, `catch onError()` is still needed.

BTW: in this pr, the operation is always submitted in background, if `shouldRunAsync`,  using `getBackgroundHandle.get` to block.

After this pr:
image

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
@@ -167,6 +167,15 @@ abstract class AbstractOperation(session: Session) extends Operation with Loggin
 
   protected def onError(): PartialFunction[Throwable, Unit]
 
+  protected def submitBackgroundOperation(r: Runnable): Future[_] = {
+    val onErrorRunnable = new Runnable {
+      override def run(): Unit = try {
+        r.run()
+      } catch onError()
+    }
+    session.sessionManager.submitBackgroundOperation(onErrorRunnable)
+  }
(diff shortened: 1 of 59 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
```diff
@@ -165,10 +165,34 @@ abstract class AbstractOperation(session: Session) extends Operation with Loggin
...
+    try {
+      beforeRun()
+    } catch {
+      onError()
+    }
+
     try {
       runInternal()
+    } catch {
+      onError()
```
turboFei, line 195: the main change is in this class

turboFei:
cc @cfmcgrady @cxzl25 @pan3793 @bowenliang123

turboFei:
close it, it is too difficult

PR #4838, opened 2023-05-13
Fix spark operation exception leak in `withLocalProperties` method

turboFei:

For `ExecutePython` operation.
https://github.com/apache/kyuubi/blob/474f0972a443c62c3d77a9977d61b59a18a0f0a6/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala#L139-L141

If exception thrown in `withLocalProperties` method(for example, python worker is died), the operation will be stuck in pending state forever.
image

We need to catch exception thrown in `withLocalProperties` method.

After this pr:
image

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
@@ -77,30 +77,31 @@ class ExecutePython(
     OperationLog.removeCurrentOperationLog()
   }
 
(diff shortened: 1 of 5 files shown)


turboFei:
https://github.com/apache/kyuubi/blob/474f0972a443c62c3d77a9977d61b59a18a0f0a6/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala#L139-L157

pan3793 (approved):

cfmcgrady (approved):
Good catch!

turboFei:
thanks, merged to 1.7.2

PR #4840, opened 2023-05-15
Return cached appInfo iif both op and app are terminated

pan3793:

When a batch job fails, there is a chance that the op state changed to ERROR, but engine_state has not been updated, w/ the current implementation, currentApplicationInfo always returns the cached outdated appInfo instead of fetching appInfo from the application manager, thus causing the client can not see the correct app status.

```
mysql> select state, engine_state, peer_instance_closed from metadata;
+-------+--------------+----------------------+
| state | engine_state | peer_instance_closed |
+-------+--------------+----------------------+
| ERROR | RUNNING      |                    0 |
| ERROR | RUNNING      |                    0 |
+-------+--------------+----------------------+
```


turboFei (approved):

pan3793:
Thanks, merged to master/1.7

PR #4845, opened 2023-05-16
[KYUUBI #4807][FOLLOWUP] Fix flaky test  "basic batch rest client"

turboFei:

Using `eventually` to fix the flaky test.


pan3793:
Please rebase, it was partially fixed by another PR just now

bowenliang123:
This problem is encountered in the real world, which puzzled me.
Compared to repeating more times for consistent local log results, I suggest checking the batch status first and then fetching the local log.

turboFei:
> This problem is encountered in the real world, which puzzled me.
> 
> Compared to repeating more times for consistent local log results, I suggest checking the batch status first and then fetching the local log. Let's see what happens with this approach in #4846 .

It is slow, because we need wait the batch finished.

bowenliang123:
Alright then. LGTM.

turboFei:
thanks, merging to master and 1.7.2

PR #4858, opened 2023-05-19
Relax test message assertion in "sync query causes engine crash"

pan3793:

The exception for such network issue changes in JDK 17, but the error message looks reasonable, here we simply relax the test message assertion to make UT happy.


bowenliang123 (approved):
LGTM. Consider attaching some comments for each case if possible.

pan3793:
Thanks, merged to master/1.7

PR #4860, opened 2023-05-19
Support to config multiple session conf advisors

turboFei:


turboFei:
how do you think about this idea? @pan3793

pan3793:
This actually can be achieved by defining a `ChainedSessionConfAdvisor`, and then there are different strategies that can be used for multi advisors, e.g. use the first who returns non-empty configurations; combine configurations returned by all advisors but keep the first/last when duplicated key exists; the latter advisor can / cannot see the configurations generated by the former, etc.

PR #4872, opened 2023-05-22
[KYUUBI #4870] Add kyuubi-util and kyuubi-util-scala modules

### LINKED ISSUE #4870: [Improvement] Create `kyuubi-util` and `kyuubi-util-scala` modules
### Code of Conduct

### Search before asking

Some modules do not depend on `kyuubi-common` since they are designed to be independent w/ Kyuubi server/engine, this cause some code duplication in whole codebase.

### How should we improve?

Create the following  modules

- `kyuubi-util` only depends on JDK and `slf4j-api`
- `kyuubi-util-scala` only depends on `kyuubi-util` and `scala-library`

pan3793:

Close #4870


kyuubi-util-scala/src/test/java/org/apache/kyuubi/tags/IcebergTest.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.scalatest.TagAnnotation;
+
+@TagAnnotation
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.TYPE})
+public @interface IcebergTest {}
```
bowenliang123, line 29: How about moving the tags for tests from `kyuubi-util-scala` module to `kyuubi-util`?

pan3793: it depends on scalatest

Issue #4881, opened 2023-05-25
[Improvement] Incorrect user parameters cause the JDBC Engine to be unavailable


rhh777:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

When the kyuubi.engine.jdbc.connection.url specified database does not exist, the Client will receive a **java.sql.SQLSyntaxErrorException: Unknown database 'ddddd'**, but the JDBCEngine process does not exit. Subsequent Client connections will still be sent to the JDBCEngine, but will not be successful, even if the change to the correct database name, because JDBCEngine will only read the parameters passed in when the process startedThis causes the JDBCEngine to be unavailable until the timeout is destroyed.

My beeline parameters
```
./beeline -u 'jdbc:kyuubi://localhost:10009/default;#
kyuubi.engine.type=jdbc;kyuubi.engine.jdbc.connection.url=jdbc:mysql://localhost:13306/ddddd;
kyuubi.engine.jdbc.connection.user=root;
kyuubi.engine.jdbc.connection.password=root;
kyuubi.engine.jdbc.driver.class=com.mysql.cj.jdbc.Driver;
kyuubi.engine.jdbc.extra.classpath=/Users/hrh/.m2/repository/mysql/mysql-connector-java/8.0.13/mysql-connector-java-8.0.13.jar;
kyuubi.engine.jdbc.connection.provider=DorisConnectionProvider;
kyuubi.engine.jdbc.type=doris'
```
<img width="2222" alt="image" src="https://github.com/apache/kyuubi/assets/41908930/6d7b3d9b-5a5f-494e-b191-1806b0f4ad7f">

### How should we improve?

Can we determine at JDBCSession.close that if the sessionConnection is not initialized successfully, the process exits

https://github.com/apache/kyuubi/blob/master/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/session/JdbcSessionImpl.scala#L76

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

rhh777:
[...]

PR #4886, opened 2023-05-25
[KYUUBI #4881] JDBCEngine performs initialization sql

rhh777:

close https://github.com/apache/kyuubi/issues/4881


externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/JdbcDialect.scala
```diff
@@ -65,6 +65,8 @@ abstract class JdbcDialect extends SupportServiceLoader with Logging {
 
   def getSchemaHelper(): SchemaHelper
 
+  def initializationSQLs(): Seq[String]
```
pan3793, line 68: why not give it a default implement `SELECT 1`? then we don't need to overwrite it in each derived classes.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/util/KyuubiJdbcUtil.scala
```diff
@@ -0,0 +1,67 @@
...
+import scala.util.{Failure, Success, Try}
+
+import org.apache.kyuubi.{KyuubiSQLException, Logging}
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.engine.jdbc.connection.ConnectionProvider
+import org.apache.kyuubi.engine.jdbc.dialect.{JdbcDialect, JdbcDialects}
+
+object KyuubiJdbcUtil extends Logging {
+
+    def initializeJdbcSession(kyuubiConf: KyuubiConf): Unit = {
```
pan3793, line 31: - if a method is used in only one place, creating a `Util` object is not a good practice.

- Kyuubi uses `AbcUtils` style, not `AbcUtil`.

- `org.apache.kyuubi.util.JdbcUtils#withCloseable` can simplify code

rhh777: > * if a method is used in only one place, creating a `Util` object is not a good practice.
> * Kyuubi uses `AbcUtils` style, not `AbcUtil`.
> * `org.apache.kyuubi.util.JdbcUtils#withCloseable` can simplify code
* emm, I am considering whether to add a method to execute Session Init SQL in Util
* I refer to the **KyuubiSparkUtil** format before, I will update later
* I'll look at the **JdbcUtils** code in more detail later

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/JdbcDialect.scala
rhh777: > why not give it a default implementation `SELECT 1`? then we don't need to overwrite it in each derived classes.

It's a good idea to default to SELECT 1, updated later

pan3793:
BTW, we provide `dev/reformat` to quickly fix the style.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/JdbcDialect.scala
```diff
@@ -65,6 +65,10 @@ abstract class JdbcDialect extends SupportServiceLoader with Logging {
 
   def getSchemaHelper(): SchemaHelper
 
+  def initializationSQLs(): Seq[String] = {
```
bowenliang123, line 68: The initialization SQLs should be configurable.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/util/KyuubiJdbcUtils.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.kyuubi.engine.jdbc.dialect.{JdbcDialect, JdbcDialects}
+import org.apache.kyuubi.util.JdbcUtils
+
+object KyuubiJdbcUtils extends Logging {
+
+  def initializeJdbcSession(kyuubiConf: KyuubiConf): Unit = {
+    try {
+      val dialect: JdbcDialect = JdbcDialects.get(kyuubiConf)
+      JdbcUtils.withCloseable(ConnectionProvider.create(kyuubiConf)) { connection =>
+        JdbcUtils.withCloseable(dialect.createStatement(connection, 100)) { statement =>
```
bowenliang123, line 32: The timeout for executing init SQL should be configurable.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/PhoenixDialect.scala
```diff
@@ -150,4 +150,5 @@ class PhoenixDialect extends JdbcDialect {
   override def name(): String = {
     "phoenix"
   }
+
```
bowenliang123, line 153: Remove the added blank line.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/DorisDialect.scala
```diff
@@ -162,4 +162,5 @@ class DorisDialect extends JdbcDialect {
   override def name(): String = {
     "doris"
   }
+
```
bowenliang123, line 165: ditto.

rhh777: > ditto.

Updated.

pan3793 (approved):
The implementation looks good to me, but I don't think the code should be put into `KyuubiJdbcUtils`, the code is not reusable totally, and it is compose of specific logic, not generic enough to put into a `*Utils`.

I would suggest move the `initializeJdbcSession` into `JdbcSQLEngine` directly. But if you consist, I'm fine to merge it as-is.

rhh777:
> The implementation looks good to me, but I don't think the code should be put into `KyuubiJdbcUtils`, the code is not reusable totally, and it is compose of specific logic, not generic enough to put into a `*Utils`.
> 
> I would suggest move the `initializeJdbcSession` into `JdbcSQLEngine` directly. But if you consist, I'm fine to merge it as-is.

Do you mean that `KyuubiJdbcUtils` methods are all move to `JdbcSQLEngine`? `JdbcSessionImpl open()` calls the `initializeJdbcSession` method of `JdbcSQLEngine`?

pan3793:
Yes, all things in `KyuubiJdbcUtils`. And the second method can be inlined, I don't see the benefit of extracting it as a method.

rhh777:
> Yes, all things in `KyuubiJdbcUtils`. And the second method can be inlined, I don't see the benefit of extracting it as a method.

After `JdbcSQLEngine` is initialized, the connection can be closed. However, after `JdbcSessionImpl open()` is initialized, the connection cannot be closed and will be used by the sql of subsequent users. So the `initializeJdbcSession` method is split.

cxzl25 (changes requested):
Overall LGTM.
Please use the following command to update `settings.md`
```bash
KYUUBI_UPDATE=1 build/mvn clean test -pl kyuubi-server -am -Pflink-provided,spark-provided,hive-provided -Dtest=none -DwildcardSuites=org.apache.kyuubi.config.AllKyuubiConfiguration
```

rhh777:
> Overall LGTM. Please use the following command to update `settings.md`
> 
> ```shell
> KYUUBI_UPDATE=1 build/mvn clean test -pl kyuubi-server -am -Pflink-provided,spark-provided,hive-provided -Dtest=none -DwildcardSuites=org.apache.kyuubi.config.AllKyuubiConfiguration
> ```

Updated

cxzl25 (approved):

pan3793:
Thanks, merged to master

PR #4898, opened 2023-05-29
fix  logOperation multiple read with missing line

huangzhir:

to close https://github.com/apache/kyuubi/issues/4897

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
@@ -133,8 +133,10 @@ class OperationLog(path: Path) {
       var line: String = reader.readLine()
       while ((i < lastRows || maxRows <= 0) && line != null) {
         logs.add(line)
-        line = reader.readLine()
         i += 1
+        if(i < lastRows || maxRows <= 0) {
+          line = reader.readLine()
+        }
       }
       (logs, i)
     } catch {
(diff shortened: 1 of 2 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -133,8 +133,10 @@ class OperationLog(path: Path) {
       var line: String = reader.readLine()
       while ((i < lastRows || maxRows <= 0) && line != null) {
         logs.add(line)
-        line = reader.readLine()
         i += 1
+        if (i < lastRows || maxRows <= 0) {
```
pan3793, line 137: thanks for catch this issue!

is `do { ... } while` pattern more suitable here?

pan3793:
```
    val logs = new JArrayList[String]
    var rows = 0
    try {
      var line: String = null
      do {
        line = reader.readLine()
        if (line != null) { logs.add(line); rows += 1 }
      } while ((rows < lastRows || maxRows <= 0) && line != null)
      (logs, rows)
    } catch {
      ...
    }
```

huangzhir: than you, this is better solution

pan3793 (approved):

pan3793:
Test failure is not related, thanks, merged to master/1.7

PR #4905, opened 2023-05-30
Generalize util method for loading class from service loader

bowenliang123:


pan3793 (approved):

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
```diff
@@ -69,27 +69,12 @@ abstract class AbstractConnectionProvider extends Logging {
...
-        case t: Throwable =>
-          warn(s"Loaded of the provider failed with the exception", t)
-      }
-    }
-
-    // TODO support disable provider
-    providers
-  }
+        provider
+      }.toSeq
```
bowenliang123, line 77: Could you have a look at this change? Trying to simplify the loading and looping of JdbcConnectionProvider. And skipping the warning message as there's no instance construction and no possible thrown exception.

cc @zhouyifan279

bowenliang123:
Thanks, merged to master.

PR #4915, opened 2023-06-01
Adapt database not exist error message change for Spark 3.4

pan3793:

Spark is migrating to the error class framework, in SPARK-40360(fixed in 3.4.0) SCHEMA_NOT_FOUND was introduced, which changes the output of the error message on switching a not existing database.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSessionImpl.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSessionImpl.scala
@@ -74,7 +74,7 @@ class SparkSessionImpl(
       } catch {
         case e
             if database == "default" && e.getMessage != null &&
-              e.getMessage.contains("not found") =>
+              (e.getMessage.contains("not found") || e.getMessage.contains("SCHEMA_NOT_FOUND")) =>
       }
     }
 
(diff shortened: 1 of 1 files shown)


cfmcgrady:
shall we address L66 as well?

pan3793:
> shall we address L66 as well?

Seems not necessary, because we catch can rethrow the exception

```
  override def setCurrentCatalog(spark: SparkSession, catalog: String): Unit = {
    // SPARK-36841(3.3.0) Ensure setCurrentCatalog method catalog must exist
    if (spark.sessionState.catalogManager.isCatalogRegistered(catalog)) {
      spark.sessionState.catalogManager.setCurrentCatalog(catalog)
    } else {
      throw new IllegalArgumentException(s"Cannot find catalog plugin class for catalog '$catalog'")
    }
  }
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSessionImpl.scala
```diff
@@ -74,7 +74,7 @@ class SparkSessionImpl(
       } catch {
         case e
             if database == "default" && e.getMessage != null &&
-              e.getMessage.contains("not found") =>
+              (e.getMessage.contains("not found") || e.getMessage.contains("SCHEMA_NOT_FOUND")) =>
```
bowenliang123, line 77:
```
          if database == "default"
            && StringUtils.containsAny(e.getMessage, "not found", "SCHEMA_NOT_FOUND") =>
```

pan3793: adopted

bowenliang123 (approved):
LGTM.
And how about adding `[LINEAGE]` in the title?

pan3793:
This PR is not related to lineage. If you are asking if it's ok to introduce a common label of `[LINEAGE]` for other PRs, I'm fine as long as the label is short and straightforward.

pan3793:
Thanks, merged to master/1.7

PR #4921, opened 2023-06-04
[KYUUBI #4920] Add OperationRestApi

lightning-L:

close https://github.com/apache/kyuubi/issues/4920


lightning-L:
to close https://github.com/apache/kyuubi/issues/4602

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
```diff
@@ -81,6 +81,9 @@ private[v1] class OperationsResource extends ApiRequestContext with Logging {
       }
       Response.ok().build()
     } catch {
+      case e: KyuubiSQLException =>
+        error(e.getMessage, e)
+        throw e
```
turboFei, line 86: also cc @pan3793

turboFei (approved):
LGTM

cxzl25 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/OperationsResource.scala
pan3793: is that aim to address no error log issue? I also find the same issue and I think we should log it in `RestExceptionMapper` for all exceptions

pan3793: Open #4977 to address it generally.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/KyuubiOperationEvent.java
```diff
@@ -0,0 +1,343 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.client.api.v1.dto;
+
+import java.util.Map;
+
+public class KyuubiOperationEvent implements KyuubiEvent {
```
pan3793, line 22: Do we need the interface KyuubiEvent? what's the relationship between the client-side KyuubiEvent and server-side KyuubiEvent

lightning-L: yes, we do not need the interface actually. Have removed it.

pan3793:
Thanks, merged to master

PR #4948, opened 2023-06-11
[BUILD] Reuse versions of NodeJS and Pnpm from maven properties

bowenliang123:

- fetch versions for NodeJS and Pnpm from Maven properties
- enabled npm dependency caching for global packages and web-ui module, as the step above takes extra time in builds.


pan3793 (approved):

bowenliang123:
`mvn help:evaluate` is notably slow for getting maven properties. Over 30s taken for fetching 2 versions from the pom file. Try improve the performance if possible in the future.

PR #4949, opened 2023-06-11
For operation getNextRowSet method, with operation lock required 

turboFei:

For the operation getNextRowSet method, we shall add lock for it.

For example, for spark operation, the result iterator is not thread-safe, it might throw exception(if the jdbc client to kyuubi server connection socket timeout).

For incremental collect mode, the fetchResult might trigger a spark task to collect the incremental result(`self.next().toIterator`).

The jdbc client to kyuubi gateway timeout, but the fetchResult request has been sent to engine.
Then the jdbc client re-send the fetchResult request.

And the getNextResultSet in spark engine side concurrent execute.

And the result iterator is not thread-safe and might cause NPE.

image
image

### DIFF
--- a/externals/kyuubi-chat-engine/src/main/scala/org/apache/kyuubi/engine/chat/operation/ChatOperation.scala
+++ b/externals/kyuubi-chat-engine/src/main/scala/org/apache/kyuubi/engine/chat/operation/ChatOperation.scala
@@ -31,7 +31,7 @@ abstract class ChatOperation(session: Session) extends AbstractOperation(session
 
   protected lazy val conf: KyuubiConf = session.sessionManager.getConf
 
-  override def getNextRowSet(order: FetchOrientation, rowSetSize: Int): TRowSet = {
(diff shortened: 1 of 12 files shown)


pan3793:
> ... it might throw exception(if the jdbc client to kyuubi server connection socket timeout).

The change makes sense to me, but I don't get your point. JDBC is the single thread model, why connection socket timeout could cause NPE?

turboFei:
> why connection socket timeout could cause NPE?

For incremental collect mode, the fetchResult might trigger a spark task to collect the incremental result.

The jdbc client to kyuubi gateway timeout, but the fetchResult request has been sent to engine.
Then the jdbc client re-send the fetchResult request.

And the getNextResultSet in spark engine side concurrent execute.

pan3793 (approved):
LGTM, please update the PR description with the detail analysis

cxzl25 (approved):

turboFei:
thanks, merged to 1.7.2

PR #4950, opened 2023-06-12
Migrate Kyuubi embedded database from Derby to SQLite

pan3793:

Apache Derby is no longer active. One major drawback of Derby is that it does not support multiple connections to single db files, making it hard to analyze the data on local development.

SQLite may be the most popular embedded DBMS in the world. It lives almost in every smartphone (at least Android and iOS integrate SQLite), which means SQLite is quite stable and may be a good choice for standalone production deployment.

SQLite provides a CLI command `sqlite3` which is easy to use to connect a data file and run queries. Multi connections to a single db file is allowed, which helps a lot to analyze the data when the Kyuubi server is running.


pan3793:
cc @turboFei @bowenliang123 @yaooqinn

kyuubi-server/src/main/resources/sql/sqlite/001-KYUUBI-3967.sqlite.sql
```diff
@@ -0,0 +1,47 @@
+SELECT '< KYUUBI-3967: Shorten column varchar length of metadata table >' AS ' ';
+
+BEGIN;
```
pan3793, line 3: Unfortunately, SQLite does not support `ALTER TABLE MODIFY COLUMN`, it was explained at 
https://www3.sqlite.org/matrix/lang_altertable.html

> "Why ALTER TABLE is such a problem for SQLite"

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStoreConf.scala
```diff
@@ -37,7 +37,9 @@ object JDBCMetadataStoreConf {
   val METADATA_STORE_JDBC_DATABASE_TYPE: ConfigEntry[String] =
     buildConf("kyuubi.metadata.store.jdbc.database.type")
       .doc("The database type for server jdbc metadata store.<ul>" +
-        " <li>DERBY: Apache Derby, JDBC driver `org.apache.derby.jdbc.AutoloadedDriver`.</li>" +
+        " <li>(Deprecated) DERBY: Apache Derby, JDBC driver " +
+        "`org.apache.derby.jdbc.AutoloadedDriver`.</li>" +
+        " <li>SQLITE: SQLite3, JDBC driver `org.sqlite.JDBC`.</li>" +
```
cfmcgrady, line 42: shall we make a note in the documentation to indicate that the support for SQLite is available starting from version 1.8.0?

bowenliang123: Good to be left as it is. The SQLITE option is not shown on the docs of 1.7.x after all.

pan3793: We don't do that for other similar cases, and we provided versioned docs.

docs/deployment/migration-guide.md
```diff
@@ -17,6 +17,13 @@
 
 # Kyuubi Migration Guide
 
+## Upgrading from Kyuubi 1.7 to 1.8
+
+* Since Kyuubi 1.8, SQLite is added and becomes the default data type of Kyuubi metastore, Derby is deprecated.
```
bowenliang123, line 22:
```suggestion
* Since Kyuubi 1.8, SQLite is added and becomes the default database type of Kyuubi metastore, as Derby has been deprecated.
```

turboFei (approved):

PR #4953, opened 2023-06-13
[KYUUBI #4950][FOLLOWUP] Misc improvement for usage of SQLite

pan3793:

- Improve log on executing init schema DDL.
- Add `IF NOT EXISTS` on `CREATE INDEX` statement.
- Fix `kyuubi.metadata.store.jdbc.url` configuration description.


bowenliang123 (approved):

kyuubi-server/src/main/resources/sql/sqlite/metadata-store-schema-1.6.0.sqlite.sql
```diff
@@ -26,10 +26,10 @@ CREATE TABLE IF NOT EXISTS metadata(
     peer_instance_closed boolean default '0' -- closed by peer kyuubi instance
 );
```
cfmcgrady, line 27: curious, as `JDBCMetadataStore#getLatestSchemaUrl` always return the latest version of metadata store schema, and the SQLite support has been available since v1.8.0, do we still need to provide `metadata-store-schema-1.6.0.sqlite.sql`?

pan3793: You are right, we don't need that. Let's fix it and remove it later, to provide a valid example for future changes.

PR #4963, opened 2023-06-14
[KYUUBI #4937] Cleanup spark catalog shim and renamed to catalog utils

### LINKED ISSUE #4937: Remove SparkCatalogShim in Spark engine
### Code of Conduct

### Search before asking

`SparkCatalogShim` was introduced as a shim layer for compatibility with Spark 2.4, but it does not happen and has less opportunity to happen in the future, let's remove it to simply the code.

### How should we improve?

_No response_

bowenliang123:

to close #4937.


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/util/SparkCatalogUtils.scala
```diff
@@ -0,0 +1,366 @@
...
+
+  private def getSchemasWithPattern(catalog: CatalogPlugin, schemaPattern: String): Seq[String] = {
+    val p = schemaPattern.r.pattern
+    listAllNamespaces(catalog).flatMap { ns =>
+      val quoted = ns.map(quoteIfNeeded).mkString(".")
+      if (p.matcher(quoted).matches()) {
+        Some(quoted)
+      } else {
+        None
+      }
```
pan3793, line 153:
```suggestion
      if (p.matcher(quoted).matches()) Some(quoted) else None
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/util/SparkCatalogUtils.scala
```diff
@@ -0,0 +1,366 @@
...
+        }
+      case tc: TableCatalog =>
+        val tp = tablePattern.r.pattern
+        val identifiers = namespaces.flatMap { ns =>
+          tc.listTables(ns).filter(i => tp.matcher(quoteIfNeeded(i.name())).matches())
+        }
+        identifiers.map { ident =>
+          // TODO: restore view type for session catalog
+          val comment = if (ignoreTableProperties) ""
+          else tc.loadTable(ident).properties().getOrDefault(TableCatalog.PROP_COMMENT, "")
```
pan3793, line 206: It was reported by dozen users that metadata refresh takes a long time, let's add the comment to the code to make the learner easy to understand.

```suggestion
          val comment = if (ignoreTableProperties) ""
          else { // load table is a time consuming operation
            tc.loadTable(ident).properties().getOrDefault(TableCatalog.PROP_COMMENT, "")
          }
```

bowenliang123: Adopted.

pan3793:
A related PR may worth to evaluate https://github.com/apache/kyuubi/pull/4465

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/util/SparkCatalogUtils.scala
```diff
@@ -0,0 +1,375 @@
...
+  def getCatalogs(spark: SparkSession): Seq[Row] = {
+
+    // A [[CatalogManager]] is session unique
+    val catalogMgr = spark.sessionState.catalogManager
+    // get the custom v2 session catalog or default spark_catalog
+    val sessionCatalog = invokeAs[AnyRef](catalogMgr, "v2SessionCatalog")
+    val defaultCatalog = catalogMgr.currentCatalog
+
+    val defaults = Seq(sessionCatalog, defaultCatalog).distinct.map(catalog =>
+      DynMethods.builder("name").impl(catalog.getClass).buildChecked(catalog).invoke[String]())
```
bowenliang123, line 60: While `ReflectUtils.invokeAs` supports declared method only, I have to use DynMethods for calling method of inherited methods here manually.  cc @pan3793

bowenliang123:
> A related PR may worth to evaluate #4465

Seems not enough evidence for the necessity of the suggested changes in that PR.
As for me, checking the catalog name to `spark_catalog` is more strict than matching in TableCatalog class, and it's appropriate to be placed in the first matching case here.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/util/SparkCatalogUtils.scala
pan3793: Is it possible to access all methods by an unified reflection api?

bowenliang123: Shall I refactor the `ReflectUtils.invokeAs` for this, by calling declared methods first and then the inherited methods? And I will throw corresponding method candidates to either one.

pan3793: sgtm

pan3793:
> Seems not enough evidence for the necessity of the suggested changes in that PR.

The key is order, there were some issues(see https://github.com/apache/kyuubi/pull/4219) when `DelegatingCatalogExtension` is used as `spark_catalog`.

pan3793:
The change of this PR is already big, it's valuable to do some further changes in follow-up PRs.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/util/SparkCatalogUtils.scala
bowenliang123: OK, let's make it in the  follow-ups.

pan3793 (approved):

bowenliang123:
Thanks, merged to master.

PR #4970, opened 2023-06-15
[WIP] Unified invokeAs

bowenliang123:

- comment https://github.com/apache/kyuubi/pull/4963#discussion_r1230490326
- simplify reflection reflection method calling with unified `invokeAs` method for either declared, inherited, or static methods


pan3793 (approved):

kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
```diff
@@ -70,20 +78,26 @@ object ReflectUtils {
...
-      DynMethods.builder(methodName)
-        .hiddenImpl(targetClass, argClasses: _*)
-        .buildChecked(target)
-        .invoke[T](args.map(_._2): _*)
+      val method = DynMethods.builder(methodName)
+        .hiddenImpl(clz, argClasses: _*)
+        .impl(clz, argClasses: _*)
+        .buildChecked
+      if (method.isStatic) {
+        method.asStatic.invoke[T](args.map(_._2): _*)
```
pan3793, line 89: not sure if it's a good idea to include static case

pan3793: I lean to separate the static case to dedicated method

bowenliang123: how about the static fields. Separate a method for them as well？

bowenliang123: After my second thought, a dedicated method for static class may cause confusion for usage, eg. a method of Sacla object class is not static. The usages for both object/static class and common classes are shown in test cases which clear enough to understand and use. I would keep it as it is with auto detecting for whether it's a static one.

PR #4977, opened 2023-06-19
Log error message when REST API invocation error occurs

pan3793:

Currently, there is no error message on Kyuubi server's log when REST API invocation error occurs, and since we are building REST API, all response media types should be JSON.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/api.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/api.scala
@@ -42,17 +42,17 @@ private[api] trait ApiRequestContext {
 @Provider
 class RestExceptionMapper extends ExceptionMapper[Exception] with Logging {
   override def toResponse(exception: Exception): Response = {
-    error("Error occurs on accessing REST API.", exception)
+    warn("Error occurs on accessing REST API.", exception)
     exception match {
       case e: WebApplicationException =>
         Response.status(e.getResponse.getStatus)
           .`type`(MediaType.APPLICATION_JSON)
-          .entity(Map("errorMessage" -> e.getMessage))
+          .entity(Map("message" -> e.getMessage))
           .build()
       case e =>
         Response.status(Response.Status.INTERNAL_SERVER_ERROR)
           .`type`(MediaType.APPLICATION_JSON)
-          .entity(Map("errorMessage" -> e.getMessage))
+          .entity(Map("message" -> e.getMessage))
           .build()
     }
   }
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/api.scala
```diff
@@ -39,18 +40,19 @@ private[api] trait ApiRequestContext {
 }
 
 @Provider
-class RestExceptionMapper extends ExceptionMapper[Exception] {
+class RestExceptionMapper extends ExceptionMapper[Exception] with Logging {
   override def toResponse(exception: Exception): Response = {
+    error("Error occurs on accessing REST API.", exception)
```
ulysses-you, line 45: `warn`

ulysses-you (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/api.scala
```diff
@@ -39,18 +40,19 @@ private[api] trait ApiRequestContext {
 }
 
 @Provider
-class RestExceptionMapper extends ExceptionMapper[Exception] {
+class RestExceptionMapper extends ExceptionMapper[Exception] with Logging {
   override def toResponse(exception: Exception): Response = {
+    warn("Error occurs on accessing REST API.", exception)
```
cxzl25, line 45: `trino.api` also has a similar `RestExceptionMapper`
Can it be updated together?

https://github.com/apache/kyuubi/blob/c5238535f6c8d70ac97efc4c1d5dee93765fef92/kyuubi-server/src/main/scala/org/apache/kyuubi/server/trino/api/api.scala#L42-L45

turboFei (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/api.scala
pan3793: Thanks, updated

cxzl25 (approved):

lightning-L (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/trino/api/api.scala
```diff
@@ -39,18 +40,19 @@ private[api] trait ApiRequestContext {
 }
 
 @Provider
-class RestExceptionMapper extends ExceptionMapper[Exception] {
+class RestExceptionMapper extends ExceptionMapper[Exception] with Logging {
   override def toResponse(exception: Exception): Response = {
+    warn("Error occurs on accessing REST API.", exception)
```
pan3793, line 45:
```suggestion
    warn("Error occurs on accessing Trino API.", exception)
```

PR #5016, opened 2023-07-04
Support ignoring custom app name for Spark engine

bowenliang123:

- support to disable custom app name for spark engine, for either share level of USER/GROUP/SERVER. As for CONNECTION share level, custom app name for spark engine is still allowed.


link3280:
I'm curious to know why users set the custom app name in the first place if they want the default one?

bowenliang123:
It happened to set a custom spark app name for one time task on the user/client side without notifying the Kyuubi's administrator. Plus, he's using USER / SERVER share level and the app name was contaminated as the one time data task name. From the perspective of Kyuubi's administrator, it's hard to tell it's a shared spark SQL engine on YARN with such custom app name.

Therefore, this feature mainly enables the administrator of Kyuubi and the resource provider, to reject or ignore custom app name with the share level of USER / SERVER / GROUP.

link3280:
@bowenliang123 Thanks for your input! It might apply to Flink as well. Perhaps I should make a related PR for the Flink engine later.

bowenliang123:
No problem, it was a real case in my production environment. It's embarrassinging that our administrator found it difficult to identify the target YARN app of engine instance in the list of mixed app names of MR/Spark/and etc when he was investigating other engine issues.

And let's see how other users interested in this minor feature.

pan3793:
Seems we have `kyuubi.session.conf.ignore.list`, the only difference is it applies to all share levels?

bowenliang123:
Yes. The CONNECTION level should not be limited to custom app name.
And this feature may be extended to combine share level and the customer app name as the final app name, if needed in the future.

pan3793:
How about introducing a `kyuubi.session.conf.sharable.engine.ignore.list`?

bowenliang123:
> How about introducing a `kyuubi.session.conf.sharable.engine.ignore.list`?

It's a considerable approach for this.
It's introducing a new concept `sharable engine` or `sharable share level`. And this config may mix configs for different engine types as well.

PR #5020, opened 2023-07-05
Fix kubernetes pod leak issue

turboFei:

Before this pr, there is pod leak.

This pr will try to delete the pod with label if no appInfo found.


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -271,4 +281,6 @@ object KubernetesApplicationOperation extends Logging {
         "mark the application state as UNKNOWN.")
       UNKNOWN
   }
+
+  def tagToKubernetesPodName(tag: String): String = s"$KUBERNETES_POD_NAME_PREFIX$tag"
```
pan3793, line 285: can we add a `-driver` suffix?

this simple way should work, I'm not sure if the user will compliant about this behavior,

PR #5023, opened 2023-07-05
[KSHC] TableIdentify don't attach catalog

Yikf:

As title, In KSHC, HiveTable's identify does not attach the catalog to prevent an incorrect catalogName. default catalog is "spark_catalog"

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
@@ -43,6 +43,7 @@ import org.apache.spark.sql.internal.StaticSQLConf.{CATALOG_IMPLEMENTATION, GLOB
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
+import org.apache.kyuubi.spark.connector.hive.HiveConnectorUtils.withSQLConf
 import org.apache.kyuubi.spark.connector.hive.HiveTableCatalog.{toCatalogDatabase, CatalogDatabaseHelper, IdentifierHelper, NamespaceHelper}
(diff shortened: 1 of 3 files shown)


extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -103,6 +103,7 @@ class HiveTableCatalog(sparkSession: SparkSession)
       s"Require setting ${CATALOG_IMPLEMENTATION.key} to `hive` to enable hive support.")
     catalogName = name
     catalogOptions = options
+    sessionState.conf.setConfString("spark.sql.legacy.v1IdentifierNoCatalog", "true")
```
ulysses-you, line 106: this is only used in SessionCatalog, does it really affect v2 catalog code path ?

Yikf: Theoretically not, but HiveCatalog depends on the SessionCatalog...

ulysses-you: to be conservative, can we wrap all session catalog methods with:
```
withSQLConf("spark.sql.legacy.v1IdentifierNoCatalog" -> "true") {
  ...
}
```

I'm afraid it would affect session catalog behavior if people use both kyuubi hive catalog and vanilla spark session catalog.

Yikf: Good point, thanks

ulysses-you (approved):

ulysses-you:
thanks, merging to master

Yikf:
Thanks, merging master

PR #5025, opened 2023-07-06
Use formatDuration instead of formatDurationVerbose in SparkUI's engine tab

bowenliang123:

- previously used `formatDurationVerbose` generates duration with dedicated unit parts
- Spark UI itself also uses `formatDuration` in its Job/Stage pages for displaying duration
   - `JobDataUtil.getFormattedDuration` in AllJobsPage, and `formatDuration` used in `getFormattedDuration` , https://github.com/apache/spark/blob/v3.3.2/core/src/main/scala/org/apache/spark/ui/jobs/JobDataUtil.scala#L32

Before:
image

After:
image

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EnginePage.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EnginePage.scala
@@ -292,7 +292,7 @@ case class EnginePage(parent: EngineTab) extends WebUIPage("") {
         <td> {session.name} </td>
         <td> {formatDate(session.startTime)} </td>
         <td> {if (session.endTime > 0) formatDate(session.endTime)} </td>
-        <td> {formatDurationVerbose(session.duration)} </td>
+        <td> {formatDuration(session.duration)} </td>
         <td> {session.totalOperations} </td>
       </tr>
     }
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/EnginePage.scala
```diff
@@ -72,7 +72,7 @@ case class EnginePage(parent: EngineTab) extends WebUIPage("") {
     }
       <li>
         <strong>Time since start: </strong>
-        {formatDurationVerbose(timeSinceStart)}
```
cxzl25, line 75: It looks like the Spark thrift server's page is also using the `formatDurationVerbose` method.

https://github.com/apache/spark/blob/f0e18284d0c8938cb35f2ad3668aff0e4ef4891c/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/ui/ThriftServerPage.scala#L65

bowenliang123: Yes, for `Time since start` , Spark uses `formatDurationVerbose`. But for the duration in the table, `formatDuration` is used.

bowenliang123: Reverted to use formatDurationVerbose for 'Time since start'.

yaooqinn (approved):

bowenliang123:
Thanks, merged to master (1.8.0) and 1.7.2.

PR #5029, opened 2023-07-06
Close the alive probe session after engine session closed

turboFei:

Now the server side session idle time is controlled by kyuubi admin, and the engine session idle time is controlled by users.

If the engine session idle time is short than that of the server session.

The engine session might be closed after engine session idle time, and now the alive probe session will still check the engine alive until the server session idle than the server session idle timeout.

image

In this pr, we will close the alive probe if the engine session transport is not open.


cxzl25 (approved):

turboFei:
thanks, merged to master

Issue #5030, opened 2023-07-06
[FEATURE] Support query id in Spark engine


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

The `SQLOperationListener` can get the `spark.sql.execution.id` when the sql execution starts.

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

lsm1:
please assign to me

PR #5037, opened 2023-07-10
[KYUUBI #5030] Support get query id in Spark engine

Yikf:

This PR aims to support `getQueryId` in Spark engine. It get `sparl.sql.execution.id` by adding a Listener.


yaooqinn:
why not use operationId directly?

Yikf:
> why not use operationId directly?

I apologize for the late reply due to the busy past, It might be better to agree with the executionId on the engine side, for example, to see Spark UI job information by this id.

yaooqinn:
That means a query without execution or failed has no id?

yaooqinn (approved):

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #5039, opened 2023-07-10
[Improvement] Use semantic versions and remove redundant version comparison methods

bowenliang123:

- Support initializing or comparing version with major version only, e.g "3" equivalent to  "3.0"
- Remove redundant version comparison methods by using semantic versions of Spark, Flink and Kyuubi
- adding common `toDouble` method

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
@@ -21,7 +21,7 @@ import java.time.{Instant, LocalDateTime, ZoneId}
 
 import scala.annotation.meta.getter
 
-import org.apache.spark.SparkContext
+import org.apache.spark.{SPARK_VERSION, SparkContext}
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.util.kvstore.KVIndex
 
@@ -97,5 +97,5 @@ object KyuubiSparkUtil extends Logging {
 
   // Given that we are on the Spark SQL engine side, the [[org.apache.spark.SPARK_VERSION]] can be
   // represented as the runtime version of the Spark SQL engine.
-  lazy val SPARK_ENGINE_RUNTIME_VERSION = SemanticVersion(org.apache.spark.SPARK_VERSION)
+  lazy val SPARK_ENGINE_RUNTIME_VERSION: SemanticVersion = SemanticVersion(SPARK_VERSION)
 }
--- a/docs/deployment/kyuubi_on_kubernetes.md
+++ b/docs/deployment/kyuubi_on_kubernetes.md
@@ -90,7 +90,7 @@ See more related details in [Using RBAC Authorization](https://kubernetes.io/doc
 
 ## Config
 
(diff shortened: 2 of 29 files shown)


pan3793 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -95,7 +95,5 @@ object KyuubiSparkUtil extends Logging {
     }
   }
 
-  // Given that we are on the Spark SQL engine side, the [[org.apache.spark.SPARK_VERSION]] can be
-  // represented as the runtime version of the Spark SQL engine.
-  lazy val SPARK_ENGINE_RUNTIME_VERSION = SemanticVersion(org.apache.spark.SPARK_VERSION)
+  lazy val sparkRuntimeVersion: SemanticVersion = SemanticVersion(SPARK_VERSION)
```
bowenliang123, line 98: Hi @cfmcgrady , I'm generalizing Spark's runtime semantic version into `sparkRuntimeVersion` as the same as in other engines and modules. Here I prefer the `import` way rather than the full name inline, which still loads Spark version when needed. WDYT ?

pan3793 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
cfmcgrady: > I'm generalizing Spark's runtime semantic version in sparkRuntimeVersion as the same as in other engines and modules.

As stated in the code comment, the [[org.apache.spark.SPARK_VERSION]] denotes the runtime version of the Spark SQL engine when the suite is executed on the Spark SQL engine side. However, when executing the suite on the client side, it is not feasible to represent the runtime version using [[org.apache.spark.SPARK_VERSION]], as shown in https://github.com/apache/kyuubi/pull/4381

> Here I prefer the import way rather than the full name inline, which still loads Spark version when needed.

I have no opinion on whether to use the `import` style, but I prefer uppercase for constants.

bowenliang123: Let me try to summarize your suggested changes:
1. preferred uppercase for constants, eg. `SPARK_ENGINE_RUNTIME_VERSION` as it is.
2. change to get Spark's runtime version via [JDBC query "SELECT version()" as in `SparkVersionUtil`](https://github.com/apache/kyuubi/blob/ecc3b4af6149ff49faff5f5360b67a6459d84856/kyuubi-common/src/test/scala/org/apache/kyuubi/util/SparkVersionUtil.scala#L30)
3. make this `SPARK_ENGINE_RUNTIME_VERSION` private and marked as visible for testing, as it's used in tests only.

bowenliang123: I reverted this to the original code here, as `lazy val SPARK_ENGINE_RUNTIME_VERSION = SemanticVersion(SPARK_VERSION)`.

bowenliang123:
Thanks, merged to master (1.8.0) .

PR #5045, opened 2023-07-12
Initialize EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH before using

zhaohehuhu:
I came cross an issue that is the the value of EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH is 0 when the param is accessed in generateExecutorPodNamePrefixForK8s method.

I tried to move the EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH ahead of generateExecutorPodNamePrefixForK8s method. Then, I found this issue was gone.

So is it necessary to declare the EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH variable before the method definition?

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -380,6 +380,7 @@ object SparkSQLEngine extends Logging {
     // only spark driver pod will build with `SPARK_APPLICATION_ID` env.
     Utils.isOnK8s && sys.env.contains("SPARK_APPLICATION_ID")
   }
+
   @VisibleForTesting
   def generateExecutorPodNamePrefixForK8s(userName: String): String = {
     val resolvedUserName =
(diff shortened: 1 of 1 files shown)


pan3793:
I don't think so, if you try to decompile the class file, the assignment should occur inside the constructor.

zhaohehuhu:
> decompile
Yes. Is there another problem possbiebly lead to this ?

pan3793:
Sorry, I missed `setupConf()` is called in side constructor too, then your analysis is right.

Please run `dev/reformat` to fix the code style.

zhaohehuhu:
> setupConf

Got it. Thanks.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -376,7 +380,6 @@ object SparkSQLEngine extends Logging {
     // only spark driver pod will build with `SPARK_APPLICATION_ID` env.
     Utils.isOnK8s && sys.env.contains("SPARK_APPLICATION_ID")
   }
-
```
pan3793, line 379: nit: restore this blank line to minimize the change

pan3793 (approved):
LGTM, only one minor thing

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
zhaohehuhu: Got it

pan3793:
Thanks, merged to master/1.7

zwangsheng:
Hi @zhaohehuhu, after this PR, we need new feature about force rewrite executor pod name prefix, would you like to take it?
See more in https://github.com/apache/kyuubi/issues/5654

PR #5058, opened 2023-07-17
[HELM] Allow the user to set the priorityClass used by Kyuubi server Pod

camper42:

This PR allows the user to set the priorityClass used by kyuubi server.

### DIFF
--- a/charts/kyuubi/values.yaml
+++ b/charts/kyuubi/values.yaml
@@ -37,7 +37,6 @@ serviceAccount:
   name: ~
 
 # priorityClass used for Kyuubi server pod
-# This priority class should be used for kyuubi server pods only.
 priorityClass:
   # Specifies whether a priorityClass should be created
   create: false
--- a/charts/kyuubi/values.yaml
+++ b/charts/kyuubi/values.yaml
@@ -37,6 +37,7 @@ serviceAccount:
   name: ~
 
 # priorityClass used for Kyuubi server pod
+# This priority class should be used for kyuubi server pods only.
 priorityClass:
   # Specifies whether a priorityClass should be created
   create: false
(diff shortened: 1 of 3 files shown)


charts/kyuubi/values.yaml
```diff
@@ -36,6 +36,16 @@ serviceAccount:
...
 
+# priorityClass used for Kyuubi server pod
+priorityClass:
+  # Specifies whether a priorityClass should be created
+  create: false
+  # Specifies priorityClass name to be used (created if `create: true`)
+  name: ~
+  # half of system-cluster-critical by default
+  value: 1000000000
+  description: "This priority class should be used for kyuubi server pods only."
```
pan3793, line 47: I think the content should be the comment instead of the description.

A meaningful description should explain the functionality/requirements of the object itself so that the K8s administers/operators know why the object should exist when they describe the object.

camper42: helm generate PriorityClass like below

as it used by helm automaticially and has label describe it's managed by Helm, maybe just delete the `description` field and leave comments only?
```
# Source: kyuubi/templates/kyuubi-priorityclass.yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: my-cluster-critical
  labels:
    helm.sh/chart: kyuubi-0.1.0
    app.kubernetes.io/name: kyuubi
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "1.7.1"
    app.kubernetes.io/managed-by: Helm
value: 1e+09
description: This priority class should be used for kyuubi server pods only.

```

charts/kyuubi/values.yaml
```diff
@@ -37,14 +37,14 @@ serviceAccount:
   name: ~
 
 # priorityClass used for Kyuubi server pod
+# This priority class should be used for kyuubi server pods only.
```
pan3793, line 40: seems redundant with the previous line ... prefer to remove it

pan3793 (approved):

pan3793:
Thanks, merged to master

Issue #4788, opened 2023-05-05
[Improvement][K8S][HELM] Use StatefulSet instead of Deployment


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Kyuubi server is actually stateful for batch API, especially the batch session recovery relies on stable IP address/hostname after restarting Kyuubi server

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #5062, opened 2023-07-18
[KYUUBI #4788][K8S][HELM] Use StatefulSet instead of Deployment

camper42:

Use StatefulSet instead of Deployment, add a headless service for statefulset

image
image

### DIFF
--- a/charts/kyuubi/templates/kyuubi-service.yaml
+++ b/charts/kyuubi/templates/kyuubi-service.yaml
@@ -23,6 +23,7 @@ metadata:
     {{- include "kyuubi.labels" $ | nindent 4 }}
 spec:
   type: ClusterIP
+  clusterIP: None
   ports:
     {{- range $name, $frontend := .Values.server }}
     - name: {{ $name | kebabcase }}
--- a/charts/kyuubi/templates/kyuubi-service.yaml
+++ b/charts/kyuubi/templates/kyuubi-service.yaml
@@ -15,6 +15,23 @@
   limitations under the License.
 */}}
 
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ $.Release.Name }}-headless
+  labels:
+    {{- include "kyuubi.labels" $ | nindent 4 }}
+spec:
+  type: ClusterIP
+  ports:
+    {{- range $name, $frontend := .Values.server }}
+    - name: {{ $name | kebabcase }}
+      port: {{ tpl $frontend.service.port $ }}
+      targetPort: {{ $frontend.port }}
+    {{- end }}
+  selector:
+    {{- include "kyuubi.selectorLabels" $ | nindent 4 }}
+---
 {{- range $name, $frontend := .Values.server }}
 {{- if $frontend.enabled }}
 apiVersion: v1
--- a/charts/kyuubi/templates/kyuubi-statefulset.yaml
+++ b/charts/kyuubi/templates/kyuubi-statefulset.yaml
@@ -29,6 +29,7 @@ spec:
   minReadySeconds: {{ .Values.minReadySeconds }}
   replicas: {{ .Values.replicaCount }}
   revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  podManagementPolicy: {{ .Values.podManagementPolicy }}
   {{- with .Values.updateStrategy }}
   updateStrategy: {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/charts/kyuubi/templates/kyuubi-statefulset.yaml
+++ b/charts/kyuubi/templates/kyuubi-statefulset.yaml
@@ -77,7 +77,7 @@ spec:
           {{- if .Values.livenessProbe.enabled }}
           livenessProbe:
             exec:
-              command: [ "/bin/bash", "-c", "bin/kyuubi status" ]
+              command: ["/bin/bash", "-c", "bin/kyuubi status"]
             initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
             timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
--- a/charts/kyuubi/templates/kyuubi-statefulset.yaml
+++ b/charts/kyuubi/templates/kyuubi-statefulset.yaml
@@ -87,7 +87,7 @@ spec:
           {{- if .Values.readinessProbe.enabled }}
           readinessProbe:
             exec:
-              command: [ "/bin/bash", "-c", "$KYUUBI_HOME/bin/kyuubi status" ]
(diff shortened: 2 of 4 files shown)


charts/kyuubi/templates/kyuubi-statefulset.yaml
```diff
@@ -71,7 +77,7 @@ spec:
           {{- if .Values.livenessProbe.enabled }}
           livenessProbe:
             exec:
-              command: ["/bin/bash", "-c", "bin/kyuubi status"]
+              command: [ "/bin/bash", "-c", "bin/kyuubi status" ]
```
pan3793, line 80:
```suggestion
              command: ["/bin/bash", "-c", "bin/kyuubi status"]
```

pan3793 (approved):
LGTM, @dnskr would you mind taking a look?

camper42:
btw, our production kyuubi changed to this version

pan3793:
@camper42 thanks for sharing, cc @turboFei @zwangsheng too

charts/kyuubi/templates/kyuubi-service.yaml
```diff
@@ -15,6 +15,23 @@
...
 */}}
 
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ $.Release.Name }}-headless
+  labels:
+    {{- include "kyuubi.labels" $ | nindent 4 }}
+spec:
+  type: ClusterIP
```
zwangsheng, line 25: we should add `clusterIp: None` here, as kubernetes doc said:

>  In this case, you can create what are termed headless Services, by explicitly specifying "None" for the cluster IP address (.spec.clusterIP).

camper42: thanks to point that, I have to delete the headless svc to fix this.
luckily, there is no direct use of this headless service at the moment.

zwangsheng (approved):
LGTM

dnskr (changes requested):
I left few requests for changes.

I'm not sure I know why these changes are needed, so it would be useful for me if you could provide more details about the PR:
 - Why is it needed to replace `Deployment` by `StetefullSet`, i.e. what kind of problem does it solve?
 - Would it be enough to use headless Service and Deployment?
 - Which Service (headless or not) clients should connect to and use?

camper42:
> - Why is it needed to replace Deployment by StetefullSet, i.e. what kind of problem does it solve?
> - Would it be enough to use headless Service and Deployment?

as our deployment, we want to persist server log later, so statefulset + pvc maybe a better choice

and ref to #4788: Kyuubi server is actually stateful for batch API, especially the batch session recovery relies on stable IP address/hostname after restarting Kyuubi server

> Which Service (headless or not) clients should connect to and use?

at the moment headless service only used to provide stable network ID. Our client is not in the same cluster as the kyuubi server, in this case we use kyuubi-thrift-binary's(LoadBalancer) external IP.

And I'm not sure which service it better for in-cluster client connect. maybe @pan3793 know more about it.

- client -> DNS: kyuubi-headless.kyuubi.svc.cluster.local -> pod IP (may have DNS cache issue)
- client -> kyuubi-protocol cluster IP -> pod ?

---


https://kyuubi.readthedocs.io/en/v1.7.1-rc0/deployment/high_availability_guide.html#client-side-usage
> then it can randomly pick one of the Kyuubi service uris from the specified ZooKeeper addresses in the /kyuubi path.

in ZK HA mode, client direct connect to a randomly picked kyuubi server.

In kubernetes, whether we use headless services (DNS service discovery) or ClusterIP (load balancing), it does not involve zk, so can HA be achieved without zk?

pan3793:
> I'm not sure which service it better for in-cluster client connect.

For THTIFT_BINARY (TCP) it does not matter, because it's a long connection, the whole lifecycle of a session while sticky in one Kyuubi Server Pod.

For REST (HTTP), I think headless should be better. In the batch submission (aka Spark Jar application), suppose the `spark-submit` was executed in the Kyuubi-Server-1, the following request to fetch local logs of `spark-submit` process must be handled by the same Pod, the current implementation is:

1. Client => [Kyuubi Server Service] => random Kyuubi Server Pod
2.1 If the Kyuubi Server Pod holds the batch session => handle and return response
2.2 If the Kyuubi Server Pod does not hold the batch session, check the state store (and external storage like MySQL) to find which Kyuubi Server holds it and forward the request to it, wait for the response, then pass through it to the Client

The 2.2 could be enhanced by
- the response contains the Kyuubi-Server-1 Pod address
- the client always(until network exception) tries to access the Kyuubi-Server-1 directly to avoid Kyuubi Server internal forwarding overhead.

In such a case, I would say Headless Service seems more suitable.

camper42:
headless service `kyuubi-headless.kyuubi.svc.cluster.local` will resolve to:
- Pod 0 IP
- Pod 1 IP
- ...
- Pod N IP

---

Pod DNS(stable network ID) will resolve to Pod IP:
- kyuubi-0.kyuubi-headless.kyuubi.svc.cluster.local -> Pod 0 IP
- kyuubi-1.kyuubi-headless.kyuubi.svc.cluster.local -> Pod 1 IP
- ...
- kyuubi-N.kyuubi-headless.kyuubi.svc.cluster.local -> Pod N IP

---

Cluster IP/External IP will proxy to one Pod IP (and stick for a client IP if sessionAffinity enabled)

---

in the current implementation, no difference between service expose methods for kyuubi REST ?

> - the response contains the Kyuubi-Server-1 Pod address
> - the client always(until network exception) tries to access the Kyuubi-Server-1 directly to avoid Kyuubi Server internal forwarding overhead.

in this implementation, client needs to reconnect to the Kyuubi-Server-1 Pod address, if using HTTP, client can connect by Pod IP address, but if using HTTPS, we need a stable address?

pan3793:
@turboFei could you share some ideas about HTTPS and the Service type chosen?

dnskr (approved):
LGTM
@camper42 thanks for your contribution and sorry for the delay with review.

turboFei:
> @turboFei could you share some ideas about HTTPS and the Service type chosen?


we are using stateful set.

A stable FQDN is needed for kyuubi batch recovery.

For HTTPS, stateful set is not a must.

pan3793:
Thanks, merged to master

pan3793:
Supply a new advantage of `StatefulSet` over `Deployment`, `Utils.findLocalInetAddress.getCanonicalHostName` returns different values:

- `Deployment`: Pod name, e.g. `kyuubi-bd8f56ddb-59jn5`
- `StatefulSet`: FQDN, e.g. `kyuubi-server-0.kyuubi-headless.spark.svc.cluster.local`

Thus, switching to `StatefulSet` could overcome the errors like 

```
Caused by: java.net.UnknownHostException: kyuubi-bd8f56ddb-59jn5
```

when `kyuubi.frontend.connection.url.use.hostname=true` (default is `true`)

PR #5079, opened 2023-07-21
[WIP] Infer constant conditions for dynamic partitions

wForget:

When dynamic partitions write data with constant partition filter, the Rebalance inserted in the last stage will result in a large partition, which is not suitable for remote shuffle service.

For example:

```
insert overwrite t1 partition(p1) select c1, p1 from t2 where p1='1';
```

The previous behavior was to insert `rebalance_by_col(p1)`, I want to change to `rebalance_by_none` to use random shuffle to avoid large partition.

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/RebalanceBeforeWriting.scala
+++ b/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/RebalanceBeforeWriting.scala
@@ -27,7 +27,18 @@ trait RepartitionBuilderWithRebalance extends RepartitionBuilder {
       query: LogicalPlan): LogicalPlan = {
     if (!conf.getConf(KyuubiSQLConf.INFER_REBALANCE_AND_SORT_ORDERS) ||
       dynamicPartitionColumns.nonEmpty) {
-      RebalancePartitions(dynamicPartitionColumns, query)
+      // TODO: make configurable
+      val constantConditions = InferDynamicPartitionConstantConditions
+        .infer(dynamicPartitionColumns, query)
+
+      val inferredConstantPartition = dynamicPartitionColumns
+        .forall(constantConditions.getOrElse(_, Nil).nonEmpty)
+
+      if (inferredConstantPartition) {
+        RebalancePartitions(Seq.empty, query)
+      } else {
+        RebalancePartitions(dynamicPartitionColumns, query)
+      }
     } else {
       val maxColumns = conf.getConf(KyuubiSQLConf.INFER_REBALANCE_AND_SORT_ORDERS_MAX_COLUMNS)
       val inferred = InferRebalanceAndSortOrders.infer(query)
--- a/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/RebalanceBeforeWriting.scala
+++ b/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/RebalanceBeforeWriting.scala
@@ -31,10 +31,10 @@ trait RepartitionBuilderWithRebalance extends RepartitionBuilder {
       val constantConditions = InferDynamicPartitionConstantConditions
         .infer(dynamicPartitionColumns, query)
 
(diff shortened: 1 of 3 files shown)


wForget:
@ulysses-you @cfmcgrady  Could you please take a quick look and give some advice?

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/RebalanceBeforeWriting.scala
```diff
@@ -27,7 +27,18 @@ trait RepartitionBuilderWithRebalance extends RepartitionBuilder {
       query: LogicalPlan): LogicalPlan = {
     if (!conf.getConf(KyuubiSQLConf.INFER_REBALANCE_AND_SORT_ORDERS) ||
       dynamicPartitionColumns.nonEmpty) {
-      RebalancePartitions(dynamicPartitionColumns, query)
+      // TODO: make configurable
+      val constantConditions = InferDynamicPartitionConstantConditions
+        .infer(dynamicPartitionColumns, query)
```
ulysses-you, line 32: We can add a new rule before `RepartitionBeforeWriting` that changes the dynamic partition write to static partition write.

wForget: Are there any other benefits of changing to static partition writing? It may infer multiple constant partition conditions here.

ulysses-you: it saves a local sort at least..

wForget: like: 

```
insert overwrite t1 partition(p1) select c1, p1 from t2 where p1 in (1', '2');
```

ulysses-you: For this case, does not `rebalance(t)` better than `rebalance` ? Can you talk more why `rebalance` is better ?

wForget: > For this case, does not `rebalance(t)` better than `rebalance` ? Can you talk more why `rebalance` is better ?

Makes sense, I was eager to avoid large partitions and ignored the original intention of this optimization.

wForget: > We can add a new rule before RepartitionBeforeWriting that changes the dynamic partition write to static partition write.

But this may be a dangerous behavior, if inference errors will lead to data errors, I think we can only control the behavior of Rebalance first, what do you think?

ulysses-you: Each optimization can cause bug, we should trust the code.. and we need a new config to control the behavior.

I think the key point is that, we should handle the multi-level partition spec carefully. e.g.,
```sql
INSERT INTO TABLE t1 (p1, p2, p3) SELECT * FROM t2 WHERE p1 = 'a'
```
For this case , it is still a dynamic partition write.

wForget: > Each optimization can cause bug, we should trust the code.. and we need a new config to control the behavior.

Thanks, i will try to implement it.

> I think the key point is that, we should handle the multi-level partition spec carefully. e.g.,

The following judgment should cover it, I will add more test cases to verify it.

https://github.com/apache/kyuubi/blob/d1c9f4d935c6e32366eac019b955f72731ecebc4/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/RebalanceBeforeWriting.scala#L34-L35

Issue #5080, opened 2023-07-21
[Bug][Flink] EmbeddedExecutorFactory not thread-safe during bootstrap


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Currently, EmbeddedExecutorFactory that Flink application mode relies on is not thread-safe. If a new SQL is submitted at the same time as the bootstrap SQL,  NPE occurs.

See the failed CI https://github.com/apache/kyuubi/actions/runs/5610938813/job/15201466145 for details.

### Affects Version(s)

1.8

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #5082, opened 2023-07-21
[KYUUBI #5080][FLINK] Fix EmbeddedExecutorFactory not thread-safe during bootstrap

link3280:
As titled.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/java/org/apache/flink/client/deployment/application/executors/EmbeddedExecutorFactory.java
+++ b/externals/kyuubi-flink-sql-engine/src/main/java/org/apache/flink/client/deployment/application/executors/EmbeddedExecutorFactory.java
@@ -48,6 +48,12 @@ public class EmbeddedExecutorFactory implements PipelineExecutorFactory {
 
   private static ScheduledExecutor retryExecutor;
 
+  private static final Object bootstrapLock = new Object();
+
+  private static final long BOOTSTRAP_WAIT_INTERVAL = 10_000L;
+
+  private static final int BOOTSTRAP_WAIT_RETRIES = 3;
+
   private static final Logger LOGGER = LoggerFactory.getLogger(EmbeddedExecutorFactory.class);
 
   public EmbeddedExecutorFactory() {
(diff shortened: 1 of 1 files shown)


externals/kyuubi-flink-sql-engine/src/main/java/org/apache/flink/client/deployment/application/executors/EmbeddedExecutorFactory.java
```diff
@@ -105,6 +114,25 @@ public boolean isCompatibleWith(final Configuration configuration) {
...
+        } catch (InterruptedException e) {
+          throw new RuntimeException("Interrupted while waiting for bootstrap.", e);
+        }
+      }
+      if (bootstrapJobIds == null) {
+        throw new RuntimeException(
+            "Bootstrap of Flink SQL engine timed out after "
+                + BOOTSTRAP_WAIT_INTERVAL * BOOTSTRAP_WAIT_RETRIES
+                + ". "
+                + "Please check the engine log for more details.");
```
pan3793, line 133:
```suggestion
                + ". Please check the engine log for more details.");
```

externals/kyuubi-flink-sql-engine/src/main/java/org/apache/flink/client/deployment/application/executors/EmbeddedExecutorFactory.java
```diff
@@ -105,6 +114,24 @@ public boolean isCompatibleWith(final Configuration configuration) {
   public PipelineExecutor getExecutor(final Configuration configuration) {
     checkNotNull(configuration);
     Collection<JobID> executorJobIDs;
+    synchronized (bootstrapLock) {
```
pan3793, line 117: Emm... I'm not sure I get the idea.

If the code reaches here and `bootstrapJobIds == null` is true, another thread has no opportunity to initialize the static variables since they are waiting for the `bootstrapLock` to be released.

I suppose we should use a `CountDownLatch` here instead, to avoid blocking other threads to initialize the variables.

link3280: IMH, if `bootstrapJobIds == null`, the current thread would release the lock and wait for `bootstrapLock.notifyAll()` to be called by the bootstrap thread.

link3280: This could be verified by removing this line https://github.com/apache/kyuubi/blob/bf0307565cf1196447403c6a6f8e240ac01d1cd5/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala#L105 or add a delay like 30 seconds (which is the bootstrap timeout) to it.

pan3793: Sorry, my bad, the change makes sense to me

link3280: No worries. I pushed a commit to improve the logs. I think it's good to merge when the CI turns green. Thanks!

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #5115, opened 2023-07-31
[Minor] Simplify single character alternation in RegExp

bowenliang123:

- Simplify Single character alternation in RegExp of `SparkOperation`

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
@@ -50,7 +50,7 @@ abstract class SparkOperation(session: Session)
 
   private val timeZone: ZoneId = {
     spark.conf.getOption(TIMEZONE_KEY).map { timeZoneId =>
-      ZoneId.of(timeZoneId.replaceFirst("(\\+|\\-)(\\d):", "$10$2:"), ZoneId.SHORT_IDS)
+      ZoneId.of(timeZoneId.replaceFirst("([+\\-])(\\d):", "$10$2:"), ZoneId.SHORT_IDS)
     }.getOrElse(ZoneId.systemDefault())
   }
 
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -50,7 +50,7 @@ abstract class SparkOperation(session: Session)
 
   private val timeZone: ZoneId = {
     spark.conf.getOption(TIMEZONE_KEY).map { timeZoneId =>
-      ZoneId.of(timeZoneId.replaceFirst("(\\+|\\-)(\\d):", "$10$2:"), ZoneId.SHORT_IDS)
+      ZoneId.of(timeZoneId.replaceFirst("([+\\-])(\\d):", "$10$2:"), ZoneId.SHORT_IDS)
```
yaooqinn, line 53:
```scala
  def getZoneId(timeZoneId: String): ZoneId = {
    val formattedZoneId = timeZoneId
      // To support the (+|-)h:mm format because it was supported before Spark 3.0.
      .replaceFirst("(\\+|\\-)(\\d):", "$10$2:")
      // To support the (+|-)hh:m format because it was supported before Spark 3.0.
      .replaceFirst("(\\+|\\-)(\\d\\d):(\\d)$", "$1$2:0$3")

    ZoneId.of(formattedZoneId, ZoneId.SHORT_IDS)
  }
```

refer to upstream changes, we may change here with the above code

bowenliang123: Is there any doc for referring these upstream changes?

yaooqinn: no

bowenliang123: all right. But I could not understan why or in what circumstance that the replacement should be `"$10$2:"` rather than padding 0 in the front as  `"0$1$2:"`

bowenliang123: OK, LGTM after reading the comments again. Has adopted your code suggestion.

PR #5120, opened 2023-08-01
[JDBC] Allow using session's user and password for connecting database in JDBC engine

bowenliang123:

- allow using the session's user/password to connect database in the JDBC engine
- it is allowed to be applied on any share level of engines, since every kyuubi session maintains a dedicated JDBC connection in the current implementation


bowenliang123:
cc @zhaomin1423 @pan3793

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2657,6 +2657,13 @@ object KyuubiConf {
       .stringConf
       .createOptional
 
+  val ENGINE_JDBC_USE_SESSION_USER_PASSWORD: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.jdbc.connection.use.session.user.password")
```
pan3793, line 2661: I recommend to

1. rename it to `kyuubi.engine.jdbc.propagateCredential`
2. make it only takes effect on CONNECTION/USER share level
3. mark as internal

bowenliang123: I have limited this feature effecting on CONNECTION/USER share level.

Meanwhile,
In the current implementation of JDBC engine, every session has a dedicated JDBC connection. Therefore, I think we should allow them to custom session's user/password in runtime, even in GROUP and SERVER level which only controls the share level of engine instance but not the exact JDBC connection.

pan3793: I wrongly thought that sessions share a connection pool in the JDBC engine, then things become intuitive.
Suggestions 2. 3. could be ignored. Sorry for making you back and forth.

bowenliang123: No problem.  You raised a combination of share levels on engines and sessions which is worth to be discussed.

pan3793 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
bowenliang123: Updated the description with the clarification above.

bowenliang123:
Thanks, merged to master.

PR #5144, opened 2023-08-08
[WIP] [AUTHZ] hive function usage authz plugin

packyan:


extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleFunctionAuthorization.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.kyuubi.plugin.spark.authz.ranger.RuleFunctionAuthorization._
+class RuleFunctionAuthorization(spark: SparkSession) extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    buildPrivileges(spark, plan)
+  }
+}
+
+object RuleFunctionAuthorization {
+
+  private val functionPrivAndOpType = new ThreadLocal[PrivilegesAndOpType]
```
bowenliang123, line 35: It might be risky to use `ThreadLocal` as the function privileges holder.
Is it possible to put them in logical plan's tag via `setTagValue` ?

packyan:
Do we need to add a toggle parameter for this feature, and set `Disabled` as default?

PR #5149, opened 2023-08-09
[Improvement] Correct error message of ReflectUtils's invokeAs when method not found

link3280:

Currently, overloaded methods are considered the same and deduplicated in ReflectUtils, thus not easy to tell why no method is found.

This PR fixes the problem by adding the argument lists. In addition, it fixes the problem that the arg classes are not printed correctly.

### DIFF
--- a/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
+++ b/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
@@ -93,11 +93,13 @@ object ReflectUtils {
     } catch {
       case e: Exception =>
         val candidates =
-          (clz.getDeclaredMethods ++ clz.getMethods).map(_.getName).distinct.sorted
-        val argClassesNames = argClasses.map(_.getClass.getName)
+          (clz.getDeclaredMethods ++ clz.getMethods)
+            .map(m => s"${m.getName}(${m.getParameterTypes.map(_.getName).mkString(", ")})")
+            .distinct.sorted
+        val argClassesNames = argClasses.map(_.getName)
         throw new RuntimeException(
-          s"Method $methodName(${argClassesNames.mkString(",")})" +
-            s" not found in $clz [${candidates.mkString(",")}]",
+          s"Method $methodName(${argClassesNames.mkString(", ")})" +
+            s" not found in $clz [${candidates.mkString(", ")}]",
           e)
     }
   }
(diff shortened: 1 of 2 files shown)


kyuubi-util-scala/src/test/java/org/apache/kyuubi/util/reflect/ReflectUtilsSuite.scala
```diff
@@ -66,6 +66,16 @@ class ReflectUtilsSuite extends AnyFunSuite {
     assertResult("field5")(getField[String](ObjectA, "field5"))
     assertResult("field6")(getField[String](ObjectA, "field6"))
   }
+
+  test("test invokeAs method not found exception") {
+    val exception = intercept[RuntimeException]{
+      invokeAs[String](ObjectA, "methodNotExists")
```
pan3793, line 72: could you supply a case with args?

link3280: Updated. PTAL.

pan3793 (approved):

kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
```diff
@@ -93,11 +93,13 @@ object ReflectUtils {
     } catch {
       case e: Exception =>
         val candidates =
-          (clz.getDeclaredMethods ++ clz.getMethods).map(_.getName).distinct.sorted
-        val argClassesNames = argClasses.map(_.getClass.getName)
+          (clz.getDeclaredMethods ++ clz.getMethods)
+            .map(m => s"${m.getName}(${m.getParameterTypes.map(_.getName).mkString(", ")})")
```
bowenliang123, line 97:
```suggestion
            .map(m => s"${m.getName}(${m.getParameterTypes.map(_.getSimpleName).mkString(", ")})")
```

A simple class name should be good enough for debugging. Full class name for common classes (other than primitive types) is too long for multiple params of all methods.

bowenliang123 (approved):
LGTM.

cxzl25 (approved):

cfmcgrady (approved):

link3280:
Merged. Thanks to all!

pan3793:
@link3280 don't forget to update the milestone and assignee

link3280:
> @link3280 don't forget to update the milestone and assignee

Thanks for reminding me. Updated.

bowenliang123:
> @link3280 don't forget to update the milestone and assignee

Maybe we could check these requirements in the request merging scripts.

pan3793:
> Maybe we could check these requirements in the request merging scripts.

Nice idea~

bowenliang123:
> > @link3280 don't forget to update the milestone and assignee
> 
> Maybe we could check these requirements in the request merging scripts.

Created PR #5152 for this.

PR #5154, opened 2023-08-11
[Doc] Move configuration docs to the top level

bowenliang123:

- Move the configuration docs to the top level of docs, which is most commonly used and referenced
- update relevant doc links

image


pan3793 (approved):

cxzl25 (approved):

yaooqinn:
By the way, it could be separated from the admin guide as configuration guide

bowenliang123:
> By the way, it could be separated from the admin guide as configuration guide

Yeah, we could do that next time (won't affect the doc URL), when config docs become more mutual and prosperous.
Would prefer to leave it in Admin Guide for now, as Configurations are part of the administrators' responsibilities of decisions and evaluation.

ulysses-you (approved):

bowenliang123:
Thanks for reviewing, merged to master (1.8.0).

PR #5158, opened 2023-08-11
Allow embedded Zookeeper binding IP address

pan3793:

This PR provides a new configuration `kyuubi.zookeeper.embedded.client.use.hostname` to control whether use hostname or IP address on binding, to improve the out-of-box experience of K8s case, similar to `kyuubi.frontend.connection.url.use.hostname`.

### DIFF
--- a/kyuubi-zookeeper/src/main/scala/org/apache/kyuubi/zookeeper/ZookeeperConf.scala
+++ b/kyuubi-zookeeper/src/main/scala/org/apache/kyuubi/zookeeper/ZookeeperConf.scala
@@ -49,6 +49,15 @@ object ZookeeperConf {
       .stringConf
       .createOptional
 
+  val ZK_CLIENT_USE_HOSTNAME: ConfigEntry[Boolean] =
+    buildConf("kyuubi.zookeeper.embedded.client.use.hostname")
+      .doc("When true, embedded Zookeeper prefer to bind hostname, otherwise, ip address. " +
+        "Note that, the default value is set to `false` when engine running on Kubernetes " +
+        "to prevent potential network issues.")
+      .version("1.7.2")
+      .booleanConf
+      .createWithDefault(true)
+
   val ZK_DATA_DIR: ConfigEntry[String] = buildConf("kyuubi.zookeeper.embedded.data.dir")
     .doc("dataDir for the embedded zookeeper server where stores the in-memory database" +
       " snapshots and, unless specified otherwise, the transaction log of updates to the database.")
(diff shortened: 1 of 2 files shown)


kyuubi-zookeeper/src/main/scala/org/apache/kyuubi/zookeeper/ZookeeperConf.scala
```diff
@@ -49,6 +49,15 @@ object ZookeeperConf {
...
       .createOptional
 
+  val ZK_CLIENT_USE_HOSTNAME: ConfigEntry[Boolean] =
+    buildConf("kyuubi.zookeeper.embedded.client.use.hostname")
+      .doc("When true, embedded Zookeeper prefer to bind hostname, otherwise, ip address. " +
+        "Note that, the default value is set to `false` when engine running on Kubernetes " +
+        "to prevent potential network issues.")
+      .version("1.7.2")
+      .booleanConf
+      .createWithDefault(true)
```
pan3793, line 59: use true to keep behavior, but maybe false is better? since it's not intend to use for production, behavior change is fine.

pan3793:
@zwangsheng

Yikf (approved):

cxzl25 (approved):

kyuubi-zookeeper/src/main/scala/org/apache/kyuubi/zookeeper/ZookeeperConf.scala
```diff
@@ -49,6 +49,13 @@ object ZookeeperConf {
       .stringConf
       .createOptional
 
+  val ZK_CLIENT_USE_HOSTNAME: ConfigEntry[Boolean] =
+    buildConf("kyuubi.zookeeper.embedded.client.use.hostname")
+      .doc("When true, embedded Zookeeper prefer to bind hostname, otherwise, ip address.")
+      .version("1.7.2")
+      .booleanConf
+      .createWithDefault(false)
```
zwangsheng, line 57: Should we keep the original logic and default set true to bind hostname?

zwangsheng: Ignore this, LGTM default to bind on ip, consider that in most cases(especially On k8s, to reduced DNS discovery), bind ip is better for bind hostname

zwangsheng (approved):

kyuubi-zookeeper/src/main/scala/org/apache/kyuubi/zookeeper/ZookeeperConf.scala
pan3793: yea, was thought that, but since it's not for production, I think behavior change should be fine.

pan3793:
Merged to master/1.7

PR #5160, opened 2023-08-14
Refactor getNextRowSetInternal to support fetch streaming data

link3280:
Currently, `getNextRowSetInternal` returns `TRowSet` which is not friendly to explicit EOS in streaming result fetch.

This PR changes the return type to `TFetchResultsResp` to allow the engines to determine the EOS.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperation.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperation.scala
@@ -25,7 +25,7 @@ import scala.collection.JavaConverters.collectionAsScalaIterableConverter
 import org.apache.flink.configuration.Configuration
 import org.apache.flink.table.gateway.service.context.SessionContext
 import org.apache.flink.table.gateway.service.operation.OperationExecutor
-import org.apache.hive.service.rpc.thrift.{TGetResultSetMetadataResp, TRowSet, TTableSchema}
+import org.apache.hive.service.rpc.thrift.{TFetchResultsResp, TGetResultSetMetadataResp, TTableSchema}
 
 import org.apache.kyuubi.{KyuubiSQLException, Utils}
 import org.apache.kyuubi.engine.flink.result.ResultSet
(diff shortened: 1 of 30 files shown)


bowenliang123:
I may have missed the key point of the changes. How does `TFetchResultsResp` improves explicit EOS in streaming result fetch than `TRowSet` ?

link3280:
> I may have missed the key point of the changes. How does `TFetchResultsResp` improves explicit EOS in streaming result fetch than `TRowSet` ?

Because the engines could determine EOS of the result set by setting 'hasMoreResults' in `TFetchResultsResp`.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperation.scala
```diff
@@ -112,7 +114,10 @@ abstract class FlinkOperation(session: Session) extends AbstractOperation(sessio
       zoneId,
       getProtocolVersion)
     resultRowSet.setStartRowOffset(resultSet.getData.getPosition)
-    resultRowSet
+    val resp = new TFetchResultsResp(OK_STATUS)
+    resp.setResults(resultRowSet)
+    resp.setHasMoreRows(resultSet.getData.hasNext)
```
pan3793, line 119: seems this line should be added in the next PR?

link3280: You're right. Removing this line.

pan3793 (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -520,23 +520,20 @@ abstract class TFrontendService(name: String)
...
-    val resp = new TFetchResultsResp
     try {
       val operationHandle = OperationHandle(req.getOperationHandle)
       val orientation = FetchOrientation.getFetchOrientation(req.getOrientation)
       // 1 means fetching log
       val fetchLog = req.getFetchType == 1
       val maxRows = req.getMaxRows.toInt
-      val rowSet = be.fetchResults(operationHandle, orientation, maxRows, fetchLog)
-      resp.setResults(rowSet)
-      resp.setHasMoreRows(false)
```
pan3793, line 532: Seems we need to call this method explicitly.

```
  public void setHasMoreRows(boolean hasMoreRows) {
    this.hasMoreRows = hasMoreRows;
    setHasMoreRowsIsSet(true);
  }
```

bowenliang123 (approved):
LGTM.

pan3793:
Thanks, merged to master

PR #5173, opened 2023-08-16
Explicitly close underlying httpclient when closing KyuubiConnection in HTTP transport mode

bowenliang123:

- Currently KyuubiConnection's`close` method closes the TTransport instance in HTTP transport mode, that only closes the InputStream of HTTP response rather than the underlying HTTP client/connection in `libthrift 0.9.3`. It may cause connection leaks.
- Explicitly close the underlying HTTP client to fix this problem. Note: close a closed HTTP client would have no side effect.

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -88,6 +88,7 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
   private JdbcConnectionParams connParams;
   private TTransport transport;
   private TCLIService.Iface client;
+  private CloseableHttpClient httpClient;
   private boolean isClosed = true;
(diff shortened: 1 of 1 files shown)


bowenliang123:
Source code of libthrift 0.9.3  viewed on local IDEA: (Closing the input stream but not the connection)
image

Wile  viewed on Github (Closing boththe input stream and the connection)
https://github.com/apache/thrift/blob/0.9.3/lib/javame/src/org/apache/thrift/transport/THttpClient.java

pan3793:
THRIFT-4130 may be related

pan3793:
BTW, you are looking `javame`, please go to `java` dir instead

bowenliang123:
> BTW, you are looking at `javame`, please go to `java` dir instead

Yes, you are right. In Java dir, it shows the same code as released in 0.9.3.

Source code in `java` folder : (Closing both the input stream and the connection)
https://github.com/apache/thrift/blob/0.9.3/lib/java/src/org/apache/thrift/transport/THttpClient.java#L166

bowenliang123:

PR #5179, opened 2023-08-18
Use Iterable instead of Seq in kyuubi-ctl commands

bowenliang123:

- Use `Iterable` instead of `Seq` in kyuubi-ctl commands for Scala  compatibility, as
1.  in Scala 2.13,  the `scala.Seq` is now an alias for `scala.collection.immutable.Seq`  (instead of `scala.collection.Seq`)
2. in Scala 2.13, `scala.collection.mutable.ListBuffer` (or Buffers) does not extend `scala.collection.immutable.Seq` according to [Scala collection migration guide](https://docs.scala-lang.org/overviews/core/collections-migration-213.html)
3. in both Scala 2.12 and 2.13, `ListBuffer` (or Buffers) extends `scala.collection.mutable.Iterable` 


cxzl25 (approved):

lightning-L (approved):

bowenliang123:
Thanks, merged to master.

Issue #4540, opened 2023-03-16
[Improvement] Support to limit some user to make connection


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

It is helpful, sometimes we need to add some user into blacklist.

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

lordk911:
Sometimes make sure some user can only connect to kyuubi from special hosts is also helpful, just like mysql. e.g. a BI system user may have a lot of select privilege, this kind of user should only active on the BI system host machines.

turboFei:
Yes，would you like to help on this？
@lordk911

goldwater2019:
Hi, I'd like to. @turboFei Please assign this to me. By the way, is it feasible or prefered to setup blocklist in kyuubi-defaults.conf?

turboFei:
Sure @goldwater2019

turboFei:
https://github.com/apache/kyuubi/pull/4360
Fyi，you can reload the conf.

666fishsun:
@turboFei hello~ may I have the chance to be assigned this issue? It seems @goldwater2019 no response.

turboFei:
thanks @666fishsun , assigned

zhuyaogai:
@turboFei @pan3793 hi, I'm new to kyuubi, but I really want to be involved, could you assign this issue for me? I will submit the PR as soon as possible. And it seems @666fishsun no response.

turboFei:
sure, thanks @zhuyaogai

PR #5182, opened 2023-08-21
[KYUUBI #4540] Support to deny some users to make connection

zhuyaogai:

Improvement: https://github.com/apache/kyuubi/issues/4540

No


zhuyaogai:
@turboFei hi, could you please do a code review and provide some suggestions? Thank you very much!

docs/configuration/settings.md
```diff
@@ -392,6 +392,7 @@ You can configure the Kyuubi properties in `$KYUUBI_HOME/conf/kyuubi-defaults.co
...
 | kyuubi.server.limit.connections.per.user.ipaddress       | &lt;undefined&gt; | Maximum kyuubi server connections per user:ipaddress combination. Any user-ipaddress exceeding this limit will not be allowed to connect.                                                                                                      | int      | 1.6.0 |
 | kyuubi.server.limit.connections.user.unlimited.list                         || The maximum connections of the user in the white list will not be limited.                                                                                                                                                                     | seq      | 1.7.0 |
+| kyuubi.server.limit.connections.user.blacklist                              || The user in the blacklist will be refused to connect to kyuubi server.                                                                                                                                                                         | seq      | 1.8.0 |
```
turboFei, line 395: kyuubi.server.limit.connections.user.limited.list

turboFei: The user in the limited list will be refused to connect to kyuubi server.

turboFei:
thanks for raising this pr, please help rename all the `blacklist` to `limited.list` to align with the config `kyuubi.server.limit.connections.user.unlimited.list`.

zhuyaogai:
> thanks for raising this pr, please help rename all the `blacklist` to `limited.list` to align with the config `kyuubi.server.limit.connections.user.unlimited.list`.

done.

turboFei:
```
error file=/home/runner/work/kyuubi/kyuubi/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala message=File line length exceeds 100 characters line=111
```

zhuyaogai:
> ```
> error file=/home/runner/work/kyuubi/kyuubi/kyuubi-ctl/src/main/scala/org/apache/kyuubi/ctl/opt/AdminCommandLine.scala message=File line length exceeds 100 characters line=111
> ```

@turboFei hi, could you tell me what commands can be used for automatic code formatting? For example: `mvn spotless:apply`?

pan3793:
> ... what commands can be used for automatic code formatting?

`dev/reformat`

https://kyuubi.readthedocs.io/en/master/contributing/code/developer.html#format-all-code

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -361,13 +368,29 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
     batchLimiter.foreach(SessionLimiter.resetUnlimitedUsers(_, unlimitedUsers))
   }
 
+  private[kyuubi] def getLimitedUsers: Set[String] = {
+    limiter.orElse(batchLimiter).map(SessionLimiter.getLimitedUsers).getOrElse(Set.empty)
+  }
+
+  private[kyuubi] def refreshLimitedUsers(conf: KyuubiConf): Unit = {
+    val limitedUsers = conf.get(SERVER_LIMIT_CONNECTIONS_USER_LIMITED_LIST).toSet
```
bowenliang123, line 376: The user list should be filtered ensuring non-empty elements.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -101,14 +101,21 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
...
     userIpAddressLimit: Int,
-    var unlimitedUsers: Set[String])
+    var unlimitedUsers: Set[String],
+    var limitedUsers: Set[String])
   extends SessionLimiterImpl(userLimit, ipAddressLimit, userIpAddressLimit) {
   override def increment(userIpAddress: UserIpAddress): Unit = {
-    if (!unlimitedUsers.contains(userIpAddress.user)) {
+    val user = userIpAddress.user
+    if (StringUtils.isNotBlank(user) && limitedUsers.contains(user)) {
+      val errorMsg = s"Connection refused because the user is limited. (user: $user)"
```
bowenliang123, line 114: Misleading message. This is not part of the connection handshake.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -101,14 +101,21 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
...
     ipAddressLimit: Int,
     userIpAddressLimit: Int,
-    var unlimitedUsers: Set[String])
+    var unlimitedUsers: Set[String],
+    var limitedUsers: Set[String])
   extends SessionLimiterImpl(userLimit, ipAddressLimit, userIpAddressLimit) {
   override def increment(userIpAddress: UserIpAddress): Unit = {
-    if (!unlimitedUsers.contains(userIpAddress.user)) {
+    val user = userIpAddress.user
+    if (StringUtils.isNotBlank(user) && limitedUsers.contains(user)) {
```
bowenliang123, line 113: I'm not familiar with the usage of SessionLimiter. Is this the right way or timing to do the rejection here?  cc @wForget

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/AdminResourceSuite.scala
```diff
@@ -125,6 +125,19 @@ class AdminResourceSuite extends KyuubiFunSuite with RestFrontendTestHelper {
     assert(200 == response.getStatus)
   }
 
+  test("refresh limited users of the kyuubi server") {
+    var response = webTarget.path("api/v1/admin/refresh/limited_users")
+      .request()
+      .post(null)
+    assert(405 == response.getStatus)
```
bowenliang123, line 132: `403 Forbidden` may be a more suitable response code than `405 Method not allowed`.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
zhuyaogai: done.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
zhuyaogai: done.

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/AdminResourceSuite.scala
zhuyaogai: I wrote the code based on other test cases. If you think the 403 status code should be better, I can modify it (including all other test cases).

turboFei:
could you check the log error and address it？

zhuyaogai:
> could you check the log error and address it？
Sorry, I thought this was not my issue, so I didn't pay attention to it. It has now been fixed.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2614,6 +2614,15 @@ object KyuubiConf {
       .toSequence()
       .createWithDefault(Nil)
 
+  val SERVER_LIMIT_CONNECTIONS_USER_LIMITED_LIST: ConfigEntry[Seq[String]] =
+    buildConf("kyuubi.server.limit.connections.user.limited.list")
+      .doc("The user in the limited list will be limited to connect to kyuubi server.")
```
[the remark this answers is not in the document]
zhuyaogai: @pan3793 hi, do you mean that the user is configured in both `user.unlimited.list` and `user.limited.list`? In my code logic, it seems that the priority of the latter is higher.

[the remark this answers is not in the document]
pan3793: yes. and the logic makes sense, but we should document it clear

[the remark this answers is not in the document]
pan3793: let's say there is a user name "tom", what if `user.unlimited.list` and `user.limited.list` is configured but neither includes "tom"?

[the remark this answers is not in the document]
zhuyaogai: > the logic makes sense, but we should document it clear

done.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
pan3793: I have the same question, the implementation looks incorrect, limited does not mean rejection.

zhuyaogai: @pan3793 hi, I am new to kyuubi, could you give me more details about the meaning of `limited user`? From the issue  context I cannot get more input.

pan3793: I suppose when "limited user" is configured, connections from those users will be limited not rejected.

e.g. 

```
kyuubi.server.limit.connections.user.limited.list=tom,jerry
kyuubi.server.limit.connections.per.user=10
```

user "tom" and "jerry" are limited to opening up to 10 connections for each user, other users are not limited.

pan3793: 1. that's why we discuss the priority of `limited.user.list` and `unlimited.user.list`, I think it's fine that `limited.user.list` takes high priority. (it means when `limited.user.list` is configured, ignore `unlimited.user.list`)
2. `user-ipaddress` should be affected by this change too. and we can introduce a similar change for `ipaddress` in another PR.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2614,6 +2614,17 @@ object KyuubiConf {
...
 
+  val SERVER_LIMIT_CONNECTIONS_USER_LIMITED_LIST: ConfigEntry[Seq[String]] =
+    buildConf("kyuubi.server.limit.connections.user.limited.list")
+      .doc("The user in the limited list will be limited to connect to kyuubi server, " +
+        "if the user has configured both user.unlimited.list and user.limited.list, " +
+        "the priority of the latter is higher.")
+      .version("1.8.0")
+      .serverOnly
+      .stringConf
+      .toSequence()
```
bowenliang123, line 2625:
```suggestion
      .toSet()
```
With #5185 introduced set collection support for configs, you could rebase to the master branch and use it.

zhuyaogai: Thank you for your reminder.

turboFei (approved):

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: LGTM after this comment addressed

zhuyaogai: @turboFei hi, do you mean that aside from the above case, my current code implementation is okay? but @pan3793  thinks it's incorrect.

pan3793: I had an offline discussion with @turboFei, and we reached a consensus that:

1. The current implementation is **deny** semantic, which is a good feature but **limited** is not a properly name
2. You could either change the configuration name to `...deny.user.list`, or keep the `...limited.user.list` and change the implementation as I commented above https://github.com/apache/kyuubi/pull/5182#discussion_r1304545505, I think the latter approach is simple, and we can implement the **deny** in another PR.

zhuyaogai: @pan3793 hi, I have made some changes to the code, and I hope I understood what you wanted. Could you please review the code and give me some suggestions? Thank you!

pan3793: @zhuyaogai will take a look tomorrow, have a good weekend :)

zhuyaogai: @pan3793 BTW, I found that is there a problem with the current master branch code? When some users open some sessions and before closing the sessions, this user is added to the `unlimited.list`, then the counter related to this user will not be cleared.

zhuyaogai: I think we don't need to check any limit in `decrement` method, and let the already opened sessions decrement limit counter. Please correct me if I am wrong.

bowenliang123: > I had an offline discussion with @turboFei, and we reached a consensus that:
> 
> 1. The current implementation is **deny** semantic, which is a good feature but **limited** is not a properly name
> 2. You could either change the configuration name to `...deny.user.list`, or keep the `...limited.user.list` and change the implementation as I commented above [[KYUUBI #4540] [Improvement] Support to limit some user to make connection. #5182 (comment)](https://github.com/apache/kyuubi/pull/5182#discussion_r1304545505), I think the latter approach is simple, and we can implement the **deny** in another PR.

I have similar concerns in these variations. I was confused by the **deny** and **limited**.
As for me, a limited user list is not the contrast to an unlimited user, and only a feature for  deny users list is necessary.
What's more I would like to see when the server reject user connection , this feature should be part of authentication ,such as User filerter,  but not an injected mechanism in such Limiter. Or did I miss any information for better understanding?

I would suggest, 
1. Add detailed and summarised design and key points to this PR description, as a simple link to the original issue is not enough
2. Add constructive discussion in proposed change and the reason for possible approaches

I'm fine if we accept this feature, especially when some of us have a mutual understanding of enough information.

zhuyaogai:
@pan3793 hi, will we have further discussions?

pan3793:
@zhuyaogai I'm not sure if we have gaps in the logic design. The code implementation and the test cases do not cover the combinations of `limited.users` and `unlimited.users`.

1. Neither `limited.users` nor `unlimited.users` are configured
2. Only `limited.users` is configured
3. Only `unlimited.users` is configured
4. Both `limited.users` and `unlimited.users` are configured

seems the current code only considers case 4?

> that's why we discuss the priority of `limited.user.list` and `unlimited.user.list`, I think it's fine that `limited.user.list` takes high priority. (it means when `limited.user.list` is configured, ignore `unlimited.user.list`)

And I suppose we reached a consensus on what the behavior should be like for case 4, but the code looks do not match.

turboFei (approved):
LGTM, thanks for the efforts

pan3793 (approved):
LGTM, thanks for your patience and great work! @zhuyaogai

bowenliang123 (approved):

pan3793:
Thanks, merged to master

PR #5199, opened 2023-08-25
Read all columns of metadata to prevent column missing 

turboFei:

Since https://github.com/apache/kyuubi/issues/4843, we need the `requestConf` column to get application manager info.

So now the stateOnly option for metadata select is not needed.
```
private val METADATA_ALL_COLUMNS = Seq(
    METADATA_STATE_ONLY_COLUMNS,
    "resource",
    "class_name",
    "request_conf",
    "request_args").mkString(",")
```
In this pr, I remove the `stateOnly` flag to select all metadata columns to prevent column missing and make it easy to maintain.
Existing UT.

No


pan3793 (approved):

pan3793:
LGTM if CI pass

turboFei:
thanks, merging to master

PR #5201, opened 2023-08-25
Allow disabling batch impl v2 on globally

pan3793:

It makes no sense to enable batch impl v2 when `kyuubi.batch.submitter.enabled` is disabled, to avoid misuse, globally disable batch impl v2 in such case.

No.


cxzl25 (approved):

pan3793:
Thanks, merged to master

PR #5206, opened 2023-08-28
Try to kill pod with label if no ApplicationInfo found to prevent pod leak

turboFei:

Now for `KubernetesApplicationOperation`, it rely on the appInfoStore.

For batch rest api, if the closeBatch request can not be send to the kyuubiInstance that created the batch, the current kyuubiInstance will try to kill the batch.

I wonder that, in this case, the applicationInfo might can not be found in the appInfoStore.

It is better to try the best to kill the pod with  `kyuubi-unique-tag` label to prevent pod leak.


pan3793:
Also cc @zwangsheng, as you are investigating issues about deleting Pod multiple times, this could be a new case.

pan3793 (approved):

cxzl25 (approved):

turboFei:
thanks, merged to master

Issue #5210, opened 2023-08-29
[Bug] Cancel operation will cause the log file node to leak


ASiegeLion:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Execute sql, then cancel, use the lsof command to find the log file node leak,
![image](https://github.com/apache/kyuubi/assets/30170267/188e8151-e7c7-40c9-bdb9-d8daf3a0946a)
![image](https://github.com/apache/kyuubi/assets/30170267/3c2f9ea2-fc3c-420f-ac70-cf23236f10de)

The KyuubiOperation class found that there is a lack of code to close the log file in the case of cancle, causing the log file node to leak
![image](https://github.com/apache/kyuubi/assets/30170267/e685df16-cb1b-4a52-ae81-9c212095a2af)

### Affects Version(s)

master/1.7.1

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

ulysses-you:
Seems we should close operation log for canceled state.

ASiegeLion:
> 看来我们应该关闭已取消状态的操作日志。



> Seems we should close operation log for canceled state.

There is no operation to delete the operation log in the cancel method, so where is the operation log finally deleted?

ASiegeLion:
> Seems we should close operation log for canceled state.

There is no operation to delete the operation log in the cancel method, so where is the operation log finally deleted?

ulysses-you:
in `close` method, we should close operation log for canceled state.

PR #5211, opened 2023-08-29
[KYUUBI #5210]Cancel operation will cause the log file node to leak

ASiegeLion:


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperation.scala
```diff
@@ -158,6 +151,13 @@ abstract class KyuubiOperation(session: Session) extends AbstractOperation(sessi
         }
       }
     }
+    try {
+      // For launch engine operation, we use OperationLog to pass engine submit log but
+      // at that time we do not have remoteOpHandle
+      getOperationLog.foreach(_.close())
```
ulysses-you, line 157: what happens if this operation is closed ?

ASiegeLion: Sorry, I replied to your message but forgot to submit it

ulysses-you:
thanks, merged to master/1.8/1.7

PR #5215, opened 2023-08-30
[KYUUBI #4540][FOLLOWUP] Create session limiter if user unlimited list or deny list nonempty

turboFei:

We shall create session limiter if user unlimited list or deny list nonempty.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
@@ -385,14 +385,18 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       ipAddressLimit: Int,
       userIpAddressLimit: Int,
       userUnlimitedList: Set[String],
-      userLimitedList: Set[String]): Option[SessionLimiter] = {
-    Seq(userLimit, ipAddressLimit, userIpAddressLimit).find(_ > 0).map(_ =>
-      SessionLimiter(
+      userDenyList: Set[String]): Option[SessionLimiter] = {
+    if (Seq(userLimit, ipAddressLimit, userIpAddressLimit).exists(_ > 0) ||
+      Seq(userUnlimitedList, userDenyList).exists(_.nonEmpty)) {
+      Some(SessionLimiter(
         userLimit,
         ipAddressLimit,
         userIpAddressLimit,
         userUnlimitedList,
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -385,14 +385,18 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       ipAddressLimit: Int,
       userIpAddressLimit: Int,
       userUnlimitedList: Set[String],
-      userLimitedList: Set[String]): Option[SessionLimiter] = {
-    Seq(userLimit, ipAddressLimit, userIpAddressLimit).find(_ > 0).map(_ =>
-      SessionLimiter(
+      userDenyList: Set[String]): Option[SessionLimiter] = {
+    if (Seq(userLimit, ipAddressLimit, userIpAddressLimit).exists(_ > 0) ||
+      Seq(userUnlimitedList, userDenyList).exists(_.nonEmpty)) {
+      Some(SessionLimiter(
```
turboFei, line 391: should we always create the session limiter?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -385,14 +385,18 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       ipAddressLimit: Int,
       userIpAddressLimit: Int,
       userUnlimitedList: Set[String],
-      userLimitedList: Set[String]): Option[SessionLimiter] = {
-    Seq(userLimit, ipAddressLimit, userIpAddressLimit).find(_ > 0).map(_ =>
-      SessionLimiter(
+      userDenyList: Set[String]): Option[SessionLimiter] = {
+    if (Seq(userLimit, ipAddressLimit, userIpAddressLimit).exists(_ > 0) ||
+      Seq(userUnlimitedList, userDenyList).exists(_.nonEmpty)) {
```
wForget, line 390: `userUnlimitedList` should not be used as the basis for whether to create a limiter. Even if it is not empty and there are no other configurations, we do not need to create a limiter.

turboFei: thanks, updated

wForget (approved):

Issue #5216, opened 2023-08-30
[Bug] Limit counter resource leak in SessionLimiter.


zhuyaogai:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

1. When user opens session and before closing the session, this user is added to the unlimited.list, then the counter related to this user will not be cleared. 
2. The limit counter in SessionLimiter cannot be less than 0.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #5217, opened 2023-08-30
[KYUUBI #5216] Workaround for negative counter in SessionLimiter

zhuyaogai:

Fix: https://github.com/apache/kyuubi/issues/5216

No

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
@@ -95,9 +95,12 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
 
   private def decrLimitCount(key: String): Unit = {
     _counters.get(key) match {
-      case count: AtomicInteger =>
-        if (count.decrementAndGet() < 0) {
-          count.incrementAndGet()
+      case count: AtomicInteger if count.get > 0 =>
+        count.synchronized {
+          val countValue = count.get
+          if (countValue > 0) {
+            count.compareAndSet(countValue, countValue - 1)
+          }
         }
       case _ =>
     }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
@@ -96,12 +96,7 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
   private def decrLimitCount(key: String): Unit = {
     _counters.get(key) match {
       case count: AtomicInteger if count.get > 0 =>
-        count.synchronized {
-          val countValue = count.get
-          if (countValue > 0) {
-            count.compareAndSet(countValue, countValue - 1)
-          }
-        }
+        count.accumulateAndGet(1, (l, r) => if (l > 0) l - r else l)
       case _ =>
     }
   }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
@@ -95,7 +95,7 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
 
   private def decrLimitCount(key: String): Unit = {
     _counters.get(key) match {
-      case count: AtomicInteger if count.get > 0 =>
+      case count: AtomicInteger =>
         count.accumulateAndGet(1, (l, r) => if (l > 0) l - r else l)
       case _ =>
     }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
@@ -95,7 +95,10 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
 
   private def decrLimitCount(key: String): Unit = {
     _counters.get(key) match {
-      case count: AtomicInteger => count.decrementAndGet()
+      case count: AtomicInteger =>
+        if (count.decrementAndGet() < 0) {
+          count.incrementAndGet()
+        }
       case _ =>
     }
   }
@@ -118,12 +121,6 @@ class SessionLimiterWithAccessControlListImpl(
 
     if (!unlimitedUsers.contains(user)) {
       super.increment(userIpAddress)
-    }
-  }
-
-  override def decrement(userIpAddress: UserIpAddress): Unit = {
-    if (!unlimitedUsers.contains(userIpAddress.user)) {
-      super.decrement(userIpAddress)
     }
   }
 
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/session/SessionLimiterSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/session/SessionLimiterSuite.scala
@@ -20,8 +20,10 @@ import java.util.concurrent.{CountDownLatch, Executors}
 import java.util.concurrent.atomic.LongAdder
 
 import scala.collection.JavaConverters._
+import scala.util.Random
 
 import org.apache.kyuubi.{KyuubiFunSuite, KyuubiSQLException}
+import org.apache.kyuubi.util.ThreadUtils
 
 class SessionLimiterSuite extends KyuubiFunSuite {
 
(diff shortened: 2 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -95,7 +95,10 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
 
   private def decrLimitCount(key: String): Unit = {
     _counters.get(key) match {
-      case count: AtomicInteger => count.decrementAndGet()
+      case count: AtomicInteger =>
+        if (count.decrementAndGet() < 0) {
+          count.incrementAndGet()
+        }
```
bowenliang123, line 101:
```suggestion
      case count: AtomicInteger if count.get > 0 =>
        val countValue = count.get
        count.compareAndSet(countValue, countValue - 1)
```
Back and forth operations on AtomicInteger will cause side effects, as no an extra thread-safe measure applied here. 
Consider using the CAS for changing the value of AtomicInteger. And even if the CAS failed, it will be ignored and it won't cause side effects or exceptions.

bowenliang123 (changes requested):
Suggest to use CAS.

bowenliang123 (changes requested):
Suggest to use CAS.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
zhuyaogai: @bowenliang123 hi, I have tried your suggestion, but tests failed.

zhuyaogai: @pan3793 hi, my implementation can also pass tests, but @bowenliang123 proposed another implementation.Could you review it and give some suggestions? Thanks!

bowenliang123: A negative value is a situation that will happen but it's not acceptable for the counter.

wForget:
I think we can use a new limiter when updating the config instead of just updating the old limiter's configurations. Updating the limiter configuration is a very rare operation, so we can completely discard the previous state and use a new limiter. WDYT? cc @pan3793 @bowenliang123

wForget:
> I think we can use a new limiter when updating the config instead of just updating the old limiter's configurations. Updating the limiter configuration is a very rare operation, so we can completely discard the previous state and use a new limiter. WDYT? cc @pan3793 @bowenliang123

cc @turboFei

bowenliang123:
> I think we can use a new limiter when updating the config instead of just updating the old limiter's configurations. Updating the limiter configuration is a very rare operation, so we can completely discard the previous state and use a new limiter. WDYT? cc @pan3793 @bowenliang123

SGTM. Two main points are again on ‘Compare and Swap’ and thread-safe.

pan3793:
@wForget's idea SGTM, creating a new Limiter on configuration updates is simpler, and is consistent with the approach on engine decommission(the old one un-register from zk and wait for the existing session to close).

zhuyaogai:
Although creating a new Limiter on configuration updates is simpler, but still some thread-safe issues that need to be addressed  in `increment` or `decrement` method? I wonder if anyone would be willing to discuss with me, or provide some guidance, so that I can complete this PR. Thanks！

wForget:
> Although creating a new Limiter on configuration updates is simpler, but still some thread-safe issues that need to be addressed in `increment` or `decrement` method? I wonder if anyone would be willing to discuss with me, or provide some guidance, so that I can complete this PR. Thanks！

We only need to set a new limiter when updating the limiter configurations and there are no thread safety problem in the `increment` and `decrement` methods because they calculate the counter with `AtomicInteger`.

wForget:
@zhuyaogai I drafted the implementation steps, do you think it's feasible?

+ remove `SessionLimiterWithAccessControlListImpl#setUnlimitedUsers` and `SessionLimiterWithAccessControlListImpl#setDenyUsers` methods
+ add `KyuubiSessionManager.refreshLimiterWithConfig` method instead of `SessionLimiter.refreshUnlimitedUsers`(and `refreshDenyUsers`) to reset new limiter with new configs
+ call `refreshLimiterWithConfig` in `refreshUnlimitedUsers` and `refreshDenyUsers`
+ add `@volatile` to `KyuubiSessionManager.limiter` var (and batchLimiter)

zhuyaogai:
@wForget hi, thanks for your guidance, but I have some questions.
1. What would be the result if  increase the counter in the old limiter, then refresh the limiter, and finally decrease the counter in the new limiter?
2. > When user opens session and before closing the session, this user is added to the unlimited.list, then the counter related to this user will not be cleared.
The limit counter in SessionLimiter cannot be less than 0. 

Should the bugs described in [#5216](https://github.com/apache/kyuubi/issues/5216) also be fixed together?

zhuyaogai:
@wForget hi, thanks for your guidance, but I have some questions.

1. What would be the result if increase the counter in the old limiter, then refresh the limiter, and finally decrease the counter in the new limiter? 
2. >When user opens session and before closing the session, this user is added to the unlimited.list, then the counter related to this user will not be cleared.
The limit counter in SessionLimiter cannot be less than 0.

Should the bugs described in #5216 also be fixed together?

wForget:
@zhuyaogai Sorry for the late reply, you are right, the prerequisite for using a new limiter is that the inc and dec calculations of the same session need to act on the same limiter. It is difficult to do that because the current limiter calculation is in the SessionManager.

wForget:
Let's go back to this issue. I think there are two inconsistencies, the user limit counter is not cleared when user is added to unlimited user list, and the user running sessions are not added to user limit counter when removing user from unlimited user list.

I offer two solutions for this:
1. Put the limiter calculation in the session, ensure that the `inc` and `dec` calculations of limiter act on the same limiter, and then use new limiter when updating configurations (as discussed above).
2. When updating configurations, clear limiter counter or add the running session to limiter counter. This needs to be considered to ensure synchronization.

zhuyaogai:
@wForget hi, thanks for your guidance! I feel that following your solutions has become a bit complicated? Could you tell me if my initial solution is feasible? Or what considerations led to the choice of 
> I think we can use a new limiter when updating the config instead of just updating the old limiter's configurations. 

I am really looking forward to your answer, thank you!

wForget:
@zhuyaogai Thanks for your work, I think your initial solution is feasible but not the best, so I +0 for this. If there are always running sessions for a user then the count will never be less than 0, so our limiter may not be accurate.
@pan3793 @bowenliang123 @turboFei WDYT?

bowenliang123:
I'm also +0 on the initial implementation, which is practicable but also dangling in fore-and-backward side effects.
It will be fine to me if any other committer approves this.

My further concern about the counter mechanism is that we may have a common util for solving similar situations in mapped counters.

zhuyaogai:
> I'm also +0 on the initial implementation, which is practicable but also dangling in fore-and-backward side effects. It will be fine to me if any one approves this.
> 
> My further concern about the counter mechanism is that we may have a common util for solving similar situations in mapped counters.

@bowenliang123 hi, I have applied your code suggestions :)

bowenliang123:
> @bowenliang123 hi, I have applied your code suggestions :)

Thanks for the feed back, thus it might be not percfect.

bowenliang123 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -95,7 +95,13 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
 
   private def decrLimitCount(key: String): Unit = {
     _counters.get(key) match {
-      case count: AtomicInteger => count.decrementAndGet()
+      case count: AtomicInteger if count.get > 0 =>
+        count.synchronized {
```
wForget, line 99: It seems like this is enough:
```
count.accumulateAndGet(1, (l, r) => if (l > 0) l - r else l)
```

zhuyaogai: @wForget @bowenliang123 hi, could you tell me, in a user session, if the call to SessionLimiter's `increment` method fails(limited by something) or TFrontendService's `OpenSession` method fails, will it finally call the SessionLimiter's `decrement` method? If not, I think there might be some issues.

wForget: It will always be called, even if close session is not called correctly, it will be cleaned up by `SessionManager#timeoutChecker`.

bowenliang123: +1

zhuyaogai: > It seems like this is enough:
> 
> ```
> count.accumulateAndGet(1, (l, r) => if (l > 0) l - r else l)
> ```

@wForget Thanks for your answer, and I have adopted your code suggestions.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/SessionLimiter.scala
```diff
@@ -95,7 +95,8 @@ class SessionLimiterImpl(userLimit: Int, ipAddressLimit: Int, userIpAddressLimit
 
   private def decrLimitCount(key: String): Unit = {
     _counters.get(key) match {
-      case count: AtomicInteger => count.decrementAndGet()
+      case count: AtomicInteger if count.get > 0 =>
```
wForget, line 98: `if count.get > 0` is no longer needed.

zhuyaogai: done.

wForget (approved):

pan3793:
Thanks for the surgery, merging to master/1.8.

PR #5220, opened 2023-08-30
Batch submitter should only block submitting stage

pan3793:

This PR changes the block phase of the batch submitter

- before: from PENDING until TERMINATED 
- after: from PENDING until RUNNING or TERMINATED

Usually, we submit Spark batch applications in cluster mode with waitAppCompletion disabled, after a Spark application goes into the RUNNING or TERMINATED stage, the `spark-submit` process exits and the batch session does not occupy too many resources. Thus, limiting the concurrency on the submitting phase instead of the whole lifecycle of the Spark app makes more sense. 

In practice, we use 16 threads for Kyuubi instance with 8C32G. A larger concurrency may result in CPU resources being exhausted and `spark-submit` process hanging. 

No.


turboFei (approved):

cxzl25 (approved):

pan3793:
Thanks, merged to master

PR #5221, opened 2023-08-30
[DO-NOT-MERGE] Flink sql engine compilation on Scala 2.13

bowenliang123:

- make Flink SQL Engine module compilable on Scala 2.13
- However, as Flink only provides and is embedded with Scala 2.12 runtime, there's no possible way or gain to use Flink SQL Engine on Scala 2.13.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -189,8 +189,7 @@ jobs:
         uses: ./.github/actions/setup-maven
       - name: Build on Scala ${{ matrix.scala }}
         run: |
-          MODULES='!externals/kyuubi-flink-sql-engine'
-          ./build/mvn clean install -pl ${MODULES} -am \
+          ./build/mvn clean install -am \
           -DskipTests -Pflink-provided,hive-provided,spark-provided \
           -Pjava-${{ matrix.java }} \
           -Pscala-${{ matrix.scala }} \
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
@@ -90,6 +90,7 @@ object FlinkSQLEngine extends Logging {
       val flinkConfFromArgs =
         kyuubiConf.getAll.filterKeys(_.startsWith("flink."))
           .map { case (k, v) => (k.stripPrefix("flink."), v) }
+          .toMap
       flinkConf.addAll(Configuration.fromMap(flinkConfFromArgs.asJava))
 
       val executionTarget = flinkConf.getString(DeploymentOptions.TARGET)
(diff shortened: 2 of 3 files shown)


pan3793:
Simply skipping the flink module should be fine?

bowenliang123:
With this patch, the Flink SQL Engine is compilable on Scala 2.13.
But as Flink only provides and embeds Scala 2.12 runtime inside, there's no possible way or gain to use Flink SQL Engine on Scala 2.13.

I would suggest locking the Scala version to 2.12 in Flink SQL Engine module.
And force to use 2.12 when submitting flink SQL engine, even if using Kyuubi Server on Scala 2.13.

@pan3793 @link3280

link3280:
I think it's a reasonable approach. My concern is that the Flink engine module depends on 2.13 modules like `kyuubi-common`, would that cause compatibility issues?

link3280:
Another approach is to upgrade Flink engine to scala 2.13 anyway. The downside is that we won't be able to use scala expressions or UDFs in Flink engine. It's acceptable because Flink is planning to drop scala support.

Yikf:
If it is confirmed that flink-table-planner_2.12 is not used, then I agree that upgrade Flink engine to scala 2.13 to keep the scala version of all modules is uniform.

bowenliang123:
> I think it's a reasonable approach. My concern is that the Flink engine module depends on 2.13 modules like `kyuubi-common`, would that cause compatibility issues?

It has been proved to cause compatibility issues when running a Flink Engine module compiled with Scala 2.13 on Scala 2.12 runtime which is embedded inside Flink. Either The engine itself or common modules like `kyuubi-common` depends on Scala API implicitly. There are unavoidable breaking changes (e.g alias, collections)between Scala 2.12 and Scala 2.13.

Take the following example, it's caused by the changes in `scala.Serilizable`, as it extends Java's Serilizable in Scala 2.12 while it IS Java's Serilizable in Scala 2.13. 

```
PlanOnlyOperationSuite:
org.apache.kyuubi.engine.flink.operation.PlanOnlyOperationSuite *** ABORTED ***
  java.lang.RuntimeException: Unable to load a Suite class org.apache.kyuubi.engine.flink.operation.PlanOnlyOperationSuite that was discovered in the runpath: java.lang.NoClassDefFoundError: scala/Serializable
  at org.scalatest.tools.DiscoverySuite$.getSuiteInstance(DiscoverySuite.scala:80)
  at org.scalatest.tools.DiscoverySuite.$anonfun$nestedSuites$1(DiscoverySuite.scala:38)
  at scala.collection.immutable.Vector1.map(Vector.scala:1872)
  at scala.collection.immutable.Vector1.map(Vector.scala:375)
  at org.scalatest.tools.DiscoverySuite.<init>(DiscoverySuite.scala:37)
  at org.scalatest.tools.Runner$.genDiscoSuites$1(Runner.scala:1131)
  at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:1225)
  at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24(Runner.scala:992)
  at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24$adapted(Runner.scala:970)
  at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:1481)
  ...
  Cause: java.lang.BootstrapMethodError: java.lang.NoClassDefFoundError: scala/Serializable
  at org.apache.kyuubi.util.EnumUtils$.isValidEnums(EnumUtils.scala:27)
  at org.apache.kyuubi.config.TypedConfigBuilder.$anonfun$checkValues$2(ConfigBuilder.scala:212)
  at org.apache.kyuubi.config.TypedConfigBuilder.$anonfun$transform$1(ConfigBuilder.scala:170)
```

PR #5228, opened 2023-08-31
Check config value in numeric or duration range

bowenliang123:

- Add and apply some helper methods to check config value in numeric or duration range, which may also helps unifying error message for invalid value
   - `checkPositive`
   - `checkNonnegative`
   - `checkRange` by NumericRange
   - `checkPattern` for string type
   - `checkDuration` by min/max range for duration type

No.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigBuilder.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigBuilder.scala
@@ -19,12 +19,15 @@ package org.apache.kyuubi.config
 
 import java.time.Duration
 import java.util.Locale
-import java.util.regex.PatternSyntaxException
+import java.util.regex.{Pattern, PatternSyntaxException}
 
 import scala.util.{Failure, Success, Try}
 import scala.util.matching.Regex
 
+import org.apache.commons.lang3.StringUtils
+
 import org.apache.kyuubi.util.EnumUtils._
+import org.apache.kyuubi.util.MathUtils._
 
 private[kyuubi] case class ConfigBuilder(key: String) {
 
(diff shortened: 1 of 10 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -342,31 +343,31 @@ object KyuubiConf {
...
+      .checkPositive
       .createWithDefault(Duration.ofMinutes(1).toMillis)
 
   val CREDENTIALS_CHECK_INTERVAL: ConfigEntry[Long] =
     buildConf("kyuubi.credentials.check.interval")
       .doc("The interval to check the expiration of cached <user, CredentialsRef> pairs.")
       .version("1.6.0")
       .timeConf
-      .checkValue(_ > Duration.ofSeconds(3).toMillis, "Minimum 3 seconds")
+      .checkDurationRange(min = Some(Duration.ofSeconds(3)))
```
cxzl25, line 362: `_ >` should have different semantics from `checkDurationRange`.

bowenliang123: It's very likely an original typing mistake from #1021. The 'Minimum' term in the error message also hints to allow `>=` inclusively rather than `>` exclusively.
cc @lightning-L

bowenliang123:
Do you have some time to have a look at this guarding checks for config values? If we docide not to accept it , let me close this PR. @pan3793

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
lightning-L: Oh....
I am okay with the change. Thanks

kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/NumUtils.scala
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @param o the object to be checked
+   * @return true if the object is a non-negative number
+   */
+  def isNonnegativeNumber(o: Any): Boolean = o match {
+    case i: Int => i >= 0
+    case l: Long => l >= 0
+    case f: Float => f >= 0
+    case d: Double => d >= 0
+    case _ => Try { o.toString.toDouble >= 0 }.isSuccess
```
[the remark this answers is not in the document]
bowenliang123: Let me use explicit data type for the zero.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigBuilder.scala
```diff
@@ -194,6 +197,72 @@ private[kyuubi] case class TypedConfigBuilder[T](
...
+      case "boolean" => "Boolean"
+      case "string" => "String"
+      case "duration" => "Duration"
+      case _ => "Unknown"
+    }
+
+  def checkPositive: TypedConfigBuilder[T] =
+    checkValue(isPositiveNumber(_), s"Must be a positive $printableUnderlyingType.")
+
+  def checkNonnegative: TypedConfigBuilder[T] =
```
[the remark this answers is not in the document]
bowenliang123: Fixed.

kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/MathUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+   */
+  def isNonNegativeNumber(o: Any): Boolean = o match {
+    case intValue: Int => intValue >= 0
+    case longValue: Long => longValue >= 0L
+    case floatValue: Float => floatValue >= 0.0f
+    case doubleValue: Double => doubleValue >= 0.0d
+    case byteValue: Byte => byteValue >= 0
+    case shortValue: Short => shortValue >= 0
+    case charValue: Char => charValue >= 0
+    case _ => Try { o.toString.toDouble >= 0.0d }.isSuccess
```
pan3793, line 50: the logic is incorrect here, it only guards the type cast, ignores the expression result

pan3793: BTW, should we narrow the `o`'s types?

bowenliang123: > the logic is incorrect here, it only guards the type cast, ignores the expression result

Yes, noticed. Fixed and added proper unit test for this.

bowenliang123: > BTW, should we narrow the `o`'s types?

In what way or what's the ranger of types?

pan3793: Are Int, Long, Float, Double, String sufficient?

bowenliang123: Primitive data types have been covered, meanwhile other types are converted to string and then to a double for comparison. Should be good enough for all cases.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigBuilder.scala
```diff
@@ -194,6 +197,72 @@ private[kyuubi] case class TypedConfigBuilder[T](
...
+  /**
+   * Check that user-provided values for the duration config
+   * with in the range of min and max.
+   * @param min inclusive min value, if provided
+   * @param max inclusive max value, if provided
+   * @param errMsg custom error message over default message if provided and not blank
+   */
+  def checkDurationRange(
+      min: Option[Duration] = None,
+      max: Option[Duration] = None,
```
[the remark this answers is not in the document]
pan3793: I prefer None, but force-inclusive is not flexible

kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/MathUtils.scala
```diff
@@ -0,0 +1,63 @@
...
+    case doubleValue: Double => doubleValue > 0.0d
+    case byteValue: Byte => byteValue > 0
+    case shortValue: Short => shortValue > 0
+    case charValue: Char => charValue > 0
+    case _ =>
+      try {
+        o.toString.toDouble > 0.0d
+      } catch {
+        case _: Exception => false
+      }
```
pan3793, line 39: is it sufficient?
```suggestion
    case stringValue: String => stringValue.toDouble > 0.0d
    case _ => throw illegalArgException ...
```

kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/MathUtils.scala
pan3793: I mean narrow the range of data types

bowenliang123: Adopted your suggestion. And it will throw an IllegalArgumentException for unsupported types.

bowenliang123:

PR #3931, opened 2022-12-07
[KYUUBI #3924] checks source table for Iceberg metadata table

bowenliang123:
### _Why are the changes needed?_

to close #3924 

### _How was this patch tested?_
- [x] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #5248, opened 2023-09-04
[AUTHZ] Check previleges for Iceberg's CALL to RewriteDataFilesProcedure procedure

yabola:

Before this PR, Kyuubi skips authentication for the call command in iceberg. After this PR, it needs `alter` permission to use this command.
This PR supports checking iceberg call rewrite_data_files command:
- CALL ice.system.rewrite_data_files(table => 'ice.lcl.test_ice', options => map('min-input-files','2'))
- some other call statements in iceberg ( https://iceberg.apache.org/docs/latest/spark-procedures ) may be supported but needs to be further determined. The next phase is planned to support all commands as documented.

When the `rewrite_data_files` command is called, there are two situations:
1. When the triggering conditions are not be met(such as input-files < min-input-files) , there is only one logical plan.  We  need to check one logical plan permission:
```
        * == Physical Plan 1 ==
       * Call (1)
```
2. When the triggering  conditions are met, two logical plan will be triggered. We should check two logical plans permission.
```
       * == Physical Plan 1 ==
       * Call (1)
       *
       * == Physical Plan 2 ==
       * AppendData (3)
       * +- * ColumnarToRow (2)
       * +- BatchScan local.iceberg_ns.call_command_table (1)
```
       
The second plan (`AppendData`) is a bit special,  causing us to have to modify the logic of extracting table names in `DataSourceV2RelationTableExtractor`. The `DataSourceV2Relation#identifier` in `AppendData` will be rewrite task id, so it is a more appropriate way to parse the table name from `DataSourceV2Relation#table`
image

### _Was this patch authored or co-authored using generative AI tooling?
No.


pan3793:
@yabola thanks for your contribution. You may want to send a minor PR(e.g. bump iceberg from 1.3.0 to 1.3.1) first to unblock the CI (auto-run without approval)

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/serde/tableExtractors.scala
```diff
@@ -145,14 +175,37 @@ class DataSourceV2RelationTableExtractor extends TableExtractor {
         val maybeCatalogPlugin = invokeAs[Option[AnyRef]](v2Relation, "catalog")
         val maybeCatalog = maybeCatalogPlugin.flatMap(catalogPlugin =>
           lookupExtractor[CatalogPluginCatalogExtractor].apply(catalogPlugin))
-        val maybeIdentifier = invokeAs[Option[AnyRef]](v2Relation, "identifier")
```
yabola, line 148: The `identifier` in `DataSourceV2Relation` is not reliable enough. In fact, datasourceV2 will parse out the table based on identifier, so table is more appropriate here.
For example, in iceberg `rewrite_data_files` command, `identifier` is rewrite task id, iceberg catalog will load table by it ( can see in `SparkCatalog#load` the `isPathIdentifier` function)

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/serde/tableExtractors.scala
```diff
@@ -132,6 +134,34 @@ class IdentifierTableExtractor extends TableExtractor {
...
+class IcebergCallCommandExtractor extends TableExtractor {
+  override def apply(spark: SparkSession, callStatementArgs: AnyRef): Option[Table] = {
+    val literals = callStatementArgs.asInstanceOf[Seq[Literal]]
+    if (literals.isEmpty) {
+      throw new IllegalArgumentException(
+        s"Unknown parameters in iceberg call command: $callStatementArgs")
+    }
+    // iceberg will rearrange the parameters according to the parameter order
+    // defined in the procedure, where the table parameters are currently always the first.
+    lookupExtractor[IcebergTableExtractor].apply(spark, literals.head)
```
yabola, line 146: iceberg will rearrange the parameters according to the parameter order defined in the procedure, where the table parameters are currently always the first.
For example `RewriteDataFilesProcedure#PARAMETERS` and in `RewriteDataFilesProcedure#call` take the first parameter.

bowenliang123:
Hi @yabola, thanks for the contribution to the Authz plugin. And pleased to see introducing support for CALL statements for Iceberg which are critical to lakehouse lifecycle.

Some advices here,
1. Move the command spec generation to `authz.gen` package's `IcebergCommands`, which then is used in `JsonSpecFileGenerator` for spec json file generation by running `dev/gen/gen_ranger_spec_json.sh` locally . And also running `KYUUBI_UPDATE=0 dev/gen/gen_ranger_spec_json.sh` to check spec file.
2. if possible, try not to introduce a new TableExtractor specifically for Iceberg. Iceberg commands are generally supported with existed extractors.

yabola:
@bowenliang123 Thank you for your advice.
1. I add new method in `TableCommands` and generate json by `KYUUBI_UPDATE=1 build/mvn clean test -Pgen-policy -pl :kyuubi-spark-authz_2.12 -Dtest=none -DwildcardSuites=org.apache.kyuubi.plugin.spark.authz.gen.JsonSpecFileGenerator` command. Please see if it is suitable.
2. I removed one unnecessary table extractor. But there may a new table extractor may need to be added because the iceberg call command is a new logical plan(`Call(procedure: Procedure, args: Seq[Expression])`). Or do you have a better suggestion?

bowenliang123:
1. Please move `IcebergCallCommand ` to `IcebergCommands` class as mentioned above, as `org.apache.spark.sql.catalyst.plans.logical.Call` is from Iceberg but not a general plan from Spark engine. And result spec file is check by CI tests.
2. It's feasible to introduce a new table extractor for the Call plan.
3. Try not to introduce `resolveIcebergTable` in `tableExtractors`, and do the extraction by reusing existed extractors following the main procedures. Reversed coupling is always unwelcomed. Any other lakehouse may and will be supported by design with in separated command extractors and specs, while sharing the same common structure.

yabola:
1. I have moved to `IcebergCallCommand`. Sorry didn't notice...
3. Yes, there should share the same common structure. I think we'd better use the `table` parameter instead of `identifier` here. The reason:
> The identifier in DataSourceV2Relation is not reliable enough. In fact, datasourceV2 will parse out the table based on identifier, so table is more appropriate here.
For example, in iceberg rewrite_data_files command, identifier is rewrite task id, iceberg catalog will load table by it ( can see in SparkCatalog#load the isPathIdentifier function)

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/gen/IcebergCommands.scala
```diff
@@ -49,10 +49,22 @@ object IcebergCommands {
     TableCommandSpec(cmd, Seq(tableDesc), queryDescs = Seq(queryDesc))
   }
 
+  val IcebergCallCommand = {
+    val cmd = "org.apache.spark.sql.catalyst.plans.logical.Call"
+    val actionTypeDesc = ActionTypeDesc(actionType = Some(INSERT))
```
bowenliang123, line 54: `UPDATE` is the more proper one for the action type.

yabola: Thanks, done

bowenliang123:
Pleased to see the progress of the added e2e tests which passed CI tests.
Still some concerns in the following changes, that I may have further investigation or proof.
1. Again, change the source field inside the `DataSourceV2RelationTableExtractor`, from "identifier" to `Table` "table"
2. the introduced `StringTableExtractor` 
3. forced casting to Literal in `IcebergCallArgsTableExtractor `

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/IcebergCatalogPrivilegesBuilderSuite.scala
```diff
@@ -125,4 +125,26 @@ class IcebergCatalogPrivilegesBuilderSuite extends V2CommandsPrivilegesSuite {
...
+        s"(table => '$tableId', options => map('min-input-files','2')) ").queryExecution.analyzed
+      val (inputs, outputs, operationType) = PrivilegesBuilder.build(plan, spark)
+      assert(operationType === QUERY)
+      assert(outputs.size === 1)
+      val po = outputs.head
+      assert(po.actionType === PrivilegeObjectActionType.UPDATE)
+      assert(po.privilegeObjectType === PrivilegeObjectType.TABLE_OR_VIEW)
+      assertEqualsIgnoreCase(namespace)(po.dbname)
+      assertEqualsIgnoreCase(tableId.split("\\.").last)(po.objectName)
+      assert(inputs.size === 0)
```
bowenliang123, line 145: The SELECT query access privilege of the source table should be required when calling the table rewriting produce.

yabola:
@bowenliang123 Thank you for the improvement

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/serde/tableExtractors.scala
bowenliang123: Oh, got your point. Iceberg procedure rewrites this identifier, and makes it irrelevant to the original table.

bowenliang123: I think we need to `TableTableExtractor` for taking `Table` input, instead of use `StringTableExtractor` directly.

bowenliang123: Please attach the evidence showing the impacts and differences in logical plan details of Iceberg call procedure.

yabola: When CALL ice.system.rewrite_data_files command does the rewrite operation, two logical plans will be generated, for example in my unit test:

> 
  * This triggers two logical plans( input-files(2) >= min-input-files(2) ):
   *
   * == Physical Plan 1 ==
   * (1) Call
   *
   * == Physical Plan 2 ==
   * AppendData (3)
   * +- * ColumnarToRow (2)
   * +- BatchScan local.iceberg_ns.call_command_table (1)

 In the second logical plan, the `DataSourceV2Relation#identifier` in `AppendData` will be rewrite task id. So it is a more appropriate way to parse the table name from `DataSourceV2Relation#table`
 
image

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/IcebergCatalogRangerSparkExtensionSuite.scala
```diff
@@ -235,13 +235,15 @@ class IcebergCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite
       s" on [$namespace1/$table1]"))
   }
 
-  test("CALL RewriteDataFilesProcedure") {
-    val tableName = "call_command_table"
-    val table = s"$catalogV2.$namespace1.$tableName"
-    val rewriteCommand1 = s"CALL $catalogV2.system.rewrite_data_files " +
+  test("CALL RewriteProcedure") {
```
bowenliang123, line 238: We'd better use RewriteDataFilesProcedure, refering to `org.apache.iceberg.spark.procedures.RewriteDataFilesProcedure` from Iceberg Spark plugin.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/IcebergCatalogRangerSparkExtensionSuite.scala
```diff
@@ -233,4 +234,69 @@ class IcebergCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite
...
+
+    withCleanTmpResources(Seq((table, "table"))) {
+      doAs(
+        admin, {
+          sql(s"CREATE TABLE IF NOT EXISTS $table  (id int, name string) USING iceberg")
+          sql(s"INSERT INTO $table VALUES (1, 'chenliang'), (2, 'tom')")
+          sql(s"INSERT INTO $table VALUES (3, 'julie'), (4, 'ross')")
+          snapshotId = sql(s"select * from $table.snapshots limit 1")
+            .collect().apply(0).getAs[Long]("snapshot_id")
+          snapshotCommand = s"CALL $catalogV2.system.set_current_snapshot ('$table', $snapshotId)"
```
bowenliang123, line 256: Let's skip testing on `CALL system.set_current_snapshot` procedure this time. Move to the follow-ups.

yabola:
@bowenliang123 Thank you for adding new ranger policy for UT and the relevant code improvements. And is UT failure due to instability?

bowenliang123:
> @bowenliang123 Thank you for adding new ranger policy for UT and the relevant code improvements. And is UT failure due to instability?

Hi, updated the table spec json just now. I changed the operation time from QUERY to ALTERTABLE_ADDPARTS, which is not a best one but a closer one.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/IcebergCatalogRangerSparkExtensionSuite.scala
```diff
@@ -226,4 +227,71 @@ class IcebergCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite
...
+        admin, {
+          val result2 = sql(rewriteDataFiles1).collect()
+          // rewrite 2 files
+          assert(result2(0)(0) === 2)
+        })
+
+      /**
+       * Situation 2:
+       * Only one logical plan triggered
+       * when ( input-files(2) < min-input-files(3) )
```
bowenliang123, line 288: Could you attach the triggered pysical plan for this situation as well? @yabola

yabola: I have added the physical plan.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/IcebergCatalogRangerSparkExtensionSuite.scala
```diff
@@ -226,4 +227,74 @@ class IcebergCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite
...
+       *
+       * == Physical Plan 2 ==
+       * AppendData (3)
+       * +- * ColumnarToRow (2)
+       * +- BatchScan local.iceberg_ns.call_command_table (1)
+       */
+      doAs(
+        admin, {
+          val result2 = sql(rewriteDataFiles1).collect()
+          // rewrite 2 files
```
bowenliang123, line 281: Do you mean 'result to 2 rewritten data files' or 'exited 2 files before/after procedure'?

yabola: It means number of data files which were re-written by this command.  The explanation of the output of this command can refer to https://iceberg.apache.org/docs/latest/spark-procedures/#rewrite_data_files

bowenliang123:
"Permission denied: user [someone] does not have [alter] privilege on [iceberg_ns/table_select_call_command_table]"
 did not contain 
"does not have [select] privilege on [iceberg_ns/table_select_call_command_table], [update] privilege on [iceberg_ns/table_select_call_command_table]"

PrivilegesBuilder extracted two privilege object for input and output.
But their action type both are `PrivilegeObjectActionType.OTHER` (Possible value of PrivilegeObjectActionType are OTHER, INSERT, INSERT_OVERWRITE, UPDATE, DELETE).
Therefore, RuleAuthorization would then treat them as the same privilege object, and check the `alter` privilege only.

1.  it seems we could not distinguish them, by 1 for `select` on input table and 1 for `alter` on output table. cc @yaooqinn 
2. Or let's check alter privilege only, and skip checking `select` privilege, as the same and only table used in one procedure. @yabola

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/IcebergCatalogRangerSparkExtensionSuite.scala
bowenliang123: OK, thanks. Please mention these details in the description.

yabola:
> ```
> "Permission denied: user [someone] does not have [alter] privilege on [iceberg_ns/table_select_call_command_table]"
>  did not contain 
> "does not have [select] privilege on [iceberg_ns/table_select_call_command_table], [update] privilege on [iceberg_ns/table_select_call_command_table]"
> ```
> 
> PrivilegesBuilder extracted two privilege objects for input and output. But their action type both are `PrivilegeObjectActionType.OTHER` (Possible values of PrivilegeObjectActionType are OTHER, INSERT, INSERT_OVERWRITE, UPDATE, DELETE). Therefore, RuleAuthorization would then treat them as the same privilege object, and check the `alter` privilege only.
> 
> 1. it seems we could not distinguish them, by 1 for `select` on input table and 1 for `alter` on output table. cc @yaooqinn
> 2. Or let's check alter privilege only, and skip checking `select` privilege, as the same and only table used in one procedure, and no data changed by calling the procedure. @yabola

I think having a single authority is more direct and easier for people to understand.

bowenliang123 (approved):
LGTM.
All Authz tests passed on the local machine.

bowenliang123:
Thank you for your contribution and patience. Looking forward to your follow-up PRs in improving Authz support for Iceberg.
Merged to master (1.9.0).

elisabetao:
Following the TableExtractor adjustments it seems the Iceberg metadata tables (snapshots,manifests,history) are being validated for authorization . Is this expected?  Thanks a lot or the help and feedback

bowenliang123:
> Following the TableExtractor adjustments it seems the Iceberg metadata tables (snapshots,manifests,history) are being validated for authorization . Is this expected? Thanks a lot or the help and feedback

#3931 may help, but it has not been accepted.

PR #5279, opened 2023-09-13
[KYUUBI #5220][FOLLOWUP] Batch submitted considers application state

pan3793:

This PR aims to fix the `SparkSubmit` concurrency limit implemented in #5220.

"submitted" judgment in #5220 only considered `OperationState`, actually, `ApplicationState` should be counted too. For instance, if a batch is pending in `ACCEPTED` state, the `SparkSubmit` process won't exit until changed to `RUNNING` or `FAILED` state, in such case, the `OperationState` is `RUNNING` and `ApplicationState` is `PENDING`, it should not be treated as "submitted".

Additionally, this PR treats metastore as the single of truth for batch instead of `SessionManager`

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
@@ -93,12 +93,12 @@ class KyuubiBatchService(
                 case Some(metadata) if metadata.opState == OperationState.RUNNING =>
                   metadata.appState match {
                     // app that is not submitted to resource manager
-                    case None | Some(ApplicationState.NOT_FOUND) => true
+                    case None | Some(ApplicationState.NOT_FOUND) => false
                     // app that is pending in resource manager
-                    case Some(ApplicationState.PENDING) => true
+                    case Some(ApplicationState.PENDING) => false
                     // not sure, added for safe
-                    case Some(ApplicationState.UNKNOWN) => true
-                    case _ => false
+                    case Some(ApplicationState.UNKNOWN) => false
+                    case _ => true
                   }
                 case Some(_) =>
                   false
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
```diff
@@ -83,26 +84,28 @@ class KyuubiBatchService(
...
-                batchState == OperationState.RUNNING || OperationState.isTerminal(batchState)
-              }.getOrElse {
-                error(s"Batch Session $batchId is not existed, marked as finished")
-                true
+              submitted = metadataManager.getBatchSessionMetadata(batchId) match {
+                case Some(metadata) if OperationState.isTerminal(metadata.opState) =>
+                  true
+                case Some(metadata) if metadata.opState == OperationState.RUNNING =>
+                  metadata.appState match {
+                    // app that is not submitted to resource manager
```
zwangsheng, line 95: Maybe we can assume when submit operation is running, 
- app state in PENDING, RUNNING, FINISHED, KILLED, FAILED means been submitted(return true)
- app state in NOT_FOUND or UNKNOWN we should marked as un-submitted and wait for out-range fail.

cfmcgrady (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
pan3793: The purpose of such concurrency limitation is that the SparkSubmit process swallows too many CPU and memory resources, and the SparkSubmit process won't exit when the app state is PENDING, thus PENDING should be treated as not submitted

zwangsheng (approved):
LGTM

pan3793:
Merged to master/1.8

PR #5280, opened 2023-09-14
Hive engine compatible with Hive 2.3

pan3793:

Try to make the Hive engine compatible with Hive 2.3.

It is not easy to align all dependencies with Hive 2.3, thus this PR just added a new integration test to make sure that `kyuubi hive engine` compiled against Hive 3.1.3 could work with Hive 2.3.9 runtime, instead of introducing new a profile to make it pass building with Hive 2.3.9

There are potential class conflict issues, because the Apache Hive 2.3.9 and 3.1.3 official binary artifacts ship Scala 2.11 jars but Kyuubi uses Scala 2.12.

Also tested with `apache-hive-2.1.1-cdh6.3.2-bin.tar.gz` locally, which can successfully bootstrap and run queries.

No

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -242,7 +242,12 @@ jobs:
       matrix:
         java:
           - 8
+        hive-archive: [ "" ]
         comment: [ "normal" ]
+        include:
+          - java: 8
+            hive-archive: '-Dhive.archive.mirror=https://archive.apache.org/dist/hive/hive-2.3.9 -Dhive.archive.name=apache-hive-2.3.9-bin.tar.gz'
+            comment: 'verify-on-hive-2.3-binary'
     steps:
       - uses: actions/checkout@v3
       - name: Tune Runner VM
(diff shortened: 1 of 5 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/hive/HiveProcessBuilder.scala
```diff
@@ -116,8 +116,6 @@ class HiveProcessBuilder(
     buffer.toArray
   }
 
-  override def toString: String = Utils.redactCommandLineArgs(conf, commands).mkString("\n")
```
pan3793, line 119: the output is ugly, prefer to use the parent implementation

.github/workflows/master.yml
```diff
@@ -261,8 +266,15 @@ jobs:
       - name: Build and test Hive with maven w/o linters
         run: |
           TEST_MODULES="externals/kyuubi-hive-sql-engine,integration-tests/kyuubi-hive-it"
-          ./build/mvn ${MVN_OPT} -pl ${TEST_MODULES} -am clean install -DskipTests
-          ./build/mvn ${MVN_OPT} -pl ${TEST_MODULES} test
+          ./build/mvn ${MVN_OPT} ${{ matrix.hive-archive }} -pl ${TEST_MODULES} -am clean install -DskipTests
+          # Hive 2.3.9 ships Derby 10.10.2.0, which may fail to boostrap on latest JDK 8
+          # https://github.com/apache/hive/pull/4895
+          if [[ "${{ matrix.hive-archive }}" == *apache-hive-2.3.9-bin.tar.gz* ]]; then
+            HIVE_239_LIB="$PWD/externals/kyuubi-download/target/apache-hive-2.3.9-bin/lib"
```
Yikf, line 273: Nice job! I suggest that to record this information in the official document.

pan3793: This only affects users who use derby as the embedded metastore storage backend, not a production solution

pan3793: but it worth to write a dedicated docs page for hive engines deployment

Yikf: > This only affects users who use derby as the embedded metastore storage backend, not a production solution

Get

cxzl25 (approved):

pan3793:
Thanks, merged to master/1.8

Issue #5212, opened 2023-08-29
[Bug] java.lang.IllegalArgumentException: 'ArrayBuffer(PROMETHEUS, REST, THRIFT_BINARY)' in kyuubi.frontend.protocols is invalid.


FourSpaces:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I got this problem when deploying Kyuubi with k8s charts of Kyuubi

### Affects Version(s)

1.7.1

### Kyuubi Server Log Output

```logtalk
2023-08-29 07:00:05.228 INFO org.apache.kyuubi.zookeeper.EmbeddedZookeeper: Service[EmbeddedZookeeper] is stopped.
Exception in thread "main" java.lang.IllegalArgumentException: 'ArrayBuffer(PROMETHEUS, REST, THRIFT_BINARY)' in kyuubi.frontend.protocols is invalid. the frontend protocol should be one or more of THRIFT_BINARY,THRIFT_HTTP,REST,MYSQL,TRINO
	at org.apache.kyuubi.config.TypedConfigBuilder.$anonfun$checkValue$1(ConfigBuilder.scala:173)
	at org.apache.kyuubi.config.TypedConfigBuilder.$anonfun$transform$1(ConfigBuilder.scala:167)
	at scala.Option.map(Option.scala:230)
	at org.apache.kyuubi.config.ConfigEntryWithDefault.readFrom(ConfigEntry.scala:106)
	at org.apache.kyuubi.config.KyuubiConf.get(KyuubiConf.scala:99)
	at org.apache.kyuubi.server.KyuubiServer.frontendServices$lzycompute(KyuubiServer.scala:150)
	at org.apache.kyuubi.server.KyuubiServer.frontendServices(KyuubiServer.scala:149)
	at org.apache.kyuubi.service.Serverable.initialize(Serverable.scala:45)
	at org.apache.kyuubi.server.KyuubiServer.initialize(KyuubiServer.scala:178)
	at org.apache.kyuubi.server.KyuubiServer$.startServer(KyuubiServer.scala:59)
	at org.apache.kyuubi.server.KyuubiServer$.main(KyuubiServer.scala:99)
	at org.apache.kyuubi.server.KyuubiServer.main(KyuubiServer.scala)
[...]

PR #5281, opened 2023-09-14
[KYUUBI #5212] Fix configuration errors causing by helm charts of prometheus services

FourSpaces:

This bug will affect k8s deployments.

https://github.com/apache/kyuubi/issues/5212

No

### DIFF
--- a/charts/kyuubi/templates/kyuubi-headless-service.yaml
+++ b/charts/kyuubi/templates/kyuubi-headless-service.yaml
@@ -25,10 +25,15 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    {{- range $name, $frontend := merge .Values.server .Values.monitoring }}
+    {{- range $name, $frontend := .Values.server }}
     - name: {{ $name | kebabcase }}
       port: {{ tpl $frontend.service.port $ }}
       targetPort: {{ $frontend.port }}
+    {{- end }}
+    {{- if .Values.monitoring.prometheus.enabled }}
+    - name: prometheus
+      port: {{ .Values.monitoring.prometheus.port}}
+      targetPort: {{ .Values.monitoring.prometheus.port}}
     {{- end }}
   selector:
     {{- include "kyuubi.selectorLabels" $ | nindent 4 }}
--- a/charts/kyuubi/templates/kyuubi-headless-service.yaml
+++ b/charts/kyuubi/templates/kyuubi-headless-service.yaml
@@ -32,8 +32,8 @@ spec:
     {{- end }}
     {{- if .Values.monitoring.prometheus.enabled }}
     - name: prometheus
-      port: {{ .Values.monitoring.prometheus.port}}
-      targetPort: {{ .Values.monitoring.prometheus.port}}
+      port: {{ .Values.monitoring.prometheus.port }}
+      targetPort: {{ .Values.monitoring.prometheus.port }}
     {{- end }}
   selector:
     {{- include "kyuubi.selectorLabels" $ | nindent 4 }}
--- a/charts/kyuubi/templates/kyuubi-headless-service.yaml
+++ b/charts/kyuubi/templates/kyuubi-headless-service.yaml
@@ -25,7 +25,7 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    {{- range $name, $frontend := .Values.server }}
+    {{- range $name, $frontend := merge .Values.server .Values.monitoring }}
     - name: {{ $name | kebabcase }}
       port: {{ tpl $frontend.service.port $ }}
       targetPort: {{ $frontend.port }}
--- a/charts/kyuubi/templates/kyuubi-service.yaml
+++ b/charts/kyuubi/templates/kyuubi-service.yaml
@@ -15,7 +15,7 @@
   limitations under the License.
 */}}
 
-{{- range $name, $frontend := merge .Values.server .Values.monitoring }}
+{{- range $name, $frontend := .Values.server }}
 {{- if $frontend.enabled }}
 apiVersion: v1
 kind: Service
--- a/charts/kyuubi/templates/kyuubi-service.yaml
+++ b/charts/kyuubi/templates/kyuubi-service.yaml
@@ -15,7 +15,7 @@
   limitations under the License.
 */}}
 
-{{- range $name, $frontend := .Values.server }}
+{{- range $name, $frontend := merge .Values.server .Values.monitoring }}
 {{- if $frontend.enabled }}
 apiVersion: v1
 kind: Service
--- a/charts/kyuubi/templates/kyuubi-statefulset.yaml
+++ b/charts/kyuubi/templates/kyuubi-statefulset.yaml
@@ -69,7 +69,7 @@ spec:
           envFrom: {{- tpl (toYaml .) $ | nindent 12 }}
           {{- end }}
           ports:
-            {{- range $name, $frontend := .Values.server }}
+            {{- range $name, $frontend := merge .Values.server .Values.monitoring }}
             {{- if $frontend.enabled }}
             - name: {{ $name | kebabcase }}
               containerPort: {{ $frontend.port }}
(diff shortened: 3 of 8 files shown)


charts/kyuubi/templates/kyuubi-headless-service.yaml
```diff
@@ -25,7 +25,7 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    {{- range $name, $frontend := .Values.server }}
+    {{- range $name, $frontend := merge .Values.server .Values.monitoring }}
```
dnskr, line 28: Similar to comment for `StatefulSet`, let's split rendering for server and monitoring ports.

dnskr (changes requested):
Thanks for the PR!
Could you please have a look into comments I left?

charts/kyuubi/templates/kyuubi-headless-service.yaml
FourSpaces: OK

charts/kyuubi/templates/kyuubi-statefulset.yaml
```diff
@@ -69,7 +69,7 @@ spec:
           envFrom: {{- tpl (toYaml .) $ | nindent 12 }}
           {{- end }}
           ports:
-            {{- range $name, $frontend := .Values.server }}
+            {{- range $name, $frontend := merge .Values.server .Values.monitoring }}
```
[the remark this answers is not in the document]
FourSpaces: OK

charts/kyuubi/templates/kyuubi-headless-service.yaml
FourSpaces: I will solve these problems tomorrow

charts/kyuubi/templates/kyuubi-service.yaml
```diff
@@ -15,7 +15,7 @@
   limitations under the License.
 */}}
 
-{{- range $name, $frontend := .Values.server }}
+{{- range $name, $frontend := merge .Values.server .Values.monitoring }}
```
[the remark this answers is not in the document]
FourSpaces: @dnskr  I don't think we need to create a separate Service resource instance.

FourSpaces:
@dnskr  Check if there are any issues.

charts/kyuubi/templates/kyuubi-headless-service.yaml
```diff
@@ -30,6 +30,11 @@ spec:
       port: {{ tpl $frontend.service.port $ }}
       targetPort: {{ $frontend.port }}
     {{- end }}
+    {{- if .Values.monitoring.prometheus.enabled }}
+    - name: prometheus
+      port: {{ .Values.monitoring.prometheus.port}}
+      targetPort: {{ .Values.monitoring.prometheus.port}}
```
dnskr, line 36: Please, add one space after `port`
```yaml
      port: {{ .Values.monitoring.prometheus.port }}
      targetPort: {{ .Values.monitoring.prometheus.port }}
```

charts/kyuubi/templates/kyuubi-service.yaml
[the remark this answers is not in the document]
dnskr: Let's do not add it here then, i.e. let's remove changes in the file.

charts/kyuubi/templates/kyuubi-statefulset.yaml
```diff
@@ -75,6 +75,10 @@ spec:
               containerPort: {{ $frontend.port }}
             {{- end }}
             {{- end }}
+            {{- if .Values.monitoring.prometheus.enabled }}
+            - name: prometheus
+              containerPort: {{ .Values.monitoring.prometheus.port}}
```
dnskr, line 80: Please, add one space after `port` here as well. I know that it's me who suggested changes without the space., sorry for that.
```yaml
containerPort: {{ .Values.monitoring.prometheus.port }}
```

dnskr (changes requested):

FourSpaces:
@dnskr  Check if there are any issues.

charts/kyuubi/values.yaml
```diff
@@ -116,13 +116,14 @@ server:
       nodePort: ~
       annotations: {}
 
+monitoring:
   # Exposes metrics in Prometheus format
   prometheus:
     enabled: true
     port: 10019
     service:
```
dnskr, line 124: Could you please check if the node `service` is not used anymore and deleted it?

dnskr (changes requested):

charts/kyuubi/values.yaml
dnskr, line 124: Could you please check if the node `service` is not used anymore and deleted it?

dnskr (changes requested):

charts/kyuubi/values.yaml
FourSpaces: @dnskr  I think "monitoring.prometheus.service" cannot be deleted because this node is used to control whether the prometheus service of kyuubi-headless is displayed.

dnskr: Only `monitoring.prometheus.enabled` and `monitoring.prometheus.port`  are used as I can see. [charts/kyuubi/templates/kyuubi-headless-service.yaml](https://github.com/apache/kyuubi/pull/5281/files#diff-a15690f05b8022769e74321d33a3f77de06a647df4f0d5c34aad721a9a92bb6a)
```yaml
    {{- if .Values.monitoring.prometheus.enabled }}
    - name: prometheus
      port: {{ .Values.monitoring.prometheus.port }}
      targetPort: {{ .Values.monitoring.prometheus.port }}
    {{- end }}
```

FourSpaces: @dnskr  After removing, how to control the port of Prometheus and whether to apply it? If the default parameters that cannot be changed are set, I think it is not a good solution.

FourSpaces: @dnskr  Are you still following this PR?

dnskr: Sorry for the delay, I missed your last comment.
I think that Prometheus metrics endpoint doesn't require dedicated `Service` and I don't remember I saw it in other Helm charts. So it's fine that you deleted it and we have `prometheus` port in Kyuubi headless service only.

In the current implementation `{{ .Values.monitoring.prometheus.port }}` value is used for `prometheus` port of the container and `Service` ports (port and targetPort), which is also fine. So, if none of `{{ .Values.monitoring.prometheus.service.* }}` values are used, I would suggest to delete them for now and add them later if needed.

FourSpaces: OK,

dnskr (approved):

dnskr:
@pan3793 Please, have a look at the PR. It should be included in 1.8.0 release, because the chart is shipped with the release distributive.
Also, what do you think about excluding the chart from Kyuubi release and creating dedicated release process for the chart?

pan3793:
> Also, what do you think about excluding the chart from Kyuubi release and creating dedicated release process for the chart?

@dnskr the idea looks cool, but I have some concerns about hadoop/hive configuration injecting for using the published helm chart to deploy.

let me share the internal practice from my side: even though we deploy Kyuubi/Spark on K8s, the Kyuubi server and Spark application still require to communicate with external HMS, HDFS, HBase, etc., thus bunches of `*-site.xml`(usually has hundreds even thousands lines per file) are required, it's not elegant to inline those XML into `values.yaml`, eventually， we choose to use [Accessing Files Inside Templates](https://helm.sh/docs/chart_template_guide/accessing_files/) as our solution, but as the docs said, it has some limitation, maybe for security purpose, which let us must deploy the helm application from helm chart source, instead of published helm chart from repo.

pan3793 (approved):

pan3793:
Thanks, merged to master/1.8

FourSpaces:
@pan3793 @dnskr  Does it mean I need to push this PR to the branch-1.8 branch? Or does it mean something else?

pan3793:
@FourSpaces I already cherry-picked this one to branch-1.8, it will be included in the 1.8.0

FourSpaces:
OK

PR #5305, opened 2023-09-18
Return dummy log when batch job before being added to session manager

zwangsheng:
…sion manager

As title

Current we open session first then hand this session with batch by SessionManager:
https://github.com/apache/kyuubi/blob/413683507079f8f97217e8099177f0b16bc9192d/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala#L97-L121

We may face `No local log `exception, when kyuubi server picked this batch job to open session but not handled by SessionManager, due to time gap and code gap.

- [X] [Run test](https://kyuubi.readthedocs.io/en/master/contributing/code/testing.html#running-tests) locally before make a pull request

No

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -415,6 +415,16 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
         } else if (fe.connectionUrl != metadata.kyuubiInstance) {
           val internalRestClient = getInternalRestClient(metadata.kyuubiInstance)
           internalRestClient.getBatchLocalLog(userName, batchId, from, size)
+        } else if (batchV2Enabled(metadata.requestConf) &&
+          // Aims to cover the case that batch job been picked by current instance
+          // but the batch id not been handled by this SessionManager.
+          // Ref to we open session first then hand the session
+          // and mark operation RUNNING also has code gap with handler session,
+          // we consider RUNNING state too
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -415,6 +415,16 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
         } else if (fe.connectionUrl != metadata.kyuubiInstance) {
           val internalRestClient = getInternalRestClient(metadata.kyuubiInstance)
           internalRestClient.getBatchLocalLog(userName, batchId, from, size)
+        } else if (batchV2Enabled(metadata.requestConf) &&
+          // Aims to cover the case that batch job been picked by current instance
+          // but the batch id not been handled by this SessionManager.
+          // Ref to we open session first then hand the session
+          // and mark operation RUNNING also has code gap with handler session,
+          // we consider RUNNING state too
```
pan3793, line 423:
```suggestion
          // in batch v2 impl, the operation state is changed from PENDING to RUNNING
          // before being added to SessionManager.
```

pan3793 (approved):

pan3793:
Thanks, merged to master/1.8

pan3793:
Unfortunately, UT failed after this change.
```
BatchesV2ResourceSuite:
- get local log internal redirection *** FAILED ***
  200 did not equal 404 (BatchesResourceSuite.scala:626)
```

zwangsheng:
Sure @pan3793 open #5307 to fix.

PR #5310, opened 2023-09-19
[BATCH] Batch session recovery should start after HTTP server getting started

zwangsheng:

In rare cases, when JettyServer is not fully started, we start the recovery task, will get the wrong port.

As a result, we cannot get the task that is not properly submitted by the current kyuubi server instance correctly.

Like :
```
mysql> select identifier, kyuubi_instance  from metadata where kyuubi_instance like '%:-1';
+--------------------------------------+----------------------+
| identifier                           | kyuubi_instance      |
+--------------------------------------+----------------------+
| b6e88262-fddb-3ccd-abdf-95ef206e612b | XXXX:-1              |
| 6a02b526-ed10-3b76-ba36-d61fbaf6b28d | XXXX:-1              |
| 70a0446e-d3d0-3b13-9d34-4cc90369c2d9 | XXXX:-1              |
| d290ee2e-b8cd-3bc4-b6c6-72f5b5dfcb9b | XXXX:-1              |
| 9bb10f4-24b4-3eec-b13b-7932b310cb5cd | XXXX:-1              |
```

In local env, start kyuubi server with this PR, kyuubi server start as excpeted.

And i remove start server and runInternal, kyuubi server will wait here.

No

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
@@ -189,7 +189,7 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
         // block until the HTTP server is started, otherwise, we may get
         // the wrong HTTP server port -1
         while (server.getState != "STARTED") {
-          debug(s"Wait for $getName server start")
+          info(s"Waiting for $getName's HTTP server getting started")
           Thread.sleep(1000)
         }
         recoverBatchSessions()
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
@@ -55,8 +55,6 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
   private[kyuubi] def sessionManager = be.sessionManager.asInstanceOf[KyuubiSessionManager]
 
   private val batchChecker = ThreadUtils.newDaemonSingleThreadScheduledExecutor("batch-checker")
-  private val batchRecoveryPicker =
-    ThreadUtils.newDaemonSingleThreadScheduledExecutor("batch-recovery-picker")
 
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -183,7 +191,14 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
     if (!isStarted.get) {
       try {
         server.start()
-        recoverBatchSessions()
```
pan3793, line 186: will move it to behind `startInternal()` solve the issue?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -183,10 +183,16 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
...
         isStarted.set(true)
         startBatchChecker()
         startInternal()
+        // if the server is not started, we should wait for it to start
+        // otherwise, we may get wrong server port with default -1
```
pan3793, line 190:
```suggestion
        // block until the HTTP server is started, otherwise, we may get
        // the wrong HTTP server port -1
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -183,10 +183,16 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
...
       try {
         server.start()
-        recoverBatchSessions()
         isStarted.set(true)
         startBatchChecker()
         startInternal()
+        // block until the HTTP server is started, otherwise, we may get
+        // the wrong HTTP server port -1
+        while (server.getState != "STARTED") {
+          debug(s"Wait for $getName server start")
```
pan3793, line 192:
```suggestion
          info(s"Waiting for $getName's HTTP server getting started")
```

pan3793 (approved):

pan3793:
Also cc @turboFei

pan3793:
Thanks, merged to master/1.8

PR #5313, opened 2023-09-19
[BATCH][V2] Marked V2 Metadata as Init instead of open session

zwangsheng:

When we try to recover all running tasks before the restart on a specified Kyuubi service, we encounter that the kyuubi server is booted in  Start -> OOM  -> Start vicious cycle when there are too many tasks.

Here we set all need recover task as `INITIALIZED`, let every kyuubi service have chance to pick and run those job, to avoid the problem of excessive single point of load

No

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
@@ -23,13 +23,10 @@ import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 import javax.servlet.DispatcherType
 import javax.ws.rs.WebApplicationException
 import javax.ws.rs.core.Response.Status
-
 import scala.collection.mutable.ListBuffer
-
 import com.google.common.annotations.VisibleForTesting
 import org.apache.hadoop.conf.Configuration
 import org.eclipse.jetty.servlet.{ErrorPageErrorHandler, FilterHolder}
-
 import org.apache.kyuubi.{KyuubiException, Utils}
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf._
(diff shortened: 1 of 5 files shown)


zwangsheng:
friendly ping @pan3793 @turboFei

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/BatchesResourceSuite.scala
```diff
@@ -576,23 +576,40 @@ abstract class BatchesResourceSuiteBase extends KyuubiFunSuite
...
+        eventually(timeout(10.seconds)) {
+          assert(session1.batchJobSubmissionOp.getStatus.state === OperationState.RUNNING ||
+            session1.batchJobSubmissionOp.getStatus.state === OperationState.FINISHED)
+          assert(session1.batchJobSubmissionOp.builder.processLaunched)
+
+          assert(session2.batchJobSubmissionOp.getStatus.state === OperationState.RUNNING ||
+            session2.batchJobSubmissionOp.getStatus.state === OperationState.FINISHED)
+          assert(!session2.batchJobSubmissionOp.builder.processLaunched)
+        }
+      case "2" =>
```
pan3793, line 599: the check for 2 seems much lenient than 1, can we make it strict?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -140,13 +144,37 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
...
+  }
+
   @VisibleForTesting
   private[kyuubi] def recoverBatchSessions(): Unit = {
     val recoveryNumThreads = conf.get(METADATA_RECOVERY_THREADS)
     val batchRecoveryExecutor =
       ThreadUtils.newDaemonFixedThreadPool(recoveryNumThreads, "batch-recovery-executor")
     try {
-      val batchSessionsToRecover = sessionManager.getBatchSessionsToRecover(connectionUrl)
+      // split metadata wait to recover as v1 and v2
```
pan3793, line 158: the code is clear enough to show **what we are doing**, the comment should explain "why should we do it"

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -140,13 +144,37 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
     batchChecker.scheduleWithFixedDelay(task, interval, interval, TimeUnit.MILLISECONDS)
   }
 
+  private def batchV2Enabled(reqConf: Map[String, String]): Boolean = {
+    KyuubiServer.kyuubiServer.getConf.get(BATCH_SUBMITTER_ENABLED) &&
+    reqConf.getOrElse(BATCH_IMPL_VERSION.key, getConf.get(BATCH_IMPL_VERSION)) == "2"
+  }
```
pan3793, line 150: copied from another place? can we make the method sharable?

PR #5321, opened 2023-09-22
[WIP][EXTENSION] Inject Rule to fast fail gluten app if there are too much un-support operator

zwangsheng:

Fast fail if there are too much un-support operators in spark plan when enabled gluten.

Too much un-support operators when enabled gluten may significantly reduce the performance, so we want to notify user by fast fail this application.

If the user can tolerate this kind of performance penalty, you can increase the value of the `spark.sql.gluten.fallbackOperatorThreshold`.

See more gluten un-support operators in [Velox Backend Support Progress](https://github.com/oap-project/gluten/blob/main/docs/velox-backend-support-progress.md)

Note:
> Gluten-Related extension is experimental and under rapid development, this configuration is added to allow user to control extension, not intended exposing to end users, it may be removed in anytime.

No

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/GlutenPlanManager.scala
+++ b/extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/GlutenPlanManager.scala
@@ -21,16 +21,23 @@ import org.apache.spark.SparkConf
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.expressions.UserDefinedExpression
 import org.apache.spark.sql.catalyst.rules.Rule
-import org.apache.spark.sql.execution.{ColumnarRule, FileSourceScanExec, SparkPlan}
+import org.apache.spark.sql.execution._
+import org.apache.spark.sql.execution.aggregate.{ObjectHashAggregateExec, SortAggregateExec}
+import org.apache.spark.sql.execution.columnar.InMemoryTableScanExec
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
+import org.apache.spark.sql.execution.exchange.ShuffleExchangeExec
+import org.apache.spark.sql.execution.joins.{BroadcastNestedLoopJoinExec, CartesianProductExec}
 
 /**
+ * Kyuubi extension for gluten enabled case.
+ * 1. check whether the plan contains too much unsupported operator
  */
 case class GlutenPlanManager(session: SparkSession) extends ColumnarRule {
   private val GLUTEN_DRIVER_PLUGIN_CLASS = "io.glutenproject.GlutenDriverPlugin"
 
   override def preColumnarTransitions: Rule[SparkPlan] =
-    if (glutenEnabled(session.sparkContext.getConf)) {
+    if (sys.props.get("spark.testing").nonEmpty ||
+      glutenEnabled(session.sparkContext.getConf)) {
       GlutenPlanAnalysis
     } else {
       (plan: SparkPlan) => plan
@@ -43,15 +50,6 @@ case class GlutenPlanManager(session: SparkSession) extends ColumnarRule {
 }
 
 object GlutenPlanAnalysis extends Rule[SparkPlan] {
-  private val unSupportExpressions = Seq(
-    "CaseWhen",
-    "BitAndAgg",
-    "BitOrAgg",
-    "BitXorAgg",
-    "Base64",
-    "Bin",
-    "BitLength",
-    "Contains")
 
   override def apply(plan: SparkPlan): SparkPlan = {
     var count = 0
@@ -59,9 +57,9 @@ object GlutenPlanAnalysis extends Rule[SparkPlan] {
     plan foreach { p =>
       if (containsUnsupportedOperator(p)) {
         count += 1
-        check(count)
       }
     }
+    check(count)
     plan
   }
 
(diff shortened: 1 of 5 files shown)


zwangsheng:
cc @ulysses-you

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/GlutenPlanManager.scala
```diff
@@ -0,0 +1,114 @@
...
+        true
+      case p: SparkPlan
+          if p.expressions.exists(e =>
+            e.exists {
+              case _: UserDefinedExpression |
+                  _: UnaryMinus |
+                  _: Bin |
+                  _: Contains |
+                  _: StartsWith |
+                  _: EndsWith |
```
ulysses-you, line 77: it seems `Contains`, `StartsWith`, `EndsWith ` are supported

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -273,4 +273,15 @@ object KyuubiSQLConf {
...
+  val GLUTEN_FALLBACK_OPERATOR_THRESHOLD =
+    buildConf("spark.sql.gluten.fallbackOperatorThreshold")
+      .doc("The threshold of fallback operator. If the number of fallback operator is " +
+        "bigger than this threshold, we will fallback to original plan." +
+        "Note: Gluten-Related extension is experimental and under rapid development," +
+        "this configuration is added to allow user to control extension, not " +
+        "intended exposing to end users, it may be removed in anytime.")
+      .version("1.8.0")
+      .intConf
+      .createWithDefault(5)
```
ulysses-you, line 286: let's use `Int.MaxValue` to avoid breaking something

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/GlutenPlanManager.scala
```diff
@@ -0,0 +1,114 @@
...
+    conf.get("spark.gluten.enabled", "true") == "true"
+  }
+}
+
+object GlutenPlanAnalysis extends Rule[SparkPlan] {
+
+  override def apply(plan: SparkPlan): SparkPlan = {
+    val count = plan.collect {
+      case p: FileSourceScanExec
+          if !p.relation.fileFormat.isInstanceOf[ParquetFileFormat] =>
```
ulysses-you, line 56: skip orc file format

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/GlutenPlanManager.scala
```diff
@@ -0,0 +1,114 @@
...
+          _: ObjectHashAggregateExec |
+          _: SortAggregateExec |
+          _: CoalesceExec |
+          _: GenerateExec |
+          _: RangeExec |
+          _: SampleExec |
+          _: BroadcastNestedLoopJoinExec =>
+        true
+      case p: SparkPlan
+          if p.expressions.exists(e =>
```
ulysses-you, line 70:
```scala
          if p.expressions.exists(_.exists {
            ...
          })
```

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/GlutenPlanManager.scala
zwangsheng: Thanks, changed this list according to black list in 
https://github.com/oap-project/gluten/blob/ad532a7b34cef1deeb6f3ae10f3ecc741c8875b9/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc#L38C1-L55

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -284,4 +284,34 @@ object KyuubiSQLConf {
       .version("1.8.0")
       .intConf
       .createWithDefault(Int.MaxValue)
+
+  val GLUTEN_NON_SUPPORT_OPERATOR_LIST =
+    buildConf("spark.sql.gluten.nonSupportOperatorList")
```
pan3793, line 289: how about naming `spark.sql.gluten.fastFallbackOperators`? and as an open source projects, we'd better make the default value match a released version instead of the current main branch of Gluten

extensions/spark/kyuubi-extension-spark-3-3/src/main/scala/org/apache/kyuubi/sql/GlutenPlanManager.scala
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+  private def glutenEnabled(conf: SparkConf): Boolean = {
+    conf.get("spark.plugins", "").contains(GLUTEN_DRIVER_PLUGIN_CLASS) &&
+    conf.get("spark.gluten.enabled", "true") == "true"
+  }
+}
+
+object GlutenPlanAnalysis extends Rule[SparkPlan] {
+  private val nonSupportedOperatorList = conf.getConf(GLUTEN_NON_SUPPORT_OPERATOR_LIST)
```
ulysses-you, line 55: please move this inside `apply`, so we can change this config at runtime

extensions/spark/kyuubi-extension-spark-common/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
ulysses-you: +1, let's remove the default value and mark it as optional.

zwangsheng:
thanks @ulysses-you @pan3793 , make some changes:
1. Rename fallback operator config key
2. make fallback operator as optional
3. make fall back operator can be changed in runtime case(Get this operators each call apply method)

PR #5329, opened 2023-09-25
[KYUUBI #5327][BATCH] Add priority field in metadata table definition

### LINKED ISSUE #5327: [Subtask][V2][DDL] Introduce priority field in metadata table.
### Code of Conduct

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues) and found no similar issues.

### Describe the subtask

Introduce priority field in metadata table.

### Parent issue

#5326

zwangsheng:

Add `priority` field in `metadata` table definition

In the current batch v2 implementation, Kyuubi supports the FIFO schedule strategy, we are planning the enhance it with `priority` support, those with a higher integer value of `priority` have a better opportunity to be scheduled.

It refers priority definition in [Apache Hadoop YARN](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html#Setup_for_application_priority) and [Apache YuniKorn](https://yunikorn.apache.org/docs/user_guide/priorities/#application-priority) to define the `priority` as:
1. use a signed 4 bits integer to represent priority with the default value of 10
2. a higher integer value of `priority` has a better opportunity to be scheduled.

In practice, we face some issues when use MySQL as metastore:

MySQL 5.7 could not benefit from index in query mixes `ASC` and `DESC`, for example: `SELECT * FROM metadata ORDER BY priority DESC, create_time ASC`, which is mentioned in [MySQL ORDER BY optimization](https://dev.mysql.com/doc/refman/5.7/en/order-by-optimization.html)
> In some cases, MySQL cannot use indexes to resolve the `ORDER BY`, although it may still use indexes to find the rows that match the `WHERE` clause. Examples:
> - ...
> - The query mixes `ASC` and `DESC`
> - ...

Fortunately, it was enhanced in MySQL 8.0
> Two columns in an ORDER BY can sort in the same direction (both ASC, or both DESC) or in opposite directions (one ASC, one DESC). A condition for index use is that the index must have the same homogeneity, but need not have the same actual direction.
> If a query mixes ASC and DESC, the optimizer can use an index on the columns if the index also uses corresponding mixed ascending and descending columns:

Thus, we create a mixed `ORDER BY` index `ALTER TABLE metadata ADD INDEX priority_create_time_index(priority DESC, create_time ASC)` to improve the query efficiency and avoid slowness in the case of large data volumes.

According to [MySQL 5.7 Index](https://dev.mysql.com/doc/refman/5.7/en/create-index.html), this won't cause SQL error when creating this index in MySQL 5.7 but not take effect.

> A key_part specification can end with ASC or DESC. These keywords are permitted for future extensions for specifying ascending or descending index value storage. Currently, they are parsed but ignored; index values are always stored in ascending order.

Close #5327

Test MySQL & Derby with DBeaver.

MySQL:
截屏2023-09-25 19 33 51
截屏2023-09-25 19 33 57

Derby:
截屏2023-09-25 19 33 34
截屏2023-09-25 19 33 38

No


pan3793:
Please enrich the PR description:

- How do resource managers like YARN, Yunikorn define priority?
- Why MySQL 5.7 does not work well, but 8.0 does?

kyuubi-server/src/main/resources/sql/mysql/005-KYUUBI-5327.mysql.sql
```diff
@@ -0,0 +1,8 @@
+SELECT '< KYUUBI-5131: Create index on metastore.create_time' AS ' ';
```
pan3793, line 1: what's this

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-1.8.0.mysql.sql
```diff
@@ -14,7 +14,7 @@ CREATE TABLE IF NOT EXISTS metadata(
     request_name varchar(1024) COMMENT 'the request name',
     request_conf mediumtext COMMENT 'the request config map',
     request_args mediumtext COMMENT 'the request arguments',
-    create_time BIGINT NOT NULL COMMENT 'the metadata create time',
+    create_time bigint NOT NULL COMMENT 'the metadata create time',
```
pan3793, line 17: revert unnecessary change

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-1.8.0.mysql.sql
```diff
@@ -24,9 +24,14 @@ CREATE TABLE IF NOT EXISTS metadata(
     engine_state varchar(32) COMMENT 'the engine application state',
     engine_error mediumtext COMMENT 'the engine application diagnose',
     end_time bigint COMMENT 'the metadata end time',
+    priority int DEFAULT 10  COMMENT 'the application priority, high value means high priority',
```
pan3793, line 27: nit: an extra space

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-1.8.0.mysql.sql
```diff
@@ -24,9 +24,14 @@ CREATE TABLE IF NOT EXISTS metadata(
...
     peer_instance_closed boolean default '0' COMMENT 'closed by peer kyuubi instance',
     UNIQUE INDEX unique_identifier_index(identifier),
     INDEX user_name_index(user_name),
     INDEX engine_type_index(engine_type),
-    INDEX create_time_index(create_time)
+    INDEX create_time_index(create_time),
+    -- mysql 5.7 does not support mix order index
+    -- mysql 8 allow create mix order index and can be used in order by
+    -- here create it in 5.7 won't cause error
```
pan3793, line 35: add something like official docs

zwangsheng:
Addressed above.

kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-1.8.0.mysql.sql
```diff
@@ -24,9 +24,12 @@ CREATE TABLE IF NOT EXISTS metadata(
     engine_state varchar(32) COMMENT 'the engine application state',
     engine_error mediumtext COMMENT 'the engine application diagnose',
     end_time bigint COMMENT 'the metadata end time',
+    priority int DEFAULT 10 COMMENT 'the application priority, high value means high priority',
```
pan3793, line 27: is restriction `NOT NULL` applicable here?

kyuubi-server/src/main/resources/sql/sqlite/metadata-store-schema-1.8.0.sqlite.sql
```diff
@@ -24,6 +24,7 @@ CREATE TABLE IF NOT EXISTS metadata(
     engine_state varchar(32), -- the engine application state
     engine_error mediumtext, -- the engine application diagnose
     end_time bigint, -- the metadata end time
+    priority INTEGER DEFAULT 10, -- the application priority, high value means high priority
```
pan3793, line 27: ditto

kyuubi-server/src/main/resources/sql/mysql/005-KYUUBI-5327.mysql.sql
```diff
@@ -0,0 +1,13 @@
+SELECT '< KYUUBI-5327: Introduce priority in metadata' AS ' ';
+
+ALTER TABLE metadata ADD COLUMN priority int default 10 COMMENT 'the application priority, high value means high priority';
+
+-- In mysql 5.7, A key_part specification can end with ASC or DESC.
```
pan3793, line 5: use the correct terminology "MySQL" if any

kyuubi-server/src/main/resources/sql/derby/metadata-store-schema-1.8.0.derby.sql
```diff
@@ -26,7 +26,7 @@ CREATE TABLE metadata(
     engine_state varchar(32), -- the engine application state
     engine_error clob, -- the engine application diagnose
     end_time bigint,  -- the metadata end time
-    priority int DEFAULT 10, -- the application priority, high value means high priority
+    priority int DEFAULT 10 NOT NULL, -- the application priority, high value means high priority
```
pan3793, line 29: put `DEFAULT 10` at the end

zwangsheng: Addressed

pan3793 (approved):

pan3793:
Thanks, merged to master/1.8

PR #5427, opened 2023-10-16
[AUTHZ] Shade spark authz plugin

Yikf:

This PR aims to shade the kyuubi spark authz plugin to simplify the user's use.

No

### DIFF
--- a/docs/security/authorization/spark/build.md
+++ b/docs/security/authorization/spark/build.md
@@ -33,7 +33,7 @@ After a while, if everything goes well, you will get the plugin finally in two p
 
 ## Build bundle jar with Apache Maven
 
-Apache Kyuubi provides the bundle jar for the Spark authZ plugin, You can run the authZ plugin using just a bundle jar without the additional dependency of jars,
+Apache Kyuubi also provides the bundle jar for the Spark AuthZ plugin, You can run the AuthZ plugin using just a bundle jar without the additional dependency of jars,
 To build it, `cd` to the root direct of kyuubi project and run:
 
 ```shell
--- a/docs/security/authorization/spark/build.md
+++ b/docs/security/authorization/spark/build.md
@@ -31,18 +31,18 @@ After a while, if everything goes well, you will get the plugin finally in two p
 - The main plugin jar, which is under `./extensions/spark/kyuubi-spark-authz/target/kyuubi-spark-authz_${scala.binary.version}-${project.version}.jar`
 - The least transitive dependencies needed, which are under `./extensions/spark/kyuubi-spark-authz/target/scala-${scala.binary.version}/jars`
 
-## Build bundle jar with Apache Maven
+## Build shaded jar with Apache Maven
 
-Apache Kyuubi also provides the bundle jar for the Spark AuthZ plugin, You can run the AuthZ plugin using just a bundle jar without the additional dependency of jars,
+Apache Kyuubi also provides the shaded jar for the Spark AuthZ plugin, You can run the AuthZ plugin using just a shaded jar without the additional dependency of jars,
 To build it, `cd` to the root direct of kyuubi project and run:
 
 ```shell
-build/mvn clean package -pl :kyuubi-spark-authz-bundle_2.12 -DskipTests -am
+build/mvn clean package -pl :kyuubi-spark-authz-shaded_2.12 -DskipTests -am
 ```
 
 After a while, if everything goes well, you will get the plugin finally:
 
-- The bundle authZ plugin jar, which is under `./extensions/spark/kyuubi-spark-authz-bundle/target/kyuubi-spark-authz_${scala.binary.version}-bundle-${project.version}.jar`
+- The shaded AuthZ plugin jar, which is under `./extensions/spark/kyuubi-spark-authz-shaded/target/kyuubi-spark-authz-shaded_${scala.binary.version}-${project.version}.jar`
 
 ### Build against Different Apache Spark Versions
 
--- a/docs/security/authorization/spark/build.md
+++ b/docs/security/authorization/spark/build.md
@@ -31,6 +31,19 @@ After a while, if everything goes well, you will get the plugin finally in two p
 - The main plugin jar, which is under `./extensions/spark/kyuubi-spark-authz/target/kyuubi-spark-authz_${scala.binary.version}-${project.version}.jar`
 - The least transitive dependencies needed, which are under `./extensions/spark/kyuubi-spark-authz/target/scala-${scala.binary.version}/jars`
 
+## Build bundle jar with Apache Maven
+
+Apache Kyuubi provides the bundle jar for the Spark authZ plugin, You can run the authZ plugin using just a bundle jar without the additional dependency of jars,
+To build it, `cd` to the root direct of kyuubi project and run:
+
+```shell
+build/mvn clean package -pl :kyuubi-spark-authz-bundle_2.12 -DskipTests -am
+```
+
+After a while, if everything goes well, you will get the plugin finally:
+
+- The bundle authZ plugin jar, which is under `./extensions/spark/kyuubi-spark-authz-bundle/target/kyuubi-spark-authz_${scala.binary.version}-bundle-${project.version}.jar`
+
 ### Build against Different Apache Spark Versions
 
 The maven option `spark.version` is used for specifying Spark version to compile with and generate corresponding transitive dependencies.
--- a/docs/security/authorization/spark/install.md
+++ b/docs/security/authorization/spark/install.md
@@ -31,7 +31,7 @@
 
 ## Install
 
-With the `kyuubi-spark-authz_*.jar` and its transitive dependencies available or use only `kyuubi-spark-authz-bundle_*.jar` for spark runtime classpath, such as
+Use either the bundle jar `kyuubi-spark-authz-bundle_*.jar` or the `kyuubi-spark-authz_*.jar` with its transitive dependencies available for spark runtime classpath, such as
 - Copied to `$SPARK_HOME/jars`, or
 - Specified to `spark.jars` configuration
 
--- a/docs/security/authorization/spark/install.md
+++ b/docs/security/authorization/spark/install.md
@@ -31,7 +31,7 @@
 
 ## Install
 
-With the `kyuubi-spark-authz_*.jar` and its transitive dependencies available for spark runtime classpath, such as
+With the `kyuubi-spark-authz_*.jar` and its transitive dependencies available or use only `kyuubi-spark-authz-bundle_*.jar` for spark runtime classpath, such as
 - Copied to `$SPARK_HOME/jars`, or
 - Specified to `spark.jars` configuration
 
--- a/docs/security/authorization/spark/install.md
+++ b/docs/security/authorization/spark/install.md
@@ -31,7 +31,7 @@
 
 ## Install
 
-Use either the bundle jar `kyuubi-spark-authz-bundle_*.jar` or the `kyuubi-spark-authz_*.jar` with its transitive dependencies available for spark runtime classpath, such as
+Use either the shaded jar `kyuubi-spark-authz-shaded_*.jar` or the `kyuubi-spark-authz_*.jar` with its transitive dependencies available for spark runtime classpath, such as
 - Copied to `$SPARK_HOME/jars`, or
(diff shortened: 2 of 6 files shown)


Yikf:
Since kyuubi supports different spark versions and ranger versions, and the cost of integration testing is too high, we initially use the default ranger version for functional testing.

In this PR, I test manually with ranger 2.4.0 and spark 3.4.0;

case 1: user root does not have select privilege on default/t1/id
image
image

case 2: user root have select privilege on default/t1/id
image

image

Yikf:
cc @pan3793

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -331,6 +331,60 @@
...
+                <artifactId>maven-shade-plugin</artifactId>
+                <configuration>
+                    <shadedArtifactAttached>false</shadedArtifactAttached>
+                    <artifactSet>
+                        <includes>
+                            <include>org.apache.kyuubi:kyuubi-util-scala_${scala.binary.version}</include>
+                            <include>org.apache.kyuubi:kyuubi-util</include>
+                            <include>org.apache.ranger:ranger-plugins-common</include>
+                            <include>org.apache.ranger:ranger-plugins-audit</include>
+                            <include>org.codehaus.jackson:jackson-jaxrs</include>
```
pan3793, line 347: it's not a good practice to create a shaded jar that does not perform relocation

pan3793: have you checked the output shaded jar, to make sure this is no class leak?

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -331,6 +331,78 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <configuration>
+                    <shadedArtifactAttached>false</shadedArtifactAttached>
+                    <artifactSet>
+                        <includes>
+                            <include>org.apache.kyuubi:kyuubi-util-scala_${scala.binary.version}</include>
+                            <include>org.apache.kyuubi:kyuubi-util</include>
+                            <include>org.apache.ranger:ranger-plugins-common</include>
+                            <include>org.apache.ranger:ranger-plugins-audit</include>
```
yaooqinn, line 346: What‘s the steps for users to enable solr audit after this patch

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -331,6 +331,78 @@
...
+                                <exclude>NOTICE.txt</exclude>
+                                <exclude>mozilla/**</exclude>
+                                <exclude>**/module-info.class</exclude>
+                            </excludes>
+                        </filter>
+                    </filters>
+                    <relocations>
+                        <relocation>
+                            <pattern>org.apache.ranger</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.org.apache.ranger</shadedPattern>
```
yaooqinn, line 377: Will this affect the ranger configuration to pick up the right ranger rest client?

extensions/spark/kyuubi-spark-authz/pom.xml
Yikf: I decompiled check shade jar and expected that class existed. Besides, I tested shade jar manually and it worked as expected.

extensions/spark/kyuubi-spark-authz/pom.xml
Yikf: Yes, it does, `org.apache.ranger` may not need relocation, What do you think?

Yikf:
In addition, I want to change the jar package name of shade so as not to affect the use of existing users, so two jars will be produced:

1. kyuubi-spark-authz_${scala.binary.version}-${project.version}.jar is original jar
2. kyuubi-spark-authz_${scala.binary.version}-${project.version} is a fat jar

Any thought? @yaooqinn @pan3793

yaooqinn:
I suggest using a separate module for bundling， naming it kyuubi-spark-authz_xx-bundle.jar is fair enough?

Also the doc needs to be updated if we provide a new installation approach

Yikf:
> Also the doc needs to be updated if we provide a new installation approach

Thank you for mentioning it. I will.


> I suggest using a separate module for bundling， naming it kyuubi-spark-authz_xx-bundle.jar is fair enough?

Just a personal question, what are the benefits of this?

extensions/spark/kyuubi-spark-authz/pom.xml
Yikf: Now, it only leaks `jna` related classes, and since jna contains so, it seems that it cannot be relocated.

Also, ranger-related classes do not require relocation, which does not change how the user is released.

I think this type of package has almost no class conflict issues, so it seems that this is ok too.

docs/security/authorization/spark/install.md
```diff
@@ -31,7 +31,7 @@
 
 ## Install
 
-With the `kyuubi-spark-authz_*.jar` and its transitive dependencies available for spark runtime classpath, such as
+With the `kyuubi-spark-authz_*.jar` and its transitive dependencies available or use only `kyuubi-spark-authz-bundle_*.jar` for spark runtime classpath, such as
```
bowenliang123, line 34:
```suggestion
Use either the bundle jar `kyuubi-spark-authz-bundle_*.jar` or the `kyuubi-spark-authz_*.jar` with its transitive dependencies available for spark runtime classpath, such as
```

docs/security/authorization/spark/build.md
```diff
@@ -31,6 +31,19 @@ After a while, if everything goes well, you will get the plugin finally in two p
 - The main plugin jar, which is under `./extensions/spark/kyuubi-spark-authz/target/kyuubi-spark-authz_${scala.binary.version}-${project.version}.jar`
 - The least transitive dependencies needed, which are under `./extensions/spark/kyuubi-spark-authz/target/scala-${scala.binary.version}/jars`
 
+## Build bundle jar with Apache Maven
+
+Apache Kyuubi provides the bundle jar for the Spark authZ plugin, You can run the authZ plugin using just a bundle jar without the additional dependency of jars,
```
bowenliang123, line 36:
```suggestion
Apache Kyuubi also provides the bundle jar for the Spark AuthZ plugin, You can run the AuthZ plugin using just a bundle jar without the additional dependency of jars,
```

yaooqinn (approved):

pom.xml
```diff
@@ -56,6 +56,7 @@
         <module>extensions/server/kyuubi-server-plugin</module>
         <module>extensions/spark/kyuubi-extension-spark-jdbc-dialect</module>
         <module>extensions/spark/kyuubi-spark-authz</module>
+        <module>extensions/spark/kyuubi-spark-authz-bundle</module>
```
pan3793, line 59: can we use the term "shaded" instead of "bundle", as we already have `kyuubi-hive-jdbc-shaded`

bowenliang123 (approved):

docs/security/authorization/spark/build.md
```diff
@@ -31,6 +31,19 @@ After a while, if everything goes well, you will get the plugin finally in two p
...
+Apache Kyuubi also provides the bundle jar for the Spark AuthZ plugin, You can run the AuthZ plugin using just a bundle jar without the additional dependency of jars,
+To build it, `cd` to the root direct of kyuubi project and run:
+
+```shell
+build/mvn clean package -pl :kyuubi-spark-authz-bundle_2.12 -DskipTests -am
+```
+
+After a while, if everything goes well, you will get the plugin finally:
+
+- The bundle authZ plugin jar, which is under `./extensions/spark/kyuubi-spark-authz-bundle/target/kyuubi-spark-authz_${scala.binary.version}-bundle-${project.version}.jar`
```
pan3793, line 45: In such case, the name pattern should be 
```
kyuubi-spark-authz-bundle_${scala.binary.version}-${project.version}.jar
```

Also, does the plugin build against one Spark version could be used for all Spark runtimes? If not, should spark binary version be added as part of jar name?

pom.xml
Yikf: Both are fine for me, and I prefer the `bundle` as it is a external client-side jar.

docs/security/authorization/spark/build.md
Yikf: > In such case, the name pattern should be

Good catch!



> Also, does the plugin build against one Spark version could be used for all Spark runtimes? If not, should spark binary version be added as part of jar name?

I think so, Could you please double check : ) @bowenliang123

extensions/spark/kyuubi-spark-authz-bundle/pom.xml
```diff
@@ -0,0 +1,317 @@
...
+                        <relocation>
+                            <pattern>com.sun.ws.rs.ext</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.com.sun.ws.rs.ext</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>com.kstruct.gethostname4j</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.com.kstruct.gethostname4j</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>org.apache.hadoop.security</pattern>
```
pan3793, line 293: does it mean the shaded jar contains hadoop classes? it's crazy

pom.xml
pan3793: `kyuubi-hive-jdbc-shaded` does too, I prefer use `shaded` for consistent

extensions/spark/kyuubi-spark-authz-bundle/pom.xml
Yikf: No, this just contains the following three classes, which are in ranger-plugins-common;

image


image

pom.xml
Yikf: Ok

extensions/spark/kyuubi-spark-authz-bundle/pom.xml
pan3793: oh, it comes from ranger, that's fine

extensions/spark/kyuubi-spark-authz-shaded/pom.xml
```diff
@@ -0,0 +1,317 @@
...
+
+    <artifactId>kyuubi-spark-authz-shaded_${scala.binary.version}</artifactId>
+    <packaging>jar</packaging>
+    <name>Kyuubi Dev Spark Authorization Extension Shaded</name>
+    <url>https://kyuubi.apache.org/</url>
+
+    <properties>
+        <!-- the following components' version may need to tune to align w/ the ranger.version-->
+        <gethostname4j.version>1.0.0</gethostname4j.version>
+        <jersey.client.version>1.19.4</jersey.client.version>
```
cfmcgrady, line 36: the `jersey-client` is a transitive dependency from the ranger client, shall we need to specify the `jersey-client` version when shading the authz plugin?

Yikf: emm, i can't find this transitive dependency

Yikf:
Hi all, Any other comments? : )

Yikf:
Thanks all , merged to master.

extensions/spark/kyuubi-spark-authz-shaded/pom.xml
```diff
@@ -0,0 +1,317 @@
...
+                <configuration>
+                    <shadedArtifactAttached>false</shadedArtifactAttached>
+                    <artifactSet>
+                        <includes>
+                            <include>org.apache.kyuubi:kyuubi-util-scala_${scala.binary.version}</include>
+                            <include>org.apache.kyuubi:kyuubi-spark-authz_${scala.binary.version}</include>
+                            <include>org.apache.kyuubi:kyuubi-util</include>
+                            <include>org.apache.ranger:ranger-plugins-common</include>
+                            <include>org.apache.ranger:ranger-plugins-audit</include>
+                            <include>org.codehaus.jackson:jackson-jaxrs</include>
```
pan3793, line 248: it's may not sufficient, please include the transitive jackson libs (spark is going to remove the jackson 1.x deps from the release tarball)

PR #5496, opened 2023-10-23
[AUTHZ] Authz shaded include the transitive jackson libs

Yikf:

As description https://github.com/apache/kyuubi/pull/5427#discussion_r1367907906. This PR aims to:

- Authz shaded include the transitive jackson libs
- Add LICENSE and NOTICE file to exactly match the content of jar
- Add scala-2.13 profile to make authz shade module support scala 2.13

No

### DIFF
--- a/extensions/spark/kyuubi-spark-authz-shaded/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz-shaded/pom.xml
@@ -94,14 +94,6 @@
                     <artifactId>jsr311-api</artifactId>
                 </exclusion>
                 <exclusion>
-                    <groupId>org.codehaus.jackson</groupId>
-                    <artifactId>jackson-core-asl</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.codehaus.jackson</groupId>
-                    <artifactId>jackson-mapper-asl</artifactId>
-                </exclusion>
-                <exclusion>
                     <groupId>com.kstruct</groupId>
                     <artifactId>gethostname4j</artifactId>
                 </exclusion>
@@ -246,6 +238,8 @@
                             <include>org.apache.ranger:ranger-plugins-common</include>
                             <include>org.apache.ranger:ranger-plugins-audit</include>
                             <include>org.codehaus.jackson:jackson-jaxrs</include>
+                            <include>org.codehaus.jackson:jackson-core-asl</include>
+                            <include>org.codehaus.jackson:jackson-mapper-asl</include>
                             <include>com.sun.jersey:jersey-client</include>
                             <include>com.sun.jersey:jersey-core</include>
                             <include>com.kstruct:gethostname4j</include>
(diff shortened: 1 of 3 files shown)


Yikf:
@pan3793

pan3793:
```
.
├── META-INF
│   ├── AL2.0
│   ├── ASL2.0
│   ├── LGPL2.1
│   ├── LICENSE
│   ├── MANIFEST.MF
│   ├── NOTICE
│   ├── jersey-module-version
│   ├── persistence.xml
...
```

I did a quick check on the shaded jar, and found the following issues:

- where does LGPL2.1 come from? LGPL2.1 is under [category-x](https://www.apache.org/legal/resolved.html#category-x) which is not allowed to be included into ASF artifacts.
- AL2.0, ASL2.0 should be excluded
- LICENSE and NOTICE should be modified to exactly match the content of jar

Yikf:
> * where does LGPL2.1 come from? LGPL2.1 is under [category-x](https://www.apache.org/legal/resolved.html#category-x) which is not allowed to be included into ASF artifacts.

It comes from `net.java.dev.jna:jna` and `net.java.dev.jna:jna-platform`, maybe we can't shade these two jars?

> * LICENSE and NOTICE should be modified to exactly match the content of jar

Sure, BTW, Is there a tool to generate LICENSE and NOTICE？

pan3793:
BTW, jna seems to be the dep of scala 2.13

Yikf:
> * where does LGPL2.1 come from? LGPL2.1 is under [category-x](https://www.apache.org/legal/resolved.html#category-x) which is not allowed to be included into ASF artifacts.

LGPL2.1 comes from `net.java.dev.jna`, and it [is licensed under the Apache License](https://github.com/java-native-access/jna#license), excluded LGPL2.1 from shaded jar looks like fine.


> BTW, jna seems to be the dep of scala 2.13

jna looks like come from ranger

pan3793:
> BTW, jna seems to be the dep of scala 2.13

I mean https://github.com/apache/spark/pull/43008#discussion_r1333838783

Yikf:
> I mean [apache/spark#43008 (comment)](https://github.com/apache/spark/pull/43008#discussion_r1333838783)

Thanks for the information. It looks like we need to provide the scala-2.13 profile, which requires exclude jna.

For the scala 2.13 profile, maybe the jna version of ranger is different from the jna version of scala 2.13, but I can't think of a good way to solve it, because jna cannot shade.

pan3793:
@Yikf I suppose we can use the jna provided in Spark, then let's excluding jna in scala-2.13 profile

extensions/spark/kyuubi-spark-authz-shaded/pom.xml
```diff
@@ -314,4 +311,27 @@
...
+        <profile>
+            <id>scala-2.13</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-shade-plugin</artifactId>
+                        <configuration>
+                            <artifactSet>
+                                <excludes>
```
pan3793, line 326: have you tested it? what's the behavior when both `includes` and `excludes` present?

Yikf: It will contain the content in the includes tag and not in the excludes tag.
image

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #5497, opened 2023-10-23
[AuthZ] Simplify debug message for missing field/method in ReflectUtils

yaooqinn:

It's easy for developers to check a member or method from a Java class using a code viewer or online Java doc.

The current debug msg is kinda noisy for them to locate the key information they want, which is 
Java class and field name. It makes our debug log unreadable.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/resources/log4j2-test.xml
+++ b/extensions/spark/kyuubi-spark-authz/src/test/resources/log4j2-test.xml
@@ -35,7 +35,7 @@
         </File>
     </Appenders>
     <Loggers>
-        <Root level="INFO">
+        <Root level="DEBUG">
             <AppenderRef ref="stdout"/>
             <AppenderRef ref="file"/>
         </Root>
--- a/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
+++ b/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
@@ -60,11 +60,7 @@ object ReflectUtils {
       }
     } catch {
       case e: Exception =>
-        val candidates =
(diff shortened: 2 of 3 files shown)


extensions/spark/kyuubi-spark-authz/src/test/resources/log4j2-test.xml
```diff
@@ -35,7 +35,7 @@
         </File>
     </Appenders>
     <Loggers>
-        <Root level="INFO">
+        <Root level="DEBUG">
```
yaooqinn, line 38: Debug message in CI for authz module can be turned on as most of its logs are at debug level and now it's less noisy.

kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
```diff
@@ -92,14 +88,8 @@ object ReflectUtils {
...
       case e: Exception =>
-        val candidates =
-          (clz.getDeclaredMethods ++ clz.getMethods)
-            .map(m => s"${m.getName}(${m.getParameterTypes.map(_.getName).mkString(", ")})")
-            .distinct.sorted
-        val argClassesNames = argClasses.map(_.getName)
         throw new RuntimeException(
-          s"Method $methodName(${argClassesNames.mkString(", ")})" +
-            s" not found in $clz [${candidates.mkString(", ")}]",
+          s"$clz does not have $methodName method or incorrect arguments",
```
bowenliang123, line 92: I suggest to keep the method signature `$methodName(${argClassesNames.mkString(", ")})` for further investigation. As for skipping candidates, LGTM.

bowenliang123 (approved):

bowenliang123:
Thanks, merged to master(1.9.0).

yaooqinn:
Thanks @bowenliang123

PR #5506, opened 2023-10-24
[AUTHZ][MINOR] Reuse loaded table extractor in Hudi table extractors

bowenliang123:

- Reuse loaded table extractor by using `lookupExtractor` in Hudi table extractors

No.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/serde/tableExtractors.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/serde/tableExtractors.scala
@@ -246,10 +246,10 @@ class HudiDataSourceV2RelationTableExtractor extends TableExtractor {
     invokeAs[LogicalPlan](v1, "table") match {
       // Match multipartIdentifier with tableAlias
       case SubqueryAlias(_, SubqueryAlias(identifier, _)) =>
-        new StringTableExtractor().apply(spark, identifier.toString())
+        lookupExtractor[StringTableExtractor].apply(spark, identifier.toString())
       // Match multipartIdentifier without tableAlias
       case SubqueryAlias(identifier, _) =>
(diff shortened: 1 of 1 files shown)


extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/serde/tableExtractors.scala
```diff
@@ -246,10 +246,10 @@ class HudiDataSourceV2RelationTableExtractor extends TableExtractor {
     invokeAs[LogicalPlan](v1, "table") match {
       // Match multipartIdentifier with tableAlias
       case SubqueryAlias(_, SubqueryAlias(identifier, _)) =>
-        new StringTableExtractor().apply(spark, identifier.toString())
+        lookupExtractor[StringTableExtractor].apply(spark, identifier.toString())
```
yaooqinn, line 249: any special reason for changing this? `new` seems straightforward

bowenliang123: No much. Just a minor improvement to reuse the stateless extractor instance. `lookupExtractor` is used all other 29 places.

yaooqinn: What's the state for stateless here? lookupExtractor seems much heavier for such a case

bowenliang123: The extractor does not own a state. And they are more like snippets that cooperate together for object extractions, so we are reusing them rather than creating an instance inside another. No much heavier in code lines and has less memory footprint.

yaooqinn (approved):

bowenliang123:
Thanks, merged to master.

Issue #5499, opened 2023-10-23
[Bug] KyuubiSessionManager timeout checker thread does not work


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The kyuubi batch connections always increasing and seems the timeout-checker does not work.

The batch session idle timeout is PT5M, but the batch connections always increasing.
<img width="1402" alt="image" src="https://github.com/apache/kyuubi/assets/6757692/772490e0-deee-4e87-947d-eb0f55c7dec3">

### Affects Version(s)

master

### Kyuubi Server Log Output

[kyuubi_jstack.txt](https://github.com/apache/kyuubi/files/13066403/kyuubi_jstack.txt), see `KyuubiSessionManager-timeout-checker`.

The result of `grep 'that has been idle for more' kyuubi.log` is empty.

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

turboFei:
only one kyuubi node has this issue. total 6 kyuubi nodes.

yaooqinn:
KyuubiSessionManager-timeout-checker seems in good shape?

turboFei:
> KyuubiSessionManager-timeout-checker seems in good shape?

I plan to log some info for debug.

wForget:
Similar to #2503?

wForget:
I guess this may be due to some exceptions not being caught.

turboFei:
> I guess this may be due to some exceptions not being caught in checkTask.

thanks @wForget let us catch any exception in checkTask https://github.com/apache/kyuubi/pull/5515

Issue #2503, opened 2022-04-28
[Bug] The scheduled task for session TimeoutChecker in SessionManager may stop.


wForget:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

There are some timeout sessions in kyuubi server that are not cleaned up.
I found that the workQueue of `SessionManager.timeoutChecker` is empty, but I didn't find the specific exception information that caused this.

> org$apache$kyuubi$session$SessionManager$$shutdown=@Boolean[false],

> workQueue=@DelayedWorkQueue[isEmpty=true;size=0],

`KyuubiSessionManager` details:

```
method=org.apache.kyuubi.session.SessionManager.getSessionList location=AtEnter
ts=2022-04-28 15:37:15; [cost=0.103374ms] result=@KyuubiSessionManager[
    operationManager=@KyuubiOperationManager[
        queryTimeout=@None$[None],
        handleToOperation=@HashMap[isEmpty=false;size=6],
        serviceName=@String[KyuubiOperationManager],
        conf=@KyuubiConf[KyuubiConf(true)],
        state=@Val[STARTED],
        startTime=@Long[1645090187643],
        org$apache$kyuubi$Logging$$log_=@Log4jLoggerAdapter[org.slf4j.impl.Log4jLoggerAdapter(org.apache.kyuubi.operation.KyuubiOperationManager)],
    ],
    credentialsManager=@HadoopCredentialsManager[
        userCredentialsRefMap=@ConcurrentHashMap[isEmpty=false;size=27],
        sessionCredentialsEpochMap=@ConcurrentHashMap[isEmpty=false;size=24],
        org$apache$kyuubi$credentials$HadoopCredentialsManager$$providers=@Map1[Map(hadoopfs -> org.apache.kyuubi.credentials.HadoopFsDelegationTokenProvider@7fc383da)],
        org$apache$kyuubi$credentials$HadoopCredentialsManager$$renewalInterval=@Long[3600000],
        org$apache$kyuubi$credentials$HadoopCredentialsManager$$renewalRetryWait=@Long[60000],
[...]

PR #5515, opened 2023-10-24
[KYUUBI #5499][KYUUBI #2503] Catch any exception when closing idle session

turboFei:

Now only the KyuubiSQLException is caught when closing idle session, we wonder some exception is not caught and cause the session timeout checkTask exit.

No.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
@@ -23,6 +23,7 @@ import java.util.concurrent.{ConcurrentHashMap, Future, ThreadPoolExecutor, Time
 
 import scala.collection.JavaConverters._
 import scala.concurrent.duration.Duration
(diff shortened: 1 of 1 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -312,8 +312,7 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
               try {
                 closeSession(session.handle)
               } catch {
-                case e: KyuubiSQLException =>
-                  warn(s"Error closing idle session ${session.handle}", e)
+                case e: Throwable => warn(s"Error closing idle session ${session.handle}", e)
```
yaooqinn, line 315: Use NonFatal here, exceptions like InterruptedException shall stay AS-IS？

pan3793 (approved):

pan3793:
Merged to master/1.8/1.7

Issue #5507, opened 2023-10-24
[TASK][EASY] Support Initialize SQL in Flink Engine


link3280:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [X] I have sufficient knowledge and experience of this task, and I volunteer to be the mentor of this task to guide contributors to complete the task.

### Skill requirements

- Basic knowledge of Flink Engine
- Basic knowledge of Flink SQL

### Background and Goals

Support initialize SQL in Flink SQL (e.g. setting up catalogs).

### Implementation steps

- Pass initialize SQL to Flink engine
- Execute the SQLs during engine bootstrap

link3280:
cc @hadoopkandy

abhinavs1920:
Please assign this issue to me

link3280:
@abhinavs1920 Thanks for reaching out! I'm afraid that @hadoopkandy has started the related work, so I'm assigning this to him first and will let you know if a hand is needed. Thanks again!

pan3793:
This task is completed by @hadoopkandy via #5507

PR #5518, opened 2023-10-25
[KYUUBI #5507][FLINK] Support Initialize SQL in Flink Engine

hadoopkandy:
We shoud support initialize SQL to init session context the  in Flink SQL  (e.g. setting up catalogs).

No.


link3280:
@hadoopkandy I think the initializing SQLs should be executed in the engine instead of Kyuubi server. Please refer to the implementation in Spark Engine. https://github.com/apache/kyuubi/blob/3646ae08bd064ac2aa53ba10a411b067d046e098/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala#L289-L293

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkEngineInitializeSuite.scala
```diff
@@ -0,0 +1,88 @@
...
+  def shareLevel: String = ShareLevel.USER.toString
+
+  def engineType: String = "flink"
+
+  test("execute statement - kyuubi engine initialize") {
+    withJdbcStatement() { statement =>
+      var resultSet = statement.executeQuery("show catalogs")
+      val expectedCatalogs = Set("default_catalog", "cat_a")
+      while (resultSet.next()) {
+        assert(expectedCatalogs.contains(resultSet.getString(1)))
```
link3280, line 71: The test may not cover the case when `cat_a` is absent.

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkEngineInitializeSuite.scala
```diff
@@ -0,0 +1,88 @@
...
+      }
+      assert(!resultSet.next())
```
link3280, line 73: I suppose this assertion is always true?

link3280 (changes requested):
Thanks a lot for the PR. The code generally looks good, but the tests could be improved a bit. Please take a look at my comments.

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkEngineInitializeSuite.scala
hadoopkandy: FlinkOperationLocalSuite extends FlinkOperationSuite, will cause FlinkOperationSuite's  unit tests to fail,e.g: "execute statement - set/get catalog" UT will thow exception : cat_a already exists , I commit is again.Please review

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkEngineInitializeSuite.scala
hadoopkandy: yes,After engine is initialized, I want to test there are only two catalogs: default_catalog and cat_a, there are two tables: blackhole and datagen

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/session/FlinkSessionImpl.scala
```diff
@@ -64,6 +65,24 @@ class FlinkSessionImpl(
   override def open(): Unit = {
     val executor = fSession.createExecutor(Configuration.fromMap(fSession.getSessionConfig))
 
+    sessionManager.getConf.get(ENGINE_INITIALIZE_SQL).foreach { sql =>
```
pan3793, line 68: it's supposed to run only one time after the engine is bootstrapped and before ready to serve connections

link3280 (approved):

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/session/FlinkSessionImpl.scala
link3280: @pan3793 I just moved the engine-level initial SQL to the engine bootstrap process. The downside is that the engine-level initial SQL would not affect sessions. Is that expected?

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkEngineInitializeSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+  protected val ENGINE_SESSION_INITIALIZE_SQL_VALUE: String =
+    s"""
+      create catalog cat_b with ('type'='generic_in_memory');
+      create table blackhole(i int) with ('connector'='blackhole');
+      create  table datagen(i int) with (
+      'connector'='datagen',
+      'fields.i.kind'='sequence',
+      'fields.i.start'='1',
+      'fields.i.end'='10')
+     """
```
pan3793, line 48: it's not common style in scala

```
val s = """the first line
          |the second line
          |the third line
          |the forth line""".stripMargin
```

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/session/FlinkSessionImpl.scala
pan3793: it's consistent with the current Implementation of the Spark engine.

Additional background: @cfmcgrady proposed a different idea, that `sparkSession.clone` should be called instead of `sparkSession.newSession` so that the newly created session could inherit the changes applied to the bootstrap session. The situation is quite different in Flink since there is no bootstrap session concept, thus I think the current behavior makes sense.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/FlinkSQLEngine.scala
```diff
@@ -129,15 +128,21 @@ object FlinkSQLEngine extends Logging {
...
+
+    if ("yarn-application".equalsIgnoreCase(executionTarget)) {
+      // trigger an execution to initiate EmbeddedExecutor with the default flink conf
+      flinkConf.set(PipelineOptions.NAME, "kyuubi-bootstrap-sql")
+      debug(s"Running bootstrap Flink SQL in application mode with flink conf: $flinkConf.")
+      tableEnv.executeSql("select 'kyuubi'").await()
+    }
+
+    kyuubiConf.get(ENGINE_INITIALIZE_SQL).foreach(
+      tableEnv.executeSql(_).await())
```
pan3793, line 143:
```suggestion
    kyuubiConf.get(ENGINE_INITIALIZE_SQL).foreach { stmt =>
      tableEnv.executeSql(stmt).await()
    }
```

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/WithFlinkSQLEngineLocal.scala
```diff
@@ -60,7 +60,6 @@ trait WithFlinkSQLEngineLocal extends KyuubiFunSuite with WithFlinkTestResources
       }
     }
     withKyuubiConf.foreach { case (k, v) =>
-      System.setProperty(k, v)
```
pan3793, line 63: what's the background of this change?

pan3793 (approved):
Overall LGTM, just left a few comments

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/WithFlinkSQLEngineLocal.scala
link3280: It's not relevant. It's probably a cleanup of unused codes.

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkEngineInitializeSuite.scala
```diff
@@ -0,0 +1,104 @@
...
+    "show databases;"
+
+  protected val ENGINE_SESSION_INITIALIZE_SQL_VALUE: String =
+    """create catalog cat_b with ('type'='generic_in_memory');
+      create table blackhole(i int) with ('connector'='blackhole');
+      create table datagen(i int) with (
+      'connector'='datagen',
+      'fields.i.kind'='sequence',
+      'fields.i.start'='1',
+      'fields.i.end'='10')""".stripMargin
```
pan3793, line 44: `|` is the default "margin"

```suggestion
    """create catalog cat_b with ('type'='generic_in_memory');
      |create table blackhole(i int) with ('connector'='blackhole');
      |create table datagen(i int) with (
      |'connector'='datagen',
      |'fields.i.kind'='sequence',
      |'fields.i.start'='1',
      |'fields.i.end'='10')""".stripMargin
```

pan3793:
Thanks, merged to master

pan3793:
I was asked frequently these days how to initialize the catalog in the Flink engine.

@link3280, if you don't mind, I propose to backport this patch to 1.8.1 so that we can deliver it soon.

link3280:
@pan3793 Sure. I'll create a backport PR in days.

pan3793:
@link3280 I cherry-pick it to branch-1.8 in 902815d5404caa71c0e3c1b5c04df6a31d11798c

Issue #5369, opened 2023-10-08
[TASK][EASY] Support setting the username to launch Kyuubi engines in USER & CONNECTION share level


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [ ] I have sufficient knowledge and experience of this task, and I volunteer to be the mentor of this task to guide contributors to complete the task.

### Skill requirements

- Basic knowledge on Scala Programing Language

### Background and Goals

Hive supports SQL operation authorization against the session user, and then runs SQL as Hive Server user.

Currently, Kyuubi supports this feature at the SERVER share level.

We want to also support this feature in other share levels, which provide fine isolation.

### Implementation steps

Currently, `EngineRef#appUser` serves two purposes:
- The username used to run the `SparkSQLEngine`
- Part of the key used to look up the `SparkSQLEngine` to connect to.

To implement this feature, we need to replace `EngineRef#appUser` with two new fields, `EngineRef#appUser` and `EngineRef#engineSpaceUser`.

- The new `EngineRef#appUser` is the username used to run the `SparkSQLEngine`.
- The `EngineRef#engineSpaceUser` is used to look up the `SparkSQLEngine`. 

`EngineRef#engineSpaceUser` is set in the same way as the old `EngineRef#appUser`.

The new `EngineRef#appUser` is set according to a new config, `kyuubi.engine.user`:
- When `kyuubi.engine.user` is not set, set `EngineRef#appUser` to `EngineRef#engineSpaceUser`
- When `kyuubi.engine.user` is set, set `EngineRef#appUser` to that value.

mrtisttt:
Hi, I am interested in this task, can you please assign it to me?

pan3793:
@mrtisttt thanks, assigned. This idea comes from @zhouyifan279, better to have a discussion before starting

cfmcgrady:
cc @cfmcgrady

mrtisttt:
@pan3793 @zhouyifan279 @cfmcgrady 

[...]

PR #5539, opened 2023-10-26
[KYUUBI #5369] Support setting the username to launch Kyuubi engines in USER & CONNECTION share level

mrtisttt:

To close https://github.com/apache/kyuubi/issues/5369

No.


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2148,6 +2148,16 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(false)
 
+  val ENGINE_USER: OptionalConfigEntry[String] =
```
yaooqinn, line 2151: Currently, we use `hive.server2.proxy.user` to keep the behavior compatible with hiveserver2

If you want to add an alternative key for it, please check how we handle `hive.server2.proxy.user` in KyuubiAuthenticationFactory. Client user changing need to go through AuthN

mrtisttt: @yaooqinn Hi, Thanks for the review! So I really forgot to do the proxy validation, because it's essentially a proxy.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -90,13 +90,19 @@ private[kyuubi] class EngineRef(
 
   private[kyuubi] def getEngineRefId(): String = engineRefId
 
-  // Launcher of the engine
-  private[kyuubi] val appUser: String = shareLevel match {
+  // Used to look up the engine
+  private[kyuubi] val sessionUser: String = shareLevel match {
```
zhouyifan279, line 94: `sessionUser` seems to be misleading (my fault to suggest this name) as it is not same to `kyuubi.session.user `.
We need a better name.

mrtisttt: How about `engineSpaceUser` or `namespaceUser`? As we use it to lookup the engine by default.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
yaooqinn: We can not add a configuration to the switch user outside the control of FrontendServices/Authentication. This varies the basic rule of how we achieve multi-tenancy architecture.

We need a configuration to be an alternative to `hive.server2.proxy.user`. `hive.server2.proxy.user` is derived from the hive server, and it's perfect for keeping the behavior the same as the hive. But hive thrift protocol is only one of the FrontendServices. There are also other FrontendServices not necessary to follow  `hive.server2.proxy.user`. So we need `kyuubi.proxy.user`.

As for #5369, IMHO, it's a misleading task

mrtisttt: @yaooqinn Ok, I see. I'm going to create another PR to implement the configuration item kyuubi.proxy.user as an option to hive.server2.proxy.user. By the way, when both parameters are provided, which takes precedence? Personally, I think it's kyuubi.proxy.user.

yaooqinn: Yes, I agree with you.

BTW, for the spark engine with USER and CONNECTION, I'm sure that whether hive.server2.proxy.user is used or not, the client user identity will be used for engine bootstrap and SQL operations. But we need to find out whether other engine types follow this rule.

If you are interested in this topic, you can also have a try.

mrtisttt: @yaooqinn Ok, I will first add the `kyuubi.proxy.user configuration`. Then I'll confirm how to converge on other engines.

yaooqinn: Thanks @mrtisttt. I truly appreciate that.

PR #5563, opened 2023-10-30
[KYUUBI #5503][FOLLOWUP][AUTHZ] Authz should skip inner plan that have been verified

### LINKED ISSUE #5503: [Bug] LATERAL SQL miss check the whole plan
### Code of Conduct

### Search before asking

### Describe the bug

For lateral sql
```
test("xxx") {
    val db1 = defaultDb
    val table1 = "table1"
    val table2 = "table2"
    val view1 = "view1"
    withCleanTmpResources(
      Seq((s"$db1.$table1", "table"), (s"$db1.$table2", "table"), (s"$db1.$view1", "view"))) {
      doAs(admin, sql(s"CREATE TABLE IF NOT EXISTS $db1.$table1 (id int, scope int)"))
      doAs(admin, sql(s"CREATE TABLE IF NOT EXISTS $db1.$table2 (id int, age int)"))
      doAs(admin, sql(
        s"""
           |SELECT t1.id, age
           |FROM $db1.$table1 t1,
           |LATERAL (
           |  SELECT *
           |  FROM $db1.$table2 t2
           |  WHERE t1.id = t2.id
           |)
           |""".stripMargin).explain(true))
    }
  }
```

When it only check the subquery, miss check whole plan
截屏2023-10-23 下午6 37 12

### Affects Version(s)

master, 1.8.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

AngersZhuuuu:
To close #5503
For sql such as lateral join in test `[KYUUBI #5503][AUTHZ] Check plan auth checked should not set tag to all child nodes`, it will first verify subquery in `lateral` then verify whole plan, if there is a view, when verify the whole plan, the `PermanentViewMarker` will be remove by spark's optimizer.
Then it will verify both source table `table1` and `table2`.
So I think we need to do 3 things:

1. Mark all PermanentViewMarker's children's all nodes as checked and Subquery's all child marks as checked.
2. `isAuthChecked` should only check the first level of the plan to avoid skipping the check of the whole plan in the demo test
3. in `buildQuery`, if the current node has the tag, we just skip it.

Without this pr, the SQL in test will both check `table1` and `table2`

No

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
@@ -957,12 +957,17 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
     assume(isSparkV32OrGreater, "Spark 3.1 not support lateral subquery.")
     val db1 = defaultDb
     val table1 = "table1"
+    val table2 = "table2"
     val perm_view = "perm_view"
     withSingleCallEnabled {
       withCleanTmpResources(
-        Seq((s"$db1.$table1", "table"), (s"$db1.$perm_view", "view"))) {
+        Seq(
+          (s"$db1.$table1", "table"),
+          (s"$db1.$table2", "table"),
+          (s"$db1.$perm_view", "view"))) {
         doAs(admin, sql(s"CREATE TABLE IF NOT EXISTS $db1.$table1 (id int, scope int)"))
-        doAs(admin, sql(s"CREATE VIEW $db1.$perm_view AS SELECT * FROM $db1.$table1"))
+        doAs(admin, sql(s"CREATE TABLE IF NOT EXISTS $db1.$table2 (id int, scope int)"))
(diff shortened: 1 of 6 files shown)


AngersZhuuuu:
ping @yaooqinn A follow up of #5503 to fix lateral join issue.

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -992,7 +997,38 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
...
+                 |SELECT t1.id
+                 |FROM $db1.$table1 t1,
+                 |LATERAL (
+                 |  SELECT *
+                 |  FROM $db1.$table2 t2
+                 |  WHERE t1.id = t2.id
+                 |)
+                 |""".stripMargin).show()))(
+          s"does not have [select] privilege on " +
+            s"[$db1/$table1/id]")
```
AngersZhuuuu, line 1031: Debug that when check the whole plan, only check table1, won't check table2 again.

AngersZhuuuu:
In our prod, this fix works well as expected.

yaooqinn (approved):

pan3793:
Thanks, merged to master

PR #5566, opened 2023-10-30
InternalRestClient respects `kyuubi.engine.security.enabled` to add HTTP auth header

pan3793:

`kyuubi.engine.security.enabled` aims to control whether enabled security mechanism internal communication, but the current implementation is not symmetrical, the auth generator ignores the conf and always produces the auth header, but the auth header handler is only activated when conf is enabled, that causes authentication failure when `kyuubi.engine.security.enabled=false`(default value)

No.


turboFei (approved):

pan3793:
Merged to master/1.8

PR #5568, opened 2023-10-30
Check administrator only when security is enabled

pan3793:

This PR aims to improve out-of-box UX.

Currently, administrators' checking is enforced even when security is disabled, and we are going to deliver the preview of Web UI in 1.8, it may confuse users who use the default configuration to launch Kyuubi sever and explore Kyuubi WebUI (because the user "anonymous" can not visit all users' sessions/engines without being configured as an administrator).

No


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -70,6 +71,17 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
 
   private lazy val port: Int = conf.get(FRONTEND_REST_BIND_PORT)
 
+  private lazy val isSecurityEnabled = {
+    val authTypes = conf.get(AUTHENTICATION_METHOD).map(AuthTypes.withName)
+    !authTypes.contains(NONE) && authTypes != Set(NOSASL)
```
lightning-L, line 76: Just a trivial question, is it possible to reuse the variables defined in `KyuubiAuthenticationFactory`.
```
  private val authTypes = conf.get(AUTHENTICATION_METHOD).map(AuthTypes.withName)
  private val none = authTypes.contains(NONE)
  private val noSasl = authTypes == Set(NOSASL)
```

lightning-L (approved):
LGTM

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
pan3793: Thanks for pointing out it, actually, I found bunches of duplicated code in THRIFT-HTTP and RESTful protocols, I'm fixing the tests and reducing the code duplications now.

pan3793: FYI: I refactored code to reuse `getValidPasswordAuthMethod` as you suggested

kyuubi-server/src/main/scala/org/apache/kyuubi/server/http/authentication/AuthenticationHandler.scala
```diff
@@ -103,23 +101,10 @@ trait AuthenticationHandler {
...
-
-  /**
-   * HTTP header used by the SPNEGO server endpoint during an authentication sequence.
-   */
-  final val WWW_AUTHENTICATE: String = HttpConstants.WWW_AUTHENTICATE_HEADER
-
-  /**
-   * HTTP header used by the client endpoint during an authentication sequence.
-   */
-  final val AUTHORIZATION_HEADER = "Authorization"
```
pan3793, line 124: moved such definitions to `HttpAuthUtils`

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -240,7 +252,7 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
       realUser
     } else {
       sessionConf.get(KyuubiAuthenticationFactory.HS2_PROXY_USER).map { proxyUser =>
-        if (!getConf.get(KyuubiConf.SERVER_ADMINISTRATORS).contains(realUser)) {
+        if (!isAdministrator(realUser)) {
```
pan3793, line 255: the behavior here has changed, previously, `Utils.currentUser` was not counted, which is inconsistent with checking administrator logic in `AdminResource`

zhouyifan279:
LGTM

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
pan3793: I mis-interpreted the suggestion, you suggested `AuthTypes` but I used `AuthMethods`, which cause `IllegalArgumentException("No valid Password Auth detected")` when configuring `kyuubi.authentication=KERBEROS`, will fix it soon.

PR #5570, opened 2023-10-31
Fix memory leak when using incremental collect mode in JDBC engine

wForget:

Similar to #3885, there is also memory leak in the jdbc engine when using incremental collect mode.

### DIFF
--- a/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteStatement.scala
@@ -18,11 +18,12 @@ package org.apache.kyuubi.engine.jdbc.operation
 
 import java.sql.{Connection, Statement, Types}
 
-import org.apache.kyuubi.Logging
+import org.apache.kyuubi.{KyuubiSQLException, Logging}
 import org.apache.kyuubi.engine.jdbc.schema.{Column, Row, Schema}
 import org.apache.kyuubi.engine.jdbc.session.JdbcSessionImpl
 import org.apache.kyuubi.engine.jdbc.util.ResultSetWrapper
-import org.apache.kyuubi.operation.{ArrayFetchIterator, IterableFetchIterator, OperationState}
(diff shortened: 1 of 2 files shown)


externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteStatement.scala
```diff
@@ -37,6 +39,8 @@ class ExecuteStatement(
   private val operationLog: OperationLog = OperationLog.createOperationLog(session, getHandle)
   override def getOperationLog: Option[OperationLog] = Option(operationLog)
 
+  @volatile private var jdbcStatement: Statement = _
```
yaooqinn, line 42: can it be a val？

wForget: It seems not, we need to set value in executeStatement

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteStatement.scala
```diff
@@ -89,10 +95,27 @@ class ExecuteStatement(
...
-        jdbcStatement.closeOnCompletion()
-      }
       shutdownTimeoutMonitor()
     }
   }
+
+  override def validateFetchOrientation(order: FetchOrientation): Unit = {
+    if (incrementalCollect && order != FetchOrientation.FETCH_NEXT) {
+      throw KyuubiSQLException(s"The fetch type $order is not supported" +
+        s" of incremental collect mode.")
```
cxzl25, line 105:
```suggestion
        " of incremental collect mode.")
```

wForget: thanks, fixed

pan3793 (approved):

cxzl25 (approved):

bowenliang123 (commented):
A late LGTM from me. Thanks for fixing this problem.

Issue #5505, opened 2023-10-23
[TASK][EASY] Introduce HELP command in Flink engine


YesOrNo828:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [ ] I have sufficient knowledge and experience of this task, and I volunteer to be the mentor of this task to guide contributors to complete the task.

### Skill requirements

- Familiar with Flink SQL

### Background and Goals

Show available commands when users type 'HELP;' in the beeline.

### Implementation steps

- Using Flink SQL Client Parser(SqlCommandParserImpl) obtain the Command enum
- Return Help message
- Add 'HELP' statement test

pan3793:
@YesOrNo828 have you started working on this task?

YesOrNo828:
> @YesOrNo828 have you started working on this task?

Yes, I found there are some incompatible APIs between Flink 1.16 and 1.17 versions. I'm trying to fix it.

btw, this is a demo of HELP command run in the Flink 1.18 
![image](https://github.com/apache/kyuubi/assets/6017591/e2224177-a561-4d4b-951e-743c4c57ce38)

pan3793:
Looks nice. One more question: do all commands listed below work on the Kyuubi Flink engine?

YesOrNo828:
> Looks nice. One more question: do all commands listed below work on the Kyuubi Flink engine?

I have checked :
QUIT/EXIT commands don't work. beeline needs to use '!exit' or '!quit' to exit the program. 
CLEAR command is invalid.
REMOVE JAR command is not supported in Flink SQL Gateway right now, I have created an [issue](https://issues.apache.org/jira/browse/FLINK-33374) to follow.
Other commands work fine.

PR #5585, opened 2023-10-31
[KYUUBI #5505][FLINK] Support HELP command

YesOrNo828:

resolve: #5505 

Show available commands when users type 'HELP;' in the beeline.

#### Solution:

- Using `ExtendedParser` parse statement and return Operation of Flink engine.
- Check whether the operation is HelpOperation or not.  
- dependency on `flink-table-planner-loader.jar`.

#### **Why not using Flink SQL Client Parser(SqlCommandParserImpl) to obtain the Command enum?**
Flink 1.16's approach:
```
val opt:Optional[Operation] = org.apache.flink.table.client.cli.SqlCommandParserImpl.parseCommand()
check opt.get() instance of HelpOperation or not
if yes return CliStrings.MESSAGE_HELP
```

Flink 1.17 & 1.18
```
val opt: Optional[Command] = org.apache.flink.table.client.cli.parser.SqlCommandParserImpl.parseStatement()
check opt.get() is Command.HELP or not
if yes return CliStrings.MESSAGE_HELP
```

The `Command` class is added from Flink1.17;
The `SqlCommandParserImpl` package is changed, and the method name is changed from Flink1.17;

This approach requires distinguishing between different Flink versions and maintaining both implementations.
It's more complicated, so abandoned.

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -111,7 +111,10 @@ class ExecuteStatement(
         .build[ClassLoader].bind(plannerModule).get
     } catch {
       case e: Exception =>
-        throw new TableException("Error obtaining Flink planner module ClassLoader", e)
+        throw new TableException(
+          "Error obtaining Flink planner module ClassLoader. " +
+            "Make sure a flink-table-planner-loader.jar is on the classpath",
+          e)
     }
   }
 }
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
@@ -17,10 +17,15 @@
 
 package org.apache.kyuubi.engine.flink.operation
 
+import java.util.Optional
+
 import scala.concurrent.duration.Duration
 
 import org.apache.flink.api.common.JobID
+import org.apache.flink.table.api.TableException
 import org.apache.flink.table.gateway.api.operation.OperationHandle
+import org.apache.flink.table.operations.Operation
+import org.apache.flink.table.operations.command.HelpOperation
 
 import org.apache.kyuubi.Logging
 import org.apache.kyuubi.engine.flink.FlinkEngineUtils
@@ -28,6 +33,8 @@ import org.apache.kyuubi.engine.flink.result.ResultSetUtil
 import org.apache.kyuubi.operation.OperationState
 import org.apache.kyuubi.operation.log.OperationLog
 import org.apache.kyuubi.session.Session
+import org.apache.kyuubi.util.reflect.DynConstructors
+import org.apache.kyuubi.util.reflect.ReflectUtils.{getField, invokeAs}
 
 class ExecuteStatement(
     session: Session,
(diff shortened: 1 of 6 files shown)


externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -59,16 +66,49 @@ class ExecuteStatement(
...
+        return
+      }
+
       val resultFetcher = executor.executeStatement(
         new OperationHandle(getHandle.identifier),
         statement)
       jobId = FlinkEngineUtils.getResultJobId(resultFetcher)
-      resultSet = ResultSetUtil.fromResultFetcher(resultFetcher, resultMaxRows, resultFetchTimeout)
+      resultSet = ResultSetUtil
+        .fromResultFetcher(resultFetcher, resultMaxRows, resultFetchTimeout)
```
pan3793, line 82: unnecessary change?

pom.xml
```diff
@@ -1385,6 +1385,13 @@
                 <scope>provided</scope>
             </dependency>
 
+            <dependency>
+                <groupId>org.apache.flink</groupId>
+                <artifactId>flink-table-planner-loader</artifactId>
+                <version>${flink.version}</version>
+                <scope>provided</scope>
```
pan3793, line 1392: it's not intended to define scope in the root pom dependencyManagement block

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -59,16 +66,49 @@ class ExecuteStatement(
...
+        .impl("org.apache.flink.table.planner.parse.ExtendedParser")
+        .build().newInstance()
+    invokeAs(extendedParser, "parse", (classOf[String], statement))
+      .asInstanceOf[Optional[Operation]]
+  }
+
+  private def getPlannerModuleClassLoader: ClassLoader = {
+    try {
+      val plannerModule =
+        invokeAs("org.apache.flink.table.planner.loader.PlannerModule", "getInstance")
```
pan3793, line 105: could this be invoked directly?

pan3793, line 105: it's encouraged to use `Dyn*` for reflection in Flink engine module

YesOrNo828: `PlannerModule` is package-private, we can't invoke it directly here.

externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/operation/ExecuteStatement.scala
```diff
@@ -71,4 +85,33 @@ class ExecuteStatement(
...
+        .hiddenImpl("org.apache.flink.table.planner.loader.PlannerModule")
+        .buildStaticChecked()
+        .invoke().asInstanceOf[AnyRef]
+
+      DynFields.builder()
+        .hiddenImpl(plannerModule.getClass, "submoduleClassLoader")
+        .build[ClassLoader].bind(plannerModule).get
+    } catch {
+      case e: Exception =>
+        throw new TableException("Error obtaining Flink planner module ClassLoader", e)
```
pan3793, line 114: Could we add some messages to guide users on how to fix it?

pan3793 (approved):
LGTM, except to one minor comment

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
```diff
@@ -1265,4 +1266,14 @@ abstract class FlinkOperationSuite extends HiveJDBCTestHelper with WithFlinkTest
...
     assert(exception.getMessage === "Futures timed out after [60000 milliseconds]")
   }
+
+  test("execute statement - help") {
+    withJdbcStatement() { stmt =>
+      val resultSet = stmt.executeQuery("help")
+      val metadata = resultSet.getMetaData
+      assert(metadata.getColumnName(1) === "result")
+      assert(resultSet.next())
+      assert(resultSet.getString(1).equals(CliStrings.MESSAGE_HELP.toString))
```
link3280, line 1276: IIUC, the command `CLEAR` and `QUIT/EXIT` in the help message are not supported. Would that be kind of misleading?

link3280 (commented):
Thanks for the PR! I left some comments please kindly take a look. @YesOrNo828

externals/kyuubi-flink-sql-engine/src/test/scala/org/apache/kyuubi/engine/flink/operation/FlinkOperationSuite.scala
YesOrNo828: Yes, I think you're right. This `CliStrings.MESSAGE_HELP` message is described in the Flink SQL Client. CLEAR and QUIT/EXIT are not supported for Beeline.

How about we define the help message in the Flink Engine? Removing the CLEAR and QUIT/EXIT command descriptions may be a little tricky from `CliStrng.MESSAGE_HELP`. We could also add some new descriptions, for example: show jobs, stop job 'job_id', create Table, drop table, etc. WDYT?

link3280: That sounds promising. +1 from my side.

YesOrNo828: This is the screenshot of `HELP;` command. Do you have a time to take a look? @link3280 

image

link3280: I think job statements are very common and should be included. Plus, we could state that the help message only contains the most frequently used statements. Thanks!

link3280:
@YesOrNo828 Merged. Thanks!

Issue #5464, opened 2023-10-18
[TASK][MEDIUM] JDBC Engine supports MySQL


cxzl25:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [X] I have sufficient knowledge and experience of this task, and I volunteer to be the mentor of this task to guide contributors to complete the task.

### Skill requirements

- Basic knowledge on Scala Programing Language

### Background and Goals

Now JDBC engine does not support MySQL Dialect (`kyuubi.engine.jdbc.type=mysql`)

Although we can use `kyuubi.engine.jdbc.type=doris` to connect to the MySQL server.

### Implementation steps

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect

- Add MySQL dialect

Kwafoor:
hello，can i take this?  I expect to achieve it in two weeks.

pan3793:
@Kwafoor assigned, good luck!

Kwafoor:
> @Kwafoor assigned, good luck!

Thanks, i will try to finish this in two weeks.

pan3793:
Escalate task level from easy to medium due to the complexity of implementation

PR #5488, opened 2023-10-20
Introduce ElasticSearch event logger for server events

bowenliang123:
### _Why are the changes needed?_

- introduce new event logger type `ELASTICSEARCH` for server
- send server events to the ElasticSearch index
- support auto-created index with dynamic mapping type
- use `elastic4s` 8.9.1 as ElasticSearch client for Scala, which depends on the ElasticSearch Java Client `elasticsearch-rest-client` 8.9.x.
   - License: `elastic4s` and `elasticsearch-rest-client` released with Apache Licence 2.0. `elasticsearch-rest-client` does not depends on the ElasticSearch server library.
- Tested with the latest release version of ElasticSearch server 7.x and 8.x

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] [Run test](https://kyuubi.readthedocs.io/en/master/contributing/code/testing.html#running-tests) locally before make a pull request

### _Was this patch authored or co-authored using generative AI tooling?_

No.

PR #5588, opened 2023-10-31
[KYUUBI #5464] JDBC Engine supports MySQL

Kwafoor:

To close https://github.com/apache/kyuubi/issues/5464.
To support JDBC engine use MySQL Dialect (kyuubi.engine.jdbc.type=mysql).

No.


externals/kyuubi-jdbc-engine/src/test/scala/org/apache/kyuubi/engine/jdbc/mysql/WithMysqlContainer.scala
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package org.apache.kyuubi.engine.jdbc.mysql
+
+import com.dimafeng.testcontainers.{GenericContainer, SingleContainer}
+import org.testcontainers.containers.wait.strategy.Wait
+
+import org.apache.kyuubi.engine.jdbc.WithJdbcServerContainer
+
+trait WithMysqlContainer extends WithJdbcServerContainer {
```
bowenliang123, line 24: Better to use testcontainers for Mysql instead.
https://github.com/testcontainers/testcontainers-scala/blob/master/docs/src/main/tut/usage.md#mysql

bowenliang123 (changes requested):
Use testcontainers for Mysql instead.

externals/kyuubi-jdbc-engine/src/test/scala/org/apache/kyuubi/engine/jdbc/mysql/WithMysqlContainer.scala
Kwafoor: > Better to use testcontainers for Mysql instead. https://github.com/testcontainers/testcontainers-scala/blob/master/docs/src/main/tut/usage.md#mysql

We are currently using testcontainers-scala-scalatest, which already meets our needs. To use Mysql testcontainers, we need to add the `testcontainers-scala-mysql` dependency. Is it necessary?

bowenliang123: It's alright to introduce test scope dependency for testing different datasource provider. In this way, it aovids redundant wrapping lower level testcontainers api. Testconainers for kafka was included and for es is being included in another pending pr.

Kwafoor: > It's alright to introduce test scope dependency for testing different datasource provider. In this way, it aovids redundant wrapping lower level testcontainers api. Testconainers for kafka was included and for es is being included in another pending pr.

OK.

externals/kyuubi-jdbc-engine/src/main/resources/META-INF/services/org.apache.kyuubi.engine.jdbc.connection.JdbcConnectionProvider
```diff
@@ -16,4 +16,5 @@
 #
 
 org.apache.kyuubi.engine.jdbc.doris.DorisConnectionProvider
-org.apache.kyuubi.engine.jdbc.phoenix.PhoenixConnectionProvider
\ No newline at end of file
+org.apache.kyuubi.engine.jdbc.phoenix.PhoenixConnectionProvider
+org.apache.kyuubi.engine.jdbc.mysql.MysqlConnectionProvider
```
pan3793, line 20: use the correct term "MySQL"

Kwafoor: fixed

Kwafoor:
retest

Kwafoor:
How can I re-run the checks?

externals/kyuubi-jdbc-engine/src/test/scala/org/apache/kyuubi/engine/jdbc/mysql/StatementSuite.scala
```diff
@@ -0,0 +1,95 @@
...
+
+class StatementSuite extends WithMySQLEngine with HiveJDBCTestHelper {
+
+  test("test select") {
+    withJdbcStatement("test1") { statement =>
+      statement.execute("create database if not exists db1")
+      statement.execute("use db1")
+      statement.execute("create table db1.test1(id bigint, name varchar(255), age int, " +
+        "PRIMARY KEY ( `id` ))" +
+        "ENGINE=InnoDB\n" +
```
bowenliang123, line 31: Remove unnecessary "\n".

externals/kyuubi-jdbc-engine/src/test/scala/org/apache/kyuubi/engine/jdbc/mysql/WithMySQLContainer.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package org.apache.kyuubi.engine.jdbc.mysql
+
+import com.dimafeng.testcontainers.{MySQLContainer, SingleContainer}
+import org.testcontainers.utility.DockerImageName
+
+import org.apache.kyuubi.engine.jdbc.WithJdbcServerContainer
+
+trait WithMySQLContainer extends WithJdbcServerContainer {
```
bowenliang123, line 24: You dont have to wrap the Mysql container again.
For creating/starting and stopping containers,  the trait`TestContainerForAll` is perfect for the test suites.
For setting and getting query string, use the `testQueryString` of `MySQLContainer`.
For image name and other common settings, they are all available in the `MySQLContainer` object.

bowenliang123 (changes requested):
Avoid to wrap the Mysql server container again, better to use `MySQLContainer` property. And if you have troubles or worries, you can refer `ServerKafkaLoggingEventHandlerSuiteForKafka3` or the changes in #5488(https://github.com/apache/kyuubi/pull/5488/files) to see how to test with testcontainers.

externals/kyuubi-jdbc-engine/src/test/scala/org/apache/kyuubi/engine/jdbc/mysql/StatementSuite.scala
Kwafoor: fixed

Kwafoor:
> Avoid to wrap the Mysql server container again, better to use `MySQLContainer` property. And if you have troubles or worries, you can refer `ServerKafkaLoggingEventHandlerSuiteForKafka3` or the changes in #5488(https://github.com/apache/kyuubi/pull/5488/files) to see how to test with testcontainers.

fixed

Kwafoor:
@pan3793 cc

bowenliang123 (approved):
LGTM.

bowenliang123:
Please fix the failed style checks.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
```diff
@@ -57,10 +57,9 @@ abstract class AbstractConnectionProvider extends Logging {
       case None =>
         // TODO
         if (filteredProviders.size != 1) {
-          throw new IllegalArgumentException(
-            "JDBC connection initiated but more than one connection provider was found. Use " +
-              s"${ENGINE_JDBC_CONNECTION_PROVIDER.key} option to select a specific provider. " +
-              s"Found active providers ${filteredProviders.mkString("[", ", ", "]")}")
+          warn("JDBC connection initiated but more than one connection provider was found. Use " +
```
Kwafoor, line 60: @bowenliang123 @pan3793 Both `MySQLConnectionProvider` and `DorisConnectionProvider` utilize `com.mysql.cj.jdbc.Driver`. This causes an exception. I have already changed it to a warn log here, will there be any problems?

bowenliang123: As both Doris and Mysql server are compatible the mysql driver, how about simply reuse `Mysql8ConnectionProvider` instead of introducing the new subclass `MySQLConnectionProvider` ?

Kwafoor:
re-run the checks

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
Kwafoor: I believe using a distinct subclass of `MySQLConnectionProvider` would be better, as it helps users distinguish between different engines(MySQL,Doris). The presence of multiple connection providers does not impact user's experience, and shouldn't throw an exception and terminate the process. Instead, it should just remind users to set the `ENGINE_JDBC_CONNECTION_PROVIDER` parameter.

Kwafoor:
image
This error appears to be unrelated to my changes. Could it be a probabilistic issue?

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
Kwafoor: > As both Doris and Mysql server are compatible the mysql driver, how about simply reuse `Mysql8ConnectionProvider` instead of introducing the new subclass `MySQLConnectionProvider` ?

This method seems good, but it may lead to confusion when users set the `kyuubi.engine.jdbc.connection.provider` parameter (since Doris and MySQL use the same provider). In contrast, I think that since the driver used by the provider is the same, it does not affect the user's usage. Users just need to be reminded to set the `kyuubi.engine.jdbc.connection.provider` parameter to specify the provider.

@pan3793 What do you think?

Kwafoor:
@pan3793 cc

externals/kyuubi-jdbc-engine/src/main/resources/META-INF/services/org.apache.kyuubi.engine.jdbc.connection.JdbcConnectionProvider
```diff
@@ -16,5 +16,6 @@
 #
 
 org.apache.kyuubi.engine.jdbc.doris.DorisConnectionProvider
+org.apache.kyuubi.engine.jdbc.mysql.MySQLConnectionProvider
 org.apache.kyuubi.engine.jdbc.phoenix.PhoenixConnectionProvider
-org.apache.kyuubi.engine.jdbc.postgresql.PostgreSQLConnectionProvider
+org.apache.kyuubi.engine.jdbc.postgresql.PostgreSQLConnectionProvider
```
pan3793, line 21: nit: keep newline at the end

externals/kyuubi-jdbc-engine/src/main/resources/META-INF/services/org.apache.kyuubi.engine.jdbc.dialect.JdbcDialect
```diff
@@ -16,5 +16,6 @@
 #
 
 org.apache.kyuubi.engine.jdbc.dialect.DorisDialect
+org.apache.kyuubi.engine.jdbc.dialect.MySQLDialect
 org.apache.kyuubi.engine.jdbc.dialect.PhoenixDialect
-org.apache.kyuubi.engine.jdbc.dialect.PostgreSQLDialect
+org.apache.kyuubi.engine.jdbc.dialect.PostgreSQLDialect
```
pan3793, line 21: ditto

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
pan3793: @Kwafoor your suggestion LGTM, but I get no idea after reading the description of `kyuubi.engine.jdbc.connection.provider`, could you please enrich the docs?

```
The connection provider is used for getting a connection from the server
```

Kwafoor:
> ate the docs of `kyuubi.engine.type`

fixed

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
Kwafoor: How about that？

pan3793: the docs of `kyuubi.session.group.provider` should be a good example

pan3793: BTW, the `settings.md` is generated by `dev/gen/gen_all_config_docs.sh`

Kwafoor: > the docs of `kyuubi.session.group.provider` should be a good example

done

Kwafoor:
retest

Kwafoor:
@pan3793 @bowenliang123 cc

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/mysql/MySQLConnectionProvider.scala
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kyuubi.engine.jdbc.mysql
+
+class MySQLConnectionProvider extends Mysql8ConnectionProvider {
```
bowenliang123, line 19: Again why not use `Mysql8ConnectionProvider` for connection provider directly? What's the key point for introducing such a dedicated provider?

bowenliang123: Or how about create an abstract MySQLConnectionProvider and make Mysql8ConnectionProvider as a subclass of it.

Kwafoor: > Again why not use `Mysql8ConnectionProvider` for connection provider directly? What's the key point for introducing such a dedicated provider?

I think the connection provider is used for creating JDBC connections. If different databases utilize the same connection provider, it might cause confusion for users. Using distinct connection providers would also facilitate future customizations of connections to Doris and MySQL.

Kwafoor:
@pan3793 cc

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/mysql/MySQLConnectionProvider.scala
bowenliang123: There are many MySQL protocol-compatible databases. We don't need separate connection providers for each of them, especially like  Doris/Starrocks which uses Mysql driver solely and does not even have its own JDBC driver.

Kwafoor:
@pan3793 cc

pan3793:
@Kwafoor ack, sorry for late, starting taking look

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/MySQLDialect.scala
```diff
@@ -0,0 +1,154 @@
...
+    throw KyuubiSQLException.featureNotSupported()
+  }
+
+  override def getPrimaryKeysOperation(session: Session): Operation = {
+    throw KyuubiSQLException.featureNotSupported()
+  }
+
+  override def getCrossReferenceOperation(session: Session): Operation = {
+    throw KyuubiSQLException.featureNotSupported()
+  }
```
pan3793, line 141: seems unnecessary, same logic in the parent class

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/MySQLDialect.scala
```diff
@@ -0,0 +1,154 @@
...
+    if (filters.nonEmpty) {
+      query.append(" WHERE ")
+      query.append(filters.mkString(" AND "))
+    }
+
+    query.toString()
+  }
+
+  override def getTableTypesOperation(session: Session): Operation = {
+    throw KyuubiSQLException.featureNotSupported()
```
pan3793, line 89: ditto, redundant

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2780,7 +2780,15 @@ object KyuubiConf {
 
   val ENGINE_JDBC_CONNECTION_PROVIDER: OptionalConfigEntry[String] =
     buildConf("kyuubi.engine.jdbc.connection.provider")
-      .doc("The connection provider is used for getting a connection from the server")
+      .doc("A JDBC connection provider plugin for the Kyuubi Server " +
+        "to establish a connection to the JDBC URL." +
+        " The configuration value should be a subclass of " +
+        "`org.apache.kyuubi.engine.jdbc.connection.JdbcConnectionProvider`. " +
+        "Kyuubi provides the following built-in implementations: " +
+        "<li>DorisConnectionProvider: For establishing Doris connections.</li> " +
```
pan3793, line 2788: this is inconsistent with other similar parameters, can we accept short names like "doris", "mysql" (case-insensitive), and full class name here? (for existing in 1.8.0, should handle properly to keep compatiblity)

pan3793:
@Kwafoor this task is more complex than I thought at the beginning, escalated the task level to medium

Kwafoor:
> @Kwafoor this task is more complex than I thought at the beginning, escalated the task level to medium

OK, I will finish it as soon as possible.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2780,9 +2780,24 @@ object KyuubiConf {
...
+        "`org.apache.kyuubi.engine.jdbc.connection.JdbcConnectionProvider`. " +
+        "Kyuubi provides the following built-in implementations: " +
+        "<li>doris: For establishing Doris connections.</li> " +
+        "<li>mysql: For establishing MySQL connections.</li> " +
+        "<li>phoenix: For establishing Phoenix connections.</li> " +
+        "<li>postgresql: For establishing PostgreSQL connections.</li>")
       .version("1.6.0")
       .stringConf
+      .transform {
+        case "Doris" | "doris" => "DorisConnectionProvider"
```
pan3793, line 2795: I thought we should transform the short name and simple class name to full class name

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
```diff
@@ -57,10 +57,9 @@ abstract class AbstractConnectionProvider extends Logging {
       case None =>
         // TODO
         if (filteredProviders.size != 1) {
-          throw new IllegalArgumentException(
```
pan3793, line 60: filter with the full class name here

Kwafoor: providers filtered by `driverClass` , and `kyuubi.engine.jdbc.connection.provider` is not set, how to filter with the full class name here?
```
providers.filter(_.canHandle(getProviderClass(kyuubiConf)))
```

Kwafoor: @pan3793 cc

Kwafoor:
@pan3793 cc

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
pan3793: I see, method name `getProviderClass` is confusing, let's rename it to `getDriverClass`, other LGTM

pan3793 (approved):

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/connection/ConnectionProvider.scala
Kwafoor: OK

Issue #5470, opened 2023-10-19
[Umberella] Support Paimon in Authz


bowenliang123:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues) and found no similar issues.

### Describe the proposal

Introduce Paimon support in Kyuubi Authz plugin.

Apache Paimon is an incubating Apache project of data lake platform for high-speed data ingestion, changelog tracking and efficient real-time analytics.

### Task list

Paimon Spark 3 Reference: https://paimon.apache.org/docs/master/engines/spark3/

- DDL Create Tables:
  - [x] Create Table: #5462
  - [ ] Create Partitioned Table: https://paimon.apache.org/docs/master/how-to/creating-tables/#partitioned-tables
  - [ ] Create Table As: https://paimon.apache.org/docs/master/how-to/creating-tables/#create-table-as
- DDL Alter Tables : https://paimon.apache.org/docs/master/how-to/altering-tables
  - [ ] Changing/Adding Table Properties
  - [ ] Rename Table Name
  - [ ] Removing Table Properties
  - [ ] Adding New Columns
  - [ ] Renaming Column Name
  - [ ] Dropping Columns
  - [ ] Changing Column Nullability
  - [ ] Changing Column Comment
  - [ ] Adding Column Position
  - [ ] Changing Column Position
  - [ ] Changing Column Type
  - [ ] Adding watermark
  - [ ] Dropping watermark
  - [ ] Changing watermark
- DML:
  - [ ] Insert Table
  - [ ] Query Table
    - [ ] Batch Time Travel: https://paimon.apache.org/docs/master/how-to/querying-tables/#batch-time-travel
  - [ ] Update Table
- Management procedures:
  - [ ] #5430
     - [ ] Rollback to snapshot: https://paimon.apache.org/docs/master/maintenance/manage-snapshots/#rollback-to-snapshot

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [x] No. I cannot submit a PR at this time.

davidyuan1223:
hello, i want try this issue, but i'm not sure how to improve it, this code is a createTableAsSql for paimon
```
  test("CreateTableAs") {
    withCleanTmpResources(Seq((s"$catalogV2.$namespace1.$table1", "table"),
[...]

PR #5590, opened 2023-10-31
[AUTHZ] Support CreateTableAs command for Paimon

bowenliang123:

Support CreateTableAs command for Paimon
  - Create Table As Command: https://paimon.apache.org/docs/master/engines/spark3/#create-table
  - to validate https://github.com/apache/kyuubi/issues/5470#issuecomment-1787008858

No.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/PaimonCatalogRangerSparkExtensionSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/PaimonCatalogRangerSparkExtensionSuite.scala
@@ -85,7 +85,8 @@ class PaimonCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
 
   test("CreateTableAs") {
     val table2 = "table2"
-    withCleanTmpResources(Seq((s"$catalogV2.$namespace1.$table1", "table"),
+    withCleanTmpResources(Seq(
+      (s"$catalogV2.$namespace1.$table1", "table"),
       (s"$catalogV2.$namespace1.$table2", "table"))) {
       doAs(
         admin,
@@ -97,10 +98,8 @@ class PaimonCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
              |OPTIONS (
              |  primaryKey = 'id'
              |)
-             |""".stripMargin
-        )
-      )
-      val createTable =
+             |""".stripMargin))
+      val createTableAs =
         s"""
            |CREATE TABLE IF NOT EXISTS $catalogV2.$namespace1.$table2
            |USING PAIMON
@@ -108,14 +107,9 @@ class PaimonCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
            |SELECT * FROM $catalogV2.$namespace1.$table1
            |""".stripMargin
       interceptContains[AccessControlException] {
-        doAs(
-          someone,
-          sql(
-            createTable
-          )
-        )
+        doAs(someone, sql(createTableAs))
       }(s"does not have [select] privilege on [$table1/id]")
-      doAs(admin, createTable)
+      doAs(admin, createTableAs)
     }
   }
 }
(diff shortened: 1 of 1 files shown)


davidyuan1223:
Look like the test case can success in CI but can not success in local env, could you give some suggests? @pan3793

davidyuan1223:
I want ask a question, why the admin to execute doAs in the next, don't use sql to execute, i found your code is `doAs(admin, createTableAs)`, why not is `doAs(admin,sql(createTableAs))`. 
And i found the DeltaCatalogRanger also use doAs(user,mySql),but the HudiCaralog use doAs(user,sql(mySql)). 
Which is the right way???

@pan3793 could you give some advice

davidyuan1223:
> I want ask a question, why the admin to execute doAs in the next, don't use sql to execute, i found your code is `doAs(admin, createTableAs)`, why not is `doAs(admin,sql(createTableAs))`. And i found the DeltaCatalogRanger also use doAs(user,mySql),but the HudiCatalog use doAs(user,sql(mySql)). Which is the right way???
> 
> @pan3793 could you give some advice

doAs code is
```
  protected def doAs[T](user: String, f: => T): T = {
    UserGroupInformation.createRemoteUser(user).doAs[T](
      new PrivilegedExceptionAction[T] {
        override def run(): T = f
      })
  }
```
If we not use spark.sql to execute our sql, it's only a string format???

yaooqinn:
@davidyuan1223 I believe you are right, and @bowenliang123 made a mistake. Nice catch.

davidyuan1223:
> @davidyuan1223 I believe you are right, and @bowenliang123 made a mistake. Nice catch.

And I found the [DeltaCatalogRanger](https://github.com/apache/kyuubi/blob/master/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/DeltaCatalogRangerSparkExtensionSuite.scala) also is this code style, I think you need to check the author module's test case to avoid this problem. This problem just like this [comment](https://github.com/apache/kyuubi/issues/5470#issuecomment-1787008858), I want know why admin create table but it can not select from the table?

bowenliang123:
Sorry about that. Will fix and have further investigation.

bowenliang123:
@davidyuan1223 More investigation in the logical plan is required to find the reason. Let me update the test case for Paimon first. And thanks for mentioning possible bug for Delta test suite as well.

davidyuan1223:
> @davidyuan1223 More investigation in the logical plan is required to find the reason. Maybe missing fields in parsed privilege objects. Let me update the test case for Paimon first. And thanks for mentioning possible bug for Delta test suite as well.

thanks your reply, if you find the root cause about this bug, wish you can teach me how to fix this, so that I can complete the issue #5470

bowenliang123:
> > @davidyuan1223 More investigation in the logical plan is required to find the reason. Maybe missing fields in parsed privilege objects. Let me update the test case for Paimon first. And thanks for mentioning possible bug for Delta test suite as well.
> 
> thanks your reply, if you find the root cause about this bug, wish you can teach me how to fix this, so that I can complete the issue #5470

Sure. Will ping you when I came to any conclusion for this. Thanks for creating a bug issue for it.

bowenliang123:
The initial reason is that missing `database`  field in RangerAccessRequest, and the admin user is granted on db-table-column `*-*-*` resource rules.

image

davidyuan1223:
> RangerAccessRequest

LGTM, I will debug in my env

davidyuan1223:
> The initial reason is that missing `database` field in RangerAccessRequest, and the admin user is granted on db-table-column `*-*-*` resource rules.
> 
> image

this is the logic plan about createTableAs, looks like it's plan have some problem
```
CreateTableAsSelect org.apache.paimon.spark.SparkCatalog@4a5f435b, paimon_ns.table2, [provider=PAIMON], true
+- RelationV2[id#6, name#7, city#8] table1
```

bowenliang123:
Yes. Will have a dive-in debugging  in extractors later.

bowenliang123:
Seems that there's a Paimon's `org.apache.paimon.AppendOnlyFileStore` as the Spark Table inside the input query plan, which is unsupported by the DataSourceV2RelationTableExtractor.

image

davidyuan1223:
> Seems that there's a Paimon's `org.apache.paimon.AppendOnlyFileStore` as the Spark Table inside the input query plan, which is unsupported by the DataSourceV2RelationTableExtractor.
> 
> image

looks like the paimon can not supported in kyuubi's authz,? we can not make sure the paimon's plan is same as spark's plan

bowenliang123:
It's normal to see lakehouse has individual implementation of spark plan for some commands. They're still part of spark plans. To support them, we could adapt our extractor for the details of the plan. But the precondition is that table identifier is carried by the plan. Should have look at filestore class of paimon for that.

davidyuan1223:
> It's normal to see lakehouse has individual implementation of spark plan in some commands. They're still part of spark plans. To support them, we could adapt our extractor for the details of the plan. But the precondition is that table identifier is carried by the plan. Should have look at filestore class of paimon for that.

image
The paimon's FileStoreTable top interface's method name is only get the tableName, what can we do to improve the paimon could return database? The source code in 0.5 only will return like `xxx.db/tableName` to tableName

PR #5591, opened 2023-10-31
[KYUUBI #5377] Spark engine query result save to file

### LINKED ISSUE #5377: [TASK][MEDIUM] Spark engine query results support reading from HDFS
### Code of Conduct

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [ ] I have sufficient knowledge and experience of this task, and I volunteer to be the mentor of this task to guide contributors to complete the task.

### Skill requirements

- Basic knowledge on Scala Programing Language
- Familiar with Apache Spark

### Background and Goals

The client's SQL query may cause the Spark engine to fail because the query data result is too large, and the driver to fail due to OOM.
Although the number of query results can be limited by configuring `kyuubi.operation.result.max.rows`, if the amount of data in one row is too large, it will still cause OOM.
If it can support writing query results to HDFS or other storage systems, when the client needs to obtain the results, the engine will obtain the results from HDFS, which can avoid the OOM problem.

### Implementation steps

1. Distinguish execution plans with query results
2. Estimate the output size of the execution plan
`org.apache.spark.sql.catalyst.plans.logical.statsEstimation.EstimationUtils#getSizePerRow`
3. Use `df.write` api to write the query results of the execution plan to HDFS
4. Implement an iterator to read data from HDFS and return it to the client

### Additional context

Original reporter is @cxzl25

lsm1:

close #5377

NO


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -101,6 +113,10 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
         exec,
         Duration(60, TimeUnit.SECONDS))
     })
+    engineSavePath.foreach { p =>
```
wForget, line 116: Since `engineSavePath` is a temporary directory, why not use `spark tmp dir` directly?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -159,6 +160,25 @@ class ExecuteStatement(
...
+        .getOrElse(session.sessionManager.getConf.get(OPERATION_RESULT_SAVE_TO_FILE))
+      lazy val threshold =
+        session.sessionManager.getConf.get(OPERATION_RESULT_SAVE_TO_FILE_THRESHOLD)
+      if (hasResultSet && sparkSave && shouldSaveResultToHdfs(resultMaxRows, threshold, result)) {
+        val sessionId = session.handle.identifier.toString
+        val savePath = session.sessionManager.getConf.get(OPERATION_RESULT_SAVE_TO_FILE_PATH)
+        val fileName = s"$savePath/$engineId/$sessionId/$statementId"
+        val colName = range(0, result.schema.size).map(x => "col" + x)
+        if (resultMaxRows > 0) {
+          result.toDF(colName: _*).limit(resultMaxRows).write
```
wForget, line 173: Use `resultDF` instead of `result`. Also, is `toDF(colName: _*)` necessary?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -0,0 +1,120 @@
...
+      val cls = Class.forName("org.apache.spark.sql.execution.datasources.orc.OrcDeserializer")
+      val constructor = cls.getDeclaredConstructors.apply(0)
+      if (constructor.getParameterCount == 3) {
+        constructor.newInstance(new StructType, orcSchema, colId).asInstanceOf[OrcDeserializer]
+      } else {
+        constructor.newInstance(orcSchema, colId).asInstanceOf[OrcDeserializer]
+      }
+    } catch {
+      case e: Throwable =>
+        throw new Exception("Failed to create OrcDeserializer", e)
```
wForget, line 79: use `KyuubiSQLException`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -0,0 +1,120 @@
...
+    if (iters(idx).hasNext) {
+      true
+    } else {
+      iters(idx).close()
+      idx < iters.size - 1
+    }
+  }
+
+  override def next(): OrcStruct = {
+    if (iters(idx).hasNext) {
```
wForget, line 100: `iters(idx).hasNext)` has been called in the `hasNext` method

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -159,6 +160,25 @@ class ExecuteStatement(
         override def iterator: Iterator[Any] = incrementalCollectResult(resultDF)
       })
     } else {
+      val sparkSave = spark.conf.getOption(OPERATION_RESULT_SAVE_TO_FILE.key).map(_.toBoolean)
+        .getOrElse(session.sessionManager.getConf.get(OPERATION_RESULT_SAVE_TO_FILE))
+      lazy val threshold =
+        session.sessionManager.getConf.get(OPERATION_RESULT_SAVE_TO_FILE_THRESHOLD)
+      if (hasResultSet && sparkSave && shouldSaveResultToHdfs(resultMaxRows, threshold, result)) {
```
wForget, line 167: `shouldSaveResultToHdfs` is inferred based on the execution plan and may not be accurate. Should we change it to `sparkSave || shouldSaveResultToHdfs`

wForget:
It might be simpler for us to make changes in the executeStatement method, like:

change

```
result = spark.sql(statement)
```

to

```
if (saveResultToPath) {
  spark.sql(statement).write.format(format).save(resultPath)
  result = spark.read.load(resultPath)
} else {
  result = spark.sql(statement)
}
```

WDYT? cc @pan3793 @cxzl25

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
lsm1: If the result has duplicate columns, we can not write it to file, so we rename all col name to avoid this case
```
spark.sql("select 1 as a,2 as a").write("/filepath")

org.apache.spark.sql.AnalysisException: Found duplicate column(s) when inserting into hdfs
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
cxzl25: `engineSavePath` can be HDFS as the storage path, so it is not the same as tmp dir.

cxzl25:
> It might be simpler for us to make changes in the `executeStatement` method, like:

> result = spark.read.load(resultPath)

This may lose the ordering of the query data, e.g. `order by xx limit 100`

wForget:
> This may lose the ordering of the query data, e.g. `order by xx limit 100`

I did a simple test and the results were as expected. (Test Env: Kyuubi 1.8.0 + Spark 3.5.0)

```
create table wangzhen_test_20231115_t1(id bigint, name string) stored as parquet;
insert into wangzhen_test_20231115_t1 values (1, 'a');
insert into wangzhen_test_20231115_t1 values (2, 'b');
insert into wangzhen_test_20231115_t1 values (3, 'c');

set kyuubi.operation.language=scala;

val df = spark.sql("select * from wangzhen_test_20231115_t1 order by id limit 2");
df.write.format("parquet").save("hfds://XXX/result.parquet");

spark.sql("set kyuubi.operation.language=sql");
select * from `parquet`.`hfds://XXX/result.parquet`;
```

```
== Physical Plan ==
Execute InsertIntoHadoopFsRelationCommand (5)
+- WriteFiles (4)
   +- TakeOrderedAndProject (3)
      +- * ColumnarToRow (2)
         +- Scan parquet spark_catalog.XXX.wangzhen_test_20231115_t1 (1)
```
image

image

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
wForget: > `engineSavePath` can be HDFS as the storage path, so it is not the same as tmp dir.

got it

lsm1:
> ```
> if (saveResultToPath) {
>   spark.sql(statement).write.format(format).save(resultPath)
>   result = spark.read.load(resultPath)
> } else {
>   result = spark.sql(statement)
> }
> ```
we still call `result.collect()` later, if the result is too large,we can not avoid driver OOM

wForget:
> we still call `result.collect()` later, if the result is too large,we can not avoid driver OOM

Can we combine incremental collection mode?

cxzl25:
> I did a simple test and the results were as expected

Maybe you can test the scenario of generating multiple files

wForget:
> Maybe you can test the scenario of generating multiple files

Do you mean a large data set or multiple tasks?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -278,4 +279,34 @@ object SparkDatasetHelper extends Logging {
...
+      case _ => resultMaxRows
+    }
+    lazy val stats = if (limit > 0) {
+      limit * EstimationUtils.getSizePerRow(
+        result.queryExecution.executedPlan.output)
+    } else {
+      result.queryExecution.optimizedPlan.stats.sizeInBytes
+    }
+    lazy val isSort = result.queryExecution.sparkPlan match {
+      case SortExec(_, global, _, _) if global => true
```
pan3793, line 296: is it binary-compatible across different spark versions? we must ensure that engine jar built against spark 3.4 also works on spark 3.1 ~ 3.5 runtimes

wForget:
> Maybe you can test the scenario of generating multiple files

The output seems to be in order even when outputting multiple files.

```
set spark.sql.files.maxRecordsPerFile=10000;

set kyuubi.operation.language=scala;

spark.range(0, 1000000, 1, numPartitions = 10)
  .selectExpr("id", "cast(id as string) as name")
  .createOrReplaceTempView("wangzhen_test_20231115_tmp1")

val df = spark.sql("select * from wangzhen_test_20231115_tmp1 order by id limit 100000");
df.write.format("parquet").save("hdfs://XXX/result");

spark.sql("set kyuubi.operation.language=sql");
select * from `parquet`.`hdfs://XXX/result`;
```

image

lsm1:
> > we still call `result.collect()` later, if the result is too large,we can not avoid driver OOM
> 
> Can we combine incremental collection mode?

When we use incremental collection mode, it may significantly impact performance.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
cxzl25, line 296:
```suggestion
      case s: SortExec => s.global
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  private def getOrcDeserializer(orcSchema: StructType, colId: Array[Int]): OrcDeserializer = {
+    try {
+      val cls = Class.forName("org.apache.spark.sql.execution.datasources.orc.OrcDeserializer")
+      val constructor = cls.getDeclaredConstructors.apply(0)
+      if (constructor.getParameterCount == 3) {
+        constructor.newInstance(new StructType, orcSchema, colId).asInstanceOf[OrcDeserializer]
+      } else {
+        constructor.newInstance(orcSchema, colId).asInstanceOf[OrcDeserializer]
+      }
```
cxzl25, line 77:
```suggestion
      if (SPARK_ENGINE_RUNTIME_VERSION >= "3.2") {
        // https://issues.apache.org/jira/browse/SPARK-34535
        DynConstructors.builder()
          .impl(
            classOf[OrcDeserializer],
            classOf[StructType],
            classOf[Array[Int]])
          .build[OrcDeserializer]()
          .newInstance(
            new StructType,
            orcSchema,
            colId)
      } else {
        DynConstructors.builder()
          .impl(
            classOf[OrcDeserializer],
            classOf[StructType],
            classOf[StructType],
            classOf[Array[Int]])
          .build[OrcDeserializer]()
          .newInstance(
            new StructType,
            orcSchema,
            colId)
      }
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  val iters = fileList.map(x => getOrcFileIterator(x))
+
+  var idx = 0
+
+  override def hasNext: Boolean = {
+    if (iters(idx).hasNext) {
+      true
+    } else {
+      iters(idx).close()
```
cxzl25, line 95: `ExecuteStatement` needs to implement the close method to close all readers of iters to prevent the client from not pulling all the results and causing reader leak.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1893,6 +1893,28 @@ object KyuubiConf {
       .intConf
       .createWithDefault(0)
 
+  val OPERATION_RESULT_SAVE_TO_FILE: ConfigEntry[Boolean] =
+    buildConf("kyuubi.operation.result.save.to.file")
```
pan3793, line 1897:
```
kyuubi.operation.result.saveToFile.enabled
kyuubi.operation.result.saveToFile.dir
kyuubi.operation.result.saveToFile.minSize
```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1893,6 +1893,28 @@ object KyuubiConf {
...
+      .doc("The switch for Spark query result save to hdfs file")
```
pan3793, line 1898: the dir is not limited to HDFS

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1893,6 +1893,28 @@ object KyuubiConf {
...
+      .stringConf
+      .createWithDefault("/tmp/kyuubi/tmp_kyuubi_result")
+
+  val OPERATION_RESULT_SAVE_TO_FILE_THRESHOLD: ConfigEntry[Int] =
+    buildConf("kyuubi.operation.result.save.to.file.threshold")
+      .doc("The threshold of Spark result save to hdfs file, default value is 200 MB")
+      .version("1.9.0")
+      .intConf
+      .checkValue(_ > 0, "must be positive value")
+      .createWithDefault(209715200)
```
pan3793, line 1916: possible to support human-readable strings? e.g. 200MiB, 200MB, 200m, 1G

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -278,4 +279,32 @@ object SparkDatasetHelper extends Logging {
     val executionId = sc.getLocalProperty(SQLExecution.EXECUTION_ID_KEY)
     SQLMetrics.postDriverMetricUpdates(sc, executionId, metrics.values.toSeq)
   }
+
+  def shouldSaveResultToHdfs(resultMaxRows: Int, threshold: Int, result: DataFrame): Boolean = {
```
pan3793, line 283:
```suggestion
  def shouldSaveResultToFs(resultMaxRows: Int, threshold: Int, result: DataFrame): Boolean = {
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -0,0 +1,150 @@
...
+        // https://issues.apache.org/jira/browse/SPARK-34535
+        DynConstructors.builder()
+          .impl(
+            classOf[OrcDeserializer],
+            classOf[StructType],
+            classOf[Array[Int]])
+          .build[OrcDeserializer]()
+          .newInstance(
+            orcSchema,
+            colId)
```
pan3793, line 89:
```suggestion
          .impl(classOf[OrcDeserializer], classOf[StructType], classOf[Array[Int]])
          .build[OrcDeserializer]()
          .newInstance(orcSchema,colId)
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
pan3793: @lsm1 let's add such information to the comments

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -158,6 +171,24 @@ class ExecuteStatement(
...
+      val sparkSave = getSessionConf(OPERATION_RESULT_SAVE_TO_FILE, spark)
+      lazy val threshold = getSessionConf(OPERATION_RESULT_SAVE_TO_FILE_THRESHOLD, spark)
+      if (hasResultSet && sparkSave && shouldSaveResultToHdfs(resultMaxRows, threshold, result)) {
+        val sessionId = session.handle.identifier.toString
+        val savePath = session.sessionManager.getConf.get(OPERATION_RESULT_SAVE_TO_FILE_PATH)
+        saveFileName = Some(s"$savePath/$engineId/$sessionId/$statementId")
+        val colName = range(0, result.schema.size).map(x => "col" + x)
+        if (resultMaxRows > 0) {
+          result.toDF(colName: _*).limit(resultMaxRows).write
+            .option("compression", "zstd").format("orc").save(saveFileName.get)
```
pan3793, line 183: IIRC, the zstd implementation of ORC is powered by aircompressor, I'm not sure if it's powerful as zstd-jni does. Anyway, it's internal implementation detail, we can change it latter if we have better solution

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
cfmcgrady: there is another known issue as I mentioned in the issue comment

> directly call `df.write` will introduce an extra shuffle for the outermost limit, and hurt performance

I think we should also add this known issue to the comment and create a new ticket to track this issue.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
cfmcgrady: should we also make the written format configurable?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
lsm1: `kyuubiConf` doesn't have `byteConf` like spark, I can submit another pr if need

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -0,0 +1,147 @@
...
+      idx < iters.size - 1
+    }
+  }
+
+  override def next(): OrcStruct = {
+    if (iters(idx).hasNext) {
+      iters(idx).next()
+    } else {
+      idx = idx + 1
+      iters(idx).next()
```
cfmcgrady, line 127: Is it possible that the `iters(idx)` can be empty?

lsm1: `hasNext()` make sure `idx` < iters.size

cfmcgrady: The index for `iters` has been adjusted from `idx` to `idx + 1` after the check condition `iters(idx).hasNext`.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -184,6 +186,12 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
       info("Session stopped due to shared level is Connection.")
       stopSession()
     }
+    if (conf.get(OPERATION_RESULT_SAVE_TO_FILE)) {
```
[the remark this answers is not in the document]
lsm1: There is no simple way to determine whether the session has executed `ExecuteStatement`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
lsm1: hasnext make sure `idx < iters.size - 1`, `idx + 1` would not out of bounds?

turboFei:
Some question:

I wonder that, If the result is order needed, if we save the result into files and then read from when client fetching result, the result return to users is not ordered as expected.

pan3793:
@turboFei from the context, I think the implementation already reserves the global order, @cxzl25 could you please clarify it?

pan3793 (approved):

lsm1:
> Some question:
> 
> I wonder that, If the result is order needed, if we save the result into files and then read from when client fetching result, the result returned to users is not ordered as expected.

1. spark save ordered result to multiple `part-X` files in the filesystem, in order of the keys
 https://github.com/lsm1/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/OrderedRDDFunctions.scala#L47
  ```
  org.apache.spark.rdd.OrderedRDDFunctions#sortByKey
   /**
  	* Sort the RDD by key, so that each partition contains a sorted range of the elements. Calling
  	* `collect` or `save` on the resulting RDD will return or output an ordered list of records
  	* (in the `save` case, they will be written to multiple `part-X` files in the filesystem, in
  	* order of the keys).  
  */
  ```
2. When fetchOrcStatement read file, it will be sorted by file name, so it will return ordered result

cfmcgrady (approved):

turboFei (approved):

cfmcgrady:
thanks all. merging to master(v1.9.0)

cxzl25 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
cxzl25: [SPARK-33978][SQL] Support ZSTD compression in ORC data source
https://issues.apache.org/jira/browse/SPARK-33978
Fix Version/s: 3.2.0

Maybe we need a configuration item, or the Spark version less than 3.2.0 is compressed with zlib.

3.1.1 bin/spark-shell
```scala
scala> spark.range(10).write.option("compression", "zstd").orc("/tmp/zstd")
java.lang.IllegalArgumentException: Codec [zstd] is not available. Available codecs are uncompressed, lzo, snappy, zlib, none.
```

pan3793: maybe it's time to have a discussion for dropping support of Spark 3.1

turboFei: 😅we are still using spark 3.1， It need time to migrate to new spark version

PR #5601, opened 2023-11-01
[KYUUBI #5566][FOLLOWUP] Check InternalSecurityAccessor is initialized only when `kyuubi.engine.security.enabled` is true

pan3793:

Internal REST client should work when `kyuubi.engine.security.enabled` is `true`/`false`.

The changes in https://github.com/apache/kyuubi/pull/5566 is not sufficient.

`KyuubiRestAuthenticationSuite` covers the `true` case

`BatchesV1ResourceSuite` and `BatchesV2ResourceSuite` cover the `false` case

No.


pan3793:
@cxzl25 I agree with you, such configuration reuse is not good

pan3793:
Thanks, merged to master/1.8

PR #5605, opened 2023-11-02
Add mock output table rule

wForget:

When debugging SQL tasks, we may need to replace the output table with a test table to avoid overwriting online data.
So I implemented a `MockOutputTable` rule to replace the output table with a mock table which ends with `_mock_${System.currentTimeMillis()}`.

No

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/MockOutputTable.scala
+++ b/extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/MockOutputTable.scala
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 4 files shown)


extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/MockOutputTable.scala
```diff
@@ -0,0 +1,103 @@
...
+    if (!conf.getConf(MOCK_OUTPUT_TABLE)) {
+      plan
+    } else {
+      val suffix = conf.getConf(MOCK_OUTPUT_TABLE_SUFFIX).getOrElse(defaultMockSuffix)
+      plan resolveOperators {
+        // InsertInto
+        case i @ InsertIntoStatement(r: HiveTableRelation, _, _, _, _, _, _)
+            if !isMockTable(r.tableMeta) =>
+          val newTable = mockTable(r.tableMeta, suffix, true)
+          i.copy(table = DDLUtils.readHiveTable(newTable))
```
ulysses-you, line 41: We should think about more. For example, the mocked table is not existed, how can we do insert into ?

PR #5616, opened 2023-11-03
Add submit SQL page in Kyuubi Web UI

zhaohehuhu:
Screenshot 2023-11-02 at 18 10 39


kyuubi-server/web-ui/src/views/editor/components/Editor.vue
```diff
@@ -0,0 +1,261 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+-->
+
+<template>
+  <div class="editor">
+    <el-space>
+      <el-select v-model="param.engineType" :placeholder="$t('engine_type')">
```
zwangsheng, line 22: We only support Spark SQL Engine now, so maybe we should disable this el-select.

kyuubi-server/web-ui/src/views/editor/components/Editor.vue
```diff
@@ -0,0 +1,261 @@
...
+    log
+  } from '@/api/editor'
+  import type { IResponse, ISqlResult, IFields, ILog } from './types'
+
+  const { t } = useI18n()
+  const param = reactive({
+    engineType: 'SPARK_SQL'
+  })
+  const limit = ref(10)
+  const limits = ref([10, 50, 100])
```
zwangsheng, line 99: We don't need to bind this limits two-way data binding, simple constants should suffice.

kyuubi-server/web-ui/src/views/editor/components/Editor.vue
```diff
@@ -0,0 +1,261 @@
...
+      })
+      .finally(() => {
+        resultLoading.value = false
+      })
+
+    sqlLog.value = await log(runSqlResponse.identifier)
+      .then((res: ILog) => {
+        return res?.logRowSet?.join('\r\n')
+      })
+      .catch(() => '')
```
zwangsheng, line 179: Should catch error

kyuubi-server/web-ui/src/views/editor/components/Editor.vue
```diff
@@ -0,0 +1,261 @@
...
+        sqlResult.value = result[0]?.rows?.map((row: IFields) => {
+          const map: { [key: string]: any } = {}
+          row.fields?.forEach(({ value }: ISqlResult, index: number) => {
+            map[result[1].columns[index]?.columnName] = value
+          })
+          return map
+        })
+      })
+      .catch(() => {
+        sqlResult.value = []
```
zwangsheng, line 169: Should catch error

kyuubi-server/web-ui/src/locales/en_US/index.ts
```diff
@@ -37,21 +37,25 @@ export default {
   engine_ui: 'Engine UI',
   failure_reason: 'Failure Reason',
   session_properties: 'Session Properties',
+  no_data: 'No data yet',
```
zwangsheng, line 40: Please correct me if this is incorrect, we return no_data when query success but without empty data, so this should be 'No data'?

kyuubi-server/web-ui/src/components/monaco-editor/type.ts
```diff
@@ -53,10 +53,10 @@ export const editorProps = {
...
   },
   theme: {
-    type: String as PropType<Theme>,
-    validator(value: string): boolean {
-      return ['vs', 'vs-dark'].includes(value)
-    },
+    type: String as PropType<any>,
+    // validator(value: string): boolean {
+    //   return ['vs', 'vs-dark'].includes(value)
+    // },
```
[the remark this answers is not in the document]
zhaohehuhu: Sure

kyuubi-server/web-ui/src/api/editor/index.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import request from '@/utils/request'
+
+export function openSession(data: any): any {
```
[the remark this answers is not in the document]
zhaohehuhu: Got it

kyuubi-server/web-ui/src/api/server/index.ts
```diff
@@ -17,7 +17,7 @@
 
 import request from '@/utils/request'
 
-export function getAllServer() {
+export function getAllServer(): any {
```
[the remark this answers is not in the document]
zhaohehuhu: Got it

kyuubi-server/web-ui/src/router/index.ts
```diff
@@ -20,7 +20,7 @@ import overviewRoutes from './overview'
 import managementRoutes from './management'
 import detailRoutes from './detail'
 import swaggerRoutes from './swagger'
-import labRoutes from './lab'
+import labRoutes from './editor'
```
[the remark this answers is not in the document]
zhaohehuhu: Got it

kyuubi-server/web-ui/src/views/editor/components/Editor.vue
```diff
@@ -0,0 +1,273 @@
...
+        sqlResult.value = result[0]?.rows?.map((row: IFields) => {
+          const map: { [key: string]: any } = {}
+          row.fields?.forEach(({ value }: ISqlResult, index: number) => {
+            map[result[1].columns[index]?.columnName] = value
+          })
+          return map
+        })
+      })
+      .catch(() => {
+        ElMessage({
```
zwangsheng, line 171: Use `errorCatch` instead

zwangsheng (approved):
LGTM with little comments, also PTAL @pan3793

kyuubi-server/web-ui/src/views/editor/components/Editor.vue
tongwl: Hi, thanks for review. errorCatch has its own logic like set sqlResult and sqlLog to empty, not very suitable to use it here.

zwangsheng: Thanks for your comment. In my opinion, when we run this SQL and encounter an error, we already have both set the result to an empty array and used the el-message component to notify the user. However, should we also try to obtain the SQL log from the error message if it is available?

tongwl: @zwangsheng 
Certainly, got your point, what you think is the api error message might be helpful to users, we also need it!

If we want to display all the error info for the APIs through the log, there might be some issues here:
The current design for executing SQL is as follows:
* First, we call the openSession API, which returns the sessionIdentifier.
* Then insert the returned sessionIdentifier into the second API, runSql, which returns the runSqlIdentifier.
These two steps must be executed synchronously because they are dependent on each other.

After the above steps, there are two more steps remaining:
* Retrieving the SQL's return results (here, there are two APIs, one for getting the table's field names, and one for getting the table's field values).
* Getting the SQL log.
Here, considering the user experience, we don't need to make the APIs execute synchronously. That is, getting the SQL results and getting the SQL log will be sent asynchronously to the service, as the SQL log API doesn't need to wait for the SQL result, and the SQL result API doesn't need to wait for the SQL log.

Now, here's the issue:
Because the last three APIs are asynchronous, if we only overwrite their error information into the log, it might provide a poor user experience. For example, suppose the getSqlResult API returns an error in 30 seconds (it could be two errors), and it logs the error, then user is watching the log, but suddenly the log gets overwritten because at 45 seconds, the getSqlLog also sends logs. Appending the log to the API isn't good enough either because the log display might not be very clear.

**If this error log is useful for user, can we consider this design below?**
* The log section will only display query log information. 
* Error messages will be placed in the error box so that users can clearly see the error content and list error messages from multiple APIs as a reference.  (see screenshot)

Thanks.

tongwl: image

zwangsheng: Thanks for your long answer, SGTM, it's fine for me to see server query log in log item and other api error log in el-message.

tongwl: image

tongwl: image

tongwl: Hi @zwangsheng 
Has added new error messages for single or multiple errors, please help review, thanks!

zwangsheng (approved):
LGTM, cc @pan3793

kyuubi-server/web-ui/src/views/editor/components/Editor.vue
pan3793: @tongwl could you put the Run button at the head of the line? and use a light color for it?

pan3793: Also, please use `SQL` instead of `Sql` in the message.

tongwl: Hi @pan3793 , sure, got!

tongwl: image
image
Hi @pan3793 , please help review again!

tongwl: image

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #5633, opened 2023-11-06
[K8S][HELM]Save ETCD certificate in a Kubernetes Secret

zhaohehuhu:
To save ETCD certificates in a Kubernetes Secret using a Helm chart,  developers can easily build the Kyuubi with ETCD  in K8S.

### DIFF
--- a/charts/kyuubi/templates/kyuubi-etcd-secret.yaml
+++ b/charts/kyuubi/templates/kyuubi-etcd-secret.yaml
@@ -1,0 +1,32 @@
+{{/*
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/}}
+
+{{- if .Values.ha.etcd.ssl.enabled }}
+{{- $ca := .Files.Get .Values.ha.etcd.ssl.ca.path | b64enc | quote }}
+{{- $client_certificate := .Files.Get .Values.ha.etcd.ssl.client.certificate.path | b64enc | quote }}
+{{- $client_key := .Files.Get .Values.ha.etcd.ssl.client.key.path | b64enc | quote }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    {{- include "kyuubi.labels" . | nindent 4 }}
+data:
+  ca.crt: {{ $ca }}
+  client.crt: {{ $client_certificate }}
+  client_key.pem: {{ $client_key }}
+{{- end }}
--- a/charts/kyuubi/templates/kyuubi-configmap.yaml
+++ b/charts/kyuubi/templates/kyuubi-configmap.yaml
@@ -41,6 +41,15 @@ data:
     kyuubi.metrics.enabled={{ .Values.monitoring.prometheus.enabled }}
     kyuubi.metrics.reporters={{ .Values.metricsReporters }}
 
+    # Kyuubi ha
+    {{- if .Values.ha.etcd.ssl.enabled }}
+    kyuubi.ha.client.class=org.apache.kyuubi.ha.client.etcd.EtcdDiscoveryClient
+    kyuubi.ha.etcd.ssl.enabled=true
+    kyuubi.ha.etcd.ssl.ca.path={{ .Values.ha.etcd.ssl.ca.mountPath }}
+    kyuubi.ha.etcd.ssl.client.certificate.path={{ .Values.ha.etcd.ssl.client.certificate.mountPath }}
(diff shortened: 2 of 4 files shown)


zhaohehuhu:
plz help review it @pan3793

charts/kyuubi/templates/kyuubi-etcd-secret.yaml
```diff
@@ -0,0 +1,32 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/}}
+
+{{- if .Values.ha.etcd.ssl.enabled }}
+{{- $ca := .Files.Get .Values.ha.etcd.ssl.ca.path | b64enc | quote }}
+{{- $client_certificate := .Files.Get .Values.ha.etcd.ssl.client.certificate.path | b64enc | quote }}
+{{- $client_key := .Files.Get .Values.ha.etcd.ssl.client.key.path | b64enc | quote }}
```
pan3793, line 21: According to https://helm.sh/docs/chart_template_guide/accessing_files/

> Some files cannot be accessed through the `.Files` object, usually for security reasons.
> - Files in `templates/` cannot be accessed.
> - Files excluded using `.helmignore` cannot be accessed.
> - Files outside of a helm application [subchart](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), including those of the parent, cannot be accessed

This approach won't work if we publish the Helm Chart to repo and users use `helm install <repo>/<kyuubi-chart>` directly.

To be clear, I am not against this approach, actually, I use the same approach to handle Hadoop configuration files internally. But I still hesitate to use this approach upstream.

See also my previous comment: https://github.com/apache/kyuubi/pull/5281#issuecomment-1769827545

zhaohehuhu: I see

pan3793: cc @dnskr, do you have any suggestion?

dnskr: Am I right that the implementation assumes the following?
 - User downloads the chart to local filesystem.
 - User puts sensitive files to the chart directory on the filesystem.
 - User changes `.Values.ha.etcd.ssl.ca.path`, `.Values.ha.etcd.ssl.client.certificate.path` and `.Values.ha.etcd.ssl.client.key.path` values to point them to the files.
 - User executes helm install <local-chart-directory>/kyuubi.


If so, the approach won't work in the following two cases:
 1. [Default] If user executes `helm install <kyuubi-chart-repo>/<kyuubi-chart>` as @pan3793 mentioned
 2. [GitOps] If user uses AgroCD or similar CI/CD tool.

Moreover, I would prefer to not create Secrets in the chart at all or at least I would not recommend it for production deployments.
I would suggest to create the `Secret` from outside the chart release flow (manually or using secret management tools) and refer to it. I.e. I propose the following structure in `values.yaml` file with expected Secret example
```yaml
ha:
  etcd:
    ssl:
      enabled: false
      # apiVersion: v1
      # kind: Secret
      # metadata:
      #   name: kyuubi-ha-etcd-ssl
      # data:
      #   ca.crt: <base64 encoded ca.crt file content>
      #   client.crt: <base64 encoded client.crt file content>
      #   client_key.pem: <base64 encoded client_key.pem file content>
      secretName: kyuubi-ha-etcd-ssl
      mountPath: /etc/kyuubi/ssl
```
_(not sure `/etc/kyuubi/ssl` is the best path for those files)_

zhaohehuhu: Thanks @dnskr

dnskr: > To be clear, I am not against this approach, actually, I use the same approach to handle Hadoop configuration files internally. But I still hesitate to use this approach upstream.
> 
> See also my previous comment: [#5281 (comment)](https://github.com/apache/kyuubi/pull/5281#issuecomment-1769827545)

I found some information regarding the approach and your previous comment. I hope it's fine to share it here.

Right, it's impossible to reference any files outside the chart with `.Files` because of security reasons. However, users [can use '--set-file' to set individual values from a file](https://helm.sh/docs/helm/helm_install/#:~:text=can%20use%20%27%2D%2Dset%2Dfile%27%20to%20set%20individual%20values%20from%20a%20file) when installing the chart.

For instance, the following command (executed from one level above kyuubi directory) sets `kyuubi/conf/log4j2.xml.template` content to `kyuubiConf.log4j2` property and prints out `templates/kyuubi-configmap.yaml`:
```shell
helm template kyuubi kyuubi/charts/kyuubi --set-file kyuubiConf.log4j2=kyuubi/conf/log4j2.xml.template --show-only templates/kyuubi-configmap.yaml
---
# Source: kyuubi/templates/kyuubi-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kyuubi
  ...
data:
  log4j2.xml: |
    <?xml version="1.0" encoding="UTF-8"?>
    ...
    <Configuration status="INFO">
        <Properties>
            ...
        </Properties>
        <Appenders>
            ...
        </Appenders>
        <Loggers>
            ...
        </Loggers>
    </Configuration>
```

So, users can store heavy configs (`*-site.xml` and others) in dedicated places outside the chart and access those config files when they install the chart using `--set-file`. 
Does it work for your use cases?

dnskr: @pan3793 I forgot to mention you in the message above

pan3793: nice! this feature looks useful.

@dnskr, thank you for information

PR #5646, opened 2023-11-07
[BUILD] Support copying custom jars to server jars in build/dist script

bowenliang123:

In practical usage, we often maintain and have some custom dependency library JAR files for logging, authentication, advisors  and etc  that need to be packaged into the distribution package. This PR supports copying all JAR files from the path `./custom-server-jars` to the 'jars' directory of the distribution, which helps automate the packaging process.

No.


pan3793:
This does not like a common practice, especially, why those jars are expected to exist in a folder before packing? are you plan to track those jars by Git internally?

bowenliang123:
> This does not like a common practice, especially, why those jars are expected to exist in a folder before packaging? do you plan to track those jars by Git internally?

It is reasonable to use pre-built library jars for different purposes. Eg. JDBC drivers for  authentication , custom server plugin, logger jars and etc.

PR #5678, opened 2023-11-13
[KYUUBI #5677][AUTHZ] Typeof expression miss column information

### LINKED ISSUE #5677: [Bug] Typeof expression miss column information
### Code of Conduct

### Search before asking

### Describe the bug

Typeof expression miss column information

### Affects Version(s)

master, 1.8.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

AngersZhuuuu:
To close #5677 
Typeof expression miss column information, in this pr we add a place holder to support this.

For sql 
```
SELECT typeof(id), typeof(age) FROM default.t1 LIMIT 1
```

Without this pr, typeof after optimizer passed to PrivilegeBuild is just an attribute, miss the origin child expression info.
```
GlobalLimit 21
+- LocalLimit 21
   +- Project [int AS typeof(id)#27, int AS typeof(age)#28]
      +- Relation default.t1[id#18,age#19,part#20] parquet
```

When handle this plan's project list, it's an attribute and miss expression when expression result is a constant value, then we can't extract the `TypeOf`'s child expression.
image

This is caused by `TypeOf`  expression is foldable
image

It will be convert to a constant value after spark optimizer, then cause we miss the child expression, then can't extract columns by `collectLeaves`

In this pr we wrap the `TypeOf` by a non-foldable expression placeholder then we can get the expression contains column of table when `mergeProjection`
After this pr, the plan is like below
```
GlobalLimit 21
+- LocalLimit 21
   +- Project [typeofplaceholder(id#21) AS typeof(id)#30, typeofplaceholder(day#23) AS typeof(day)#31]
      +- HiveTableRelation [`default`.`table1`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#21, scope#22, day#23], Partition Cols: []]

```

No

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateTypeOf.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateTypeOf.scala
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.rule
+
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+
+import org.apache.kyuubi.plugin.spark.authz.rule.typeof.TypeOfPlaceHolder
+
+class RuleEliminateTypeOf extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    plan.transformUp { case p =>
+      p.transformExpressionsUp {
+        case toph: TypeOfPlaceHolder => toph.expr
+      }
+    }
+  }
+}
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateTypeOf.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateTypeOf.scala
@@ -17,6 +17,7 @@
 
 package org.apache.kyuubi.plugin.spark.authz.rule
 
+import org.apache.spark.sql.catalyst.expressions.TypeOf
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
 
@@ -26,7 +27,7 @@ class RuleEliminateTypeOf extends Rule[LogicalPlan] {
   override def apply(plan: LogicalPlan): LogicalPlan = {
     plan.transformUp { case p =>
       p.transformExpressionsUp {
-        case toph: TypeOfPlaceHolder => toph.expr
+        case toph: TypeOfPlaceHolder => TypeOf(toph.expr)
       }
     }
   }
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateTypeOf.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateTypeOf.scala
@@ -21,7 +21,7 @@ import org.apache.spark.sql.catalyst.expressions.TypeOf
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
 
-import org.apache.kyuubi.plugin.spark.authz.rule.typeof.TypeOfPlaceHolder
+import org.apache.kyuubi.plugin.spark.authz.rule.expression.TypeOfPlaceHolder
 
 class RuleEliminateTypeOf extends Rule[LogicalPlan] {
   override def apply(plan: LogicalPlan): LogicalPlan = {
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
@@ -1241,6 +1241,7 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
     }
   }
 
+<<<<<<< Updated upstream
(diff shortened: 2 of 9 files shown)


AngersZhuuuu:
ping @yaooqinn

yaooqinn:
What is the problem here?

AngersZhuuuu:
> What is the problem here?

Without this pr, it will check the privilege of all table, not column

yaooqinn:
I see the id, age field in the project. Why not pull them out?

AngersZhuuuu:
> I see the id, age field in the project. Why not pull them out?

Yea

AngersZhuuuu:
Without this pr the UT's result is  as below cc @yaooqinn 
```
"Permission denied: user [someone] does not have [select] privilege on [default/table1]" did not contain "does not have [select] privilege on [default/table1/id,default/table1/day]"
ScalaTestFailureLocation: org.apache.kyuubi.plugin.spark.authz.ranger.HiveCatalogRangerSparkExtensionSuite at (RangerSparkExtensionSuite.scala:1316)
org.scalatest.exceptions.TestFailedException: "Permission denied: user [someone] does not have [select] privilege on [default/table1]" did not contain "does not have [select] privilege on [default/table1/id,default/table1/day]"
	at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)
	at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)

```

yaooqinn:
> > I see the id, age field in the project. Why not pull them out?
> 
> Yea 截屏2023-11-13 下午2 01 10


Why not pull them out?

AngersZhuuuu:
> > > I see the id, age field in the project. Why not pull them out?
> > 
> > 
> > Yea 截屏2023-11-13 下午2 01 10
> 
> Why not pull them out?

The name and expr id is not same

yaooqinn:
Why is typeof special, or how many are they?

AngersZhuuuu:
> Why is typeof special, or how many are they?

Since typeof's result only needs the expression's datatype, it doesn't need real value.  
Only see one such case:
1. need expression
2. Result is constant

yaooqinn:
> Since typeof's result only needs the expression's datatype, it doesn't need real value.


Shall we bypass it?

AngersZhuuuu:
> > Since typeof's result only needs the expression's datatype, it doesn't need real value.
> 
> Shall we bypass it?

But it need table's read privilege, can't just ignore, better extract the correct column?

yaooqinn:
It's much more like `explain select a from src` to me

AngersZhuuuu:
> It's much more like `explain select a from src` to me

But only when the query's result is empty, then this sql have the output. If there is no row, show an empty result

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -1296,4 +1296,37 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
...
+          doAs(
+            someone,
+            sql(
+              s"""
+                 |SELECT
+                 |typeof(cast(id as string)),
+                 |typeof(substring(day, 1, 3))
+                 |FROM $db1.$table1""".stripMargin).show()))(
+          s"does not have [select] privilege on [$db1/$table1/id,$db1/$table1/day]")
+        doAs(admin, sql(s"SELECT typeof(id), typeof(day) FROM $db1.$table1").show())
```
yaooqinn, line 1328: check values

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -1296,4 +1296,37 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
...
+          sql(
+            s"""
+               |CREATE TABLE IF NOT EXISTS $db1.$table1(
+               |id int,
+               |scope int,
+               |day string)
+               |""".stripMargin))
+        doAs(admin, sql(s"INSERT INTO $db1.$table1 SELECT 1, 2, 'TONY'"))
+        interceptContains[AccessControlException](
+          doAs(someone, sql(s"SELECT typeof(id), typeof(day) FROM $db1.$table1").show()))(
```
yaooqinn, line 1316: do not use show

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
yaooqinn: SELECT typeof(typeof)

typeof typeof(id + age)

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
AngersZhuuuu: Done

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
AngersZhuuuu: How about current?

bowenliang123:
Could you add the example for the plan in either the issue or the pr description to show the problem and the plan?

AngersZhuuuu:
> Could you add the example for the plan in either the issue or the pr description to show the problem and the plan?

How about current?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateTypeOf.scala
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.rule
+
+import org.apache.spark.sql.catalyst.expressions.TypeOf
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+
+import org.apache.kyuubi.plugin.spark.authz.rule.typeof.TypeOfPlaceHolder
```
yaooqinn, line 24: typeof -> expression

AngersZhuuuu: > typeof -> expression

Done

AngersZhuuuu:
Any more suggestion?

yaooqinn (approved):

bowenliang123:
> > Could you add the example for the plan in either the issue or the pr description to show the problem and the plan?
> 
> How about current?

Sorry, it's still too difficult for me. I'm out of the context.
Show the SQL or situation first, then the plan, and the suggested solution.

bowenliang123:
If you don't have time next time, there's no need to ping me. Lately, I haven't been keeping up with your series of PRs. They always get merged quickly, and my suggestions are being skipped. Not enough information for me to understand.  If the Kyuubi community is okay with this 'no-explanation-just-go' approach, I don't have any objections. I attribute all of my willingness to my own shortcomings in abilities.

AngersZhuuuu:
> > > Could you add the example for the plan in either the issue or the pr description to show the problem and the plan?
> > 
> > 
> > How about current?
> 
> Sorry, it's still too difficult for me. I'm out of the context. Show the SQL or situation first, then the plan, and the suggested solution.

Add explain about root cause and solution.

Issue #5680, opened 2023-11-13
[TASK][EASY] Optimize kyuubi spark engine driver pod name generation


zwangsheng:
### What's the level of this task?

EASY

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [X] I have sufficient expertise on this task, and I volunteer to be a mentor of this task to guide contributors through the task.

### Skill requirements

- Scala
- Spark Config about Spark Driver Pod Name and Spark App Name

### Background and Goals

Background:
In default case, when kyuubi submit spark sql engine on kubernetes, will generate spark driver pod name with `kyuubi--${app_name}-${engine_ref_id}-driver`. And app_name will be `kyuubi_${shareLevel}_${engineType}_${appUser}_${engineRefId}` if not set by user. In result, we may get spark driver pod name with `kyuubi-kyuubi-${shareLevel}-${engineType}-${appUser}-${engineRefId}-${engineRefId}-driver`.

We were hoping for a more concise name, such as `kyuubi-${shareLevel}-${engineType}-${appUser}-${engineRefId}-driver`.

### Implementation steps

1. modify logic in `SparkProcessBuilder#appendPodNameConf` method. 

### Additional context

Introduction of [2023 Kyuubi Code Contribution Program](https://github.com/apache/kyuubi/issues/5357)

zhaohehuhu:
someone is doing this task, right ? @zwangsheng

zwangsheng:
> someone is doing this task, right ? @zwangsheng

Yes, there is PR #5695 to fix this

liaoyt:
does this still in progress?
i'm facing the issue that the pod name is too long, and hoping this issue is fixed. 
<img width="1518" alt="image" src="https://github.com/apache/kyuubi/assets/12195059/605b6346-8f35-497f-bfed-c55c8c0c67b5">

PR #5695, opened 2023-11-15
[KYUUBI #5680] Optimize kyuubi spark engine driver&executor pod name generation

Ocean22:
[KYUUBI #5680] Optimize kyuubi spark engine driver and executor pod name generation

**Background**:
1) In default case, when kyuubi submit spark sql engine on kubernetes, will generate spark driver pod name with "kyuubi-${app_name}-${engine_ref_id}-driver". 
2) And app_name will be "kyuubi_${shareLevel}_${engineType}_${appUser}_${engineRefId}" if not set by user. 
3) In result, we may get spark driver pod name with "kyuubi-kyuubi-${shareLevel}-${engineType}-${appUser}-${engineRefId}-${engineRefId}-driver".
4) We were hoping for a more concise and readable name, such as "kyuubi-${shareLevel}-${engineType}-${appUser}-${app_name}-${engineRefId}-driver".

**1) Before modification：**
Left is unset `spark.app.name`, and right is set to “ocean”
image

**2) Modify the code**
image

**3) Build module and get the jar**
image

**4) Build a new images**
image

**5) After modification:**
image

No

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -146,7 +146,7 @@ private[kyuubi] class EngineRef(
         s"${shareLevel}_${engineType}_${appUser}_${appKey}"
       } else {
         s"${shareLevel}_${engineType}_${appUser}"
-      }/*  */
+      }
     commonNamePrefix
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -136,19 +136,19 @@ private[kyuubi] class EngineRef(
   }
 
   /**
- * Generate app key, enrich `spark.app.name` content
- * Example: connection_spark-sql_root_myappname
- */
-@VisibleForTesting
-private[kyuubi] def generateAppKey(appKey: String): String = {
-  var commonNamePrefix: String = ""
-  if (appKey != null){
-    commonNamePrefix = s"${shareLevel}_${engineType}_${appUser}_${appKey}"
-  }else {
-    commonNamePrefix = s"${shareLevel}_${engineType}_${appUser}"
+   * Generate app key, enrich `spark.app.name` content
+   * Example: connection_spark-sql_root_myappname
+   */
+  @VisibleForTesting
+  private[kyuubi] def generateAppKey(appKey: String): String = {
+    val commonNamePrefix: String =
+      if (appKey != null) {
+        s"${shareLevel}_${engineType}_${appUser}_${appKey}"
+      } else {
+        s"${shareLevel}_${engineType}_${appUser}"
+      }/*  */
+    commonNamePrefix
   }
-  commonNamePrefix
-}
 
   /**
    * The EngineSpace used to expose itself to the KyuubiServers in `serverSpace`
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -136,6 +136,21 @@ private[kyuubi] class EngineRef(
   }
 
   /**
+ * Generate app key, enrich `spark.app.name` content
+ * Example: connection_spark-sql_root_myappname
+ */
+@VisibleForTesting
+private[kyuubi] def generateAppKey(appKey: String): String = {
+  var commonNamePrefix: String = ""
+  if (appKey != null){
+    commonNamePrefix = s"${shareLevel}_${engineType}_${appUser}_${appKey}"
+  }else {
+    commonNamePrefix = s"${shareLevel}_${engineType}_${appUser}"
+  }
+  commonNamePrefix
+}
+
+  /**
    * The EngineSpace used to expose itself to the KyuubiServers in `serverSpace`
    *
    * For `CONNECTION` share level:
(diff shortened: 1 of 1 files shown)


yaooqinn:
Hi @Ocean22, thanks for the PR.

There is also an original purpose for `spark.app.name` generation I want to share. In early versions of kyuubi, I used(just an attemption) it as a unique key for retrieval. However, we do not rely on it for this purpose.

So, since we are now touching this part, I suggest we make it as simple as possible. for example, the appUser is not necessary here as both YARN UI or Spark History UI display with a USER field, uuid is not necessary as it is also in the app tags,

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -187,7 +202,7 @@ private[kyuubi] class EngineRef(
     conf.set(KYUUBI_ENGINE_SUBMIT_TIME_KEY, String.valueOf(started))
     builder = engineType match {
       case SPARK_SQL =>
-        conf.setIfMissing(SparkProcessBuilder.APP_KEY, defaultEngineName)
+        conf.set(SparkProcessBuilder.APP_KEY, generateAppKey(conf.getOption(SparkProcessBuilder.APP_KEY).getOrElse(null)))
```
zwangsheng, line 205: IMO, if user set spark app name, we should not over write this, so we should keep setIfMissing here.

zwangsheng (commented):
We can just simply changed appendPodName logic in `SparkProcessBuilder#appendPodNameConf`

1. remove `kyuubi` prefix
2. only append `engineRefId`, when `appName` not contains it.

Ocean22:
Thanks for guidance,I will continue to improve it when I have free time.

Ocean22:
@yaooqinn @zwangsheng Hi, following our previous discussion, I have changed a way to handle this issue.

Ocean22:
> @yaooqinn @zwangsheng Hi, following our previous discussion, I have changed a way to handle this issue.

Sorry, it still need some else change,I still need to do more testing

Ocean22:
> > @yaooqinn @zwangsheng Hi, following our previous discussion, I have changed a way to handle this issue.
> 
> Sorry, it still need some else change,I still need to do more testing

Now is OK, and this photo is the result that I test.

kyuubi-server/src/main/scala/org/apache/kyuubi/util/KubernetesUtils.scala
```diff
@@ -137,6 +137,8 @@ object KubernetesUtils extends Logging {
     lazy val resolvedResourceName = s"kyuubi-${getResourceNamePrefix(appName, engineRefId)}-driver"
     if (forciblyRewrite || resolvedResourceName.length > DRIVER_POD_NAME_MAX_LENGTH) {
       s"kyuubi-$engineRefId-driver"
+    } else if (appName.contains(engineRefId)){
+      getResourceNamePrefix(appName, "driver")
     } else {
       resolvedResourceName
     }
```
zwangsheng, line 144:
```suggestion
    val resourceNamePrefix = if (appName.contains(engineRefId)) {
      getResourceNamePrefix(appName, "")
    } else {
      getResourceNamePrefix(appName, engineRefId)
    }
    val resolvedResourceName = if (resourceNamePrefix.startsWith("kyuubi")) {
      s"$resourceNamePrefix-driver"
    } else {
      s"kyuubi-$resourceNamePrefix-driver"
    }
    if (resolvedResourceName.length <= DRIVER_POD_NAME_MAX_LENGTH) {
      resolvedResourceName
    } else {
      s"kyuubi-$engineRefId-driver"
    }
```

zwangsheng (commented):
Sorry for late review, we should determine if `resolvedResourceName ` carries engine ref id before determining if it's too long.

kyuubi-server/src/main/scala/org/apache/kyuubi/util/KubernetesUtils.scala
Ocean22: > 
Yes,you are right.And I tested with your code and it passed, but about `resourceNamePrefix. startsWith ("kyuubi")`, if  `spark. app. name` is set to the name starting with kyuubi, it will have the effect of the arrow in the photo. So should it be changed to `resourceNamePrefix. startsWith ("kyuubi-")`?
企业微信截图_17004520968190

zwangsheng (approved):
LGTM, also FYI @pan3793

pan3793 (approved):
LGTM, though @yaooqinn's suggestion is not applied, we can do it in the future

pan3793:
image
there still is a minor issue on the final result

Ocean22:
> image there still is a minor issue on the final result

Sorry,I thought this method would avoid this situation.Let me improve it
image

pan3793:
how about this?
```patch
- private def getResourceNamePrefix(appName: String, engineRefId: String): String = {
-   s"$appName-$engineRefId"
+ private def getResourceNamePrefix(appName: String, engineRefId: Option[String]): String = {
+   engineRefId.map(refId => "$appName-$refId").getOrElse(appName)
```

Ocean22:
> how about this?
> 
> ```diff
> - private def getResourceNamePrefix(appName: String, engineRefId: String): String = {
> -   s"$appName-$engineRefId"
> + private def getResourceNamePrefix(appName: String, engineRefId: Option[String]): String = {
> +   engineRefId.map(refId => s"$appName-$refId").getOrElse(appName)
> ```
Is it right what I modified?
image

Ocean22:
> @Ocean22 sorry I missed the message, it's correct, please push it

Good,I will push it later.

pan3793 (approved):

pan3793:
@Ocean22 could you please provide a Public Name and email (which should be linked to your GitHub account), otherwise your contribution will not be counted by GitHub

pan3793:
Thanks, merged to master/1.8

PR #5696, opened 2023-11-15
[REST] Return remote session id and remote operation id

turboFei:

 Not all the engine sessionIds/operationIds are unified between kyuubi server and kyuubi engines, so we need to expose  these remote details to provide more insights.

No.

### DIFF
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/SessionData.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/SessionData.java
@@ -25,6 +25,7 @@ import org.apache.commons.lang3.builder.ToStringStyle;
 
 public class SessionData {
   private String identifier;
+  private String remoteId;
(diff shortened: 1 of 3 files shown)


pan3793 (approved):

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/SessionData.java
```diff
@@ -25,6 +25,7 @@
 
 public class SessionData {
   private String identifier;
+  private String remoteSessionId;
```
pan3793, line 28:
```suggestion
  private String remoteId;
```

pan3793: let's unify the naming style

cxzl25 (approved):

turboFei:
thanks all, merged to master and branch-1.8

PR #5703, opened 2023-11-15
[KYUUBI #5700][AUTHZ] DropTableCommand should check it's table or view

### LINKED ISSUE #5700: [Bug] DropTableCommand should check it's table or view
### Code of Conduct

### Search before asking

### Describe the bug

When it's a DropTableCommand, we did't check if it's table a permanent view, both use drop table

### Affects Version(s)

master, 1.8.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

AngersZhuuuu:
To close #5700 
Now all  DropTableCommand return DROPTABLE, but DROPVIEW also use this command, we should check it's table or view or temp view

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
@@ -20,6 +20,7 @@ package org.apache.kyuubi.plugin.spark.authz
 import scala.collection.mutable.ArrayBuffer
 
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.expressions.{Expression, NamedExpression}
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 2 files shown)


AngersZhuuuu:
ping @yaooqinn

yaooqinn:
Changes look a bit ugly:), and not work for 3.5 or higher

AngersZhuuuu:
> Changes look a bit ugly:), and not work for 3.5 or higher

Yea, spark 3.5 split the command.

AngersZhuuuu:
> Changes look a bit ugly:), and not work for 3.5 or higher

If not handle here, we may need to add a OperationType extractor to do this....but add a new extractor is too heavey...

yaooqinn:
Can we have a e2e test about what's failing for now?

AngersZhuuuu:
> Can we have a e2e test about what's failing for now?

In current code, for ranger, DROPTABLE and DROPVIEW have same accessType, won't cause failing case.
截屏2023-11-16 上午10 53 27


But maybe for some other auth system we need to divide these two type of command, such as our auth system.

yaooqinn:
> But maybe for some other auth system we need to divide these two type of command, such as our auth system.

Is there some thing wrong with your auth system. Here we only define DROP, not define whether the  underlying object is a table or a view,

AngersZhuuuu:
> > But maybe for some other auth system we need to divide these two type of command, such as our auth system.
> 
> Is there some thing wrong with your auth system. Here we only define DROP, not define whether the underlying object is a table or a view,

Just  make a audit info that it's a drop view or table.

yaooqinn:
Why audit log has to be related to OperationType

AngersZhuuuu:
> Why audit log has to be related to OperationType

Hmm, our platform's requirement...

yaooqinn:
The audit module has nothing to do with OperationType. It's not a standard protocol and changeable in the future

AngersZhuuuu:
> The audit module has nothing to do with OperationType. It's not a standard protocol and changeable in the future

oK

PR #5711, opened 2023-11-16
[K8S] Support to build application state and error from pod container

turboFei:
# :mag: Description

This pr is to support to build the application state from pod container instead of pod.

The use case is that:
For spark on k8s, the pod contains two containers.
- one is the spark driver main container
- the another one is a sidecar container

If the spark driver main container completed but the sidecar container not, the batch application state is always RUNNING as the pod state is `NotReady`. 

It is not expected.

So in this pr, I want to introduce a new option config to support to build application state from the pod container.

## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklists
## 📝 Author Self Checklist

- [x] My changes generate no new warnings

- [x] Pull request title is okay.
- [x] No license issues.
- [x] Milestone correctly set?
- [ ] Test coverage is ok
- [ ] Assignees are selected.

**Be nice. Be informative.**


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -274,16 +293,66 @@ object KubernetesApplicationOperation extends Logging {
 
   def toLabel(tag: String): String = s"label: $LABEL_KYUUBI_UNIQUE_KEY=$tag"
 
-  def toApplicationState(state: String): ApplicationState = state match {
-    // https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2396
```
turboFei, line 278: remove this one as it is out dated

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1231,6 +1231,21 @@ object KyuubiConf {
...
+  val KUBERNETES_APPLICATION_STATE_FROM_CONTAINER: ConfigEntry[Boolean] =
+    buildConf("kyuubi.kubernetes.application.state.fromContainer")
+      .doc("If set to true then the application state will be retrieved from the container " +
+        "instead of the pod.")
+      .version("1.8.1")
+      .booleanConf
+      .createWithDefault(false)
+
+  val KUBERNETES_APPLICATION_STATE_CONTAINER: OptionalConfigEntry[String] =
+    buildConf("kyuubi.kubernetes.application.state.container")
```
pan3793, line 1243: I couldn’t recall the details well. how can we configure the container name? does it have a default name? throw pod template?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1231,6 +1231,22 @@ object KyuubiConf {
...
+      .booleanConf
+      .createWithDefault(false)
+
+  val KUBERNETES_APPLICATION_STATE_CONTAINERS: ConfigEntry[Set[String]] =
+    buildConf("kyuubi.kubernetes.application.state.containers")
+      .doc("A comma-separated list of container names to retrieve the application state from.")
+      .version("1.8.1")
+      .stringConf
+      .toSet()
+      .createWithDefault(Set.empty)
```
zwangsheng, line 1248: how about set default to `spark-kubernetes-driver`(default spark driver container name)

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1231,6 +1231,22 @@ object KyuubiConf {
...
+      .version("1.8.1")
+      .booleanConf
+      .createWithDefault(false)
+
+  val KUBERNETES_APPLICATION_STATE_CONTAINERS: ConfigEntry[Set[String]] =
+    buildConf("kyuubi.kubernetes.application.state.containers")
+      .doc("A comma-separated list of container names to retrieve the application state from.")
+      .version("1.8.1")
+      .stringConf
+      .toSet()
```
pan3793, line 1247: why should we care about a set of containers' status? and how can we determine the final application status from each container's status? e.g. use OR or AND strategy?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: SGTM

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: will give a default value `spark-kubernetes-driver`

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: changed to use single container name

zwangsheng (commented):
LGTM at all and this is a good idea to check application status by container status!

But have one more question about: we set this `kyuubi.kubernetes.application.state.fromContainer ` as server level conf, what will happen when user submit spark app with custom container name(by setting pod template and `spark.kubernetes.driver.podTemplateContainerName`)?

turboFei:
> But have one more question about: we set this kyuubi.kubernetes.application.state.fromContainer  as server level conf, what will happen when user submit spark app with custom container name(by setting pod template and spark.kubernetes.driver.podTemplateContainerName)?

If not container matched with the server level `kyuubi.kubernetes.application.state.container`, it will fallback to use the pod state to build application state.

So, I think it does not matter.

turboFei:
Or I can even ignore the user specified `spark.kubernetes.driver.podTemplateContainerName` in kyuubi server side.

With kyuubi conf:
`kyuubi.session.conf.ignore.list` and `kyuubi.batch.conf.ignore.list`.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1239,6 +1239,21 @@ object KyuubiConf {
       .booleanConf
       .createWithDefault(false)
 
+  val KUBERNETES_APPLICATION_STATE_FROM_CONTAINER: ConfigEntry[Boolean] =
+    buildConf("kyuubi.kubernetes.application.state.fromContainer")
```
pan3793, line 1243: actually, there are other potential ways for app state judgement, e.g. service, or CRD object, so I would suggest using an emum instead of bool for it.
```
kyuubi.kubernetes.application.state.from=[pod|contianer]
```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1239,6 +1239,21 @@ object KyuubiConf {
...
+      .doc("If set to true then the application state will be retrieved from the container " +
+        "instead of the pod.")
+      .version("1.8.1")
+      .booleanConf
+      .createWithDefault(false)
+
+  val KUBERNETES_APPLICATION_STATE_CONTAINER: ConfigEntry[String] =
+    buildConf("kyuubi.kubernetes.application.state.container")
+      .doc("The container names to retrieve the application state from.")
```
pan3793, line 1252: there are bunches of tricks under the hood, users may not get enough information from such a brief description.

e.g. in which conditions, this configuration takes effect? how to set the value(e.g. it must be same as xxx)?

pan3793:
> If no container matched with the server level kyuubi.kubernetes.application.state.container, it will fallback to use the pod state to build application state.

such behavior should be documented clearly.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -295,16 +314,64 @@ object KubernetesApplicationOperation extends Logging {
...
+      Option(containerState.getTerminated) match {
+        case Some(terminated) =>
+          if (terminated.getExitCode == 0) {
+            FINISHED
+          } else {
+            FAILED
+          }
+        case None => UNKNOWN
+      }
+    }
```
pan3793, line 357:
```suggestion
    } else if (containerState.getTerminated == null) {
      UNKNOWN
    } else if (containerState.getTerminated.getExitCode == 0) {
      FINISHED
    } else {
      FAILED
    }
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -295,16 +314,64 @@ object KubernetesApplicationOperation extends Logging {
...
+    toApplicationStateAndError(pod, appStateFromContainer, appStateContainer)._1
+  }
+
+  def toApplicationStateAndError(
+      pod: Pod,
+      appStateFromContainer: Boolean,
+      appStateContainer: String): (ApplicationState, Option[String]) = {
+    val containerStateToBuildAppState = if (appStateFromContainer) {
+      pod.getStatus.getContainerStatuses.asScala.find(_.getState == appStateContainer).map(
+        _.getState)
```
pan3793, line 330: avoid breaking line after `func(\n lambda ...`

```suggestion
      pod.getStatus.getContainerStatuses.asScala
        .find(_.getState == appStateContainer).map(_.getState)
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -295,16 +314,64 @@ object KubernetesApplicationOperation extends Logging {
...
+  def podStateToApplicationState(podState: String): ApplicationState = podState match {
+    // https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
     case "Pending" => PENDING
     case "Running" => RUNNING
     case "Succeeded" => FINISHED
     case "Failed" | "Error" => FAILED
     case "Unknown" => UNKNOWN
     case _ =>
-      warn(s"The kubernetes driver pod state: $state is not supported, " +
+      warn(s"The kubernetes driver pod state: $podState is not supported, " +
```
pan3793, line 374:
```suggestion
      warn(s"The Spark driver pod state: $podState is not supported, " +
```

turboFei:
cc @pan3793 @zwangsheng

pan3793 (approved):

PR #5750, opened 2023-11-22
[Spark] Fix the inaccurate issue of obtaining COLUMN_SIZE in the decimal field jdbc of spark engine

waywtdcc:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

Fix the inaccuracy in getting COLUMN_SIZE from spark engine's decimal field jdbc. The current jdbc's get column size gets the decimal field as default size, which is inaccurate; if it is decimal(20,3), the obtained column size is 16; the actual is 20.

- [X] Bugfix (non-breaking change which fixes an issue)

#### Related Unit Tests

---

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
@@ -140,14 +140,11 @@ object SchemaHelper {
     case dt
         if Array(TIMESTAMP_NTZ, DAY_TIME_INTERVAL, YEAR_MONTH_INTERVAL)
           .contains(dt.getClass.getSimpleName) => Some(dt.defaultSize)
+    case dt @ (_: DecimalType) =>
+      Some(dt.precision)
     case dt @ (BooleanType | _: NumericType | DateType | TimestampType |
         CalendarIntervalType | NullType) =>
-      // decimal type
-      if (dt.isInstanceOf[DecimalType]) {
-        Some(dt.asInstanceOf[DecimalType].precision)
-      } else {
-        Some(dt.defaultSize)
-      }
+      Some(dt.defaultSize)
     case StructType(fields) =>
       val sizeArr = fields.map(f => getColumnSize(f.dataType))
       if (sizeArr.contains(None)) {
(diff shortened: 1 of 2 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -142,7 +142,12 @@ object SchemaHelper {
           .contains(dt.getClass.getSimpleName) => Some(dt.defaultSize)
     case dt @ (BooleanType | _: NumericType | DateType | TimestampType |
         CalendarIntervalType | NullType) =>
-      Some(dt.defaultSize)
```
yaooqinn, line 145: We can add another case branch like `case dt: DecimalType =>`

yaooqinn:
Nice catch! Hi @waywtdcc , can we restore the checklist from PR desc?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -140,6 +140,8 @@ object SchemaHelper {
     case dt
         if Array(TIMESTAMP_NTZ, DAY_TIME_INTERVAL, YEAR_MONTH_INTERVAL)
           .contains(dt.getClass.getSimpleName) => Some(dt.defaultSize)
+    case dt @ (_: DecimalType) =>
```
yaooqinn, line 143: why do we need @ here？

yaooqinn (changes requested):
Hi, please restore the PR description

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
waywtdcc: case dt: DecimalType =>
      Some(dt.precision) ？

yaooqinn: yeah

yaooqinn: @waywtdcc can you fix this?

yaooqinn (approved):

yaooqinn:
Merged to master, 1.8, 1.7

PR #5772, opened 2023-11-24
[KYUUBI #5687] Using session user as proxy user to execute statements at the server and group share level

### LINKED ISSUE #5687: [Improvement] Using [session user] as [proxy user] to execute statements at the server share level
### Code of Conduct

### Search before asking

**At the server share level, If I don't use the Ranger AuthZ Plugin,** multiple users use the system user's credential, who submits the spark job, to access HDFS, rather than their owner credential. 

So on the HDFS side, all actions are executed as the system user, so the HDFS Ranger Plugin loses control.

I want to find a way for all session users can use their own Identity to access HDFS. Thus at the server share level, session users could be under minimum access control.

### How should we improve?

I read Kyuubi's code And found the function setting appropriate configs for each session user. 👇

https://github.com/apache/kyuubi/blob/master/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala#L130

Maybe we can set [session user] as [proxy user] to execute statements, like this: 👇

image

Are there some lurking issues or concerns at the system level?

Looking forward to your opinions very much!

theoryxu:
# :mag: Description
## Issue References 🔗

This pull request fixes #5687

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklists
## 📝 Author Self Checklist

- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

- [ ] Pull request title is okay.
- [ ] No license issues.
- [ ] Milestone correctly set?
- [ ] Test coverage is ok
- [ ] Assignees are selected.

**Be nice. Be informative.**

For caution, the parameter set false as the default


pan3793:
@theoryxu it would be nice if you could add a UT for this feature, to demonstrate the behavior differences with this feature enabled/disabled.

theoryxu:
> @theoryxu it would be nice if you could add a UT for this feature, to demonstrate the behavior differences with this feature enabled/disabled.

Thank you for your review.

I‘ll figure out how Kyuubi's UT framework works and add one for this. 

By the way, Are there other flaws I should know about?  I could fix it at once.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala
```diff
@@ -144,7 +151,14 @@ abstract class SparkOperation(session: Session)
           setSessionUserSign()
         }
 
-        f
+        if (isSessionUserAsProxyUser) {
+          UserGroupInformation.createProxyUser(session.user, UserGroupInformation.getCurrentUser)
```
cxzl25, line 155: Have you tested this solution?

My tests locally are not working as expected.

Create a proxy user. In the Kerberos environment, you should obtain the corresponding proxy user token.

https://github.com/apache/spark/blob/685f536a594700df32c22d62563ba3ec3dd26081/sql/hive/src/main/scala/org/apache/spark/sql/hive/security/HiveDelegationTokenProvider.scala#L104

And regardless of whether there is a Kerberos environment or not, SharedState is only initialized once, which means that the user is determined when the hive meta store client establishes a connection.

https://github.com/apache/spark/blob/685f536a594700df32c22d62563ba3ec3dd26081/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala#L268-L271

https://github.com/apache/spark/blob/685f536a594700df32c22d62563ba3ec3dd26081/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/IsolatedClientLoader.scala#L324

theoryxu: Thanks for your response.

I had tested in a simple auth environment before submitting this PR. It's worked. But you are right; It does not work well above the Kerberos (I confirmed it in my new clusters). 

 I'll try to make sure how it works with  Kerberos in detail. But I'm not very familiar with Kerberos. I'd appreciate it if someone could guide it.

Issue #5808, opened 2023-12-03
[Umbrella] Improvements and evaluation for TRowSet generation of Spark Engine


bowenliang123:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues) and found no similar issues.

### Describe the proposal

RowSet generation is that taking the results from result iterator and serializing them into column-based or row-based TRowSet, which is the key point for transportation and performance in most common cases.

1. It's been reported possibility drawbacks in looping the result iterator by wrapped stream in `SparkOperation` (https://github.com/apache/kyuubi/blob/master/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/SparkOperation.scala#L253C21-L253C26)
2. the performance of Spark Engine's RowSet.toTRowSet should be evaluated by benchmarks, for overall performance and for each data type with different mode of Thrift version and arrow based.
3. Performance Improvements in Spark Engine's RowSet implementation
4. Code cleanup in Spark Engine's RowSet generation

### Task list

- [ ] benchmark ut 
    -  #5809 
       - [ ] Add benchmark unit test for RowSet generation covering supported data types
       - [ ] Add benchmark dedicated unit test for each supported data type for RowSet generation 
- [ ] Replace looping the iterator from `toSeq` (`toStream` of Iterator) to immutable collection
    - [ ] Compare toStream/toSeq/toList/toVector
    - #5804
- [ ] Parallel processing for column-based TRowSet generation
- [ ]  Performance in data types
    - [ ] DecimalType with column-based mode #5810
    - [ ] ArrayType of primitive data types with column based mode
- [ ] Code cleanup in RowSet of Spark Engine
    - [ ] Avoid redundant accessing the row by index  `rows(i)` for nullable checks and conversion
    - [ ] Avoid redundant accessing the column by index `row(i)` by index for nullable checks and conversion
    - [ ] #5811

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

bowenliang123:
Closing this PR with no enough consensus on the purposes, the design, the changes and the approaches.

PR #5811, opened 2023-12-04
[SPARK] Reuse time formatters instance in value serialization of TRowSet generation

bowenliang123:
# :mag: Description
## Issue References 🔗

Subtask of #5808.

## Describe Your Solution 🔧
Value serialization to Hive style string by  `HiveResult.toHiveString`  requires a `TimeFormatters` instance handling the date/time data types. Reusing the pre-created time formatters's instance, it dramatically reduces the overhead and improves the TRowset generation.

This may help to reduce memory footprints and fewer operations for TRowSet generation performance.

This is also aligned to the Spark's `RowSetUtils` in the implementation of RowSet generation for  [SPARK-39041](https://issues.apache.org/jira/browse/SPARK-39041) by @yaooqinn , with explicitly declared TimeFormatters instance.

#### Related Unit Tests

---

# Checklists
## 📝 Author Self Checklist

- [ ] My changes generate no new warnings

- [x] Pull request title is okay.
- [x] No license issues.
- [x] Milestone correctly set?
- [x] Test coverage is ok
- [x] Assignees are selected.
- [x] Minimum number of approvals
- [x] No changes are requested

**Be nice. Be informative.**

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
@@ -24,15 +24,17 @@ import scala.collection.JavaConverters._
 import org.apache.hive.service.rpc.thrift._
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.execution.HiveResult
+import org.apache.spark.sql.execution.HiveResult.TimeFormatters
 import org.apache.spark.sql.types._
 
 import org.apache.kyuubi.util.RowSetUtils._
 
 object RowSet {
 
-  def toHiveString(valueAndType: (Any, DataType), nested: Boolean = false): String = {
-    // compatible w/ Spark 3.1 and above
-    val timeFormatters = HiveResult.getTimeFormatters
+  def toHiveString(
+      valueAndType: (Any, DataType),
+      nested: Boolean = false,
+      timeFormatters: TimeFormatters): String = {
     HiveResult.toHiveString(valueAndType, nested, timeFormatters)
   }
 
@@ -71,6 +73,7 @@ object RowSet {
   def toRowBasedSet(rows: Seq[Row], schema: StructType): TRowSet = {
     val rowSize = rows.length
     val tRows = new java.util.ArrayList[TRow](rowSize)
+    val timeFormatters = HiveResult.getTimeFormatters
     var i = 0
     while (i < rowSize) {
       val row = rows(i)
@@ -78,7 +81,7 @@ object RowSet {
       var j = 0
       val columnSize = row.length
       while (j < columnSize) {
-        val columnValue = toTColumnValue(j, row, schema)
+        val columnValue = toTColumnValue(j, row, schema, timeFormatters)
         tRow.addToColVals(columnValue)
         j += 1
       }
@@ -91,18 +94,23 @@ object RowSet {
   def toColumnBasedSet(rows: Seq[Row], schema: StructType): TRowSet = {
     val rowSize = rows.length
     val tRowSet = new TRowSet(0, new java.util.ArrayList[TRow](rowSize))
+    val timeFormatters = HiveResult.getTimeFormatters
     var i = 0
     val columnSize = schema.length
     while (i < columnSize) {
       val field = schema(i)
-      val tColumn = toTColumn(rows, i, field.dataType)
+      val tColumn = toTColumn(rows, i, field.dataType, timeFormatters)
       tRowSet.addToColumns(tColumn)
       i += 1
     }
     tRowSet
   }
 
-  private def toTColumn(rows: Seq[Row], ordinal: Int, typ: DataType): TColumn = {
+  private def toTColumn(
+      rows: Seq[Row],
+      ordinal: Int,
+      typ: DataType,
+      timeFormatters: TimeFormatters): TColumn = {
     val nulls = new java.util.BitSet()
     typ match {
       case BooleanType =>
@@ -152,7 +160,7 @@ object RowSet {
         while (i < rowSize) {
           val row = rows(i)
           nulls.set(i, row.isNullAt(ordinal))
-          values.add(toHiveString(row.get(ordinal) -> typ))
+          values.add(toHiveString(row.get(ordinal) -> typ, timeFormatters = timeFormatters))
           i += 1
         }
         TColumn.stringVal(new TStringColumn(values, nulls))
@@ -184,7 +192,8 @@ object RowSet {
   private def toTColumnValue(
       ordinal: Int,
       row: Row,
-      types: StructType): TColumnValue = {
+      types: StructType,
+      timeFormatters: TimeFormatters): TColumnValue = {
     types(ordinal).dataType match {
       case BooleanType =>
         val boolValue = new TBoolValue
@@ -232,7 +241,9 @@ object RowSet {
       case _ =>
         val tStrValue = new TStringValue
         if (!row.isNullAt(ordinal)) {
-          tStrValue.setValue(toHiveString(row.get(ordinal) -> types(ordinal).dataType))
+          tStrValue.setValue(toHiveString(
+            row.get(ordinal) -> types(ordinal).dataType,
+            timeFormatters = timeFormatters))
         }
         TColumnValue.stringVal(tStrValue)
     }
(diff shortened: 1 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
```diff
@@ -24,15 +24,24 @@ import scala.collection.JavaConverters._
...
 
 object RowSet {
 
+  private val timeUnrelatedDataTypes: Set[DataType] =
+    Set(BooleanType, FloatType, BinaryType, StringType)
+
   def toHiveString(valueAndType: (Any, DataType), nested: Boolean = false): String = {
     // compatible w/ Spark 3.1 and above
-    val timeFormatters = HiveResult.getTimeFormatters
+    val timeFormatters: TimeFormatters = valueAndType match {
```
wForget, line 39: We may need to create a timeFormatters for a dataset instead of each cell.

wForget: Refer to the implementation in spark: https://github.com/apache/spark/blob/master/sql/hive-thriftserver/src/main/scala/org/apache/spark/sql/hive/thriftserver/RowSetUtils.scala#L37

wForget: I think we can reuse it in the `toTRowSet` method

bowenliang123:

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
bowenliang123: Sure. Adopted.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
```diff
@@ -24,15 +24,17 @@ import scala.collection.JavaConverters._
...
+import org.apache.spark.sql.execution.HiveResult.{getTimeFormatters, TimeFormatters}
 import org.apache.spark.sql.types._
 
 import org.apache.kyuubi.util.RowSetUtils._
 
 object RowSet {
 
-  def toHiveString(valueAndType: (Any, DataType), nested: Boolean = false): String = {
-    // compatible w/ Spark 3.1 and above
-    val timeFormatters = HiveResult.getTimeFormatters
```
bowenliang123, line 35: PS: The key change of this PR, skipping time formatters recreation.

cfmcgrady: I think the reason that the time formatter is created each time is because the session time zone can be modified by querying `set spark.sql.session.timeZone=<new time zone>`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -103,19 +104,20 @@ object SparkDatasetHelper extends Logging {
       timestampAsString: Boolean): DataFrame = {
 
     val quotedCol = (name: String) => col(quoteIfNeeded(name))
+    val tf = getTimeFormatters
```
wForget, line 107: I'm not sure if `HiveResult.getTimeFormatters` is thread safe and serializable, `convertTopLevelComplexTypeToHiveString` method is only for handling complex types in arrow mode, @cfmcgrady do you have any suggestions?

bowenliang123: Shall we roll back to create TimeFormatters inside `convertTopLevelComplexTypeToHiveString`, if `HiveResult.getTimeFormatters` is probably not thread-safe ?

wForget: > Shall we roll back to create TimeFormatters inside `convertTopLevelComplexTypeToHiveString`, if `HiveResult.getTimeFormatters` is probably not thread-safe ?

In that case, we could just do an adjustment like this

```
 RowSet.toHiveString((row, st), nested = true, timeFormatters = getTimeFormatters)
```

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
```diff
@@ -91,18 +94,23 @@ object RowSet {
   def toColumnBasedSet(rows: Seq[Row], schema: StructType): TRowSet = {
     val rowSize = rows.length
     val tRowSet = new TRowSet(0, new java.util.ArrayList[TRow](rowSize))
+    val timeFormatters = getTimeFormatters
```
pan3793, line 97: this is a good place as the time formatter is session timezone aware
```
  def getTimeFormatters: TimeFormatters = {
    val dateFormatter = DateFormatter()
    val timestampFormatter = TimestampFormatter.getFractionFormatter(
      DateTimeUtils.getZoneId(SQLConf.get.sessionLocalTimeZone))
    TimeFormatters(dateFormatter, timestampFormatter)
  }
```

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/schema/RowSetSuite.scala
```diff
@@ -97,6 +98,8 @@ class RowSetSuite extends KyuubiFunSuite {
 
   private val rows: Seq[Row] = (0 to 10).map(genRow) ++ Seq(Row.fromSeq(Seq.fill(17)(null)))
 
+  private val tf: TimeFormatters = getTimeFormatters
```
wForget, line 101: Please recreate it for each test

bowenliang123: Done.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
bowenliang123: Yes, good point.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
cfmcgrady: > In that case, we could just do an adjustment like this
> ```
>  RowSet.toHiveString((row, st), nested = true, timeFormatters = getTimeFormatters)
> ```

the partition/task level LGTM.

wForget: It may not be thread-safe, so we should try to control its usage scope.

FYI: https://github.com/apache/spark/blob/43ca0b929ab3c2f10d1879e5df622195564f8885/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/DateFormatter.scala#L198

https://github.com/apache/spark/blob/43ca0b929ab3c2f10d1879e5df622195564f8885/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/TimestampFormatter.scala#L532

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
bowenliang123: Yes, it's session config awared.

bowenliang123:
image

The green block shows the approximatly from 5x to 10x Improvements on decimal, date, timestamp, array, map, interval, localdate and instant.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
pan3793, line 97: super nit: keep the object name. avoid eliminating the object name unless it can improve the code style significantly.
```suggestion
    val timeFormatters = HiveResult.getTimeFormatters
```

pan3793 (approved):
LGTM, only a minor comment

pan3793:
@bowenliang123 the benchmark result looks impressive. would be better to group the results by data type.

cfmcgrady (approved):
LGTM. good catch!

pan3793:
Thanks, merged to master/1.8

PR #5812, opened 2023-12-04
[Spark] Reuse the column value skipping redundant access on row(i)

bowenliang123:
# :mag: Description
## Issue References 🔗

Subtask of #5808 

## Describe Your Solution 🔧
- Avoid redundant access on row(i) in both the `isNullAt(ordinal)` and `row.getAs[T](ordinal)` by Reusing the column value

#### Related Unit Tests

---

# Checklists
## 📝 Author Self Checklist

- [ ] My code follows the [style guidelines](https://kyuubi.readthedocs.io/en/master/contributing/code/style.html) of this project
- [ ] I have performed a self-review
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

- [ ] Pull request title is okay.
- [ ] No license issues.
- [ ] Milestone correctly set?
- [ ] Test coverage is ok
- [ ] Assignees are selected.

**Be nice. Be informative.**

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
@@ -169,12 +169,13 @@ object RowSet {
     var idx = 0
     while (idx < size) {
       val row = rows(idx)
-      val isNull = row.isNullAt(ordinal)
+      val value = row.get(ordinal)
+      val isNull = value == null
       if (isNull) {
         nulls.set(idx, true)
         ret.add(idx, defaultVal)
       } else {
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
```diff
@@ -169,12 +169,13 @@ object RowSet {
     var idx = 0
     while (idx < size) {
       val row = rows(idx)
-      val isNull = row.isNullAt(ordinal)
+      val value = row.get(ordinal)
+      val isNull = value == null
```
pan3793, line 173: I think we can not make such an assumption, the implementation may be different, for example, the developer uses a custom value to represent `NULL` in its own `CustomRow`, especially for those non-JVM implementations

Issue #5935, opened 2023-12-30
[FEATURE] Engine yarn mode support proxy user


Yikf:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Engine yarn mode support proxy user

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #5936, opened 2023-12-30
[KYUUBI #5935] Engine YARN mode support proxy appUser

Yikf:
# :mag: Description
## Issue References 🔗

This pull request fixes https://github.com/apache/kyuubi/issues/5935

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
@@ -19,6 +19,7 @@ package org.apache.kyuubi.engine.deploy.yarn
 import java.io._
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
+import java.security.PrivilegedExceptionAction
 import java.util
 import java.util.{Locale, Properties}
 import java.util.zip.{ZipEntry, ZipOutputStream}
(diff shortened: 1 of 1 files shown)


Yikf:
image

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
```diff
@@ -92,6 +94,28 @@ abstract class EngineYarnModeSubmitter extends Logging {
   def engineExtraJars(): Seq[File] = Seq.empty
 
   protected def submitApplication(): Unit = {
+    val user = kyuubiConf.getOption(KYUUBI_SESSION_USER_KEY)
```
pan3793, line 97: we actually have seen that limitation of always using session user as the engine launch user, let's define a dedicated `proxyUser: Option[String]`

pan3793:
BTW, for hadoop-testing environment, the recommended `HIVE_HADOOP_CLASSPATH` is `/opt/hadoop/share/hadoop/client/*`, there are only Hadoop Shaded client jars

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
Yikf: The user retrieved in this context is identified as appUser, which is supplied via the ProcessBuilder using the --conf parameter. However, the naming of the configuration key associated with this user appears to be somewhat unclear. 

Are you suggesting that the key name should be updated for clarity, or are you indicating that appUser does not represent the intended proxy user?

PR #5940, opened 2024-01-03
[Flink] Reduce excessively long thread sleep

### LINKED ISSUE #5891: [Bug] Flink engine jdbc execution time is much slower than the flink web display time
### Code of Conduct

### Search before asking

### Describe the bug

 Flink engine jdbc execution time is much slower than the flink web display time

SELECT *
FROM TABLEA
image

image

### Affects Version(s)

1.8.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

waywtdcc:
This pull request fixes #5891 
Reduce excessively long thread sleep

### DIFF
--- a/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/result/IncrementalResultFetchIterator.scala
+++ b/externals/kyuubi-flink-sql-engine/src/main/scala/org/apache/kyuubi/engine/flink/result/IncrementalResultFetchIterator.scala
@@ -58,7 +58,7 @@ class IncrementalResultFetchIterator(
 
   var hasNext: Boolean = true
 
-  val FETCH_INTERVAL_MS: Long = 1000
+  val FETCH_INTERVAL_MS: Long = 50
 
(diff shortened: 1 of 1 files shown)


pan3793:
@waywtdcc thanks for your contribution, please restore and fill in the PR description seriously, the detailed change description and review history are extremely important to an open-source project.

pan3793:
@link3280 @YesOrNo828 do you have suggestions/concerns on changing this value? Or maybe we should make it configurable?

link3280:
The default fetch interval is the same as [CliTableResultView in Flink SQL client](https://github.com/apache/flink/blob/6e78eb18524ead3abd60da0ca41751b45e0e2482/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliTableResultView.java#L62), which I think is a reasonable default.

I would rather make it configurable if necessary.

YesOrNo828:
+1 make it configurable.

PR #5953, opened 2024-01-09
[LICENSE] Update NOTICE

pan3793:
# :mag: Description
## Issue References 🔗

This pull request fixes the NOTICE issue.

## Describe Your Solution 🔧

According to https://infra.apache.org/licensing-howto.html#alv2-dep, we should analyze and bubble up the relevant portions into the top-level NOTICE file, no matter whether they're ASF projects or not.

> If the dependency supplies a NOTICE file, its contents must be analyzed and the relevant portions bubbled up into the top-level NOTICE file.

Kyuubi has some code copied/derived from Apache Iceberg/Hive/Spark/Parquet MR, so we should update NOTICE to respect that.

## Test Plan 🧪

Non-code change, review is required.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -14,5 +14,22 @@ Copyright 2017-2022 The Apache Software Foundation
 Apache Parquet MR
 Copyright 2014-2024 The Apache Software Foundation
 
+This project includes code from Kite, developed at Cloudera, Inc. with
+the following copyright notice:
+
+| Copyright 2013 Cloudera Inc.
+|
+| Licensed under the Apache License, Version 2.0 (the "License");
+| you may not use this file except in compliance with the License.
+| You may obtain a copy of the License at
+|
+|   http://www.apache.org/licenses/LICENSE-2.0
+|
+| Unless required by applicable law or agreed to in writing, software
+| distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


yaooqinn:
> Kyuubi has some code copied/derived from Apache Iceberg/Hive/Spark/Parquet MR, so we should update NOTICE to respect that.

Please also address these in the LICENSE file.

pan3793:
@yaooqinn listing them under LICENSE is not mandatory, just to facilitate traceability. We already use explicit words like "copied from XXXX" in class comments to achieve that.

justinmclean:
It is best to list the files in LICENSE even if they are ALv2 licensed, as users know which files they are. As all ASFs have the same header files, it is hard to determine where a file comes from, and you may want to update it.

pan3793:
> ... as users know which files they are ...

for project-level auditing, it could be done by searching the keyword "copied from".

for the developer/user reading code, the inline comments are more convenient

image

In my experience, the tracking list maintained externally without an automated script to check is easily out-of-sync, and eventually corrupt.

pan3793:
cc @cfmcgrady you may have the full context of this PR

cfmcgrady (approved):
LGTM

pan3793:
Thanks, merged to master/1.8/1.7

PR #5961, opened 2024-01-10
Support to specify client kerberosAuthType as fromTicketCache 

turboFei:
# :mag: Description

The use case:
- 1. the user code running in flink cluster and wrapped by transparent `ugi.doAs`
- 2. the current UserGroupInfomation is the proxy user not login/real user
- 3. user specify url with kyuubiClientTicketCache but does not work
- 4. the proxy UserGroupInfomation is used and then throw GSS issue.
```
Caused by: org.apache.kyuubi.shade.org.apache.thrift.transport.TTransportException: GSS initiate failed
    at org.apache.kyuubi.shade.org.apache.thrift.transport.TSaslTransport.sendAndThrowMessage(TSaslTransport.java:232)
    at org.apache.kyuubi.shade.org.apache.thrift.transport.TSaslTransport.open(TSaslTransport.java:316)
    at org.apache.kyuubi.shade.org.apache.thrift.transport.TSaslClientTransport.open(TSaslClientTransport.java:37)
    at org.apache.kyuubi.jdbc.hive.auth.TSubjectTransport.lambda$open$0(TSubjectTransport.java:47)
    at java.security.AccessController.doPrivileged(Native Method)
    at javax.security.auth.Subject.doAs(Subject.java:422)
    at org.apache.kyuubi.jdbc.hive.auth.TSubjectTransport.open(TSubjectTransport.java:42)
    at org.apache.kyuubi.jdbc.hive.KyuubiConnection.openTransport(KyuubiConnection.java:458)
    at org.apache.kyuubi.jdbc.hive.KyuubiConnection.<init>(KyuubiConnection.java:207)
    ... 22 more
```

The root cause is that, for this case, the result of `isHadoopUserGroupInformationDoAs` is true.

So, `isFromSubjectAuthMode` is true.

In this pr, I want to specify the kerberosAuthType to fromTicketCache, and do not check `isHadoopUserGroupInformationDoAs`. 

After this pr, customer can specify `kerberosAuthType=fromTicketCache` to leverage ticket cache prefer than `ugi.doAs`.

## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -847,30 +847,59 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
     }
   }
 
+  private boolean isForciblyFromKeytabAuthMode() {
+    return AUTH_KERBEROS_AUTH_TYPE_FROM_KEYTAB.equalsIgnoreCase(
+        sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE));
+  }
+
+  private boolean isForciblyFromSubjectAuthMode() {
+    return AUTH_KERBEROS_AUTH_TYPE_FROM_SUBJECT.equalsIgnoreCase(
+        sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE));
+  }
+
+  private boolean isForciblyTgtCacheAuthMode() {
+    return AUTH_KERBEROS_AUTH_TYPE_FROM_TICKET_CACHE.equalsIgnoreCase(
+        sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE));
+  }
+
   private boolean isKeytabAuthMode() {
-    return isSaslAuthMode()
-        && hasSessionValue(AUTH_PRINCIPAL)
+    // handle explicit cases first
+    if (isForciblyFromSubjectAuthMode() || isForciblyTgtCacheAuthMode()) {
+      return false;
+    }
+    if (isKerberosAuthMode() && isForciblyFromKeytabAuthMode()) {
+      return true;
+    }
+    // handle implicit cases then
+    return isKerberosAuthMode()
         && hasSessionValue(AUTH_KYUUBI_CLIENT_PRINCIPAL)
         && hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB);
   }
 
   private boolean isFromSubjectAuthMode() {
-    return isSaslAuthMode()
-        && hasSessionValue(AUTH_PRINCIPAL)
-        && !hasSessionValue(AUTH_KYUUBI_CLIENT_PRINCIPAL)
+    // handle explicit cases first
+    if (isForciblyFromKeytabAuthMode() || isForciblyTgtCacheAuthMode()) {
+      return false;
+    }
+    if (isKerberosAuthMode() && isForciblyFromSubjectAuthMode()) {
+      return true;
+    }
+    // handle implicit cases then
+    return isKerberosAuthMode()
         && !hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB)
-        && (AUTH_KERBEROS_AUTH_TYPE_FROM_SUBJECT.equalsIgnoreCase(
-                sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE))
-            || (!AUTH_KERBEROS_AUTH_TYPE_FROM_TICKET_CACHE.equalsIgnoreCase(
-                    sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE))
-                && isHadoopUserGroupInformationDoAs()));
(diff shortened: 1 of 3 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -861,7 +861,9 @@ && hasSessionValue(AUTH_PRINCIPAL)
         && !hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB)
         && (AUTH_KERBEROS_AUTH_TYPE_FROM_SUBJECT.equalsIgnoreCase(
                 sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE))
-            || isHadoopUserGroupInformationDoAs());
+            || (!AUTH_KERBEROS_AUTH_TYPE_FROM_TICKET_CACHE.equalsIgnoreCase(
+                    sessConfMap.get(AUTH_KERBEROS_AUTH_TYPE))
+                && isHadoopUserGroupInformationDoAs()));
```
pan3793, line 866: if use specific `kerberosAuthType=fromTicketCache` we should always respect that, thus `isFromSubjectAuthMode()` should always return `false`, no matter under UGI context or not, right?

pan3793: if so, I suggest changing it to

(I use string literal in the example, please replace them with const variable)
```java
  private boolean isForciblyFromSubjectAuthMode() {
    return "fromSubject".equalsIgnoreCase(sessConfMap.get("kerberosAuthType"));
  }

  private boolean isForciblyTgtCacheAuthMode() {
    return "fromTicketCache".equalsIgnoreCase(sessConfMap.get("kerberosAuthType"));
  }

  ...

  private boolean isFromSubjectAuthMode() {
    return isSaslAuthMode()
        && hasSessionValue(AUTH_PRINCIPAL)
        && !hasSessionValue(AUTH_KYUUBI_CLIENT_PRINCIPAL)
        && !hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB)
        && !isForciblyTgtCacheAuthMode()
        && (isForciblyFromSubjectAuthMode() || isHadoopUserGroupInformationDoAs());
  }

```

pan3793: I found another case, if the user provides `kyuubiClientPrinciple` and `kyuubiClientKeytab` while explicitly setting `kerberosAuthType=fromTicketCache`, the code path always goes keytab mode, still intuitive ...

pan3793: maybe something like
```java
  private boolean isForciblyFromKeytabAuthMode() {
    return "fromKeytab".equalsIgnoreCase(sessConfMap.get("kerberosAuthType"));
  }

  private boolean isForciblyFromSubjectAuthMode() {
    return "fromSubject".equalsIgnoreCase(sessConfMap.get("kerberosAuthType"));
  }

  private boolean isForciblyTgtCacheAuthMode() {
    return "fromTicketCache".equalsIgnoreCase(sessConfMap.get("kerberosAuthType"));
  }

  private boolean isKeytabAuthMode() {
    // handle explicit cases first
    if (isForciblyFromSubjectAuthMode() || isForciblyTgtCacheAuthMode()) {
      return false;
    }
    if (isKerberosAuthMode() && isForciblyFromKeytabAuthMode()) {
      return true;
    }
    // handle implicit cases then
    return isKerberosAuthMode()
        && hasSessionValue(AUTH_KYUUBI_CLIENT_PRINCIPAL)
        && hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB);
  }

  private boolean isFromSubjectAuthMode() {
    // handle explicit cases first
    if (isForciblyFromKeytabAuthMode() || isForciblyTgtCacheAuthMode()) {
      return false;
    }
    if (isKerberosAuthMode() && isForciblyFromSubjectAuthMode()) {
      return true;
    }
    // handle implicit cases then
    return isKerberosAuthMode()
        && !hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB)
        && isHadoopUserGroupInformationDoAs();
  }

  private boolean isTgtCacheAuthMode() {
    // handle explicit cases first
    if (isForciblyFromKeytabAuthMode() || isForciblyFromSubjectAuthMode()) {
      return false;
    }
    if (isKerberosAuthMode() && isForciblyTgtCacheAuthMode()) {
      return true;
    }
    // handle implicit cases then
    return isKerberosAuthMode()
        && !hasSessionValue(AUTH_KYUUBI_CLIENT_KEYTAB);
  }
```

turboFei: thanks for the suggestions, adopted.

pan3793 (approved):

PR #5988, opened 2024-01-17
[KYUUBI #5987] Enhance KYUUBI_HOME detection in shell script to handle softlink cases

### LINKED ISSUE #5987: [Improvement] [Bug] Modify the logic in the beeline script that specifies the KYUUBI_HOME environment variable.
### Code of Conduct

### Search before asking

### Describe the bug

When using Kyuubi in my company, there is a scenario where we create a symbolic link from ${KYUUBI_HOME}/bin/beeline to /usr/bin/kyuubi-beeline. This allows distinguishing its usage from Hive's beeline, and users can globally use kyuubi-beeline without specifying a directory. However, when executing kyuubi-beeline, an error occurs, stating that it cannot find load-kyuubi-env.sh. The specific error message is as follows:**_" /bin/kyuubi-beeline: line 25: //bin/load-kyuubi-env.sh: No such file or directory."_** 

Personally, I believe the logic of exporting KYUUBI_HOME in ${KYUUBI_HOME}/bin/beeline should be modified. If the environment has already declared KYUUBI_HOME, it should use the pre-declared KYUUBI_HOME instead of forcefully specifying the parent directory of the current one.

### Affects Version(s)

master/1.8.0

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

Emor-nj:
# :mag: Description
## Issue References 🔗

This pull request fixes #5987
Modify the logic in the beeline script that specifies the KYUUBI_HOME environment variable.

## Describe Your Solution 🔧

 If the environment has already declared KYUUBI_HOME, it should use the pre-declared KYUUBI_HOME instead of forcefully specifying the parent directory of the current one.

1. set soft link:
 `ln -s /opt/soft/kyuubi/kyuubi/bin/beeline /usr/bin/kyuubi-beeline`
2. set env like:
`export KYUUBI_HOME=/opt/soft/kyuubi/kyuubi`
3. run kyuubi-beeline:
`kyuubi-beeline -u "jdbc:hive2://xx.xx.xx.xx:2181,xx.xx.xx.xx:2181,xx.xx.xx.xx:2181/tmp;serviceDiscoveryMode=zooKeeper;zooKeeperNamespace=kyuubi;principal=hive/_HOST@xxx"`
4. command will fail:
`/bin/kyuubi-beeline: line 27: //bin/load-kyuubi-env.sh: No such file or directory
Error: JAVA_HOME IS NOT SET! CANNOT PROCEED.`

1. set soft link:
 `ln -s /opt/soft/kyuubi/kyuubi/bin/beeline /usr/bin/kyuubi-beeline`
2. set env like:
`export KYUUBI_HOME=/opt/soft/kyuubi/kyuubi`
3. run kyuubi-beeline:
`kyuubi-beeline -u "jdbc:hive2://xx.xx.xx.xx:2181,xx.xx.xx.xx:2181,xx.xx.xx.xx:2181/tmp;serviceDiscoveryMode=zooKeeper;zooKeeperNamespace=kyuubi;principal=hive/_HOST@xxx"`
4. command will success
#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/bin/beeline
+++ b/bin/beeline
@@ -20,7 +20,7 @@
 CLASS="org.apache.hive.beeline.KyuubiBeeLine"
 
 if [ -z "${KYUUBI_HOME}" ]; then
-  export KYUUBI_HOME="$(cd "`dirname "$0"`"/..; pwd)"
+  KYUUBI_HOME="$(cd "`dirname "$0"`"/..; pwd)"
 fi
 
 . "${KYUUBI_HOME}/bin/load-kyuubi-env.sh" -s
--- a/bin/beeline
+++ b/bin/beeline
@@ -19,7 +19,7 @@
 ## Kyuubi BeeLine Entrance
 CLASS="org.apache.hive.beeline.KyuubiBeeLine"
 
-export KYUUBI_HOME="$(cd "$(dirname "$0")"/..; pwd)"
+export KYUUBI_HOME="${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}"
 
 . "${KYUUBI_HOME}/bin/load-kyuubi-env.sh" -s
 
--- a/bin/beeline
+++ b/bin/beeline
@@ -19,7 +19,9 @@
 ## Kyuubi BeeLine Entrance
 CLASS="org.apache.hive.beeline.KyuubiBeeLine"
 
-export KYUUBI_HOME="${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}"
+if [ -z "${KYUUBI_HOME}" ]; then
+  export KYUUBI_HOME="$(cd "`dirname "$0"`"/..; pwd)"
+fi
 
 . "${KYUUBI_HOME}/bin/load-kyuubi-env.sh" -s
 
--- a/bin/load-kyuubi-env.sh
+++ b/bin/load-kyuubi-env.sh
@@ -18,7 +18,7 @@
 
 
 if [ -z "${KYUUBI_HOME}" ]; then
-  export ${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}
+  export KYUUBI_HOME="${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}"
 fi
 export KYUUBI_CONF_DIR="${KYUUBI_CONF_DIR:-"${KYUUBI_HOME}"/conf}"
 
--- a/bin/load-kyuubi-env.sh
+++ b/bin/load-kyuubi-env.sh
@@ -18,7 +18,7 @@
 
 
 if [ -z "${KYUUBI_HOME}" ]; then
-  export KYUUBI_HOME="$(cd "`dirname "$0"`"/..; pwd)"
+  export ${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}
 fi
 export KYUUBI_CONF_DIR="${KYUUBI_CONF_DIR:-"${KYUUBI_HOME}"/conf}"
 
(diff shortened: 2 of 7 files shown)


Emor-nj:
cc @pan3793

pan3793:
seems we have the same issue on the other scripts.

actually, what we want to do here is to follow the soft link to find the real file, then locale the `load-kyuubi-env.sh` via the relative path. if we can do that, we don't even declare the evn KYUUBI_HOME before executing the soft link of those scripts

pan3793:
I checked Spark's scripts, let's simply use this pattern

```
if [ -z "${KYUUBI_HOME}" ]; then
  KYUUBI_HOME="$(cd "`dirname "$0"`"/..; pwd)"
fi
```

Emor-nj:
> seems we have the same issue on the other scripts.
> 
> actually, what we want to do here is to follow the soft link to find the real file, then locale the `load-kyuubi-env.sh` via the relative path. if we can do that, we don't even declare the evn KYUUBI_HOME before executing the soft link of those scripts

Good idea! Here's the modified line:
`export KYUUBI_HOME="$(cd "$(dirname "$(realpath "$0")")"/..; pwd)"`
If there are no issues, I will proceed to uniformly modify all related scripts.

Emor-nj:
> I checked Spark's scripts, let's simply use this pattern
> 
> ```
> if [ -z "${KYUUBI_HOME}" ]; then
>   KYUUBI_HOME="$(cd "`dirname "$0"`"/..; pwd)"
> fi
> ```

Is the intention to maintain consistency with Spark by prioritizing the use of environment variables rather than directly obtaining the absolute path?

pan3793:
is `realpath` available on all UNIX-like OS? I saw Spark [handles it exceptionally](https://github.com/apache/spark/blob/master/build/util.sh)

> Is the intention to maintain consistency with Spark ...

I haven't seen users report similar issues to Spark, thus I suppose the approach Spark used is reliable, it always is my first fallback candidate if no better options.

Emor-nj:
> is `realpath` available on all UNIX-like OS? I saw Spark [handles it exceptionally](https://github.com/apache/spark/blob/master/build/util.sh)
> 
> > Is the intention to maintain consistency with Spark ...
> 
> I haven't seen users report similar issues to Spark, thus I suppose the approach Spark used is reliable, it always is my first fallback candidate if no better options.

Get it, let's align with Spark. I will uniformly modify all related scripts.

bin/beeline
```diff
@@ -19,7 +19,9 @@
 ## Kyuubi BeeLine Entrance
 CLASS="org.apache.hive.beeline.KyuubiBeeLine"
 
-export KYUUBI_HOME="${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}"
+if [ -z "${KYUUBI_HOME}" ]; then
+  export KYUUBI_HOME="$(cd "`dirname "$0"`"/..; pwd)"
```
pan3793, line 23: export is not required

pan3793 (approved):
LGTM

bin/load-kyuubi-env.sh
```diff
@@ -18,7 +18,7 @@
 
 
 if [ -z "${KYUUBI_HOME}" ]; then
-  export ${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}
+  export KYUUBI_HOME="${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}"
```
pan3793, line 21: KYUUBI_HOME is always empty
```suggestion
  export KYUUBI_HOME="$(cd "$(dirname "$0")"/.. || exit; pwd)"
```

Emor-nj: get it, thanks

pan3793:
Thanks, merged to master/1.8

PR #5990, opened 2024-01-17
Always take the first declared SASL/PLAIN auth type

pan3793:
# :mag: Description
## Issue References 🔗

https://github.com/apache/kyuubi/pull/5185 changed the type of `kyuubi.authentication` from `Seq`(ordered) to `Set`(unordered), which break the assumption of 
```
Note that: for SASL authentication, KERBEROS and PLAIN auth types are supported
at the same time, and only the first specified PLAIN auth type is valid.
```

## Describe Your Solution 🔧

Restore the type to Seq

## Test Plan 🧪

UT is updated

---

# Checklist 📝

**Be nice. Be informative.**


bowenliang123 (approved):
LGTM. Thanks for fixing it.

pan3793:
Thanks, merged to master/1.8

PR #5994, opened 2024-01-18
Support to build kyuubi-spark-authz-shaded into kyuubi-spark-sql-engine module

turboFei:
# :mag: Description

In this PR, I add a profile `spark-authz-compile` to make kyuubi-spark-authz-shaded module built into kyuubi-spark-sql-engine module.

It can make the authz integration more easily for kyuubi-spark-sql-engine.
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


bowenliang123:
With the new profile activated, is the `kyuubi-spark-authz-shaded` shipped as a jar with kyuubi server or just embedded inside the spark SQL engine ?

pan3793:
this may not be a good idea. engine's plugins use lots of engine's internal APIs which can not work across the engine's versions, we usually leave the engine's plugin separately.

PR #6003, opened 2024-01-22
[KYUUBI #6003] Allow disabling user impersonation on launching engine

pan3793:
# :mag: Description
## Issue References 🔗

HiveServer2 has a configuration `hive.server2.enable.doAs` to control the execution user between the session user and the server user, Kyuubi's CONNECTION and USER share levels always perform like doAs enabled do. In CDH 5/6, this is disabled by default, users who want to migrate from CDH to Kyuubi may encounter permission issues with the current implementation.

## Describe Your Solution 🔧

This pull request introduces a new configuration `kyuubi.engine.doAs.enabled` to allow enable/disable user impersonation on launching engine. For security purpose, it's not allowed to be overridden by session conf.

The change in this PR has certain limitations:

- only supports Spark engine
- only supports interactive mode, specifically, it does not take effect on Spark batch mode now.

## Test Plan 🧪

The first step is passing all existing UTs when `kyuubi.engine.doAs.enabled=true`.

Tested on internal Kerberized-environment, when `kyuubi.engine.share.level=CONNECTION` and `kyuubi.engine.doAs.enabled=false`, use user 'spark' to launch engine, and the engine submitted without `--proxy-user spark`, thus engine launched by server user `hive`, then run `select session_user(), current_user()` and returns

```
+-----------------+-----------------+
| session_user()  | current_user()  |
+-----------------+-----------------+
| spark           | hive            |
+-----------------+-----------------+
```

And I checked the `spark.app.name` and registered path on Zookeeper also expected.
```
+-----------------+--------------------------------------------------------------------------+
|       key       |                       value                                              |
+-----------------+--------------------------------------------------------------------------+
| spark.app.name  | kyuubi_USER_SPARK_SQL_spark_default_51a416e5-6023-4bac-a964-cd9605f17c61 |
+-----------------+--------------------------------------------------------------------------+
```

---

# Checklist 📝

**Be nice. Be informative.**


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2087,20 +2087,33 @@ object KyuubiConf {
       .version("1.5.0")
       .fallbackConf(ENGINE_CONNECTION_URL_USE_HOSTNAME)
 
+  val ENGINE_DO_AS_ENABLED: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.doAs.enabled")
```
pan3793, line 2091: cc @yaooqinn @zhouyifan279 WDYT of this configuration name? My another candidate is `kyuubi.engine.impersonation.enabled`

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -175,11 +177,15 @@ class SparkProcessBuilder(
...
         val ugi = UserGroupInformation
           .loginUserFromKeytabAndReturnUGI(principal.get, keytab.get)
-        if (ugi.getShortUserName != proxyUser) {
+        if (doAsEnabled && ugi.getShortUserName != proxyUser) {
           warn(s"The session proxy user: $proxyUser is not same with " +
             s"spark principal: ${ugi.getShortUserName}, so we can't support use keytab. " +
             s"Fallback to use proxy user.")
           None
+        } else if (!doAsEnabled && ugi.getShortUserName != Utils.currentUser) {
+          warn(s"The server's user: ${Utils.currentUser} is not same with " +
```
wForget, line 186: Is there any issue with using keytab in this case?

pan3793: the checking rule is same: the login user must be the same as the launch engine user. that said, when doAs is disabled, only the server's principle and keytab is allowed.

wForget: > that said, when doAs is disabled, only the server's principle and keytab is allowed.

Got it, thanks.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -97,6 +98,8 @@ private[kyuubi] class EngineRef(
     case _ => user
   }
 
+  private[kyuubi] val effectiveUser: String = if (doAsEnabled) appUser else Utils.currentUser
```
pan3793, line 101: there are three user variables in this class now,

- `user`: session user
- `appUser`: for engine routing
- `effectiveUser`: for engine launching

those name are not much straightforward, maybe I need to rename them to

- `user` => `sessionUser`
- `appUser` => `routingUser`
- `effectiveUser` => `appUser`

thought?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2087,20 +2087,33 @@ object KyuubiConf {
...
+      .doc("Whether to enable user impersonation on launching engine. When enabled, " +
+        "for engines which supports user impersonation, e.g. SPARK, depends on the " +
+        s"`kyuubi.engine.share.level`, different users will be used to launch the engine. " +
+        "Otherwise, Kyuubi Server's user will always be used to launch the engine.")
+      .version("1.8.1")
+      .booleanConf
```
yaooqinn, line 2097: Let's add this to 1.9.0 instead of a bugfix version

pan3793: Okay

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
zhouyifan279: +1. It's better to let appUser be consistent with the user on YARN.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
zhouyifan279: +1 for kyuubi.engine.doAs.enabled.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -175,11 +177,15 @@ class SparkProcessBuilder(
...
+            s"spark principal: ${ugi.getShortUserName}, skip to use keytab.")
```
zhouyifan279, line 187: Change `skip to use keytab.` to `Fallback to use server's user.` can avoid misunderstanding.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2116,7 +2129,7 @@ object KyuubiConf {
       " <li>HIVE_SQL: specify this engine type will launch a Hive engine which can provide" +
       " all the capacity of the Hive Server2.</li>" +
       " <li>JDBC: specify this engine type will launch a JDBC engine which can forward " +
```
wForget, line 2131: There is also an extra space at the end of this line

pan3793: fixed

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
pan3793: updated. I kept both

wForget (approved):
LGTM

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -175,11 +177,16 @@ class SparkProcessBuilder(
...
           .loginUserFromKeytabAndReturnUGI(principal.get, keytab.get)
-        if (ugi.getShortUserName != proxyUser) {
+        if (doAsEnabled && ugi.getShortUserName != proxyUser) {
           warn(s"The session proxy user: $proxyUser is not same with " +
             s"spark principal: ${ugi.getShortUserName}, so we can't support use keytab. " +
             s"Fallback to use proxy user.")
           None
+        } else if (!doAsEnabled && ugi.getShortUserName != Utils.currentUser) {
+          warn(s"The server's user: ${Utils.currentUser} is not same with " +
+            s"spark principal: ${ugi.getShortUserName}, skip to use keytab. " +
```
zhouyifan279, line 187: Should be `skip using keytab`

zhouyifan279:
LGTM

pan3793:
Thanks, merged to master

PR #6004, opened 2024-01-22
[KYUUBI #5987] Always export KYUUBI_HOME in `load-kyuubi-env.sh`

pan3793:
# :mag: Description
## Issue References 🔗

This is a follow-up of #5987, after applying #5987, I found a regression, the Kyuubi Server Java process can not see KYUUBI_HOME env if we don't declare KYUUBI_HOME as an env var explicitly, this breaks the main resources jar discovery of the engine, and eventually fails the engine bootstrap.

## Describe Your Solution 🔧

Restore `export KYUUBI_HOME` logic in `load-kyuubi-env.sh`

## Test Plan 🧪

Build and start Kyuubi Server.
```
export SPARK_HOME=/Users/chengpan/app/spark-3.5.0-bin-hadoop3
build/dist --spark-provided --flink-provided --hive-provided
dist/bin/kyuubi run
```

Open session to launch a Spark engine
```
dist/bin/beeline -u 'jdbc:hive2://0.0.0.0:10009/default'
```

A wrong `kyuubi-spark-sql-engine_2.12-1.9.0-SNAPSHOT.jar` location was found.
```
/Users/chengpan/app/spark-3.5.0-bin-hadoop3/bin/spark-submit \
        ...
        --proxy-user chengpan /Users/chengpan/Projects/apache-kyuubi/dist/jars/externals/kyuubi-spark-sql-engine/target/kyuubi-spark-sql-engine_2.12-1.9.0-SNAPSHOT.jar
```

The correct `kyuubi-spark-sql-engine_2.12-1.9.0-SNAPSHOT.jar` location was found.
```
/Users/chengpan/app/spark-3.5.0-bin-hadoop3/bin/spark-submit \
        ...
        --proxy-user chengpan /Users/chengpan/Projects/apache-kyuubi/dist/externals/engines/spark/kyuubi-spark-sql-engine_2.12-1.9.0-SNAPSHOT.jar
```

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/bin/load-kyuubi-env.sh
+++ b/bin/load-kyuubi-env.sh
@@ -16,10 +16,7 @@
 # limitations under the License.
 #
 
-
-if [ -z "${KYUUBI_HOME}" ]; then
-  export KYUUBI_HOME="$(cd "$(dirname "$0")"/.. || exit; pwd)"
-fi
+export KYUUBI_HOME="${KYUUBI_HOME:-"$(cd "$(dirname "$0")"/.. || exit; pwd)"}"
 export KYUUBI_CONF_DIR="${KYUUBI_CONF_DIR:-"${KYUUBI_HOME}"/conf}"
 
 silent=0


pan3793:
cc @Emor-nj, sorry I didn't spot this issue during the previous review, would you please take a look?

Emor-nj:
> cc @Emor-nj, sorry I didn't spot this issue during the previous review, would you please take a look?
I think the reason is that when the Kyuubi service starts, the Kyuubi script([https://github.com/apache/kyuubi/blob/master/bin/kyuubi](url)) does not declare KYUUBI_HOME as an environment variable. It is only passed as a context variable to load-kyuubi-env.sh. Therefore, although load-kyuubi-env.sh detects that KYUUBI_HOME has a value, it still does not declare it as an environment variable. As a result, when launching the Kyuubi service process, Java cannot retrieve KYUUBI_HOME from the system environment variables. Consequently, ProcBuilder(https://github.com/apache/kyuubi/blob/master/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala) cannot obtain the correct main resources JAR when constructing the spark-submit command.
The code in ProcBuilder is:
`env.get(KYUUBI_HOME).toSeq
        .flatMap { p =>
          Seq(
            Paths.get(p, "externals", "engines", shortName, jarName),
            Paths.get(p, "externals", module, "target", jarName))
        }
        .find(Files.exists(_)).map(_.toAbsolutePath.toFile.getCanonicalPath)`
env.get(KYUUBI_HOME) will be empty.

pan3793:
@Emor-nj yea, I have the same conclusion, do you have concerns about this fix?

Emor-nj:
> @Emor-nj yea, I have the same conclusion, do you have concerns about this fix?

No，it looks good to me.

pan3793:
Thanks, merged to master/1.8

PR #6008, opened 2024-01-23
RESTful API supports killing engine forcibly

zhaohehuhu:
# :mag: Description
## Describe Your Solution 🔧

I'd like to introduce the feature that allows users to forcibly kill an engine through API.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/docs/client/rest/rest_api.md
+++ b/docs/client/rest/rest_api.md
@@ -457,13 +457,14 @@ Delete the specified engine.
 
 #### Request Parameters
 
-| Name                    | Description                   | Type             |
-|:------------------------|:------------------------------|:-----------------|
-| type                    | the engine type               | String(optional) |
-| sharelevel              | the engine share level        | String(optional) |
-| subdomain               | the engine subdomain          | String(optional) |
-| proxyUser               | the proxy user to impersonate | String(optional) |
-| hive.server2.proxy.user | the proxy user to impersonate | String(optional) |
+| Name                    | Description                         | Type              |
+|:------------------------|:------------------------------------|:------------------|
+| type                    | the engine type                     | String(optional)  |
+| sharelevel              | the engine share level              | String(optional)  |
+| subdomain               | the engine subdomain                | String(optional)  |
+| proxyUser               | the proxy user to impersonate       | String(optional)  |
+| hive.server2.proxy.user | the proxy user to impersonate       | String(optional)  |
+| kill                    | whether to kill the engine forcibly | Boolean(optional) |
 
 `proxyUser` is an alternative to `hive.server2.proxy.user`, and the current behavior is consistent with
 `hive.server2.proxy.user`. When both parameters are set, `proxyUser` takes precedence.
--- a/docs/client/rest/rest_api.md
+++ b/docs/client/rest/rest_api.md
@@ -457,14 +457,14 @@ Delete the specified engine.
 
 #### Request Parameters
 
-| Name                    | Description                         | Type              |
-|:------------------------|:------------------------------------|:------------------|
-| type                    | the engine type                     | String(optional)  |
-| sharelevel              | the engine share level              | String(optional)  |
-| subdomain               | the engine subdomain                | String(optional)  |
-| proxyUser               | the proxy user to impersonate       | String(optional)  |
-| hive.server2.proxy.user | the proxy user to impersonate       | String(optional)  |
-| kill                    | whether to kill the engine forcibly | Boolean(optional) |
+| Name                    | Description                                                  | Type              |
+|:------------------------|:-------------------------------------------------------------|:------------------|
+| type                    | the engine type                                              | String(optional)  |
+| sharelevel              | the engine share level                                       | String(optional)  |
+| subdomain               | the engine subdomain                                         | String(optional)  |
+| proxyUser               | the proxy user to impersonate                                | String(optional)  |
+| hive.server2.proxy.user | the proxy user to impersonate                                | String(optional)  |
+| kill                    | whether to kill the engine forcibly. Default value is false. | Boolean(optional) |
 
 `proxyUser` is an alternative to `hive.server2.proxy.user`, and the current behavior is consistent with
 `hive.server2.proxy.user`. When both parameters are set, `proxyUser` takes precedence.
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -3544,7 +3544,12 @@ object KyuubiConf {
       .version("1.8.0")
       .stringConf
       .toSequence()
-      .createWithDefault(Seq("spark.driver.memory", "spark.executor.memory"))
+      .createWithDefault(Seq(
+        "spark.driver.memory",
+        "spark.executor.memory",
+        "spark.kubernetes.context",
+        "spark.kubernetes.namespace",
+        "spark.master"))
 
   val ENGINE_SPARK_INITIALIZE_SQL: ConfigEntry[Seq[String]] =
     buildConf("kyuubi.engine.spark.initialize.sql")
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -3544,9 +3544,7 @@ object KyuubiConf {
       .version("1.8.0")
       .stringConf
       .toSequence()
-      .createWithDefault(Seq(
-        "spark.driver.memory",
-        "spark.executor.memory"))
+      .createWithDefault(Seq("spark.driver.memory", "spark.executor.memory"))
 
   val ENGINE_SPARK_INITIALIZE_SQL: ConfigEntry[Seq[String]] =
     buildConf("kyuubi.engine.spark.initialize.sql")
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -3546,10 +3546,7 @@ object KyuubiConf {
       .toSequence()
       .createWithDefault(Seq(
         "spark.driver.memory",
-        "spark.executor.memory",
-        "spark.kubernetes.context",
-        "spark.kubernetes.namespace",
-        "spark.master"))
+        "spark.executor.memory"))
 
   val ENGINE_SPARK_INITIALIZE_SQL: ConfigEntry[Seq[String]] =
     buildConf("kyuubi.engine.spark.initialize.sql")
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -161,13 +161,14 @@ object ApplicationManagerInfo extends Logging {
 
   def deserialize(encodedStr: String): ApplicationManagerInfo = {
     try {
-      info(s"The original string encoded:$encodedStr")
       val json = new String(
         Base64.getDecoder.decode(encodedStr.getBytes),
         StandardCharsets.UTF_8)
       mapper.readValue(json, classOf[ApplicationManagerInfo])
     } catch {
-      case _: Throwable => ApplicationManagerInfo(None)
+      case _: Throwable =>
+        error(s"The original string encoded:$encodedStr")
+        ApplicationManagerInfo(None)
     }
   }
 }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -19,8 +19,10 @@ package org.apache.kyuubi.engine
 
 import java.nio.charset.StandardCharsets
 import java.util.Base64
+
 import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
+
 import org.apache.kyuubi.Logging
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.ApplicationState.ApplicationState
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -17,6 +17,12 @@
 
 package org.apache.kyuubi.engine
 
+import java.nio.charset.StandardCharsets
+import java.util.Base64
+
+import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.ApplicationState.ApplicationState
 
@@ -134,6 +140,9 @@ case class ApplicationManagerInfo(
 
 object ApplicationManagerInfo {
   final val DEFAULT_KUBERNETES_NAMESPACE = "default"
+  val mapper: ObjectMapper = new ObjectMapper()
+    .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+    .registerModule(DefaultScalaModule)
 
   def apply(
       resourceManager: Option[String],
@@ -142,5 +151,21 @@ object ApplicationManagerInfo {
     new ApplicationManagerInfo(
       resourceManager,
       KubernetesInfo(kubernetesContext, kubernetesNamespace))
+  }
+
+  def serialize(appMgrInfo: ApplicationManagerInfo): String = {
+    Base64.getEncoder.encodeToString(
+      mapper.writeValueAsString(appMgrInfo).getBytes(StandardCharsets.UTF_8))
+  }
+
+  def deserialize(encodedStr: String): ApplicationManagerInfo = {
+    try {
+      val json = new String(
+        Base64.getDecoder.decode(encodedStr.getBytes),
+        StandardCharsets.UTF_8)
+      mapper.readValue(json, classOf[ApplicationManagerInfo])
+    } catch {
+      case _: Throwable => ApplicationManagerInfo(None)
+    }
   }
 }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -167,7 +167,7 @@ object ApplicationManagerInfo extends Logging {
       mapper.readValue(json, classOf[ApplicationManagerInfo])
     } catch {
       case _: Throwable =>
-        error(s"The original string encoded:$encodedStr")
+        error(s"Fail to desirable the encoded string: $encodedStr")
         ApplicationManagerInfo(None)
     }
   }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -19,10 +19,9 @@ package org.apache.kyuubi.engine
 
 import java.nio.charset.StandardCharsets
 import java.util.Base64
-
 import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
-
+import org.apache.kyuubi.Logging
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.ApplicationState.ApplicationState
 
@@ -138,7 +137,7 @@ case class ApplicationManagerInfo(
     resourceManager: Option[String],
     kubernetesInfo: KubernetesInfo = KubernetesInfo())
 
-object ApplicationManagerInfo {
+object ApplicationManagerInfo extends Logging {
   final val DEFAULT_KUBERNETES_NAMESPACE = "default"
   val mapper: ObjectMapper = new ObjectMapper()
     .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
@@ -160,6 +159,7 @@ object ApplicationManagerInfo {
 
   def deserialize(encodedStr: String): ApplicationManagerInfo = {
     try {
+      info(s"The original string encoded:$encodedStr")
       val json = new String(
         Base64.getDecoder.decode(encodedStr.getBytes),
         StandardCharsets.UTF_8)
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -167,7 +167,7 @@ object ApplicationManagerInfo extends Logging {
       mapper.readValue(json, classOf[ApplicationManagerInfo])
     } catch {
       case _: Throwable =>
-        error(s"Fail to desirable the encoded string: $encodedStr")
+        error(s"Fail to deserialize the encoded string: $encodedStr")
         ApplicationManagerInfo(None)
     }
   }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
@@ -45,6 +45,7 @@ import org.apache.kyuubi.session.{KyuubiSession, KyuubiSessionManager, SessionHa
 @Tag(name = "Admin")
 @Produces(Array(MediaType.APPLICATION_JSON))
 private[v1] class AdminResource extends ApiRequestContext with Logging {
+
   @ApiResponse(
     responseCode = "200",
     content = Array(new Content(mediaType = MediaType.APPLICATION_JSON)),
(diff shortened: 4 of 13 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -270,6 +275,18 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
               s"${e.getMessage}")
         }
       }
+
+      if (Option(refId).isDefined) {
```
wForget, line 279: Can we add a new parameter to decide whether to kill forcefully or gracefully？

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -270,6 +275,18 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
+
+        val applicationManagerInfo = ApplicationManagerInfo(
+          Option(sparkMaster),
+          Option(sparK8Context),
+          Option(sparK8Namespace))
+
+        fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]
```
wForget, line 286: We should further handle the result of `killApplication`.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
zhaohehuhu: Of course.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -33,18 +33,19 @@ import org.apache.kyuubi.{KYUUBI_VERSION, Logging}
...
 import org.apache.kyuubi.server.KyuubiServer
 import org.apache.kyuubi.server.api.{ApiRequestContext, ApiUtils}
-import org.apache.kyuubi.session.{KyuubiSession, SessionHandle}
+import org.apache.kyuubi.session.{KyuubiSession, KyuubiSessionManager, SessionHandle}
 
 @Tag(name = "Admin")
 @Produces(Array(MediaType.APPLICATION_JSON))
 private[v1] class AdminResource extends ApiRequestContext with Logging {
-
+  private var killMessage: KillResponse = (false, "UNKNOWN")
```
wForget, line 48: Why is it defined outside of kill method?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -255,7 +261,7 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
     }
     val engine = normalizeEngineInfo(userName, engineType, shareLevel, subdomain, "default")
     val engineSpace = calculateEngineSpace(engine)
-
+    var msg = s"Engine $engineSpace is deleted successfully."
     withDiscoveryClient(fe.getConf) { discoveryClient =>
```
wForget, line 265: kill gracefully only when forceKill is false.

zhaohehuhu:
Done. Thanks

zhaohehuhu:
plz review it again @wForget

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -255,24 +260,44 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
     }
     val engine = normalizeEngineInfo(userName, engineType, shareLevel, subdomain, "default")
     val engineSpace = calculateEngineSpace(engine)
+    var msg = s"Engine $engineSpace is deleted successfully."
+    if (forceKill) {
+      if (StringUtils.isBlank(refId)) {
+        error(s"Invalid refId: $refId")
+        throw new NotFoundException(s"Invalid refId: $refId")
```
wForget, line 267: change to `IllegalArgumentException`

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -255,24 +260,44 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
     }
     val engine = normalizeEngineInfo(userName, engineType, shareLevel, subdomain, "default")
     val engineSpace = calculateEngineSpace(engine)
+    var msg = s"Engine $engineSpace is deleted successfully."
+    if (forceKill) {
+      if (StringUtils.isBlank(refId)) {
+        error(s"Invalid refId: $refId")
```
wForget, line 266: remove this log

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -255,24 +260,44 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
+        Option(sparK8Context),
+        Option(sparK8Namespace))
+
+      var killMessage: KillResponse = (false, "UNKNOWN")
+      killMessage = fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]
+        .applicationManager.killApplication(applicationManagerInfo, refId)
+      if (!killMessage._1) {
+        msg = s"Engine $engineSpace failed to get deleted forcibly," +
+          s"cause ${killMessage._2}"
+        error(msg)
```
wForget, line 281: Do we need to throw an exception if kill fails?

zhaohehuhu: done. Thanks！

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -246,7 +246,12 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
       @QueryParam("sharelevel") shareLevel: String,
       @QueryParam("subdomain") subdomain: String,
       @QueryParam("proxyUser") kyuubiProxyUser: String,
-      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String): Response = {
+      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String,
+      @QueryParam("forceKill") forceKill: Boolean,
```
wForget, line 250: add default value
```
@QueryParam("forceKill") @DefaultValue("false") forceKill: Boolean,
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -255,24 +260,44 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
-            error(s"Failed to delete engine node:$nodePath", e)
-            throw new NotFoundException(s"Failed to delete engine node:$nodePath," +
-              s"${e.getMessage}")
+      val applicationManagerInfo = ApplicationManagerInfo(
+        Option(sparkMaster),
+        Option(sparK8Context),
+        Option(sparK8Namespace))
+
+      var killMessage: KillResponse = (false, "UNKNOWN")
+      killMessage = fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]
```
wForget, line 275: use `val`
```
    val killMessage = fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]
```

wForget (commented):
Overall LGTM, cc @turboFei could you take a look?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
zhaohehuhu: Got it

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
zhaohehuhu: killMessage need to be assigned a new value next line. Could we keep var here ?

wForget: > killMessage need to be assigned a new value next line. Could we keep var here ?

which line?

zhaohehuhu: killMessage = fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]

zhaohehuhu: Done. Thanks!!

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -246,7 +246,12 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
       @QueryParam("sharelevel") shareLevel: String,
       @QueryParam("subdomain") subdomain: String,
       @QueryParam("proxyUser") kyuubiProxyUser: String,
-      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String): Response = {
+      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String,
+      @QueryParam("forceKill") @DefaultValue("false") forceKill: Boolean,
+      @QueryParam("refId") refId: String,
+      @QueryParam("sparK8Context") sparK8Context: String,
```
turboFei, line 252: I am not sure why here is specified for spark.

zhaohehuhu:
We're using Spark engine in most of time, so we would like a way to kill Spark engine forcibly thus fail to take other engines into account. Maybe I can make some improvements to support multiple engines by passing the ApplicationManagerInfo instead of Spark Info.
@turboFei

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -246,7 +246,12 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
       @QueryParam("sharelevel") shareLevel: String,
       @QueryParam("subdomain") subdomain: String,
       @QueryParam("proxyUser") kyuubiProxyUser: String,
-      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String): Response = {
+      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String,
+      @QueryParam("forceKill") @DefaultValue("false") forceKill: Boolean,
+      @QueryParam("refId") refId: String,
```
turboFei, line 251: refId is not needed, we can get it with user and specified `sharelevel`, `subdomain`.

turboFei: otherwise, Kyuubiserver can not check whether you have permission to kill the engine.

You can specify any refId to kill, it does not make senses.

turboFei: See DiscoveryClient::getServiceNodesInfo

zhaohehuhu: OK, that makes sense. I can get refId by below listEngines method, right?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -255,24 +260,47 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
+              s"cause ${killMessage._2}"
+            error(msg)
+            throw new NotFoundException(msg)
+          }
+      })
+    } else {
+      withDiscoveryClient(fe.getConf) { discoveryClient =>
+        val engineNodes = discoveryClient.getChildren(engineSpace)
+        engineNodes.foreach { node =>
+          val nodePath = s"$engineSpace/$node"
```
turboFei, line 290: how about checking `forceKill` inside here?

zhaohehuhu: OK. Let me take a look.

zhaohehuhu:
Could you plz review it again ? @turboFei

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -255,24 +259,47 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
+          })
+        } else {
+          info(s"Deleting engine node:$nodePath")
+          try {
+            discoveryClient.delete(nodePath)
+          } catch {
+            case e: Exception =>
+              error(s"Failed to delete engine node:$nodePath", e)
+              throw new NotFoundException(s"Failed to delete engine node:$nodePath," +
+                s"${e.getMessage}")
```
turboFei, line 297: FYI:
```
....
    withDiscoveryClient(fe.getConf) { discoveryClient =>
      val engineNodes = discoveryClient.getServiceNodesInfo(engineSpace, silent = true)
      engineNodes.foreach { engineNode =>
        val nodePath = s"$engineSpace/${engineNode.nodeName}"
        info(s"Deleting engine node:$nodePath")
        try {
          discoveryClient.delete(nodePath)
        } catch {
          case e: Exception =>
            error(s"Failed to delete engine node:$nodePath", e)
            throw new NotFoundException(s"Failed to delete engine node:$nodePath," +
              s"${e.getMessage}")
        }

        if (forceKill) {
          val applicationManagerInfo = ApplicationManagerInfo(
            Option(resourceManager),
            Option(kubernetesContext),
            Option(kubernetesNamespace))
          engineNode.engineRefId.foreach { engineRefId =>
            fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]
              .applicationManager.killApplication(applicationManagerInfo, engineRefId)
.....
          }
        }
      }
    }
....
```

zhaohehuhu: Great ! So we have to delete the path and then kill the application.

turboFei: yes

zhaohehuhu: Thanks for your guidance.

zhaohehuhu:
Thanks all for reviewing this pr@wForget @turboFei

zhaohehuhu:
I fixed UT. Plz take a look @turboFei

turboFei (approved):
LGTM, cc @wForget @pan3793

pan3793:
a little bit concern about the design, which exposed too many details to end users, let me think this weekend

turboFei:
How about register the app mgr into to discovery service?

https://github.com/apache/kyuubi/pull/6360/commits/4cf2852ebb2343a47c7e5a182a58832076c0255b

pan3793:
@turboFei @zhaohehuhu Sorry for forgetting this one, the new approach lgtm, let fix the UT the get it in.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -3544,7 +3544,12 @@ object KyuubiConf {
       .version("1.8.0")
       .stringConf
       .toSequence()
-      .createWithDefault(Seq("spark.driver.memory", "spark.executor.memory"))
+      .createWithDefault(Seq(
+        "spark.driver.memory",
+        "spark.executor.memory",
+        "spark.kubernetes.context",
+        "spark.kubernetes.namespace",
+        "spark.master"))
```
pan3793, line 3552: we don't need to expose those extra attrs as we decided to use `kyuubi.engine.appMgrInfo` to stored a base64 encoded string

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -277,7 +278,8 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
       @QueryParam("sharelevel") shareLevel: String,
       @QueryParam("subdomain") subdomain: String,
       @QueryParam("proxyUser") kyuubiProxyUser: String,
-      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String): Response = {
+      @QueryParam("hive.server2.proxy.user") hs2ProxyUser: String,
+      @QueryParam("forceKill") @DefaultValue("false") forceKill: Boolean): Response = {
```
pan3793, line 282:
```suggestion
      @QueryParam("kill") @DefaultValue("false") kill: Boolean): Response = {
```

pan3793:
the failed UTs are related, because this PR appends a new configuration to the engine launch command, they should be easy to debug.

pan3793:
Please don't forget to update the RESTful API docs.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
zhaohehuhu: Got it. Thanks.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
zhaohehuhu: Got it. Thanks.

zhaohehuhu:
> Please don't forget to update the RESTful API docs.

Done.

zhaohehuhu:
> the failed UTs are related, because this PR appends a new configuration to the engine launch command, they should be easy to debug.

OK. I'm looking into it

docs/client/rest/rest_api.md
```diff
@@ -457,13 +457,14 @@ Delete the specified engine.
...
-| proxyUser               | the proxy user to impersonate | String(optional) |
-| hive.server2.proxy.user | the proxy user to impersonate | String(optional) |
+| Name                    | Description                         | Type              |
+|:------------------------|:------------------------------------|:------------------|
+| type                    | the engine type                     | String(optional)  |
+| sharelevel              | the engine share level              | String(optional)  |
+| subdomain               | the engine subdomain                | String(optional)  |
+| proxyUser               | the proxy user to impersonate       | String(optional)  |
+| hive.server2.proxy.user | the proxy user to impersonate       | String(optional)  |
+| kill                    | whether to kill the engine forcibly | Boolean(optional) |
```
pan3793, line 467: let's mention the default value is false

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
```diff
@@ -143,4 +152,20 @@ object ApplicationManagerInfo {
...
+  }
+
+  def deserialize(encodedStr: String): ApplicationManagerInfo = {
+    try {
+      val json = new String(
+        Base64.getDecoder.decode(encodedStr.getBytes),
+        StandardCharsets.UTF_8)
+      mapper.readValue(json, classOf[ApplicationManagerInfo])
+    } catch {
+      case _: Throwable => ApplicationManagerInfo(None)
```
pan3793, line 168: log the original encodedStr

docs/client/rest/rest_api.md
zhaohehuhu: Got it.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
zhaohehuhu: OK. Thanks.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
```diff
@@ -143,4 +153,21 @@ object ApplicationManagerInfo {
...
   }
+
+  def serialize(appMgrInfo: ApplicationManagerInfo): String = {
+    Base64.getEncoder.encodeToString(
+      mapper.writeValueAsString(appMgrInfo).getBytes(StandardCharsets.UTF_8))
+  }
+
+  def deserialize(encodedStr: String): ApplicationManagerInfo = {
+    try {
+      info(s"The original string encoded:$encodedStr")
```
pan3793, line 164: move to catch block. we only need to print the base64 string when something goes wrong

zhaohehuhu: Thanks. Done

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
```diff
@@ -143,4 +153,22 @@ object ApplicationManagerInfo {
...
+
+  def deserialize(encodedStr: String): ApplicationManagerInfo = {
+    try {
+      val json = new String(
+        Base64.getDecoder.decode(encodedStr.getBytes),
+        StandardCharsets.UTF_8)
+      mapper.readValue(json, classOf[ApplicationManagerInfo])
+    } catch {
+      case _: Throwable =>
+        error(s"The original string encoded:$encodedStr")
```
pan3793, line 170:
```suggestion
        error(s"Fail to desirable the encoded string: $encodedStr")
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
```diff
@@ -143,4 +153,22 @@ object ApplicationManagerInfo {
...
+
+  def deserialize(encodedStr: String): ApplicationManagerInfo = {
+    try {
+      val json = new String(
+        Base64.getDecoder.decode(encodedStr.getBytes),
+        StandardCharsets.UTF_8)
+      mapper.readValue(json, classOf[ApplicationManagerInfo])
+    } catch {
+      case _: Throwable =>
+        error(s"Fail to desirable the encoded string: $encodedStr")
```
pan3793, line 170:
```suggestion
        error(s"Fail to deserialize the encoded string: $encodedStr")
```

kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/AdminResourceSuite.scala
```diff
@@ -301,6 +301,57 @@ class AdminResourceSuite extends KyuubiFunSuite with RestFrontendTestHelper {
...
+        assert(client.getChildren(engineSpace).isEmpty, s"refId same with $id?")
+      }
+
+      eventually(timeout(30.seconds), interval(100.milliseconds)) {
+        assert(engineMgr.getApplicationInfo(
+          ApplicationManagerInfo(
+            None,
+            KubernetesInfo(None, None)),
+          id)
+          .exists(_.state == ApplicationState.NOT_FOUND))
```
pan3793, line 350:
```suggestion
        val appMgrInfo = ApplicationManagerInfo(None, KubernetesInfo(None, None))
        assert(engineMgr.getApplicationInfo(appMgrInfo, id)
          .exists(_.state == ApplicationState.NOT_FOUND))
```

pan3793 (approved):

pan3793:
cc @turboFei, this patch now adopts your suggestions, please take a look when you have time.

turboFei (approved):

turboFei:
would you update the AdminRestApi and CommandLine?

pan3793:
Thanks, merged to master

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -286,24 +288,38 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
 
     withDiscoveryClient(fe.getConf) { discoveryClient =>
-      val engineNodes = discoveryClient.getChildren(engineSpace)
-      engineNodes.foreach { node =>
-        val nodePath = s"$engineSpace/$node"
+      val engineNodes = discoveryClient.getServiceNodesInfo(engineSpace, silent = true)
+      engineNodes.foreach { engineNode =>
+        val nodePath = s"$engineSpace/${engineNode.nodeName}"
+        val engineRefId = engineNode.engineRefId.orNull
         info(s"Deleting engine node:$nodePath")
```
HaoYang670, line 298: Sorry for the late comment. 
the `engineSpace` is not cleaned right? Why don't we delete it?

pan3793: except for CONNECTION level engine, we cannot guarantee that there are no other engines access the same `engineSpace`

HaoYang670: Do you mean there can be multiple engines under same subdomain?

HaoYang670: I don't think so. Even in GROUP level, we can only create one engine per subdomain, right. And subdomain has been included in `engineSpace` right?

pan3793: there are some corner cases. for example, if the engine has no response due to full GC or overload, kyuubi will create a new one on the same engine space. I also see sometimes the dist lock does not work properly(which should not, may be bugs), that also causes multi engines live in the same subdomain. finally, in theory, there are race conditions for kyuubi server deleting and engine registering under the same subdomain.

HaoYang670: Thanks for the explanation!

Issue #5367, opened 2023-10-08
[TASK][EASY] Spark crashes with ClassCastException when resolving a join of masked tables


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [ ] I have sufficient knowledge and experience of this task, and I volunteer to be the mentor of this task to guide contributors to complete the task.

### Skill requirements

- Knowledge about Spark Catalyst
- Basic Knowledge of SQL authorization model, e.g. Apache Ranger

### Background and Goals

Spark authz plugin may generate an invalid query plan which fails the query
https://github.com/apache/kyuubi/issues/5092

### Implementation steps

Analysis the stack trace and fix the invalid query plan

### Additional context

Original reporter is @yaooqinn

xza-m:
Hello, I am interested in this task, can i try? I estimate it will take two weeks.

pan3793:
@xza-m thanks, assigned

PR #6012, opened 2024-01-24
[KYUUBI #5367] Spark crashes with ClassCastException when resolving a join of masked tables

### LINKED ISSUE #5092: [Bug] [AUTHZ] Spark crashes with ClassCastException when resolving a join of masked tables
### Code of Conduct

### Search before asking

### Describe the bug

Hi, we're using your `spark-authz` lib which started crashing for some workflows after upgrading 1.6.1 -> 1.7.1
This only happens when joining a table that contains masked columns even though they're not used in the join condition. 

```
Exception in thread "main" java.lang.ClassCastException: class org.apache.kyuubi.plugin.spark.authz.ranger.datamasking.DataMaskingStage1Marker cannot be cast to class org.apache.spark.sql.catalyst.plans.logical.Join (org.apache.kyuubi.plugin.spark.authz.ranger.datamasking.DataMaskingStage1Marker and org.apache.spark.sql.catalyst.plans.logical.Join are in unnamed module of loader 'app')
 	at org.apache.spark.sql.Dataset.resolveSelfJoinCondition(Dataset.scala:1069)
 	at org.apache.spark.sql.Dataset.join(Dataset.scala:1117)
       [..] our code
```

This should be enough to reproduce it:
```
val df0 = sqlContext.table("table_with_masked_col") // only col2 is masked which is not used
      .select($"col0", $"col1")

    df0.as("a")
      .join(
        right = df0.as("b"),
        joinExprs = $"a.col0" === $"b.col0" && $"a.col1" === $"b.col1",
        joinType = "left_outer") // crashes
```

We had to downgrade to 1.6.1 where this doesn't happen.

### Affects Version(s)

1.7.1

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

pilietis93:
# :mag: Description
## Issue References 🔗

This pull request fixes #5092

Spark crashes with ClassCastException when resolving a join of masked tables

## Describe Your Solution 🔧

As per @sauliusvl [comment](https://github.com/apache/kyuubi/pull/5468#issuecomment-1831774255) in this unfinished [PR](https://github.com/apache/kyuubi/pull/5468) `Stage0Marker` column ids are not updated on the right side of the join which leads to `Stage1Marker` being incorrectly applied which in turn causes 
```
java.lang.ClassCastException: org.apache.kyuubi.plugin.spark.authz.ranger.datamasking.DataMaskingStage1Marker cannot be cast to org.apache.spark.sql.catalyst.plans.logical.Join
```

So the solution was to copy the scan plan together with the child plan in `withNewChildInternal` method so that column ids are aligned.

Self join of tables containing masked columns using `joinExprs` arg fails with ClassCastException.

Self join of tables containing masked columns using `joinExprs` arg succeeds.

#### Related Unit Tests
DataMaskingTestBase

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/datamasking/DataMaskingStage0Marker.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/datamasking/DataMaskingStage0Marker.scala
@@ -18,9 +18,10 @@
 package org.apache.kyuubi.plugin.spark.authz.rule.datamasking
 
 import org.apache.spark.sql.catalyst.expressions.{Attribute, ExprId}
-import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, UnaryNode}
+import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, Project, UnaryNode}
 
 import org.apache.kyuubi.plugin.spark.authz.util.WithInternalChild
+
 case class DataMaskingStage0Marker(child: LogicalPlan, scan: LogicalPlan)
   extends UnaryNode with WithInternalChild {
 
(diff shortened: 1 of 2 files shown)


pan3793 (approved):
The change makes sense to me, thanks for contribution!

cxzl25 (approved):

pan3793:
seems the test failed in Spark 3.1 and 3.5

PR #6020, opened 2024-01-25
Introduce service stop graceful period to prevent stuck because of active connections

turboFei:
# :mag: Description
## Issue References 🔗

This pr introduce service stop graceful period to prevent stuck because of active connections.

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
```diff
@@ -125,10 +125,6 @@ case class SparkSQLEngine(spark: SparkSession) extends Serverable("SparkSQLEngin
       if (!shutdown.get) {
         info(s"Spark engine is de-registering from engine discovery space.")
         frontendServices.flatMap(_.discoveryService).foreach(_.stop())
-        while (backendService.sessionManager.getOpenSessionCount > 0) {
-          Thread.sleep(1000 * 60)
-        }
-        info(s"Spark engine has no open session now, terminating.")
```
turboFei, line 131: This is duplicated, so I remove it.

PR #6036, opened 2024-01-31
JDBC driver conditional sets fetchSize on opening session

pan3793:
# :mag: Description
## Issue References 🔗

I get reported by a user that using Kyuubi JDBC driver 1.8.0 to access Spark Thrift Server 2.4 (with Hive 1.2.1) with getting an error on opening the session even with `clientProtocol=7`

```
org.apache.hive.service.cli.HiveSQLException: java.lang.IllegalArgumentException: hive configuration hive.server2.thrift.resultset.default.fetch.size does not exists.
        at org.apache.hive.service.cli.session.HiveSessionImpl.configureSession(HiveSessionImpl.java:220)
        at org.apache.hive.service.cli.session.HiveSessionImpl.open(HiveSessionImpl.java:154)
        at org.apache.hive.service.cli.session.SessionManager.openSession(SessionManager.java:258)
        ... 13 more
```

## Describe Your Solution 🔧

When `hive.conf.validation` is `true` (it also is the default value), an IllegalArgumentException will be thrown if the provided configurations start with `hive.` and can not be recognized by the server.

One solution is to disable `hive.conf.validation` on the server side, but we can also address it by avoiding passing this configuration if we know that the server does not support it.

HIVE-14901 (2.3.0, HIVE_CLI_SERVICE_PROTOCOL_V10) introduces `hive.server2.thrift.resultset.default.fetch.size`, so we can set this configuration only when protocol >= HIVE_CLI_SERVICE_PROTOCOL_V10

## Test Plan 🧪

Verified locally by connecting to HS2 2.1.1 and HS2 2.3.9

IllegalArgumentException throws when using  `jdbc:hive2://localhost:10000/default;clientProtocol=8` to connect HS2 2.1.1

Everything is OK when using `jdbc:hive2://localhost:10000/default` to connect HS2 2.3.9

Everything is OK when using `jdbc:hive2://localhost:10000/default;clientProtocol=8` to connect HS2 2.1.1

Everything is OK when using `jdbc:hive2://localhost:10000/default` to connect HS2 2.3.9

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -746,13 +746,13 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
               "Unsupported Hive2 protocol version %s specified by session conf key %s",
               clientProtocolStr, CLIENT_PROTOCOL_VERSION));
     }
+    openReq.setClient_protocol(clientProtocol);
     // HIVE-14901: set the fetchSize
     if (clientProtocol.compareTo(TProtocolVersion.HIVE_CLI_SERVICE_PROTOCOL_V10) >= 0) {
       openConf.put(
           "set:hiveconf:hive.server2.thrift.resultset.default.fetch.size",
           Integer.toString(fetchSize));
     }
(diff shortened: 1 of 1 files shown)


zyx199693 (approved):
look good to me

cxzl25 (approved):

pan3793:
Thanks, merged to master/1.8

PR #6046, opened 2024-02-04
[K8S][HELM] Add Hadoop configuration support

dnskr:
# :mag: Description

## Issue References 🔗
The PR adds basic support for Hadoop configuration files to be used by Apache Kyuubi, Apache Spark etc.

## Describe Your Solution 🔧

The PR adds:
 - Apache Hadoop related `ConfigMap` that has to be mounted to Kyuubi pods as files.
 - `HADOOP_CONF_DIR` environment variable to Kyuubi pods.

Notes:
- The `ConfigMap` change restarts Kyuubi pods!
- Only `core-site.xml` and `hdfs-site.xml` files are added. Not sure `yarn-site.xml`, `mapred-site.xml` or other files should be added. Please, let me know if so.

## Test Plan 🧪

Install the chart with configured Spark and the following values:
```yaml
# Hadoop configuration files
hadoopConf:
  # The value (templated string) is used for core-site.xml file
  # See https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/resources/core-default.xml and Hadoop documentation for more details
  coreSite: |
    <?xml version="1.0"?>
    <?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
    <configuration>
      <property>
        <name>test.core.site</name>
        <value>test-core-site</value>
      </property>
    </configuration>

  # The value (templated string) is used for hdfs-site.xml file
  # See https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/resources/hdfs-default.xml and Hadoop documentation for more details
  hdfsSite: |
    <?xml version="1.0"?>
    <?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
    <configuration>
      <property>
        <name>test.hdfs.site</name>
        <value>test-hdfs-site</value>
      </property>
    </configuration>
```
Run `beeline` from Kyuubi pod:
```shell
./bin/beeline -u 'jdbc:hive2://kyuubi-thrift-binary:10009' -n test-user
```
Check specified property values:
```shell
0: jdbc:hive2://kyuubi-thrift-binary:10009> set test.core.site;
+-----------------+-----------------+
|       key       |      value      |
+-----------------+-----------------+
| test.core.site  | test-core-site  |
+-----------------+-----------------+


0: jdbc:hive2://kyuubi-thrift-binary:10009> set test.hdfs.site;
+-----------------+-----------------+
|       key       |      value      |
+-----------------+-----------------+
| test.hdfs.site  | test-hdfs-site  |
+-----------------+-----------------+
```

---

# Checklist 📝

**Be nice. Be informative.**


charts/kyuubi/templates/kyuubi-configmap-hadoop.yaml
```diff
@@ -0,0 +1,32 @@
...
+  name: {{ .Release.Name }}-hadoop
+  labels:
+    {{- include "kyuubi.labels" . | nindent 4 }}
+data:
+  {{- with .Values.hadoopConf.coreSite }}
+  core-site.xml: |
+    {{- tpl . $ | nindent 4 }}
+  {{- end }}
+  {{- with .Values.hadoopConf.hdfsSite }}
+  hdfs-site.xml: |
```
pan3793, line 30: Actually, it's likely to put additional configuration files under HADOOP_CONF_DIR in practice, for example, `hive-site.xml`, `hbase-site.xml`, so we may want to allow the user to inject custom files.

Another concern is that embedded large-XML into YAML may be not user-friendly, it's easy to mess up the indent.

chgl: Definitely agree. The approach I often see is providing both options: either provide a (templated) inline config - which as you said can get unwieldy fast - or provide the name of an existing config map. The latter allows creating that config map from a file or some other way, without relying on functionality in the chart.

pan3793: @chgl are there examples of some projects that have similar cases?

chgl: The ones that I am aware of: 

- https://github.com/bitnami/charts/blob/main/bitnami/zookeeper/values.yaml#L203-L214, - https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L293-L309. 
- Superset: https://github.com/apache/superset/blob/master/helm/superset/values.yaml#L135 (fairly extreme since entire python snippets are inlined) with https://github.com/apache/superset/blob/master/helm/superset/values.yaml#L50 as an option to reference an existing secret.
- akhq: https://github.com/tchiotludo/akhq/blob/dev/helm/akhq/values.yaml#L36-L46

dnskr:
Closed in favor to:
 - https://github.com/apache/kyuubi/issues/6123
 - https://github.com/apache/kyuubi/pull/6521

PR #6027, opened 2024-01-29
[KYUUBI #5674] Add PostgreSQL as backend database for kyuubi metadata

tigrulya-exe:
# :mag: Description
## Issue References 🔗

This pull request fixes #5674

## Describe Your Solution 🔧

Currently, Kyuubi supports the use of SQLite and MySQL to store metadata. Although PostgreSQL can be supported through the 'CUSTOM' method, there is a certain usage cost for users, therefore this MR adds support for PostgreSQL as a backend database for kyuubi metadata.

## Types of changes :bookmark:

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Related Unit Tests
`org.apache.kyuubi.server.metadata.jdbc.JDBCMetadataStoreSuite`

---

# Checklist 📝

- [x] This patch was not authored or co-authored using [Generative Tooling](https://www.apache.org/legal/generative-tooling.html)

**Be nice. Be informative.**

Issue #5674, opened 2023-11-12
[FEATURE] Add PostgreSQL as backend database for kyuubi metadata


lukeyan2023:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add PostgreSQL as Kyuubi's metadata storage database

### Motivation

Currently, Kyuubi supports the use of SQLite and MySQL to store metadata. Although PostgreSQL can be supported through the 'CUSTOM' method, there is a certain usage cost for users. PostgreSQL ，as one of the most widely used databases,  if it can be used out of the box, will further lower the user's usage threshold

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

pan3793:
@lukeyan2023 Great, looking forward your contribution～

tigrulya-exe:
@pan3793 Hi! I've implemented this feature and opened PR #6027. Could you take a look, please?

pan3793:
@tigrulya-exe thanks, I'm looking

PR #6066, opened 2024-02-20
Fix PostgreSQL metastore migration script failure if indexes already exist

tigrulya-exe:
# :mag: Description
## Issue References 🔗

This pull request is a small fix for #5674 

## Describe Your Solution 🔧

PostgreSQL support as backend for metadata store was added in #6027. If we choose it and run schema init scripts several times, then the following errors will be logged: `Error executing sql: CREATE UNIQUE INDEX unique_identifier_index ON metadata(identifier), with params: . ERROR: relation "unique_identifier_index" already exists`. This PR adds lacking `IF NOT EXISTS` to db indexes creation ddls in the PostgreSQL metastore init script.

#### Related Unit Tests

---

# Checklist 📝


pan3793 (approved):

pan3793:
Thanks, merged to master

PR #6077, opened 2024-02-22
[KYUUBI #6070] Improve perf on assembling row-based TRowSet

### LINKED ISSUE #6070: [Improvement] Performance Improvement for converting spark rows to thrift rows
### Code of Conduct

### Search before asking

https://issues.apache.org/jira/browse/SPARK-47085 reported an issue, scala Seq.apply has O(n) complexity so that we shall avoid it

### How should we improve?

_No response_

Kwafoor:
# :mag: Description
## Issue References 🔗

This pull request fixes #6070

## Describe Your Solution 🔧

https://issues.apache.org/jira/browse/SPARK-47085

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TRowSetGenerator.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TRowSetGenerator.scala
@@ -18,6 +18,8 @@
 package org.apache.kyuubi.engine.result
 import java.util.{ArrayList => JArrayList}
 
+import scala.collection.JavaConverters._
+
 import org.apache.kyuubi.shaded.hive.service.rpc.thrift._
 
 trait TRowSetGenerator[SchemaT, RowT, ColumnT]
@@ -40,23 +42,17 @@ trait TRowSetGenerator[SchemaT, RowT, ColumnT]
   }
 
   def toRowBasedSet(rows: Seq[RowT], schema: SchemaT): TRowSet = {
-    val rowSize = rows.length
-    val tRows = new JArrayList[TRow](rowSize)
-    var i = 0
-    while (i < rowSize) {
-      val row = rows(i)
+    val tRows = rows.map { row =>
+      val tRow = new TRow()
       var j = 0
       val columnSize = getColumnSizeFromSchemaType(schema)
-      val tColumnValues = new JArrayList[TColumnValue](columnSize)
       while (j < columnSize) {
         val columnValue = toTColumnValue(row, j, schema)
-        tColumnValues.add(columnValue)
+        tRow.addToColVals(columnValue)
         j += 1
       }
-      i += 1
-      val tRow = new TRow(tColumnValues)
-      tRows.add(tRow)
-    }
+      tRow
+    }.asJava
     new TRowSet(0, tRows)
   }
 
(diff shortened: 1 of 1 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TRowSetGenerator.scala
```diff
@@ -65,13 +61,11 @@ trait TRowSetGenerator[SchemaT, RowT, ColumnT]
     val tRowSet = new TRowSet(0, new JArrayList[TRow](rowSize))
     var i = 0
     val columnSize = getColumnSizeFromSchemaType(schema)
-    val tColumns = new JArrayList[TColumn](columnSize)
```
pan3793, line 68: it's by design, see https://github.com/apache/kyuubi/pull/5831

pan3793: ditto in row based implementation

Kwafoor: sorry for mistake , thanks.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TRowSetGenerator.scala
```diff
@@ -40,11 +42,7 @@ trait TRowSetGenerator[SchemaT, RowT, ColumnT]
   }
 
   def toRowBasedSet(rows: Seq[RowT], schema: SchemaT): TRowSet = {
-    val rowSize = rows.length
-    val tRows = new JArrayList[TRow](rowSize)
-    var i = 0
-    while (i < rowSize) {
-      val row = rows(i)
+    val tRows = rows.map { row =>
       var j = 0
```
pan3793, line 46: since there is no `i`, can we replace `j` with `i`?

pan3793 (approved):

cxzl25 (approved):

pan3793:
Thanks, merged to master

yaooqinn:
TColumnGenerator.getColumnToList was missed?

hh-cn:
@beryllw @yaooqinn 

I'm sorry to bother you, but it seems that this issue hasn't been fully resolved. The logic related to toRowBasedSet has been fixed, but in the toColumnBasedSet, the getColumnToList method still involves similar access to a non-IndexedSeq with val row = rows(idx). This results in significantly impacted serialization speed when Hive JDBC statements have a large fetchSize (> 300).

> trait TColumnGenerator[RowT] extends TRowSetColumnGetter[RowT] {
>   protected def getColumnToList[T](
>       rows: Seq[RowT],
>       ordinal: Int,
>       defaultVal: T,
>       convertFunc: (RowT, Int) => T = null): (JList[T], ByteBuffer) = {
>     val rowSize = rows.length
>     val ret = new JArrayList[T](rowSize)
>     val nulls = new JBitSet()
>     var idx = 0
>     while (idx < rowSize) {
>       val row = rows(idx)
>       val isNull = isColumnNullAt(row, ordinal)
>       if (isNull) {
>         nulls.set(idx, true)
>         ret.add(defaultVal)
>       } else {
>         val value = Option(convertFunc) match {
>           case Some(f) => f(row, ordinal)
>           case _ => getColumnAs[T](row, ordinal)
>         }
>         ret.add(value)
>       }
>       idx += 1
>     }
>     (ret, ByteBuffer.wrap(nulls.toByteArray))
>   }

yaooqinn:
@hh-cn Can you send a PR for that issue?

hh-cn:
> @hh-cn Can you send a PR for that issue?

I am not familiar with the process and standards for submitting a PR. I will glad to see you fixing it soon.

yaooqinn:
Can you create an issue then? I'm fully booked and don't have time to fix this. But I'm sure someone will be interested in fixing it.

PR #6080, opened 2024-02-23
Improve instruction for deploy Kyuubi on K8s

pan3793:
# :mag: Description
## Issue References 🔗

Historically, we provided a few standalone YAML files to allow users to deploy Kyuubi on K8s, and later a Helm Chart was introduced, which is a simple and mature way for production use cases.

## Describe Your Solution 🔧

Remove the legacy standalone YAML files way and update docs to guide user to use Helm chart to deploy Kyuubi on K8s.

## Test Plan 🧪

Review

---

# Checklist 📝

**Be nice. Be informative.**


charts/kyuubi/values.yaml
```diff
@@ -45,7 +45,7 @@ updateStrategy:
 image:
   repository: apache/kyuubi
   pullPolicy: IfNotPresent
-  tag: ~
+  tag: 1.8.1-spark
```
pan3793, line 48: use an image that contains Spark binary as the default value.

docker/Dockerfile
```diff
@@ -15,14 +15,12 @@
...
 
-# Usage:
-#   1. use ./build/dist to make binary distributions of Kyuubi or download a release
-#   2. Untar it and run the docker command below
-#      docker build -f docker/Dockerfile -t repository/kyuubi:tagname .
-#   Options:
-#     -f this docker file
-#     -t the target repo and tag name
-#     more options can be found with -h
+# It's recommended to use ./bin/docker-image-tool.sh to build the image rather than the vanilla docker command.
```
pan3793, line 18: cc @zwangsheng, am I right?

charts/kyuubi/values.yaml
dnskr: I would prefer to leave `~` for now to use `{{ .Chart.AppVersion }}` as image tag value by default.
The change makes change, however it's not enough to deploy Kyuubi Helm chart with Spark support.
Let me create `Minimal Helm chart configuration to run Kyuubi with Spark` issue to discuss all problems, limitations and options there.

docker/Dockerfile
zwangsheng: Yes, although we may need to optimize this script for now, we still recommend that users use script builds to reduce the complexity of building docker images.

zwangsheng (approved):
LGTM

zwangsheng:
BYW, should we mentions update `docker-image-tool.sh` in release guide?

charts/kyuubi/values.yaml
pan3793: reverted

pan3793:
Thanks, merged to master

Issue #6078, opened 2024-02-22
[Bug] failed to insert data to hive partition table if using `INSERT INTO` for spark-hive-connector


FANNG1:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

failed to insert data to hive partition table if using `INSERT INTO` for spark-hive-connector, could be reproduced by replace `INSERT OVERWRITE` with `INSERT INTO` in HiveQuerySuite.scala
```
  private def readPartitionedTable(format: String, hiveTable: Boolean): Unit = {
    withSparkSession() { spark =>
      val table = "hive.default.employee"
      withTempPartitionedTable(spark, table, format, hiveTable) {
        spark.sql(
          s"""
             | INSERT INTO
             | $table PARTITION(year = '2023')
             | VALUES("zhao", "09")
             |""".stripMargin)
        checkQueryResult(s"select * from $table", spark, Array(Row.apply("zhao", "2023", "09")))
      }
    }
  }
```

### Affects Version(s)

master

### Kyuubi Engine Log Output

```logtalk
org.apache.kyuubi.spark.connector.hive.KyuubiHiveConnectorException: Dynamic partition strict mode requires at least one static partition column. To turn this off set hive.exec.dynamic.partition.mode=nonstrict
	at org.apache.kyuubi.spark.connector.hive.write.HiveWrite.extractAndValidatePartitionCols(HiveWrite.scala:205)
	at org.apache.kyuubi.spark.connector.hive.write.HiveWrite.toBatch(HiveWrite.scala:84)
	at org.apache.spark.sql.execution.datasources.v2.V2ExistingTableWriteExec.run(WriteToDataSourceV2Exec.scala:360)
	at org.apache.spark.sql.execution.datasources.v2.V2ExistingTableWriteExec.run$(WriteToDataSourceV2Exec.scala:359)
```

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

pan3793:
@Yikf do u have time to take a look?

Yikf:
[...]

PR #6082, opened 2024-02-25
[KYUUBI #6078] KSHC should handle the commit of the partitioned table as dynamic partition at write path

Yikf:
# :mag: Description
## Issue References 🔗

This pull request fixes https://github.com/apache/kyuubi/issues/6078, KSHC should handle the commit of the partitioned table as dynamic partition at write path, that's beacuse the process of writing with Apache Spark DataSourceV2 using dynamic partitioning to handle static partitions.

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWriteBuilder.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWriteBuilder.scala
@@ -19,11 +19,10 @@ package org.apache.kyuubi.spark.connector.hive.write
 
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.catalog.CatalogTable
-import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
 import org.apache.spark.sql.connector.write._
-import org.apache.spark.sql.sources.{AlwaysTrue, And, EqualNullSafe, EqualTo, Filter}
+import org.apache.spark.sql.sources.Filter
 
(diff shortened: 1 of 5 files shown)


Yikf:
Please take a look if you have a time, @FANNG1 @pan3793

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWriteBuilder.scala
```diff
@@ -47,45 +41,22 @@ case class HiveWriteBuilder(
...
+      dynamicPartitionSpec())
   }
 
   override def overwrite(filters: Array[Filter]): WriteBuilder = {
-    filters match {
-      case Array(AlwaysTrue) => // no partition, do nothing
-      case _ => staticPartition = deduplicateFilters(filters)
-    }
-    overwriteDynamicPartitions()
+    forceOverwrite = true
```
pan3793, line 48: why `filters` can be ignored here?

Yikf: The current DSv2 SupportsOverwrite#overwrite is designed to filter out (delete) data that matches deleteExpr when writer commit, But in the current catalyst implementation, these filterExprs are static partition EqualNullSafe expressions, KSHC does not currently require these processing filterExprs.

pan3793: can we add an `requires` guard here?

pan3793 (approved):

pan3793:
Overall LGTM, let me merge it first, and address comment in follow-up PRs.

PR #6099, opened 2024-02-27
Remove support of Derby for Kyuubi metastore

pan3793:
# :mag: Description
## Issue References 🔗

This pull request removes support of Derby for Kyuubi metastore.

## Describe Your Solution 🔧

Previously, we migrated the embedded DB of Kyuubi metastore from Derby to SQLite, and also marked Derby as deprecated (#4950), now, I propose to remove support of Derby for Kyuubi metastore.

Note, that both Derby and SQLite are mainly for testing purposes, and they're not supposed to be used in production. Users should not be surprised by this removal.

Also, the Derby we used suffers CVE-2022-46337

> Mitigation:
>
> Users should upgrade to Java 21 and Derby 10.17.1.0.
>
> Alternatively, users who wish to remain on older Java versions should build their own Derby distribution from one of the release families to which the fix was backported: 10.16, 10.15, and 10.14. Those are the releases which correspond, respectively, with Java LTS versions 17, 11, and 8.

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Test Plan 🧪

Some tests are modified accordingly. Pass GA.

---

# Checklist 📝

**Be nice. Be informative.**


yaooqinn (approved):

kyuubi-server/pom.xml
```diff
@@ -410,6 +405,12 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.derby</groupId>
+            <artifactId>derby</artifactId>
+            <scope>test</scope>
```
pan3793, line 411: we still relies Derby to bootstrap a embedded HMS for testing Hadoop Token refresh feature.

pan3793:
Merged to master

PR #6115, opened 2024-02-28
[KYUUBI #6114] Do not throw exception in KyuubiStatement#getMoreResults(int) in case of closing current result set

### LINKED ISSUE #6114: [Bug] Do not throw exception in KyuubiStatement#getMoreResults(int) in case of closing current result set
### Code of Conduct

### Search before asking

### Describe the bug

After fixing [HIVE-7680](https://issues.apache.org/jira/browse/HIVE-7680) issue the method `getMoreResults()` of `HiveStatement` and respectively `KyuubiStatement`classes returns false instead of throwing exception. The javadoc of the `Statement#getMoreResults()`method says, that it should implicitly close any current ResultSet object, i.e. is similar to calling the `KyuubiStatement#getMoreResults(int)` method with `Statement.CLOSE_CURRENT_RESULT` argument. Therefore, in the latter case, we should also return false instead of failing with an exception.

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

tigrulya-exe:
# :mag: Description
## Issue References 🔗

This pull request fixes #6114

## Describe Your Solution 🔧

After fixing [HIVE-7680](https://issues.apache.org/jira/browse/HIVE-7680) issue the method getMoreResults() of HiveStatement and respectively KyuubiStatementclasses returns false instead of throwing exception. The javadoc of the Statement#getMoreResults()method says, that it should implicitly close any current ResultSet object, i.e. is similar to calling the KyuubiStatement#getMoreResults(int) method with Statement.CLOSE_CURRENT_RESULT argument. Therefore, in the latter case, we should also return false instead of failing with an exception. This fix also allows Kyuubi JDBC driver to be used in the JDBC engine. 

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
@@ -544,6 +544,18 @@ public class KyuubiStatement implements SQLStatement, KyuubiLoggable {
   }
 
   @Override
+  public boolean getMoreResults(int current) throws SQLException {
+    // getMoreResults() javadoc says, that it should implicitly close
+    // any current ResultSet object, so here in case of Statement.CLOSE_CURRENT_RESULT
+    // we can implement the same logic
+    // see also https://issues.apache.org/jira/browse/HIVE-7680
+    if (current == Statement.CLOSE_CURRENT_RESULT) {
+      return false;
+    }
+    throw new SQLFeatureNotSupportedException("Method not supported");
+  }
+
+  @Override
   public boolean getMoreResults() throws SQLException {
     return false;
   }
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/adapter/SQLStatement.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/adapter/SQLStatement.java
@@ -65,11 +65,6 @@ public interface SQLStatement extends Statement {
   }
 
   @Override
-  default boolean getMoreResults(int current) throws SQLException {
-    throw new SQLFeatureNotSupportedException("Method not supported");
-  }
-
-  @Override
   default ResultSet getGeneratedKeys() throws SQLException {
     throw new SQLFeatureNotSupportedException("Method not supported");
   }


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -543,6 +543,18 @@ public int getMaxRows() throws SQLException {
...
+  @Override
+  public boolean getMoreResults(int current) throws SQLException {
+    // getMoreResults() javadoc says, that it should implicitly close
+    // any current ResultSet object, so here in case of Statement.CLOSE_CURRENT_RESULT
+    // we can implement the same logic
+    // see also https://issues.apache.org/jira/browse/HIVE-7680
+    if (current == Statement.CLOSE_CURRENT_RESULT) {
+      return false;
+    }
+    throw new SQLFeatureNotSupportedException("Method not supported");
```
pan3793, line 555: let's put the args into the exception message. something like `Method not supported. (current=$current)`

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -543,6 +543,18 @@ public int getMaxRows() throws SQLException {
     return maxRows;
   }
 
+  @Override
+  public boolean getMoreResults(int current) throws SQLException {
+    // getMoreResults() javadoc says, that it should implicitly close
+    // any current ResultSet object, so here in case of Statement.CLOSE_CURRENT_RESULT
+    // we can implement the same logic
+    // see also https://issues.apache.org/jira/browse/HIVE-7680
```
pan3793, line 551: we usually use such words in comments for backport cases, but it's not. let's make it more clear.

please change the comment to something like
```suggestion
    // this is an additional enhancement on top of HIVE-7680
```

pan3793 (approved):
Almost good, just nits. Thanks!

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -543,6 +543,18 @@ public int getMaxRows() throws SQLException {
...
+    if (current == Statement.CLOSE_CURRENT_RESULT) {
+      return false;
```
pan3793, line 553: Hmm, should we perform `result.close` here?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
pan3793: The `TrinoStatement`'s implementation looks better
```
    @Override
    public boolean getMoreResults(int current)
            throws SQLException
    {
        checkOpen();

        currentUpdateCount.set(-1);
        currentUpdateType.set(null);

        if (current == CLOSE_CURRENT_RESULT) {
            closeResultSet();
            return false;
        }

        if (current != KEEP_CURRENT_RESULT && current != CLOSE_ALL_RESULTS) {
            throw new SQLException("Invalid argument: " + current);
        }

        throw new SQLFeatureNotSupportedException("Multiple open results not supported");
    }
```

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
tigrulya-exe: I think there might be reasons why this wasn't added in the `HIVE-7680` fix, but it seems like we can actually close the resultSet here. Added resultSet closing.

PR #6118, opened 2024-02-29
Migrate to shaded HMS client for getting delegation token on server

zhouyifan279:
# :mag: Description
## Describe Your Solution 🔧

Kyuubi Shaded 0.3 introduces a light kyuubi-relocated-hive-metastore-client, for refreshing 
delegation token, this PR aims to migrate from the vanilla HMS client to this light shaded HMS client, then we can get rid of Hive dependencies, especially the vulnerable thrift 0.9, from the Kyuubi server.

## Test Plan 🧪

Pass GA.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/NOTICE-binary
+++ b/NOTICE-binary
@@ -189,42 +189,6 @@ JUnit (4.12)
 
 * License: Eclipse Public License
 
-Hive Beeline
-Copyright 2022 The Apache Software Foundation
-
-Hive Common
-Copyright 2022 The Apache Software Foundation
-
-Hive JDBC
-Copyright 2022 The Apache Software Foundation
-
-Hive Llap Client
-Copyright 2022 The Apache Software Foundation
-
-Hive Metastore
-Copyright 2022 The Apache Software Foundation
-
-Hive Serde
-Copyright 2022 The Apache Software Foundation
-
-Hive Service
-Copyright 2022 The Apache Software Foundation
-
-Hive Service RPC
-Copyright 2022 The Apache Software Foundation
-
-Hive Shims 0.23
-Copyright 2022 The Apache Software Foundation
-
-Hive Shims Common
-Copyright 2022 The Apache Software Foundation
-
-Hive Standalone Metastore
-Copyright 2022 The Apache Software Foundation
-
-Hive Storage API
-Copyright 2020 The Apache Software Foundation
-
 Apache HttpClient
 Copyright 1999-2020 The Apache Software Foundation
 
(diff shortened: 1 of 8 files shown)


NOTICE-binary
```diff
@@ -189,42 +189,6 @@ JUnit (4.12)
 
 * License: Eclipse Public License
 
-Hive Beeline
```
ulysses-you, line 192: is it leaked before this pr ?

kyuubi-server/pom.xml
```diff
@@ -78,6 +78,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-relocated-hive-metastore-client</artifactId>
+            <version>${kyuubi-relocated.version}</version>
+        </dependency>
```
ulysses-you, line 85: can we add comment to say it is only used for hive delegation token ?

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HiveDelegationTokenProviderSuite.scala
```diff
@@ -118,7 +120,7 @@ class HiveDelegationTokenProviderSuite extends KerberizedTestHelper {
       assert(aliasAndToken._2 != null)
 
       val token = aliasAndToken._2
-      val tokenIdent = token.decodeIdentifier().asInstanceOf[DelegationTokenIdentifier]
+      val tokenIdent = token.decodeIdentifier().asInstanceOf[AbstractDelegationTokenIdentifier]
```
ulysses-you, line 123: we still have hive dependencies for tests, why do we need change this ?

NOTICE-binary
pan3793: should be retained, the following jars were shipped under `beeline-jars`
```
hive-beeline-3.1.3.jar
hive-common-3.1.3.jar
hive-jdbc-3.1.3.jar
hive-service-3.1.3.jar
hive-service-rpc-3.1.3.jar
```

kyuubi-server/pom.xml
pan3793: see https://github.com/apache/kyuubi-shaded/pull/35

so far, it's only for token, but it will be changed over time

kyuubi-server/src/test/scala/org/apache/kyuubi/credentials/HiveDelegationTokenProviderSuite.scala
pan3793: actually, we should use the shaded `DelegationTokenIdentifier` instead

pan3793: BTW, the test scope hive jars are going to be removed soon

ulysses-you:
lgtm if tests pass

pan3793:
Thanks, merged to master

PR #6120, opened 2024-03-01
Trino create client session should use engine seesion user

lsm1:
# :mag: Description
## Issue References 🔗

In engine share level such as SERVER, KYUUBI_SESSION_USER_KEY is a fixed value, we should use the engine session user as the trino client session user

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


cxzl25 (approved):

pan3793:
> In engine share level such as SERVER, KYUUBI_SESSION_USER_KEY is a fixed value

this seems not an expected behavior, will take a look later

pan3793:
The issue should be `val sessionConf: KyuubiConf = sessionManager.getConf`, it's actually an engine level conf, but we expect a session level conf.

PR #6131, opened 2024-03-05
Simplify Maven dependency management after dropping building support for Spark 3.1

pan3793:
# :mag: Description
## Issue References 🔗

SPARK-33212 (fixed in 3.2.0) moves from `hadoop-client` to shaded hadoop client, to simplify the dependency management, previously , we add some workaround to handle Spark 3.1 dependency issues. As we removed building support for Spark 3.1 now, we can remove those workaround to simplify `pom.xml`

## Describe Your Solution 🔧

As above.

## Test Plan 🧪

Pass GA.

---

# Checklist 📝

**Be nice. Be informative.**


pan3793:
Thanks, merged to master

PR #6140, opened 2024-03-07
Move getCodeSourceLocation to kyuubi-util

pan3793:
# :mag: Description
## Issue References 🔗

The `getCodeSourceLocation` lives on `kyuubi-common`, which is not reachable for modules like `kyuubi-hive-beeline`.

## Describe Your Solution 🔧

Move it to `kyuubi-util`.

## Test Plan 🧪

Pass GA.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/JavaUtils.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/JavaUtils.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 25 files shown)


kyuubi-util/src/main/java/org/apache/kyuubi/util/JavaUtils.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.apache.kyuubi.util;
+
+import java.io.File;
+import java.net.URISyntaxException;
+
+public class JavaUtils {
+
+  public static String getCodeSourceLocation(Class<?> clazz) {
```
ulysses-you, line 27: shall we drop Utils.getCodeSourceLocation ?

pan3793: yes, we should, just forgot to delete. fixed

ulysses-you (approved):

pan3793:
Thanks, merged to master

PR #6148, opened 2024-03-08
[KYUUBI #6147][TRINO] Use the overlay conf as session conf

### LINKED ISSUE #6147: [Bug][TRINO] Trino engine use the engine level conf as session level conf
### Code of Conduct

### Search before asking

### Describe the bug

In TrinoSessionImpl, it's wrong consider sessionManager.getConf as session level conf. Actually, it's engine level conf.

https://github.com/apache/kyuubi/blob/6fc7552ef05c83cfb8a2cd56d80c02c6e8a033c7/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala#L52

We should use the passed arg conf as session level conf.

https://github.com/apache/kyuubi/blob/6fc7552ef05c83cfb8a2cd56d80c02c6e8a033c7/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala#L43-L50

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

zwangsheng:
# :mag: Description
## Issue References 🔗

This pull request fixes #6147

## Describe Your Solution 🔧

Overwriting the engine conf with the passed conf consider as the full session conf.

#### Related Unit Tests
None

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala
+++ b/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala
@@ -51,7 +51,7 @@ class TrinoSessionImpl(
 
   val sessionConf: KyuubiConf = {
     val engineConf = sessionManager.getConf.clone
-    conf.foreach(kv => engineConf.set(kv._1, kv._2))
+    conf.foreach { case (k, v) => engineConf.set(k, v) }
     engineConf
   }
 
--- a/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala
+++ b/externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala
@@ -49,7 +49,11 @@ class TrinoSessionImpl(
     sessionManager: SessionManager)
   extends AbstractSession(protocol, user, password, ipAddress, conf, sessionManager) {
 
-  val sessionConf: KyuubiConf = sessionManager.getConf
+  val sessionConf: KyuubiConf = {
+    val engineConf = sessionManager.getConf
+    conf.foreach(kv => engineConf.set(kv._1, kv._2))
+    engineConf
+  }
 
   override val handle: SessionHandle =
     conf.get(KYUUBI_SESSION_HANDLE_KEY).map(SessionHandle.fromUUID).getOrElse(SessionHandle())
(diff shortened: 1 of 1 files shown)


externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala
```diff
@@ -49,7 +49,11 @@ class TrinoSessionImpl(
     sessionManager: SessionManager)
   extends AbstractSession(protocol, user, password, ipAddress, conf, sessionManager) {
 
-  val sessionConf: KyuubiConf = sessionManager.getConf
+  val sessionConf: KyuubiConf = {
+    val engineConf = sessionManager.getConf
+    conf.foreach(kv => engineConf.set(kv._1, kv._2))
```
cxzl25, line 54: Will this not pollute the engine conf?

zwangsheng: thanks for catch, using `sessionManager.getConf.clone` instead.

pan3793: @cxzl25 seems we lack a pre-assembled session conf in AbstractSession, we need some major refactor to simplify the current conf evaluation, I roughly remember you have raised a similar idea previously.

pan3793 (approved):
LGTM, I think we can adopt this ad-hoc fix and reconsider the https://github.com/apache/kyuubi/commit/d0f479831a5fca72dc10bf97f2627b05ac07946b 's idea later

externals/kyuubi-trino-engine/src/main/scala/org/apache/kyuubi/engine/trino/session/TrinoSessionImpl.scala
```diff
@@ -49,7 +49,11 @@ class TrinoSessionImpl(
     sessionManager: SessionManager)
   extends AbstractSession(protocol, user, password, ipAddress, conf, sessionManager) {
 
-  val sessionConf: KyuubiConf = sessionManager.getConf
+  val sessionConf: KyuubiConf = {
+    val engineConf = sessionManager.getConf.clone
+    conf.foreach(kv => engineConf.set(kv._1, kv._2))
```
pan3793, line 54:
```suggestion
    conf.foreach { case (k, v) => engineConf.set(k, v) }
```

pan3793:
Thanks, merged to master/1.8.2

PR #6150, opened 2024-03-08
Invoke setStopAtShutdown for jetty server

turboFei:
# :mag: Description
## Issue References 🔗

Invoke org.eclipse.jetty.server.Server::setStopAtShutdown
```
    /**
     * Set stop server at shutdown behaviour.
     *
     * @param stop If true, this server instance will be explicitly stopped when the
     * JVM is shutdown. Otherwise the JVM is stopped with the server running.
     * @see Runtime#addShutdownHook(Thread)
     * @see ShutdownThread
     */
    public void setStopAtShutdown(boolean stop)
```

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/ui/JettyServer.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/ui/JettyServer.scala
@@ -34,6 +34,7 @@ private[kyuubi] case class JettyServer(
       server.start()
       connector.start()
       server.addConnector(connector)
+      server.setStopAtShutdown(true)
     } catch {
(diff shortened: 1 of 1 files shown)


pan3793:
This adds a shutdown hook to close the Jetty server, actually, Kyuubi also adds shutdown hooks to close all frontend services and recursively close the Jetty server. So is it necessary?

turboFei:
> This adds a shutdown hook to close the Jetty server, actually, Kyuubi also adds shutdown hooks to close all frontend services and recursively close the Jetty server. So is it necessary?

Make senses, will close it.

PR #6153, opened 2024-03-11
Cut out Hive JDBC and Hive Service deps in beeline module

pan3793:
# :mag: Description
## Issue References 🔗

This is the next step of https://github.com/apache/kyuubi/issues/6146, cutting out most Hive deps(excepting `hive-common`) and recovering the skipped tests via minor code tunning.

## Describe Your Solution 🔧

- Drop `hive-jdbc`, `hive-service`, `hive-service-rpc` deps in the beeline module.
- Migrate from `commons-lang` to `commons-lang3` in the beeline module.
- Recover the skipped test `TestClientCommandHookFactory#connectHook`

## Test Plan 🧪

Pass GA, and manually test to ensure the following error has gone.

Before
```
root@hadoop-master1:/opt/kyuubi# bin/beeline --version
Warn: Not find kyuubi environment file /etc/kyuubi/conf/kyuubi-env.sh, using default ones...
java.lang.NoClassDefFoundError: org/apache/curator/RetryPolicy
```

After
```
root@hadoop-master1:/opt/kyuubi# bin/beeline --version
Connecting to jdbc:hive2://hadoop-master1.orb.local:10000/default;password=hive;user=hive
Connected to: Apache Hive (version 2.3.9)
Driver: Kyuubi Project Hive JDBC Client (version 1.9.0-SNAPSHOT)
Beeline version 1.9.0-SNAPSHOT by Apache Kyuubi
0: jdbc:hive2://hadoop-master1.orb.local:1000>
```

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-hive-beeline/src/test/java/org/apache/hive/beeline/ProxyAuthTest.java
+++ b/kyuubi-hive-beeline/src/test/java/org/apache/hive/beeline/ProxyAuthTest.java
@@ -26,10 +26,11 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import javax.security.auth.login.LoginException;
 import org.apache.commons.io.FileUtils;
+import org.apache.hadoop.hdfs.security.token.delegation.DelegationTokenIdentifier;
+import org.apache.hadoop.io.Text;
 import org.apache.hadoop.security.UserGroupInformation;
-import org.apache.hive.jdbc.HiveConnection;
-import org.apache.hive.service.auth.HiveAuthConstants;
-import org.apache.hive.service.cli.session.SessionUtils;
+import org.apache.hadoop.security.token.Token;
+import org.apache.kyuubi.jdbc.hive.KyuubiConnection;
 
 /**
  * Simple client application to test various direct and proxy connection to HiveServer2 Note that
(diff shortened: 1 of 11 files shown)


kyuubi-hive-beeline/src/test/java/org/apache/hive/beeline/ProxyAuthTest.java
```diff
@@ -40,7 +41,7 @@
  * <client-principal>
  */
 public class ProxyAuthTest {
-  private static final String driverName = "org.apache.hive.jdbc.HiveDriver";
+  private static final String driverName = "org.apache.kyuubi.jdbc.KyuubiHiveDriver";
```
ulysses-you, line 44: we still support `org.apache.hive.jdbc.HiveDriver` right ?

ulysses-you:
lgtm if tests pass

kyuubi-hive-beeline/src/test/java/org/apache/hive/beeline/ProxyAuthTest.java
pan3793: I think we have no such guarantee, the internally used JDBC driver is not user-faced stuff, we only need to ensure that user could smoothly migrate from Hive Beeline to Kyuubi Beeline without changing the JDBC URL.

pan3793:
Thanks, merged to master

PR #6162, opened 2024-03-11
Cut out hive-common deps in beeline module

pan3793:
# :mag: Description
## Issue References 🔗

This PR removes the `hive-common` dep from the beeline module, and it is the final step of cutting out all Hive deps from the beeline module.

## Describe Your Solution 🔧

- Remove support of Hive CLI mode (a.k.a. `beelineMode == false`) on beeline. 
- Copy ~7 classes (~600 lines) from `org.apache.hive:hive-common:3.1.3`
- Drop `org.apache.hive:hive-common:3.1.3` from beeline module

Note: Hive CLI mode removal is necessary, because it is useless for Kyuubi cases and seriously coupled with `HiveConf`

## Test Plan 🧪

Pass GA, and manually tested.

```
root@hadoop-master1:/opt/kyuubi# beeline
Connecting to jdbc:hive2://hadoop-master1.orb.local:10000/default;password=hive;user=hive
Connected to: Apache Hive (version 2.3.9)
Driver: Kyuubi Project Hive JDBC Client (version 1.9.0-SNAPSHOT)
Beeline version 1.9.0-SNAPSHOT by Apache Kyuubi
0: jdbc:hive2://hadoop-master1.orb.local:1000> select version();
+--------------------------------------------------+
|                       _c0                        |
+--------------------------------------------------+
| 2.3.9 r92dd0159f440ca7863be3232f3a683a510a62b9d  |
+--------------------------------------------------+
1 row selected (1.443 seconds)
0: jdbc:hive2://hadoop-master1.orb.local:1000>
```

---

# Checklist 📝

**Be nice. Be informative.**


kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/hs2connection/HiveSiteHS2ConnectionFileParser.java
```diff
@@ -18,42 +18,98 @@
...
 import org.slf4j.LoggerFactory;
 
 /*
  * Looks for a hive-site.xml from the classpath. If found this class parses the hive-site.xml
  * to return a set of connection properties which can be used to construct the connection url
  * for Beeline connection
+ *
+ * This class is modified to get rid of dependency on HiveConf
  */
 public class HiveSiteHS2ConnectionFileParser implements HS2ConnectionFileParser {
```
pan3793, line 38: it's easy to extend the `HS2ConnectionFileParser` to make the beeline support auto-getting connection URL from `kyuubi-defaults.conf`, we should make such extension in the future.

ulysses-you (approved):

pan3793:
@ulysses-you As we discussed offline, in case users may pull `hive-beeline` as a dependency, to reduce the effort of migration to `kyuubi-hive-beeline`, I moved all classes copied from `hive-common` to the package `org.apache.hive.beeline.common`.

pan3793:
Thanks, merged to master

PR #6166, opened 2024-03-12
[MINOR] Fix wrong copy path about SPARK_HOME in docker

yanghua:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

After running `build/dist`, the old structure is like `dist/spark-binary/spark-3.4.2-bin-hadoop3/xxx`.
In `Dockerfile`, after the command `ONBUILD COPY spark-binary ${SPARK_HOME}` takes effect. The path in the container is `/opt/spark/spark-3.4.2-bin-hadoop3`. So I suggest only copy the content of the origin spark binary, making the structure looks like `dist/spark-binary/xxx`. Finally, the path in the container (`/opt/spark/xxx`) should be correct.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/bin/docker-image-tool.sh
+++ b/bin/docker-image-tool.sh
@@ -110,7 +110,7 @@ function build {
         error "Cannot found dir SPARK_HOME $SPARK_HOME, you must configure SPARK_HOME correct."
       fi
     fi
-    cp -r "$SPARK_HOME/" "$KYUUBI_ROOT/spark-binary/"
+    cp -r "$SPARK_HOME/." "$KYUUBI_ROOT/spark-binary/"
   fi
 
   # Verify that the Docker image content directory is present


zwangsheng:
Hi @yanghua, i test in my local env with following steps:
1. download Kyuubi tgz from website and tar
2. run `./bin/docker-image-tool.sh -s ~/work/spark -t test build`
3. run `docker run kyuubi:test ls -l /opt/spark`
And it's expected.

zwangsheng (approved):

yanghua:
> Hi @yanghua, I test in my local macOs with following steps and old script:
> 
> 1. download Kyuubi tgz from website and tar
> 2. run `./bin/docker-image-tool.sh -s ~/work/spark -t test build`
> 3. run `docker run kyuubi:test ls -l /opt/spark`
> 
> And it's expected. I'm not sure those issue caused by os or other problem.
> 
> After all, I also test this PR in local, it fit goal. I'm fine to accept this. cc @pan3793

My steps in 1.8.1 branch:

```
./build/dist --name 1.8.1 --tgz

cd dist

sudo ./bin/docker-image-tool.sh -r xxx -t v1.8.1 build

sudo ./bin/docker-image-tool.sh -r xxx -t v1.8.1 push
```

@zwangsheng Can you try it without `-s` option?

yanghua:
In short, if we want to copy all the contact of `src` dir to `dist`.

`cp -r src/ dist/`

 `cp` command has this behavior: if `dist` exists, the result is `dist/src`

So, if we want to get the right behavior, it should be `cp -r src/. dist/` or `cp -r src/* dist/`.

zwangsheng:
LGTM

zwangsheng:
Thanks, merge to master.

bin/docker-image-tool.sh
```diff
@@ -110,7 +110,7 @@ function build {
         error "Cannot found dir SPARK_HOME $SPARK_HOME, you must configure SPARK_HOME correct."
       fi
     fi
-    cp -r "$SPARK_HOME/" "$KYUUBI_ROOT/spark-binary/"
+    cp -r "$SPARK_HOME/." "$KYUUBI_ROOT/spark-binary/"
```
pan3793, line 113: does `"$SPARK_HOME/*"` work? it's a more common case than that dot

pan3793:
if this is a bug fix, we should backport to branch-1.8

bin/docker-image-tool.sh
yanghua: > does `"$SPARK_HOME/*"` work? it's a more common case than that dot

oh... `*` does not work. It throws:

```
cp: cannot stat '/data00/home/yanghua/kyuubi/dist/externals/spark-3.4.2-bin-hadoop3/*': No such file or directory
```

yanghua:
> if this is a bug fix, we should backport to branch-1.8

Do a cherry-pick?

pan3793:
Cherry-picked to 1.8

PR #6185, opened 2024-03-15
Audit kyuubi operation state change log

turboFei:
# :mag: Description
## Issue References 🔗

We meet some issue, I want to check the operation sessionHandle and the operation timeline.
I found that, it is difficult for me to check it from kyuubi log.
So, In this pr, I log the operation change log into separate file.
## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/test/resources/log4j2-test.xml
+++ b/kyuubi-server/src/test/resources/log4j2-test.xml
@@ -21,6 +21,7 @@
 <Configuration status="WARN">
     <Properties>
         <Property name="restAuditLogPath">target/rest-audit.log</Property>
+        <Property name="opAuditLogPath">target/operation-audit.log</Property>
     </Properties>
     <Appenders>
         <Console name="stdout" target="SYSTEM_OUT">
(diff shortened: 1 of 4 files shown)


turboFei:
cc @pan3793

kyuubi-server/src/test/resources/log4j2-test.xml
```diff
@@ -21,6 +21,7 @@
 <Configuration status="WARN">
```
pan3793, line 21: we have bunches of `log4j2-test.xml`, can we sync them?

cxzl25 (approved):

kyuubi-server/src/test/resources/log4j2-test.xml
turboFei: I think it is not needed, we just need to enable it in kyuubi-server side.

pan3793:
Thanks, merged to master

PR #6196, opened 2024-03-19
Add log4j2 template for kyuubi repl commands

wForget:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

Add log4j2 template for kyuubi repl command.

Since I want to add rolling logs for kyuubi server, we need to add independent log4j properties for kyuubi repl command.

- [X] New feature (non-breaking change which adds functionality)

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


conf/log4j2-repl.xml.template
```diff
@@ -0,0 +1,36 @@
...
+
+<!-- Provide log4j2.xml.template to fix `ERROR Filters contains invalid attributes "onMatch", "onMismatch"`, see KYUUBI-2247 -->
+<!-- Extra logging related to initialization of Log4j.
+ Set to debug or trace if log4j initialization is failing. -->
+<Configuration status="INFO">
+    <Appenders>
+        <Console name="stdout" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} %p %tn %c: %m%n%ex"/>
+            <Filters>
+                <RegexFilter regex=".*Thrift error occurred during processing of message.*" onMatch="DENY" onMismatch="NEUTRAL"/>
```
pan3793, line 27: IIRC, this is to address a Thrift exception, seems not required now

conf/log4j2.xml.template
```diff
@@ -55,7 +55,7 @@
         <Root level="INFO">
             <AppenderRef ref="stdout"/>
         </Root>
-        <Logger name="org.apache.kyuubi.ctl.ServiceControlCli" level="error" additivity="false">
```
wForget, line 58: This class has been removed

conf/log4j2-repl.xml.template
wForget: > IIRC, this is to address a Thrift issue, which seems is not required as we already upgraded to 0.16

Thanks, I will remove it. Do we need to remove it in `log4j2.xml.template` too?

conf/log4j2.xml.template
pan3793: since you created a separate `log4j2-repl.xml`, do we need such conf in `log4j2.xml` then?

wForget: > since you created a separate `log4j2-repl.xml`, do we need such conf in `log4j2.xml` then?

I think it can be kept because user may not have created log4j2-repl.xml, in which case log4j2.xml can also be used.

pan3793: SGTM if it works, please manually check it

wForget: > SGTM if it works, please manually check it

I have tested manually and it works fine.

pan3793 (approved):

PR #6199, opened 2024-03-19
Support to run HiveSQLEngine on kerberized YARN

zhouyifan279:
# :mag: Description
## Issue References 🔗

This pull request implement a feature -  Run HiveSQLEngine on kerberized YARN

## Describe Your Solution 🔧
Introduced two configs:
- kyuubi.engine.principal
- kyuubi.engine.keytab

When do submit to a kerberized YARN, submitter uploads `kyuubi.engine.keytab` to application's staging dir.
YARN NodeManager downloads keytab to AM's working directory. AM logins to Kerberos using the principal and keytab

**Note**
I've tried to run HiveSQLEngine with only DelegationTokens but failed. 

Take SQL `SELECT * FROM a` as an example:
Hive handles this simple TableScan SQL by reading directly from table's hdfs file.
When Hive invokes `FileInputFormat.getSplits` during reading, `java.io.IOException: Delegation Token can be issued only with kerberos or web authentication` will be thrown. 
The simplified stacktrace from IDEA is as below:
```
getDelegationToken:734, DFSClient (org.apache.hadoop.hdfs)
getDelegationToken:2072, DistributedFileSystem (org.apache.hadoop.hdfs)
collectDelegationTokens:108, DelegationTokenIssuer (org.apache.hadoop.security.token)
addDelegationTokens:83, DelegationTokenIssuer (org.apache.hadoop.security.token)
obtainTokensForNamenodesInternal:143, TokenCache (org.apache.hadoop.mapreduce.security)
obtainTokensForNamenodesInternal:102, TokenCache (org.apache.hadoop.mapreduce.security)
obtainTokensForNamenodes:81, TokenCache (org.apache.hadoop.mapreduce.security)
listStatus:221, FileInputFormat (org.apache.hadoop.mapred)
getSplits:332, FileInputFormat (org.apache.hadoop.mapred)
getNextSplits:372, FetchOperator (org.apache.hadoop.hive.ql.exec)
getRecordReader:304, FetchOperator (org.apache.hadoop.hive.ql.exec)
getNextRow:459, FetchOperator (org.apache.hadoop.hive.ql.exec)
pushRow:428, FetchOperator (org.apache.hadoop.hive.ql.exec)
fetch:147, FetchTask (org.apache.hadoop.hive.ql.exec)
getResults:2208, Driver (org.apache.hadoop.hive.ql)
getNextRowSet:494, SQLOperation (org.apache.hive.service.cli.operation)
getNextRowSetInternal:105, HiveOperation (org.apache.kyuubi.engine.hive.operation)
```

Theoretically, it can be solved by add AM DelegationTokens into 
 `org.apache.hadoop.hive.ql.exec.FetchOperator.job.credentials`.
But actually, it is impossible without modifying Hive's source code.

HiveSQLEngine can not run on a kerberized YARN

HiveSQLEngine can run on a kerberized YARN

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


zhouyifan279:
cc @pan3793 @Yikf

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
```diff
@@ -91,9 +98,36 @@ abstract class EngineYarnModeSubmitter extends Logging {
    */
   def engineExtraJars(): Seq[File] = Seq.empty
 
-  protected def submitApplication(): Unit = {
+  def run(): Unit = {
     yarnConf = KyuubiHadoopUtils.newYarnConfiguration(kyuubiConf)
     hadoopConf = KyuubiHadoopUtils.newHadoopConf(kyuubiConf)
+    appUser = kyuubiConf.getOption(KYUUBI_SESSION_USER_KEY).orNull
```
pan3793, line 104: is this correct for GROUP and SERVER share level?

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
```diff
@@ -134,6 +168,29 @@ abstract class EngineYarnModeSubmitter extends Logging {
...
+    if (UserGroupInformation.isSecurityEnabled) {
+      val credentials = obtainHadoopFsDelegationToken()
+      val serializedCreds = KyuubiHadoopUtils.serializeCredentials(credentials)
+      amContainer.setTokens(ByteBuffer.wrap(serializedCreds))
+    }
+  }
+
+  private def obtainHadoopFsDelegationToken(): Credentials = {
+    val tokenRenewer = Master.getMasterPrincipal(hadoopConf)
+    info("Delegation token renewer is: " + tokenRenewer)
```
pan3793, line 181:
```suggestion
    info("Delegation token renewer is: $tokenRenewer")
```

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
zhouyifan279: `KYUUBI_SESSION_USER_KEY`  is set to `EngineRef#appUser` when submit Engines. So it's correct for all share levels.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/ApplicationMaster.scala
```diff
@@ -71,7 +73,37 @@ object ApplicationMaster extends Logging {
...
+            info("Attempting to login to Kerberos " +
+              s"using principal: ${principalName} and keytab: ${keytabFilename}")
+            UserGroupInformation.loginUserFromKeytab(principalName, keytabFilename)
+          }
+
+          val newUGI = UserGroupInformation.getCurrentUser()
+
+          // Transfer the original user's tokens to the new user, since it may contain needed tokens
+          // (such as those user to connect to YARN).
+          // TODO originalCreds may have been expired if Application retries.
```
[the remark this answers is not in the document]
zhouyifan279: I just found we also had disabled retry in `EngineYarnModeSubmitter`.  This comment can be removed.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
pan3793: It's weird. I remember we use this key to distinguish the session user and app user in some cases ...

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
zhouyifan279: Done

pan3793:
> Theoretically, it can be solved by add AM DelegationTokens into org.apache.hadoop.mapred.JobConf#credentials.
But actually, it is impossible without modifying Hive's source code.

If possible, would you mind leaving some notes about the invocation chain? It may be asked by future explores or contributors

cxzl25:
> When Hive invokes FileInputFormat#getSplits, java.io.IOException: Delegation Token can be issued only with kerberos or web authentication will be thrown.
Theoretically, it can be solved by add AM DelegationTokens into org.apache.hadoop.mapred.JobConf#credentials.
But actually, it is impossible without modifying Hive's source code.

Agreed, many years ago we removed the `TokenCache.obtainTokensForNamenodes` logic from Hadoop's `FileInputFormat`.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
```diff
@@ -187,6 +245,19 @@ abstract class EngineYarnModeSubmitter extends Logging {
 
     distributeJars(localResources, env)
     distributeConf(localResources, env)
+
+    // If we passed in a keytab, make sure we copy the keytab to the staging directory on
+    // HDFS, and setup the relevant environment vars, so the AM can login again.
+    amKeytabFileName.foreach { kt =>
+      info("To enable the AM to login from keytab, credentials are being copied over to the AM" +
```
cxzl25, line 252: This log seems a bit redundant, because in the Kerberos environment, we explicitly require users to configure this configuration item.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/ApplicationMaster.scala
```diff
@@ -71,7 +73,36 @@ object ApplicationMaster extends Logging {
...
+
+          val newUGI = UserGroupInformation.getCurrentUser()
+
+          // Transfer the original user's tokens to the new user, since it may contain needed tokens
+          // (such as those user to connect to YARN).
+          newUGI.addCredentials(originalCreds)
+          newUGI
+        case _ =>
+          val appUser = kyuubiConf.getOption(KyuubiReservedKeys.KYUUBI_SESSION_USER_KEY)
+          require(appUser.isDefined)
```
Yikf, line 97: It will be nice if add error message for the require.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
```diff
@@ -91,9 +98,36 @@ abstract class EngineYarnModeSubmitter extends Logging {
...
+      // with any user file.
+      Some(new File(keytab).getName() + "-" + UUID.randomUUID().toString)
+    } else {
+      None
+    }
+
+    val ugi = if (UserGroupInformation.getCurrentUser.getShortUserName == appUser) {
+      UserGroupInformation.getCurrentUser
+    } else {
+      UserGroupInformation.createProxyUser(appUser, UserGroupInformation.getCurrentUser)
```
Yikf, line 123: This also supports proxy users, right?

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
```diff
@@ -134,6 +168,29 @@ abstract class EngineYarnModeSubmitter extends Logging {
     }
   }
 
+  private def setupSecurityToken(amContainer: ContainerLaunchContext): Unit = {
+    if (UserGroupInformation.isSecurityEnabled) {
+      val credentials = obtainHadoopFsDelegationToken()
+      val serializedCreds = KyuubiHadoopUtils.serializeCredentials(credentials)
+      amContainer.setTokens(ByteBuffer.wrap(serializedCreds))
```
Yikf, line 175: Why does the ApplicationMaster need to login even if the token is set on this block?

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
```diff
@@ -17,10 +17,12 @@
 package org.apache.kyuubi.engine.deploy.yarn
```
Yikf, line 17: The  submitter process determine whether to login as well in kerberized environment?

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
zhouyifan279: Without`UserGroupInformation.loginUserFromKeytab`, the login user can only authenticate by tokens.
But HiveSQLEngine requires the login user to authenticate by kerberos tgt to obtain new tokens.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
zhouyifan279: Keytabs are security-sensitive files. I think it's better to tell user why we upload keytabs in place.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
zhouyifan279: Yes

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
zhouyifan279: I'm not sure what "determine whether to login as well" means. Can you elaborate it?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/hive/HiveProcessBuilder.scala
```diff
@@ -17,16 +17,17 @@
 
 package org.apache.kyuubi.engine.hive
 
-import java.io.File
+import java.io.{File, IOException}
```
pan3793, line 20:
```suggestion
import java.io.File
```

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
pan3793: SGTM

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
pan3793: This should only happen when using TGT cache with YARN mode, though it does not work ... it's good to keep those code path in case some one modifies the Hive source code to make it workable, but we'd better add a warning message to explicitly tell users this is not likely to work.

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/ApplicationMaster.scala
zhouyifan279: Done

kyuubi-common/src/main/scala/org/apache/kyuubi/engine/deploy/yarn/EngineYarnModeSubmitter.scala
zhouyifan279: More precisely speaking, proxy user mechanism works fine during the submitting progress. AM can be started successfully.
But HiveSQLEngine currently only works with principal & keytab provided.
I think we should check this prerequisite in HiveYarnModeProcessBuilder.

pan3793: yes, exactly

PR #6201, opened 2024-03-20
Distinguish the collect mode of different engines

lsm1:
# :mag: Description
## Describe Your Solution 🔧

Currently, Spark, JDBC, and TRINO engines support incremental collection, but they share the same configuration option `kyuubi.operation.incremental.collect`. Sometimes, it's necessary to enable incremental collection only for specific engines, which requires distinguishing between them.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2652,6 +2651,14 @@ object KyuubiConf {
       .stringConf
       .createWithDefault("yyyy-MM-dd HH:mm:ss.SSS")
 
+  val ENGINE_SPARK_OPERATION_INCREMENTAL_COLLECT: ConfigEntry[Boolean] =
+    buildConf("kyuubi.engine.spark.operation.incremental.collect")
+      .doc("When true, the result will be sequentially calculated and returned to" +
+        s" the Spark driver. Note that, ${OPERATION_RESULT_MAX_ROWS.key} will be ignored" +
+        " on incremental collect mode. It fallback to `kyuubi.operation.incremental.collect`")
+      .version("1.9.0")
```
pan3793, line 2659: 1.9.0 is out, use 1.10.0

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #6207, opened 2024-03-25
Support to retrieve Spark UserDefinedType result

turboFei:
# :mag: Description
## Issue References 🔗

This pull request fixes #

To fix below issue:
```
24/03/25 00:47:10 ERROR SparkTBinaryFrontendService: Error getting result set metadata: 
java.lang.IllegalArgumentException: Unrecognized type name: struct<type:tinyint,size:int,indices:array<int>,values:array<double>>
```
image

## Describe Your Solution 🔧

#### Related Unit Tests
After this pr:
image

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
@@ -21,6 +21,7 @@ import java.util.Collections
 
 import scala.collection.JavaConverters._
 
+import org.apache.spark.sql.kyuubi.SparkDataTypeHelper
 import org.apache.spark.sql.types._
 
 import org.apache.kyuubi.shaded.hive.service.rpc.thrift._
@@ -65,7 +66,7 @@ object SchemaHelper {
     case _: ArrayType => TTypeId.ARRAY_TYPE
     case _: MapType => TTypeId.MAP_TYPE
     case _: StructType => TTypeId.STRUCT_TYPE
-    // TODO: it is private now, case udt: UserDefinedType => TTypeId.USER_DEFINED_TYPE
+    case _ if SparkDataTypeHelper.isUserDefinedType(typ) => TTypeId.USER_DEFINED_TYPE
     case other =>
       throw new IllegalArgumentException(s"Unrecognized type name: ${other.catalogString}")
   }
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
@@ -66,6 +66,7 @@ object SchemaHelper {
     case _: ArrayType => TTypeId.ARRAY_TYPE
     case _: MapType => TTypeId.MAP_TYPE
     case _: StructType => TTypeId.STRUCT_TYPE
+    // SPARK-7768(fixed in 3.2.0) promoted UserDefinedType to DeveloperApi
     case _ if SparkDataTypeHelper.isUserDefinedType(typ) => TTypeId.USER_DEFINED_TYPE
     case other =>
       throw new IllegalArgumentException(s"Unrecognized type name: ${other.catalogString}")
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
@@ -241,6 +241,8 @@ public class JdbcColumn {
         return "struct";
       case NULL_TYPE:
         return "void";
+      case UNION_TYPE:
+        return "uniontype";
       case USER_DEFINED_TYPE:
         return "user_defined";
       default:
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
@@ -241,8 +241,6 @@ public class JdbcColumn {
         return "struct";
       case NULL_TYPE:
         return "void";
-      case UNION_TYPE:
-        return "uniontype";
       case USER_DEFINED_TYPE:
         return "user_defined";
       default:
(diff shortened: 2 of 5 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
```diff
@@ -178,6 +178,7 @@ public static int convertTTypeIdToSqlType(TTypeId type) throws SQLException {
       case INTERVAL_YEAR_MONTH_TYPE:
       case INTERVAL_DAY_TIME_TYPE:
       case UNION_TYPE:
+      case USER_DEFINED_TYPE:
```
yaooqinn, line 181: please implement getColumnTypeName

yaooqinn: And other APIs

turboFei: thanks, addressed

turboFei: Seems other apis all leverage `convertTTypeIdToSqlType`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -65,7 +66,7 @@ object SchemaHelper {
     case _: ArrayType => TTypeId.ARRAY_TYPE
     case _: MapType => TTypeId.MAP_TYPE
     case _: StructType => TTypeId.STRUCT_TYPE
-    // TODO: it is private now, case udt: UserDefinedType => TTypeId.USER_DEFINED_TYPE
+    case _ if SparkDataTypeHelper.isUserDefinedType(typ) => TTypeId.USER_DEFINED_TYPE
```
pan3793, line 69: We are likely to drop support for 3.1 soon, then we can simplify the code
```suggestion
    // SPARK-7768(fixed in 3.2.0) promoted UserDefinedType to DeveloperApi
    case _ if SparkDataTypeHelper.isUserDefinedType(typ) => TTypeId.USER_DEFINED_TYPE
```

turboFei: done

pan3793 (approved):

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
```diff
@@ -240,6 +241,8 @@ static String getColumnTypeName(TTypeId type) throws SQLException {
         return "struct";
       case NULL_TYPE:
         return "void";
+      case USER_DEFINED_TYPE:
+        return "user_defined";
```
yaooqinn, line 245: Just a question, comply with hive?

turboFei: https://github.com/apache/hive/blob/6dc073b673ec82900424654e5e148625020acd6d/serde/src/java/org/apache/hadoop/hive/serde2/thrift/Type.java#L107C1-L110C20

Yes，it is compatible, just lower case.

turboFei:
```
  USER_DEFINED_TYPE("USER_DEFINED",
      java.sql.Types.OTHER,
      TTypeId.USER_DEFINED_TYPE,
      true, false);

  private final String name;
  private final TTypeId tType;
  private final int javaSQLType;
  private final boolean isQualified;
  private final boolean isComplex;
  private final boolean isCollection;

  Type(String name, int javaSQLType, TTypeId tType, boolean isQualified, boolean isComplex,
      boolean isCollection) {
    this.name = name;
    this.javaSQLType = javaSQLType;
    this.tType = tType;
    this.isQualified = isQualified;
    this.isComplex = isComplex;
    this.isCollection = isCollection;
  }
```

turboFei: Also cc @jiaoqingbo related pr: https://github.com/apache/kyuubi/pull/2878

yaooqinn (approved):

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
```diff
@@ -240,6 +241,10 @@ static String getColumnTypeName(TTypeId type) throws SQLException {
         return "struct";
       case NULL_TYPE:
         return "void";
+      case UNION_TYPE:
+        return "uniontype";
```
turboFei, line 245: UNION_TYPE was missed here before.
```
  UNION_TYPE("UNIONTYPE",
      java.sql.Types.OTHER,
      TTypeId.UNION_TYPE,
      true, false),
```

pan3793: the original copy is based on 2.3 / 3.1, if things happen after 3.1, we'd better leave a Hive JIRA ticket in the comments

turboFei: It has been 4 years since last commit for Type.java.
image

turboFei: https://github.com/apache/hive/blob/master/serde/src/java/org/apache/hadoop/hive/serde2/thrift/Type.java

turboFei: Missed here: https://github.com/apache/kyuubi/pull/2878/files#r1538554887

turboFei: should I raise another pr to address it? @pan3793

pan3793: let's fix it first, independently

turboFei:
thanks, merged to master and 1.9.

PR #6258, opened 2024-04-04
[KYUUBI #6079] Web UI support Basic authN

### LINKED ISSUE #6079: [TASK][MEDIUM]  Web UI support Basic authN
### Code of Conduct

### Search before asking

https://github.com/apache/kyuubi/pull/6042#issuecomment-1959437519

### How should we improve?

Support Basic authN for Web UI - that means, when SASL/Plain(LDAP/JDBC/CUSTOM) auth method is configured, pop something like below to allow user login via username and password 

image

beryllw:
# :mag: Description
## Issue References 🔗

This pull request fixes #6079 

## Describe Your Solution 🔧

image

zpHy0SGauD

image

## Test Plan 🧪

---

# Checklist 📝

**Be nice. Be informative.**


beryllw:
> the frontend code needs to be modified so that all http request will attach a basic auth header

frontend code means web-ui code? 
If user login success, other http request will auto attach a basic auth header.

pan3793:
> If user login success, other http request will auto attach a basic auth header.

interesting, let me verify it locally

pan3793:
found 2 issues:
- if user input the wrong username/password, no way to correct
- only SQL editor works

beryllw:
> * if user input the wrong username/password, no way to correct

Fixing this will change the current authentication logic, i will fix it soon.

> * only SQL editor works

Other api return http status code 405? Because the current user is not an administrator.

```
org.apache.kyuubi.server.api.v1.AdminResource#sessions
    if (!fe.isAdministrator(userName)) {
      throw new NotAllowedException(
        s"$userName is not allowed to list all live sessions")
    }
```

pan3793:
> Other api return http status code 405? Because the current user is not an administrator.

Oh, I see ...

beryllw:
@pan3793 cc

pan3793:
seems a explicitly logout button is required, otherwise user can not switch login user

beryllw:
> seems a explicitly logout button is required, otherwise user can not switch login user


https://stackoverflow.com/questions/233507/how-to-log-out-user-from-web-site-using-basic-authentication

> What you have to do is have the user click a logout link, and send a ‘401 Unauthorized’ in response, using the same realm and at the same URL folder level as the normal 401 you send requesting a login.

Simple implementation is that when the user clicks the logout button, the server returns 401.

pan3793:
> Simple implementation is that when the user clicks the logout button, the server returns 401.

We can use a specific endpoint for that, so other endpoints behavior won't be affected

beryllw:
> Simple implementation is that when the user clicks the logout button, the server returns 401.

Login users cannot be displayed in this way.Implemented on the frontend code instead.
phoSgHuZKz
mssFmLoEiD

pan3793 (approved):

pan3793:
Thanks, merged to master/1.9.1

PR #6260, opened 2024-04-06
Clean up and improve comments for spark extensions

pan3793:
# :mag: Description

This pull request 

- improves comments for SPARK-33832
- removes unused `spark.sql.analyzer.classification.enabled` (I didn't update the migration rules because this configuration seems never to work properly)

## Test Plan 🧪

Review

---

# Checklist 📝

**Be nice. Be informative.**


pan3793:
cc @ulysses-you, do we still require the Kyuubi customized `InsertShuffleNodeBeforeJoin` after SPARK-33832, seems SPARK-33832 is already smart enough

wForget:
> cc @ulysses-you, do we still require the Kyuubi customized `InsertShuffleNodeBeforeJoin` after SPARK-33832, seems SPARK-33832 is already smart enough

I think we may still need `InsertShuffleNodeBeforeJoin`, which can also be used to optimize multiple table join skew situations.

pan3793:
@wForget I see, they are handling different things, SPARK-33832 for skew optimization after Join, but this one is for before Join

extensions/spark/kyuubi-extension-spark-3-4/src/main/scala/org/apache/kyuubi/sql/InsertShuffleNodeBeforeJoin.scala
```diff
@@ -49,7 +49,7 @@ object InsertShuffleNodeBeforeJoin extends Rule[SparkPlan] {
     }
   }
 
-  // Since spark 3.3, insertShuffleBeforeJoin shouldn't be applied if join is skewed.
+  // SPARK-33832 (Spark 3.3) already handles skew join cases
```
wForget, line 52: As https://github.com/apache/kyuubi/pull/6260#issuecomment-2041317569 commented, should we revert this change?

pan3793: yes, the comment is incorrect then. but seems SPARK-46170 recovers the skew join cases

wForget (approved):
Thanks, LGTM.

pan3793:
Merged to master/1.9

PR #6284, opened 2024-04-10
[KYUUBI #6278] Update DB state when the query fallback to resource manager and the batch app is terminal

### LINKED ISSUE #6278: [TASK][EASY] Update DB state when the query fallback to yarn and the batch app is terminal
### Code of Conduct

### Search before asking

In scenarios with high concurrency for submitting batch sessions and querying status concurrently, especially when there are a large number of short-period batch session submissions, if some Kyuubi servers crash, it will impose significant pressure on the remaining Kyuubi servers. This is because at this time, they are unable to query the information of some batch sessions in their memory. Therefore, we should try to stack the pressure of this scenario onto YARN as much as possible. 

### How should we improve?

Therefore, I suggest that if a job reaches its final state, we should update the database as soon as possible to avoid further degradation of performance as much as possible.

yanghua:
# :mag: Description
## Issue References 🔗
Close #6278 

This pull request fixes #

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -341,9 +341,14 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
                 Some(metadata.engineOpenTime).filter(_ > 0).orElse(Some(System.currentTimeMillis)))
               // if the batch app is terminated, update the metadata in db.
               if (BatchJobSubmission.applicationTerminated(batchAppStatus)) {
+                val appInfo = batchAppStatus.get
                 sessionManager.updateMetadata(Metadata(
                   identifier = batchId,
-                  engineState = batchAppStatus.get.state.toString))
+                  engineId = appInfo.id,
+                  engineName = appInfo.name,
+                  engineUrl = appInfo.url.orNull,
+                  engineState = appInfo.state.toString,
+                  engineError = appInfo.error))
               }
               buildBatch(metadata, batchAppStatus)
           }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -31,6 +31,7 @@ import scala.util.control.NonFatal
 import io.swagger.v3.oas.annotations.media.{Content, Schema}
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.tags.Tag
+import org.apache.commons.lang3.StringUtils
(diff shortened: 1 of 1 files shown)


yanghua:
WIP.  Regardless of whether to change the code, local tests always fail. Use the community's CI to run.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -317,6 +317,7 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       sessionManager.getBatchMetadata(batchId).map { metadata =>
         if (batchV2Enabled(metadata.requestConf) ||
           OperationState.isTerminal(OperationState.withName(metadata.state)) ||
+          ApplicationState.isTerminated(ApplicationState.withName(metadata.engineState)) ||
```
turboFei, line 320: engineState might be null

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -315,8 +315,14 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       buildBatch(batchSession)
     }.getOrElse {
       sessionManager.getBatchMetadata(batchId).map { metadata =>
+        val isOperationTerminated = (metadata.state.nonEmpty
+          && OperationState.isTerminal(OperationState.withName(metadata.state)))
+        val isApplicationTerminated = (metadata.engineState.nonEmpty
```
turboFei, line 320:
```
StringUtils.isNotBlank(metadata.engineState)
```

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -332,6 +339,12 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
                 Some(userName),
                 // prevent that the batch be marked as terminated if application state is NOT_FOUND
                 Some(metadata.engineOpenTime).filter(_ > 0).orElse(Some(System.currentTimeMillis)))
+              // if the batch app is terminated, update the metadata in db.
+              if (BatchJobSubmission.applicationTerminated(batchAppStatus)) {
+                sessionManager.updateMetadata(Metadata(
+                  identifier = batchId,
+                  engineState = batchAppStatus.get.state.toString))
```
turboFei, line 346: should we update all the batchAppStatus into metadata?

Including appId, name, state, error.

turboFei (approved):
Left minor comments, cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
yanghua: +1

pan3793 (approved):

PR #6305, opened 2024-04-13
[KYUUBI #6297] Package Spark SQL engine both Scala 2.12 and 2.13

### LINKED ISSUE #6297: [TASK][EASY] Package Spark SQL engine both Scala 2.12 and 2.13
### What's the level of this task?

EASY

### Code of Conduct

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/343) and found no similar tasks.

### Mentor

- [ ] I have sufficient expertise on this task, and I volunteer to be a mentor of this task to guide contributors through the task.

### Skill requirements

- Shell scripts, Maven packaging
- knowledge of how Kyuubi launches Spark applications

### Background and Goals

The Kyuubi Spark SQL engine is a version-agnostic jar, which means the `kyuubi-spark-sql-engine_2.12-1.9.0.jar` built with Spark 3.5 should also work on Spark 3.2 and all supported versions.
Since 1.8, the whole Kyuubi project has supported building with Scala 2.13, so users could create a custom distribution with Scala 2.13. Another idea is, that one Kyuubi tgz ships both `kyuubi-spark-sql-engine_2.12-1.9.0.jar` and `kyuubi-spark-sql-engine_2.13-1.9.0.jar`, then it can support Scala 2.12 and 2.13 in same time.

### Implementation steps

Kyuubi Server already supports extracting Spark's scala version from `$SPARK_HOME/jars/spark-core_*.jar` and choosing the right `kyuubi-spark-sql-engine_*.jar`, so we need to modify `dev/dist` to make the final binary tarball always(no matter the Kyuubi server compiled with Scala 2.12 or 2.13) ships
```
externals/engines/spark/kyuubi-spark-sql-engine_2.12-1.10.0-SNAPSHOT.jar
externals/engines/spark/kyuubi-spark-sql-engine_2.13-1.10.0-SNAPSHOT.jar
```

### Additional context

Introduction of [2024H1 Kyuubi Code Contribution Program](https://github.com/apache/kyuubi/issues/6232)

PorterZhang2021:
# :mag: Description
## Issue References 🔗

This pull request fixes #6297 

## Describe Your Solution 🔧

Here are my main modifications:
1. In the `build/dist` file, I added an additional packaging command that detects the version of `SCALA_VERSION`. If it is version 2.12, then it will additionally package version 2.13, otherwise it will package version 2.12
2. In the section of `# Copy spark engines`, the Jar packages of versions 2.12 and 2.13 will be moved to `$DISTDIR/externals/engines/spark/`

This is my first time submitting a PR, so there may be many unfamiliar areas, and I don't know if my changes meet the needs of the community. More guidance may be needed. When testing locally, it was able to be packaged successfully

## Test Plan 🧪

image

image

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/build/dist
+++ b/build/dist
@@ -250,16 +250,16 @@ echo -e "\$ ${BUILD_COMMAND[@]}\n"
 
 # shellcheck disable=SC2050
 if [ "$SCALA_VERSION" = "2.12" ]; then
-  EXTRA_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@  -Pscala-2.13 -pl :kyuubi-spark-sql-engine_2.13 )
+  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@ -Pscala-2.13 -pl :kyuubi-spark-sql-engine_2.13 -am)
 else
-  EXTRA_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@  -Pscala-2.12 -pl :kyuubi-spark-sql-engine_2.12 )
+  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT -pl :kyuubi-spark-sql-engine_2.12 -am)
 fi
 
 # shellcheck disable=SC2145
-echo -e "\$ ${EXTRA_BUILD_COMMAND[@]}\n"
+echo -e "\$ ${EXTRA_SPARK_ENGINE_BUILD_COMMAND[@]}\n"
 
 "${BUILD_COMMAND[@]}"
-"${EXTRA_BUILD_COMMAND[@]}"
+"${EXTRA_SPARK_ENGINE_BUILD_COMMAND}"
 
 # Make directories
 rm -rf "$DISTDIR"
@@ -307,8 +307,8 @@ done
 cp "$KYUUBI_HOME/externals/kyuubi-flink-sql-engine/target/kyuubi-flink-sql-engine_${SCALA_VERSION}-${VERSION}.jar" "$DISTDIR/externals/engines/flink/"
 
 # Copy spark engines
-for SUPPORT_VERSION in 2.12 2.13; do
-  cp "$KYUUBI_HOME/externals/kyuubi-spark-sql-engine/target/kyuubi-spark-sql-engine_${SUPPORT_VERSION}-${VERSION}.jar" "$DISTDIR/externals/engines/spark/"
+for scala_version in 2.12 2.13; do
+  cp "$KYUUBI_HOME/externals/kyuubi-spark-sql-engine/target/kyuubi-spark-sql-engine_${scala_version}-${VERSION}.jar" "$DISTDIR/externals/engines/spark/"
 done
 
 # Copy trino engines
--- a/build/dist
+++ b/build/dist
@@ -252,14 +252,14 @@ echo -e "\$ ${BUILD_COMMAND[@]}\n"
 if [ "$SCALA_VERSION" = "2.12" ]; then
   EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@ -Pscala-2.13 -pl :kyuubi-spark-sql-engine_2.13 -am)
 else
-  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT -pl :kyuubi-spark-sql-engine_2.12 -am)
+  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@ -Pscala-2.12 -pl :kyuubi-spark-sql-engine_2.12 -am)
 fi
 
 # shellcheck disable=SC2145
 echo -e "\$ ${EXTRA_SPARK_ENGINE_BUILD_COMMAND[@]}\n"
 
 "${BUILD_COMMAND[@]}"
-"${EXTRA_SPARK_ENGINE_BUILD_COMMAND}"
+"${EXTRA_SPARK_ENGINE_BUILD_COMMAND[@]}"
 
 # Make directories
 rm -rf "$DISTDIR"
--- a/build/dist
+++ b/build/dist
@@ -248,7 +248,18 @@ echo -e "\nBuilding with..."
 # shellcheck disable=SC2145
 echo -e "\$ ${BUILD_COMMAND[@]}\n"
 
+# shellcheck disable=SC2050
+if [ "$SCALA_VERSION" = "2.12" ]; then
+  EXTRA_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@  -Pscala-2.13 -pl :kyuubi-spark-sql-engine_2.13 )
+else
+  EXTRA_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@  -Pscala-2.12 -pl :kyuubi-spark-sql-engine_2.12 )
+fi
+
+# shellcheck disable=SC2145
+echo -e "\$ ${EXTRA_BUILD_COMMAND[@]}\n"
+
 "${BUILD_COMMAND[@]}"
+"${EXTRA_BUILD_COMMAND[@]}"
 
 # Make directories
 rm -rf "$DISTDIR"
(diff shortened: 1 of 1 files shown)


build/dist
```diff
@@ -248,7 +248,18 @@ echo -e "\nBuilding with..."
 # shellcheck disable=SC2145
 echo -e "\$ ${BUILD_COMMAND[@]}\n"
 
+# shellcheck disable=SC2050
+if [ "$SCALA_VERSION" = "2.12" ]; then
+  EXTRA_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@  -Pscala-2.13 -pl :kyuubi-spark-sql-engine_2.13 )
```
pan3793, line 253: `-am` is required, otherwise the mvn command should fail if you don't have .m2 cache

nit: one space between words. EXTRA_BUILD_COMMAND => EXTRA_SPARK_ENGINE_BUILD_COMMAND. `else` is required to safety

build/dist
```diff
@@ -296,7 +307,9 @@ done
 cp "$KYUUBI_HOME/externals/kyuubi-flink-sql-engine/target/kyuubi-flink-sql-engine_${SCALA_VERSION}-${VERSION}.jar" "$DISTDIR/externals/engines/flink/"
 
 # Copy spark engines
-cp "$KYUUBI_HOME/externals/kyuubi-spark-sql-engine/target/kyuubi-spark-sql-engine_${SCALA_VERSION}-${VERSION}.jar" "$DISTDIR/externals/engines/spark/"
+for SUPPORT_VERSION in 2.12 2.13; do
```
pan3793, line 310: SUPPORT_VERSION is not a good var name here, as SCALA_VERSION is occupied, simplify use `i`

pan3793:
@PorterZhang2021 thanks for making this PR. We haven't tested the scripts under Cygwin, but only macOS and Linux, it would be good to add a manual E2E test - create a kyuubi tarball, and use it to bootstrap both Spark 2.12 and 2.13 engines.

build/dist
```diff
@@ -248,7 +248,18 @@ echo -e "\nBuilding with..."
 # shellcheck disable=SC2145
 echo -e "\$ ${BUILD_COMMAND[@]}\n"
 
+# shellcheck disable=SC2050
+if [ "$SCALA_VERSION" = "2.12" ]; then
+  EXTRA_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@  -Pscala-2.13 -pl :kyuubi-spark-sql-engine_2.13 )
+else
+  EXTRA_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@  -Pscala-2.12 -pl :kyuubi-spark-sql-engine_2.12 )
```
yikf, line 255: Thanks @PorterZhang2021 contribution, there are some minor comments:

- Apache kyuubi does not have a profile of the scala-2.12, you can use the default profile for scala-2.12;
- `$@` is actually unnecessary;

You can also [create a kyuubi tarball](https://kyuubi.readthedocs.io/en/master/contributing/code/distribution.html), and use it to manual E2E test.

PorterZhang2021:
> @PorterZhang2021 thanks for making this PR. We haven't tested the scripts under Cygwin, but only macOS and Linux, it would be good to add a manual E2E test - create a kyuubi tarball, and use it to bootstrap both Spark 2.12 and 2.13 engines.感谢你做这个PR。我们还没有在Cygwin下测试这些脚本，但只测试了macOS和Linux下的脚本，最好添加一个手动的E2E测试——创建一个kyubi tarball，并用它来引导Spark 2.12和2.13引擎。

ok, I will try it, thanks for your help.

PorterZhang2021:
@pan3793 @yikf Hello everyone, I have attempted to make a new modification based on your feedback. However, removing `'$@'` resulted in a build error, so I did not make any changes.
The following results are from E2E test:
1. when I use spark 1.2
2 12版本spark
2 12版本spark启动
2 12语句验证情况
2. when I use spark 1.3
2 13版本spark启动
设置2 13环境
image

build/dist
pan3793: `$@` should be persisted, it's user passed custom mvn args

pan3793 (approved):
LGTM, one minor tip, you don't need to reboot the Kyuubi to switch the SPARK_HOME and SPARK_CONF_DIR.

- How does one Kyuubi instance support multiple versions/types of the engine(Spark or Flink or other engines)?

> Kyuubi runs a shell command to launch an engine instance if necessary. For the Spark engine, it calls `$SPARK_HOME/bin/spark-submit ...` to launch a Spark application as an engine instance, so users can overwrite `kyuubi.engine.type` and [the related environment variables](https://kyuubi.readthedocs.io/en/master/configuration/settings.html#environments) to use a different engine client.
> ```
> jdbc:hive2://kyuubi:10009/default;#spark.app.name=my-app;kyuubi.engine.type=SPARK_SQL;kyuubi.engineEnv.SPARK_HOME=/opt/spark-3.2
> ```
[FAQ](https://github.com/apache/kyuubi/discussions/2481)

pan3793:
Thanks, merged to master

build/dist
yikf: yea, but in the case of actually compiling only the spark engine jar, the additional parameters may not be necessary and may instead affect the compilation of the current module;

For example, if the user create a kyuubi tarball compiled by `./dev/dist -Pscala-2.13`, we need to compile the spark engine jar of `scala-2.12`. then the compile command for this block becomes 
```shell
"$MVN" install $MVN_DIST_OPT **-Pscala-2.13** ~~-Pscala-2.12~~ -pl: Kyuubi-spark-SQl-engine_2.2.12 -am?
```

BTW, kyuubi doesn't seem to have a profile for scala-2.12?

pan3793: Hmm... this definitely is an issue, but we can not skip all user-provided mvn args too. we need to rethink and make a followup

pan3793: > kyuubi doesn't seem to have a profile for scala-2.12?

yes, one more thing we need to verify, if both `scala-2.12` and `scala-2.13` exist, which one will take effect if both profiles are activated

yikf: It is better not to rely on some small feature of maven to complicate things, if only the spark engine jar is compiled, what user-provided parameters might be useful?

pan3793: for example, override dep's version

yikf: Oh, I see. then we need another way to solve this problem.

PorterZhang2021:
Hello, everyone @pan3793 @yikf, I have tried two solutions, and the following is my train of thought:
# Solution 1
## Steps
I found a way to use `<profiles>`, roughly as follows:
```xml
<profiles>
        <profile>
            <id>scala-2.12</id>
            <activation>
                <activeByDefault>true</activeByDefault>
            </activation>
            <properties>
            </properties>
        </profile>
        <profile>
            <id>scala-2.13</id>
            <properties>
                <scala.binary.version>2.13</scala.binary.version>
            </properties>
        </profile>
    </profiles>
```
After specifying, I attempted to use 
1. `/ Build/mvn install - DskipTests - Dmaven. savadoc. skip=true - Dmaven. scaladoc. skip=true - Dmaven. source. skip - Pscala-2.12- Pscala-2.13- pl: kyuubi spark SQL engine - am ` 
2. `/ Build/mvn install - DskipTests - Dmaven. savadoc. skip=true - Dmaven. scaladoc. skip=true - Dmaven. source. skip - Pscala-2.13- Pscala-2.12- pl: kyuubi spark SQL engine - am ` 
## Problem
But in the end, it was found that if both '- Pscala2.12' and '- Pscala2.13' are used at the same time, '- Pscala2.13' will be selected by default, which may not be a good solution to this problem.
# Solution 2
Later, I thought about whether it was possible to filter the parameter '$@' internally. I tried writing the following code
## Steps

1. code
```sh
filtered_args=()
for arg in "$@"; do
    if [ "$arg" != "-Pscala2.12" ]; then
        filtered_args+=("$arg")
    fi
done

echo "mvn install ${filtered_args[@]}"
```
2. test
then I try it, after using the ` /test-shell.sh -Pscala2.12 -Etest` command, the final result is as follows：
image

This approach seems to meet our requirements. If you think this solution can be tried, I will conduct testing. At the same time, I also hope that you can provide me with some parameters that may be needed for `$@`, so that I can conduct some testing to prevent any problems from occurring.

Thank you for your help and review.

pan3793:
Cherry-picked to 1.9.1 to support Spark 4.0.0-preview1

PR #6306, opened 2024-04-15
[KYUUBI #5314] Support JDK 21

### LINKED ISSUE #5314: [TASK][EASY] Support JDK 21
### Code of Conduct

### Search before asking

### Skill requirements

- familiar with GitHub Action
- basic knowledge of Apache Maven, including profiles

Java 21 and the JDK 21 release under a general availability (GA) status occurred Sep 19, 2023, as the newest major update to the Java programming language.

More details on Java 21 (JDK 21) GA release via the [JDK-dev mailing list](https://mail.openjdk.org/pipermail/jdk-dev/2023-September/008267.html). More details on all of the major improvements with Java 21 can be found via [OpenJDK.org](https://openjdk.org/projects/jdk/21/).

Kyuubi has supported Java 17 since 1.8, it would be nice if we could support Java 21.

### How should we improve?

Step 0: wait for https://github.com/apache/kyuubi/issues/6293
Step 1: test Kyuubi on JDK 21 and fix errors encountered
Step 2: add CI in GitHub Actions
Step 3(out of scope): leverage new features introduced in JDK 21, e.g. Virtual Thread

Spark 3.5 almost supports Java 21 (except for Arrow, see SPARK-45781), to pass Kyuubi's CI, we need to either 
- disable arrow-related tests; or
- replace arrow jars under Spark 3.5 runtime in CI workflow, just as we did for Hive https://github.com/apache/kyuubi/pull/5280

dupen01:
# :mag: Description
## Issue References 🔗

This pull request fixes #5314 

## Describe Your Solution 🔧
To support Java 21, it would be necessary to upgrade the `spotless-maven-plugin` to a more recent version, such as from 2.30.0 to 2.43.0, as per https://github.com/diffplug/spotless/issues/1819. 
Correspondingly, `googlejavaformat` would also need to be upgraded, which would result in the loss of support for Java 8 (as documented in https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md). This upgrade would cause some existing code formatting checks to fail. 
Drawing inspiration from the Apache Flink community's approach, I propose skipping the `spotless-maven-plugin`'s check behavior in `java-21` profile. Once the Kyuubi community decides to drop support for Java 8, upgrading both `spotless-maven-plugin` and `googlejavaformat` to their newer versions could then be considered.

Executed `mvn clean package -DskipTests` locally under a JDK 21 environment, with no exceptions encountered. 

Packaged into a tar file and did some basic Spark engine test on my Mac, no exceptions were observed.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


pom.xml
```diff
@@ -2136,6 +2136,41 @@
...
+                    --add-opens=java.base/sun.nio.cs=ALL-UNNAMED
+                    --add-opens=java.base/sun.security.action=ALL-UNNAMED
+                    --add-opens=java.base/sun.security.tools.keytool=ALL-UNNAMED
+                    --add-opens=java.base/sun.security.x509=ALL-UNNAMED
+                    --add-opens=java.base/sun.util.calendar=ALL-UNNAMED
+                    -Djdk.reflect.useDirectMethodHandle=false
+                    -Dio.netty.tryReflectionSetAccessible=true</maven.plugin.surefire.argLine>
+                <!-- Current google format does not run on Java 21.
+                     Don't upgrade it in this profile because it formats code differently.
+                     Re-evaluate once support for Java 8 is dropped. -->
```
pan3793, line 2169: can we recover it by override `<spotless.java.googlejavaformat.version>` in this profile?

pom.xml
```diff
@@ -2136,6 +2136,41 @@
...
+            <id>java-21</id>
+            <activation>
+                <jdk>21</jdk>
+            </activation>
+            <properties>
+                <java.version>21</java.version>
+                <maven.compiler.source></maven.compiler.source>
+                <maven.compiler.target></maven.compiler.target>
+                <maven.compiler.release>${java.version}</maven.compiler.release>
+                <maven.plugin.surefire.argLine>-XX:+IgnoreUnrecognizedVMOptions
```
pan3793, line 2149: actually, we can set this property unconditionally, instead of duplicating it in different profiles

pan3793:
I merged your arrow upgrading PR, please rebase this PR to master, and add a CI coverage for Java 21

pom.xml
dupen01: But if using a new version of `googleJavaFormat` for code formatting check under Java 21, one problem is that this would cause some existing code formatting checks to fail, like this:
```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.43.0:check (default) on project kyuubi-hive-jdbc: The following files had format violations:
[ERROR]     src/main/java/org/apache/kyuubi/jdbc/KyuubiDriver.java
[ERROR]         @@ -17,6 +17,8 @@
[ERROR]          
[ERROR]          package·org.apache.kyuubi.jdbc;
[ERROR]          
[ERROR]         -/**·@deprecated·Use·`KyuubiHiveDriver`·instead.·*/
[ERROR]         +/**
[ERROR]         +·*·@deprecated·Use·`KyuubiHiveDriver`·instead.
[ERROR]         +·*/
[ERROR]          @Deprecated
[ERROR]          public·class·KyuubiDriver·extends·KyuubiHiveDriver·{}
```
If set `check` to `apply`, in which case it would work.

pan3793: If we accepts those changes, will the Java8 with old `googleJavaFormat` be happy?

dupen01:
> I merged your arrow upgrading PR, please rebase this PR on master, and add a CI coverage for Java 21

Thanks, I will synchronize from the master branch.

pan3793:
@dupen01 you don't need to recreate new PR, just sync the master branch and perform `git rebase master`

pom.xml
dupen01: I will take come tests using both Java 8 and Java 21, and then report back here.

dupen01: I upgraded the versions of `spotless` and `googlejavaformat` in the java21 profile and manually executed `mvn spotless:apply`, some code formats were adjusted by `spotless`. 

But when I switched to Java8 with old `googlejavaformat` for checking, errors occurred. It seems that the rules of the new version of `googlejavaformat` are not compatible with the old version. Perhaps before we drop support for Java8, we can only skip spotless's check in Java21. Once we abandon support for Java8, we can upgrade the versions of spotless and googlejavaformat.

pan3793: thanks for investigating, let's leave a brief comment to explain the decision, i.e. "Skip Java formatting because Java 21 requires a new version of googlejavaformat, but the format rules conflict with the existing one"

pan3793:
you can add the following part to `.github/workflows/master.yml` line 74 to add the Java 21 CI
```
          - java: 21
            spark: '3.5'
            spark-archive: ''
            exclude-tags: ''
            comment: 'normal'
```

dupen01:
My new submission includes other PR commit, and I'm not sure if there are any conflicts. Could you please review it for me? I apologize for any inconvenience this may cause you.

pan3793:
> My new submission includes other PR commit, and I'm not sure if there are any conflicts

Git is a little bit confusing for new contributors, I hope the following article and tool helps you to learn it.

[1] https://dev.to/lydiahallie/cs-visualized-useful-git-commands-37p1
[2] https://git-fork.com/

pan3793:
> java.lang.IllegalAccessException: class org.apache.kyuubi.client.ServerTestHelper cannot access class sun.security.tools.keytool.CertAndKeyGen (in module java.base) because module java.base does not export sun.security.tools.keytool to unnamed module @b121a8

looks additional module export is required

dupen01:
> > java.lang.IllegalAccessException: class org.apache.kyuubi.client.ServerTestHelper cannot access class sun.security.tools.keytool.CertAndKeyGen (in module java.base) because module java.base does not export sun.security.tools.keytool to unnamed module @b121a8
> 
> looks additional module export is required

I reproduced the exception, and it appears to be caused by the configuration in the `java-17` profile not being passed to `java-21`. After copying the configuration from the `java-17` profile to `java-21`, the exception disappeared.
```
<profile>
            <id>java-21</id>
            <activation>
                <jdk>21</jdk>
            </activation>
            <properties>
                <java.version>21</java.version>
                <maven.plugin.surefire.argLine>-XX:+IgnoreUnrecognizedVMOptions
                    --add-opens=java.base/java.lang=ALL-UNNAMED
                    --add-opens=java.base/java.lang.invoke=ALL-UNNAMED
                    --add-opens=java.base/java.lang.reflect=ALL-UNNAMED
                    --add-opens=java.base/java.io=ALL-UNNAMED
                    --add-opens=java.base/java.net=ALL-UNNAMED
                    --add-opens=java.base/java.nio=ALL-UNNAMED
                    --add-opens=java.base/java.util=ALL-UNNAMED
                    --add-opens=java.base/java.util.concurrent=ALL-UNNAMED
                    --add-opens=java.base/java.util.concurrent.atomic=ALL-UNNAMED
                    --add-opens=java.base/sun.nio.ch=ALL-UNNAMED
                    --add-opens=java.base/sun.nio.cs=ALL-UNNAMED
                    --add-opens=java.base/sun.security.action=ALL-UNNAMED
                    --add-opens=java.base/sun.security.tools.keytool=ALL-UNNAMED
                    --add-opens=java.base/sun.security.x509=ALL-UNNAMED
                    --add-opens=java.base/sun.util.calendar=ALL-UNNAMED
                    -Djdk.reflect.useDirectMethodHandle=false
                    -Dio.netty.tryReflectionSetAccessible=true</maven.plugin.surefire.argLine>
            </properties>
            <build>
...
```

dupen01:
Hi, @pan3793. 
Regarding the two errors that occurred in the CI workflow, here is my investigation:

1. Scala Test (2.13, 8, 3.5): 
Exception message:
```
Error: ] /home/runner/work/kyuubi/kyuubi/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala:83: Implicit definition should have explicit type (inferred com.zaxxer.hikari.HikariDataSource)
```
Analysis:
This is likely a new feature of Scala 2.13 that does not support the omission of explicit types when defining implicit variables. When I made the following change to the problematic code `kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala`:
```
83  -  implicit private[kyuubi] val hikariDataSource = new HikariDataSource(hikariConfig)
83  +  implicit private[kyuubi] val hikariDataSource: HikariDataSource = new HikariDataSource(hikariConfig)
```
After re-executing the Scala Test (2.13, 8, 3.5) workflow, it succeeded.
This exception seems unrelated to this issue.

2. Kyuubi and Spark Test (21, 3.5, normal): 
Exception message:
```
- max result rows *** FAILED ***
  org.apache.kyuubi.jdbc.hive.KyuubiSQLException: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 68.0 failed 1 times, most recent failure: Lost task 0.0 in stage 68.0 (TID 68) (localhost executor driver): java.lang.UnsupportedOperationException: sun.misc.Unsafe or java.nio.DirectByteBuffer.<init>(long, int) not available
	at org.apache.arrow.memory.util.MemoryUtil.directBuffer(MemoryUtil.java:174)
	at org.apache.arrow.memory.ArrowBuf.getDirectBuffer(ArrowBuf.java:229)
	at org.apache.arrow.memory.ArrowBuf.nioBuffer(ArrowBuf.java:224)
	...
```
Analysis:
I reproduced this exception locally, and it seems to be caused by a version mismatch between the arrow dependency included in the Spark 3.5 distribution package (version 12.0.0) and in Kyuubi's master branch(15.0.2) .
Testing:
After replacing the built-in `arrow-12.0.0` in the Spark 3.5 distribution package with `arrow-15.0.2`, and performing the "max result rows" test locally, the exception disappeared.

pan3793:
@dupen01

for issue 1, we can do Scala 2.13 upgrading independently.

issue 2 was mentioned in https://github.com/apache/kyuubi/issues/5314

> replace arrow jars under Spark 3.5 runtime in CI workflow, just as we did for Hive https://github.com/apache/kyuubi/pull/5280

pom.xml
pan3793: this is done in 5623ead6bdea8a9b5948c07ca48c857c2e518940

pom.xml
```diff
@@ -2136,12 +2138,66 @@
...
+        </profile>
+
         <!-- For development only, not generally applicable for all modules -->
         <profile>
             <id>scala-2.13</id>
             <properties>
                 <scala.binary.version>2.13</scala.binary.version>
-                <scala.version>2.13.8</scala.version>
+                <!-- Scala version above 2.13.11 support Java 21. -->
+                <scala.version>2.13.11</scala.version>
```
pan3793, line 2200: `maven.plugin.silencer.version` might be required to update accordingly. anyway, let's do this upgrading independently.

.github/workflows/master.yml
```diff
@@ -71,6 +71,11 @@ jobs:
             spark-archive: '-Dspark.archive.mirror=https://archive.apache.org/dist/spark/spark-3.4.2 -Dspark.archive.name=spark-3.4.2-bin-hadoop3.tgz -Pzookeeper-3.6'
             exclude-tags: '-Dmaven.plugin.scalatest.exclude.tags=org.scalatest.tags.Slow,org.apache.kyuubi.tags.DeltaTest,org.apache.kyuubi.tags.IcebergTest,org.apache.kyuubi.tags.PaimonTest,org.apache.kyuubi.tags.SparkLocalClusterTest'
             comment: 'verify-on-spark-3.4-binary'
+          - java: 21
+            spark: '3.5'
```
pan3793, line 75: FYI. 03dcedd89ed26fb21d0475b974a637e805e06d3b enabled CI for Spark 4.0, with JDK 17, and Spark 4.0 is fully compatible with Java 21.

pan3793:
@dupen01 the master branch is ready to support Java 21, please let me know if you still interested in this PR :)

dupen01:
> @dupen01 the master branch is ready to support Java 21, please let me know if you still interested in this PR :)

sure, i will pull master and do my work for this.

pan3793:
thanks for making this change, one last request, could you please enable it with Spark 4.0 test in CI https://github.com/apache/kyuubi/blob/aaef76cc168bf629768e0a3372687fdc210edf59/.github/workflows/master.yml#L58

PR #6319, opened 2024-04-18
[KYUUBI #6320] Fix terminated application pods not deleted issue when kyuubi server restarted

turboFei:
# :mag: Description
## Issue References 🔗

This pull request fixes #

We found that some pods were not deleted if kyuubi server restarted

The root cause is that:
com.google.common.cache::cleanup will not trigger removalListener action, we shall use `invalidateAll` instead.

Testing:
Nothing print for below code:
```
import com.google.common.cache.{Cache, CacheBuilder, RemovalNotification}

var removed = 0
val cache: Cache[String, String] = CacheBuilder.newBuilder()
  .removalListener((notification: RemovalNotification[String, String]) => {
    removed += 1
    println("removed: " + removed)
  }).build()

(0 until 1000).foreach { i =>
  cache.put(i.toString, i.toString)
}

cache.cleanup()
```

Replacing `cache.cleanup()` with `cache.invalidateAll()`, the output is expected.

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
@@ -265,7 +265,7 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
     enginePodInformers.clear()
 
     if (cleanupTerminatedAppInfoTrigger != null) {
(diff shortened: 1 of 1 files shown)


pan3793 (approved):

pan3793:
Thanks, merged to master/1.9/1.8

Actually, we have a CronJob to clean up terminated Pod too.

PR #6330, opened 2024-04-21
[KYUUBI #6305][FOLLOWUP] Improve package Spark SQL engine both Scala 2.12 and 2.13

PorterZhang2021:
# :mag: Description
## Issue References 🔗

This pull request fixes #6305 

### Solution 1 use `<profile>` - Inappropriate
I found a way to use <profiles>, roughly as follows:
```xml
<profile>
    <id>scala-2.12</id>
    <properties>
    </properties>
</profile>
```
After specifying, I attempted to use
1. `build/mvn install -DskipTests -Dmaven.javadoc.skip=true -Dmaven.scaladoc.skip=true - Dmaven.source.skip -Pscala-2.12 -Pscala-2.13 -pl: kyuubi-spark-sql-engine -am`
2. `build/mvn install -DskipTests -Dmaven.javadoc.skip=true -Dmaven.scaladoc.skip=true - Dmaven.source.skip -Pscala-2.13 -Pscala-2.12 -pl: kyuubi-spark-sql-engine -am`

#### Problem
But in the end, it was found that if both '-Pscala-2.12' and '-Pscala-2.13' are used at the same time, '-Pscala2.13' will be selected by default, which may not be a good solution to this problem.

### Solution2 
Later, I thought about whether it was possible to filter the parameter '$@' internally. It is effective.

## Test Plan 🧪
1. scala2.12
scala_2 12版本spark-sql-engine
scala_2 12_select 测试
2. scala2.13
image
scala_2 13版本测试
---

# Checklist 📝

**Be nice. Be informative.**


build/dist
```diff
@@ -247,18 +247,25 @@ BUILD_COMMAND=("$MVN" clean install $MVN_DIST_OPT $@)
...
+        FILTERED_ARGS+=("$arg")
+    fi
+done
 # shellcheck disable=SC2050
 if [ "$SCALA_VERSION" = "2.12" ]; then
-  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@ -Pscala-2.13 -pl :kyuubi-spark-sql-engine_2.13 -am)
+  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT -Pscala-2.13 ${FILTERED_ARGS[@]} -pl :kyuubi-spark-sql-engine_2.13 -am)
 else
-  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT $@ -Pscala-2.12 -pl :kyuubi-spark-sql-engine_2.12 -am)
+  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT -pl ${FILTERED_ARGS[@]} :kyuubi-spark-sql-engine_2.12 -am)
```
yikf, line 263:
```suggestion
  EXTRA_SPARK_ENGINE_BUILD_COMMAND=("$MVN" install $MVN_DIST_OPT ${FILTERED_ARGS[@]} -pl :kyuubi-spark-sql-engine_2.12 -am)
```

build/dist
```diff
@@ -247,18 +247,25 @@ BUILD_COMMAND=("$MVN" clean install $MVN_DIST_OPT $@)
 echo -e "\nBuilding with..."
 # shellcheck disable=SC2145
 echo -e "\$ ${BUILD_COMMAND[@]}\n"
+"${BUILD_COMMAND[@]}"
 
+FILTERED_ARGS=()
+# shellcheck disable=SC2045
+for arg in "$@"; do
+    if [[ $arg != -Pscala* ]]; then
```
yikf, line 255: I agree with the approach of dealing with parameters If there's no better way, even if it is hack to compare.

Not only are we dealing with arguments like -Pscala*, but we also need -P p1,scala-2.13, -P p1,scala-2.13,p2... -Dscala.version=2.13 and so on.

In addition, we can add some comments to describe it, and if the scala version changes in the future, it can guide the developer to clarify the meaning of this piece and determine whether it needs to change.

build/dist
```diff
@@ -247,18 +247,25 @@ BUILD_COMMAND=("$MVN" clean install $MVN_DIST_OPT $@)
 echo -e "\nBuilding with..."
 # shellcheck disable=SC2145
 echo -e "\$ ${BUILD_COMMAND[@]}\n"
+"${BUILD_COMMAND[@]}"
 
+FILTERED_ARGS=()
+# shellcheck disable=SC2045
+for arg in "$@"; do
```
yikf, line 254: The logic can be a function so that other modules/logics do not need to rely on the context of the script.

PorterZhang2021:
Hello , I have revised the filtering according to @yikf 's suggestion, but it is not easy to solve the problem of `- p scala-2.13 p1`. Therefore, I have written a simple prompt for users to use the `- pscala-2.13,p1` method, cc @yikf @pan3793

pan3793:
@PorterZhang2021 thanks for keeping work on this issue. actually, we can replace all mvn args `scala-2.12` with `scala-2.13` and vice versa

pan3793:
BTW, the code you pasted is weird, like something produced by OCR?

PorterZhang2021:
> @PorterZhang2021 thanks for keeping work on this issue. actually, we can replace all mvn args `scala-2.12` with `scala-2.13` and vice versa

ok, I will try it again, I'm not quite sure what ocr means. Perhaps there is a part of the content that I searched for Google and inquired about GPT, and then made similar implementations myself, which looks a bit strange. I will make some modifications later.

pan3793:
OCR means "Optical Character Recognition", what I mean strange part is 
image

PorterZhang2021:
cc @pan3793

pan3793 (approved):
Verified locally.

pan3793:
Thanks, merged to master

pan3793:
Cherry-picked to 1.9.1 to support Spark 4.0.0-preview1

PR #6335, opened 2024-04-24
[REST] Support uploading extra resources in creating batch jobs via REST API

bowenliang123:
# :mag: Description
## Describe Your Solution 🔧
- support creating batch jobs with uploading extra resource files
- allow uploading extra resource when creating batch jobs via REST API
- support binding the subresources to configs by customed configs, eg.`spark.submit.pyFiles`.

#### Related Unit Tests
+ new test

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -137,17 +137,17 @@ object Utils extends Logging {
   /**
    * Delete a directory recursively.
    */
-  def deleteDirectoryRecursively(f: File, forceDelete: Boolean = true): Unit = {
+  def deleteDirectoryRecursively(f: File, ignoreException: Boolean = true): Unit = {
     if (f.isDirectory) {
       val files = f.listFiles
       if (files != null && files.nonEmpty) {
-        files.foreach(deleteDirectoryRecursively(_, forceDelete))
+        files.foreach(deleteDirectoryRecursively(_, ignoreException))
       }
     }
     try {
       f.delete()
     } catch {
-      case e: Exception if !forceDelete => throw e
+      case e: Exception if !ignoreException => throw e
     }
   }
 
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -149,7 +149,7 @@ object Utils extends Logging {
     } catch {
       case e: Exception =>
         if (ignoreException) {
-          warn(s"ignored the exception when deleting file: ${f.toPath}", e)
+          warn(s"Ignoring the exception in deleting file, path: ${f.toPath}", e)
         } else {
           throw e
         }
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -147,7 +147,12 @@ object Utils extends Logging {
     try {
       f.delete()
     } catch {
-      case e: Exception if !ignoreException => throw e
+      case e: Exception =>
+        if (ignoreException) {
+          warn(s"ignored the exception when deleting file: ${f.toPath}")
+        } else {
+          throw e
+        }
     }
   }
 
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -149,7 +149,7 @@ object Utils extends Logging {
     } catch {
       case e: Exception =>
         if (ignoreException) {
-          warn(s"ignored the exception when deleting file: ${f.toPath}")
+          warn(s"ignored the exception when deleting file: ${f.toPath}", e)
         } else {
           throw e
         }
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -137,12 +137,18 @@ object Utils extends Logging {
   /**
    * Delete a directory recursively.
    */
-  def deleteDirectoryRecursively(f: File): Boolean = {
-    if (f.isDirectory) f.listFiles match {
-      case files: Array[File] => files.foreach(deleteDirectoryRecursively)
-      case _ =>
+  def deleteDirectoryRecursively(f: File, forceDelete: Boolean = true): Unit = {
+    if (f.isDirectory) {
+      val files = f.listFiles
+      if (files != null && files.nonEmpty) {
+        files.foreach(deleteDirectoryRecursively(_, forceDelete))
+      }
     }
-    f.delete()
+    try {
+      f.delete()
+    } catch {
+      case e: Exception if !forceDelete => throw e
+    }
   }
 
   /**
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
@@ -58,8 +58,14 @@ public class BatchRestApi {
     extraResources.stream()
         .distinct()
         .filter(StringUtils::isNotBlank)
-        .map(path -> Paths.get(path).toFile())
-        .filter(File::exists)
+        .map(
+            path -> {
+              File file = Paths.get(path).toFile();
+              if (!file.exists()) {
+                throw new RuntimeException("File not existed, path: " + path);
+              }
+              return file;
+            })
         .forEach(
             file ->
                 multiPartMap.put(
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
@@ -18,8 +18,9 @@
 package org.apache.kyuubi.client;
 
 import java.io.File;
-import java.util.HashMap;
-import java.util.Map;
+import java.nio.file.Paths;
+import java.util.*;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.kyuubi.client.api.v1.dto.*;
 import org.apache.kyuubi.client.util.JsonUtils;
 import org.apache.kyuubi.client.util.VersionUtils;
@@ -46,10 +47,23 @@ public class BatchRestApi {
   }
 
   public Batch createBatch(BatchRequest request, File resourceFile) {
+    return createBatch(request, resourceFile, Collections.emptyList());
+  }
+
+  public Batch createBatch(BatchRequest request, File resourceFile, List<String> extraResources) {
     setClientVersion(request);
     Map<String, MultiPart> multiPartMap = new HashMap<>();
     multiPartMap.put("batchRequest", new MultiPart(MultiPart.MultiPartType.JSON, request));
     multiPartMap.put("resourceFile", new MultiPart(MultiPart.MultiPartType.FILE, resourceFile));
+    extraResources.stream()
+        .distinct()
+        .filter(StringUtils::isNotBlank)
+        .map(path -> Paths.get(path).toFile())
+        .filter(File::exists)
+        .forEach(
+            file ->
+                multiPartMap.put(
+                    file.getName(), new MultiPart(MultiPart.MultiPartType.FILE, file)));
     return this.getClient().post(API_BASE_PATH, multiPartMap, Batch.class, client.getAuthHeader());
   }
 
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
@@ -31,6 +31,7 @@ public class BatchRequest {
   private String name;
   private Map<String, String> conf = Collections.emptyMap();
   private List<String> args = Collections.emptyList();
+  private Map<String, String> extraResourcesMap = Collections.emptyMap();
 
   public BatchRequest() {}
 
@@ -110,6 +111,14 @@ public class BatchRequest {
     this.args = args;
   }
 
+  public Map<String, String> getExtraResourcesMap() {
+    return extraResourcesMap;
+  }
+
+  public void setExtraResourcesMap(Map<String, String> extraResourcesMap) {
+    this.extraResourcesMap = extraResourcesMap;
+  }
+
   @Override
   public boolean equals(Object o) {
     if (this == o) return true;
@@ -120,13 +129,20 @@ public class BatchRequest {
         && Objects.equals(getClassName(), that.getClassName())
         && Objects.equals(getName(), that.getName())
         && Objects.equals(getConf(), that.getConf())
-        && Objects.equals(getArgs(), that.getArgs());
+        && Objects.equals(getArgs(), that.getArgs())
+        && Objects.equals(getExtraResourcesMap(), that.getExtraResourcesMap());
   }
 
   @Override
   public int hashCode() {
     return Objects.hash(
-        getBatchType(), getResource(), getClassName(), getName(), getConf(), getArgs());
+        getBatchType(),
+        getResource(),
+        getClassName(),
+        getName(),
+        getConf(),
+        getArgs(),
+        getExtraResourcesMap());
   }
 
   @Override
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -17,7 +17,6 @@
 
 package org.apache.kyuubi.operation
 
-import java.nio.file.{Files, Paths}
 import java.util.Locale
 import java.util.concurrent.TimeUnit
 
@@ -395,11 +394,7 @@ class BatchJobSubmission(
 
   private def cleanupUploadedResourceIfNeeded(): Unit = {
     if (session.isResourceUploaded) {
-      try {
-        Files.deleteIfExists(Paths.get(resource))
-      } catch {
-        case e: Throwable => error(s"Error deleting the uploaded resource: $resource", e)
-      }
+      Utils.deleteDirectoryRecursively(session.resourceUploadFolderPath.toFile)
     }
   }
 }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -546,13 +546,19 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       resourceFileName: String,
       formDataMultiPartOpt: Option[FormDataMultiPart]): Option[JPath] = {
     val uploadFileFolderPath = batchResourceUploadFolderPath(batchId)
-    handleUploadingResourceFile(
-      request,
-      resourceFileInputStream,
-      resourceFileName,
-      uploadFileFolderPath)
-    handleUploadingExtraResourcesFiles(request, formDataMultiPartOpt, uploadFileFolderPath)
-    Some(uploadFileFolderPath)
+    try {
+      handleUploadingResourceFile(
+        request,
+        resourceFileInputStream,
+        resourceFileName,
+        uploadFileFolderPath)
+      handleUploadingExtraResourcesFiles(request, formDataMultiPartOpt, uploadFileFolderPath)
+      Some(uploadFileFolderPath)
+    } catch {
+      case e: Exception =>
+        Utils.deleteDirectoryRecursively(uploadFileFolderPath.toFile)
+        throw e
+    }
   }
 
   private def handleUploadingResourceFile(
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -18,6 +18,7 @@
 package org.apache.kyuubi.server.api.v1
 
 import java.io.InputStream
+import java.nio.file.{Path => JPath}
 import java.util
 import java.util.{Collections, Locale, UUID}
 import java.util.concurrent.ConcurrentHashMap
@@ -32,7 +33,7 @@ import io.swagger.v3.oas.annotations.media.{Content, Schema}
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.tags.Tag
 import org.apache.commons.lang3.StringUtils
-import org.glassfish.jersey.media.multipart.{FormDataContentDisposition, FormDataParam}
+import org.glassfish.jersey.media.multipart.{FormDataContentDisposition, FormDataMultiPart, FormDataParam}
 
 import org.apache.kyuubi.{Logging, Utils}
 import org.apache.kyuubi.client.api.v1.dto._
@@ -190,7 +191,8 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
   def openBatchSessionWithUpload(
       @FormDataParam("batchRequest") batchRequest: BatchRequest,
       @FormDataParam("resourceFile") resourceFileInputStream: InputStream,
-      @FormDataParam("resourceFile") resourceFileMetadata: FormDataContentDisposition): Batch = {
+      @FormDataParam("resourceFile") resourceFileMetadata: FormDataContentDisposition,
+      formDataMultiPart: FormDataMultiPart): Batch = {
     require(
       fe.getConf.get(BATCH_RESOURCE_UPLOAD_ENABLED),
       "Batch resource upload function is disabled.")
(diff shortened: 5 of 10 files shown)


.rat-excludes
```diff
@@ -54,3 +54,4 @@ build/scala-*/**
 **/BeeLine.properties
 **/sql-keywords.properties
 **/python/**
+**/*.py
```
turboFei, line 57: we also have other py source files in project, so we can ignore all the py files

bowenliang123: I do ignore all the pyfiles here, as rat checks may not be suitable for python scripts?
Do you mean 'we cannnot'  ignore all the pyfiles?

turboFei: yes, we can not

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
```diff
@@ -32,6 +32,10 @@ public class BatchRequest {
   private Map<String, String> conf = Collections.emptyMap();
   private List<String> args = Collections.emptyList();
 
+  private Map<String, String> subresourcesMap = Collections.emptyMap();
+
+  private Map<String, String> transformedSubresourcesMap = Collections.emptyMap();
```
turboFei, line 37: can we reuse the `subresourcesMap` just similar with `resource`?

turboFei: I mean `setSubresourcesMap` in server side.

turboFei: otherwise, it is confused, as it is internal but visible for customers.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -314,7 +317,9 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
           metadata.className,
           metadata.requestArgs,
           Some(metadata),
-          fromRecovery = true)
+          fromRecovery = true,
+          // todo: subResources
+          subResources = Map.empty[String, String])
```
turboFei, line 322: seems all the uploaded sub resources are used by batch config,

can we add a reserved key into batch conf?

And get the uploaded files list(to cleanup) from batch conf when batch got recovered.

turboFei: then we do not need to persist the sub resources anymore.

kyuubi-server/src/test/resources/python/app.py
```diff
@@ -0,0 +1,20 @@
+from module1.module import func1
```
turboFei, line 1: add license to pass the rat check?

.rat-excludes
bowenliang123: OK. Removed this exclusion for python scripts.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -190,7 +190,8 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
   def openBatchSessionWithUpload(
       @FormDataParam("batchRequest") batchRequest: BatchRequest,
       @FormDataParam("resourceFile") resourceFileInputStream: InputStream,
-      @FormDataParam("resourceFile") resourceFileMetadata: FormDataContentDisposition): Batch = {
+      @FormDataParam("resourceFile") resourceFileMetadata: FormDataContentDisposition,
+      multiPart: FormDataMultiPart): Batch = {
```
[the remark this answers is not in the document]
bowenliang123: The consumed media type is not changed, it remains `MediaType.MULTIPART_FORM_DATA`.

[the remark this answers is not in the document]
turboFei: but it is APPLICATION_JSON in the api annotation.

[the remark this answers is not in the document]
turboFei: should we separate these two APIs? cc @pan3793 

As it provide the wrong prompt in the swagger UI.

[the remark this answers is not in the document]
bowenliang123: Line 189 `@Consumes(Array(MediaType.MULTIPART_FORM_DATA))` contorls the accepted media type for the request.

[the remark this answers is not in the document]
turboFei: ok, thanks for the info.

pan3793:
I'm not sure the `subresources` is good name here, I would call it `extraResources`

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
```diff
@@ -53,6 +54,20 @@ public Batch createBatch(BatchRequest request, File resourceFile) {
     return this.getClient().post(API_BASE_PATH, multiPartMap, Batch.class, client.getAuthHeader());
   }
 
+  public Batch createBatchWithSubResources(
+      BatchRequest request, File resourceFile, List<String> subResources) {
```
pan3793, line 58: we don't need a different name, just use `createBatch`
```suggestion
  public Batch createBatch(
      BatchRequest request, File mainResource, List<String> extraResources) {
```

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
pan3793: I have the same concerns. we should not leak a internal used parameter to user faced api

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -203,6 +204,20 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+      val targetConfKey = a.getKey
+      val subresourceFileName = a.getValue
+      val filePart = multiPart.getField(subresourceFileName)
+      val subresourceFile = Utils.writeToTempFile(
+        filePart.getValueAs(classOf[InputStream]),
+        KyuubiApplicationManager.uploadWorkDir,
+        filePart.getContentDisposition.getFileName)
+      (targetConfKey, subresourceFile.getPath)
+    }.toMap.asJava
+    batchRequest.setTransformedSubresourcesMap(transformedSubMap)
```
pan3793, line 219: I'm confused here, for the main resource, we just replace the reference directly, for extra resources, why not find the references and replace them directly too?

bowenliang123:
> I'm not sure the `subresources` is good name here, I would call it `extraResources`

SGTM. Let me change to extra resources.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
bowenliang123: LGTM. Fixed.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
bowenliang123: Thanks for the review. `transformedSubresourcesMap` has been removed from the request DTO. Please have a look.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: Yes, changed to replacing them directly.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
```diff
@@ -53,6 +54,19 @@ public Batch createBatch(BatchRequest request, File resourceFile) {
...
 
+  public Batch createBatch(BatchRequest request, File resourceFile, List<String> extraResources) {
+    setClientVersion(request);
+    Map<String, MultiPart> multiPartMap = new HashMap<>();
+    multiPartMap.put("batchRequest", new MultiPart(MultiPart.MultiPartType.JSON, request));
+    multiPartMap.put("resourceFile", new MultiPart(MultiPart.MultiPartType.FILE, resourceFile));
+    for (String path : extraResources) {
+      File file = new File(path);
+      String fileName = file.getName();
+      multiPartMap.put(fileName, new MultiPart(MultiPart.MultiPartType.FILE, file));
```
turboFei, line 65: how do you define the mapping between config and uploaded files?

For example: `spark.submit.pyFiles` -> extraResource?

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
```diff
@@ -110,6 +111,14 @@ public void setArgs(List<String> args) {
     this.args = args;
   }
 
+  public Map<String, String> getExtraResourcesMap() {
+    return extraResourcesMap;
```
turboFei, line 115: return emptyMap if it is null

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
bowenliang123: Here in L165 https://github.com/apache/kyuubi/pull/6335/files/d35219f9f388a1fb91e12c2f008730e5aabf015b#diff-bb6880107c3d6bf608f287bbbc8f59f8705b72818473bff46d348844855d91acR165.
Mapping is marked via  extraResourcesMap.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -401,6 +402,17 @@ class BatchJobSubmission(
         case e: Throwable => error(s"Error deleting the uploaded resource: $resource", e)
       }
     }
+
+    if (session.isExtraResourcesUploaded) {
+      // cleanup extra resources
+      extraResources.values.foreach { extraResource =>
```
turboFei, line 408: if the `extraResources` field is only used when cleaning up.

Could you delete the session upload path directly?

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
bowenliang123: extraResourcesMap was initialed as empty map in L34. (https://github.com/apache/kyuubi/pull/6335/files/d35219f9f388a1fb91e12c2f008730e5aabf015b#diff-5dffcdb403904677ed5c8e8d4c997513aa236dc4cfa828a1e0dad1d9bec93644L34)

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/api/Metadata.scala
```diff
@@ -81,7 +81,8 @@ case class Metadata(
     // keep consistent with table creation DDL
     // find why we set 10 as default in KYUUBI #5329
     priority: Int = 10,
-    peerInstanceClosed: Boolean = false) {
+    peerInstanceClosed: Boolean = false,
+    extraResources: Map[String, String] = Map.empty) {
```
turboFei, line 85: Seems not used here, we can remove it.

turboFei:
one more question, in this PR, does it support batch recovery with extra sub resources?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
pan3793: +1 for this idea. We can upload all resources into `$BASE_DIR/<session_id>/` and clean the folder recursively, then we don't need to track the whole resource file list.

pan3793:
general idea:
1. for batch impl v1, `spark-submit` runs immediately on the same kyuubi instance after "create batch" request is accepted, and the existing mechanism ensures that each batch job must be recovered from the same kyuubi instance, plus the above "session folder for extra resources idea", we don't need to track the resource file list, so no extra metadata(and DB schema change) is required, right?
2. for batch impl v2, "create batch" request just inserts a record into DB and waits for all kyuubi instances to pick up, but for batch jobs that have uploaded resource files, we must ensure that the job will be picked up by the kyuubi instance that accepted the "create batch" request. I think we currently have such issues, which can be addressed later.

bowenliang123:
Agree with @pan3793. We could consider address the problem for V2 in the future, or just disable the uploading files for V1.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/api/Metadata.scala
bowenliang123: Fixed.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
bowenliang123: Good suggestion. Please review the latest version. It puts all the uploaded files of resource and extra resources into `$BASE_DIR/<batch_id>/` folder which will be cleanup afterwards.

dev/dependencyList
```diff
@@ -30,6 +30,7 @@ arrow-vector/16.0.0//arrow-vector-16.0.0.jar
 checker-qual/3.42.0//checker-qual-3.42.0.jar
 classgraph/4.8.138//classgraph-4.8.138.jar
 commons-codec/1.15//commons-codec-1.15.jar
+commons-io/2.11.0//commons-io-2.11.0.jar
```
pan3793, line 33: LICENSE-binary and NOTICE-binary should be updated too

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -190,20 +191,21 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
       batchRequest != null,
       "batchRequest is required and please check the content type" +
         " of batchRequest is application/json")
-    val tempFile = Utils.writeToTempFile(
-      resourceFileInputStream,
-      KyuubiApplicationManager.uploadWorkDir,
-      resourceFileMetadata.getFileName)
-    batchRequest.setResource(tempFile.getPath)
-    openBatchSessionInternal(batchRequest, isResourceFromUpload = true)
+    openBatchSessionInternal(
```
pan3793, line 203: can we handle upload inside function `openBatchSessionWithUpload`? otherwise, this function looks a little bit redundant

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -525,22 +538,106 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+  private def handleUploadingExtraResourcesFiles(
+      request: BatchRequest,
+      formDataMultiPartOpt: Option[FormDataMultiPart],
+      uploadFileFolderPath: JPath): Unit = {
+    val extraResourceMap = request.getExtraResourcesMap.asScala
+    if (extraResourceMap.nonEmpty) {
+      val fileNameSeperator = ","
+      val formDataMultiPart = formDataMultiPartOpt.get
+      val transformedExtraResourcesMap = extraResourceMap
+        .map { case (confKey, confValue) =>
```
pan3793, line 583: I remember there is a function `mapValues`

pan3793 (commented):
overall lgtm, some nits

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: Sure. Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: No. In my first draft, I do the uploading and file saving in function `openBatchSessionWithUpload`. But then moved into `openBatchSessionInternal`, for two reasons, 1. we need the batchId for uploading path , 2. do more checks on request before saving file preventing unnecessary local files.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -525,22 +538,105 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+          confValue.split(fileNameSeperator).filter(StringUtils.isNotBlank(_)))
+        .filter { case (confKey, fileNames) =>
+          fileNames.nonEmpty && StringUtils.isNotBlank(confKey)
+        }.mapValues { fileNames =>
+          fileNames.map(fileName =>
+            Option(formDataMultiPart.getField(fileName))
+              .getOrElse(throw new RuntimeException(s"File part for file $fileName not found")))
+        }.map {
+          case (confKey, fileParts) =>
+            val tempFilePaths = fileParts.map(filePart => {
```
pan3793, line 593: nit: for multi lines lambda, `{` is always preferred than `(`
```suggestion
            val tempFilePaths = fileParts.map { filePart =>
```

pan3793 (approved):

dev/dependencyList
bowenliang123: Addressed.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: Thanks for the hint. Resoved.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -395,11 +395,7 @@ class BatchJobSubmission(
 
   private def cleanupUploadedResourceIfNeeded(): Unit = {
     if (session.isResourceUploaded) {
-      try {
-        Files.deleteIfExists(Paths.get(resource))
-      } catch {
-        case e: Throwable => error(s"Error deleting the uploaded resource: $resource", e)
-      }
+      FileUtils.deleteQuietly(session.resourceUploadFolderPath.toFile)
```
turboFei, line 398: You can use `org.apache.kyuubi.Utils.deleteDirectoryRecursively` and remove the common-io dependency.

turboFei:
raised a PR for resource uploaded: https://github.com/apache/kyuubi/pull/6592

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: The only concern is that, deleteDirectoryRecursively will break on exception.

bowenliang123: OK, changed to use `org.apache.kyuubi.Utils.deleteDirectoryRecursively`. And it's made sure it won't break on exceptions and continue to delete the whole folder.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/BatchRestApi.java
```diff
@@ -46,10 +47,23 @@ public Batch createBatch(BatchRequest request) {
...
+  public Batch createBatch(BatchRequest request, File resourceFile, List<String> extraResources) {
     setClientVersion(request);
     Map<String, MultiPart> multiPartMap = new HashMap<>();
     multiPartMap.put("batchRequest", new MultiPart(MultiPart.MultiPartType.JSON, request));
     multiPartMap.put("resourceFile", new MultiPart(MultiPart.MultiPartType.FILE, resourceFile));
+    extraResources.stream()
+        .distinct()
+        .filter(StringUtils::isNotBlank)
+        .map(path -> Paths.get(path).toFile())
+        .filter(File::exists)
```
turboFei, line 62: do not need to filter, throw RuntimeException if not exists.

bowenliang123: OK. Fixed.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -243,6 +248,14 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
         markDuplicated(batch)
       case None =>
         val batchId = userProvidedBatchId.getOrElse(UUID.randomUUID().toString)
+        if (isResourceFromUpload) {
+          handleUploadingFiles(
+            batchId,
+            request,
+            resourceFileInputStream.get,
+            resourceFileMetadata.get.getFileName,
+            formDataMultiPartOpt)
```
turboFei, line 257: set BATCH_IMPL_VERSION to 1 if isResourceFromUpload?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -525,22 +538,104 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+      resourceFileName: String,
+      formDataMultiPartOpt: Option[FormDataMultiPart]): Option[JPath] = {
+    val uploadFileFolderPath = batchResourceUploadFolderPath(batchId)
+    handleUploadingResourceFile(
+      request,
+      resourceFileInputStream,
+      resourceFileName,
+      uploadFileFolderPath)
+    handleUploadingExtraResourcesFiles(request, formDataMultiPartOpt, uploadFileFolderPath)
+    Some(uploadFileFolderPath)
```
turboFei, line 555: cleanup the session upload path if exception thrown

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: How about raising a separate PR to check or override the batch implement version ?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
bowenliang123: Sure. Fixed.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -137,12 +137,18 @@ object Utils extends Logging {
...
+    if (f.isDirectory) {
+      val files = f.listFiles
+      if (files != null && files.nonEmpty) {
+        files.foreach(deleteDirectoryRecursively(_, forceDelete))
+      }
+    }
+    try {
+      f.delete()
+    } catch {
+      case e: Exception if !forceDelete => throw e
```
turboFei, line 150: I think you can rename `forceDelete` to `ignoreException`, or others, due it does not force delete the file if exception thrown.

And we can log exception error here.

bowenliang123: No problem. `ignoreException` was the original param name in my first draft, changing back to it.  Fixed.

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -147,7 +147,12 @@ object Utils extends Logging {
     try {
       f.delete()
     } catch {
-      case e: Exception if !ignoreException => throw e
+      case e: Exception =>
+        if (ignoreException) {
+          warn(s"ignored the exception when deleting file: ${f.toPath}")
```
turboFei, line 152: warn("...", e)

turboFei (approved):
LGTM overall, thanks

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
turboFei: cc @pan3793

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
bowenliang123: nice. Fixed.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
pan3793: > set BATCH_IMPL_VERSION to 1 if isResourceFromUpload?

yes, we should do that for now, I'm fine to do it in place or a separate PR

pan3793 (approved):

bowenliang123:
Thanks. Merged to master (1.10.0).

Issue #4847, opened 2023-05-17
[Bug] close the session immediately when engine corrupt


huangzhir:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

While using the Kyuubi REST functionality, I encountered the following issues:

 I created a session and submitted a statement. However, this statement had an issue that caused the Spark engine to crash. For example, I submitted a spark.stop Scala code snippet.
POST http://127.0.0.1:10099/api/v1/sessions/5536200f-cfbc-44ff-a5fd-05a34c9892ab/operations/statement
``` 
{
    "statement": "spark.close",
    "runAsync": true,
    "confOverlay": {
        "kyuubi.operation.language":"SCALA"
    }
}
```

Here, because the engine crashed, the Spark session is closed. However, Kyuubi still maintains this session. When using the 'list session' and 'session count' interfaces, I can see that the session is still present.Even after half an hour, the situation remains the same.

GET  http://127.0.0.1:10099/api/v1/sessions
return
```
[
    {
        "identifier": "5536200f-cfbc-44ff-a5fd-05a34c9892ab",
        "user": "anonymous",
        "ipAddr": "127.0.0.1",
        "conf": {
            "kyuubi.session.connection.url": "0.0.0.0:10099",
            "kyuubi.server.ipAddress": "0.0.0.0",
            "kyuubi.session.real.user": "anonymous",
            "kyuubi.client.ipAddress": "127.0.0.1"
        },
        "createTime": 1684309480586,
        "duration": 39452,
        "idleTime": 0,
        "exception": "",
        "sessionType": "INTERACTIVE",
        "kyuubiInstance": "0.0.0.0:10099",
        "engineId": "local-1684309482569"
    }
]
```

### Affects Version(s)

1.6,1.7,master

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

pan3793:
Thanks for digging into this issue, I would not say the current behavior is wrong, because it returns the error state and message to the client, and the client has the responsibility to close the session, no matter if it's healthy or corrupt.

What's the behavior of HiveServer2?

huangzhir:
[...]

PR #6355, opened 2024-05-01
[KYUUBI #4847][FOLLOWUP] Exclude the alive probe sessions in terminating checker 

turboFei:
# :mag: Description
## Issue References 🔗

This pull request fixes #
follow up of #4847 

Address comments: https://github.com/apache/kyuubi/issues/4847#issuecomment-2072945805 
## Describe Your Solution 🔧

In this pr, when checking the engine terminating, it will ignore the alive probe sessions. 

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
@@ -122,10 +122,12 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
   }
 
   def closeSession(sessionHandle: SessionHandle): Unit = {
-    _latestLogoutTime = System.currentTimeMillis()
     val session = handleToSession.remove(sessionHandle)
     if (session == null) {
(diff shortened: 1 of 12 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/session/SessionManager.scala
```diff
@@ -161,6 +163,11 @@ abstract class SessionManager(name: String) extends CompositeService(name) {
 
   def getOpenSessionCount: Int = handleToSession.size()
```
wForget, line 164: Does this method still make sense? If not, I think we can remove it to avoid it being used incorrectly.

turboFei: addressed

wForget (approved):
LGTM

turboFei:
thanks, merged to master/1.9.1/1.8.3

PR #6359, opened 2024-05-05
[REST] Return more fields within session/batch data

turboFei:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

Return `sessionName` and `totalOperations` within `SessionData`.
Return `appStartTime` with list batches.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/SessionData.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/SessionData.java
@@ -36,6 +36,8 @@ public class SessionData {
   private String sessionType;
   private String kyuubiInstance;
   private String engineId;
+  private String sessionName;
+  private int totalOperations;
 
   public SessionData() {}
 
@@ -51,7 +53,9 @@ public class SessionData {
       String exception,
       String sessionType,
       String kyuubiInstance,
-      String engineId) {
+      String engineId,
+      String sessionName,
+      int totalOperations) {
     this.identifier = identifier;
     this.remoteId = remoteId;
     this.user = user;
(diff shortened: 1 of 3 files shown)


kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/SessionData.java
```diff
@@ -36,6 +36,8 @@ public class SessionData {
   private String sessionType;
   private String kyuubiInstance;
   private String engineId;
+  private String sessionName;
+  private int totalOperations;
```
wForget, line 40: use `Integer`

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -432,6 +432,7 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
         val createTime = resultSet.getLong("create_time")
         val engineType = resultSet.getString("engine_type")
         val clusterManager = Option(resultSet.getString("cluster_manager"))
+        val engineOpenTime = resultSet.getLong("engine_open_time")
```
turboFei, line 435:
```
    engine_open_time bigint, -- the engine open time
```

wForget: Where is it assigned a value?

turboFei: let me check

turboFei: https://github.com/apache/kyuubi/blob/78e104d18b467dff5c97516c9dbe541eb8f7f276/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala#L151

wForget (approved):

turboFei:
thanks, merged to master

PR #6363, opened 2024-05-06
[INFRA] Increase CI concurrency

pan3793:
# :mag: Description

Previously, we were asked to limit the job concurrency to 10, while the new policy is more relaxed.

> All workflows SHOULD have a job concurrency level less than or equal to 15. Just because 20 is the max, doesn't mean you should strive for 20.

https://infra.apache.org/github-actions-policy.html

## Test Plan 🧪

Monitor CI

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -42,7 +42,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


wForget (approved):

pan3793:
Merged to master/1.9

PR #6365, opened 2024-05-06
SparkConnectEngineDemo

davidyuan1223:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


davidyuan1223:
@pan3793 @yaooqinn @ulysses-you this pr is only a draft for discussion. What do you think about this method to compatible spark grpc ?

ulysses-you:
Thank you @davidyuan1223 . We should add the grpc frontend service first, it is used to receive spark connect client request and send to connect engine.

externals/kyuubi-spark-connect-engine/src/main/scala/org/apache/kyuubi/engine/spark/connect/SparkConnectFrontendService.scala
davidyuan1223, line 1: @ulysses-you This is the grpc frontendService,currently, only support config

ulysses-you: It's engine. I mean the frontend of Kyuubi server. We can create a new module `kyuubi-grpc` for it.

davidyuan1223:
@ulysses-you what do you think about the new module kyuubi-grpc(only structure, not write test case yet)

externals/kyuubi-spark-connect-engine/src/main/resources/protobuf/kyuubi/spark/connect/base.proto
ulysses-you, line 1: @davidyuan1223 is it possbile add spark-connect dependency and just delegate request to the  SparkConnectService ? If so, we do not need to copy proto files.

ulysses-you:
> @ulysses-you what do you think about the new module kyuubi-grpc(only structure, not write test case yet)


@davidyuan1223 yeah, a new module for kyuubi grpc server is better. We can refactor and improve code after completing the connect frontend in future.

externals/kyuubi-spark-connect-engine/src/main/resources/protobuf/kyuubi/spark/connect/base.proto
davidyuan1223: yes, but i think, maybe we need to add other request/response, so my plan is put the proto files in kyuubi grpc, we can add kyuubi's interface based on the spark-connect proto, so not add the spark-connect common dependency

ulysses-you: Can you elaborate your plan ? I'm afrait things becomes complex if we decide to support connect based on proto.

davidyuan1223: > Can you elaborate your plan ? I'm afrait things becomes complex if we decide to support connect based on proto.

Actually, it's a consume, because the proto is based on spark, if we wanna support it with another engine, maybe need modify the proto? So i think put the proto in kyuubi-grpc could help us compite with other sql engine in the future. And if we wanna add spark-connect-common dependency, we need the spark version > 3.4, i think it's not compitable with other versions, because the rpc request will be handle by kyuubi's grpc backendService, not spark's rpc server, based on the proto, we could support rpc process in other spark version

ulysses-you: > So i think put the proto in kyuubi-grpc could help us compite with other sql engine in the future.

It's too early to consider that. Even if we add proto, there is no standard proto interface, how can Spark proto compatible with other engine ? For each proto based engine, I think we can also add their server dependency.

> And if we wanna add spark-connect-common dependency, we need the spark version > 3.4, i think it's not compitable with other versions, because the rpc request will be handle by kyuubi's grpc backendService, not spark's rpc server, based on the proto, we could support rpc process in other spark version

It sounds like just like copy all Spark connect server code into Kyuubi repo ? I think we only need to support connect since Spark3.5 and later, that's why we need a new module for Kyuubi Server. We do not need support older Spark version.

----

Besides, Add a Kyuubi grpc proto is fine to me for those statements, e.g., show kyuubi session/operators etc which are for Kyuubi itself.

davidyuan1223: > consume



> > So i think put the proto in kyuubi-grpc could help us compite with other sql engine in the future.
> 
> It's too early to consider that. Even if we add proto, there is no standard proto interface, how can Spark proto compatible with other engine ? For each proto based engine, I think we can also add their server dependency.
> 
> > And if we wanna add spark-connect-common dependency, we need the spark version > 3.4, i think it's not compitable with other versions, because the rpc request will be handle by kyuubi's grpc backendService, not spark's rpc server, based on the proto, we could support rpc process in other spark version
> 
> It sounds like just like copy all Spark connect server code into Kyuubi repo ? I think we only need to support connect since Spark3.5 and later, that's why we need a new module for Kyuubi Server. We do not need support older Spark version.
> 
> Besides, Add a Kyuubi grpc proto is fine to me for those statements, e.g., show kyuubi session/operators etc which are for Kyuubi itself.

You are right, i will modify them.

Issue #4515, opened 2023-03-14
[Bug] Executing `kyuubi status` returns 'Kyuubi is not running' in docker container


dnskr:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

I'm using the official docker image and helm chart to deploy and manage Kyuubi on Kubernetes.
I found that `bin/kyuubi status` returns `Kyuubi is not running` when the instance is actually up and running.

The PID value in the file should be `15`, however it is `1301` for the main instance in this case:
```bash
kyuubi@kyuubi-5b6784d6b4-gf2d4:/opt/kyuubi$ bin/kyuubi status
Warn: Not find kyuubi environment file /opt/kyuubi/conf/kyuubi-env.sh, using default ones...
Kyuubi is not running

kyuubi@kyuubi-5b6784d6b4-gf2d4:/opt/kyuubi$ ls -la pid
total 16
drwxrwxr-x 1 root   root 4096 Mar 14 19:30 .
drwxrwxr-x 1 root   root 4096 Mar 14 19:41 ..
-rw-r--r-- 1 kyuubi root    5 Mar 14 19:30 kyuubi--org.apache.kyuubi.server.KyuubiServer.pid

kyuubi@kyuubi-5b6784d6b4-gf2d4:/opt/kyuubi$ cat pid/kyuubi--org.apache.kyuubi.server.KyuubiServer.pid 
1301

kyuubi@kyuubi-5b6784d6b4-gf2d4:/opt/kyuubi$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
kyuubi         1  0.0  0.0   7112  3388 ?        Ss   19:23   0:00 bash ./bin/kyuubi run
kyuubi        15  1.0 17.2 3933096 1020332 ?     Sl   19:23   0:26 /opt/java/openjdk/bin/java -cp /opt/kyuubi/jars/*:/opt/kyuubi/conf: org.apache.kyuubi.server.KyuubiServer
kyuubi      7952  0.0  0.0   7244  3956 pts/0    Ss   20:04   0:00 bash
kyuubi      8135  0.0  0.0   8896  3184 pts/0    R+   20:05   0:00 ps aux

[...]

PR #6374, opened 2024-05-08
[KYUUBI #4515] Capturing process id for kyuubi server launched using run command

Madhukar98:
…run command

# :mag: Description
## Issue References 🔗

This pull request fixes # https://github.com/apache/kyuubi/issues/4515

## Describe Your Solution 🔧

Currently, the process ID (PID) for a Kyuubi server launched using bin/kyuubi run is not being captured. This leads to the bin/kyuubi status command incorrectly reporting that the server is not running.
Furthermore, if the bin/kyuubi run command is initiated at the launch of a Docker container, subsequent attempts to run bin/kyuubi start will fail due to the error "ports already occupied".
To resolve these issues and ensure synchronization between the status of the Kyuubi server whether it's launched in the foreground with bin/kyuubi run or in the background with bin/kyuubi start, two changes have been made:
1. A get_pid function has been created to capture the PID of the process launched using the run command.
2. A check has been added before starting a Kyuubi server to ensure that it's not already running in a different mode.

bin/kyuubi run -> Launches kyuubi server
bin/kyuubi status -> Kyuubi is not running
bin/kyuubi start -> Fails with port already occupied

bin/kyuubi run -> Launches kyuubi server
bin/kyuubi status -> Kyuubi is running (PID)
bin/kyuubi start -> Kyuubi running as process PID. Stop it first.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/bin/kyuubi
+++ b/bin/kyuubi
@@ -117,6 +117,14 @@ cmd="${RUNNER} ${KYUUBI_JAVA_OPTS} -cp ${KYUUBI_CLASSPATH} $CLASS $@"
 
 pid="${KYUUBI_PID_DIR}/kyuubi-$USER-$CLASS.pid"
 
+function get_pid(){
+  run_pid="$(ps -ef | grep org.apache.kyuubi.server.KyuubiServer |
+  grep java | grep "$KYUUBI_CONF_DIR" | grep -v grep | awk '{print $2}')"
+  if [[ -n $run_pid ]]; then
+    echo "$run_pid" > "$pid"
+  fi
+}
+
 function start_kyuubi() {
   if [[ ! -w ${KYUUBI_PID_DIR} ]]; then
     echo "${USER} does not have 'w' permission to ${KYUUBI_PID_DIR}"
@@ -127,6 +135,8 @@ function start_kyuubi() {
     echo "${USER} does not have 'w' permission to ${KYUUBI_LOG_DIR}"
     exit 1
   fi
+
+  get_pid
 
   if [ -f "$pid" ]; then
     TARGET_ID="$(cat "$pid")"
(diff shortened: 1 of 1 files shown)


bin/kyuubi
```diff
@@ -106,7 +106,20 @@ cmd="${RUNNER} ${KYUUBI_JAVA_OPTS} -cp ${KYUUBI_CLASSPATH} $CLASS"
 
 pid="${KYUUBI_PID_DIR}/kyuubi-$USER-$CLASS.pid"
 
+function get_pid(){
+  if [[ ! -w ${KYUUBI_PID_DIR} ]]; then
+    echo "${USER} does not have 'w' permission to ${KYUUBI_PID_DIR}"
+    exit 1
+  fi
+  
+  run_pid="$(ps -ef | grep kyuubi | grep java | grep -v grep | awk '{print $2}')"
```
zhouyifan279, line 115: 1. `grep kyuubi` is not sufficient to ensure that this is a Kyuubi Server process. For example, Kyuubi Engine processes also have `kyuubi` in their info.
2. User may have multiple Kyuubi installations on a single machine. We need to ensure that this Kyuubi Server process is launched using current Kyuubi installation.

Madhukar98: Thanks @zhouyifan279 for your feedback. I will work on these.

Madhukar525722: Hi @zhouyifan279 , I have updated the implementation, now I am making the selection on
1. org.apache.kyuubi.server.KyuubiServer - Ensures its an server process
2. $KYUUBI_CONF_DIR - Kyuubi conf dir path, displayed in the complete process details. This identifier is choosen to differentiate between the instances, because different instances cant run with same conf, as it will cause port already in use error.

Please review the changes.

Thanks

PR #6399, opened 2024-05-20
Spark Kyuubi UI supports both javax and jakarta servlet namespaces

pan3793:
# :mag: Description

Spark 4.0 migrated from `javax.servlet` to `jakarta.servlet` in SPARK-47118, which breaks the binary compatibility of `SparkUITab` and `WebUIPage` that Kyuubi used, thus breaking the previous assumption of Kyuubi Spark SQL engine: single jar built with default Spark version, compatible with all supported versions of Spark runtime.

## Describe Your Solution 🔧

This PR uses bytebuddy to dynamically generate classes and Java reflection find and dispatch method invocation in runtime, to recover the existing compatibility of Kyuubi Spark SQL engine.

## Test Plan 🧪

Build with Spark 3.5
```
build/dist --tgz --web-ui --spark-provided --flink-provided --hive-provided -Pspark-3.5
```

It produces both Scala 2.12 and 2.13 Spark SQL engine jars
- `kyuubi-spark-sql-engine_2.12-1.10.0-SNAPSHOT.jar`
- `kyuubi-spark-sql-engine_2.13-1.10.0-SNAPSHOT.jar`

Run with Spark 3.4 Scala 2.12

image

Run with Spark 3.5 Scala 2.13

image

Run with Spark 4.0.0-preview1 Scala 2.13

image

---

# Checklist 📝

**Be nice. Be informative.**


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/ui/JakartaHttpServletRequest.scala
```diff
@@ -0,0 +1,173 @@
...
+
+import java.io.BufferedReader
+import java.security.Principal
+import java.util
+import java.util.Locale
+
+import jakarta.servlet._
+import jakarta.servlet.http._
+
+class JakartaHttpServletRequest(req: HttpServletRequest)
```
pan3793, line 28: Unfortunately, neither Java nor Scala support delegate syntax, I originally wanted to implement it through JDK Dynamic Proxy, but met classloader issues. Eventually, I chose such a naive approach.

pan3793:
@link3280 this approach is similar to that we used to make Flink support both 1.16 and 1.17

pan3793:
Thanks, merged to master/1.9

yikf:
late lgtm due to busy with other things, thanks @pan3793 for the nice work!

PR #6400, opened 2024-05-21
Fix memory leak when using saveToFile

turboFei:
# :mag: Description
## Issue References 🔗

Fix memory leak when using saveToFile mode.

FYI: https://stackoverflow.com/questions/45649044/scala-stream-iterate-and-memory-management

Stream is IterableAgain, which means, that it will keep all the elements you iterate through in case you want to see them again.

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


turboFei:
Similar with https://github.com/apache/kyuubi/pull/3885

turboFei:
cc @wForget

wForget (approved):

turboFei:
thanks, merged to 1.9.1

turboFei:
also cc @pan3793 It is the second time we meet this issue. we can enhance it later.

yaooqinn:
Late LGTM

PR #6416, opened 2024-05-24
Generate flattened POM

pan3793:
# :mag: Description

Leverage [Flatten Maven Plugin](https://www.mojohaus.org/flatten-maven-plugin/) to generate resolved `pom.xml`.

This recovers a dev case for Scala 2.13.

We usually perform `clean install` with `-am` and `-DskipTests` one time and perform `test` after changing code on specific modules, e.g.
```
build/mvn -Pscala-2.13 -pl externals/kyuubi-spark-sql-engine clean install -am -DskipTests
build/mvn -Pscala-2.13 -pl externals/kyuubi-spark-sql-engine test
```

Currently, the second step wrongly resolves to `*_2.12` dependencies due to the installed `pom.xml` referring to `*_${scala.binary.version}` while the default value of `scala.binary.version` is `2.12`.

## Test Plan 🧪

This change is verified locally by the above case.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ spark-warehouse/
 metastore_db
 derby.log
 rest-audit.log
+**/.flattened-pom.xml
 **/dependency-reduced-pom.xml
 metrics/
 /kyuubi-ha/embedded_zookeeper/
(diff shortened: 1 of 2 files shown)


bowenliang123 (approved):

pan3793:
Thanks, merged to master/1.9

PR #6456, opened 2024-06-06
Opens jdk.internal.ref

pan3793:
# :mag: Description

This PR follows SPARK-45508 to modify the JVM args by adding `--add-opens=java.base/jdk.internal.ref=ALL-UNNAMED`, without this arg should not affect Spark engines as `spark-submit` will handle that by launching a sub-process and adding those JVM args.

The change benefits Spark in-process cases in UT, a little.

This PR also renames maven properties `maven.plugin.surefire.argLine` back to `extraJavaTestArgs` as it is used in both surefire and scalatest plugins, and makes it always takes effect instead of only JDK 17, then we don't need a copy when adding a new profile `java-21`

## Test Plan 🧪

Pass GHA.

---

# Checklist 📝

**Be nice. Be informative.**


cxzl25 (approved):

pan3793:
Thanks, merged to master

PR #6458, opened 2024-06-07
Remove commons-logging from binary release

pan3793:
# :mag: Description

[`jcl-over-slf4j`](https://www.slf4j.org/legacy.html#jcl-over-slf4j) is a drop-in replacement of `commons-logging`, the latter one should not be present in the final classpath, otherwise, there are potential class conflict issues.

The current dep check is problematic, this PR also changes it to always perform "install" to fix the false negative report.

## Test Plan 🧪

Simply delete `commons-logging-1.1.3.jar` from `apache-kyuubi-1.9.1-bin.tgz` and everything goes well.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/.github/workflows/dep.yml
+++ b/.github/workflows/dep.yml
@@ -48,27 +48,15 @@ jobs:
           check-latest: false
       - name: Setup Maven
         uses: ./.github/actions/setup-maven
-      - name: Check kyuubi modules available
-        id: modules-check
-        run: >-
-          build/mvn dependency:resolve validate
-          -DincludeGroupIds="org.apache.kyuubi" -DincludeScope="compile"
-          -Pfast -Denforcer.skip=false
-          -pl kyuubi-ctl,kyuubi-server,kyuubi-assembly -am
-        continue-on-error: true
-      - name: build
+      - name: Install modules
         env:
-          MAVEN_OPTS: -Dorg.slf4j.simpleLogger.defaultLogLevel=error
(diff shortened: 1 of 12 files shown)


.github/workflows/dep.yml
```diff
@@ -48,27 +48,15 @@ jobs:
...
-        run: >-
-          build/mvn clean install
-          -Pflink-provided,spark-provided,hive-provided
-          -Dmaven.javadoc.skip=true
-          -Drat.skip=true
-          -Dscalastyle.skip=true
-          -Dspotless.check.skip
-          -DskipTests
-          -pl kyuubi-ctl,kyuubi-server,kyuubi-assembly -am
+          MAVEN_ARGS: -Dorg.slf4j.simpleLogger.defaultLogLevel=error
```
pan3793, line 53: I suppose it should be MAVEN_ARGS, though it does not work before Maven 3.9
https://maven.apache.org/configure.html#maven_args-environment-variable

.github/workflows/dep.yml
```diff
@@ -48,27 +48,15 @@ jobs:
...
+        run: |
+          build/mvn clean install \
```
pan3793, line 55: use `|` to make the command "copy-paste and run" easily

cxzl25 (approved):

pan3793:
Thanks, merged to master/1.9.2

PR #6463, opened 2024-06-13
Release semaphore immediately after startup process exit

ic4y:
# :mag: Description
## Issue References 🔗

The concurrency limit for the engine startup process is mainly used to avoid overload on the machine(or container) of the Kyuubi server, the current implementation holds startupProcessSemaphore until the session is established successfully. While for Spark on YARN cluster mode, some YARN queue resource insufficiency may block the subsequent Spark application submissions to other queues, significantly affecting the Kyuubi server's resource utilization.

## Describe Your Solution 🔧

We should immediately release the `startupProcessSemaphore` after the engine startup process exits (i.e., after the `spark-submit` process exits) as the load has already disappeared.

## Test Plan 🧪

I tested it on a cluster of 50 kyuubi Servers, and kyuubi server resource utilization increased by 70%

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
@@ -239,16 +239,16 @@ private[kyuubi] class EngineRef(
       while (engineRef.isEmpty) {
         if (exitValue.isEmpty && process.waitFor(1, TimeUnit.SECONDS)) {
           exitValue = Some(process.exitValue())
-          if (!exitValue.contains(0)) {
+          if (exitValue.contains(0)) {
+            acquiredPermit = false
+            startupProcessSemaphore.foreach(_.release())
+          } else {
             val error = builder.getError
             MetricsSystem.tracing { ms =>
               ms.incCount(MetricRegistry.name(ENGINE_FAIL, appUser))
               ms.incCount(MetricRegistry.name(ENGINE_FAIL, error.getClass.getSimpleName))
             }
             throw error
-          } else {
-            acquiredPermit = false
-            startupProcessSemaphore.foreach(_.release())
           }
         }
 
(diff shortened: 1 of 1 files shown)


pan3793:
The idea makes sense, and a similar mechanism was applied to batch submission too https://github.com/apache/kyuubi/pull/6028

cxzl25 (commented):
LGTM

cxzl25 (approved):

wForget (approved):
Thanks, LGTM

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -246,6 +246,9 @@ private[kyuubi] class EngineRef(
               ms.incCount(MetricRegistry.name(ENGINE_FAIL, error.getClass.getSimpleName))
             }
             throw error
+          } else {
+            acquiredPermit = false
+            startupProcessSemaphore.foreach(_.release())
```
pan3793, line 251: not a big deal, let's invert the `if else` to avoid anti-condition to gain clearer code

```
if (exitValue.contains(0)) {
  startupProcessSemaphore.foreach(_.release())
  acquiredPermit = false
} else {
  ...
}
```

ic4y: I got it

pan3793:

ic4y:
> @ic4y you may want to link the email `Authored-by: ic4y <ic4y@apache.org>` to your GitHub account, otherwise your contributions won't be counted by GitHub
> 
> You can check [95ed748](https://github.com/apache/kyuubi/commit/95ed74821c3a2d3a3f402033de7b463966a4bc28) after updating the GitHub profile

I have added my email, thank you for reminding me.

Issue #6469, opened 2024-06-15
[Bug] Memory Overhead of OrcMapreduceRecordReader Initialization results in driver oom when fetch big result using FetchOrcStatement


Z1Wu:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

# Problem 

When fetching big result using feature introduced by https://github.com/apache/kyuubi/issues/5377 ，driver will be oom if there are too many `RecordReaderIterator[OrcStruct]` being instantiated in memory.

In current implementation, the number of `RecordReaderIterator[OrcStruct]` will be determined  by the number orc files written by insert command which is hard to control when run with AQE or other configuration.

https://github.com/apache/kyuubi/blob/95ed74821c3a2d3a3f402033de7b463966a4bc28/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala#L77-L79

Simple script to reproduce the problem:

``` sh

# connect to kyuubi server and enable fetch file using FetchOrcStatement
beeline -u 'jdbc:hive2://<kyuubi-server>:<port>/?spark.executor.instances=40;kyuubi.operation.result.saveToFile.enabled=true;kyuubi.operation.result.saveToFile.minSize=1;spark.submit.deployMode=client;spark.driver.memory=5g;spark.sql.catalog.tpcds=org.apache.kyuubi.spark.connector.tpcds.TPCDSCatalog;' -e 'select * from tpcds.sf3000.catalog_returns;'

```

Error Stack :

```
Error: org.apache.kyuubi.KyuubiSQLException: org.apache.kyuubi.KyuubiSQLException: Error operating ExecuteStatement: java.lang.OutOfMemoryError: Java heap space
	at org.apache.orc.impl.RecordReaderUtils.readRanges(RecordReaderUtils.java:423)
	at org.apache.orc.impl.RecordReaderUtils.readDiskRanges(RecordReaderUtils.java:491)
	at org.apache.orc.impl.RecordReaderUtils$DefaultDataReader.readFileData(RecordReaderUtils.java:103)
	at org.apache.orc.impl.reader.StripePlanner.readData(StripePlanner.java:152)
[...]

PR #6470, opened 2024-06-15
[KYUUBI #6469] Lazily initialize RecordReaderIterator to avoid driver oom when fetching big result set

Z1Wu:
# :mag: Description
## Issue References 🔗

This pull request fixes https://github.com/apache/kyuubi/issues/6469

## Describe Your Solution 🔧

Instead of initializing all RecordReaderIterator when create OrcFileIterator，we can lazily initialize the RecordReaderIterator to make sure that there is only one RecordReaderIterator which reads file current fetching by client in driver memory.

#### Related Unit Tests

---

# Checklist 📝

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
@@ -35,7 +35,6 @@ import org.apache.spark.sql.execution.datasources.RecordReaderIterator
 import org.apache.spark.sql.execution.datasources.orc.OrcDeserializer
 import org.apache.spark.sql.types.StructType
 
-import org.apache.kyuubi.Logging
 import org.apache.kyuubi.operation.{FetchIterator, IterableFetchIterator}
 
 class FetchOrcStatement(spark: SparkSession) {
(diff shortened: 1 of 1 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -74,37 +75,44 @@ class FetchOrcStatement(spark: SparkSession) {
   }
 }
 
-class OrcFileIterator(fileList: ListBuffer[LocatedFileStatus]) extends Iterator[OrcStruct] {
-
-  private val iters = fileList.map(x => getOrcFileIterator(x))
+class OrcFileIterator(fileList: ListBuffer[LocatedFileStatus]) extends Iterator[OrcStruct]
+  with Logging {
```
cxzl25, line 79: It looks like `Logging` is not used.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
```diff
@@ -74,37 +75,44 @@ class FetchOrcStatement(spark: SparkSession) {
...
 
   var idx = 0
```
cxzl25, line 81: nit private

cxzl25 (commented):
LGTM overall. I forgot to comment on this potential problem internally or on github. cc @lsm1

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
Z1Wu: removed

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/FetchOrcStatement.scala
Z1Wu: fixed

pan3793 (approved):

pan3793:
Thanks, merged to master/1.9

PR #6473, opened 2024-06-16
[KYUUBI #6302] Call cancelJobGroup immediately after statement execution finished

### LINKED ISSUE #6302: [Bug] Kyuubi return query status (FINISH_STATE) quickly, but job still executing.
### Code of Conduct

### Search before asking

### Describe the bug

When I submit query by kyuubi to Spark, kyuubi return query status (FINISH_STATE) quickly, but spark job still executing.
image

image
image

Then I find similar bug description in issue. @XorSum https://github.com/apache/kyuubi/issues/5232, in this issue, said this fix has been merged into 1.8.0, but I try both 1.8.0, 1.8.1, 1.9.0, this problem will always arise, Is there any configuration that needs to be modified？

I find my SQL is also `(empty table) left join (big table)`.

my SQL is as follow:
`select
  *
from
  (
      select
      uid
    from
      db.table1
    where
      d_s = 20240408
      and abtest_id like 'sadjjfljlajf'
  ) t1
  left join (
    select
          vendor_id
    FROM
      db.table2
    LATERAL VIEW explode (json_object_keys (countdata)) t1 AS kv 
    WHERE
      d_s = 20240408
  ) t2 on t2.vendor_id = t1.uid
limit
  5;`

in this SQL, the left sub query result is empty (0 rows).
`select
      count(1)
    from
      db.table1
    where
      d_s = 20240408
      and abtest_id like 'sadjjfljlajf'`
image

right sub-query result is relative big (4432411935 rows).
`select
          count(1)
    FROM
      db.table2
    LATERAL VIEW explode (json_object_keys (countdata)) t1 AS kv 
    WHERE
      d_s = 20240408`
image

### Affects Version(s)

kyuubi 1.8.0 / 1.8.1/1.9.0    | Spark 3.4.2

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

XorSum:
# :mag: Description
## Issue References 🔗

This pull request fixes #6302 : when the SQL is submitted by Kyuubi Rest Api (the same kyuubi path with BEELINE), the legacy job will not be cancelled, https://github.com/apache/kyuubi/issues/6302#issuecomment-2160572624 .

The reason:
    Beeline session calls `cancelJobGroup` in `SparkOperation#cleanup`. 
    But  Rest Api session doesn't call `SparkOperation#cleanup`, so the legacy job submitted by Rest Api will not be canceled.

## Describe Your Solution 🔧

The modification:
    call `SparkOperation#cleanup` in `ExecuteStatement#executeStatement`'s finally clause.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
@@ -104,6 +104,7 @@ class ExecuteStatement(
       onError(cancel = true)
     } finally {
       shutdownTimeoutMonitor()
+      if (!spark.sparkContext.isStopped) spark.sparkContext.cancelJobGroup(statementId)
     }
 
   override protected def runInternal(): Unit = {


cxzl25 (approved):

pan3793 (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
```diff
@@ -104,6 +104,7 @@ class ExecuteStatement(
       onError(cancel = true)
     } finally {
       shutdownTimeoutMonitor()
+      if (!spark.sparkContext.isStopped) spark.sparkContext.cancelJobGroup(statementId)
```
wForget, line 107: Will this cause tasks using incremental collect mode to fail immediately?

pan3793:

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala
pan3793: thought, this should be fine in most cases, but still a corner case that may fail incremental collect queries.

when incremental collect is enabled,

1. trigger the job to calculate the first partition of the final stage result RDD
2. when the above job is finished, change the operation state from RUNNING to FINISHED
3.1 can job group
3.2 the client starts to consume the first partition result, after consuming all records, triggers a new job to calculate the second partition of result RDD

3.1 and 3.2 could happen in parallel, but in most cases, 3.1 completes quickly while 3.2 takes a few seconds, so the whole process is fine. while there is a chance that the result of the first partition is small and the client consumes the result quickly, and the scheduler is under high pressure that jobGroup cancellation takes a few seconds, then such a code could fail the query.

pan3793: so it would be better to

- skip the cancellation here for incremental collect queries. or
- add an internal flag to control whether to skip the cancellation here for incremental collect queries, skipping by default for safety.

PR #6490, opened 2024-06-20
[KYUUBI #6489] [PYTHON] PyKyuubi get_table_names also supports Spark SQL dialect

### LINKED ISSUE #6489: [Bug] function 'get_table_names' in the python client returns incorrect results
### Code of Conduct

### Search before asking

### Describe the bug

image
image

Hi,
I encountered the problem in the above image when I tried to connect Kyuubi with Superset.
The table structure and table name are incorrect, and appear to return a schema.

It's not what I want.

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

```yaml
default
```

_No response_

### Additional context

_No response_

BruceWong96:
# :mag: Description
## Issue References 🔗

This pull request fixes #6489

## Describe Your Solution 🔧

After my investigation, I found the bug and solution.
The function get_table_names returns an incorrect value when I used Superset to connect to Kyuubi for Spark SQL.
[get_table_names](https://github.com/apache/kyuubi/blob/master/python/pyhive/sqlalchemy_hive.py#L380)

The following code is used to connect to hive directly.
`return [row[0] for row in connection.execute(text(query))]` 

Because The following value is returned when the Hive is connected.

show tables in default :
[('student',), ('student_scores',)]

The following code is used to connect to Kyuubi.
`return [row[1] for row in connection.execute(text(query))]` 

Because The following value is returned when the Kyuubi is connected.

show tables in default :
[('default', 'employees', False), ('default', 'student', False), ('default', 'student_scores', False)]

So, for the difference in return value, I modified the code.

And I test them in Superset. The code works.

Hive
image

Kyuubi
image

Spark SQL also works properly.
image

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/python/pyhive/sqlalchemy_hive.py
+++ b/python/pyhive/sqlalchemy_hive.py
@@ -377,7 +377,17 @@ class HiveDialect(default.DefaultDialect):
         query = 'SHOW TABLES'
         if schema:
             query += ' IN ' + self.identifier_preparer.quote_identifier(schema)
-        return [row[1] for row in connection.execute(text(query))]
+
+        table_names = []
+
+        for row in connection.execute(text(query)):
+            # Hive returns 1 columns
+            if len(row) == 1:
+                table_names.append(row[0])
+            # Kyuubi returns 3 columns
+            elif len(row) == 3:
+                table_names.append(row[1])
+        return table_names
 
     def do_rollback(self, dbapi_connection):
         # No transactions for Hive
--- a/python/pyhive/sqlalchemy_hive.py
+++ b/python/pyhive/sqlalchemy_hive.py
@@ -10,6 +10,7 @@ from __future__ import unicode_literals
 
 import datetime
 import decimal
+import logging
 
 import re
 from sqlalchemy import exc
@@ -39,6 +40,7 @@ from pyhive.common import UniversalSet
 from dateutil.parser import parse
 from decimal import Decimal
 
+_logger = logging.getLogger(__name__)
 
 class HiveStringTypeBase(types.TypeDecorator):
     """Translates strings returned by Thrift into something else"""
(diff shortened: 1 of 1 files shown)


python/pyhive/sqlalchemy_hive.py
```diff
@@ -377,7 +377,7 @@ def get_table_names(self, connection, schema=None, **kw):
         query = 'SHOW TABLES'
         if schema:
             query += ' IN ' + self.identifier_preparer.quote_identifier(schema)
-        return [row[0] for row in connection.execute(text(query))]
+        return [row[1] for row in connection.execute(text(query))]
```
pan3793, line 380: this breaks the Hive case.

The PyKyuubi is supposed to work with Kyuubi, Spark Thrift Server, and HiveServer2. In this case, we may need to either switch to the Metadata Thrift API or create different dialects

BruceWong96: > this breaks the Hive case.
> 
> The PyKyuubi is supposed to work with Kyuubi, Spark Thrift Server, and HiveServer2. In this case, we may need to either switch to the Metadata Thrift API or create different dialects

Thank you for your reply. 
I have made modifications and tests. 
The specific details are in the issue.

python/pyhive/sqlalchemy_hive.py
```diff
@@ -377,7 +377,17 @@ def get_table_names(self, connection, schema=None, **kw):
...
             query += ' IN ' + self.identifier_preparer.quote_identifier(schema)
-        return [row[1] for row in connection.execute(text(query))]
+
+        table_names = []
+
+        for row in connection.execute(text(query)):
+            # Hive returns 1 columns
+            if len(row) == 1:
+                table_names.append(row[0])
+            # Kyuubi returns 3 columns
```
pan3793, line 387: actually, for SQL cases, Kyuubi's behavior totally depends on the computing engine
```suggestion
            # Spark SQL returns 3 columns
```

python/pyhive/sqlalchemy_hive.py
```diff
@@ -377,7 +377,17 @@ def get_table_names(self, connection, schema=None, **kw):
...
+            elif len(row) == 3:
+                table_names.append(row[1])
```
pan3793, line 389: can we have an else branch to handle the other cases?

BruceWong96: > can we have an else branch to handle the other cases?

So is it better for the else branch to throw an exception or return UNKNOWN as a tablename ?

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #6492, opened 2024-06-20
[K8S][HELM] Align Helm templates with formatting approach

dnskr:
# :mag: Description
## Issue References 🔗
There are few formatting misalignments in the Helm chart.

## Describe Your Solution 🔧
The PR aligns the chart templates formatting without any functional changes.

#### Test 1
All templates (and rendered `labels`) are correct with default chart values:
```sh
helm template --debug charts/kyuubi
```

#### Test 2
Templates `kyuubi-alert.yaml`, `kyuubi-podmonitor.yaml` and `kyuubi-servicemonitor.yaml` are correct with additionally provided values:
```sh
helm template --debug charts/kyuubi -f values-monitoring.yaml -s templates/kyuubi-alert.yaml -s templates/kyuubi-podmonitor.yaml -s templates/kyuubi-servicemonitor.yaml
```
File `values-monitoring.yaml` used for testing
```yaml
metricsReporters: PROMETHEUS

podMonitor:
  enabled: true
  podMetricsEndpoint:
    - port: web

serviceMonitor:
  enabled: true
  endpoints:
    - port: web
  labels:
    first: label1
    second: label2

prometheusRule:
  enabled: true
  groups:
    - name: ./example.rules
      rules:
        - alert: ExampleAlert
          expr: vector(1)

```

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/charts/kyuubi/templates/kyuubi-alert.yaml
+++ b/charts/kyuubi/templates/kyuubi-alert.yaml
@@ -20,9 +20,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
   name: {{ .Release.Name }}
-  labels:
-    {{- include "kyuubi.labels" . | nindent 4 }}
+  labels: {{- include "kyuubi.labels" . | nindent 4 }}
 spec:
-  groups:
-    {{- toYaml .Values.prometheusRule.groups | nindent 4 }}
+  groups: {{- toYaml .Values.prometheusRule.groups | nindent 4 }}
 {{- end }}
--- a/charts/kyuubi/templates/kyuubi-configmap.yaml
+++ b/charts/kyuubi/templates/kyuubi-configmap.yaml
@@ -19,8 +19,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ .Release.Name }}
-  labels:
-    {{- include "kyuubi.labels" . | nindent 4 }}
+  labels: {{- include "kyuubi.labels" . | nindent 4 }}
 data:
   {{- with .Values.kyuubiConf.kyuubiEnv }}
   kyuubi-env.sh: |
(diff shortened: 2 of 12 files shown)


charts/kyuubi/templates/kyuubi-alert.yaml
```diff
@@ -20,9 +20,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
   name: {{ .Release.Name }}
-  labels:
-    {{- include "kyuubi.labels" . | nindent 4 }}
+  labels: {{- include "kyuubi.labels" . | nindent 4 }}
```
pan3793, line 23: well, but the previous one is more intuitive for me, since the rendered result of template
```
  labels:
    {{- include "kyuubi.labels" . | nindent 4 }}
```
is something like
```
  labels:
    label1: value1
    label2: value2
```

any specify reason to use the proposed one?

dnskr: There is no specific reason, rather than following one style across all templates. The idea is to keep templates more compact if it is possible. We use this formatting style in many places, for instance: https://github.com/apache/kyuubi/blob/0a53415d929dfe477f9c31f707e559c09dd175a0/charts/kyuubi/templates/kyuubi-role.yaml#L25 https://github.com/apache/kyuubi/blob/0a53415d929dfe477f9c31f707e559c09dd175a0/charts/kyuubi/templates/kyuubi-service.yaml#L27 https://github.com/apache/kyuubi/blob/0a53415d929dfe477f9c31f707e559c09dd175a0/charts/kyuubi/templates/kyuubi-service.yaml#L44

pan3793: I agree to unify the style, but lean to invert the style.

dnskr: There is no strong convention at the moment as far as I know, so both approaches are valid. Personally I prefer more compact style, because from my experience it's more convenient to read and review the code, especially if a template has many lines. One argument here might be that Bitnami uses this approach as well for all their [charts](https://github.com/bitnami/charts/tree/main/bitnami).

Issue #6464, opened 2024-06-13
[Bug] Kyuubi Server register HA&ZK  serverUri error when serverHost bind 0.0.0.0


zcx513566:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When serverHost bind 0.0.0.0 to listen all NIC(ip double stack ipv4 / ipv6) request,  Kyuubi Server register HA&ZK  serverUri  is 0.0.0.0:10009

### Affects Version(s)

1.9.0 and above

### Kyuubi Server Configurations

```yaml
kyuubi.frontend.bind.host=0.0.0.0
or
kyuubi.frontend.thrift.binary.bind.host=0.0.0.0
```

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #6495, opened 2024-06-21
[KYUUBI #6464] Kyuubi Server register HA&ZK serverUri and insert db i…

zcx513566:
…n metadata table error when serverHost bind 0.0.0.0

# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

in connectionUrl function use Utils.findLocalInetAddress.getHostAddress or Utils.findLocalInetAddress.getCanonicalHostName  get host adress if bind host 0.0.0.0

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -114,6 +114,12 @@ abstract class TFrontendService(name: String)
     checkInitialized()
     val host = (conf.get(FRONTEND_ADVERTISED_HOST), serverHost) match {
       case (Some(advertisedHost), _) => advertisedHost
+      case (None, Some(h)) if (h.equals("0.0.0.0")) =>
```
pan3793, line 117: Let's define a utility function in `JavaUtils` instead of writing `h.equals("0.0.0.0")` many times. Meanwhile, IPv6 might be considered in the future.

According to [Wikipedia](https://en.wikipedia.org/wiki/0.0.0.0), the official name of `0.0.0.0` is "This host on this network", defined at [RFC 1122](https://datatracker.ietf.org/doc/html/rfc1122) section 3.2.1.3.

In Linux, [INADDR_ANY](https://www.man7.org/linux/man-pages/man7/ip.7.html) and [IN6ADDR_ANY_INIT](https://www.man7.org/linux/man-pages/man7/ipv6.7.html) are used to represent IPv4 and IPv6 "This host on this network". So we can name the function "isAnyInetAddress"

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -114,6 +114,12 @@ abstract class TFrontendService(name: String)
...
+        if (conf.get(FRONTEND_CONNECTION_URL_USE_HOSTNAME)) {
+          Utils.findLocalInetAddress.getCanonicalHostName
```
pan3793, line 119: this function is moved to `JavaUtils` in https://github.com/apache/kyuubi/commit/315adda3539bcfab2bb2b9cb2ac6882f8444721c, could you please rebase your branch to latest master?

PR #6526, opened 2024-07-04
Kyuubi BeeLine wrongly process JDBC URL that contains `--`

zcx513566:
# :mag: Description
## Issue References 🔗

This pull request fixes https://github.com/apache/kyuubi/issues/6522

## Describe Your Solution 🔧

Remove comments in dispatch commands only execute command

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiBeeLine.java
+++ b/kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiBeeLine.java
@@ -218,7 +218,7 @@ public class KyuubiBeeLine extends BeeLine {
     if (!commands.isEmpty()) {
       for (String command : commands) {
         debug(loc("executing-command", command));
-        if (!dispatch(command)) {
+        if (!removeCommentsDispatch(command)) {
           code++;
         }
       }
@@ -282,9 +282,8 @@ public class KyuubiBeeLine extends BeeLine {
     return executionResult;
   }
 
-  // see HIVE-15820: comment at the head of beeline -e
-  @Override
-  boolean dispatch(String line) {
+  // see HIVE-15820: comment at the head of beeline -e only dispatch commands
+  boolean removeCommentsDispatch(String line) {
     return super.dispatch(isPythonMode() ? line : HiveStringUtils.removeComments(line));
   }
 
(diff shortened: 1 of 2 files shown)


pan3793:
cc @cfmcgrady @turboFei

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiBeeLine.java
```diff
@@ -282,9 +282,8 @@ int runInit() {
     return executionResult;
   }
 
-  // see HIVE-15820: comment at the head of beeline -e
-  @Override
-  boolean dispatch(String line) {
+  // see HIVE-15820: comment at the head of beeline -e only dispatch commands
+  boolean removeCommentsDispatch(String line) {
```
turboFei, line 286: what isthe difference

pan3793: The existing implementation also affects invocations in BeeLine
image

turboFei (approved):

turboFei:
could you add UT?

kyuubi-hive-beeline/src/main/java/org/apache/hive/beeline/KyuubiBeeLine.java
zcx513566: > what isthe difference
Remove comments in dispatch commands only execute command. 
for example：beeline -e '--   xxx' or -f script.sql remove comments，but not do anything for other case.

zcx513566:
> could you add UT?
ut file testKyuubiBeelineComment covered remove comments in beeline -e '-- comment' and add ut for url contain '--' case.

pan3793 (approved):

pan3793:
Merge to master/1.9

pan3793:
@zcx513566 with an additional look at this issue and the [upstream change](https://github.com/apache/hive/commit/425e1ff7c054f87c4db87e77d004282d529599ae), I think the fix is invalid. And the UT you added is invalid too, it also passes without this patch.

zcx513566:
> @zcx513566 with an additional look at this issue and the [upstream change](https://github.com/apache/hive/commit/425e1ff7c054f87c4db87e77d004282d529599ae), I think the fix is invalid. And the UT you added is invalid too, it also passes without this patch.
First the url should not remove comments. it also passes without this patch is  because the UT only check sql content，not verified url.

Issue #6523, opened 2024-07-03
[Improvement] Provide Kyuubi configuration properties with cli options


dnskr:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

I propose to add support for `--conf` options to Kyuubi similar to Spark and kyuubi-beeline. It would be useful for users and operational experience to be able to provide configuration properties with cli options in addition to config files.

### How should we improve?

Configuration provided via cli option should have higher priority over config files. For instance, the following command should launch Kyuubi with `kyuubi.frontend.rest.bind.port` equal to `9999`:
```shell
kyuubi run --conf “kyuubi.frontend.rest.bind.port=9999”
```

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #6530, opened 2024-07-10
[KYUUBI #6523] Support configuration with `--conf` options

dnskr:
# :mag: Description
## Issue References 🔗

This pull request fixes https://github.com/apache/kyuubi/issues/6523

## Describe Your Solution 🔧

`KyuubiServer` respects configuration properties provided as `--conf` cli options with highest priority.

REST endpoint starts at `10099` port:
```shell
bin/kyuubi run --conf kyuubi.frontend.rest.bind.port=9999
...
2024-07-10 17:57:24.998 INFO main org.apache.kyuubi.server.KyuubiRestFrontendService: Service[KyuubiRestFrontendService] is started.
2024-07-10 17:57:24.999 INFO main org.apache.kyuubi.server.KyuubiRestFrontendService: Exposing REST endpoint at: http://0.0.0.0:10099
2024-07-10 17:57:24.999 INFO main org.apache.kyuubi.server.KyuubiServer: Service[KyuubiServer] is started.
```

REST endpoint starts at `9999` port:
```shell
bin/kyuubi run --conf kyuubi.frontend.rest.bind.port=9999
...
2024-07-10 17:49:41.659 INFO main org.apache.kyuubi.server.KyuubiRestFrontendService: Service[KyuubiRestFrontendService] is started.
2024-07-10 17:49:41.660 INFO main org.apache.kyuubi.server.KyuubiRestFrontendService: Exposing REST endpoint at: http://0.0.0.0:9999
2024-07-10 17:49:41.660 INFO main org.apache.kyuubi.server.KyuubiServer: Service[KyuubiServer] is started.
```

#### Related Unit Tests
`kyuubi-common/src/test/scala/org/apache/kyuubi/config/KyuubiConfSuite.scala`

---

# Checklist 📝

**Be nice. Be informative.**


dnskr:
I'm not familiar with the source code well enough yet, so any feedback, suggestions or even an alternative PR are highly appreciated.

bin/kyuubi
```diff
@@ -102,7 +102,7 @@ if [[ -n ${YARN_CONF_DIR} ]]; then
   KYUUBI_CLASSPATH="${KYUUBI_CLASSPATH}:${YARN_CONF_DIR}"
 fi
 
-cmd="${RUNNER} ${KYUUBI_JAVA_OPTS} -cp ${KYUUBI_CLASSPATH} $CLASS"
+cmd="${RUNNER} ${KYUUBI_JAVA_OPTS} -cp ${KYUUBI_CLASSPATH} $CLASS ${@:2}"
```
pan3793, line 105: it also allows `bin/kyuubi` without command. I think we can shift the consumed args and pass `$@` here

pan3793:
the rest change lgtm, thanks for working on it.

bin/kyuubi
dnskr: @pan3793 Thanks, I missed that case.

Why do we need to support running Kyuubi as a daemon with `bin/kyuubi` without `start` command? At first look, it would be more expected to me to print help message if no command provided by user. I can create a PR if you think it is a good change.

Also I found another interesting case. As a user, I would expect `restart` to launch Kyuubi with the same configs was used to launch Kyuubi with `start` command initially. Not sure how to manage it automatically, so in current implementation `restart` command requires users to provide the same `--conf` options was used with `start` command.

pan3793: > Why do we need to support running Kyuubi as a daemon with `bin/kyuubi` without `start` command?

this is a behavior from the very beginning, let's keep it to avoid breaking the user experience

> ... so in current implementation `restart` command requires users to provide the same `--conf` options was used with `start` command.

I see, and it is reasonable to me.

dnskr: @pan3793 I added commit to address your initial suggestion. Please, have a look if it looks good.

---
> > Why do we need to support running Kyuubi as a daemon with `bin/kyuubi` without `start` command?
> 
> this is a behavior from the very beginning, let's keep it to avoid breaking the user experience
> 

Right. It is make sense to leave it as is, at least before Kyuubi 2.0

pan3793 (approved):
Okay, I think it should work properly now.

pan3793:
also cc @cxzl25 @cfmcgrady @hddong

cxzl25 (approved):

pan3793:
Thanks, merged to master

PR #6542, opened 2024-07-17
KyuubiBatchService should wait for HTTP server started before picking jobs

pan3793:
# :mag: Description

This is similar to https://github.com/apache/kyuubi/pull/5310.

We need to wait for Jetty Server started before picking jobs, otherwise, it may get the wrong local HTTP server port -1.

This only affects Batch V2

## Test Plan 🧪

Tested in internal workloads.

---

# Checklist 📝

**Be nice. Be informative.**


pan3793:
cc @zwangsheng

cxzl25 (approved):

pan3793:

PR #6561, opened 2024-07-24
[KYUUBI #6560] Support removing user-specified repartition before writing when using zorder

### LINKED ISSUE #6560: [Improvement] Support removing user-specified repartition before writing when using zorder
### Code of Conduct

### Search before asking

In our company, users will use the repartition hint to merge small files, and the number of partitions in the repetition may not be reasonable. This will result in the zorder compression rate not being high enough, or because the number of repartition partitions is too small, resulting in a long sort time overhead. So we hope to have a configuration to remove the user-specified repartition.

### How should we improve?

Add a resolve rule to remove user-specified repartition before writing when using zorder

huangxiaopingRD:
# :mag: Description
## Issue References 🔗

This pull request fixes #6560

## Describe Your Solution 🔧

-   Add `RemoveRepartitionBeforeInsertInto` to remove the user-specified repartition

Because the user added unreasonable repartition, the compression rate will not be high enough after turning on zorder, and the overall time consumption will increase significantly.

After removing the user-specified repartition, the parallelism of Rebalance is more reasonable, resulting in higher compression rate and shorter execution time.
#### Related Unit Tests
`ZorderSuiteBase / test("Check remove user specify repartition as expected")`

---

# Checklist 📝

**Be nice. Be informative.**


pan3793:
Generally, we should always respect user explicitly requested data distribution to unsurprise users. My view may be too pessimistic, breaking user requested data distribution may cause correctness issue in corner cases, I'm -0 on introduing this feature. But I'm OK if other committers accept it, as long as we disable it by default and explain the dangerous in the docs

huangxiaopingRD:
> Generally, we should always respect user explicitly requested data distribution to unsurprise users. My view may be too pessimistic, breaking user requested data distribution may cause correctness issue in corner cases, I'm -0 on introduing this feature. But I'm OK if other committers accept it, as long as we disable it by default and explain the dangerous in the docs

Your concern is correct. This feature is only decided by the user, not by the platform. This feature is similar to Spark's [RemoveAllHints](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/ResolveHints.scala#L298)

PR #6585, opened 2024-08-05
Support Multiple Configuration Files in FileSessionConfAdvisor

lsm1:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

The current FileSessionConfAdvisor only supports configuring a single file through kyuubi.session.conf.profiler. However, users may need to use multiple configuration files, and the existing setup does not support this. To address this need, we have updated the FileSessionConfAdvisor to support multiple configuration files.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


cxzl25 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/session/FileSessionConfAdvisor.scala
```diff
@@ -51,15 +51,18 @@ object FileSessionConfAdvisor extends Logging {
...
       .build(new CacheLoader[String, JMap[String, String]] {
-        override def load(profile: String): JMap[String, String] = {
-          val propsFile = Utils.getPropertiesFile(s"kyuubi-session-$profile.conf")
-          propsFile match {
-            case None =>
-              error("File not found: $KYUUBI_CONF_DIR/" + s"kyuubi-session-$profile.conf")
-              Collections.emptyMap()
-            case Some(_) =>
-              Utils.getPropertiesFromFile(propsFile).asJava
+        override def load(profiles: String): JMap[String, String] = {
```
wForget, line 54: Should the key of sessionConfCache be one profile?

kyuubi-server/src/test/scala/org/apache/kyuubi/plugin/PluginLoaderSuite.scala
```diff
@@ -67,6 +67,19 @@ class PluginLoaderSuite extends KyuubiFunSuite {
     assert(clusterAConfFromCache.get("kyuubi.ha.namespace") == "kyuubi-ns-a")
     assert(clusterAConfFromCache.get("kyuubi.zk.ha.namespace") == null)
     assert(clusterAConfFromCache.size() == 5)
+
+    conf.set(KyuubiConf.SESSION_CONF_PROFILE, Seq("cluster-a", "cluster-b"))
+    val clusterABConf =
```
wForget, line 72: This case does not seem to reflect the scenario of multiple profiles.

lsm1: In this case, since the value of `kyuubi.ha.namespace` is different, the value from `cluster-b` will replace the value from `cluster-a`. I will add more key-value pairs to `cluster-b` to refine this case.

wForget (approved):
LGTM

lsm1:
Merge to master

PR #6587, opened 2024-08-06
Periodically expire temp files and operation logs on server to avoid memeory leak by Files.deleteOnExit

bowenliang123:
# :mag: Description
## Issue References 🔗

-

## Describe Your Solution 🔧
Fix the memory leak on server caused by `Files.deleteOnExit`.
For long-running Kyuubi server instances, some operation log files and batch job upload files are marked for deletion at exit using `Files.deleteOnExit`. However, the `files` list within the `DeleteOnExitHook`  by `Files.deleteOnExit` method continuously accumulates file paths without being cleaned up, leading to a memory leak issue.

This PR fix this issue by:
1. introduce a new util `FileExpirationUtils` for similar use of `Files.deleteOnExit`, with exposed method for evict file path from the list to prevent accumulative path list
2. adding a service `TempFileService ` in server module, periodical clean-up the files for operation logging path, uploaded resources and etc. And it evict the paths in `TempFileCleanupUtils` instance after cleanup.
3. add the new config `kyuubi.server.tempFile.expireTime` with a default value of 7 days, to control How often to trigger a file expiration clean-up for stale files

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


kyuubi-common/src/main/scala/org/apache/kyuubi/util/FileExpirationUtils.scala
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.{Long => JLong}
+import java.nio.file.{Files, Path}
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicLong
+
+import com.google.common.cache.{Cache, CacheBuilder, RemovalNotification}
+
+import org.apache.kyuubi.Utils
+
+object FileExpirationUtils {
```
pan3793, line 28: IMO this is more like a service instead of a util

bowenliang123: But if it's put into a service, the service has to accept the path exposed by common tools like `OperationLog`.

bowenliang123: It's now refactored into a server's service. Please have a check.

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -51,17 +51,19 @@ object OperationLog extends Logging {
   /**
    * The operation log root directory, this directory will delete when JVM exit.
```
pan3793, line 52: comments are outdated

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
```diff
@@ -51,17 +51,19 @@ object OperationLog extends Logging {
...
-        path.toFile.deleteOnExit()
+        FileExpirationUtils.deleteFileOnExit(path)
+        path
       } catch {
         case e: IOException =>
           error(s"Failed to create operation log root directory: $path", e)
+          null
       }
-    }
+    }.orNull
```
pan3793, line 66: why not just return a `Option[Path]`?

kyuubi-common/src/main/scala/org/apache/kyuubi/util/FileExpirationUtils.scala
```diff
@@ -0,0 +1,57 @@
...
+  private lazy val isCleanupShutdownHookInstalled = {
+    addFilesCleanupOnExitShutdownHook()
+    new AtomicBoolean(true)
+  }
+
+  private lazy val files: util.Set[String] =
+    Collections.synchronizedSet(new util.LinkedHashSet[String]())
```
pan3793, line 36: it's weird to maintain state in a Utils class, can we merge it into the Service class? BTW, we don't need to use an external collection to track those files, a Cache instance is sufficient, see https://github.com/apache/kyuubi/pull/6319/files

kyuubi-server/src/main/scala/org/apache/kyuubi/server/FileCleanupService.scala
```diff
@@ -0,0 +1,75 @@
...
+import java.util.concurrent.atomic.AtomicLong
+
+import com.google.common.cache.{Cache, CacheBuilder, RemovalNotification}
+
+import org.apache.kyuubi.Utils
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.service.AbstractService
+import org.apache.kyuubi.util.FileExpirationUtils
+
+class FileCleanupService(name: String) extends AbstractService(name) {
```
pan3793, line 31: seems we also need such a service for engine.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/FileCleanupService.scala
```diff
@@ -0,0 +1,75 @@
...
+
+    cleanupScheduler.scheduleAtFixedRate(
+      () => expiringFiles.cleanUp(),
+      0,
+      Math.max(interval / 10, 100),
+      TimeUnit.MILLISECONDS)
+  }
+
+  override def stop(): Unit = {
+    expiringFiles.invalidateAll()
```
pan3793, line 58: Do we guarantee it will be called after SessionManager stops?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/FileCleanupService.scala
pan3793: maybe we can give it a more generic name, like `TempFileManager`

kyuubi-common/src/main/scala/org/apache/kyuubi/util/FileExpirationUtils.scala
bowenliang123: It's a list not a state? Inside the Java util's `Files.deleteOnExit` method, there's a file list maintained in `DeleteOnExitHook`. So it is a similar design here.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/FileCleanupService.scala
bowenliang123: OK. Renamed to `TempFileManager` .

I would keep this inside the service module, as
1. it require's Guava cache for eviction and removal listener. the common module does not depend on Guava.
2. server is required to maintain running status for much longer time than engine. so the leaking problem is more urgent in server.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/FileCleanupService.scala
bowenliang123: It's not guaranteed. Are you suggesting to use another shutdown hook for cleanup expiring files here?

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
bowenliang123: Updated.

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/log/OperationLog.scala
bowenliang123: The existed code Line 86 has the similar usage with `operationLogRoot`.
https://github.com/apache/kyuubi/pull/6587/files#diff-7e407151fdede81d22e23cc44075af5ed3ce5d54c702a84cbbf74d54ac459537R86

kyuubi-common/src/main/scala/org/apache/kyuubi/util/FileExpirationUtils.scala
pan3793: In fact, the `java.io.*` classes were one of the most controversial designs in the Java standard library, and the `java.nio.*` classes were later introduced in subsequent versions. Sometimes the JDK standard library does not follow common design patterns, especially parts closely related to the OS platform.

I would suggest moving TempFileManager to the common module, and centralizing the logic into it. Guava dep is not an issue for me, as it has no real effects on the leaf modules, e.g. `kyuubi-server`, `spark-sql-engine`

pan3793: BTW, I'm investigating the Spark Connect support, which based on the gRPC protobuf, Guava will eventually land common, and everywhere ...

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TempFileService.scala
```diff
@@ -0,0 +1,92 @@
...
+    ThreadUtils.newDaemonSingleThreadScheduledExecutor(s"temp-file-cleanup-scheduler")
+
+  override def initialize(conf: KyuubiConf): Unit = {
+    super.initialize(conf)
+    val interval = conf.get(KyuubiConf.SERVER_STALE_FILE_EXPIRATION_INTERVAL)
+    expiringFiles = CacheBuilder.newBuilder()
+      .expireAfterWrite(interval, TimeUnit.MILLISECONDS)
+      .removalListener((notification: RemovalNotification[String, String]) => {
+        val pathStr = notification.getValue
+        cleanupFilePath(pathStr)
```
pan3793, line 45: add a info log

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -118,6 +120,8 @@ class KyuubiSessionImpl(
     // we should call super.open before running launch engine operation
     super.open()
 
+    Option(tempFileService).foreach(_.addPathToExpiration(operationalLogRootDir.orNull))
```
pan3793, line 123: `addPathToExpiration` requires arg not null

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -97,7 +98,8 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       this,
       userConf,
       userConf.get(ENGINE_DO_AS_ENABLED),
-      parser)
+      parser,
+      tempFileService = kyuubiServer.tempFileService)
```
pan3793, line 102: do we need each session to hold one TempFileService instance? if not, just access the global TempFileService through SessionManager

kyuubi-common/src/main/scala/org/apache/kyuubi/util/TempFileCleanupUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  private def installFilesCleanupOnExitShutdownHook(): Unit = {
+    Utils.addShutdownHook(
+      hook = () => {
+        deleteTargets.foreach(pathStr =>
+          Utils.deleteDirectoryRecursively(Paths.get(pathStr).toFile))
+        deleteTargets.clear()
+      },
+      priority = Integer.MIN_VALUE)
```
pan3793, line 47: is it same as the JDK's one?

pan3793:
overall lgtm, just some nits

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TempFileService.scala
bowenliang123: I would prefer a debug log here. It's maybe noisy to have a info log, as every session log root path will be cleaned-up by expiration.

Added the debug log on removing expired file here.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
bowenliang123: Yes. Let me add the guard check for it. BTW, the operationalLogRootDir is never null here as it's inited in L121 `super.open()`.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
bowenliang123: No. And all sessions share one TempFileService instance.
Here the server's `KyuubiSessionManager` passes the one "global" TempFileService in KyuubiServer to init `KyuubiSessionImpl` in server module, just like the way in Line101 for shared parser crossing sessions in `KyuubiSessionManager`.

pan3793: looks like the `parser` is an exception here. we don't need to pass a global sharable instance into each session, that makes us clear to know we are touching a sharable or session-isolated instance

for exmaple
```
class KyuubiSessionImpl {

  sessionManager.abc // we are touching a sharable instance

  conf.xxx // we are touching a session-isolated instance
}
```

bowenliang123: Thanks for the explanation and the example. Let me change to use the `kyuubiServer.tempFileService` global instance directly in `KyuubiSessionImpl` class.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -118,6 +118,10 @@ class KyuubiSessionImpl(
     // we should call super.open before running launch engine operation
     super.open()
 
+    operationalLogRootDir.foreach { dir =>
+      Option(sessionManager.tempFileService).foreach(_.addPathToExpiration(dir))
```
pan3793, line 122: `sessionManager.tempFileService` should always be present, right?

bowenliang123: Yes. You're right. `sessionManager.tempFileService` is always present, let me skip the option wrapper here.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -3106,6 +3106,14 @@ object KyuubiConf {
       .timeConf
       .createWithDefaultString("PT30M")
 
+  val SERVER_STALE_FILE_EXPIRATION_INTERVAL: ConfigEntry[Long] =
+    buildConf("kyuubi.server.stale.file.expiration.interval")
+      .doc("How often to trigger a file expiration clean-up for stale files")
```
bowenliang123, line 3111: Is there any better suggestion for the config name ? @pan3793

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TempFileService.scala
```diff
@@ -0,0 +1,93 @@
...
+   * add the file path to the expiration list
+   * ensuring the path will be deleted
+   * either after duration
+   * or on the JVM exit
+   *
+   * @param path the path of file or directory
+   */
+  def addPathToExpiration(path: Path): Unit = {
+    require(path != null)
+    if (!expiringFiles.asMap().values().contains(path)) {
```
cxzl25, line 82: Why is the key value of excitingFiles designed like this id -> path?

How efficient is the contains value method?

bowenliang123: > Why is the key value of excitingFiles designed like this id -> path?

In a way to make use of Guava's Cache eviction mechanism preventing manual maintenance for expiration time and the ticker state. The Guava cache is in key-value form. The cache eviction lazily evicts the expired entries, and here we have a dedicate scheduled thread to call `cleanup` method to actively evict them in a much longer interval.

> How efficient is the contains value method?

After the second thoughts, I have skipped the existence check for two reasons. One is that Guava's LoadCache loops the all the tables of segments inside `containsValue` method. Second, the tradeoff between saving the repeated values and maintaining value set.

kyuubi-common/src/main/scala/org/apache/kyuubi/util/TempFileCleanupUtils.scala
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.atomic.AtomicBoolean
+
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
+import org.apache.kyuubi.Utils
+
+object TempFileCleanupUtils {
+  private lazy val deleteTargets: mutable.Set[String] =
+    Collections.synchronizedSet(new util.HashSet[String]).asScala
```
cxzl25, line 33: `ConcurrentHashMap.newKeySet[String]()`

bowenliang123: OK, applied.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
bowenliang123: Changed to `kyuubi.server.tempFile.expireTime`, as discussed offline.

bowenliang123:
Thanks, merged to master (1.10.0)

PR #6588, opened 2024-08-06
Get effective USER if env var is unavailable

pan3793:
# :mag: Description

In some cases, env var USER is unavailable, for example, docker container, we should evaluate it via `id -nu`

ref: https://github.com/apache/hadoop/blob/rel/release-3.4.0/hadoop-common-project/hadoop-common/src/main/bin/hadoop-functions.sh#L893-L896

## Test Plan 🧪

```
$ docker run -t -i ubuntu:latest
root@1dbeaefd6cd4:/# echo $USER

root@1dbeaefd6cd4:/# id -nu
root
root@1dbeaefd6cd4:/# exit
```

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/bin/kyuubi
+++ b/bin/kyuubi
@@ -35,6 +35,11 @@ if [[ "$*" = *--help ]] || [[ "$*" = *-h ]]; then
   exit 0
 fi
 
+# if for some reason the shell doesn't have $USER defined
+# (e.g., ssh'd in to execute a command)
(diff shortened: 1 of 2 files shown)


cxzl25 (approved):

pan3793:
Merged to 1.10.0/1.9.3/1.8.3

PR #6592, opened 2024-08-06
Add uploadWorkDir into local dir allow list and do not skip check for resource uploaded use case

turboFei:
# :mag: Description
## Issue References 🔗

As title, for security concern. 

Before, it skips to check the local dir access for resource upload use case, which is not expected.
## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


pan3793 (approved):

pan3793:
Merged to 1.10.0/1.9.3/1.8.3

PR #6606, opened 2024-08-12
`build/mvn` checks project local downloaded binary ahead

pan3793:
# :mag: Description

I found sometimes `build/mvn -version` is pretty slow even `build/apache-maven-${MVN_VERSION}` is already cached, and I found it may stuck at the following call when network quality is not good.

https://github.com/apache/kyuubi/blob/bdd91f45396ab20b6dfe0a266fb8bfe340d99192/build/mvn#L86

## Test Plan 🧪

Tested `build/mvn -version` in the following cases:
1. when `build/apache-maven-${MVN_VERSION}/bin/mvn` is available, the command always finishes quickly.
2. when `build/apache-maven-${MVN_VERSION}/bin/mvn` is unavailable but global installed `mvn` version matches `maven.version` defined in root `pom.xml`, the command always finishes quickly too.
3. otherwise, it automatically downloads the maven binary tarball from network.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/build/mvn
+++ b/build/mvn
@@ -45,59 +45,58 @@ install_app() {
   curl_opts="--progress-bar ${curl_opts}"
   wget_opts="--progress=bar:force ${wget_opts}"
 
-  if [ -z "$3" -o ! -f "$binary" ]; then
-    # check if we already have the tarball
-    # check if we have curl installed
-    # download application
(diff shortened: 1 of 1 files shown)


bowenliang123 (approved):

pan3793:
Thanks, merged to master/1.9.3/1.8.3

PR #6626, opened 2024-08-17
Fix operation never expired issue with periodical listOperations api calls

turboFei:
# :mag: Description
## Issue References 🔗

This pull request fixes operation never expired issue.
I found that, some Kyuubi operations never expired.

After investigation, the root cause:
1. our kyuubi console lookup the operations periodically
2. it retrieves the kyuubi operation events
3. the KyuubiOperationEvent::apply will get the operation status and update the operation the last access time, https://github.com/apache/kyuubi/pull/2452 
4. then the operation never expired.
## Describe Your Solution 🔧

Just get the operation event without updating the operation last access time.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/events/KyuubiOperationEvent.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/events/KyuubiOperationEvent.scala
@@ -18,8 +18,6 @@
 package org.apache.kyuubi.events
 
 import org.apache.kyuubi.Utils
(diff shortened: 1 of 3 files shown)


turboFei:
cc @pan3793

wForget (approved):
LGTM, Thanks

turboFei:
thanks, merged to 1.10.0 and 1.9.3

PR #6627, opened 2024-08-17
Using concurrent collections for session operation handle set

turboFei:
# :mag: Description
## Issue References 🔗

As mentioned in https://github.com/apache/kyuubi/pull/6626, the operation never idle because of periodical get operation status, but in my opinion, the operations should be closed after session closed, but it did not.

Currently, for session operation handles set, it does not use concurrent collections, I wonder there is concurrent update issue.

## Describe Your Solution 🔧

#### Related Unit Tests
Not needed.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala
@@ -17,6 +17,8 @@
 
 package org.apache.kyuubi.operation
 
+import java.util.concurrent.ConcurrentHashMap
+
 import scala.collection.JavaConverters._
 
 import org.apache.kyuubi.KyuubiSQLException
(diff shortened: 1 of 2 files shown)


turboFei:
cc @pan3793 @wForget

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala
```diff
@@ -36,7 +38,7 @@ import org.apache.kyuubi.shaded.hive.service.rpc.thrift._
  */
 abstract class OperationManager(name: String) extends AbstractService(name) {
 
-  final private val handleToOperation = new java.util.HashMap[OperationHandle, Operation]()
+  final private val handleToOperation = new ConcurrentHashMap[OperationHandle, Operation]()
```
wForget, line 41: The method using `handleToOperation` has synchronized block, do we still need this change?

https://github.com/apache/kyuubi/blob/0165d5e3a26cc8353a9f7135e8fa4a9c4ae99e78/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala#L98-L101

turboFei: thanks, reverted

wForget (approved):

turboFei:
thanks, merged to 1.10.0 and 1.9.3

PR #6629, opened 2024-08-19
[KYUUBI #6335][FOLLOWUP] Using sessionId for sessionUploadFolderPath

turboFei:
# :mag: Description
## Issue References 🔗

Followup for #6335 

Just use the session handle identifier to build the batchResourceUploadFolderPath, and do not rely on the BatchJobSubmission operation initialization. 
## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
@@ -24,6 +24,8 @@ import java.util.Locale
 
 import scala.util.control.NonFatal
 
+import org.apache.commons.lang3.StringUtils
+
 import org.apache.kyuubi.{KyuubiException, Utils}
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.KubernetesApplicationOperation.LABEL_KYUUBI_UNIQUE_KEY
@@ -210,5 +212,10 @@ object KyuubiApplicationManager {
       case appType if appType.startsWith("FLINK") => // TODO: check flink app access local paths
       case _ =>
     }
+  }
+
+  def sessionUploadFolderPath(sessionId: String): Path = {
+    require(StringUtils.isNotBlank(sessionId))
+    uploadWorkDir.resolve(sessionId)
   }
 }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -545,7 +545,7 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       resourceFileInputStream: InputStream,
       resourceFileName: String,
       formDataMultiPartOpt: Option[FormDataMultiPart]): Option[JPath] = {
-    val uploadFileFolderPath = batchResourceUploadFolderPath(batchId)
+    val uploadFileFolderPath = KyuubiApplicationManager.sessionUploadFolderPath(batchId)
     try {
       handleUploadingResourceFile(
         request,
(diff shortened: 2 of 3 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSession.scala
```diff
@@ -82,8 +82,8 @@ class KyuubiBatchSession(
   override val normalizedConf: Map[String, String] =
     sessionConf.getBatchConf(batchType) ++ sessionManager.validateBatchConf(conf)
 
-  private[kyuubi] def resourceUploadFolderPath: Path =
-    BatchesResource.batchResourceUploadFolderPath(batchJobSubmissionOp.batchId)
+  private[kyuubi] lazy val resourceUploadFolderPath: Path =
```
bowenliang123, line 85:
```suggestion
  private[kyuubi] def resourceUploadFolderPath: Path =
```
Minor optional suggestion: function instead of holding the variable, as it's able to be recalculated.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -642,6 +642,8 @@ object BatchesResource {
     Option(batchState).exists(bt => VALID_BATCH_STATES.contains(bt.toUpperCase(Locale.ROOT)))
   }
 
-  def batchResourceUploadFolderPath(batchId: String): JPath =
-    KyuubiApplicationManager.uploadWorkDir.resolve(s"batch-$batchId")
```
bowenliang123, line 646: I have removed the prefix `batch-` here.

bowenliang123 (approved):
LGTM.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
```diff
@@ -211,4 +213,9 @@ object KyuubiApplicationManager {
       case _ =>
     }
   }
+
+  def sessionUploadFolderPath(sessionId: String): Path = {
+    require(StringUtils.isNotBlank(sessionId))
+    uploadWorkDir.resolve(sessionId)
+  }
```
turboFei, line 220: how about moving to `KyuubiApplicationManager.scala`? @bowenliang123

bowenliang123: SGTM.

turboFei:
thanks, merged to 1.10.0

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -545,7 +545,7 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       resourceFileInputStream: InputStream,
       resourceFileName: String,
       formDataMultiPartOpt: Option[FormDataMultiPart]): Option[JPath] = {
-    val uploadFileFolderPath = batchResourceUploadFolderPath(batchId)
+    val uploadFileFolderPath = KyuubiApplicationManager.sessionUploadFolderPath(batchId)
```
bowenliang123, line 548: Hi @turboFei @pan3793 , as batch id is not the same as the session id, is it a misuse here, uploaded files are put along with `batch id`, but the cleanup method cleans up the path with `session id` ?

turboFei: batchId == batch sessionId

bowenliang123: OK. Thanks

PR #6632, opened 2024-08-21
[KYUUBI #6619][FOLLOWUP] Do not log auth audit log for thrift http cookie auth

turboFei:
# :mag: Description
## Issue References 🔗

Follow up for #6619, I found that, there are many duplicate auth audit logs for thrift http.

We shall only log when make authentication with auth handler, and skip it for cookie auth.
## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


pan3793 (approved):

turboFei:
thanks, merged to 1.10.0

PR #6639, opened 2024-08-22
Port HIVE-27815: Support update numModifiedRows

pan3793:
# :mag: Description

Backport https://github.com/apache/hive/pull/4819

Note: it's only the JDBC driver side change, to make it work, we also need to modify the engines.

## Test Plan 🧪

We can not test this feature so far because we don't have engine support that. Pass GHA to ensure it breaks nothing.

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
@@ -380,7 +380,7 @@ public class KyuubiStatement implements SQLStatement, KyuubiLoggable {
     TGetOperationStatusResp statusResp = null;
 
     // Poll on the operation status, till the operation is complete
-    while (!isOperationComplete) {
+    do {
       try {
         /**
          * For an async SQLOperation, GetOperationStatus will use the long polling approach It will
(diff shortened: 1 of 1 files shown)


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiStatement.java
```diff
@@ -590,7 +590,16 @@ public int getUpdateCount() throws SQLException {
      * finished.
      */
     waitForOperationToComplete();
```
turboFei, line 592: Seems this line can be removed.

pan3793: right, thanks for checking

turboFei (approved):

pan3793:
interesting error, CI runs at mid-night at UTC timezone, thus failed with

> Path does not exist: file:/home/runner/work/kyuubi/kyuubi/kyuubi-server/target/tmp/kyuubi-9be91795-ccaa-4905-b249-8d147597f34f/kyuubi_operation/day=20240825

let me re-trigger it.

pan3793:
Merged to master

pan3793:
Given this causes significant perf regression(see https://github.com/apache/kyuubi/issues/6796) and the newly added functionality is not completed, let me revert it first.

PR #6645, opened 2024-08-28
Size based eviction for server-side temp files cleanup

bowenliang123:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

- adding `maximumSize` to support size based eviction for server-side temp files cleanup in `TempFileService`
- size-based eviction is disabled by default , with `maximumSize` set to optional by default
- time-based eviction time is now extended from 14 days to 30 days by default

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -3121,6 +3121,14 @@ object KyuubiConf {
       .timeConf
       .createWithDefaultString("P14D")
 
+  val SERVER_TEMP_FILE_EXPIRE_MAX_COUNT: ConfigEntry[Long] =
+    buildConf("kyuubi.server.tempFile.maxCount")
+      .doc("The upper threshold size of server-side temporary file paths to cleanup")
+      .version("1.10.0")
+      .serverOnly
+      .longConf
+      .createWithDefault(10000)
```
pan3793, line 3130: how much memory will be occupied?

bowenliang123: Estimated in a map struct, the total memory consumption would be approximately:
max count 10000 * (key length 13+6 + value length ~50 ) =  2,020,000B ~= 2MB

bowenliang123: Or do you want me to increase the default value?

pan3793: A potential issue, what will happen if an active file expires.

bowenliang123: Both time-based and size-based eviction is a fallback approach to clean up the temp files. It's assumed that the threshold is large enough for use.

pan3793: 14d might be long enough, 10000 is relatively small, how about disabling the size-based eviction strategy by default?

bowenliang123: OK. 1. Disabled size-based eviction strategy by default ; 2. Increased the default time-base eviction time to 30days.

pan3793 (approved):

bowenliang123:
Thanks, merged to master (1.10.0).

PR #6646, opened 2024-08-28
Fix RESTful API NPE when restarting Kyuubi server

turboFei:
# :mag: Description
## Issue References 🔗

We meet issue when restarting the kyuubi server:
```
2024-08-28 09:48:41.507 WARN org.apache.kyuubi.server.api.RestExceptionMapper: Error occurs on accessing REST API.
java.lang.NullPointerException: Cannot invoke "org.apache.kyuubi.server.KyuubiServer.getConf()" because the return value of "org.apache.kyuubi.server.KyuubiServer$.kyuubiServer()" is null
	at org.apache.kyuubi.server.api.v1.BatchesResource.batchV2Enabled(BatchesResource.scala:72) ~[kyuubi-server_2.12-1.10.0.0.1.0.jar:1.10.0.0.1.0]
	at org.apache.kyuubi.server.api.v1.BatchesResource.$anonfun$batchInfo$3(BatchesResource.scala:345) ~[kyuubi-server_2.12-1.10.0.0.1.0.jar:1.10.0.0.1.0]
	at scala.Option.map(Option.scala:230) ~[scala-library-2.12.18.jar:?]
```

The root cause is that, the `KyuubiServer.kyuubiServer` is null until the kyuubi server start finished.

https://github.com/apache/kyuubi/blob/db57e9365d7933942197936ac7ff711d58f7ea91/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiServer.scala#L231-L234

But the RESTful api initialization finished before than that, so, NPE will be thrown between REST front started to all the kyuubi service started.
## Describe Your Solution 🔧

In this PR, 
1. move the KyuubiBatchService into KyuubiRestFrontendService
2. move the TempFileService (introduced in #6587) into KyuubiSessionManager

image

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/TempFileService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/TempFileService.scala
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.kyuubi.service
+package org.apache.kyuubi.server
 
 import java.nio.file.{Path, Paths}
 import java.util.concurrent.TimeUnit
@@ -25,7 +25,7 @@ import com.google.common.cache.{Cache, CacheBuilder, RemovalNotification}
 
 import org.apache.kyuubi.Utils
 import org.apache.kyuubi.config.KyuubiConf
-import org.apache.kyuubi.service.TempFileService.tempFileCounter
+import org.apache.kyuubi.service.AbstractService
 import org.apache.kyuubi.util.{TempFileCleanupUtils, ThreadUtils}
 
 class TempFileService(name: String) extends AbstractService(name) {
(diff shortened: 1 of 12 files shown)


turboFei:
The UT passed.

gentle ping @pan3793 @bowenliang123

turboFei:
also cc @wForget

kyuubi-server/src/main/scala/org/apache/kyuubi/server/TempFileService.scala
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.kyuubi.service
+package org.apache.kyuubi.server
```
bowenliang123, line 18: How about keeping the `TempFileService` class in the `org.apache.kyuubi.service` package?

bowenliang123 (approved):
LGTM.

turboFei:
thanks, merging to master

pan3793:
late lgtm too, thanks for fixing this.

PR #6648, opened 2024-08-30
Support to show launch engine log for thrift http mode

turboFei:
# :mag: Description
## Issue References 🔗

This PR is to return the launch engine operation handle for thrift http frontend service when opening session.

So that, the kyuubi connection show the launch engine log.

Refer https://github.com/apache/kyuubi/pull/1377

## Describe Your Solution 🔧

Just return the launch engine op handle for thrift http connection.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
@@ -228,7 +228,7 @@ final class KyuubiTHttpFrontendService(
     super.initialize(conf)
   }
 
-  // same as KyuubiTBinaryFrontendService, to return launch engine op handle
+  /** Same as KyuubiTBinaryFrontendService, to return launch engine op handle. */
   override def OpenSession(req: TOpenSessionReq): TOpenSessionResp = {
     debug(req.toString)
     info("Client protocol version: " + req.getClient_protocol)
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
```diff
@@ -221,6 +228,45 @@ final class KyuubiTHttpFrontendService(
     super.initialize(conf)
   }
 
+  // same as KyuubiTBinaryFrontendService, to return launch engine op handle
```
pan3793, line 231: can we have an abstract class layer to share the code then?

pan3793 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiTHttpFrontendService.scala
pan3793: ah, let's leave it then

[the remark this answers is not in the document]
turboFei: done

turboFei:
thanks, merged to 1.10.0

pan3793:
As this also cover https://github.com/apache/kyuubi/issues/6338 for thrift HTTP mode, backported to 1.9.3

PR #6688, opened 2024-09-12
[SPARK] Avoid trigger execution when getting result schema

wForget:
# :mag: Description
## Issue References 🔗

`DataFrame.isEmpty` may trigger execution again, we should avoid it.

- [X] Bugfix (non-breaking change which fixes an issue)

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
@@ -59,14 +59,14 @@ class ExecutePython(
   override protected def supportProgress: Boolean = true
 
   override protected def resultSchema: StructType = {
-    if (result == null || result.schema.isEmpty) {
+    if (result == null) {
       new StructType().add("output", "string")
         .add("status", "string")
         .add("ename", "string")
         .add("evalue", "string")
         .add("traceback", "array<string>")
     } else {
-      result.schema
+      super.resultSchema
     }
   }
 
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
@@ -59,7 +59,7 @@ class PlanOnlyStatement(
   override protected def resultSchema: StructType = {
     if (result == null) {
       new StructType().add("plan", "string")
-    } else if (result.isEmpty) {
+    } else if (result.schema.isEmpty) {
       new StructType().add("result", "string")
     } else result.schema
   }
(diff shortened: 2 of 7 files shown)


pan3793:
cc @iodone

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -59,7 +59,7 @@ class PlanOnlyStatement(
   override protected def resultSchema: StructType = {
     if (result == null) {
       new StructType().add("plan", "string")
-    } else if (result.isEmpty) {
+    } else if (result.schema.isEmpty) {
```
ulysses-you, line 62: If I ger it correctly, when we go into thie else if branch, that means the query is in `planExcludes` ? We return the result rather than the plan. Can we add a new flag to distinguish them ? I'm not sure if the result schema of `planExcludes` is always empty.

wForget: I think we don't need to add a flag to distinguish them, because we handle all cases：

```
if (result == null) {
    ...
else if (result.schema.isEmpty) {
    ...
} else {
    result.schema
}
```

wForget: And it is consistent with `ExecuteStatement`

https://github.com/apache/kyuubi/blob/635c79373df921018a48cb7c354719e5a9254641/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecuteStatement.scala#L53-L58

ulysses-you: ah, I see. How about reduce the else if and else by using `super.resultSchema()` ?

wForget: Sounds good, thanks, I will change that

ulysses-you (approved):

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -59,14 +59,14 @@ class ExecutePython(
...
-    if (result == null || result.schema.isEmpty) {
+    if (result == null) {
       new StructType().add("output", "string")
         .add("status", "string")
         .add("ename", "string")
         .add("evalue", "string")
         .add("traceback", "array<string>")
     } else {
-      result.schema
+      super.resultSchema
```
iodone, line 69: The logic here is no different from before?

wForget: It looks no different, `result` is never assigned in `ExecutePython`.

bowenliang123:
Thanks, merged to master (1.10.0) and branch-1.9 (1.9.3).

PR #6702, opened 2024-09-18
support list server configs by admin cli tool

bowenliang123:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

- add command `list config` in admin cli tool , for listing all the server configs, which is useful for checking runtime configs when before/after refreshing the configs of user-defaults or non-user-defaults

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


pan3793:
listing configurations should restricted to admin only. additionally, we may need to have another parameter to define the display mode for configuration values, e.g.
- ALL - display all configuration values as-is
- REDACTED - to avoid displaying password/secret values

PS: I remember there was a similar request to have a configuration display mode of configurations for retrieving batch info

PR #6708, opened 2024-09-22
spark-sql-engine's classes should not be visible in server module

pan3793:
# :mag: Description
## Issue References 🔗

The server module does not consume the `spark-sql-engine` classes, but it supposes that
`spark-sql-engine`'s jar existed on the target folder, so here we declare it as a dependency
to make sure that Maven always processes `spark-sql-engine` module before the server module.
IntelliJ IDEA 2024.1 fixed the IDEA-93855, thus the relocated classes inside the
`spark-sql-engine`'s shaded jar are visible in the server module in IDEA, for example,
`org.apache.kyuubi.shaded.spark.connect.proto.ExecutePlanRequest`, which silently breaks
the IDEA code analysis and jumping capabilities.

## Describe Your Solution 🔧

Changing the dependency type from `jar`(default value) to `pom` seems to be a workaround.

## Test Plan 🧪

image

image

image

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/pom.xml
+++ b/kyuubi-server/pom.xml
@@ -77,7 +77,7 @@
 
         <dependency>
             <groupId>org.apache.kyuubi</groupId>
-            <artifactId>kyuubi-hive-jdbc</artifactId>
+            <artifactId>${hive.jdbc.artifact}</artifactId>
             <version>${project.version}</version>
         </dependency>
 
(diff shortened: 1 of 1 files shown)


cxzl25 (approved):

pan3793:
Thanks, merged to master/1.9.3/1.8.3

PR #6727, opened 2024-10-15
replace immutable empty list and map in BatchRequest initialization 

bowenliang123:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

- replace immutable empty list and map with mutable one for BatchRequest initialization, for easier BatchRequest construction for rest-client use

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
+++ b/kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
@@ -17,10 +17,7 @@
 
 package org.apache.kyuubi.client.api.v1.dto;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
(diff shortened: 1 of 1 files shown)


pan3793:
> for easier BatchRequest construction for rest-client use

what's the real use case?

bowenliang123:
eg. 
```
BatchRequest a = new BatchRequest();
a.getConf().put("abc","def");
a.getExtraResourcesMap().put("opq","xyz");
```

pan3793:
let's update the getters too.
```
  public Map<String, String> getConf() {
    if (null == conf) {
      return Collections.emptyMap();
    }
    return conf;
  }

  public List<String> getArgs() {
    if (null == args) {
      return Collections.emptyList();
    }
    return args;
  }
```

bowenliang123:
Cool, thanks for the reminder. Cleaned up.

kyuubi-rest-client/src/main/java/org/apache/kyuubi/client/api/v1/dto/BatchRequest.java
```diff
@@ -90,9 +87,6 @@ public void setName(String name) {
   }
 
   public Map<String, String> getConf() {
-    if (null == conf) {
-      return Collections.emptyMap();
-    }
```
pan3793, line 95: I mean keep the null check but change the default value, otherwise, it may introduce regression if user calls `setConf(null)` explicitly before calling `getConf`

bowenliang123: SGTM.

pan3793 (approved):

bowenliang123:
Thanks, merged to master (1.10.0).

PR #6750, opened 2024-10-17
[REST] Using ForbiddenException instead of NotAllowedException

turboFei:
# :mag: Description
## Issue References 🔗

Seems NotAllowedException is used for method not allowed, and currently, we use false constructor, the error message we expected would not be return to client end.

It only told:
```
{"message":"HTTP 405 Method Not Allowed"}
```
Because the message we used to build the NotAllowedException was treated as `allowed` method, not as `message`.

image

## Describe Your Solution 🔧

We should use the ForbidenException instead, and then the error message we excepted can be visible in client end.

https://github.com/apache/kyuubi/blob/85dd5a52efa790195ed46b713c53938ffd31c5d9/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/api.scala#L47-L51

#### Related Unit Tests

image

---

# Checklist 📝

**Be nice. Be informative.**


bowenliang123 (approved):
LGTM.

bowenliang123:
Thanks, merged to master(1.10.0) and branch-1.9(1.9.3).

PR #6758, opened 2024-10-18
[KYUUBI #6754] Improve the performance of ranger access requests

### LINKED ISSUE #6754: [Improvement] Improve the performance of ranger access requests
### Code of Conduct

### Search before asking

Right now in RuleAuthorization we use an ArrayBuffer to collect access requests, which is very slow because each new PrivilegeObject needs to be compared with all access requests.

### How should we improve?

We can use a HashMap to optimize this.

wankunde:
# :mag: Description
## Issue References 🔗

This pull request fixes #6754

## Describe Your Solution 🔧

Right now in RuleAuthorization we use an ArrayBuffer to collect access requests, which is very slow because each new PrivilegeObject needs to be compared with all access requests.

Before

After

#### Related Unit Tests

Test with local 50000 files:
```java
test("KYUUBI #6754: improve the performance of ranger access requests") {
    val outputPath = "/private/var/folders/tr/scn8dgl13_l6_sh17bghtln1b35kn1/T/kyuubi-test-5492934124608743789/"
    println("output path: "+ outputPath)

    val plugin = mock[SparkRangerAdminPlugin.type]
    when(plugin.verify(Seq(any[RangerAccessRequest]), any[SparkRangerAuditHandler]))
      .thenAnswer(_ => ())

    val df = spark.read.parquet(outputPath + "/*/*.parquet")
    val plan = df.queryExecution.optimizedPlan
    val start = System.currentTimeMillis()
    RuleAuthorization(spark).checkPrivileges(spark, plan)
    val end = System.currentTimeMillis()
    println(s"Time elapsed : ${end - start} ms")
  }
```

---

# Checklist 📝

- [ x] This patch was not authored or co-authored using [Generative Tooling](https://www.apache.org/legal/generative-tooling.html)

**Be nice. Be informative.**

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz/pom.xml
@@ -380,6 +380,19 @@
             <artifactId>${hudi.artifact}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.scalatestplus</groupId>
+            <artifactId>mockito-4-11_${scala.binary.version}</artifactId>
+            <version>${scalatestplus.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
@@ -17,7 +17,7 @@
 
 package org.apache.kyuubi.plugin.spark.authz.ranger
 
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.immutable.HashMap
 
 import org.apache.ranger.plugin.policyengine.RangerAccessRequest
 import org.apache.spark.sql.SparkSession
@@ -25,6 +25,7 @@ import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 
 import org.apache.kyuubi.plugin.spark.authz._
 import org.apache.kyuubi.plugin.spark.authz.ObjectType._
+import org.apache.kyuubi.plugin.spark.authz.ranger.AccessType.AccessType
 import org.apache.kyuubi.plugin.spark.authz.ranger.SparkRangerAdminPlugin._
 import org.apache.kyuubi.plugin.spark.authz.rule.Authorization
 import org.apache.kyuubi.plugin.spark.authz.util.AuthZUtils._
@@ -34,15 +35,14 @@ case class RuleAuthorization(spark: SparkSession) extends Authorization(spark) {
     val auditHandler = new SparkRangerAuditHandler
     val ugi = getAuthzUgi(spark.sparkContext)
     val (inputs, outputs, opType) = PrivilegesBuilder.build(plan, spark)
-    val requests = new ArrayBuffer[AccessRequest]()
+    var requests = new HashMap[(AccessResource, AccessType), AccessRequest]()
 
     def addAccessRequest(objects: Iterable[PrivilegeObject], isInput: Boolean): Unit = {
       objects.foreach { obj =>
         val resource = AccessResource(obj, opType)
         val accessType = ranger.AccessType(obj, opType, isInput)
-        if (accessType != AccessType.NONE && !requests.exists(o =>
-            o.accessType == accessType && o.getResource == resource)) {
-          requests += AccessRequest(resource, ugi, opType, accessType)
+        if (accessType != AccessType.NONE) {
+          requests += (resource, accessType) -> AccessRequest(resource, ugi, opType, accessType)
         }
       }
     }
@@ -50,7 +50,7 @@ case class RuleAuthorization(spark: SparkSession) extends Authorization(spark) {
     addAccessRequest(inputs, isInput = true)
     addAccessRequest(outputs, isInput = false)
 
-    val requestArrays = requests.map { request =>
+    val requestArrays = requests.values.map { request =>
       val resource = request.getResource.asInstanceOf[AccessResource]
       resource.objectType match {
         case ObjectType.COLUMN if resource.getColumns.nonEmpty =>
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorizationSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorizationSuite.scala
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.ranger
+
+import java.io.File
+import java.nio.file.Files
+
+import scala.reflect.io.Path.jfile2path
+
+import org.apache.ranger.plugin.policyengine.RangerAccessRequest
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito.when
+import org.scalatestplus.mockito.MockitoSugar
+// scalastyle:off
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.BeforeAndAfterAll
+
+import org.apache.kyuubi.plugin.spark.authz.SparkSessionProvider
+
+class RuleAuthorizationSuite extends AnyFunSuite
+  with SparkSessionProvider with BeforeAndAfterAll with MockitoSugar {
+  // scalastyle:on
+  override protected val catalogImpl: String = "hive"
+
+  private var tempDir: File = _
+
+  override def beforeAll(): Unit = {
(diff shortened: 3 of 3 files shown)


extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -380,6 +380,19 @@
...
+        <dependency>
+            <groupId>org.scalatestplus</groupId>
+            <artifactId>mockito-4-11_${scala.binary.version}</artifactId>
+            <version>${scalatestplus.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <version>${mockito.version}</version>
```
pan3793, line 392:
```suggestion
```

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -380,6 +380,19 @@
             <artifactId>${hudi.artifact}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.scalatestplus</groupId>
+            <artifactId>mockito-4-11_${scala.binary.version}</artifactId>
+            <version>${scalatestplus.version}</version>
```
pan3793, line 386:
```suggestion
```

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
```diff
@@ -34,23 +35,22 @@ case class RuleAuthorization(spark: SparkSession) extends Authorization(spark) {
     val auditHandler = new SparkRangerAuditHandler
     val ugi = getAuthzUgi(spark.sparkContext)
     val (inputs, outputs, opType) = PrivilegesBuilder.build(plan, spark)
-    val requests = new ArrayBuffer[AccessRequest]()
+    var requests = new HashMap[(AccessResource, AccessType), AccessRequest]()
```
pan3793, line 38: Why `immutable.HashMap` instead of `mutable.HashMap` is used here?

BTW, Scala 2.13 changed the alias of `HashMap` from `mutable.HashMap` to `immutable.HashMap`, to make the code clear, please write `new immutable.HashMap` or use `new mutable.HashMap` inline

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorizationSuite.scala
```diff
@@ -0,0 +1,71 @@
...
+  override def afterAll(): Unit = {
+    if (tempDir != null) {
+      tempDir.deleteRecursively()
+    }
+    spark.stop()
+    super.afterAll()
+  }
+
+  // scalastyle:on
+  test("KYUUBI #6754: improve the performance of ranger access requests") {
```
pan3793, line 56: benchmark cases are not expected to run in regular test workflow, I remember we ported Spark benchmark framework, can we write it in that style?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
pan3793: and please leave a few comments here to explain the background

extensions/spark/kyuubi-spark-authz/pom.xml
pan3793: we prefer to manage all deps version/exclusions in root pom's `dependencyManagement` block, and simply reference the groupId/artifactId/scope in the sub-modules pom's `dependencies` scope

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
pan3793: and, was `immutable.Map.newBuilder` considered?

wankunde: 1. Change to `mutable.HashMap`
2. Added some comments
3. I'm not sure if `immutable.Map.newBuilder` is better than `mutable.HashMap`?

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorizationSuite.scala
wankunde: Added benchmark, thanks

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
bowenliang123: How about using `mutable.LinkedHashSet[AccessRequest]` as the type for `requests`?
1. it maintains the original adding order
2. it guarantees the element's uniqueness reliing on the `hashCode` 
3. good for element looping as fast as the array

And make sure related methods of AccessRequest properly implemented. Add some unit test for assuring it.

wankunde: I'm not sure if there is any overhead in maintaining the linked elements? And would the `mutable.LinkedHashSet[AccessRequest]` be slower than `ArrayBuffer` when looping the elements?

extensions/spark/kyuubi-spark-authz/pom.xml
```diff
@@ -380,6 +380,14 @@
             <artifactId>${hudi.artifact}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-core_${scala.binary.version}</artifactId>
+            <version>${spark.version}</version>
```
pan3793, line 386: no need to duplicate version in sub-modules' pom
```suggestion
```

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
```diff
@@ -34,15 +35,19 @@ case class RuleAuthorization(spark: SparkSession) extends Authorization(spark) {
     val auditHandler = new SparkRangerAuditHandler
     val ugi = getAuthzUgi(spark.sparkContext)
     val (inputs, outputs, opType) = PrivilegesBuilder.build(plan, spark)
-    val requests = new ArrayBuffer[AccessRequest]()
+
+    // Use a HashSet to deduplicate the same AccessResource and AccessType, the requests will be all
+    // the non-duplicate requests.
```
pan3793, line 40: also mention the requirements of keep ordering

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/spark/sql/RuleAuthorizationBenchmark.scala
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Benchmark to measure performance of collecting ranger access requests.
+ *
+ * {{{
+ *
+ *   RUN_BENCHMARK=1 ./build/mvn clean test \
+ *   -pl extensions/spark/kyuubi-spark-authz -am \
+ *   -Pspark-3.3 \
+ *   -Dtest=none -DwildcardSuites=org.apache.spark.sql.RuleAuthorizationBenchmark
```
pan3793, line 46: authz module is activated unconditionally

```suggestion
 *   RUN_BENCHMARK=1 ./build/mvn clean test \
 *   -pl extensions/spark/kyuubi-spark-authz -am \
 *   -Dtest=none -DwildcardSuites=org.apache.spark.sql.RuleAuthorizationBenchmark
```

pan3793 (approved):
thanks, lgtm, only a few comments

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
bowenliang123: The uniqueness is guaranteed by `hashCode`, which could be told from its name `HashSet` of `LinkedHashSet`. Not much overhead cost for maintaining the order inside, just adding a reference to a linked list. And the linked list perform well in looping elements.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
wankunde: Done

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RuleAuthorization.scala
wankunde: I mean `mutable.LinkedHashSet[AccessRequest]` may be slower than HashSet + ArrayBuffer in our case.

pan3793: the unique key is NOT the `AccessRequest` itself but part of properties, the combination of ArrayBuffer with a HashSet looks fine to me.

bowenliang123: > I mean `mutable.LinkedHashSet[AccessRequest]` may be slower than HashSet + ArrayBuffer in our case.

No, it wont be slowed down, as LinkedHashSet maintaining both HashSet and Linked list inside.

bowenliang123: > the unique key is NOT the `AccessRequest` itself but part of properties, the combination of ArrayBuffer with a HashSet looks fine to me.

Noticed. But the `HashSet` still reily on the same logic with `hashCode` inside. And no proper customed method of  either `AccessRequest` or its parent class.

bowenliang123:
Thanks for your contribution! Merged to master (1.10.0) .

PR #6762, opened 2024-10-21
[BUILD] cleanup dockerfile warnings

bowenliang123:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

- fix the dockerfile warning reported by `GitHub Actions
/ Kyuubi Server On Kubernetes Integration Test` :
  - Legacy key/value format with whitespace separator should not be used
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format
More info: https://docs.docker.com/go/dockerfile/rule/legacy-key-value-format/

  - The 'as' keyword should match the case of the 'from' keyword
FromAsCasing: 'as' and 'FROM' keywords' casing do not match
More info: https://docs.docker.com/go/dockerfile/rule/from-as-casing/

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/build/Dockerfile.CI
+++ b/build/Dockerfile.CI
@@ -33,7 +33,7 @@
 # see: https://github.com/moby/moby/issues/38379
 ARG BASE_IMAGE=eclipse-temurin:8-jdk-focal
 
-FROM eclipse-temurin:8-jdk-focal as builder
+FROM eclipse-temurin:8-jdk-focal AS builder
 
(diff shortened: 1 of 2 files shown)


wForget (approved):

bowenliang123:
Thanks, merged to master(1.10.0).

PR #6779, opened 2024-10-23
[KYUUBI #6402]: engine.share.level=GROUP enable for a list of hadoop …

### LINKED ISSUE #6402: [Bug] engine.share.level=GROUP takes only first AD Group if the user is part of multiple AD Groups
### Code of Conduct

### Search before asking

### Describe the bug

If the user is part of multiple AD Groups and when engine.share.level is set to group, hadoop returns all groups that the user is part of but kyuubi takes the first group from that list and submits job as that particular group.

In the below error logs, user is aprt of multiple AD Groups. But kyuubi picks the first on i.e Internet_User to submit the job. And also it expects Internet_User to be part of YARN user which will not be found either

### Affects Version(s)

1.8.2

### Kyuubi Server Log Output

```logtalk
Failing this attempt.Diagnostics: [2024-05-06 09:14:47.685]Application application_1714974956360_0003 initialization failed (exitCode=255) with output: main : command provided 0
main : run as user is Internet_User
main : requested yarn user is Internet_User
User Internet_User not found
```

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

Madhukar525722:
…groups

# :mag: Description
## Issue References 🔗

This pull request fixes #6402 

## Describe Your Solution 🔧

Currently, the group level engine is getting launched with the first user group itself. This change will allow user to pass in which group they want to launch the engine. 

The flow of implementation is :
- Take input of preferred groups
    - If it is not defined, take the default route, pick the first group from the users groups list
    - If its defined
        - It will iterate the list, and look for the first match
        - If no match found, it will return the first group from the users groups list

1. In this way, for every group there will be single engine, which can be re-used.
2. First user launching the engine is implemented to handle the scenarios where Hadoop groups are not the yarn users.
3. For security concerns in 2nd point, we can have fine grained access control on session level, using Apache Ranger. 

## Test Results🧪
1. When PREFERRED_GROUPS were not defined, it used the first group from the list
Spark application name: kyuubi_GROUP_SPARK_SQL_Internet_Users_default_d2826b5b-3f1e-48a0-b42f-d248da914b7c
application ID: application_1728291907264_43966
User: madlnu

2. When valid PREFERRED_GROUPS were defined
Spark application name: kyuubi_GROUP_SPARK_SQL_kyuubi_test_b_default_be9a16a8-be38-4ab6-bee9-1934f8556f18
application ID: application_1728291907264_43968
User: madlnu

3. When no PREFERRED_GROUPS were matching, it used the first group from the list
Spark application name: kyuubi_GROUP_SPARK_SQL_Internet_Users_default_d2826b5b-3f1e-48a0-b42f-d248da914b7c
application ID: application_1728291907264_43966
User: madlnu

4. When any other user X tries to access the existing GROUP engine, it uses the same engine
Spark application name: kyuubi_GROUP_SPARK_SQL_Internet_Users_default_d2826b5b-3f1e-48a0-b42f-d248da914b7c
application ID: application_1728291907264_43966
User: madlnu

---

# Checklist 📝

**Be nice. Be informative.**


Madhukar525722:
Hi @pan3793 @bowenliang123 . Please review the implementation. Thanks

bowenliang123:
To support preferred the group name in session conf, I would suggest to do the following:
1. add a session conf key in `KyuubiReservedKeys` rather than in adding a server side config for a global preferred group (you are not even using its value  anyway).
2. add a server config for choosing the select policy, eg. 'head', 'prefered_session_conf'

And I have worries about using the session conf will interference choosing the engine reference, as it's easy to be changed at runtime. Is there any better approach with solid connection variables?

pan3793:
@Madhukar525722 thanks for taking care of this feature, I leave the comments to add a configuration "kyuubi.session.preferGroup" previously, but I have a little different idea now.

Now I would suggest having a "kyuubi.session.preferredGroups" (Seq[String]), when present, we select the most preferred group from the whole group list, otherwise, take the head, instead of failing fast. For the implementatio, we can use a custom `Comparator` to acheive that.

> And I have worries about using the session conf will interference choosing the engine reference, as it's easy to be changed at runtime. Is there any better approach with solid connection variables?

seems there is no much differences from the existing properties like `kyuubi.session.user`?

Madhukar525722:
Hi @pan3793 , I have considered your suggestion for using kyuubi.session.preferredGroups. Please review, once approved, I will rebase it.

Madhukar525722:
Hi @pan3793 , Please review the change, integration tests failed due to timeout. In previous it was green

pan3793:
@Madhukar525722 sorry for late reply, I'm quite busy these days, will take a look soon

PR #6783, opened 2024-10-28
Support save WithCTE for insertRepartitionBeforeWrite

ic4y:
# :mag: Description
## Issue References 🔗

First, I'd like to thank @wForget for the help with this issue.

When using the "save to HDFS" feature, queries ending with an `ORDER BY` sometimes lose their sort order in the results. Upon investigating the code, I discovered that when using `WITH` statements and saving SQL results with `toDF.write.save`, a `WithCTE` node is generated after the Sort node. This causes the `canInsertRepartitionByExpression` check to fail, leading to an incorrect `Repartition` node insertion after the Sort node, which ultimately disrupts the sort order.

However, this issue does not occur when using `INSERT INTO TABLE` with `WithCTE` nodes.

The provided unit test can reproduce this issue, but after using `toDF.write.save`, I am unable to access the complete execution plan to assert whether a `Repartition` node is present. Therefore, the current test is ineffective. 

Hope someone can help figure out how to write this unit test.

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**


ulysses-you (commented):
lgtm if tests pass

wForget:
There is an issue with the logic of check write in base unit test. Please wait #6793

PR #6785, opened 2024-10-28
Shutdown the executor service in KubernetesApplicationOperation and prevent NPE

turboFei:
# :mag: Description
## Issue References 🔗

As title.

Fix NPE, because the cleanupTerminatedAppInfoTrigger will be set to `null`.
https://github.com/apache/kyuubi/blob/d3520ddbcea96ec55c525600126047c44c7adb35/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala#L269

Also shutdown the ExecutorService when KubernetesApplicationOperation stoped. 
## Describe Your Solution 🔧

Shutdown the thread executor service and check the null.

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
@@ -266,6 +266,7 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
 
     if (cleanupTerminatedAppInfoTrigger != null) {
       cleanupTerminatedAppInfoTrigger.invalidateAll()
+      ThreadUtils.shutdown(expireCleanUpTriggerCacheExecutor)
       cleanupTerminatedAppInfoTrigger = null
     }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
@@ -151,6 +151,7 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
       expireCleanUpTriggerCacheExecutor,
       () => {
         try {
+          if (cleanupTerminatedAppInfoTrigger == null) return
           cleanupTerminatedAppInfoTrigger.asMap().asScala.foreach {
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -273,6 +274,16 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
       Utils.tryLogNonFatalError(client.close())
     }
     kubernetesClients.clear()
+
+    if (expireCleanUpTriggerCacheExecutor != null) {
```
wForget, line 278: This may need to be placed before assigning cleanupTerminatedAppInfoTrigger to null.

https://github.com/apache/kyuubi/blob/d3520ddbcea96ec55c525600126047c44c7adb35/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala#L267

turboFei: No, we need to invalidate the cache before shutdown the executor.

turboFei: So that, the remaining terminated application can be cleaned before server stop.

wForget: The non-null check here is not atomic, so NPE may still occur.

https://github.com/apache/kyuubi/blob/78c2a6c0ecf67e0056aef613415e113ba5a3a013/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala#L154

turboFei: How about?
```
         Option(cleanupTerminatedAppInfoTrigger).foreach { trigger =>
            trigger.asMap().asScala.foreach {
              case (key, _) =>
                // do get to trigger cache eviction
                trigger.getIfPresent(key)
            }
          }
```

wForget (approved):
LGTM, thanks.

turboFei:

PR #6786, opened 2024-10-29
Skip repeated checks on convert function

bowenliang123:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TColumnGenerator.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TColumnGenerator.scala
@@ -32,7 +32,8 @@ trait TColumnGenerator[RowT] extends TRowSetColumnGetter[RowT] {
       convertFunc: (RowT, Int) => T = null): (JList[T], ByteBuffer) = {
     val rowSize = rows.length
     val ret = new JArrayList[T](rowSize)
-    val nulls = new JBitSet()
+    val nulls = new JBitSet(rowSize)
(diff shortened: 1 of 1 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TColumnGenerator.scala
```diff
@@ -33,16 +33,18 @@ trait TColumnGenerator[RowT] extends TRowSetColumnGetter[RowT] {
...
         nulls.set(idx, true)
         ret.add(defaultVal)
       } else {
-        val value = Option(convertFunc) match {
-          case Some(f) => f(row, ordinal)
-          case _ => getColumnAs[T](row, ordinal)
+        val value = if (isConvertFuncDefined) {
+          convertFunc(row, ordinal)
+        } else {
+          getColumnAs[T](row, ordinal)
```
pan3793, line 47: let's use a simple way
```
if (convertFunc == null) {
  ...
} else {
  ...
}
```

bowenliang123: Sure.

yaooqinn (approved):

bowenliang123:
Thanks, merged to master (1.11.0) and branch-1.10 (1.10.1).

PR #6789, opened 2024-11-01
Prevent triggering spark job when executing the initialization sql

turboFei:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

#### Related Unit Tests

---

# Checklist 📝

**Be nice. Be informative.**

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
@@ -48,7 +48,7 @@ object KyuubiSparkUtil extends Logging {
         interruptOnCancel = true)
       debug(s"Execute initialization sql: $sql")
       try {
-        spark.sql(sql).isEmpty
+        spark.sql(sql).take(1).isEmpty
       } finally {
         spark.sparkContext.clearJobGroup()
       }


bowenliang123 (approved):
LGTM. Nice catch.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/WithSparkSQLEngine.scala
```diff
@@ -35,7 +35,9 @@ trait WithSparkSQLEngine extends KyuubiFunSuite {
   // Behavior is affected by the initialization SQL: 'SHOW DATABASES'
   // SPARK-35378 (3.2.0) makes it triggers job
   // SPARK-43124 (4.0.0) makes it avoid triggering job
-  protected val initJobId: Int = if (SPARK_ENGINE_RUNTIME_VERSION >= "4.0") 0 else 1
+  // KYUUBI #6789 makes it avoid triggering job
+  // protected val initJobId: Int = if (SPARK_ENGINE_RUNTIME_VERSION >= "4.0") 0 else 1
+  protected val initJobId: Int = 0
```
turboFei, line 40: how do you think about? @pan3793

turboFei: Fixed by @wForget  https://github.com/apache/spark/pull/45397

turboFei: In our use case, I set initial/mini executors to 0 for notebook connections.

If job triggered by initialization SQL, it will need at least 1 executor before the connection ready.

turboFei: And for notebook connections, we will init the spark driver in a temporary queue, and then move it to the final user queues, so it should be fast to make the connection ready even the user queue has no available resources.

But if job triggered during initialization, it will make it slow if the  user queue has no available resources.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/KyuubiSparkUtil.scala
```diff
@@ -48,7 +48,7 @@ object KyuubiSparkUtil extends Logging {
         interruptOnCancel = true)
       debug(s"Execute initialization sql: $sql")
       try {
-        spark.sql(sql).isEmpty
+        spark.sql(sql).take(1).isEmpty
```
ulysses-you, line 51: `spark.sql(sql).take(1)` has two issues:
- it will do an row conversion from `InternalRow` to `Row`
- it does not do column pruning

how about using the same way of the pr https://github.com/apache/spark/pull/45373 to improve it ?

turboFei: Not sure whether it is easy to maintain this workaround in kyuubi end.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -290,4 +290,13 @@ object SparkDatasetHelper extends Logging {
     nodeName == "org.apache.spark.sql.execution.command.ExecutedCommandExec" ||
     nodeName == "org.apache.spark.sql.execution.CommandResultExec"
   }
+
+  /** SPARK-47270: Returns a optimized plan for CommandResult, convert to `LocalRelation`. */
+  def commandResultOptimized[T](dataset: Dataset[T]): Dataset[T] = {
```
pan3793, line 295: I'm -0 on this change, I would consider it a Spark side issue. Spark’s master branch is undergoing a major refactoring, I'm worried about accessing Spark's non-public API in the engine's "core code path".

For users who want to avoid triggering executor launch, they can either patch their spark or set init SQL as something like: `SET spark.app.id`

turboFei: Ok. I will cherry-pick the pr into our managed spark 3.5.

turboFei:
Prefer to backport the spark PR to managed spark

PR #6822, opened 2024-11-25
[BUILD] Using project version property

turboFei:

Easy to maintain the project version.

GA.

No.


pan3793:
I remember that Maven would complain properties reference in that block is not suggested and may not work in the future versions

Issue #6263, opened 2024-04-07
[TASK][MEDIUM] Hive engine supports Apache Hive 4.0.0


pan3793:
### What's the level of this task?

MEDIUM

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/343) and found no similar tasks.

### Mentor

- [ ] I have sufficient expertise on this task, and I volunteer to be a mentor of this task to guide contributors through the task.

### Skill requirements

- Scala
- Hive/MR/Tez
- YARN
- Hadoop Security / Kerberos (optional)

### Background and Goals

Currently, the Kyuubi Hive engine supports
- Cloudera Hive 2.1.1-cdh6
- Apache Hive 2.3
- Apache Hive 3.1

in both local and YARN deploy mode, and both SIMPLE and KERBEROS security mode.

Apache Hive released 4.0.0 on 29 March 2024 https://hive.apache.org/general/downloads/, we'd like to add support for this latest version too.

### Implementation steps

change code/test/GHA workflows and make CI happy.

### Additional context

Introduction of [2024H1 Kyuubi Code Contribution Program](https://github.com/apache/kyuubi/issues/6232)

This may help for testing https://github.com/awesome-kyuubi/hadoop-testing

slfan1989:
@pan3793 Can I assign this issue? Thank you!

slfan1989:
> @slfan1989 thanks, assigned. BTW, I don't know how difficult to support Hive 4, the task level will be adjusted based on the real efforts.

Thank you for your explanation! I will try my best to complete it.

Madhukar525722:
Hi @pan3793 and @slfan1989 are you guys working on this? Can we work together?

PR #6827, opened 2024-11-26
[KYUUBI #6263] Use HiveConf.getConfVars or Hive conf names directly

Madhukar525722:

As initial step to bring hive 4 compatibility in Kyuubi, has addressed the following incompatible changes on ConfVars enum via HIVE-27925.
https://github.com/apache/hive/pull/4919
https://github.com/apache/hive/pull/5107
HiveConf.getConfVars or config names is more robust way to handle this incompatibility.

Build locally and CI

No


pan3793:
@Madhukar525722 Thank you for your effort in working on upgrading Hive 4, it seems you touched both the Spark engine/plugin modules and the Hive engine module, I would say it's IMPOSSIBLE to upgrade the Spark's built-in Hive to 4 at the current stage, all Kyuubi supported Spark versions(Spark 3.2 to 3.5 and 4.0 as of today) use Hive 2.3. So I think you are focusing on upgrading the Hive engine to Hive 4, right?

Madhukar525722:
Thanks @pan3793 for the review. Now, I understood what I missed, I forgot about spark dependency of hive on 2.3.10 and 2.3.9. Yes, I was focusing on hive engine upgrade, I will upgrade hive jdbc.

PR #6829, opened 2024-12-01
Add metrics for batch pending max elapse time

turboFei:

1. add metrics `kyuubi.operartion.batch_pending_max_elapse` for the batch pending max elapse time, which is helpful for batch health monitoring, and we can send alert if the batch pending elapse time too long
2. For `GET /api/v1/batches` api, limit the max time window for listing batches, which is helpful that, we want to reserve more metadata in kyuubi server end, for example: 90 days, but for list batches, we just want to allow user to search the last 7 days. It is optional. And if `create_time` is specified, order by `create_time` instead of `key_id`.
https://github.com/apache/kyuubi/blob/68a6f48da53dd0ad2e20b450a41ca600b8c1e1d2/kyuubi-server/src/main/resources/sql/mysql/metadata-store-schema-1.8.0.mysql.sql#L32

GA.

No.


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -420,15 +421,24 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+      math.max(createTime, metadataSearchWindow.map(System.currentTimeMillis() - _).getOrElse(0L))
     val filter = MetadataFilter(
       sessionType = SessionType.BATCH,
       engineType = batchType,
       username = batchUser,
       state = batchState,
       requestName = batchName,
-      createTime = createTime,
+      createTime = createTimeFilter,
       endTime = endTime)
```
pan3793, line 433: not related to this PR, but `createTime` and `endTime` are confusing, should be `minCreateTime` and `maxEndTime`, and say if they are inclusive or exclusive in the comments, with an explanation for specific values

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -257,14 +257,15 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
...
+      desc: Boolean = false,
+      orderByKeyId: Boolean = true): Seq[Metadata] = {
     val queryBuilder = new StringBuilder
     val params = ListBuffer[Any]()
     queryBuilder.append("SELECT ")
     queryBuilder.append(METADATA_COLUMNS)
     queryBuilder.append(s" FROM $METADATA_TABLE")
     queryBuilder.append(s" ${assembleWhereClause(filter, params)}")
-    queryBuilder.append(" ORDER BY key_id ")
+    queryBuilder.append(" ORDER BY ").append(if (orderByKeyId) "key_id " else "create_time ")
```
pan3793, line 268: have you tried without `ORDER BY` clause? if the result ordering is stable (for example, ordered by pk), it might be the fastest approach

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2023,6 +2023,14 @@ object KyuubiConf {
       .intConf
       .createWithDefault(65536)
 
+  val METADATA_SEARCH_WINDOW: OptionalConfigEntry[Long] =
+    buildConf("kyuubi.metadata.search.window")
+      .doc("The time window to search the metadata from metadata store.")
```
pan3793, line 2028: from a user perspective, I have no idea what I should do with this conf after reading the docs

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
pan3793: update: the predication name might be more intuitive if we use `greatThanXXX` `lessThanOrEqualsXXX`

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -257,14 +257,15 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
       filter: MetadataFilter,
       from: Int,
       size: Int,
-      desc: Boolean = false): Seq[Metadata] = {
+      desc: Boolean = false,
+      orderByKeyId: Boolean = true): Seq[Metadata] = {
```
pan3793, line 261: the parameters are evolving to Spark's `SortOrder`

```
case class SortOrder(
    child: Expression,
    direction: SortDirection,
    nullOrdering: NullOrdering,
    sameOrderExpressions: Seq[Expression])
```

how about using some SQL terminologies to redesign the method's signature?
```
override def getMetadataList(
      filter: MetadataFilter,
      offset: Int,
      limit: Int,
      orderBy: Option[String],
      direction: String = "ASC"): Seq[Metadata]
```

turboFei:
thanks for the comments, will address it today

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
turboFei: Will address it in another PR

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
turboFei: It is not guaranteed.

image

turboFei: But for small dataset, it looks good. I think we can have a try.
image

turboFei: And the UT can not pass without `order by`.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
turboFei: updated

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2023,6 +2023,16 @@ object KyuubiConf {
       .intConf
       .createWithDefault(65536)
 
+  val METADATA_SEARCH_WINDOW: OptionalConfigEntry[Long] =
+    buildConf("kyuubi.metadata.search.window")
+      .doc("The time window to restrict user queries to metadata within a specific period. " +
+        "For example, if the window is set to P7D, only metadata from the past 7 days can be " +
+        "queried. If not set, it allows searching all metadata information in the metadata store.")
```
pan3793, line 2030: the doc should declare the affected user-faced interfaces, for example, it only affects xxx API, and with suggestions like, you may want to set this to XXX when XXX, also should include the side-effects if any.

turboFei: thanks, addressed

pan3793:
Thanks, merged to master/1.10

Issue #6830, opened 2024-12-02
[Improvement] Allow indicate advisory shuffle partition size when merge small files


yabola:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

when merging small files(set `spark.sql.optimizer.insertRepartitionBeforeWrite.enabled`=true) , the default session advisory partition size (64MB) will be used as target. This default value can still lead to small files because the written data can be compressed nicely using columnar file formats (usually 1/4 or smaller of the shuffle exchange size, the result is often around 15MB).

Spark now support configuring the rebalance expression advisory size in https://github.com/apache/spark/pull/40421 . So we can have a configuration that can configure the merge size separately.

### How should we improve?

add one configuration to control the size separately

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #6831, opened 2024-12-02
[KYUUBI #6830] Allow indicate advisory shuffle partition size when me…

yabola:
when merging small files(set `spark.sql.optimizer.insertRepartitionBeforeWrite.enabled`=true) , the default session advisory partition size (64MB) will be used as target. This default value can still lead to small files because the written data can be compressed nicely using columnar file formats (usually 1/4 or smaller of the shuffle exchange size, the result is often around 15MB).

Spark now support configuring the rebalance expression advisory size in https://github.com/apache/spark/pull/40421 . So we can have a configuration that can configure the merge size separately.

no

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
+++ b/extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
@@ -103,6 +103,14 @@ object KyuubiSQLConf {
       .booleanConf
       .createWithDefault(false)
 
+  val REBALANCE_ADVISORY_PARTITION_SIZE_IN_BYTES =
+    buildConf("spark.sql.optimizer.rebalanceAdvisoryPartitionSize")
+      .doc("The advisory size in bytes of the shuffle partition during merge small files. " +
+        s"It takes effect when Spark coalesces small shuffle partitions or splits skewed " +
+        s"shuffle partition.")
+      .version("1.10.1")
+      .fallbackConf(ADVISORY_PARTITION_SIZE_IN_BYTES)
+
   val WATCHDOG_MAX_PARTITIONS =
     buildConf("spark.sql.watchdog.maxPartitions")
       .doc("Set the max partition number when spark scans a data source. " +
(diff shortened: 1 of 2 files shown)


pan3793:
the use case already covered by 
```
spark.sql.optimizer.finalStageConfigIsolation.enabled=true
spark.sql.finalStage.adaptive.advisoryPartitionSizeInBytes=512m
```
https://kyuubi.readthedocs.io/en/master/extensions/engines/spark/rules.html#additional-configurations

yabola:
@pan3793 Yes, I hadn't noticed before. 
Another question is, can we add  compression ratios to control different `advisoryPartitionSizeInBytes`  for different file format (parquet, orc,avro, text, etc.) . It can make it more automated. Iceberg has similar functionality.

pan3793:
@yabola the real compression ratio is affected by several things, e.g. the input shuffle compression codec and level, the data itself, the written format as you mentioned, the write compression codec and level, I don't think we can do that automatically and correctly

yabola:
@pan3793 emmm, but in the scenario of merging small files, we only need to consider the shuffle data size (this rule is only for shuffle data to file, doesn't matter what the data source is).
Due to the row storage and estimation method of shuffle data, there is still a significant difference between the shuffle size and the actual written file size, especially for Parquet , usually less than 1/3 of the size of the shuffle data.
iceberg Implementation:
https://github.com/apache/iceberg/blob/38c8daa4eae8a75ab46571f1efce1609100f53dd/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkCompressionUtil.java#L60-L69

pan3793:
> in the scenario of merging small files, we only need to consider the shuffle data size (this rule is only for shuffle data to file, doesn't matter what the data source is).

I overlook this, you are right.

I read the the Iceberg's code and understand the how it works, I am a little bit pessimistic to adopt it, because the real compression ratio is affected by data itself, the assumption is not always true, when the estimate deviates significantly from the real value, it's hard to explain to users how this happen. Files written by a Spark job are likely read by other Spark jobs, and the data will be covnerted to Spark InternelRow layout(same as Shuffle) again, have the compression ratio been considered on the read code path too?

I think it would be nice if we add a section in docs to explain how `advisoryPartitionSizeInBytes` affects the written file size, with some tune guidance.

yabola:
@pan3793 I will close this, if anyone finds it helpful, can revisit this ~
In addition, we may consider the situation to be simpler, because the recommended size of parquet (or other format) is generally 128MB, and the files written are almost certainly smaller than the shuffle size, which is enough, because this is a general rule.

PR #6836, opened 2024-12-04
Ship `kafka-clients` in binary distribution tarball without compression libs

pan3793:

I'd like to include `kafka-clients` in the Kyuubi binary distribution tarball to enable the out-of-box support for sinking Kyuubi events to Kafka.

- Kafka is an important component in modern data platforms, and is a defacto message queue implementation, especially in the big data domain
- `kafka-clients` is released under Apache License V2, has no legal issue
- `kafka-clients` is quite a light lib, has no third-party deps except for `slf4j-api` and a few optional compression libs
- `kafka-clients` uses "none" compression as default, in practice, "gzip"(delegate to JDK gzip algorithm, no additional libs are required) already performs well for non-extreme cases

Additionally, LOG4J2 has a built-in `KafkaAppender` that supports sinking logging to Kafka, which also requires `kafka-clients` in the classpath, I have some initial ideas to forward both Kyuubi server's and engine bootstrap processes log to Kafka in a structured format, will send another PR to add docs to guide users in configuring that.

Review

No.

### DIFF
--- a/LICENSE-binary
+++ b/LICENSE-binary
@@ -306,8 +306,6 @@ io.vertx:vertx-grpc
 com.squareup.retrofit2:retrofit
 com.squareup.okhttp3:okhttp
 org.apache.kafka:kafka-clients
-org.lz4:lz4-java
-org.xerial.snappy:snappy-java
 org.xerial:sqlite-jdbc
 
 BSD
@@ -319,7 +317,6 @@ jline:jline
 com.thoughtworks.paranamer:paranamer
 com.google.protobuf:protobuf-java-util
 com.google.protobuf:protobuf-java
-com.github.luben:zstd-jni
 org.postgresql:postgresql
 
 Eclipse Distribution License - v 1.0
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -1053,7 +1053,20 @@
                 <groupId>org.apache.kafka</groupId>
                 <artifactId>kafka-clients</artifactId>
                 <version>${kafka.version}</version>
-                <optional>true</optional>
```
pan3793, line 1056: seems the `build/dependency.sh` does not correctly handle `optional`

LICENSE-binary
```diff
@@ -306,8 +306,6 @@ io.vertx:vertx-grpc
 com.squareup.retrofit2:retrofit
 com.squareup.okhttp3:okhttp
 org.apache.kafka:kafka-clients
-org.lz4:lz4-java
-org.xerial.snappy:snappy-java
```
pan3793, line 310: the LICENSE/NOTICE should match the real content of artifacts

turboFei (approved):

bowenliang123 (approved):

pan3793:
Thanks, merged to master/1.10

PR #6837, opened 2024-12-04
[INFRA] Disable K8s CI image GHA cache

pan3793:

The GHA cache `engine-archives` seems frequently to be evicted because of exceeding the project cache capacity(10GiB), and I found the K8s IT job produces fresh large cache objects frequently, we should disable that to let `engine-archives` survive, to avoid downloading from the fragile `archive.apache.org` each time.

```
Connect to archive.apache.org:443 [archive.apache.org/65.108.204.189, archive.apache.org/2a01:4f9:1a:a084:0:0:0:2] failed: Network is unreachable (connect failed)
```

Xnip2024-12-04_17-36-03

PS: I deleted bunches of `buildkit-blob-*` manually to save `engine-archives`

Monitor GHA after merging.

No.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -420,6 +420,8 @@ jobs:
       # https://github.com/docker/build-push-action
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
+        with:
+          cache-binary: false
       - name: Build Kyuubi Docker Image
         uses: docker/build-push-action@v6
         with:


pan3793:
The K8s CI has room to improve, will find time to refactor it later.

bowenliang123 (approved):

pan3793:
Thanks, merged to master/1.10/1.9, I will monitor the CI in the next few days.

Issue #6726, opened 2024-10-14
[FEATURE] support trino stage progress


taylor12805:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

kyuubi now have interface to query operation progress of spark sql job. trino jdbc also have function to query statement progress while job is running, we should also support query progress of trino statement.

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #6841, opened 2024-12-08
[KYUUBI #6726][FOLLOWUP] Fix compilation on scala-2.13

naive-zhang:
# Description

Currently, Kyuubi supports Trino progress in scala-2.12, but got some error in scala-2.13.
* Make Trino progress available in scala-2.13

## Test Plan 🧪

---

# Checklist 📝

**Be nice. Be informative.**


pan3793:
Can we fix it in place? e.g. by adding `.Seq` to make the collection conversion work in both Scala 2.12 and 2.13

externals/kyuubi-trino-engine/pom.xml
```diff
@@ -98,6 +98,36 @@
...
+                        </goals>
+                        <phase>generate-test-sources</phase>
+                        <configuration>
+                            <sources>
+                                <source>src/test/scala-${scala.binary.version}</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
```
pan3793, line 130:
```suggestion
```

PR #6859, opened 2024-12-24
Exclude log4j12 from hive engine module classpath

pan3793:

Kyuubi uses log4j2 as the logging framework, while I found that the Hive SQL engine module still polls log4j 1.2 to the classpath unexpectedly, we should exclude it to avoid potential issues.

```
build/mvn dependency:tree -pl :kyuubi-hive-sql-engine_2.12
```
```
...
[INFO] +- org.apache.hive:hive-service:jar:3.1.3:provided
[INFO] |  +- org.apache.hive:hive-exec:jar:3.1.3:provided
[INFO] |  |  +- org.apache.zookeeper:zookeeper:jar:3.4.6:provided
[INFO] |  |  |  +- log4j:log4j:jar:1.2.16:provided
...
```

Checks `build/mvn dependency:tree | grep 'log4j:log4j:jar:1.2'` returns nothing and pass GHA.

No.

### DIFF
--- a/externals/kyuubi-hive-sql-engine/pom.xml
+++ b/externals/kyuubi-hive-sql-engine/pom.xml
@@ -93,6 +93,10 @@
                     <groupId>org.apache.hadoop</groupId>
                     <artifactId>hadoop-yarn-registry</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>log4j</groupId>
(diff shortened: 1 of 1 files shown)


cxzl25 (approved):

turboFei (approved):

pan3793:
Thanks, merged to master/1.10.2/1.9.4

PR #6864, opened 2024-12-25
Support to return prometheus metrics with instance label

turboFei:

For my use case, the instances are not human readable, so I prefer to return the FQDN.
image

Integration testing.
```
(base) ➜  dist git:(prometheus_label_2) cat conf/kyuubi-defaults.conf
kyuubi.metrics.prometheus.metrics.instance.enabled=true
kyuubi.zookeeper.embedded.client.port.address=localhost
kyuubi.frontend.bind.host=localhost
```

image

No.


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConf.scala
```diff
@@ -80,6 +80,13 @@ object MetricsConf {
     .checkValue(path => path.startsWith("/"), "Context path must start with '/'")
     .createWithDefault("/metrics")
 
+  val METRICS_PROMETHEUS_METRICS_INSTANCE_ENABLED: ConfigEntry[Boolean] =
+    buildConf("kyuubi.metrics.prometheus.metrics.instance.enabled")
```
pan3793, line 84: maybe
```suggestion
    buildConf("kyuubi.metrics.prometheus.labels.instance.enabled")
```
?

turboFei: Sounds good.

kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/PrometheusReporterService.scala
```diff
@@ -100,4 +112,46 @@ class PrometheusReporterService(registry: MetricRegistry)
...
+    val metricsSnapshotWriter = new java.io.StringWriter
+    val contentType = TextFormat.chooseContentType(null)
+    TextFormat.writeFormat(contentType, metricsSnapshotWriter, bridgeRegistry.metricFamilySamples())
+    val labelStr = labelString(labels)
+    metricsSnapshotWriter.toString.split("\n").map { line =>
+      if (line.startsWith("#")) {
+        line
+      } else {
+        val Array(metrics, value) = line.split("\\s+", 2)
+        s"""$metrics${labelStr} $value"""
```
pan3793, line 149: a little bit hacky, but given the feature is disabled by default, let's merge it in first.

pan3793 (approved):

kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/PrometheusReporterService.scala
turboFei: changed to

```
       line.split("\\s+", 2) match {
          case Array(metrics, rest) => s"""$metrics${labelStr} $rest"""
          case _ => line
        }
```

pan3793:
Thanks, merged to master/1.10

PR #6866, opened 2024-12-25
Add metrics for SSL keystore expiration time

turboFei:

Add metrics for SSL keystore expiration, then we can add alert if the keystore will expire in 1 month.

Integration testing.
image

No.

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
@@ -37,6 +37,8 @@ object MetricsConstants {
   final private val THRIFT_BINARY_CONN = KYUUBI + "thrift.binary.connection."
   final private val REST_CONN = KYUUBI + "rest.connection."
 
+  final val THRIFT_SSL_CERT_EXPIRATION = KYUUBI + "thrift.ssl.cert.expiration"
+
   final val CONN_OPEN: String = CONN + "opened"
   final val CONN_FAIL: String = CONN + "failed"
   final val CONN_TOTAL: String = CONN + "total"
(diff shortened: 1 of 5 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsConstants.scala
```diff
@@ -37,6 +37,8 @@ object MetricsConstants {
   final private val THRIFT_BINARY_CONN = KYUUBI + "thrift.binary.connection."
   final private val REST_CONN = KYUUBI + "rest.connection."
 
+  final val SSL_CERT_EXPIRATION = KYUUBI + "ssl.cert.expiration"
```
pan3793, line 40: will ssl cert be used in other places? do we need an additional namespace?

turboFei: https://github.com/apache/kyuubi/blob/fd539130764b466778e88144390668a59154b0eb/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala#L768-L774

turboFei: Currently, the keystore is used for thrift binary and http SSL.

turboFei: Should we add `thrift` ?
```
kyuubi.thrift.ssl.cert.expiration
```

turboFei: Currently, the keystore is not used for RESTful yet.

We use the certificate of Loadbalancer for http SSL instead of maintain the keystore for every kyuubi instance.

pan3793: > Should we add `thrift` ?
> 
> ```
> kyuubi.thrift.ssl.cert.expiration
> ```

sgtm

pan3793:
thanks, merged to master/1.10

PR #6867, opened 2024-12-26
Refactor the grafana dashboard template

turboFei:

The original dashboard dashboard template does not work for my use case:

In this PR:
1. I add the `baseFilter` variable, default value is `job=~"kyuubi"`.
2. apply the `baseFilter` for all expressions
3. add `baseLegend` for the legendFormat
4. Fix hardcode typos 
5. remove the `sum` from the expressions for JVM

image

No.

### DIFF
--- a/grafana/dashboard-template.json
+++ b/grafana/dashboard-template.json
@@ -136,7 +136,7 @@
           "expr": "kyuubi_jvm_uptime{instance=~\"$instance\",$baseFilter}",
           "format": "time_series",
           "instant": true,
-          "legendFormat": "{{instance}}",
+          "legendFormat": "${baseLegend}",
           "range": false,
           "refId": "A"
         }
@@ -231,7 +231,7 @@
           "includeNullMetadata": true,
           "instant": false,
           "interval": "",
-          "legendFormat": "{{instance}}-batch",
+          "legendFormat": "${baseLegend}-batch",
           "range": true,
           "refId": "A",
           "useBackend": false
@@ -245,7 +245,7 @@
           "expr": "kyuubi_connection_total_INTERACTIVE{instance=~\"$instance\",$baseFilter}",
           "hide": false,
           "instant": false,
-          "legendFormat": "{{instance}}-interactive",
+          "legendFormat": "${baseLegend}-interactive",
           "range": true,
           "refId": "B"
         }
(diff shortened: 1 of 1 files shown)


grafana/dashboard-template.json
```diff
@@ -113,7 +147,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "PBFA97CFB590B2093"
```
turboFei, line 116: hardcode before

grafana/dashboard-template.json
```diff
@@ -191,27 +225,27 @@
...
-          "legendFormat": "{{instance}}-batch",
+          "legendFormat": "${baseLegend}-batch",
           "range": true,
           "refId": "A",
           "useBackend": false
         },
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "PBFA97CFB590B2093"
```
turboFei, line 208: hardcode before

grafana/dashboard-template.json
```diff
@@ -684,27 +718,27 @@
...
           "hide": false,
-          "legendFormat": "Eden Space Committed",
+          "legendFormat": "${baseLegend} Eden Space Committed",
           "range": true,
           "refId": "A"
         },
         {
           "datasource": "${DS_PROMETHEUS}",
           "editorMode": "code",
-          "expr": "sum(kyuubi_memory_usage_pools_Par_Eden_Space_max)",
```
turboFei, line 705: not sure whether the `sum` is necessary, but I think we can remove such kind of metrics.

Because seems it does not exists for my kyuubi server + JDK17.

turboFei: I think only reserve the Memery Usage  + GC count is enough.

cc @pan3793

pan3793:
verified locally, lgtm.

nit, can we put `$baseFilter` at first?
```patch
- {instance=~\"$instance\",$baseFilter}
+ {$baseFilter,instance=~\"$instance\"}
```

turboFei:
> verified locally, lgtm.
> 
> nit, can we put `$baseFilter` at first?
> 
> ```diff
> - {instance=~\"$instance\",$baseFilter}
> + {$baseFilter,instance=~\"$instance\"}
> ```


Updated.

yaooqinn (approved):

pan3793 (approved):

pan3793:
thanks, merged to master

PR #6865, opened 2024-12-25
[TEST] Set `kyuubi.zookeeper.embedded.client.port.address` to `localhost` for testing

turboFei:
### Why are the changes needed?

Similar with `kyuubi.frontend.bind.host`, it is better to set `kyuubi.zookeeper.embedded.client.port.address` to `localhost` for testing as well.

### How was this patch tested?

GA.

### Was this patch authored or co-authored using generative AI tooling?

No.

PR #6870, opened 2024-12-26
Add more metrics for grafana dashboard

turboFei:

Add more metrics for monitoring.

1. Overall
image2. Connection(repeat for connectionType - `thrift_binary_connection,rest_connection,thrift_http_connection,metadata_request`)
For example:
image
image

image

3. Operation(repeat for operationType - `ExecuteStatement,BatchJobSubmission,LaunchEngine`)

image
image

4. Extra
image

5. Move the JVM statistics down.

No.


pan3793:
Thanks for making this change, I apply it in Grafana v9.0.8 and it works well.

One suggestion, can we use apply `increase` for `MetadataRequest` panel?
image

pan3793:
btw, I think we should keep the 1m trend interval, "x times per minute" is a very common concept

pan3793:
test failure looks related to #6865

pan3793:
Some JVM panels seem does not work for Java17, which can be fixed later independently.

turboFei:
> Some JVM panels seem does not work for Java17, which can be fixed later independently.

image




> btw, I think we should keep the 1m trend interval, "x times per minute" is a very common concept

Ok

turboFei:
Also add total connections in the overview.
image


@pan3793 It should be almost done.

I will add the `keystore expiration` pannel in next PR.

turboFei:
GC Count for increase instead of total

image

turboFei:
Added keystore expiration:
image

turboFei:
If you select  `ALL` for the `connType` or `opType, it will repeat the rows.
image

pan3793 (approved):
Verified locally, thanks!

pan3793:
Merged to master

PR #6872, opened 2024-12-26
[KYUUBI #6866][FOLLOWUP] Prevent register gauge conflicts if both thrift binary SSL and thrift http SSL enabled

turboFei:

Followup for https://github.com/apache/kyuubi/pull/6866
It would throw exception if both thrift binary SSL and thrift http SSL enabled 

No.


pan3793 (approved):

pan3793:
https://github.com/apache/kyuubi/pull/6866 goes 1.10.2 too, but this goes olny 1.11.0?

turboFei:
Thanks, merged to main and 1.10.2

Issue #6843, opened 2024-12-09
[Bug] “query-timeout-thread“  thread leak 


ASiegeLion:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

![image](https://github.com/user-attachments/assets/677444fa-51f3-4389-8f15-b1caf067f173)
If the session manager's ThreadPoolExecutor refuses to execute asyncOperation, query-timeout-thread will not be shut down.

### Affects Version(s)

master

### Are you willing to submit PR?

- [X] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #6873, opened 2024-12-27
[KYUUBI #6843] [FOLLOWUP] Fix 'query-timeout-thread' thread leak

lsm1:

If the session manager's ThreadPoolExecutor refuses to execute the asyncOperation, then we need to shut down the query-timeout-thread in the catch block. This should also be done in JDBC and the CHAT engine.


cxzl25 (approved):

lsm1:
Thanks, Merged to master, branch-1.10, branch-1.9

PR #6883, opened 2025-01-09
Using withOauthTokenProvider instead of withOauthToken to support token refresh 

turboFei:

Address comments: https://github.com/apache/kyuubi/discussions/6877#discussioncomment-11743818

> I guess this is a Kyuubi implementation issue, we just read the content from the kyuubi.kubernetes.authenticate.oauthTokenFile and call ConfigBuilder.withOauthToken, I guess this approach does not support token refresh...

Existing GA.

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/util/KubernetesUtils.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/util/KubernetesUtils.scala
@@ -78,6 +78,8 @@ object KubernetesUtils extends Logging {
       .withOption(oauthTokenValue) { (token, configBuilder) =>
         configBuilder.withOauthToken(token)
       }.withOption(oauthTokenFile) { (file, configBuilder) =>
+        // Prior to Kubernetes 1.24, the default token never expired.
+        // In newer versions, it expires after 1 hour by defaults.
         configBuilder.withOauthTokenProvider(new OAuthTokenProvider {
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/util/KubernetesUtils.scala
```diff
@@ -78,7 +78,9 @@ object KubernetesUtils extends Logging {
       .withOption(oauthTokenValue) { (token, configBuilder) =>
         configBuilder.withOauthToken(token)
       }.withOption(oauthTokenFile) { (file, configBuilder) =>
-        configBuilder.withOauthToken(Files.asCharSource(file, Charsets.UTF_8).read())
+        configBuilder.withOauthTokenProvider(new OAuthTokenProvider {
```
pan3793, line 81: seems before k8s 1.24, the default token has no expiration time, but the newer version is 1h, let's leave some comments here, including links about K8s behavior change

turboFei: done

pan3793 (approved):

pan3793:
thanks, merged to master/1.10.2/1.9.4

turboFei:
1.9.4 has not been released.

pan3793:
@venkatmedida Unfortunately, there is no released version that ships this change, maybe it's time to have a discussion for the next release ...

PR #6891, opened 2025-01-15
Fix get existing gauge issue

turboFei:

For the `com.codahale.metrics.MetricRegistry::gauge`.
It `getOrAdd` the gauge with name.
```
    public <T extends Gauge> T gauge(String name) {
        return (Gauge)this.getOrAdd(name, MetricRegistry.MetricBuilder.GAUGES);
    }
```

So we have to get all the gauges to check whether the gauge exists.

UT.

No.

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
@@ -20,6 +20,8 @@ package org.apache.kyuubi.metrics
 import java.lang.management.ManagementFactory
 import java.util.concurrent.TimeUnit
 
+import scala.collection.JavaConverters._
+
 import com.codahale.metrics.{Gauge, MetricRegistry, Snapshot}
 import com.codahale.metrics.jvm._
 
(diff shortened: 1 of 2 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/MetricsSystem.scala
```diff
@@ -58,8 +60,8 @@ class MetricsSystem extends CompositeService("MetricsSystem") {
     meter.mark(value)
   }
 
-  def getGauge[T](name: String): Option[Gauge[T]] = {
-    Option(registry.gauge(name))
+  def getGauge(name: String): Option[Gauge[_]] = {
+    registry.getGauges((metricsName, _) => { metricsName == name }).asScala.map(_._2).headOption
```
pan3793, line 64: semantic is correct, but performance is poor, is it possible to have an O(1) approach? if impossible and perf is acceptable, please leave a brief comment to explain that.

pan3793:
this looks like an independent issue, though it affects #6866, please update the title with a new ticket id or leave it to let the merge script to assign automatically.

turboFei:
> this looks like an independent issue, though it affects #6866, please update the title with a new ticket id or leave it to let the merge script to assign automatically.

done

turboFei:
thanks, merged to master and 1.10.2

PR #6893, opened 2025-01-16
Upgrading commons-collections to commons-collections4

avishnus:
commons-collections has reached EOL. Need to migrate to commons-collections4

No


pan3793:
This won't work.

Kyuubi itself does not use `commons-collections` but the supported engines like Hive do.

pan3793:
close as purposed change is invalid

PR #6899, opened 2025-01-21
Update sqlalchemy_presto.py add try_cast

pinohans:
Update sqlalchemy_presto.py add try_cast

New API.
When users encounter dirty data when using `CAST`, `TRY_CAST` becomes more useful.

Local test with sqlalchemy.

No


pan3793 (changes requested):
Changes without reasonable "Why" won't be considered.

Issue #6901, opened 2025-01-24
[Bug] [authz] Default policy not defined


developster:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When Apache Ranger cannot be reached and the policy cache is empty no policy is enforced and everything is permitted.

### Affects Version(s)

v1.10.1

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #6902, opened 2025-01-24
[KYUUBI #6901] Default policy for spark

developster:
Added a service definition for spark which in turn enables the creation of a default policy for the spark service.
Default policy will block access until another policy is downloaded from Apache Ranger.

Tested manually.
Configure Kyuubi Authz plugin. Do not start Apache Ranger, it must not be reachable.
Make sure that policy cache is empty.
Start Kyuubi engine and try to query any tables. The default policy should not allow any access.
Previously the access was not restricted because there wasn't a default policy defined.

No.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/resources/service-defs/ranger-servicedef-spark.json
+++ b/extensions/spark/kyuubi-spark-authz/src/main/resources/service-defs/ranger-servicedef-spark.json
@@ -1,0 +1,464 @@
+{
+	"id":3,
+	"name": "hive",
+	"displayName": "Hadoop SQL",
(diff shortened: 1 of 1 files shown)


extensions/spark/kyuubi-spark-authz/src/main/resources/service-defs/ranger-servicedef-spark.json
```diff
@@ -0,0 +1,464 @@
+{
+	"id":3,
+	"name": "hive",
```
yaooqinn, line 3: Would it conflict with the hive def?

yaooqinn (approved):

Issue #6905, opened 2025-01-29
[Improvement] [pyhive] HTTP connections always connect to the 'default' database


developster:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

HiveHTTPDialect is ignoring the database name from the url and thus the 'default' database is used even if another database was specified. Updated the HTTP dialect to use the database specified in the url.

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #6906, opened 2025-01-29
[KYUUBI #6905] PyHive HTTP/HTTPS dialect to use the database name from url

developster:
HTTP dialect ignores the database specified in the URL and uses the "default" instead.

Tested manually.

No.

### DIFF
--- a/python/pyhive/sqlalchemy_hive.py
+++ b/python/pyhive/sqlalchemy_hive.py
@@ -421,6 +421,7 @@ class HiveHTTPDialect(HiveDialect):
             "scheme": self.scheme,
             "username": url.username or None,
             "password": url.password or None,
+            'database': url.database or 'default',
         }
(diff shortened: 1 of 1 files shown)


python/pyhive/sqlalchemy_hive.py
```diff
@@ -421,6 +421,7 @@ def create_connect_args(self, url):
             "scheme": self.scheme,
             "username": url.username or None,
             "password": url.password or None,
+            'database': url.database or 'default',
```
pan3793, line 424:
```suggestion
            "database": url.database or "default",
```

pan3793 (approved):

cxzl25 (approved):

yaooqinn (approved):

pan3793:
Thanks, merged to master

PR #6915, opened 2025-02-13
Fix ClickHouse integration tests

pan3793:

I observed ClickHouse integration test failure in GHA, after some investigation, the root cause is https://github.com/testcontainers/testcontainers-java/pull/9942

```
/entrypoint.sh: neither CLICKHOUSE_USER nor CLICKHOUSE_PASSWORD is set, disabling network access for user 'default'
```

In short, the recent ClickHouse docker image does not allow the `default` user to connect without a password, unfortunately, `testcontainers-scala-clickhosue` does not expose API to set CLICKHOSUE_USER and CLICKHOUSE_PASSWORD, as a workaround, I pin `clickhouse-server:24.3.15`(the latest version has no such restriction) until a fixed version of Testcontainers available.

This PR also switches the `clickhouse-jdbc`'s classifier from `http` to `shaded`, the reason is, `http` does not ship ApacheHttpClient5, previously, it happened to work because `iceberg-runtime-spark3.5_2.12` packaged un-relocated ApacheHttpClient5 classes, but it gets fixed in Iceberg 1.8.0, then `clickhouse-jdbc:http` stop working.
```
java.lang.NoClassDefFoundError: org/apache/hc/core5/http/HttpRequest
```

Additionally, this PR bumps `clickhouse-jdbc` from 0.6.0 to 0.6.5.

Pass GHA.

No.

### DIFF
--- a/externals/kyuubi-jdbc-engine/pom.xml
+++ b/externals/kyuubi-jdbc-engine/pom.xml
@@ -112,7 +112,7 @@
         <dependency>
             <groupId>com.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
-            <classifier>http</classifier>
+            <classifier>shaded</classifier>
             <scope>test</scope>
         </dependency>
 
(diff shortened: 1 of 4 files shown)


pan3793:
Pin `clickhouse-server:24.3.15` instead of downgrading to 23.8

lsm1 (approved):

pan3793:
Thanks, merged to master/1.10

PR #6945, opened 2025-03-01
[KYUUBI #6941] Test Add new Column for paimon

davidyuan1223:

Cuurently, ranger check test case missing check paimon add new column command, add it.
#6941 

Test Add New Column for paimin with ranger

No


davidyuan1223:
@yaooqinn @pan3793 TBR

PR #6953, opened 2025-03-01
[KYUUBI #6948] Test Changing Column Comment

davidyuan1223:

Range check Test case missing paimon changing column comment, add the test case
#6948

Test Ranger check with paimon changing column comment command

No


yaooqinn (approved):

pan3793:
Thanks, merged to master

PR #6956, opened 2025-03-01
[KYUUBI #6951] Test changing column type

davidyuan1223:

Ranger check test case missing paimon changing column type command, add the test case
#6951

Test ranger check paimon changing column type command

No


yaooqinn (approved):

pan3793:
Merged to master

PR #6974, opened 2025-03-08
[KYUUBI #6973][AUTHZ] Support Paimon DELETE FROM / UPDATE / MERGE INTO commands

davidyuan1223:

Support ranger check with paimon Update & Delete & MergeInti Table Command

Test Paimon Update & Delete Table Command with ranger check
#6973 

No


extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/PaimonCatalogRangerSparkExtensionSuite.scala
```diff
@@ -31,6 +31,8 @@ import org.apache.kyuubi.util.AssertionUtils._
 @PaimonTest
 class PaimonCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
   override protected val catalogImpl: String = "hive"
+  override protected val sqlExtensions: String =
+    "org.apache.paimon.spark.extensions.PaimonSparkSessionExtensions"
```
pan3793, line 35: I remember this would break the Scala 2.13 UT, if so, you may want to enable this extension conditionally, or ask the Paimon community to handle Scala 2.13 properly.

davidyuan1223: > I remember this would break the Scala 2.13 UT, if so, you may want to enable this extension conditionally, or ask the Paimon community to handle Scala 2.13 properly.

you are right, i found paimon-spark profile has some questions, it only support spark3 with scala2.12, then the spark4 with scala2.13, i will jump the test case when scala version is 2.13, then i will submit an issue in paimon community

pan3793 (approved):

pan3793:
Thanks, merged to master

Issue #5376, opened 2023-10-08
[TASK][MEDIUM] AuthZ plugin supports all SparkOperation


pan3793:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before creating

- [X] I have searched in the [task list](https://github.com/orgs/apache/projects/296) and found no similar tasks.

### Mentor

- [ ] I have sufficient knowledge and experience of this task, and I volunteer to be the mentor of this task to guide contributors to complete the task.

### Skill requirements

- Basic knowledge of Scala programming language
- Familiar with Spark Catalyst

### Background and Goals

Presently, the Authorization (AuthZ) plugin exclusively offers access control for the ExecuteStatement operation. Enhancing its functionality to encompass all operations, such as GetCatalogs and GetTables.

### Implementation steps

N/A

### Additional context

Original reporter is @cf

pan3793:
@cfmcgrady this task is missing "implementation steps" before adding into https://github.com/orgs/apache/projects/296

phionex0571:
@pan3793 I think this issue is a little challenging for me but I really want to have a try, and I will most probably finish it in two months. can you aisgn it to me? thx.

zxl-333:
> [@pan3793](https://github.com/pan3793) I think this issue is a little challenging for me but I really want to have a try, and I will most probably finish it in two months. can you aisgn it to me? thx.

Is there any progress on this issue

wForget:
> > [@pan3793](https://github.com/pan3793) I think this issue is a little challenging for me but I really want to have a try, and I will most probably finish it in two months. can you aisgn it to me? thx.
> 
> Is there any progress on this issue

It looks like there is no further progress, do you want to take it?

wForget:
In addition, I have an idea for implementation. 

Steps:
+ Define RunnableCommand in Engine, like: KyuubiGetTablesCommand (keep logic of current engine)
+ Replace KyuubiGetTablesCommand in Authz plugin with authz processing logic (ref: RuleReplaceShowObjectCommands)
+ The logic of GetTables Operator in Engine: val plan = new KyuubiGetTablesCommand; spark.sessionState.executePlan(plan).executedPlan.executeCollect()

cc @pan3793

zxl-333:
[...]

PR #6978, opened 2025-03-12
[KYUUBI #5376] Change the GetSchemas and GetTables functions on kyuubi to use DDL to obtain information

zxl-333:
Change the GetSchemas and GetTables functions on kyuubi to use DDL to obtain information, thereby resolving the issue of DBeaver bypassing Ranger's permission restrictions.

The ranger plugin has been enabled to control permissions. When using DBeaver to connect to Kyuubi, users without permission for the database and tables can still see the database and tables on the left side.
After applying this modification, the access to the database tables can be controlled through the Ranger plugin. When users use DBeaver, they can only see the database tables for which they have permissions in the "Tables" column on the left side.

Use the ranger plugin of Kyuubi for authorization, and use DBeaver to connect to Kyuubi for testing.

No


pan3793:
@zxl-333 the test failure is related, could you please fix them first?

zxl-333:
> @zxl-333 the test failure is related, could you please fix them first?

ok

PR #6980, opened 2025-03-12
[KYUUBI #6979] Support check paimon system producers

davidyuan1223:

Cuurently, ranger check missing paimom system producers command, need to support these command
1. create_tag
2. delete_tag
3. rollback

#6979

PS: There has a question about paimon, paimon'sparkCatalog need the currentCatalog Env is the paimon's catalog, use default spark_catalog will throw exception, maybe we should add this hint to the documentation.
such as
If you wanna support producers check with paimon, you need use sql `use $paimon_catalog` to ensure the session currentCatalog is paimon_catalog

PS: paimon-spark-3.3:0.8.2 has some compaitable question, suggest upgrade the paimon version

producers test cases
1. create_tag
2. delete_tag
3. rollback

No


pan3793 (approved):

pan3793:
@davidyuan1223 the failure looks related, could you please fix it?

```
- Producers *** FAILED ***
  Expected exception org.apache.kyuubi.plugin.spark.authz.AccessControlException to be thrown, but java.lang.NoSuchMethodError was thrown (PaimonCatalogRangerSparkExtensionSuite.scala:553)
```

davidyuan1223:
> @davidyuan1223 the failure looks related, could you please fix it?
> 
> ```
> - Producers *** FAILED ***
>   Expected exception org.apache.kyuubi.plugin.spark.authz.AccessControlException to be thrown, but java.lang.NoSuchMethodError was thrown (PaimonCatalogRangerSparkExtensionSuite.scala:553)
> ```
 Ok

pan3793:
> paimon-spark-3.3:0.8.2 has some compaitable question, suggest upgrade the paimon version

@davidyuan1223 so we can upgrade Paimon to bring Spark 3.3 support back? If so, let's upgrade Paimon first

davidyuan1223:
> > paimon-spark-3.3:0.8.2 has some compaitable question, suggest upgrade the paimon version
> 
> @davidyuan1223 so we can upgrade Paimon to bring Spark 3.3 support back? If so, let's upgrade Paimon first

https://github.com/apache/paimon/pull/3465

https://github.com/apache/paimon/blob/release-0.9/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/catalyst/analysis/PaimonProcedureResolver.scala

i think we could upgrade the version to 0.9, from this branch, paimon has support the case compatibile

pan3793:
> i think we could upgrade the version to 0.9, from this branch ...

@davidyuan1223 please go ahead

pan3793:
@davidyuan1223 I don't understand what the exact issue is, we use Paimon spark 3-3 artifacts for testing, why ClassNotFoundException `org.apache.spark.sql.connector.catalog.Column`(introduced in Spark 3.5) is thrown? can you elaborate more about what the underlying happened? it's Kyuubi authz side issue or Paimon issue?

davidyuan1223:
> @davidyuan1223 I don't understand what the exact issue is, we use Paimon spark 3-3 artifacts for testing, why ClassNotFoundException `org.apache.spark.sql.connector.catalog.Column`(introduced in Spark 3.5) is thrown? can you elaborate more about what the underlying happened? it's Kyuubi authz side issue or Paimon issue?

this is the error trace(we only talk about spark3.3
```
  Cause: java.lang.ClassNotFoundException: org.apache.spark.sql.connector.catalog.Column
  at java.net.URLClassLoader.findClass(URLClassLoader.java:387)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
  at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
  at java.lang.Class.getDeclaredMethod(Class.java:2128)
  at org.apache.kyuubi.util.reflect.DynMethods$Builder.hiddenImpl(DynMethods.java:408)
  at org.apache.kyuubi.util.reflect.DynMethods$Builder.hiddenImpl(DynMethods.java:429)
  at org.apache.kyuubi.util.reflect.ReflectUtils$.invokeAs(ReflectUtils.scala:81)
```
this question, i think the root case is the ReflectUtils,which will execute reflect to find targetClass's method instance, in the moment, the code env is from our spark3.3 version, when call XXX.properties method, which need Column class, but in spark3.3 env, the Column not exists. Just like classLoader will first use the bootstrapClassLoader? i'm not sure, but i think the root cause may be the ReflectUtils

pan3793:
This is likely caused by the Paimon improperly packaged issue, `paimon-spark-common` is compiled with Spark 3.5, and those classes are assembled into every version of Spark bundle jars. So technically only the `paimon-spark-3.5` is reliable

davidyuan1223:
> This is likely a Paimon side issue caused by improper packaging, `paimon-spark-common` is compiled with Spark 3.5, and those classes are assembled into every version of Spark bundle jars. So technically only the `paimon-spark-3.5` is reliable

so rollback the commit?

pan3793:
either rollback or keep upgrading to the latest version is fine, as long as tests pass with Spark 3.5

davidyuan1223:
> either rollback or keep upgrading to the latest version is fine, as long as tests pass with Spark 3.5

i think rollback is better, which could compatible with 3.4+

pan3793 (approved):

pan3793:
Thanks, merged to master

Issue #6984, opened 2025-03-15
[Bug] Python magic table dictionary iteration error


JustFeng:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Issue Report: ValueError Occurs When Rendering Map Type Data Using %table Magic Command in Kyuubi

---

#### **Title**  
`ValueError: too many values to unpack (expected 2)` When Rendering Map-Type Data with `%table` Magic Command

---

#### **Description**  
When using the Kyuubi magic command `%table` to display query results containing **Map-type data**, a `ValueError` is thrown during the rendering process. The error indicates that an attempt was made to unpack more values than expected.

---

#### **Steps to Reproduce**  
1. **Execute a Query Returning Map-Type Data**:  
   Run a Kyuubi SQL query that involves a `MAP<KEY, VALUE>` column.  
   Example:  
   ```python
    from pyspark.sql import SparkSession
    from pyspark.sql.types import MapType, StringType, IntegerType
    spark = SparkSession.builder \
        .appName("MapFieldExample") \
        .getOrCreate()

    data = [
        (1, {"a": "1", "b": "2"}),     
        (2, {"x": "10"}),
        (3, {"key": "value"})
    ]

    schema = "id INT, map_col MAP<STRING, STRING>"
    df = spark.createDataFrame(data, schema=schema)
    df.printSchema()
    df2=df.collect()
   ```

2. **Trigger Rendering with `%table`**:  
   Use the `%table` magic command to render the result:  
   ```python
   %table df2
   ```

3. **Observe the Error**:  
   The error occurs during column data processing, specifically when iterating over Map entries.

---

#### **Expected Result**  
The `%table` command should render all columns, including Map-type data 
```text
{'application/vnd.livy.table.v1+json': {'headers': [{'name': 'id', 'type': 'INT_TYPE'}, {'name': 'map_col', 'type': 'MAP_TYPE'}], 'data': [[1, {'a': '1', 'b': '2'}], [2, {'x': '10'}], [3, {'key': 'value'}]]}}
```

---

#### **Actual Result**  
```python
ValueError: too many values to unpack (expected 2)
```
---

#### **Environment Details**  
- **Kyuubi Server Version**:1.10.0
- **Client Environment**:
  - Notebooks: Jupyter
[...]

PR #6985, opened 2025-03-15
[KYUUBI #6984] Fix ValueError when rendering MapType data

JustFeng:
[
[KYUUBI #6984] Fix ValueError when rendering MapType data
](https://github.com/apache/kyuubi/issues/6984)

The issue was caused by an incorrect iteration of MapType data in the `%table` magic command. When iterating over a `MapType` column, the code used `for k, v in m` directly, which leads to a `ValueError` because raw `Map` entries may not be properly unpacked 

- [x] Manual testing:  
  Executed a query with a `MapType` column and confirmed that the `%table` command now renders it without errors.  
```python
 from pyspark.sql import SparkSession
 from pyspark.sql.types import MapType, StringType, IntegerType
 spark = SparkSession.builder \
     .appName("MapFieldExample") \
     .getOrCreate()

 data = [
     (1, {"a": "1", "b": "2"}),     
     (2, {"x": "10"}),
     (3, {"key": "value"})
 ]

 schema = "id INT, map_col MAP<STRING, STRING>"
 df = spark.createDataFrame(data, schema=schema)
 df.printSchema()
 df2=df.collect()
```
using `%table` render table
```python
 %table df2
```

result
```python
{'application/vnd.livy.table.v1+json': {'headers': [{'name': 'id', 'type': 'INT_TYPE'}, {'name': 'map_col', 'type': 'MAP_TYPE'}], 'data': [[1, {'a': '1', 'b': '2'}], [2, {'x': '10'}], [3, {'key': 'value'}]]}}

```

No

**notice** This PR was co-authored by DeepSeek-R1.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/resources/python/execute_python.py
+++ b/externals/kyuubi-spark-sql-engine/src/main/resources/python/execute_python.py
@@ -314,7 +314,7 @@ def magic_table_convert_map(m):
     last_value_type = None
     converted_items = {}
 
-    for key, value in m:
+    for key, value in m.items():
         key_type, key = magic_table_convert(key)
         value_type, value = magic_table_convert(value)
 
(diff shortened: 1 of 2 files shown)


turboFei:
@JustFeng  could you add UT into https://github.com/apache/kyuubi/blob/master/kyuubi-server/src/test/scala/org/apache/kyuubi/engine/spark/PySparkTests.scala? Thanks

turboFei (approved):

turboFei:
thanks

turboFei:
thanks @JustFeng, merged to master, branch-1.10 and branch-1.9

PR #6989, opened 2025-03-18
[KYUUBI #6989] Calculate expected join partitions based on scanned table size

wForget:

Avoid unstable test case caused by table size changes, this is likely to happen when upgrading Parquet/ORC/Spark.

unit test

No


extensions/spark/kyuubi-extension-spark-3-5/src/test/scala/org/apache/spark/sql/DynamicShufflePartitionsSuite.scala
```diff
@@ -75,7 +85,7 @@ class DynamicShufflePartitionsSuite extends KyuubiSparkSQLExtensionTest {
             if (dynamicShufflePartitions) {
               joinExchanges.foreach(e =>
                 assert(e.outputPartitioning.numPartitions
-                  == Math.min(22, maxDynamicShufflePartitionNum)))
+                  == Math.min(expectedJoinPartitionNum, maxDynamicShufflePartitionNum)))
```
pan3793, line 88: should line 130 be updated too?

wForget: thank you, changed

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #6994, opened 2025-03-23
Get the latest batch app info after submit process terminated to prevent batch ERROR due to engine submit timeout

turboFei:

We meet below issue:
For spark on yarn:
```
spark.yarn.submit.waitAppCompletion=false
kyuubi.engine.yarn.submit.timeout=PT10M
```

Due to network issue, the application submission was very slow.

It was submitted after 15 minutes.
image

image

Then the batch failed from PENDING state to ERRO state directly, due to application state NOT_FOUND(exceeds the kyuubi.engine.yarn.submit.timeout).

https://github.com/apache/kyuubi/blob/a54ee39ab338e310c6b9a508ad8f14c0bd82fa0f/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala#L99-L106

image

Here is the operation event:
image

But from the batch log, the current application status should be `PENDING`.
```
:2025-03-21 17:36:19.350 INFO [KyuubiSessionManager-exec-pool: Thread-176922] org.apache.kyuubi.operation.BatchJobSubmission: Batch report for bbba09c8-3704-4a87-8394-9bcbbd39cc34, Some(ApplicationInfo(application_1741747369441_2258235,6042072c-e8fa-425d-a6a3-3d5bbb4ec1e3-275732_6042072c-e8fa-425d-a6a3-3d5bbb4ec1e3-275732.e3a34b86-7fc7-43ea-b4a5-1b6f27df54b5.0_20250322002147.stm,PENDING,Some(https://apollo-rno-rm-2.vip.hadoop.ebay.com:50030/proxy/application_1741747369441_2258235/),Some()))
```

So, we should retrieve the batch application info after the submission process terminated before checking the application failed, to get the current application information to prevent the corner case:
1. the application submission time exceeds the `kyuubi.engine.yarn.submit.timeout` and the app state is NOT FOUND
2. can not get the application report before the submission process terminated
3. then the batch state to ERROR from PENDING directly. 

Conclusion:

The application state transition was:

UNKNOWN(before submit timeout) -> NOT_FOUND(reach submit timeout) -> processExit -> batchOpError -> PENDING(updateApplicationInfoMetadataIfNeeded) -> UNKNOWN(batchError but app not terminated)

After this PR, it should be:

UNKNOWN(before submit timeout) -> NOT_FOUND(reach submit timeout) ->  processExit-> PENDING(after process terminated) -> ....

Existing GA.

No.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
@@ -124,6 +124,7 @@ abstract class AbstractOperation(session: Session) extends Operation with Loggin
 
   protected def setOperationException(opEx: KyuubiSQLException): Unit = {
     this.operationException = opEx
+    withOperationLog(error(s"Error operating $opType: ${opEx.getMessage}", opEx))
   }
 
   def getOperationJobProgress: TProgressUpdateResp = operationJobProgress
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -165,6 +165,8 @@ class BatchJobSubmission(
       opState: OperationState,
       appState: ApplicationState.ApplicationState): ApplicationState.ApplicationState = {
     if (opState == OperationState.ERROR && !ApplicationState.isTerminated(appState)) {
+      withOperationLog(error(s"Batch $batchId state is $opState," +
+        s" but the application state is $appState and not terminated."))
       ApplicationState.UNKNOWN
     } else {
       appState
@@ -272,6 +274,9 @@ class BatchJobSubmission(
         }
         process.waitFor(applicationCheckInterval, TimeUnit.MILLISECONDS)
       }
+
+      // get the current application info after the process is terminated
+      _applicationInfo = currentApplicationInfo()
 
       if (applicationFailed(_applicationInfo)) {
         Utils.terminateProcess(process, applicationStartupDestroyTimeout)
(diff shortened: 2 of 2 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/operation/AbstractOperation.scala
```diff
@@ -124,6 +124,7 @@ abstract class AbstractOperation(session: Session) extends Operation with Loggin
 
   protected def setOperationException(opEx: KyuubiSQLException): Unit = {
     this.operationException = opEx
+    withOperationLog(error(s"Error operating $opType: ${opEx.getMessage}", opEx))
```
turboFei, line 127: Due I do not find the batch operation exception from kyuubi log(I got it from kyuubi operation event stored in elasticsearch), so I think we can log the ex here.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -273,6 +275,9 @@ class BatchJobSubmission(
         process.waitFor(applicationCheckInterval, TimeUnit.MILLISECONDS)
```
pan3793, line 275: so the key point is - in the current round, app state is NOT_FOUND because the submit stage exceeds the `kyuubi.engine.yarn.submit.timeout`, while during the `process.waitFor` period, submit success, then `process.isAlive` returns `false`, it has no chance to retrieve the app state from the cluster manager?

turboFei: Yes,

I think with `spark.yarn.submit.waitAppCompletion=false`, usually the application state is UNKNOWN/NOT_FOUND when process exit.

turboFei: When the process terminated, if the current running time not exceed `kyuubi.engine.yarn.submit.timeout`, the app state is UNKNOWN.

If it exceeds the engine submit timeout, the app state would be `NOT_FOUND`.

turboFei:
close this one, prefer https://github.com/apache/kyuubi/pull/6997

Issue #6926, opened 2025-02-17
[Improvement] Add Server Local Engine Share Level


taylor12805:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

There are some scenarios where we want to assign kyuubi engines on the same server where kyuubi server is running, in which case we can allocate server resources more evenly.
Therefore, I would like to introduce a new ShareLevel that functions as described above.

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #7013, opened 2025-04-03
[KYUUBI #6926] Add SERVER_LOCAL engine share level

taylor12805:

As clarified in https://github.com/apache/kyuubi/issues/6926, there are some scenarios user want to launch engine on each kyuubi server. SERVER_LOCAL engine share level implement this function by extracting local host address as subdomain, in which case each kyuubi server's engine is unique.

No


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -147,6 +160,8 @@ private[kyuubi] class EngineRef(
    *   /`serverSpace_version_USER_engineType`/`user`[/`subdomain`]
    * For `GROUP` share level:
    *   /`serverSpace_version_GROUP_engineType`/`primary group name`[/`subdomain`]
+   * For `SERVER_LOCAL` share level:
+   *   /`serverSpace_version_SERVER_LOCAL_engineType`/`kyuubi server user`/`hostAddress`
```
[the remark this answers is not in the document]
taylor12805: append subdomain to hostAddress, like `hostAddress_sumDomain`?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -82,6 +83,14 @@ private[kyuubi] class EngineRef(
 
   private val enginePoolSelectPolicy: String = conf.get(ENGINE_POOL_SELECT_POLICY)
 
+  private lazy val localHostAddr = JavaUtils.findLocalInetAddress.getHostAddress
```
wForget, line 86:
```
private lazy val localHostAddr = {
  val host = JavaUtils.findLocalInetAddress.getHostAddress
  if (StringUtils.isBlank(host)) {
      throw KyuubiSQLException(
      s"Local host address can not be empty if ShareLevel set to SERVER_LOCAL")
  }
  host
}
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -147,6 +157,9 @@ private[kyuubi] class EngineRef(
    *   /`serverSpace_version_USER_engineType`/`user`[/`subdomain`]
    * For `GROUP` share level:
    *   /`serverSpace_version_GROUP_engineType`/`primary group name`[/`subdomain`]
+   * For `SERVER_LOCAL` share level:
+   *   /`serverSpace_version_SERVER_LOCAL_engineType`
+   *   /`kyuubi server user`/`hostAddress`[/`subdomain`]
```
wForget, line 162: This seems to have an extra layer than the original, @pan3793 @yaooqinn is this allowed? 

Or do we add server host to subdomain? like:

```
/`serverSpace_version_SERVER_LOCAL_engineType`/`kyuubi server user`/`hostAddress`[_`subdomain`]
```

taylor12805: I think it make sense to add server host to subdomain, I will update it

yaooqinn (approved):

PR #7017, opened 2025-04-09
Using mutable JettyServer uri to prevent batch kyuubi instance mismatch

turboFei:

To fix the batch kyuubi instance port is negative issue. 
image

It happen after I stop the kyuubi service.
We should use variable instead of function for jetty server serverUri.
After the server connector stopped, the localPort would be `-2`.

image

Existing UT.

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/ui/JettyServer.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/ui/JettyServer.scala
@@ -49,7 +49,7 @@ private[kyuubi] class JettyServer(
       case _ =>
     }
   }
-  val serverUri: String = connector.getHost + ":" + connector.getLocalPort
+  def getServerUri: String = connector.getHost + ":" + connector.getLocalPort
 
   def addHandler(handler: Handler): Unit = synchronized {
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/ui/JettyServer.scala
```diff
@@ -49,7 +52,10 @@ private[kyuubi] class JettyServer(
       case _ =>
     }
   }
-  def getServerUri: String = connector.getHost + ":" + connector.getLocalPort
+
+  @volatile private var _serverUri: String = _
+  def getServerUri: String =
+    Option(_serverUri).getOrElse(connector.getHost + ":" + connector.getLocalPort)
```
pan3793, line 58: can we at least log some warning message in `getOrElse` block, to indicate the Jetty server has not been started?

turboFei: done

pan3793 (approved):

turboFei:
thanks, merging to master and branch-1.10

PR #7020, opened 2025-04-12
Supports fully asynchronous Engine initialization

HanShuhang:

When `SESSION_ENGINE_LAUNCH_ASYNC` is set to `True`, the client should not block while waiting for the engine to initialize. This is particularly important in scenarios where cluster resources are constrained, as engine initialization can be significantly delayed, potentially causing the client to block during execute calls. This can lead to serious issues—for example, we found that PyHive may lose connection with Kyuubi under such circumstances. This PR enables fully asynchronous engine startup, thereby reducing client response latency and improving the overall user experience.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -104,7 +104,12 @@ class KyuubiSessionImpl(
   }
 
   @volatile private var _client: KyuubiSyncThriftClient = _
-  def client: KyuubiSyncThriftClient = _client
+  def client: KyuubiSyncThriftClient = {
+    // When `SESSION_ENGINE_LAUNCH_ASYNC` is set to true,
+    // launch engine when client used for the first time.
+    waitForEngineLaunched()
+    _client
+  }
 
   @volatile private var _engineSessionHandle: SessionHandle = _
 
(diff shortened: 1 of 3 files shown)


turboFei:
could you check the UT failure?

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -180,8 +185,9 @@ class KyuubiSessionImpl(
             _engineSessionHandle =
               engineClient.openSession(protocol, user, passwd, openEngineSessionConf)
             _client = engineClient
-            logSessionInfo(s"Connected to engine [$host:$port]/[${client.engineId.getOrElse("")}]" +
-              s" with ${_engineSessionHandle}]")
+            logSessionInfo(
+              s"Connected to engine [$host:$port]/[${_client.engineId.getOrElse("")}]" +
+                s" with ${_engineSessionHandle}]")
```
turboFei, line 190: please rebase the code

turboFei: yes, it has been fixed, could you rebase again?

HanShuhang: 👌done

turboFei:
This PR is a little tricky for me.

Did you test whether this PR resolve the issue?

Could you check the KyuubiConnection?
https://github.com/apache/kyuubi/blob/4cbff4d192f75dd2a66b259c309d28f8c34f3e89/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java#L195-L198

HanShuhang:
Our users connect to kyuubi by **pyhive** with **CONNECTION** mode. If the engine is not started for a long time (maybe dozens of minutes) due to waiting for resources, the client's execute sql will be blocked in kyuubi. The final phenomenon is that **even if the spark engine is successfully started finally, the client can not receive any response**. This PR solved this problem, execute request (async) from pyhive can return quickly without blocking while waiting for the engine to start. 
In addition, asynchronous execution should not be blocked waiting for resources in theory, which is obviously not asynchronous enough.

PR #7024, opened 2025-04-15
kyuubi-sparksql support udf permission control

liujinhui1994:
kyuubi-sparksql, supports UDF permission control

We have already run it in production without any problems, and solved the problem of controlling the permissions used by our UDF.

NO

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
@@ -18,6 +18,7 @@
 package org.apache.kyuubi.plugin.spark.authz
 
 import scala.collection.mutable.ArrayBuffer
+import scala.util.{Failure, Success, Try}
 
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.expressions.{AttributeSet, Expression, NamedExpression}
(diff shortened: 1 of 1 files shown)


turboFei:
FYI: You can run `dev/reformat` to refine the style.

liujinhui1994:
@turboFei  please review thanks

pan3793:
@liujinhui1994 any chance to supply a unit test to cover your changes, to avoid being broken accidentally in the future?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
```diff
@@ -341,4 +343,50 @@ object PrivilegesBuilder {
...
+
+    def checkExpressionForUDF(expr: Expression): Unit = {
+      expr match {
+        case udfExpr
+            if udfExpr.nodeName.equals("HiveGenericUDF") ||
+              udfExpr.nodeName.equals("HiveGenericUDTF") ||
+              udfExpr.nodeName.equals("HiveSimpleUDF") ||
+              udfExpr.nodeName.equals("HiveUDAFFunction") =>
+          val funName: String = getFieldValNoError(udfExpr, "name")
+          val isFun: Boolean = getFieldValNoError(udfExpr, "isUDFDeterministic")
```
pan3793, line 359: why should we check `isUDFDeterministic`?

pan3793:
after a closer look at the code, the direction might be incorrect, there is an existing `FunctionExtractor` that supports extracting Hive UDFs, but seems was not applied properly for commands and queries.

liujinhui1994:
> @liujinhui1994 any chance to supply a unit test to cover your changes, to avoid being broken accidentally in the future?
ok

liujinhui1994:
Please take a look. Any other questions? @pan3793

Issue #7028, opened 2025-04-16
[Improvement] Persist the kubernetes metastore


turboFei:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Persist the kubernetes metadata into metastore.

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #7029, opened 2025-04-16
[KYUUBI #7028] Persist the kubernetes application terminate state into metastore for app info store fallback

turboFei:

1. Persist the kubernetes application terminate info into metastore to prevent the event lose.
2. If it can not get the application info from informer application info store, fallback to get the application info from metastore instead of return NOT_FOUND directly. 
3. It is critical because if we return false application state, it might cause data quality issue.

UT and IT.

image

No.

### DIFF
--- a/kyuubi-server/src/main/resources/sql/mysql/006-KYUUBI-7028.mysql.sql
+++ b/kyuubi-server/src/main/resources/sql/mysql/006-KYUUBI-7028.mysql.sql
@@ -1,0 +1,18 @@
+SELECT '< KYUUBI-7028: Persist Kubernetes metadata into metastore' AS ' ';
+
+CREATE TABLE IF NOT EXISTS k8s_engine_info(
+    key_id bigint PRIMARY KEY AUTO_INCREMENT COMMENT 'the auto increment key id',
+    identifier varchar(36) NOT NULL COMMENT 'the identifier id, which is an UUID',
+    context varchar(32) COMMENT 'the kubernetes context',
+    namespace varchar(255) COMMENT 'the kubernetes namespace',
+    pod_name varchar(255) NOT NULL COMMENT 'the kubernetes pod name',
+    pod_state varchar(32) COMMENT 'the kubernetes pod state',
+    container_state mediumtext COMMENT 'the kubernetes container state',
+    engine_id varchar(128) COMMENT 'the engine id',
+    engine_name mediumtext COMMENT 'the engine name',
+    engine_state varchar(32) COMMENT 'the engine state',
+    engine_error mediumtext COMMENT 'the engine diagnose',
+    create_time bigint COMMENT 'the metadata create time',
+    update_time bigint COMMENT 'the metadata update time',
+    UNIQUE INDEX unique_identifier_index(identifier)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/kyuubi-server/src/main/resources/sql/mysql/006-KYUUBI-7028.mysql.sql
+++ b/kyuubi-server/src/main/resources/sql/mysql/006-KYUUBI-7028.mysql.sql
@@ -12,7 +12,6 @@ CREATE TABLE IF NOT EXISTS k8s_engine_info(
     engine_name mediumtext COMMENT 'the engine name',
     engine_state varchar(32) COMMENT 'the engine state',
     engine_error mediumtext COMMENT 'the engine diagnose',
-    create_time bigint COMMENT 'the metadata create time',
     update_time bigint COMMENT 'the metadata update time',
     UNIQUE INDEX unique_identifier_index(identifier)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/kyuubi-server/src/main/resources/sql/postgresql/001-KYUUBI-7028.postgresql.sql
+++ b/kyuubi-server/src/main/resources/sql/postgresql/001-KYUUBI-7028.postgresql.sql
@@ -12,7 +12,6 @@ CREATE TABLE IF NOT EXISTS k8s_engine_info(
     engine_name text,
     engine_state varchar(32),
     engine_error text,
-    create_time bigint NOT NULL,
     update_time bigint NOT NULL
     );
 
@@ -27,7 +26,6 @@ COMMENT ON COLUMN k8s_engine_info.engine_id IS 'the engine id';
 COMMENT ON COLUMN k8s_engine_info.engine_name IS 'the engine name';
 COMMENT ON COLUMN k8s_engine_info.engine_state IS 'the engine state';
 COMMENT ON COLUMN k8s_engine_info.engine_error IS 'the engine diagnose';
-COMMENT ON COLUMN k8s_engine_info.create_time IS 'the metadata create time';
 COMMENT ON COLUMN k8s_engine_info.update_time IS 'the metadata update time';
 
 CREATE UNIQUE INDEX IF NOT EXISTS k8s_engine_info_unique_identifier_index ON k8s_engine_info(identifier);
--- a/kyuubi-server/src/main/resources/sql/postgresql/001-KYUUBI-7028.postgresql.sql
+++ b/kyuubi-server/src/main/resources/sql/postgresql/001-KYUUBI-7028.postgresql.sql
@@ -1,0 +1,33 @@
+SELECT '< KYUUBI-7028: Persist Kubernetes metadata into metastore' AS ' ';
+
+CREATE TABLE IF NOT EXISTS k8s_engine_info(
+    key_id bigserial PRIMARY KEY,
+    identifier varchar(36) NOT NULL,
+    context varchar(32),
+    namespace varchar(255),
+    pod_name varchar(255) NOT NULL,
+    pod_state varchar(32),
+    container_state text,
+    engine_id varchar(128),
+    engine_name text,
+    engine_state varchar(32),
+    engine_error text,
+    create_time bigint NOT NULL,
+    update_time bigint NOT NULL
+    );
+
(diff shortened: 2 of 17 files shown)


turboFei:
image


Testing passed, cc @pan3793

turboFei:
This PR has been well tested. cc @pan3793

kyuubi-server/src/main/resources/sql/mysql/006-KYUUBI-7028.mysql.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE TABLE IF NOT EXISTS kubernetes_metadata(
+    key_id bigint PRIMARY KEY AUTO_INCREMENT COMMENT 'the auto increment key id',
+    identifier varchar(36) NOT NULL COMMENT 'the identifier id, which is an UUID',
+    context varchar(32) COMMENT 'the kubernetes context',
+    namespace varchar(255) COMMENT 'the kubernetes namespace',
+    pod_name varchar(255) NOT NULL COMMENT 'the kubernetes pod name',
+    app_id varchar(128) COMMENT 'the application id',
+    app_name mediumtext COMMENT 'the application name',
+    app_state varchar(32) COMMENT 'the application state',
+    app_error mediumtext COMMENT 'the application diagnose',
```
pan3793, line 12: Why name those columns `app_*`, instead of `engine_*`, as we did for `metadata` table?

kyuubi-server/src/main/resources/sql/mysql/006-KYUUBI-7028.mysql.sql
```diff
@@ -0,0 +1,16 @@
...
+
+CREATE TABLE IF NOT EXISTS kubernetes_metadata(
+    key_id bigint PRIMARY KEY AUTO_INCREMENT COMMENT 'the auto increment key id',
+    identifier varchar(36) NOT NULL COMMENT 'the identifier id, which is an UUID',
+    context varchar(32) COMMENT 'the kubernetes context',
+    namespace varchar(255) COMMENT 'the kubernetes namespace',
+    pod_name varchar(255) NOT NULL COMMENT 'the kubernetes pod name',
+    app_id varchar(128) COMMENT 'the application id',
+    app_name mediumtext COMMENT 'the application name',
+    app_state varchar(32) COMMENT 'the application state',
```
pan3793, line 11: do we need to store pod/container state?

kyuubi-server/src/main/resources/sql/mysql/006-KYUUBI-7028.mysql.sql
```diff
@@ -0,0 +1,16 @@
+SELECT '< KYUUBI-7028: Persist Kubernetes metadata into metastore' AS ' ';
+
+CREATE TABLE IF NOT EXISTS kubernetes_metadata(
```
pan3793, line 3: the `metadata` is a quite general concept, thus it cannot help the user understand what the table is for. I saw we use `appInfoStore` in the code, it's a good name, maybe the table name could be `k8s_app_info`, or `k8s_engine_info`, or `k8s_engine_info_store`

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -419,6 +419,142 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
...
+         |pod_state,
+         |container_state,
+         |engine_id,
+         |engine_name,
+         |engine_state,
+         |engine_error,
+         |create_time,
+         |update_time
+         |)
+         |SELECT ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,?,?
```
pan3793, line 439: nit: spaces

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -419,6 +419,142 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
...
+         |WHERE NOT EXISTS (
```
pan3793, line 440: not sure if it's better to use the DB-specific dialect, e.g. MySQL supports `INSERT IGNORE INTO`

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -419,6 +419,142 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
...
+              s"with params: ${params.mkString(", ")}")
+        }
+      }
+    }
+
+  }
+
+  override def upsertKubernetesEngineInfo(metadata: KubernetesEngineInfo): Unit = {
+    insertKubernetesEngineInfo(metadata)
+    updateKubernetesEngineInfo(metadata)
```
pan3793, line 529: MySQL has native support for `INSERT ... ON DUPLICATE KEY UPDATE`, it's the simplest and most efficient way, and if we use DB-specific implementations, we don't need to care about the multi-instance issue for SQLite.

pan3793 (commented):
DB schema LGTM, the upsert implementation might have improvement room

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JdbcDatabaseDialect.scala
```diff
@@ -27,6 +32,47 @@ class GenericDatabaseDialect extends JdbcDatabaseDialect {
...
+      table: String,
+      colsToInsert: Seq[String],
+      colsToReplace: Seq[String],
+      keyCol: String): Option[String] = {
+    Some(
+      s"""
+         |INSERT INTO $table (${colsToInsert.mkString(",")})
+         |VALUES (${colsToInsert.map(_ => "?").mkString(",")})
+         |ON DUPLICATE KEY UPDATE
+         |${colsToReplace.map(c => s"$c = VALUES($c)").mkString(",")}
```
turboFei: Refer: https://dev.mysql.com/doc/refman/8.4/en/insert-on-duplicate.html

image

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JdbcDatabaseDialect.scala
```diff
@@ -19,6 +19,10 @@ package org.apache.kyuubi.server.metadata.jdbc
 
 trait JdbcDatabaseDialect {
   def limitClause(limit: Int, offset: Int): String
+  def insertOrReplace(
+      table: String,
+      cols: Seq[String],
+      keyCol: String): Option[String] = None
```
pan3793, line 25: can we have a default implementation instead of returning `Option[String]`? Or simply leave empty default implementation as all supported RDBMSs have dedicated dialects

kyuubi-server/src/main/resources/sql/postgresql/001-KYUUBI-7028.postgresql.sql
```diff
@@ -0,0 +1,31 @@
...
+    namespace varchar(255),
+    pod_name varchar(255) NOT NULL,
+    pod_state varchar(32),
+    container_state text,
+    engine_id varchar(128),
+    engine_name text,
+    engine_state varchar(32),
+    engine_error text,
+    update_time bigint NOT NULL
+    );
```
pan3793, line 16: super nit:
```suggestion
);
```

pan3793 (approved):
LGTM, except for the dialect API style

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JdbcDatabaseDialect.scala
pan3793: oh, I see the issue, it might require more than one SQL to complete the logic

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JdbcDatabaseDialect.scala
```diff
@@ -19,14 +19,71 @@ package org.apache.kyuubi.server.metadata.jdbc
...
+      cols: Seq[String],
+      keyCol: String,
+      keyVal: String): String = {
+    s"""
+       |INSERT INTO $table (${cols.mkString(",")})
+       |SELECT ${cols.map(_ => "?").mkString(",")}
+       |WHERE NOT EXISTS (
+       |  SELECT 1 FROM $table WHERE $keyCol = '$keyVal')
+       |)
+       |""".stripMargin
```
turboFei, line 45: cc @pan3793 how about now?

I have to inject the keyVal here.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JdbcDatabaseDialect.scala
turboFei: Updated.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JdbcDatabaseDialect.scala
pan3793: lgtm

turboFei:
thanks, merged to 1.11.0

PR #7033, opened 2025-04-16
Treat YARN/Kubernetes application NOT_FOUND as failed to prevent data quality issue

turboFei:

Currently, NOT_FOUND application stated is treated as a terminated but not failed state.

It might cause some data quality issue if downstream application depends on the batch state for data processing.

So, I think we should treat NOT_FOUND as a failed state instead.

Currently, we support 3 types of application manager.
1. [JpsApplicationOperation](https://github.com/apache/kyuubi/blob/master/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/JpsApplicationOperation.scala)
2. [YarnApplicationOperation](https://github.com/apache/kyuubi/blob/master/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/YarnApplicationOperation.scala)
3. [KubernetesApplicationOperation](https://github.com/apache/kyuubi/blob/master/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala)

YarnApplicationOperation and KubernetesApplicationOperation are widely used in production use case.

And in multiple kyuubi instance mode, the NOT_FOUND case should rarely happen.
1.  https://github.com/apache/kyuubi/blob/7e199d6fdbdf52222bb3eadd056b9e5a2295f36e/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala#L369-L385

3. https://github.com/apache/kyuubi/pull/7029

So, I think we should treat NOT_FOUND as a failed state in production use case.
It is better to fail some corner cases than to mistakenly set unsuccessful batches to the finished state.

GA.

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -99,10 +99,14 @@ object ApplicationState extends Enumeration {
 
   def isFailed(
       state: ApplicationState,
-      appOperation: Option[ApplicationOperation]): Boolean = state match {
+      appOperation: Option[ApplicationOperation]): Boolean = {
+    isFailed(state, appOperation.exists(_.supportPersistedAppState))
+  }
+
+  def isFailed(state: ApplicationState, supportPersistedAppState: Boolean): Boolean = state match {
     case FAILED => true
     case KILLED => true
-    case NOT_FOUND => appOperation.exists(_.supportPersistedAppState)
+    case NOT_FOUND => supportPersistedAppState
     case _ => false
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -94,6 +94,7 @@ object ApplicationState extends Enumeration {
   def isFailed(state: ApplicationState): Boolean = state match {
     case FAILED => true
     case KILLED => true
+    case NOT_FOUND => true
     case _ => false
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -91,10 +91,17 @@ object ApplicationState extends Enumeration {
   type ApplicationState = Value
   val PENDING, RUNNING, FINISHED, KILLED, FAILED, ZOMBIE, NOT_FOUND, UNKNOWN = Value
 
-  def isFailed(state: ApplicationState): Boolean = state match {
+  def isFailed(
+      state: ApplicationState,
+      appOperation: Option[ApplicationOperation]): Boolean = state match {
     case FAILED => true
     case KILLED => true
-    case NOT_FOUND => true
+    case NOT_FOUND => appOperation match {
+        // For YARN and Kubernetes operations, if the application is not found, treat it as failed
+        // to prevent mistakenly set unsuccessful applications to the finished state.
+        case Some(_: YarnApplicationOperation) | Some(_: KubernetesApplicationOperation) => true
+        case _ => false
+      }
     case _ => false
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -106,7 +106,7 @@ object ApplicationState extends Enumeration {
   def isFailed(state: ApplicationState, supportPersistedAppState: Boolean): Boolean = state match {
     case FAILED => true
     case KILLED => true
-    case NOT_FOUND => supportPersistedAppState
+    case NOT_FOUND if supportPersistedAppState => true
     case _ => false
   }
 
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
@@ -85,6 +85,12 @@ trait ApplicationOperation {
       tag: String,
       proxyUser: Option[String] = None,
       submitTime: Option[Long] = None): ApplicationInfo
+
+  /**
+   * Whether the application state can be persisted and retrieved after finished.
+   * @return true if the application state can be persisted
+   */
+  def supportPersistedAppState: Boolean
 }
 
 object ApplicationState extends Enumeration {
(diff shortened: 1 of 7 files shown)


turboFei:
cc @pan3793

turboFei:
I am confident for this PR after https://github.com/apache/kyuubi/pull/7029 

cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -212,7 +215,8 @@ class BatchJobSubmission(
         metadata match {
           case Some(metadata) if metadata.peerInstanceClosed =>
             setState(OperationState.CANCELED)
-          case Some(metadata) if metadata.state == OperationState.PENDING.toString =>
+          case Some(metadata)
+              if metadata.state == OperationState.PENDING.toString && metadata.engineId == null =>
```
turboFei, line 219: In case the applicationInfo is set by peer

https://github.com/apache/kyuubi/blob/7e199d6fdbdf52222bb3eadd056b9e5a2295f36e/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala#L369-L385

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -223,6 +227,14 @@ class BatchJobSubmission(
               case Some(appId) => monitorBatchJob(appId)
             }
           case Some(metadata) =>
+            if (StringUtils.isNoneBlank(metadata.engineId, metadata.engineState)) {
+              _applicationInfo = Some(new ApplicationInfo(
+                id = metadata.engineId,
+                name = metadata.engineName,
+                state = ApplicationState.withName(metadata.engineState),
+                url = Option(metadata.engineUrl),
+                error = metadata.engineError))
```
turboFei, line 236: try to read the application info from metadata first.

In case it has been finished and metadata updated by peer.
https://github.com/apache/kyuubi/blob/7e199d6fdbdf52222bb3eadd056b9e5a2295f36e/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala#L369-L385

turboFei:
cc @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
```diff
@@ -91,9 +91,17 @@ object ApplicationState extends Enumeration {
...
-  def isFailed(state: ApplicationState): Boolean = state match {
+  def isFailed(
+      state: ApplicationState,
+      appOperation: Option[ApplicationOperation]): Boolean = state match {
     case FAILED => true
     case KILLED => true
+    case NOT_FOUND => appOperation match {
+        // For YARN and Kubernetes operations, if the application is not found, treat it as failed
+        // to prevent mistakenly set unsuccessful applications to the finished state.
+        case Some(_: YarnApplicationOperation) | Some(_: KubernetesApplicationOperation) => true
```
pan3793, line 102: Instead of judging the class, how about exposing a method, i.e. `supportPersistedAppState`

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -212,6 +215,20 @@ class BatchJobSubmission(
         metadata match {
           case Some(metadata) if metadata.peerInstanceClosed =>
             setState(OperationState.CANCELED)
+          case Some(metadata)
+              // in case it has been updated by peer kyuubi instance, see KYUUBI-6278
```
pan3793, line 219: nit: `KYUUBI-6278` => `KYUUBI #6278`

does this fix another issue? what happens without this part of the code before/after this PR?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -566,8 +566,9 @@ object KubernetesApplicationOperation extends Logging {
       case Some(containerAppState) => containerAppState
       case None => podAppState
     }
-    val applicationError =
-      if (ApplicationState.isFailed(applicationState)) {
+    val applicationError = {
+      // here the applicationState could not been NOT_FOUND, safe to use None ApplicationOperation
+      if (ApplicationState.isFailed(applicationState, appOperation = None)) {
```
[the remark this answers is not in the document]
turboFei: The function is in Object, no `appOperation` instance to transfer.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
turboFei: Sounds good.

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: Let me move it to another separate PR.

turboFei:
@pan3793 addresses all the comments

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: The issue is that,

1.  if the batch state already updated by peer with terminated state
2. without this change, it would try to get the current application state and might NOT_FOUND

turboFei: Before this PR:

1. it might get NOT_FOUND
2. then batch finished

turboFei: After this PR:
1. respect the terminated application state set by peer, and then FINISH/FAIL the batch.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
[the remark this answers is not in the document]
turboFei: resolved.

```
  def isFailed(state: ApplicationState, supportPersistedAppState: Boolean): Boolean = state match {
    case FAILED => true
    case KILLED => true
    case NOT_FOUND => supportPersistedAppState
    case _ => false
  }
```

Here invoke:

```
ApplicationState.isFailed(applicationState, supportPersistedAppState = true)
```

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
turboFei: FYI:
https://github.com/apache/kyuubi/blob/70c03ef998bb695d62202f2378f3a09c5fcdfd83/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala#L323-L325

For monitorBatchJob, it would get the currentApplicationInfo first, and does not respect the metadata store application state.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ApplicationOperation.scala
```diff
@@ -85,15 +85,28 @@ trait ApplicationOperation {
...
+  def isFailed(
+      state: ApplicationState,
+      appOperation: Option[ApplicationOperation]): Boolean = {
+    isFailed(state, appOperation.exists(_.supportPersistedAppState))
+  }
+
+  def isFailed(state: ApplicationState, supportPersistedAppState: Boolean): Boolean = state match {
     case FAILED => true
     case KILLED => true
+    case NOT_FOUND => supportPersistedAppState
```
pan3793, line 109:
```suggestion
    case NOT_FOUND if supportPersistedAppState => true
```

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #7034, opened 2025-04-16
[KUBERNETES] Prefer to use pod `spark-app-name` label as application name than pod name

turboFei:

After https://github.com/apache/spark/pull/34460 (Since Spark 3.3.0), the `spark-app-name` is available.

We shall use it as the application name if it exists.

Minor change.

No.


pan3793 (approved):

turboFei:
thanks, merged to 1.11.0 and 1.10.2

PR #7035, opened 2025-04-18
Close the operation by operation manager to prevent operation leak

turboFei:

To fix the operation leak if the session init timeout.

1. the operationHandle has not been added into session `opHandleSet` (super.runOperation).
2. The `operation.close()` only close the operation, but it does not remove it from `handleToOperation` map.
3. the session close would not remove the opHandle from `handleToOperation` as it has not been added into session `opHandleSet`

So here we can resolve the operation leak by invoking `operationManager.closeOperation` to remove the operation handle and close session.
https://github.com/apache/kyuubi/blob/cc68cb4c852dccd920ba53a4d853aa407ff83faa/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala#L235-L246

https://github.com/apache/kyuubi/blob/cc68cb4c852dccd920ba53a4d853aa407ff83faa/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala#L100-L103 

https://github.com/apache/kyuubi/blob/cc68cb4c852dccd920ba53a4d853aa407ff83faa/kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala#L127-L130

https://github.com/apache/kyuubi/blob/cc68cb4c852dccd920ba53a4d853aa407ff83faa/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala#L89-L92

FYI: the operation was added into `handleToOperation` during new operation.
https://github.com/apache/kyuubi/blob/cc68cb4c852dccd920ba53a4d853aa407ff83faa/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/KyuubiOperationManager.scala#L56-L64

Minor change.

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -238,7 +238,7 @@ class KyuubiSessionImpl(
         waitForEngineLaunched()
       } catch {
         case t: Throwable =>
-          operation.close()
+          sessionManager.operationManager.closeOperation(operation.getHandle)
           throw t
       }
       sessionEvent.totalOperations += 1


pan3793:
> 1. the operationHandle has not been added into session opHandleSet (`super.runOperation`).

do you rememeber why we don't process `LaunchEngine` op like others? 

> 2. The operation.close() only close the session ...

maybe close the operation?

turboFei:
> maybe close the operation?

Fixed.

turboFei:
> do you remember why we don't process LaunchEngine op like others?

Maybe for engine log retrieve?


BTW, the leaked operation usually is ExecuteStatementOperation.

pan3793:
> > do you remember why we don't process LaunchEngine op like others?
> 
> Maybe for engine log retrieve?

NVM, I misinterpreted the code. LGTM

pan3793:
Merged to master/1.10.2/1.9.4

Issue #6884, opened 2025-01-09
[FEATURE] Support to reassign the batches to alternative kyuubi instance in case kyuubi instance lost 


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Support to reassign the batches to alternative kyuubi instance in case kyuubi instance lost.

### Motivation

For our use case, we deploy kyuubi server on k8s.

But if the pod can not re-create after termination because of resource or k8s issue, the kyuubi instance can not get back in short time.

It is better that we can support to reassign the batchs that are in `PENDING`, `RUNNING` status to alternative kyuubi instance.

### Describe the solution

1. we need to ensure the kyuubi instance is unreachable and the alternative kyuubi instance is reachable 
2. change the batch metadata `kyuubi_instance` to alternative kyuubiInstance 
3. call the alternative kyuubiInstance to recover the specified batch

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

turboFei:
cc @George314159

PR #7037, opened 2025-04-22
[KYUUBI #6884] [FEATURE] Support to reassign the batches to alternative kyuubi instance in case kyuubi instance lost 

George314159:

Support to reassign the batches to alternative kyuubi instance in case kyuubi instance lost.
https://github.com/apache/kyuubi/issues/6884

Unit Test

No


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -561,6 +561,58 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+      schema = new Schema(implementation = classOf[ReassignBatchResponse]))),
+    description = "Reassign instance's all batch sessions to a new instance")
+  @GET
+  @Path("{kyuubiInstance}/{batchId}/reassign")
+  def reassignBatchSessions(
+      @PathParam("kyuubiInstance") kyuubiInstance: String,
+      @PathParam("batchId") batchId: String,
+      @QueryParam("force")
+      @DefaultValue("false") force: Boolean): ReassignBatchResponse = {
+    sessionManager.getBatchMetadata(batchId).map { metadata =>
```
turboFei, line 577: some comments:

1. for single batchId recovery, we can get the kyuubiInstance from metastore and respect the real kyuubi instance
2. for kyuubi instance batches recovery, we should add a `size` to only assign the limited batches to the target kyuubi instance
3. The response struct is better be `List[Batch], originalKyuubiInsance: String, newKyuubiInstance: String`, we can throw exception if does not successful, so `success` flag seems not necessary.
4. there is side impact if call `fe.recoverBatchSessions()`, it might recover all the `pending` and `running` batches  again

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -561,6 +561,58 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
+      val url = fe.connectionUrl
+      if (kyuubiInstance == url) {
+        throw new IllegalStateException(s"KyuubiInstance is alive: $kyuubiInstance")
+      }
+      if (!force && checkInstanceNetwork(kyuubiInstance, batchId)) {
+        throw new IllegalStateException(s"KyuubiInstance is alive: $kyuubiInstance," +
+          s"please use force option to bypass this check")
+      }
+      try {
+        sessionManager.reassignBatchSessions(kyuubiInstance, fe.connectionUrl)
```
turboFei, line 591: the logical is not correct,

you check the `batch` exists and then try to reassign all the batches on the instance. 

It is not align.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -327,6 +327,20 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
     }
   }
 
+  def reassignBatchSessions(kyuubiInstance: String, newKyuubiInstance: String): Unit = {
```
turboFei, line 330: return Seq[String]

turboFei:
rebased the code

turboFei (approved):
LGTM, cc @pan3793

pan3793:
quite useful and nice feature for failure tolerance, I will take a closer look next week

turboFei:
ping @pan3793

PR #7041, opened 2025-04-23
Fix NPE when getting metadtamanager in KubernetesApplicationOperation

turboFei:

To fix NPE.

Before, we use below method to get `metadataManager`.
```
private def metadataManager = KyuubiServer.kyuubiServer.backendService
    .sessionManager.asInstanceOf[KyuubiSessionManager].metadataManager
```
But before the kyuubi server fully restarted, the `KyuubiServer.kyuubiServer` is null and might throw NPE during batch recovery phase.

For example:

```
:2025-04-23 14:06:24.040 ERROR [KyuubiSessionManager-exec-pool: Thread-231] org.apache.kyuubi.engine.KubernetesApplicationOperation: Failed to get application by label: kyuubi-unique-tag=95116703-4240-4cc1-9886-ccae3a2ac879, due to Cannot invoke "org.apache.kyuubi.server.KyuubiServer.backendService()" because the return value of "org.apache.kyuubi.server.KyuubiServer$.kyuubiServer()" is null
```

Existing GA.

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
@@ -52,12 +52,11 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
 
   val operationManager = new KyuubiOperationManager()
   val credentialsManager = new HadoopCredentialsManager()
-  val applicationManager = new KyuubiApplicationManager()
 
   // Currently, the metadata manager is used by the REST frontend which provides batch job APIs,
   // so we initialize it only when Kyuubi starts with the REST frontend.
-  lazy val metadataManager: Option[MetadataManager] =
-    if (conf.isRESTEnabled) Some(new MetadataManager()) else None
+  var metadataManager: Option[MetadataManager] = None
+  var applicationManager: KyuubiApplicationManager = _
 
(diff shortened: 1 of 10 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -52,12 +52,8 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
 
   val operationManager = new KyuubiOperationManager()
   val credentialsManager = new HadoopCredentialsManager()
-  val applicationManager = new KyuubiApplicationManager()
-
-  // Currently, the metadata manager is used by the REST frontend which provides batch job APIs,
-  // so we initialize it only when Kyuubi starts with the REST frontend.
-  lazy val metadataManager: Option[MetadataManager] =
-    if (conf.isRESTEnabled) Some(new MetadataManager()) else None
+  val metadataManager: MetadataManager = new MetadataManager()
```
turboFei, line 55: It is fine to always create metadatamanager, especially we are going to persist the kubernetes information into database.

turboFei:
cc @pan3793 could you help take a look?

This PR fix the kubernetes appliction recovery issue.

pan3793 (approved):
thanks for fixing this. lgtm if UT pass

turboFei:

PR #7043, opened 2025-04-24
Support to construct the batch info from metadata directly

turboFei:

Add an option to allow construct the batch info from metadata directly instead of redirecting the requests to reduce the RPC latency. 

Minor change and Existing GA.

No.


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1951,6 +1951,16 @@ object KyuubiConf {
       .stringConf
       .createWithDefault("1")
 
+  val BATCH_INFO_INTERNAL_REDIRECT = buildConf("kyuubi.batch.info.internal.redirect")
+    .serverOnly
+    .doc("When set to true, the batch info is retrieved by forwarding the request to the " +
+      "corresponding Kyuubi instance using the internal REST client. If false, the batch info is " +
```
pan3793, line 1957: let's make the "corresponding" clear

to the corresponding Kyuubi instance using the internal REST client.
=>
to the Kyuubi instance that submitted this job via internal RPC.

pan3793: I remember there may have been some state inconsistent risk for batch v1 implementation previously, is that still true?

turboFei: I can't think of any risks with the current V1 implementation

pan3793 (approved):

turboFei:
thanks, merged to 1.11.0

Issue #7049, opened 2025-04-28
[FEATURE] Ranger check support iceberg all operation permission check


davidyuan1223:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

see #7040 

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

baotran306:
Hi @davidyuan1223, I hope you are doing well, any update for this change?

davidyuan1223:
@baotran306 hello, could you list todo list for me, the test cases i can not make sure could all resolve currently, maybe we could test it one by one, like #5470

davidyuan1223:
> [@baotran306](https://github.com/baotran306) hello, could you list todo list for me, the test cases i can not make sure could all resolve currently, maybe we could test it one by one, like [#5470](https://github.com/apache/kyuubi/issues/5470)

In your issue, you could list that, then i will fix the missing test case one by one, it could help us easy to merge request and check

davidyuan1223:
#7065 this pr will update 5 files to fix on test case, if we make all test case in one pr, it's not good for review

baotran306:
> [@baotran306](https://github.com/baotran306) hello, could you list todo list for me, the test cases i can not make sure could all resolve currently, maybe we could test it one by one, like [#5470](https://github.com/apache/kyuubi/issues/5470)

Hi @davidyuan1223, I've listed the to-do list for you in #7040. Please help me double-check it. Thanks for your effort.

davidyuan1223:
> > [@baotran306](https://github.com/baotran306) hello, could you list todo list for me, the test cases i can not make sure could all resolve currently, maybe we could test it one by one, like [#5470](https://github.com/apache/kyuubi/issues/5470)
> 
> Hi [@davidyuan1223](https://github.com/davidyuan1223), I've listed the to-do list for you in [#7040](https://github.com/apache/kyuubi/issues/7040). Please help me double-check it. Thanks for your effort.

great!

PR #7052, opened 2025-05-02
[KYUUBI #7049] Iceberg Ranger Check test case improvement

davidyuan1223:
#7049


davidyuan1223:
This pr is a draft(currently), i change the derby mode from path to memory, because which will has conflict when iceberg execute alter command

nqvuong1998:
Hi @davidyuan1223 , any update for these test cases?

davidyuan1223:
> Hi @davidyuan1223 @turboFei , any update for these test cases?

Iceberg has compatible error with current branch, I need more time to resolve that. Such has alter table rename table name will has error about current derby mode

Issue #7051, opened 2025-05-02
[Bug] JDBC Driver with hive2 protocol throw HadoopConfiguration class not found when `trustStorePassword` is not provided


StephenOTT:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Using https://kyuubi.readthedocs.io/en/master/client/jdbc/kyuubi_jdbc.html#hiveserver2-compatibility with 

```
<dependency>
    <groupId>org.apache.kyuubi</groupId>
    <artifactId>kyuubi-hive-jdbc-shaded</artifactId>
    <version>1.10.1</version>
</dependency>
```

If you use a url such as 

`jdbc:hive2://some_url:1234/my_db;transportMode=http;ssl=true;httpPath=/some_path`

and you provide credentials using user/password props as part of the lib,

then you get a Class Not Found exception when trying to creation a connection.

The error traces back to:

https://github.com/apache/kyuubi/blob/master/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java#L566-L567

Which says essentially: If there is ssl=true then look for the password in the `trustStorePassword` arg.

But then we can see in the getPassword method:

https://github.com/apache/kyuubi/blob/master/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java#L665-L668

If the password was null / not found in the `trustStorePassword` arg / the arg did not exist, then try to create the HadoopConfiguration class instance.

But this class is NOT part of the `kyuubi-hive-jdbc-shaded` jar.  And thus we get an error.

Current Workaround:  If you add a arg into the JDBC url such as `trustStorePassword=fake_value` then the connection is successful and no issues.

### Affects Version(s)

JDBC Shaded 1.10.1

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

StephenOTT:
[...]

PR #7058, opened 2025-05-09
[KYUUBI #7051] Fix usage without sslTrustStore of JDBC driver

Khrol:

If `sslTrustStore` is not provided `org.apache.hadoop.conf.Configuration` class existence becomes a hard dependency.
This makes jdbc client too complex to configure: extra Hadoop jars should be provided.

`hadoopCredentialProviderAvailable` variable is useless in the previous implementation logic because it's always `true` or the code is not reachable.

Screenshot 2025-05-09 at 13 05 12

Build jar and used it to connect from DataGrip.
Screenshot 2025-05-09 at 13 01 29

No.


turboFei (approved):

pan3793 (approved):

pan3793:
Thanks, merged to master/1.10.2/1.9.4

PR #6920, opened 2025-02-14
Spark SQL engine supports Spark 4.0

pan3793:
### Why are the changes needed?

Spark 4.0 continues to receive breaking changes since 4.0.0-preview2, and the 4.0.0 RC1 is scheduled at 20250215, this PR fixes all compatibility for the latest Spark 4.0.0-SNAPSHOT for Spark SQL engine.

### How was this patch tested?

Pass GHA with `spark-master`

### Was this patch authored or co-authored using generative AI tooling?

No.

PR #7061, opened 2025-05-15
[KYUUBI #6920][FOLLOWUP] Spark SQL engine supports Spark 4.0

pan3793:

There were some breaking changes after we fixed compatibility for Spark 4.0.0 RC1 in #6920, but now Spark has reached 4.0.0 RC6, which has less chance to receive more breaking changes.

Changes are extracted from https://github.com/apache/kyuubi/pull/6928, which passed CI with Spark 4.0.0 RC6

No.

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
@@ -17,17 +17,19 @@
 
 package org.apache.kyuubi.spark.connector.hive
 
+import java.lang.{Boolean => JBoolean, Long => JLong}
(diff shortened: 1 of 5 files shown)


extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
```diff
@@ -166,6 +183,129 @@ class HiveTableCatalog(sparkSession: SparkSession)
       HiveTable(sparkSession, catalog.getTableMetadata(ident.asTableIdentifier), this)
     }
 
+  // scalastyle:off
```
cfmcgrady, line 186: why do we need to disable the style check for this method?

cfmcgrady (approved):

pan3793:
Merged to master/1.10

PR #7064, opened 2025-05-15
[WIP][KYUUBI #7064] Fix thrift GetSchemas/GetTables with custom session catalog

pan3793:

The thrift `GetSchemas`/`GetTables` operations assumed `spark_catalog` is always the builtin V1 session catalog and listed databases/tables via `spark.sessionState.catalog`. When a custom session catalog is configured, e.g.

```
spark.sql.catalog.spark_catalog=org.apache.iceberg.spark.SparkSessionCatalog
```

these operations bypass the DSv2 catalog APIs (`SupportsNamespaces`/`TableCatalog`) and the V1 session catalog path cannot even parse the V2 table metadata the custom catalog persists in HMS - it may return wrong results, or throw and fail the whole request. #7673 is a real-world example: `GetTables` fails with `ClassNotFoundException: org.apache.iceberg.mr.hive.HiveIcebergSerDe` when the V1 `HiveExternalCatalog` tries to load tables registered by Iceberg with `storage_handler = org.apache.iceberg.mr.hive.HiveIcebergStorageHandler`.

### What changes are in this patch?

- `SparkCatalogUtils.hasCustomSessionCatalog`: detects a custom session catalog via the `builtin` magic value (SPARK-50700)
- `getSchemas`: when `spark_catalog` has a custom session catalog, list namespaces through the DSv2 API instead of `spark.sessionState.catalog`
- `getCatalogTablesOrViews`: match `TableCatalog` before the builtin session-catalog case so a custom session catalog takes the DSv2 listing path
- test infra: `WithSimpleHMSContainer` (testcontainers) runs a real HMS; `SparkIcebergOperationSuite` now runs with `catalogImplementation=hive`, an Iceberg `SparkSessionCatalog` as session catalog and a hadoop-type Iceberg catalog as default catalog; the get tables test covers both `spark_catalog` and the v2 catalog

### Current status (WIP, takeover welcome)

Rebased onto apache/master (`b090cc28e`) as single commit `8bd7d9fc8`; includes local WIP test extensions (null catalogName coverage, per-catalog namespace creation) not yet on the remote branch.

Local run on JDK 17 / Spark 3.5.8:

```
build/mvn test -pl externals/kyuubi-spark-sql-engine -am -Dtest=none \
  -DwildcardSuites=org.apache.kyuubi.engine.spark.operation.SparkIcebergOperationSuite \
  -Dmaven.plugin.scalatest.include.tags=org.apache.kyuubi.tags.IcebergTest \
  -Dmaven.plugin.scalatest.exclude.tags=
```

Result: 5 passed, 3 failed, all in `IcebergMetadataTests`:

1. `get schemas` / `get schemas with multipart namespaces`: `metaData.getSchemas(null, ...)` returns `null` as TABLE_CAT, expected the resolved default catalog name (`hadoop_prod`). Regression of this PR's `getSchemas` rewrite: rows are tagged with the raw `catalogName` argument, while master tags non-session-catalog rows with the resolved `catalog.name`. Fix: use the resolved catalog name in the DSv2 branch.
2. `get tables`: `CREATE NAMESPACE spark_catalog.`a.b`` fails with `` `a.b` is not a valid name for tables/databases`` - Spark's V1 session catalog rejects dotted names and Iceberg's `SparkSessionCatalog` delegates to it. Dotted namespaces are only creatable on the v2 catalog; the test should create them only under the v2 catalog (master does so against the default catalog only).

Remaining work:

- fix the `getSchemas` catalogName tagging (item 1)
- restructure the get tables test data (item 2): flat namespaces under both catalogs, dotted only under the v2 catalog
- decide whether `nekyuubi/kyuubi-hive-metastore:latest` should live somewhere more official/maintained
- TODO already in code: view type is not restored for the session catalog in `getCatalogTablesOrViews`
- verify against Spark 4.x profiles

`SparkIcebergOperationSuite` extended (real HMS via testcontainers + custom session catalog coverage). Currently 3 failures as described above; the other 5 tests in the suite pass.

Assisted-by: GLM 5.3

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/DataFrameHolder.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/DataFrameHolder.scala
@@ -17,7 +17,7 @@
 
 package org.apache.kyuubi.engine.spark.repl
 
-import java.util
+import java.util.HashMap
 
 import org.apache.spark.kyuubi.SparkContextHelper
 import org.apache.spark.sql.{DataFrame, SparkSession}
@@ -29,7 +29,7 @@ import org.apache.spark.sql.{DataFrame, SparkSession}
  */
 class DataFrameHolder(spark: SparkSession) {
 
-  private val results = new util.HashMap[String, DataFrame]()
+  private val results = new HashMap[String, DataFrame]()
 
   private def currentId: String = {
     SparkContextHelper.getCurrentStatementId(spark.sparkContext)
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/DataFrameHolder.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/repl/DataFrameHolder.scala
@@ -17,7 +17,7 @@
 
 package org.apache.kyuubi.engine.spark.repl
 
-import java.util.HashMap
+import java.util
 
 import org.apache.spark.kyuubi.SparkContextHelper
 import org.apache.spark.sql.{DataFrame, SparkSession}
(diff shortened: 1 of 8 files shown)


pan3793:
I don't have cycles to finish this PR in the near future. The current state, remaining work, and known test failures are described in the PR description above.

To be explicit: I'm the sole author of the changes on this branch, and I authorize any Apache Kyuubi contributor to take over this work - feel free to fork the branch, reuse or rework any part of the code in your own PR, with or without attribution, under the Apache License 2.0.

Happy to review a takeover PR.

PR #7095, opened 2025-06-11
Respect terminated app state when building batch info from metadata

turboFei:

Respect terminated app state when building batch info from metadata

It is a followup for https://github.com/apache/kyuubi/pull/2911, 
https://github.com/apache/kyuubi/blob/9e40e39c39566c435ad92f0659c6120b9f2b8578/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala#L128-L142

1. if the kyuubi instance is unreachable during maintain window.
2. the batch app state has been terminated, and the app stated was backfilled by another kyuubi instance peer, see #2911
3. the batch state in the metadata table is still PENDING/RUNNING
4. return the terminated batch state for such case instead of `PENDING or RUNNING`.

GA and IT.

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
@@ -91,7 +91,7 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
 
   private def sessionManager = fe.be.sessionManager.asInstanceOf[KyuubiSessionManager]
 
-  private def buildBatch(session: KyuubiBatchSession): Batch = {
+  private def buildBatchFromSession(session: KyuubiBatchSession): Batch = {
     val batchOp = session.batchJobSubmissionOp
     val batchOpStatus = batchOp.getStatus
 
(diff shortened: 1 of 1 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -362,7 +362,18 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
...
-          MetadataManager.buildBatch(metadata)
+          if (isApplicationTerminated) {
+            buildBatch(
+              metadata,
+              Some(ApplicationInfo(
+                metadata.engineId,
+                metadata.engineName,
+                metadata.appState.orNull,
+                Option(metadata.engineUrl),
+                metadata.engineError)))
```
turboFei, line 373: https://github.com/apache/kyuubi/blob/9e40e39c39566c435ad92f0659c6120b9f2b8578/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala#L128-L142

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/BatchesResource.scala
```diff
@@ -125,7 +125,7 @@ private[v1] class BatchesResource extends ApiRequestContext with Logging {
       Map.empty[String, String].asJava)
   }
 
-  private def buildBatch(
+  private def buildBatchFromMetadataAndAppInfo(
```
turboFei, line 128: how about now? @pan3793

pan3793 (approved):

PR #7098, opened 2025-06-13
Add ability to annotate pods and headless service in Kyuubi helm chart

jasonj:

Support adding arbitrary annotations to Kyuubi pods and services - for example, those needed for annotation-based auto-discovery via [k8s-monitoring-helm](https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/docs/examples/features/annotation-autodiscovery/default/README.md)

Helm chart installs with and without annotations added

No


yaooqinn:
Merged to master, thank you @jasonj

PR #7055, opened 2025-05-03
Support to filter out server only configs with prefixes

turboFei:
### Why are the changes needed?

To filter out server only configs with prefixes.

For some kyuubi configs, there is no related defined ConfigEntry, and we can not filter out them and have to populate them to engien end.

For example:
```
kyuubi.kubernetes.28.master.address=k8s://master
kyuubi.backend.server.event.kafka.broker=localhost:9092
kyuubi.metadata.store.jdbc.driver=com.mysql.cj.jdbc.Driver
kyuubi.metadata.store.jdbc.datasource.maximumPoolSize=600
kyuubi.metadata.store.jdbc.datasource.minimumIdle=100
kyuubi.metadata.store.jdbc.datasource.idleTimeout=60000
```

This PR supports to exclude them by setting:
```
kyuubi.config.server.only.prefixes=kyuubi.backend.server.event.kafka.,kyuubi.metadata.store.jdbc.datasource.,kyuubi.kubernetes.28.
```

### How was this patch tested?

UT
### Was this patch authored or co-authored using generative AI tooling?

No.

PR #7110, opened 2025-06-23
Fix `serverOnlyPrefixConfigKeys` is iterator issue

turboFei:

Followup for #7055 
Before this PR, the `serverOnlyPrefixConfigKeys` is type of iterator.

After one time iteration, it become empty.

In this PR, we convert it to `Set` to fix this issue.

UT.

No.


pan3793 (approved):

turboFei:
thanks, merging to master and 1.10.3

PR #7121, opened 2025-06-29
Improve operation timeout management with configurable executors

wangzhigang1999:

The current mechanism for handling operation timeouts in Kyuubi creates a new `ScheduledExecutorService` with a dedicated thread for each operation. In scenarios with a large number of concurrent operations, this results in excessive thread creation, which consumes substantial system resources and may adversely affect server performance and stability.

This PR introduces a shared `ScheduledThreadPool` within the Operation Manager to centrally schedule operation timeouts. This approach avoids the overhead of creating an excessive number of threads, thereby reducing the system load. Additionally, both the pool size and thread keep-alive time are configurable via the `OPERATION_TIMEOUT_POOL_SIZE` and `OPERATION_TIMEOUT_POOL_KEEPALIVE_TIME` parameters.

A new unit test for `newDaemonScheduledThreadPool` was added to `ThreadUtilsSuite.scala`. Furthermore, a dedicated `TimeoutSchedulerSuite` was introduced to verify operation timeout behavior.

No.


wangzhigang1999:
@pan3793 Thank you very much for your suggestions and support! I understand your concerns and agree that directly replacing the existing method would be more concise, clearer, and reduce complexity. I will adjust the PR based on your suggestions and update the code as soon as possible. Thank you again for your guidance!

wangzhigang1999:
@pan3793  Hello, thank you very much for your suggestions and support! Based on your feedback, I have simplified the code by completely replacing the mechanism of creating a separate thread for each operation with a unified timeout management implementation based on a thread pool.  I look forward to your further review and guidance, thank you!

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/ThreadPoolTimeoutExecutor.scala
```diff
@@ -0,0 +1,120 @@
...
+  import com.google.common.annotations.VisibleForTesting
+
+  import org.apache.kyuubi.config.KyuubiConf
+
+  @volatile private var _instance: Option[ThreadPoolTimeoutExecutor] = None
+
+  /**
+   * Return the global timeout executor, creating it if necessary (thread-safe).
+   */
+  def getOrCreate(conf: KyuubiConf): ThreadPoolTimeoutExecutor = {
```
pan3793: I understand you create an `object ThreadPoolTimeoutExecutor` and add this `getOrCreate` method to make it resettable for testing, but I feel we can manage the pool lifecycle in the `OperationManager`, thought?

wangzhigang1999: I’m thinking of updating the patch like this:

1. Add a helper in `ThreadUtils` to create a daemon `ScheduledThreadPoolExecutor`.
2. Let `OperationManager` create the pool in `initialize()` and shut it down in `stop()`.
3. Make `AbstractOperation` get the pool from `OperationManager`, so no need for the singleton.

This way the scheduler lifecycle stays inside `OperationManager`, but we still reuse the helper. Does that sound okay to you? Or do you have any other ideas?

pan3793: SGTM

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2212,6 +2212,22 @@ object KyuubiConf {
...
+      .version("1.11.0")
+      .intConf
+      .createWithDefault(8)
+
+  val OPERATION_TIMEOUT_POOL_KEEPALIVE_TIME: ConfigEntry[Long] =
+    buildConf("kyuubi.operation.timeout.pool.keepalive.time")
+      .doc(
+        "The keep-alive time in seconds for idle threads in the shared timeout executor pool.")
+      .version("1.11.0")
+      .longConf
```
pan3793: timeConf

kyuubi-common/src/test/scala/org/apache/kyuubi/operation/timeout/TimeoutSchedulerSuite.scala
```diff
@@ -0,0 +1,54 @@
...
+    val latch = new CountDownLatch(1)
+    val future = manager.scheduleTimeout(
+      new Runnable {
+        override def run(): Unit = latch.countDown()
+      },
+      1)
+
+    // Ensure the task completed
+    assert(latch.await(5, TimeUnit.SECONDS))
+    assert(!future.isCancelled)
```
pan3793, line 46: ~~scalatest support `eventually` for these case~~
misread, this is also fine

kyuubi-common/src/main/scala/org/apache/kyuubi/operation/OperationManager.scala
```diff
@@ -40,13 +43,42 @@ abstract class OperationManager(name: String) extends AbstractService(name) {
...
   def getOperationCount: Int = handleToOperation.size()
 
   def allOperations(): Iterable[Operation] = handleToOperation.values().asScala
 
   override def initialize(conf: KyuubiConf): Unit = {
     LogDivertAppender.initialize(skipOperationLog)
     super.initialize(conf)
+
+    val poolSize = conf.get(KyuubiConf.OPERATION_TIMEOUT_POOL_SIZE)
+    val keepAlive = conf.get(KyuubiConf.OPERATION_TIMEOUT_POOL_KEEPALIVE_TIME)
```
pan3793: let's have a concrete prefix for these vars name, i.e. `timeoutPool`

pan3793 (approved):
It looks much simpler now, LGTM except for minor comments

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2212,6 +2212,22 @@ object KyuubiConf {
       .checkValue(_ >= 1000, "must >= 1s if set")
       .createOptional
 
+  val OPERATION_TIMEOUT_POOL_SIZE: ConfigEntry[Int] =
+    buildConf("kyuubi.operation.timeout.pool.size")
+      .doc("Number of threads in the per-OperationManager timeout scheduler used for " +
+        "operation-level timeout monitoring.")
```
pan3793: we should not expose OperationManager concept to user docs, for user, it's a global pool

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2212,6 +2212,22 @@ object KyuubiConf {
...
+      .version("1.11.0")
+      .intConf
+      .createWithDefault(8)
+
+  val OPERATION_TIMEOUT_POOL_KEEPALIVE_TIME: ConfigEntry[Long] =
+    buildConf("kyuubi.operation.timeout.pool.keepalive.time")
+      .doc(
+        "Keep-alive time for idle threads in the per-OperationManager timeout scheduler.")
```
pan3793: ditto, and can u make the string start at the previous line?

wangzhigang1999: Each tweak is small, but together they make the code easier to read and maintain. Thanks for catching them.

wangzhigang1999:
> @wangzhigang1999 could you update the PR description to match the code? I can merge this PR after then

done

pan3793:
Thanks, merged to master

Issue #7122, opened 2025-06-30
[FEATURE] Spark Hive connector supports ORC hive table pushdown filter


flaming-archer:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

The ORC hive table donnot support pushdown filter yet.  This will greatly affect performance.

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #7123, opened 2025-06-30
[KYUUBI #7122] Support ORC hive table pushdown filter

flaming-archer:

Previously, the `HiveScan` class was used to read data. If it is determined to be ORC type, the `ORCScan` from Spark datasourcev2 can be used. `ORCScan` supports pushfilter down, but `HiveScan` does not yet support it. 

In our testing, we are able to achieve approximately 2x performance improvement.

The conversation can be controlled by setting `spark.sql.kyuubi.hive.connector.read.convertMetastoreOrc`. When enabled, the data source ORC reader is used to process ORC tables created by using the HiveQL syntax, instead of Hive SerDe.

close https://github.com/apache/kyuubi/issues/7122

added unit test

No

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
@@ -37,6 +37,7 @@ import org.apache.spark.sql.hive.kyuubi.connector.HiveBridgeHelper.{BucketSpecHe
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
+import org.apache.kyuubi.spark.connector.hive.KyuubiHiveConnectorConf.READ_CONVERT_METASTORE_ORC
 import org.apache.kyuubi.spark.connector.hive.read.{HiveCatalogFileIndex, HiveScanBuilder}
 import org.apache.kyuubi.spark.connector.hive.write.HiveWriteBuilder
 
@@ -61,14 +62,14 @@ case class HiveTable(
       catalogTable.stats.map(_.sizeInBytes.toLong).getOrElse(defaultTableSize))
   }
 
-  lazy val finalTableType: Option[String] = {
-    val serde = catalogTable.storage.serde.getOrElse("").toLowerCase(Locale.ROOT)
-    val parquet = serde.contains("parquet")
-    val orc = serde.contains("orc")
+  lazy val convertedProvider: Option[String] = {
+    val serde = catalogTable.storage.serde.getOrElse("").toUpperCase(Locale.ROOT)
+    val parquet = serde.contains("PARQUET")
+    val orc = serde.contains("ORC")
     val provider = catalogTable.provider.map(_.toUpperCase(Locale.ROOT))
-    if (orc | provider.contains("ORC")) {
+    if (orc || provider.contains("ORC")) {
       Some("ORC")
-    } else if (parquet | provider.contains("PARQUET")) {
+    } else if (parquet || provider.contains("PARQUET")) {
       Some("PARQUET")
     } else {
       None
@@ -93,8 +94,9 @@ case class HiveTable(
   }
 
   override def newScanBuilder(options: CaseInsensitiveStringMap): ScanBuilder = {
-    finalTableType match {
-      case Some("ORC") => OrcScanBuilder(sparkSession, fileIndex, schema, dataSchema, options)
+    convertedProvider match {
+      case Some("ORC") if sparkSession.sessionState.conf.getConf(READ_CONVERT_METASTORE_ORC) =>
+        OrcScanBuilder(sparkSession, fileIndex, schema, dataSchema, options)
       case _ => HiveScanBuilder(sparkSession, fileIndex, dataSchema, catalogTable)
     }
   }
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
@@ -18,6 +18,7 @@
 package org.apache.kyuubi.spark.connector.hive
 
 import java.util
+import java.util.Locale
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
@@ -31,6 +32,7 @@ import org.apache.spark.sql.connector.catalog.TableCapability.{BATCH_READ, BATCH
 import org.apache.spark.sql.connector.expressions.Transform
 import org.apache.spark.sql.connector.read.ScanBuilder
 import org.apache.spark.sql.connector.write.{LogicalWriteInfo, WriteBuilder}
+import org.apache.spark.sql.execution.datasources.v2.orc.OrcScanBuilder
 import org.apache.spark.sql.hive.kyuubi.connector.HiveBridgeHelper.{BucketSpecHelper, LogicalExpressions}
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
@@ -59,6 +61,20 @@ case class HiveTable(
       catalogTable.stats.map(_.sizeInBytes.toLong).getOrElse(defaultTableSize))
   }
 
+  lazy val finalTableType: Option[String] = {
+    val serde = catalogTable.storage.serde.getOrElse("").toLowerCase(Locale.ROOT)
+    val parquet = serde.contains("parquet")
+    val orc = serde.contains("orc")
+    val provider = catalogTable.provider.map(_.toUpperCase(Locale.ROOT))
+    if (orc | provider.contains("ORC")) {
+      Some("ORC")
+    } else if (parquet | provider.contains("PARQUET")) {
+      Some("PARQUET")
+    } else {
+      None
+    }
+  }
+
   override def name(): String = catalogTable.identifier.unquotedString
 
   override def schema(): StructType = catalogTable.schema
@@ -77,7 +93,10 @@ case class HiveTable(
   }
 
   override def newScanBuilder(options: CaseInsensitiveStringMap): ScanBuilder = {
-    HiveScanBuilder(sparkSession, fileIndex, dataSchema, catalogTable)
+    finalTableType match {
+      case Some("ORC") => OrcScanBuilder(sparkSession, fileIndex, schema, dataSchema, options)
+      case _ => HiveScanBuilder(sparkSession, fileIndex, dataSchema, catalogTable)
+    }
   }
 
   override def newWriteBuilder(info: LogicalWriteInfo): WriteBuilder = {
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/KyuubiHiveConnectorConf.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/KyuubiHiveConnectorConf.scala
@@ -19,6 +19,8 @@ package org.apache.kyuubi.spark.connector.hive
 
 import java.util.Locale
 
+import org.apache.spark.sql.internal.SQLConf.buildConf
+
 object KyuubiHiveConnectorConf {
 
   import org.apache.spark.sql.internal.SQLConf.buildStaticConf
@@ -39,4 +41,12 @@ object KyuubiHiveConnectorConf {
         "Invalid value for 'spark.sql.kyuubi.hive.connector.externalCatalog.share.policy'." +
           "Valid values are 'ONE_FOR_ONE', 'ONE_FOR_ALL'.")
       .createWithDefault(OneForAllPolicy.name)
+
+  val READ_CONVERT_METASTORE_ORC =
+    buildConf("spark.sql.kyuubi.hive.connector.read.convertMetastoreOrc")
+      .doc(s"When set to true, the built-in ORC reader are used to process " +
+        "ORC tables created by using the HiveQL syntax, instead of Hive serde.")
+      .version("1.11.0")
+      .booleanConf
+      .createWithDefault(true)
 }
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveFileIndex.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveFileIndex.scala
@@ -54,6 +54,13 @@ class HiveCatalogFileIndex(
 
   override def partitionSchema: StructType = table.partitionSchema
 
+  override def listFiles(
+      partitionFilters: Seq[Expression],
+      dataFilters: Seq[Expression]): Seq[PartitionDirectory] = {
+    val fileIndex = filterPartitions(partitionFilters)
+    fileIndex.listFiles(partitionFilters, dataFilters)
+  }
+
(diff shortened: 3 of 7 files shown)


extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
```diff
@@ -77,7 +93,10 @@ case class HiveTable(
   }
 
   override def newScanBuilder(options: CaseInsensitiveStringMap): ScanBuilder = {
-    HiveScanBuilder(sparkSession, fileIndex, dataSchema, catalogTable)
+    finalTableType match {
+      case Some("ORC") => OrcScanBuilder(sparkSession, fileIndex, schema, dataSchema, options)
```
pan3793: the conversion is not always compatible, you must have a switch to allow user to disable it.

for the built-in V1 Hive ORC tables, Spark has a `spark.sql.hive.convertMetastoreOrc`, we can have a `spark.sql.kyuubi.hive.connector.read.convertMetastoreOrc`

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
```diff
@@ -59,6 +61,20 @@ case class HiveTable(
       catalogTable.stats.map(_.sizeInBytes.toLong).getOrElse(defaultTableSize))
   }
 
+  lazy val finalTableType: Option[String] = {
+    val serde = catalogTable.storage.serde.getOrElse("").toLowerCase(Locale.ROOT)
+    val parquet = serde.contains("parquet")
+    val orc = serde.contains("orc")
+    val provider = catalogTable.provider.map(_.toUpperCase(Locale.ROOT))
```
pan3793, line 69: why lowercase `serde` but uppercase `provider`? can we have a consistent style?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
```diff
@@ -59,6 +61,20 @@ case class HiveTable(
...
+    if (orc | provider.contains("ORC")) {
```
pan3793: I then you mean `||` instead of `|`

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveQuerySuite.scala
```diff
@@ -260,6 +269,89 @@ class HiveQuerySuite extends KyuubiHiveTest {
...
+      spark,
+      table,
+      s"""
+        CREATE TABLE $table (
+          id INT,
+          data STRING,
+          value INT
+        ) PARTITIONED BY (dt STRING, region STRING)
+       STORED AS ORC
+      """) {
```
pan3793: multi-lines string should be written like

```suggestion
      s"""CREATE TABLE $table (
         |  id INT,
         |  data STRING,
         |  value INT
         |) PARTITIONED BY (dt STRING, region STRING)
         |STORED AS ORC
         |""".stripMargin) {
```

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
```diff
@@ -59,6 +61,20 @@ case class HiveTable(
       catalogTable.stats.map(_.sizeInBytes.toLong).getOrElse(defaultTableSize))
   }
 
+  lazy val finalTableType: Option[String] = {
```
pan3793: `provider` is a term in the Spark data source table

```suggestion
  lazy val convertedProvider: Option[String] = {
```

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveFileIndex.scala
```diff
@@ -54,6 +54,13 @@ class HiveCatalogFileIndex(
 
   override def partitionSchema: StructType = table.partitionSchema
 
+  override def listFiles(
```
pan3793, line 57: why need to override this method?

flaming-archer: 1. If this method is not overridden, it will call the `listFiles` method of the parent class's PartitioningAvareFileIndex, which needs to override the `pathSpec` method. 

      Rewriting this method is a bit difficult because it requires knowing the specific value of the partition. 
      
      See definition   `def partitionSpec(): PartitionSpec`
      And the definitions of PartitionSpec are
      
      `case class PartitionPath(values: InternalRow, path: Path)` , pls notice the **values**.
      
      `case class PartitionSpec(`
      `partitionColumns: StructType,` 
      `partitions: Seq[PartitionPath])`

2. I think hiveFileidex has actually rewritten the listFiles method in HiveFileIndex.scala, so if we reuse it directly, the risk should be lower.

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveCatalogSuite.scala
```diff
@@ -355,8 +357,17 @@ class HiveCatalogSuite extends KyuubiHiveTest {
     val orcProps: util.Map[String, String] = new util.HashMap[String, String]()
     orcProps.put(TableCatalog.PROP_PROVIDER, "orc")
     val ot = catalog.createTable(orc_table, schema, Array.empty[Transform], orcProps)
+
     val orcScan = ot.asInstanceOf[HiveTable]
-      .newScanBuilder(CaseInsensitiveStringMap.empty()).build().asInstanceOf[HiveScan]
+      .newScanBuilder(CaseInsensitiveStringMap.empty()).build().asInstanceOf[OrcScan]
     assert(orcScan.isSplitable(new Path("empty")))
+
+    withSparkSession(Map(READ_CONVERT_METASTORE_ORC.key -> "false")) {
```
pan3793: both true and false should be tested

```
    Seq("true", "false").foreach { value =>
      withSparkSession(Map(READ_CONVERT_METASTORE_ORC.key -> value)) { _ =>
        val orcScan = ot.asInstanceOf[HiveTable]
          .newScanBuilder(CaseInsensitiveStringMap.empty()).build().asInstanceOf[FileScan]
        assert(orcScan.isSplitable(new Path("empty")))
      }
    }
```

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveFileIndex.scala
pan3793: I see, previously the method was not called, now it is passed to `OrcScanBuilder`, and finally will be called by `OrcScan#partitions`, overwriting is required so that filter pruning takes effect to avoid listing unnecessary partition files.

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveQuerySuite.scala
```diff
@@ -52,6 +52,15 @@ class HiveQuerySuite extends KyuubiHiveTest {
     finally spark.sql(s"DROP TABLE $table")
   }
 
+  def withTempPushFilterPartitionedTable(
```
pan3793: this does not look like an eligible encapsulation, let's define

```
import org.apache.spark.sql.hive.kyuubi.connector.HiveBridgeHelper.Utils

abstract class KyuubiHiveTest extends QueryTest with Logging {
...
  /**
   * Drops table `tableName` after calling `f`.
   */
  protected def withTable(tableNames: String*)(f: => Unit): Unit = {
    Utils.tryWithSafeFinally(f) {
      tableNames.foreach { name =>
        spark.sql(s"DROP TABLE IF EXISTS $name")
      }
    }
  }
...
}
```

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
```diff
@@ -77,7 +94,11 @@ case class HiveTable(
   }
 
   override def newScanBuilder(options: CaseInsensitiveStringMap): ScanBuilder = {
-    HiveScanBuilder(sparkSession, fileIndex, dataSchema, catalogTable)
+    convertedProvider match {
+      case Some("ORC") if sparkSession.sessionState.conf.getConf(READ_CONVERT_METASTORE_ORC) =>
```
yikf, line 98: Why not the implementation mechanism of ConvertRelation?

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
```diff
@@ -59,6 +62,20 @@ case class HiveTable(
...
   }
 
+  lazy val convertedProvider: Option[String] = {
+    val serde = catalogTable.storage.serde.getOrElse("").toUpperCase(Locale.ROOT)
+    val parquet = serde.contains("PARQUET")
+    val orc = serde.contains("ORC")
+    val provider = catalogTable.provider.map(_.toUpperCase(Locale.ROOT))
+    if (orc || provider.contains("ORC")) {
+      Some("ORC")
+    } else if (parquet || provider.contains("PARQUET")) {
```
yikf, line 72: Could `provider` possibly be parquet? It seems only can be hive

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveCatalogSuite.scala
yikf: +1

yikf:
@flaming-archer Thanks for your effort, left some comments.

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/KyuubiHiveConnectorConf.scala
```diff
@@ -39,4 +41,12 @@ object KyuubiHiveConnectorConf {
         "Invalid value for 'spark.sql.kyuubi.hive.connector.externalCatalog.share.policy'." +
           "Valid values are 'ONE_FOR_ONE', 'ONE_FOR_ALL'.")
       .createWithDefault(OneForAllPolicy.name)
+
+  val READ_CONVERT_METASTORE_ORC =
+    buildConf("spark.sql.kyuubi.hive.connector.read.convertMetastoreOrc")
+      .doc(s"When set to true, the built-in ORC reader are used to process " +
+        "ORC tables created by using the HiveQL syntax, instead of Hive serde.")
```
pan3793:
```suggestion
      .doc("When enabled, the data source ORC reader is used to process " +
        "ORC tables created by using the HiveQL syntax, instead of Hive SerDe.")
```

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveQuerySuite.scala
pan3793: can be removed then?

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveQuerySuite.scala
```diff
@@ -260,6 +269,90 @@ class HiveQuerySuite extends KyuubiHiveTest {
     }
   }
 
+  test("ORC filter pushdown") {
+    val table = "hive.default.orc_filter_pushdown"
+    withTable(table) {
+      spark.sql(s"""
```
pan3793: can you move the String definition to the next line, to avoid having a huge leading blank block?

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveCatalogSuite.scala
```diff
@@ -355,8 +357,17 @@ class HiveCatalogSuite extends KyuubiHiveTest {
...
-      .newScanBuilder(CaseInsensitiveStringMap.empty()).build().asInstanceOf[HiveScan]
-    assert(orcScan.isSplitable(new Path("empty")))
+
+    Seq("true", "false").foreach { value =>
+      withSparkSession(Map(READ_CONVERT_METASTORE_ORC.key -> value)) { _ =>
+        val scan = ot.asInstanceOf[HiveTable]
+          .newScanBuilder(CaseInsensitiveStringMap.empty()).build()
+        val orcScan = value match {
+          case "true" => scan.asInstanceOf[OrcScan]
+          case "false" => scan.asInstanceOf[HiveScan]
```
pan3793: rewrite with `assert`

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
flaming-archer: Actually, I copy from here , maybe parquet?

https://github.com/apache/kyuubi/blob/4e40f9457dfcde9b2748eb6da877e1ee32193fca/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/read/HiveScan.scala#L57-L59

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
pan3793: that requires more work and is more invasive compared to the current approach, let's try this simple way first, and the extended rule approach can be evaluated later

extensions/spark/kyuubi-spark-connector-hive/src/test/scala/org/apache/kyuubi/spark/connector/hive/HiveCatalogSuite.scala
```diff
@@ -355,8 +357,26 @@ class HiveCatalogSuite extends KyuubiHiveTest {
...
+          case "true" =>
+            assert(
+              scan.isInstanceOf[OrcScan],
+              s"Expected OrcScan, got ${scan.getClass.getSimpleName}")
+            scan.asInstanceOf[OrcScan]
+          case "false" =>
+            assert(
+              scan.isInstanceOf[HiveScan],
+              s"Expected HiveScan, got ${scan.getClass.getSimpleName}")
+            scan.asInstanceOf[HiveScan]
```
pan3793, line 376: we need a `case _` branch

pan3793:
also cc @cfmcgrady

pan3793:
will merge it this afternoon if no further comments

pan3793:
Thanks, merged to master

Issue #7109, opened 2025-06-19
[Bug] Ignore the ? in backticks


ruanwenjun:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

https://github.com/apache/kyuubi/issues/6221 has fixed some case, but there still exist bug.

sql:
```SQL
set hive.support.quoted.identifiers=none

SELECT
    regexp_replace(col2, '\n|\r|\t', '') as col2,
    `(col2|col2)?+.+`
FROM
    test_table
```

`(col2|col2)?+.+` should as 

```code
String sql = "SELECT\n" +
                "    regexp_replace(col2, '\\n|\\r|\\t', '') as col2,\n" +
                "    `(col2|col2)?+.+`\n" +
                "FROM\n" +
                "    test_table";
Utils.updateSql(sql, new HashMap<>());
```

got exception
```java
Exception in thread "main" org.apache.kyuubi.jdbc.hive.KyuubiSQLException: Parameter #1 is unset
	at org.apache.kyuubi.jdbc.hive.Utils.updateSql(Utils.java:170)
	at org.apache.kyuubi.jdbc.hive.Test.main(Test.java:14)
```

### Affects Version(s)

1.10

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #7125, opened 2025-07-02
[KYUUBI #7109] Ignore the ? in backticks

ruanwenjun:
We will split the sql by `?` when we use `KyuubiPreparedStatement`. But there exist corner case when ? exist in backticks.
For example, below sql contains `?`, but we shouldn't split it by `?`.
```sql
SELECT `(ds|hr)?+.+` FROM sales
```
More details can find at https://hive.apache.org/docs/latest/languagemanual-select_27362043/#regex-column-specification

Hive upstream fix - HIVE-29060

UT.

NO.

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
@@ -124,6 +124,7 @@ public class Utils {
     boolean inSingleQuote = false;
     boolean inDoubleQuote = false;
     boolean inComment = false;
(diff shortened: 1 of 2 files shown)


pan3793:
Kyuubi JDBC Driver is derived from Hive JDBC Driver, was this fixed in Hive project?

ruanwenjun:
> Kyuubi JDBC Driver is derived from Hive JDBC Driver, was this fixed in Hive project?

Hive doesn't have a fix for this. 😢

pan3793:
let me take a look at how `spark-sql` processes this.

pan3793:
Thanks, merged to master/1.10

PR #7130, opened 2025-07-09
[KYUUBI #7129] Support PARQUET hive table pushdown filter

flaming-archer:

Previously, the `HiveScan` class was used to read data. If it is determined to be PARQUET type, the `ParquetScan` from Spark datasourcev2 can be used. `ParquetScan` supports pushfilter down, but `HiveScan` does not yet support it.

The conversation can be controlled by setting `spark.sql.kyuubi.hive.connector.read.convertMetastoreParquet`. When enabled, the data source PARQUET reader is used to process PARQUET tables created by using the HiveQL syntax, instead of Hive SerDe.

close https://github.com/apache/kyuubi/issues/7129

added unit test

No


flaming-archer:
@cfmcgrady please take a look at it.

flaming-archer:
In our testing, the performance can be improved by 1.4 times

PR #7132, opened 2025-07-09
Respect `kyuubi.session.engine.startup.waitCompletion` for wait engine completion

turboFei:

We should not fail the batch submission if the submit process is alive and wait engine completion is false.

Especially for spark on kubernetes, the app might failed with NOT_FOUND state if the spark submit process running time more than the submit timeout.

In this PR, if the `kyuubi.session.engine.startup.waitCompletion` is false, when getting the application info, it use current timestamp as submit time to prevent the app failed with NOT_FOUND state due to submit timeout.

Pass current GA and manually testing.

No.


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -292,13 +292,23 @@ class BatchJobSubmission(
     try {
       info(s"Submitting $batchType batch[$batchId] job:\n$builder")
       val process = builder.start
-      while (process.isAlive && !applicationFailed(_applicationInfo, appOperation)) {
+
+      // continue polling if process is alive and application not started yet
+      // or application is not failed
+      while (process.isAlive && (
+          applicationId(_applicationInfo).isEmpty ||
+            !applicationFailed(_applicationInfo, appOperation))) {
```
pan3793: what do you mean "application not started yet"? Pod is not started, or Pod can not be found (no matter the status)?

for the latter case, previously we introduced `kyuubi.engine.kubernetes.submit.timeout` to allow Kyuubi wait for more time

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -3473,7 +3473,7 @@ object KyuubiConf {
         "after `spark-submit` is returned.")
       .version("1.7.1")
       .timeConf
-      .createWithDefaultString("PT30S")
+      .createWithDefaultString("PT5M")
```
turboFei: It is too short now, we shall enlarge it for production.

pan3793: I think the default value great than `kyuubi.session.engine.initialize.timeout` (3 min) does not make sense

pan3793: in most time, 30s is sufficent to make app visible on YARN / K8s

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala
```diff
@@ -372,6 +372,25 @@ class SparkProcessBuilder(
...
+      .getOrElse(KyuubiConf.SESSION_ENGINE_STARTUP_WAIT_COMPLETION.defaultValStr)
+      .toBoolean
+  }
+
+  def engineWaitCompletionConf(): Map[String, String] =
+    clusterManager().map(_.toLowerCase(Locale.ROOT)) match {
+      case Some(m) if m.startsWith("yarn") =>
+        Map(YARN_SUBMIT_WAIT_APP_COMPLETION -> waitEngineCompletion.toString)
+      case Some(m) if m.startsWith("k8s") =>
+        Map(KUBERNETES_SUBMISSION_WAIT_APP_COMPLETION -> waitEngineCompletion.toString)
```
pan3793, line 392: this is a behavior change, could you document it on migration guide?

turboFei: sure

turboFei: added

turboFei:
How about now? @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -79,7 +79,18 @@ class BatchJobSubmission(
   def appStartTime: Long = _appStartTime
   def appStarted: Boolean = _appStartTime > 0
 
-  private lazy val _submitTime = if (appStarted) _appStartTime else System.currentTimeMillis
+  private lazy val _submitTime = System.currentTimeMillis
+  private lazy val waitEngineCompletion = builder.waitEngineCompletion
+  private def submitTime: Long = if (appStarted) {
```
pan3793, line 95: could you add comments to clarify the meaning of `appStartTime`, `appStarted`, `submitTime`?

kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -79,7 +79,23 @@ class BatchJobSubmission(
...
+
+  /**
+   * If the application has been submitted (applicationId is not empty), return the application
+   * start time. Otherwise, if the builder process does not wait for the engine completion and
+   * the process is not terminated, return the current time as the submit time to prevent
+   * application failed within NOT_FOUND state due to the process duration exceeds the engine submit
+   * timeout. Otherwise, return the time that start to run the batch job submission operation.
+   * @return the time that start to count the application submit timeout.
+   */
+  private def submitTime: Long = if (appStarted) {
```
turboFei, line 95: updated the method description @pan3793

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/YarnApplicationOperation.scala
```diff
@@ -131,19 +131,20 @@ class YarnApplicationOperation extends ApplicationOperation with Logging {
       proxyUser: Option[String] = None,
       submitTime: Option[Long] = None): ApplicationInfo = withYarnClient(proxyUser) { yarnClient =>
     debug(s"Getting application info from YARN cluster by tag: $tag")
+    val startTime = System.currentTimeMillis()
```
turboFei, line 134: this is used to prevent the elapsedTime too long caused by yarn client stuck, especially for the waitAppCompletion=false use case, it would cause the application failure.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -292,27 +292,34 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
       throw new IllegalStateException("Methods initialize and isSupported must be called ahead")
     }
     debug(s"Getting application[${toLabel(tag)}]'s info from Kubernetes cluster")
+    val startTime = System.currentTimeMillis()
```
turboFei, line 295: this is used to prevent the elapsedTime too long caused by metadataManager or other slowness, which might cause the app failure for waitAppCompletion=false

PR #7138, opened 2025-07-13
Support Per-Client Initialization SQL in Shared Engine Mode

1358035421:
image
**Current Behavior:**

When "kyuubi.engine.share.level = USER/GROUP/SERVER", the first client (Client A) calling openSession creates a Kyuubi-Spark-SQL-Engine (Spark Driver), where the initialization SQL configured in "kyuubi.engine.spark.initialize.sql" takes effect.

Subsequent clients (e.g., Client B) connecting via openSession will reuse the existing Kyuubi-Spark-SQL-Engine (Spark Driver) created in step 1, where the initialization SQL configured in "kyuubi.engine.spark.initialize.sql" becomes ineffective.

**Why This Capability Is Needed:**

Currently, kyuubi.engine.spark.initialize.sql only applies to the first openSession client. All subsequent SQL operations inherit the initialization SQL configuration from the first client (this appears to be a potential bug).

Client A may need to set "USE dbA" in its current SQL context, while Client B may need "USE dbB" in its own context - such scenarios should be supported.

Tested on local Kyuubi/Spark cluster. No existing unit tests cover this scenario. Please point me to any relevant tests so I can add them

No

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
@@ -105,7 +105,7 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
     userIsolatedSparkSessionThread.foreach(_.shutdown())
   }
 
-  private def getOrNewSparkSession(user: String): SparkSession = {
+  private def getOrNewSparkSession(user: String, sessionConf: Map[String, String]): SparkSession = {
     if (singleSparkSession) {
       spark
     } else {
@@ -113,8 +113,8 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
         // it's unnecessary to create a new spark session in connection share level
         // since the session is only one
         case CONNECTION => spark
-        case USER => newSparkSession(spark)
-        case GROUP | SERVER if userIsolatedSparkSession => newSparkSession(spark)
+        case USER => newSparkSession(spark, sessionConf)
+        case GROUP | SERVER if userIsolatedSparkSession => newSparkSession(spark, sessionConf)
         case GROUP | SERVER =>
           userIsolatedCacheLock.synchronized {
             if (userIsolatedCache.containsKey(user)) {
@@ -123,7 +123,7 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
               userIsolatedCache.get(user)
             } else {
               userIsolatedCacheCount.put(user, (1, System.currentTimeMillis()))
-              val newSession = newSparkSession(spark)
+              val newSession = newSparkSession(spark, sessionConf)
               userIsolatedCache.put(user, newSession)
               newSession
             }
@@ -132,11 +132,16 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
     }
   }
 
-  private def newSparkSession(rootSparkSession: SparkSession): SparkSession = {
+  private def newSparkSession(
+      rootSparkSession: SparkSession,
+      sessionConf: Map[String, String]): SparkSession = {
     val newSparkSession = rootSparkSession.newSession()
     KyuubiSparkUtil.initializeSparkSession(
       newSparkSession,
-      conf.get(ENGINE_SESSION_SPARK_INITIALIZE_SQL))
+      conf.get(ENGINE_SESSION_SPARK_INITIALIZE_SQL)) ++
+         sessionConf.get(ENGINE_SESSION_SPARK_INITIALIZE_SQL.key).filter(_.nonEmpty)
+           .map(_.split(";").toSeq
+           .getOrElse(Nil))
     newSparkSession
   }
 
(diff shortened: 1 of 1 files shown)


1358035421:
> We have another config `kyuubi.session.engine.spark.initialize.sql` for session level

The kyuubi.session.engine.spark.initialize.sql configuration is only executed when the kyuubi-spark-sql-engine (Spark driver) starts for the first time.
When connecting new sessions to an already running kyuubi-spark-sql-engine(Spark driver）, this configuration does not execute the initialization SQL as expected.
You can refer to the illustration above for more details.

pan3793:
You write `kyuubi.engine.spark.initialize.sql` in the PR description, which is indeed for engine level, I mean `kyuubi.session.engine.spark.initialize.sql`, there are two different configs. Did I misread something?

1358035421:
> You write `kyuubi.engine.spark.initialize.sql` in the PR description, which is indeed for engine level, I mean `kyuubi.session.engine.spark.initialize.sql`, there are two different configs. Did I misread something?

Thank you for your comment and for catching this mistake.

You are correct, there are two different configurations: `kyuubi.engine.spark.initialize.sql`  and `kyuubi.session.engine.spark.initialize.sql`. In my previous PR description and image, I mistakenly wrote `kyuubi.engine.spark.initialize.sql` instead of `kyuubi.session.engine.spark.initialize.sql`. I apologize for the confusion this may have caused.

I have now updated the PR description and related image to use the correct configuration key, kyuubi.session.engine.spark.initialize.sql.

In addition, I would like to clarify that currently the kyuubi.session.engine.spark.initialize.sql configuration does not work as expected

Thank you again for your careful review.

1358035421:
https://github.com/apache/kyuubi/blob/master/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala#L135-L141

You can take a look at this code and see if my understanding is correct. This code runs on the remote kyuubi-spark-sql-engine (spark driver), the `conf` here is actually the `KyuubiConf` generated by `spark-submit` when the first session is created. As a result, when a new session is created with the `kyuubi.session.engine.spark.initialize.sql` configuration, the `conf` still does not contain the new session’s configuration.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSQLSessionManager.scala
```diff
@@ -132,11 +132,17 @@ class SparkSQLSessionManager private (name: String, spark: SparkSession)
...
+      rootSparkSession: SparkSession,
+      sessionConf: Map[String, String]): SparkSession = {
     val newSparkSession = rootSparkSession.newSession()
     KyuubiSparkUtil.initializeSparkSession(
       newSparkSession,
-      conf.get(ENGINE_SESSION_SPARK_INITIALIZE_SQL))
+      sessionConf.get(ENGINE_SESSION_SPARK_INITIALIZE_SQL.key)
+        .filter(_.nonEmpty)
+        .map(_.split(";")
+          .toSeq)
```
pan3793:
```suggestion
        .map(_.split(";").toSeq)
```

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/session/MultiSessionSuiteInitSQLSuite.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  override def withKyuubiConf: Map[String, String] = {
+    Map(
+      ENGINE_SHARE_LEVEL.key -> "SERVER",
+      ENGINE_SINGLE_SPARK_SESSION.key -> "false",
+      (
+        ENGINE_SESSION_SPARK_INITIALIZE_SQL.key,
+        "CREATE DATABASE IF NOT EXISTS INIT_DB_SOLO;" +
+          "CREATE TABLE IF NOT EXISTS INIT_DB_SOLO.test(a int) USING CSV;" +
+          "INSERT INTO INIT_DB_SOLO.test VALUES (2);"))
```
pan3793: this looks too overkill, can we use a simple `SET a=1` and check if the `abc` is set correctly

docs/deployment/migration-guide.md
```diff
@@ -21,6 +21,8 @@
 
 * Since Kyuubi 1.11, the configuration `spark.sql.watchdog.forcedMaxOutputRows` provided by Kyuubi Spark extension is removed, consider using `kyuubi.operation.result.max.rows` instead. Note, the latter works without requirement of installing Kyuubi Spark extension.
 
+* Since Kyuubi 1.11, the configuration `kyuubi.session.engine.spark.initialize.sql` is now applied for every session in shared engines (USER/GROUP/SERVER), correcting previous behavior where it was only applied for the first session.
```
pan3793: should it be "the configuration ... **set by client**"? I think the server-side conf (present in `kyuubi-defaults.conf`) also works previously, right?

pan3793: nit: append new record at the end if there is no special reason

pan3793:
Thanks, merged to master

PR #7141, opened 2025-07-18
Support to get spark app url with pattern `http://{{SPARK_DRIVER_POD_IP}}:{{SPARK_UI_PORT}}`

turboFei:

We are using [virtual-kubelet](https://github.com/virtual-kubelet/virtual-kubelet) for spark on kubernetes, and spark kubernetes pods would be allocated across kubernetes clusters.

And we use the driver POD ip as driver host, see https://github.com/apache/spark/pull/40392, which is supported since spark-3.5.

The kubernetes context and namespace are virtual and we can not build the app URL by spark driver svc.

And the spark driver pod IP is accessible for our use case, so raise this PR to build the spark app url by spark driver pod id and spark ui port.

UT.

image
image

No.


kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -474,6 +485,11 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
             id = getPodAppId(pod),
             name = getPodAppName(pod),
             state = appState,
+            url = appInfo.url.orElse(getPodAppUrl(
+              sparkAppUrlSource,
+              sparkAppUrlPattern,
+              kubernetesInfo,
+              pod)),
```
pan3793:
```suggestion
            url = appInfo.url.orElse {
              getPodAppUrl(sparkAppUrlSource, sparkAppUrlPattern, kubernetesInfo, pod)
            },
```

pan3793:
LGTM, one question, SPARK-42769 is only available since Spark 3.5, do we write something in docs for this requirements?

turboFei:
Have updated the PR desc.

```
And we use the driver POD ip as driver host, see https://github.com/apache/spark/pull/40392, which is supported since spark-3.5.
```

pan3793 (approved):
I see, Pod IP is retrieved by K8s client

PR #7143, opened 2025-07-19
[WIP][KYUUBI #3723]Add SQL session state management and save metadata

z1131392774:
- Add a new SessionState enum to define possible session states.
- Insert or update metadata at different stages of the session lifecycle.

#3723

No,it's just a draft.

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
@@ -27,13 +27,16 @@ import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf._
 import org.apache.kyuubi.config.KyuubiConf.EngineOpenOnFailure._
 import org.apache.kyuubi.config.KyuubiReservedKeys.{KYUUBI_ENGINE_CREDENTIALS_KEY, KYUUBI_SESSION_HANDLE_KEY, KYUUBI_SESSION_SIGN_PUBLICKEY, KYUUBI_SESSION_USER_SIGN}
-import org.apache.kyuubi.engine.{EngineRef, KyuubiApplicationManager}
+import org.apache.kyuubi.engine.{EngineRef, EngineType, KyuubiApplicationManager}
+import org.apache.kyuubi.engine.spark.SparkProcessBuilder
 import org.apache.kyuubi.events.{EventBus, KyuubiSessionEvent}
 import org.apache.kyuubi.ha.client.DiscoveryClientProvider._
 import org.apache.kyuubi.ha.client.ServiceNodeInfo
 import org.apache.kyuubi.operation.{Operation, OperationHandle}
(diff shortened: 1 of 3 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -73,6 +77,39 @@ class KyuubiSessionImpl(
     case (key, value) => sessionConf.set(key, value)
   }
 
+  val (clusterManager, kubernetesInfo) = engine.engineType match {
```
turboFei, line 80: it is not accurate here

turboFei: For the `clusterManager`, I think you can leave it empty and update it after getting the engine info.

But seems it does not matter.

turboFei: FYI: https://github.com/apache/kyuubi/blob/5f4b1f0de513ad1085dcea8ab3d4911a82733b9e/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala#L311-L312

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -326,4 +381,40 @@ class KyuubiSessionImpl(
...
+  private def isTimedOut: Boolean = {
+    lastAccessTime + sessionIdleTimeoutThreshold <= System.currentTimeMillis &&
+    getNoOperationTime > sessionIdleTimeoutThreshold
+  }
+
+  private def setState(newState: SessionState): Unit = {
+    SessionState.validateTransition(state, newState)
+    newState match {
+      case SessionState.RUNNING =>
+        info(s"Session is ready.")
```
turboFei, line 394: this kind of log is meaningless without session identifier

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -326,4 +381,40 @@ class KyuubiSessionImpl(
...
+      case SessionState.ERROR | SessionState.TIMEOUT | SessionState.CLOSED =>
+        info(s"Session is terminated.")
```
turboFei, line 396: ditto

turboFei: and please log the state

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
z1131392774: Got it, thanks for the guidance! I will get this updated.

PR #7144, opened 2025-07-21
Remove direct usage of `sun.misc.Signal`

pan3793:

This PR replaces the `sun.misc.Signal` with the Kyuubi wrapped one, see https://github.com/apache/kyuubi-shaded/pull/64, which allows Kyuubi to use any of Java 8+ to compile with `-release:8` while still ensuring compatibility with Java 8.

Pass GHA.

Local tested by building against JDK 21, running on JDK 8, everything works as expected.

No.

### DIFF
--- a/kyuubi-common/pom.xml
+++ b/kyuubi-common/pom.xml
@@ -109,6 +109,11 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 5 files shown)


bowenliang123 (approved):
LGTM

PR #7148, opened 2025-07-21
Fix spark.kubernetes.file.upload.path permission

pan3793:

The default behavior of HDFS is to set the permission of a file created with `FileSystem.create` or `FileSystem.mkdirs` to `(P & ^umask)`, where `P` is the permission in the API call and umask is a system value set by `fs.permissions.umask-mode` and defaults to `0022`. This means, with default settings, any mkdirs call can have at most `755` permissions.

The same issue also got reported in [SPARK-30860](https://issues.apache.org/jira/browse/SPARK-30860)

Manual test.

No.


cxzl25 (approved):

pan3793:
Thanks, merged to master

Issue #7153, opened 2025-07-22
[Bug] Server OOM due to nested Zookeeper Jaas Client Configurations


Z1Wu:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

There are  lots of nested Jass Configuration in heapdump.

<img width="1636" height="687" alt="Image" src="https://github.com/user-attachments/assets/8df3f168-8f6e-4ea7-afd5-b8ac36a86323" />

Server heap keeps increasing.

<img width="1114" height="256" alt="Image" src="https://github.com/user-attachments/assets/a7e6d920-2d87-4b95-a93c-b7e5f9e87b7f" />

When invoking `org.apache.kyuubi.ha.client.DiscoveryClientProvider$#withDiscoveryClient` , a new zookeeper client will be created with `JaasConf` whose `baseConfig` refers to previous-created JassConf.

``` scala 
def setupZkAuth(): Unit = (conf.get(HA_ZK_AUTH_PRINCIPAL), getKeyTabFile(conf)) match {
      case (Some(principal), Some(keytab)) if UserGroupInformation.isSecurityEnabled =>
        if (!new File(keytab).exists()) {
          throw new IOException(s"${HA_ZK_AUTH_KEYTAB.key}: $keytab does not exists")
        }
        System.setProperty("zookeeper.sasl.clientconfig", "KyuubiZooKeeperClient")
        conf.get(HA_ZK_AUTH_SERVER_PRINCIPAL).foreach { zkServerPrincipal =>
          // ZOOKEEPER-1467 allows configuring SPN in client
          System.setProperty("zookeeper.server.principal", zkServerPrincipal)
        }
        val zkClientPrincipal = KyuubiHadoopUtils.getServerPrincipal(principal)
        // HDFS-16591 makes breaking change on JaasConfiguration
        val jaasConf = DynConstructors.builder()
          .impl( // Hadoop 3.3.5 and above
            "org.apache.hadoop.security.authentication.util.JaasConfiguration",
            classOf[String],
            classOf[String],
            classOf[String])
          .impl( // Hadoop 3.3.4 and previous
            // scalastyle:off
            "org.apache.hadoop.security.token.delegation.ZKDelegationTokenSecretManager$JaasConfiguration",
[...]

PR #7154, opened 2025-07-22
[KYUUBI #7153]  Share JAAS configuration for Zookeeper client to avoid server OOM

Z1Wu:

Sharing jaas configuration for zookeeper client with same keytab and principal to avoid server oom due to nested jaas configuration. 

fix issue https://github.com/apache/kyuubi/issues/7153

ut

no

### DIFF
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
@@ -40,7 +40,7 @@ import org.apache.kyuubi.util.reflect.DynConstructors
 object ZookeeperClientProvider extends Logging {
 
   /**
-   * Sharing JAAS configuration for Zookeeper client with same keytab and principal to
+   * Share JAAS configuration for Zookeeper client with same keytab and principal to
    * avoid server OOM due to each new JAAS configuration references the previous instance.
    * See KYUUBI #7154 for more details.
    */
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
@@ -40,8 +40,9 @@ import org.apache.kyuubi.util.reflect.DynConstructors
 object ZookeeperClientProvider extends Logging {
 
   /**
-   * Sharing jaas configuration for zookeeper client with same keytab and principal to
-   * avoid server oom due to nested jaas configuration. see KYUUBI #7154 for more details.
+   * Sharing JAAS configuration for Zookeeper client with same keytab and principal to
+   * avoid server OOM due to each new JAAS configuration references the previous instance.
+   * See KYUUBI #7154 for more details.
    */
   val jaasConfigurationCache = new ConcurrentHashMap[(String, String), Configuration]()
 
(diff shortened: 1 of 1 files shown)


kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
```diff
@@ -114,21 +117,26 @@ object ZookeeperClientProvider extends Logging {
         }
         val zkClientPrincipal = KyuubiHadoopUtils.getServerPrincipal(principal)
         // HDFS-16591 makes breaking change on JaasConfiguration
```
pan3793: nit: remove this line

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
```diff
@@ -38,6 +39,8 @@ import org.apache.kyuubi.util.reflect.DynConstructors
 
 object ZookeeperClientProvider extends Logging {
 
+  val jaasConfigurationCache = new ConcurrentHashMap[(String, String), Configuration]()
```
pan3793, line 47: can you add comment here to explain the benefit of caching the JaasConfiguration?

pan3793:
@Z1Wu I see your point now, let's continue with your orignal proposal. TBH, some JDK APIs related to security is a mess.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
Z1Wu: fixed

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
Z1Wu: fixed

Z1Wu:
> @Z1Wu does Zookeeper allow using non-global configuration? if so, we don't need to hack it in that way ...

AFAIK, there is not such method.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
```diff
@@ -38,6 +39,12 @@ import org.apache.kyuubi.util.reflect.DynConstructors
 
 object ZookeeperClientProvider extends Logging {
 
+  /**
+   * Sharing jaas configuration for zookeeper client with same keytab and principal to
+   * avoid server oom due to nested jaas configuration. see KYUUBI #7154 for more details.
+   */
```
pan3793:
```suggestion
  /**
   * Sharing JAAS configuration for Zookeeper client with same keytab and principal to
   * avoid server OOM due to each new JAAS configuration references the previous instance.
   * See KYUUBI #7154 for more details.
   */
```

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperClientProvider.scala
```diff
@@ -38,6 +39,13 @@ import org.apache.kyuubi.util.reflect.DynConstructors
 
 object ZookeeperClientProvider extends Logging {
 
+  /**
+   * Sharing JAAS configuration for Zookeeper client with same keytab and principal to
```
pan3793:
```suggestion
   * Share JAAS configuration for Zookeeper client with same keytab and principal to
```

PR #7158, opened 2025-07-27
Spark engine respects session-level idle timeout threshold

1358035421:
Fixes the same class of issue as 
https://github.com/apache/kyuubi/pull/7138

Previously, `sessionIdleTimeoutThreshold` was initialized only once during session creation using `sessionManager.getConf`, preventing dynamic updates when clients pass new configurations during connection.

we now:
- Allow clients to set session-specific kyuubi.session.idle.timeout` during connection
- Dynamically adjust idle timeout per session
- Prevent connection pile-up by timely recycling idle sessions

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala
@@ -56,7 +56,12 @@ abstract class AbstractSession(
     if (lastIdleTime > 0) System.currentTimeMillis() - _lastIdleTime else 0
   }
 
-  override val sessionIdleTimeoutThreshold: Long = sessionManager.getConf.get(SESSION_IDLE_TIMEOUT)
+  override val sessionIdleTimeoutThreshold: Long = {
+    conf.get(SESSION_IDLE_TIMEOUT.key)
+      .map(_.toLong)
+      .getOrElse(
+        sessionManager.getConf.get(SESSION_IDLE_TIMEOUT))
+  }
 
   val normalizedConf: Map[String, String] = sessionManager.validateAndNormalizeConf(conf)
 
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala
@@ -56,12 +56,7 @@ abstract class AbstractSession(
     if (lastIdleTime > 0) System.currentTimeMillis() - _lastIdleTime else 0
   }
 
-  override val sessionIdleTimeoutThreshold: Long = {
-    conf.get(SESSION_IDLE_TIMEOUT.key)
-      .map(_.toLong)
-      .getOrElse(
-        sessionManager.getConf.get(SESSION_IDLE_TIMEOUT))
-  }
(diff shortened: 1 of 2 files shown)


1358035421:
Hi @turboFei,

Thank you for your feedback. I agree that from the Kyuubi service side, it's reasonable to restrict the session idle timeout set by clients, in order to protect the overall service stability.

However, in the Spark scenario, the value of kyuubi.session.idle.timeout is actually determined by the first user who creates the Spark driver pod. So in practice, it is already user-controlled at that point.

Given this, I think the logic to restrict or validate the session idle timeout should be handled specifically in
class SparkSQLSessionManager private (name: String, spark: SparkSession) extends SessionManager(name)
rather than changing the general behavior in the SessionManager base class.

This way, we can address the Spark-specific behavior without affecting other engines or the general session management logic.

Let me know if you agree with this direction, or if you have other suggestions!

kyuubi-common/src/main/scala/org/apache/kyuubi/session/AbstractSession.scala
```diff
@@ -56,7 +56,12 @@ abstract class AbstractSession(
     if (lastIdleTime > 0) System.currentTimeMillis() - _lastIdleTime else 0
   }
 
-  override val sessionIdleTimeoutThreshold: Long = sessionManager.getConf.get(SESSION_IDLE_TIMEOUT)
+  override val sessionIdleTimeoutThreshold: Long = {
+    conf.get(SESSION_IDLE_TIMEOUT.key)
+      .map(_.toLong)
+      .getOrElse(
+        sessionManager.getConf.get(SESSION_IDLE_TIMEOUT))
+  }
```
turboFei: How about override this method in https://github.com/apache/kyuubi/blob/master/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSession.scala?


or KyuubiSessionImpl?
```
override val sessionIdleTimeoutThreshold: Long = sessionManager.getConf.get(SESSION_IDLE_TIMEOUT)
```

turboFei: so that we can keep the behavior in kyuubi server end.

1358035421: Thank you very much for your reply and suggestions, @turboFei!

Your comments reminded me that it would actually be more appropriate to make the changes in
(https://github.com/apache/kyuubi/blob/master/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/session/SparkSessionImpl.scala)
so that the behavior only affects Spark, and does not impact other engines.

I haven't participated in the implementation or testing of other engines, so I'm not sure whether they have similar issues related to session idle timeout.
Do you know if other engines also allow users to override session idle timeout in a way that could affect their stability, or is this only specific to Spark?

Thanks again for your advice!

turboFei: In my option, the engine session idle timeout is open for users, and the kyuubi session idle timeout should be only set by kyuubi admin.

So, it is fine for me to ether apply this change to spark engine or other engines.

turboFei (approved):
LGTM

pan3793 (approved):

PR #7160, opened 2025-07-29
Use single try-with-resources blocks in docs example

jagij:
Gets rid of the nested try-with-resources blocks. Improves legibility slightly.

Make the example code just a bit nicer.

It still compiles. I created a simple (maven) project using only these two java files.

No


turboFei (approved):

cxzl25 (approved):

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #7166, opened 2025-08-06
Bump arrow version to fix CVE-2024-52338

turboFei:

To fix CVE, https://nvd.nist.gov/vuln/detail/CVE-2024-52338

GA.

No.

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkArrowbasedOperationSuite.scala
+++ b/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkArrowbasedOperationSuite.scala
@@ -23,6 +23,7 @@ import java.util.Locale
 
 import scala.util.Try
 
+import org.apache.commons.lang3.{JavaVersion, SystemUtils}
 import org.apache.spark.{KyuubiSparkContextHelper, TaskContext}
 import org.apache.spark.scheduler.{SparkListener, SparkListenerJobStart}
 import org.apache.spark.sql.{QueryTest, Row, SparkSession}
@@ -60,6 +61,8 @@ class SparkArrowbasedOperationSuite extends WithSparkSQLEngine with SparkDataTyp
   override def resultFormat: String = "arrow"
 
   override def beforeEach(): Unit = {
+    assume(SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_11), "Test requires Java 11 or later")
+
     super.beforeEach()
     withJdbcStatement() { statement =>
       checkResultSetFormat(statement, "arrow")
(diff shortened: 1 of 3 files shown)


turboFei:
```
Caused by: java.lang.NoSuchMethodError: java.nio.ByteBuffer.position(I)Ljava/nio/ByteBuffer;
	at org.apache.arrow.memory.ArrowBuf.setBytes(ArrowBuf.java:902)
	at org.apache.arrow.vector.BaseVariableWidthVector.setSafe(BaseVariableWidthVector.java:1193)
	at org.apache.spark.sql.execution.arrow.StringWriter.setValue(ArrowWriter.scala:256)
	at org.apache.spark.sql.execution.arrow.ArrowFieldWriter.write(ArrowWriter.scala:135)
	at org.apache.spark.sql.execution.arrow.ArrowWriter.write(ArrowWriter.scala:100)
	at org.apache.spark.sql.execution.arrow.KyuubiArrowConverters$ArrowBatchIterator.$anonfun$next$1(KyuubiArrowConverters.scala:291)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally(SparkErrorUtils.scala:64)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally$(SparkErrorUtils.scala:61)
```

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkArrowbasedOperationSuite.scala
```diff
@@ -60,6 +61,8 @@ class SparkArrowbasedOperationSuite extends WithSparkSQLEngine with SparkDataTyp
   override def resultFormat: String = "arrow"
 
   override def beforeEach(): Unit = {
+    assume(SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_11), "Test requires Java 11 or later")
```
turboFei, line 64: arrow is experimental in kyuubi, I think it is fine to only have testing on JDK17

pan3793: the JDBC driver must be compatible with Java 8

turboFei: Seems impossible with arrow 17.0.0 ...

pan3793: Many widely used projects have abandoned Java 8, we must follow that to escalate our supported JDK baseline if the CVE is a real matter

turboFei:
> I think we are not affected.


Agree, we are not impacted.

PR #7171, opened 2025-08-13
Fix empty list engine result when etcd is used as the service registry

zhaohehuhu:

In etcd, keys are stored in a flat namespace where / is just part of the key name. For example, /kyuubi_version_USER_SPARK_SQL/test/default is treated as a complete single key. 

To retrieve all entries under a similar path, a prefix query (--prefix) is required.

In ZooKeeper, data is organized in a hierarchical tree structure where / represents parent-child relationships.

added a UT to verify code.

No

### DIFF
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/DiscoveryClient.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/DiscoveryClient.scala
@@ -66,6 +66,7 @@ trait DiscoveryClient extends Logging {
    */
   def pathNonExists(path: String): Boolean
 
+  def pathNonExists(path: String, isPrefix: Boolean): Boolean
   /**
    * Delete a path.
    * @param path the path to be deleted
--- a/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/DiscoveryClient.scala
+++ b/kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/DiscoveryClient.scala
@@ -67,6 +67,7 @@ trait DiscoveryClient extends Logging {
   def pathNonExists(path: String): Boolean
 
   def pathNonExists(path: String, isPrefix: Boolean): Boolean
+
   /**
    * Delete a path.
    * @param path the path to be deleted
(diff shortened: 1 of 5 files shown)


kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/etcd/EtcdDiscoveryClient.scala
```diff
@@ -151,6 +151,12 @@ class EtcdDiscoveryClient(conf: KyuubiConf) extends DiscoveryClient {
     kvClient.get(ByteSequence.from(path.getBytes())).get().getKvs.isEmpty
   }
 
+  override def pathNonExists(path: String, isPrefix: Boolean): Boolean = {
+    kvClient.get(
+      ByteSequence.from(path.getBytes()),
+      GetOption.newBuilder().isPrefix(isPrefix).build()).get().getKvs.isEmpty
```
pan3793, line 153: what will happen if we always set isPrefix to true?

zhaohehuhu: If isPrefix = true, it gets all keys starting with path. But I believe it won't break current behavior in Kyuubi. @pan3793

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/DiscoveryClient.scala
```diff
@@ -66,6 +66,8 @@ trait DiscoveryClient extends Logging {
    */
   def pathNonExists(path: String): Boolean
 
+  def pathNonExists(path: String, isPrefix: Boolean): Boolean
```
pan3793: let's delete the method `def pathNonExists(path: String): Boolean` and give `isPrefix` a default value

pan3793: also, add javadocs to this method

zhaohehuhu: Thanks. I will improve it.

kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/DiscoveryClient.scala
```diff
@@ -62,9 +62,14 @@ trait DiscoveryClient extends Logging {
   def pathExists(path: String): Boolean
 
   /**
-   * Check if the path non exists.
-   */
-  def pathNonExists(path: String): Boolean
+   * Checks whether the given path and all its child paths (prefix matches) do not exist.
+   * The isPrefix is set to true by default for Etcd to retrieve all entries under the given path.
+   * For other discovery service, it can be set false by default to only check the exact path.
+   * @param path The path to check
```
pan3793, line 68:
```suggestion
   *
   * @param path The path to check
```

pan3793 (approved):

pan3793:
Thanks, merged to master/1.10

PR #7174, opened 2025-08-15
Kyuubi Spark Extention supports Spark 4.0

pan3793:

Port Kyuubi Spark Extension from 3.5 to 4.0.

Note: the `KyuubiEnsureRequirements` does not handle DSv2 bucketing properly now.

UTs are ported too.

No.

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-4-0/src/main/scala/org/apache/kyuubi/sql/KyuubiEnsureRequirements.scala
+++ b/extensions/spark/kyuubi-extension-spark-4-0/src/main/scala/org/apache/kyuubi/sql/KyuubiEnsureRequirements.scala
@@ -1,0 +1,328 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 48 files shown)


extensions/spark/kyuubi-extension-spark-4-0/src/main/scala/org/apache/kyuubi/sql/KyuubiEnsureRequirements.scala
```diff
@@ -0,0 +1,328 @@
...
+import org.apache.spark.sql.execution.exchange._
+import org.apache.spark.sql.execution.joins.{ShuffledHashJoinExec, SortMergeJoinExec}
+import org.apache.spark.sql.internal.SQLConf
+
+/**
+ * Copy from Apache Spark `EnsureRequirements`
+ *   1. remove reorder join predicates
+ *   2. remove shuffle pruning
+ *
+ * TODO: bring back the v2 bucketing join handling logic
```
pan3793, line 37: Unfortunately, we don't have UT for this case

wForget (approved):
Thanks, LGTM

pan3793:
Merged to master

PR #7176, opened 2025-08-20
Cleanup metadata with batch size and interval

turboFei:

Cleanup metadata with batch size and interval to prevent hold the lock for long time if the metadata size is huge.

GA.

No.


kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JdbcDatabaseDialect.scala
```diff
@@ -31,6 +32,10 @@ class GenericDatabaseDialect extends JdbcDatabaseDialect {
     s"LIMIT $limit OFFSET $offset"
   }
 
+  override def deleteFromLimitClause(limit: Int): String = {
+    "" // Generic dialect does not support LIMIT in DELETE statements
```
pan3793: let's add a warning log here. it won't bother the user since generally, it won't fall back to `GenericDatabaseDialect`

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2077,6 +2077,24 @@ object KyuubiConf {
...
 
+  val METADATA_CLEANER_BATCH_SIZE: ConfigEntry[Int] =
+    buildConf("kyuubi.metadata.cleaner.batch.size")
+      .serverOnly
+      .doc("The batch size for cleaning expired metadata. " +
+        "This is used to avoid holding the delete lock for a long time " +
+        "when there are too many expired metadata to be cleaned.")
+      .version("1.11.0")
+      .intConf
+      .createWithDefault(Int.MaxValue)
```
pan3793: I expect this is a general perf optimization, should we give a more reasonable default value? e.g. 1000 or 3000

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2077,6 +2077,24 @@ object KyuubiConf {
...
+
+  val METADATA_CLEANER_BATCH_INTERVAL: ConfigEntry[Long] =
+    buildConf("kyuubi.metadata.cleaner.batch.interval")
+      .serverOnly
+      .internal
+      .doc("The interval to wait before next batch of cleaning expired metadata.")
+      .timeConf
+      .createWithDefault(Duration.ofMillis(100).toMillis)
```
pan3793: this is a lower-priority work, let's have at least 1s interval to avoid impacting batch scheduling

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataStore.scala
```diff
@@ -102,8 +102,13 @@ trait MetadataStore extends Closeable {
   /**
    * Check and cleanup the terminated batches information with maxAge limitation.
    * @param maxAge the batch state info maximum age.
+   * @param limit the maximum number of metadata to be cleaned up.
    */
-  def cleanupMetadataByAge(maxAge: Long): Unit
+  def cleanupMetadataByAge(maxAge: Long, limit: Int): Int
+
+  def cleanupMetadataByAge(maxAge: Long): Int = {
+    cleanupMetadataByAge(maxAge, Int.MaxValue)
```
pan3793: it worth nothing to keep this API, since the return value changed

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
```diff
@@ -253,6 +256,31 @@ class MetadataManager extends AbstractService("MetadataManager") {
...
+            batchSize)) >= batchSize
+      }
+      if (needToCleanKubernetesInfo) {
+        needToCleanKubernetesInfo =
+          withMetadataRequestMetrics(_metadataStore.cleanupKubernetesEngineInfoByAge(
+            maxAge,
+            batchSize)) >= batchSize
+      }
+      if (needToCleanMetadata || needToCleanKubernetesInfo) {
+        info("Sleep for " + batchInterval + "ms before next metadata cleanup batch")
```
pan3793: why not use an interpreted string?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
```diff
@@ -253,6 +256,31 @@ class MetadataManager extends AbstractService("MetadataManager") {
       TimeUnit.MILLISECONDS)
   }
 
+  @VisibleForTesting
+  private[metadata] def cleanupMetadata(maxAge: Long, batchSize: Int, batchInterval: Long): Unit = {
+    var needToCleanMetadata = true
+    var needToCleanKubernetesInfo = true
+
+    while (needToCleanMetadata || needToCleanKubernetesInfo) {
```
pan3793: let's have an enforced loop round threshold(maybe `100`, no need to be configurable， hardcode is fine) to guard this while loop.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
```diff
@@ -253,6 +256,32 @@ class MetadataManager extends AbstractService("MetadataManager") {
...
+        needToCleanKubernetesInfo =
+          withMetadataRequestMetrics(_metadataStore.cleanupKubernetesEngineInfoByAge(
+            maxAge,
+            batchSize)) >= batchSize
+      }
+      if (needToCleanMetadata || needToCleanKubernetesInfo) {
+        info(s"Sleep for $batchInterval ms before next metadata cleanup batch")
+        Thread.sleep(batchInterval)
+      }
+      cleanupLoop += 1
```
pan3793: log a warning message if it exceeds the threshold, so we know there might be some bugs or unexpected data in metastore

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
```diff
@@ -346,6 +375,8 @@ class MetadataManager extends AbstractService("MetadataManager") {
 }
 
 object MetadataManager extends Logging {
+  final val MAX_CLEANUP_LOOPS = 100
```
pan3793: it won't be used in other places, move it to the used function

turboFei:
addressed the comments

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
```diff
@@ -253,6 +256,39 @@ class MetadataManager extends AbstractService("MetadataManager") {
       TimeUnit.MILLISECONDS)
   }
 
+  @VisibleForTesting
+  private[metadata] def cleanupMetadata(maxAge: Long, batchSize: Int, batchInterval: Long): Unit = {
+    var needToCleanMetadata = true
+    var needToCleanKubernetesInfo = true
+    var cleanupLoop = 0
+    val MAX_CLEANUP_LOOPS = 100
+
```
pan3793:
```suggestion

    val MAX_CLEANUP_LOOPS = 100 // a guard in case it runs into an infinite loop
```

pan3793 (approved):

turboFei:
thanks, merged into master(1.11.0)

Issue #7180, opened 2025-09-01
[Bug] [LINEAGE] Subquery in the project doesn't drill down to get the column lineage


yabola:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

For example, the following SQL statement will get the wrong column lineage result:
create table table0(a int, b string, c string)
create table table1(a int, b string, c string)
select (select sum(a) from table0 where table1.b = table0.b) as aa, b from table1

`aa` should come from table0.a , not from table0.b

### Affects Version(s)

1.7.0

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #7181, opened 2025-09-01
[KYUUBI #7180][LINEAGE] Subquery in the project should drill down to get the column lineage relationships

yabola:

The following SQL statement will get the wrong column lineage result:
create table table0(a int, b string, c string)
create table table1(a int, b string, c string)
select (select sum(a) from table0 where table1.b = table0.b) as aa, b from table1

The root cause:  
From  https://github.com/apache/spark/pull/32687 ,  we can know the references for a subquery expression are defined as outer attribute references.  So we should always drill down to get the corresponding column lineage relationship for the subquery plan.

add new ut

no

### DIFF
--- a/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
+++ b/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
@@ -119,14 +119,15 @@ trait LineageParser {
     val exps = named.map {
       case exp: Alias =>
         val references =
-          if (exp.references.nonEmpty) exp.references
-          else {
+          if (exp.references.isEmpty || exp.child.isInstanceOf[ScalarSubquery]) {
             val attrRefs = getExpressionSubqueryPlans(exp.child)
               .map(extractColumnsLineage(_, ListMap[Attribute, AttributeSet]()))
               .foldLeft(ListMap[Attribute, AttributeSet]())(mergeColumnsLineage).values
               .foldLeft(AttributeSet.empty)(_ ++ _)
               .map(attr => attr.withQualifier(attr.qualifier :+ SUBQUERY_COLUMN_IDENTIFIER))
             AttributeSet(attrRefs)
+          } else {
+            exp.references
           }
         (
           exp.toAttribute,
--- a/extensions/spark/kyuubi-spark-lineage/src/test/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParserHelperSuite.scala
+++ b/extensions/spark/kyuubi-spark-lineage/src/test/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParserHelperSuite.scala
@@ -1143,6 +1143,18 @@ abstract class SparkSQLLineageParserHelperSuite extends KyuubiFunSuite
         List(
           ("a", Set(s"$DEFAULT_CATALOG.default.table1.a")),
           ("b", Set(s"$DEFAULT_CATALOG.default.table1.b")))))
+
+      val sql12 =
+        """
+          |select (select sum(a) from table0 where table1.b = table0.b) as aa, b from table1
+          |""".stripMargin
+      val ret12 = extractLineage(sql12)
+      assert(ret12 == Lineage(
+        List(s"$DEFAULT_CATALOG.default.table0", s"$DEFAULT_CATALOG.default.table1"),
+        List(),
+        List(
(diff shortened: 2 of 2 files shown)


yabola:
@wForget @pan3793  please help review when you have time, thank you~

yabola:
@pan3793 Hi~ could you help review this when you have time? We had tested much other sql and didn't find more problems.

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
```diff
@@ -124,19 +115,19 @@ trait LineageParser {
...
   private def getSelectColumnLineage(
-      named: Seq[NamedExpression],
-      inputTablesByPlan: mutable.HashSet[String]): AttributeMap[AttributeSet] = {
+      named: Seq[NamedExpression]): AttributeMap[AttributeSet] = {
     val exps = named.map {
       case exp: Alias =>
         val references =
-          if (exp.references.nonEmpty) exp.references
-          else {
+          if (exp.references.isEmpty || exp.child.isInstanceOf[ScalarSubquery]) {
```
wForget: Is it enough to simply check if exp.child is a ScalarSubquery? Do we need to recursively check the children of exp?

For example, could you verify this sql?

```
select if((select sum(a) from table0 where table1.b = table0.b) > 100, 1, 2) as aa, b from table1
```

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
```diff
@@ -124,19 +115,19 @@ trait LineageParser {
...
             val attrRefs = getExpressionSubqueryPlans(exp.child)
-              .map(extractColumnsLineage(_, ListMap[Attribute, AttributeSet](), inputTablesByPlan))
+              .map(extractColumnsLineage(_, ListMap[Attribute, AttributeSet]()))
               .foldLeft(ListMap[Attribute, AttributeSet]())(mergeColumnsLineage).values
               .foldLeft(AttributeSet.empty)(_ ++ _)
               .map(attr => attr.withQualifier(attr.qualifier :+ SUBQUERY_COLUMN_IDENTIFIER))
             AttributeSet(attrRefs)
```
wForget: Additionally, non-subquery references of exp may be ignored here.

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
yabola: Thank you for your review. Yes, this problem exists. I add  recursively check for the children exp.

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
yabola: This is a good catch. I add `exp.references` to the returned variables  and add more UT for this case.

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
```diff
@@ -128,15 +128,17 @@ trait LineageParser {
       inputTablesByPlan: mutable.HashSet[String]): AttributeMap[AttributeSet] = {
     val exps = named.map {
       case exp: Alias =>
+        val subqueryPlans = getExpressionSubqueryPlans(exp.child)
         val references =
-          if (exp.references.nonEmpty) exp.references
-          else {
-            val attrRefs = getExpressionSubqueryPlans(exp.child)
+          if (exp.references.isEmpty || subqueryPlans.nonEmpty) {
```
wForget: `exp.references.isEmpty` is not needed.

yabola: Yes, this is redundant code, I deleted it.

wForget (approved):
Thanks @yabola , LGTM

wForget:
Thanks @yabola , merged to master

Issue #7183, opened 2025-09-02
[FEATURE] [LINEAGE] Support collect all the input tables by the plan


yabola:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

The current input tables is derived from column which has lineage result, which will cause some input tables to be missing. We need to support one way to collect the complete input tables by the plan.
For example:
insert overwrite v2_catalog.db.tb3 select t1.col1, t1.col2 , t1.col3 from v2_catalog.db.tb1 t1 join v2_catalog.db.tb2 t2 on t1.col1 = t2.col1

The input tables for v2_catalog.db.tb3  should be v2_catalog.db.tb1  and v2_catalog.db.tb2

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #7184, opened 2025-09-02
[KYUUBI #7183][LINEAGE] Support collect all input tables by the plan

yabola:

The current input tables is derived from column which has lineage result, which will cause some input tables to be missing. We need to support one way to collect the complete input tables by the plan.
For example:
insert overwrite v2_catalog.db.tb3 select t1.col1, t1.col2 , t1.col3 from v2_catalog.db.tb1 t1 join v2_catalog.db.tb2 t2 on t1.col1 = t2.col1

The input tables for v2_catalog.db.tb3 should be v2_catalog.db.tb1 and v2_catalog.db.tb2

add new UT

no

### DIFF
--- a/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
+++ b/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
@@ -72,8 +72,7 @@ trait LineageParser {
     if (SparkContextHelper.getConf(LineageConf.LEGACY_COLLECT_INPUT_TABLES_ENABLED)) {
       Lineage(inputTablesByColumn.distinct, outputTables.distinct, columnsLineage)
     } else {
-      val filteredInputTables = inputTablesByPlan -- outputTables.distinct
-      Lineage(filteredInputTables.toList, outputTables.distinct, columnsLineage)
+      Lineage(inputTablesByPlan.toList, outputTables.distinct, columnsLineage)
     }
   }
 
--- a/extensions/spark/kyuubi-spark-lineage/src/test/scala/org/apache/kyuubi/plugin/lineage/helper/RowLevelCatalogLineageParserSuite.scala
+++ b/extensions/spark/kyuubi-spark-lineage/src/test/scala/org/apache/kyuubi/plugin/lineage/helper/RowLevelCatalogLineageParserSuite.scala
@@ -52,7 +52,7 @@ class RowLevelCatalogLineageParserSuite extends SparkSQLLineageParserHelperSuite
           "  INSERT (pk, name, price) VALUES (cast(source.pk as int), source.name, source.price)" +
           "WHEN NOT MATCHED BY SOURCE THEN  UPDATE SET target.name = 'abc' ")
       assert(ret0 == Lineage(
-        List("v2_catalog.db.source_t"),
+        List("v2_catalog.db.source_t", "v2_catalog.db.target_t"),
         List("v2_catalog.db.target_t"),
         List(
           (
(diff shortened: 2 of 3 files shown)


yabola:
@wForget @pan3793 cc, please help review it when you have time, thank you

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/spark/kyuubi/lineage/LineageConf.scala
```diff
@@ -46,6 +46,14 @@ object LineageConf {
       "Unsupported lineage dispatchers")
     .createWithDefault(Seq(LineageDispatcherType.SPARK_EVENT.toString))
 
+  val COLLECT_INPUT_TABLES_BY_PLAN_ENABLED =
+    ConfigBuilder("spark.kyuubi.plugin.lineage.collectInputTablesByPlan.enabled")
+      .doc("Whether to collect input tables by the plan." +
+        " By Default collect input tables by the column lineages.")
+      .version("1.11.0")
+      .booleanConf
+      .createWithDefault(false)
```
pan3793, line 56: this is more like a bug fix, let's enable it by default and mark it `internal`

yabola: Thank you for review, I modified the default value and config name, see if it is appropriate

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
```diff
@@ -53,18 +54,27 @@ trait LineageParser {
...
       }
-    val (inputTables, outputTables) = columnsLineage.foldLeft((List[String](), List[String]())) {
-      case ((inputs, outputs), (out, in)) =>
-        val x = (inputs ++ in.map(_.split('.').init.mkString("."))).filter(_.nonEmpty)
-        val y = outputs ++ List(out.split('.').init.mkString(".")).filter(_.nonEmpty)
-        (x, y)
+    if (SparkContextHelper.getConf(LineageConf.LEGACY_COLLECT_INPUT_TABLES_ENABLED)) {
+      Lineage(inputTablesByColumn.distinct, outputTables.distinct, columnsLineage)
+    } else {
+      val filteredInputTables = inputTablesByPlan -- outputTables.distinct
```
yabola: I also filtered out the outputTables here. Some plans, such as merge into, will insert their own data into themselves, but databricks does not count this part of the input table. I am not sure if this is appropriate.

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/spark/kyuubi/lineage/LineageConf.scala
```diff
@@ -46,6 +46,14 @@ object LineageConf {
       "Unsupported lineage dispatchers")
     .createWithDefault(Seq(LineageDispatcherType.SPARK_EVENT.toString))
 
+  val LEGACY_COLLECT_INPUT_TABLES_ENABLED =
+    ConfigBuilder("spark.kyuubi.plugin.lineage.legacy.collectInputTablesByColumn")
+      .doc(" When true collect input tables by column lineage." +
+        " When false collect all the input tables by the plan.")
```
pan3793:
```suggestion
      .internal
      .doc("When true, collect input tables by column lineage. " +
        "When false, collect all the input tables by the plan.")
```

pan3793 (dismissed):

extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
pan3793: what about `insert into a select * from a`?

yabola: Sorry, my fault. I reconfirmed that databricks will include  `a`  as upstream table , but merge into will not (although the source table is also scanned ). I'm confused about this behavior.

pan3793 (approved):

PR #7190, opened 2025-09-04
Fix - Presto SQLAlchemy dialect did not implement get_view_names

rad-pat:
Presto SQLAlchemy dialect did not implement the `get_view_names` method and resulted in an exception when trying to inspect the schema. This was discovered in Superset repo whilst trying to update the pandas package which now makes a call to `get_view_names`. 

Very basic Python tests have been added here, but all the SQLAlchemy dialects in this repo would benefit from running the full SQLAlchemy dialect test suite instead of these bespoke tests.

### DIFF
--- a/python/pyhive/sqlalchemy_presto.py
+++ b/python/pyhive/sqlalchemy_presto.py
@@ -231,9 +231,9 @@ class PrestoDialect(default.DefaultDialect):
             )
         else:
             # N.B. This is incorrect, if no schema is provided, the current/default schema should
-            #   be used with a call to self._get_default_schema_name(connection), but I could not
-            #   see how to implement that
-            #   default_schema = self._get_default_schema_name(connection)
+            #  be used with a call to self._get_default_schema_name(connection), but I could not
+            #  see how to implement that
+            #  default_schema = self._get_default_schema_name(connection)
             view_name_query = """
                 SELECT table_name
                 FROM information_schema.views


pan3793:
It doesn't seem lucky, the change causes test failure. Could you take a look?

rad-pat:
@pan3793, I have hopefully fixed it - I also extended the python range of tests to cover supported versions of Python, and bumped the package version in the hopes that a release could be made to PyPI when merged.

rad-pat:
Looking at failures, new python versions failing because of old packages. The previous versions of python still failing for same reason, I thought I had fixed it, checking...

rad-pat:
@pan3793 - definitely no broken Python tests now, I removed the addition of current Python versions as they fail because dependencies are not up to date with Python > 3.10. So, just the changes required and tests now working. 🤞 More work than I was hoping for.

python/pyhive/__init__.py
```diff
@@ -1,3 +1,3 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
-__version__ = '0.7.0'
+__version__ = '0.7.1'
```
pan3793:
```suggestion
__version__ = '0.7.0'
```

python/pyhive/sqlalchemy_presto.py
```diff
@@ -204,12 +204,44 @@ def get_indexes(self, connection, table_name, schema=None, **kw):
...
 
+    def _get_default_schema_name(self, connection):
+        #'SELECT CURRENT_SCHEMA()'
+        return super()._get_default_schema_name(connection)
+
     def get_table_names(self, connection, schema=None, **kw):
         query = 'SHOW TABLES'
+        # N.B. This is incorrect, if no schema is provided, the current/default schema should be used
+        #  with a call to an overridden self._get_default_schema_name(connection), but I could not see how to implement
+        #  that as there is no CURRENT_SCHEMA function
```
pan3793:
```suggestion
        #  with a call to an overridden self._get_default_schema_name(connection), but I could not
        # see how to implement that as there is no CURRENT_SCHEMA function
```

python/pyhive/sqlalchemy_presto.py
```diff
@@ -204,12 +204,44 @@ def get_indexes(self, connection, table_name, schema=None, **kw):
...
 
+    def _get_default_schema_name(self, connection):
+        #'SELECT CURRENT_SCHEMA()'
+        return super()._get_default_schema_name(connection)
+
     def get_table_names(self, connection, schema=None, **kw):
         query = 'SHOW TABLES'
+        # N.B. This is incorrect, if no schema is provided, the current/default schema should be used
+        #  with a call to an overridden self._get_default_schema_name(connection), but I could not
+        # see how to implement that as there is no CURRENT_SCHEMA function
```
pan3793:
```suggestion
        #  with a call to an overridden self._get_default_schema_name(connection), but I could not
        #  see how to implement that as there is no CURRENT_SCHEMA function
```

python/pyhive/sqlalchemy_presto.py
```diff
@@ -204,12 +204,44 @@ def get_indexes(self, connection, table_name, schema=None, **kw):
...
+                FROM information_schema.views
+                WHERE table_schema = :schema
+            """
+            query = text(view_name_query).bindparams(
+                bindparam("schema", type_=types.Unicode)
+            )
+        else:
+            # N.B. This is incorrect, if no schema is provided, the current/default schema should be used
+            #  with a call to self._get_default_schema_name(connection), but I could not see how to implement that
+            #  default_schema = self._get_default_schema_name(connection)
```
pan3793:
```suggestion
            # N.B. This is incorrect, if no schema is provided, the current/default schema should
            #   be used with a call to self._get_default_schema_name(connection), but I could not
            #   see how to implement that
            #   default_schema = self._get_default_schema_name(connection)
```

python/pyhive/sqlalchemy_presto.py
```diff
@@ -204,12 +204,45 @@ def get_indexes(self, connection, table_name, schema=None, **kw):
...
+                WHERE table_schema = :schema
+            """
+            query = text(view_name_query).bindparams(
+                bindparam("schema", type_=types.Unicode)
+            )
+        else:
+            # N.B. This is incorrect, if no schema is provided, the current/default schema should
+            #   be used with a call to self._get_default_schema_name(connection), but I could not
+            #   see how to implement that
+            #   default_schema = self._get_default_schema_name(connection)
```
pan3793:
```suggestion
            # N.B. This is incorrect, if no schema is provided, the current/default schema should
            #  be used with a call to self._get_default_schema_name(connection), but I could not
            #  see how to implement that
            #  default_schema = self._get_default_schema_name(connection)
```

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #7202, opened 2025-09-12
[AUTHZ] make FilteredShowObjectsExec extends V2CommandExec to avoid running executor side task

LennonChin:
Now FilteredShowObjectsExec is not a subclass of V2CommandExec, when it was executed it will invoke `def doExecute(): RDD[InternalRow]` method to submit executor side task to run, which is slow and resource wasting.  V2CommandExec implemented `def executeCollect(): Array[InternalRow]` to avoid running `def doExecute(): RDD[InternalRow]`.

Improve perfomance.

exists unit tests.

No.


bowenliang123 (approved):
LGTM.

PR #7204, opened 2025-09-12
[AUTHZ] count(*)/count(1) should check sub nodes' privileges

LennonChin:
Spark Optimizer's ColumnPruning will replace `count(*)/count(1)` `Aggregate` plan's child to a `Project` node with empty projection list:

```scala
object ColumnPruning extends Rule[LogicalPlan] {
  def apply(plan: LogicalPlan): LogicalPlan = removeProjectBeforeFilter(
    plan.transformWithPruning(AlwaysProcess.fn, ruleId) {
    ...
    // Prunes the unused columns from child of Aggregate/Expand/Generate/ScriptTransformation
    case a @ Aggregate(_, _, child) if !child.outputSet.subsetOf(a.references) =>
      a.copy(child = prunedChild(child, a.references))
    ...
}
```

but AuthZ plugin's `PrivilegesBuilder.buildQuery` method will ignore to check child node when plan's `inputSet` is empty, in this scenario, `Aggregate` node's child plan's privileges are ignored, which cause `count(*)/ count(1)` will ignored all privileges that should be checked.

this patch add a holder node `ChildOutputHolder` when `Aggregate` node's `references` and it's child node's `outputSet` hava no intersection, it will hold the child node's `outputSet` used for build privilege objects, and `ChildOutputHolder` node will be eliminated after `RuleAuthorization` rule work completed.

updated old unit tests and added new unit tests.

No.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
@@ -107,8 +107,11 @@ object PrivilegesBuilder {
           //   1. `MapInPandas`, `ScriptTransformation`
           //   2. `Project` output only have constant value
           if (columnPrune(p.references.toSeq ++ p.output, p.inputSet).isEmpty) {
-            // If plan is project and output don't have relation to input, can ignore.
-            // If plan tree exists ChildOutputHolder, we should build child logic plan.
+            // 1. If plan is project and output don't have relation to input, can ignore.
+            // 2. If sub logic plan tree exists ChildOutputHolder node, it means that the output of
+            //    some nodes in the tree is fixed by RuleChildOutputMarker in some special
+            //    scenarios, such as the Aggregate(count(*)) child node. To avoid missing child node
+            //    permissions, we need to continue checking down.
             if (!p.isInstanceOf[Project] || existsChildOutputHolder) {
               buildQuery(
                 child,
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/ChildOutputHolder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/ChildOutputHolder.scala
@@ -22,10 +22,10 @@ import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, UnaryNode}
 
 import org.apache.kyuubi.plugin.spark.authz.util.WithInternalChild
 
-case class ChildOutputHolder(child: LogicalPlan, fixedOutput: Seq[Attribute])
+case class ChildOutputHolder(child: LogicalPlan, childOutput: Seq[Attribute])
   extends UnaryNode with WithInternalChild {
 
-  val output: Seq[Attribute] = fixedOutput
+  val output: Seq[Attribute] = childOutput
 
   override def withNewChildInternal(newChild: LogicalPlan): LogicalPlan = copy(child = newChild)
 }
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateChildOutputHolder.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/RuleEliminateChildOutputHolder.scala
@@ -22,6 +22,9 @@ import org.apache.spark.sql.catalyst.rules.Rule
 
 import org.apache.kyuubi.plugin.spark.authz.rule.plan.ChildOutputHolder
 
+/**
+ * Transforming down [[ChildOutputHolder]]
+ */
 object RuleEliminateChildOutputHolder extends Rule[LogicalPlan] {
   override def apply(plan: LogicalPlan): LogicalPlan = {
     plan.transform {
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/RuleChildOutputMarker.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/RuleChildOutputMarker.scala
@@ -20,6 +20,13 @@ package org.apache.kyuubi.plugin.spark.authz.rule.plan
 import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, LogicalPlan}
 import org.apache.spark.sql.catalyst.rules.Rule
 
+/**
+ * Rule to add [[ChildOutputHolder]] used to fixed logic plan's child node's output,
+ * now used for following cases:
+ *
+ *  1. Aggregate(count(*)/count(1)), it's child node will be pruned in Spark optimizer
+ *     rule [[org.apache.spark.sql.catalyst.optimizer.ColumnPruning]].
+ */
 object RuleChildOutputMarker extends Rule[LogicalPlan] {
   override def apply(plan: LogicalPlan): LogicalPlan = {
     plan.transform {


extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/ChildOutputHolder.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.rule.plan
+
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, UnaryNode}
+
+import org.apache.kyuubi.plugin.spark.authz.util.WithInternalChild
+
+case class ChildOutputHolder(child: LogicalPlan, fixedOutput: Seq[Attribute])
```
pan3793: let's add comments to explain why we need this.

`fixedOutput` is not a good name, it `fixed` what? maybe just call it `childOutput`

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
```diff
@@ -100,13 +101,15 @@ object PrivilegesBuilder {
...
       case p =>
+        val existsChildOutputHolder = p.exists(_.isInstanceOf[ChildOutputHolder])
         for (child <- p.children) {
           // If current plan's references don't have relation to it's input, have two cases
           //   1. `MapInPandas`, `ScriptTransformation`
           //   2. `Project` output only have constant value
           if (columnPrune(p.references.toSeq ++ p.output, p.inputSet).isEmpty) {
             // If plan is project and output don't have relation to input, can ignore.
-            if (!p.isInstanceOf[Project]) {
+            // If plan tree exists ChildOutputHolder, we should build child logic plan.
```
pan3793: the variable name `existsChildOutputHolder` itself explains what it does, a good comment explains WHY, not WHAT.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/ChildOutputHolder.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, UnaryNode}
+
+import org.apache.kyuubi.plugin.spark.authz.util.WithInternalChild
+
+case class ChildOutputHolder(child: LogicalPlan, childOutput: Seq[Attribute])
+  extends UnaryNode with WithInternalChild {
+
+  val output: Seq[Attribute] = childOutput
```
wForget, line 28: Why not just return `child.output`?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
```diff
@@ -100,13 +101,18 @@ object PrivilegesBuilder {
...
           //   2. `Project` output only have constant value
           if (columnPrune(p.references.toSeq ++ p.output, p.inputSet).isEmpty) {
-            // If plan is project and output don't have relation to input, can ignore.
-            if (!p.isInstanceOf[Project]) {
+            // 1. If plan is project and output don't have relation to input, can ignore.
+            // 2. If sub logic plan tree exists ChildOutputHolder node, it means that the output of
+            //    some nodes in the tree is fixed by RuleChildOutputMarker in some special
+            //    scenarios, such as the Aggregate(count(*)) child node. To avoid missing child node
+            //    permissions, we need to continue checking down.
+            if (!p.isInstanceOf[Project] || existsChildOutputHolder) {
```
wForget, line 115: Does this affect children that are not `ChildOutputHolder` ?

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/ChildOutputHolder.scala
LennonChin: > Why not just return `child.output`?

Suppose now we have a table like:

```sql
CREATE TABLE IF NOT EXISTS default.test
(id int, scope int, part string)
PARTITIONED BY(part)
```

when we query like `select count(*) from (select id from default.test where part = 'part-1') t`, the original plan is following:

```
GlobalLimit 21
+- LocalLimit 21
   +- Project [cast(count(1)#10L as string) AS count(1)#13]
      +- Aggregate [count(1) AS count(1)#10L]
         +- Project [id#7]
            +- Filter (part#9 = part-1)
               +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]] 
```

because the `references` of `Aggregate [count(1) AS count(1)#21L]` is empty, `ColumnPruning` will add a `Project` with empty projection list onto `Project [id#49]` and `CollapseProject` will collapsing this empty `Project` node:

```
=== Applying Rule org.apache.spark.sql.catalyst.optimizer.ColumnPruning ===
 GlobalLimit 21                                                                                                                                                     GlobalLimit 21
 +- LocalLimit 21                                                                                                                                                   +- LocalLimit 21
    +- Project [cast(count(1)#10L as string) AS count(1)#13]                                                                                                           +- Project [cast(count(1)#10L as string) AS count(1)#13]
       +- Aggregate [count(1) AS count(1)#10L]                                                                                                                            +- Aggregate [count(1) AS count(1)#10L]
!         +- Project [id#7]                                                                                                                                                  +- Project
!            +- Filter (part#9 = part-1)                                                                                                                                        +- Project
!               +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]                  +- Filter (part#9 = part-1)
!                                                                                                                                                                                     +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]

=== Applying Rule org.apache.spark.sql.catalyst.optimizer.CollapseProject ===
 GlobalLimit 21                                                                                                                                                        GlobalLimit 21
 +- LocalLimit 21                                                                                                                                                      +- LocalLimit 21
!   +- Project [cast(count(1)#10L as string) AS count(1)#13]                                                                                                              +- Aggregate [cast(count(1) as string) AS count(1)#13]
!      +- Aggregate [count(1) AS count(1)#10L]                                                                                                                               +- Project
!         +- Project                                                                                                                                                            +- Filter (part#9 = part-1)
!            +- Project                                                                                                                                                            +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]
!               +- Filter (part#9 = part-1)                                                                                                                            
!                  +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]   
```

the plan tree will be modified as following result:

```
GlobalLimit 21
+- LocalLimit 21
   +- Aggregate [cast(count(1) as string) AS count(1)#13]
      +- Project
         +- Filter (part#9 = part-1)
            +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]
```

in `PrivilegeBuilder.buildQuery`, follwing code will skip `Aggregate [count(1) AS count(1)#21L]` because it's `p.references.toSeq ++ p.output` is empty, so all child nodes' privileges in the tree will be ignored:

```scala
if (columnPrune(p.references.toSeq ++ p.output, p.inputSet).isEmpty) {
  // If plan is project and output don't have relation to input, can ignore.
  if (!p.isInstanceOf[Project]) {
   ...
  }
}
```

if we add a `ChildOutputHolder`, the plan tree checked by `PrivilegeBuilder.buildQuery` will be:

```scala
Aggregate [cast(count(1) as string) AS count(1)#13]
+- Project
   +- ChildOutputHolder [id#7]
      +- Project [id#7]
         +- Filter (part#9 = part-1)
            +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]
```

`ChildOutputHolder [id#7]` fixed the output of sub node `Project [id#7]` and because plan tree exists `ChildOutputHolder` node, follwing code will continue check sub nodes' privileges:

```scala
case p =>
   val existsChildOutputHolder = p.exists(_.isInstanceOf[ChildOutputHolder])
   for (child <- p.children) {
      if (columnPrune(p.references.toSeq ++ p.output, p.inputSet).isEmpty) {
      // 1. If plan is project and output don't have relation to input, can ignore.
      // 2. If sub logic plan tree exists ChildOutputHolder node, it means that the output of
      //    some nodes in the tree is fixed by RuleChildOutputMarker in some special
      //    scenarios, such as the Aggregate(count(*)) child node. To avoid missing child node
      //    permissions, we need to continue checking down.
      if (!p.isInstanceOf[Project] || existsChildOutputHolder) {
         buildQuery(
            child,
            privilegeObjects,
            p.inputSet.map(_.toAttribute).toSeq,
            Nil,
            spark)
      }
      } else {
      ...
      }
   }
```

Why does `ChildOutputHolder` need to hold the output of the `Aggregate` node's original child node? if we just return `child.output` in `ChildOutputHolder` like this:

```scala
case class ChildOutputHolder(child: LogicalPlan)
  extends UnaryNode with WithInternalChild {

  val output: Seq[Attribute] = child.output

  override def withNewChildInternal(newChild: LogicalPlan): LogicalPlan = copy(child = newChild)
}
```

after `ColumnPruning` and `CollapseProject` conversion, `ChildOutputHolder`'s child will be a `Project` with empty projection list, `ChildOutputHolder`'s input and output are empty, we lost original output of `Aggregate`'s child node `Project [id#7]`, therefore we will lost the privilege checking of `id` column:

```
=== Applying Rule org.apache.spark.sql.catalyst.optimizer.ColumnPruning ===
 GlobalLimit 21                                                                                                                                                        GlobalLimit 21
 +- LocalLimit 21                                                                                                                                                      +- LocalLimit 21
    +- Project [cast(count(1)#10L as string) AS count(1)#13]                                                                                                              +- Project [cast(count(1)#10L as string) AS count(1)#13]
       +- Aggregate [count(1) AS count(1)#10L]                                                                                                                               +- Aggregate [count(1) AS count(1)#10L]
!         +- ChildOutputHolder                                                                                                                                                  +- Project
!            +- Project [id#7]                                                                                                                                                     +- ChildOutputHolder
!               +- Filter (part#9 = part-1)                                                                                                                                           +- Project
!                  +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]                     +- Project
!                                                                                                                                                                                           +- Filter (part#9 = part-1)
!                                                                                                                                                                                              +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]

=== Applying Rule org.apache.spark.sql.catalyst.optimizer.CollapseProject ===
 GlobalLimit 21                                                                                                                                                              GlobalLimit 21
 +- LocalLimit 21                                                                                                                                                            +- LocalLimit 21
!   +- Project [cast(count(1)#10L as string) AS count(1)#13]                                                                                                                    +- Aggregate [cast(count(1) as string) AS count(1)#13]
!      +- Aggregate [count(1) AS count(1)#10L]                                                                                                                                     +- Project
!         +- Project                                                                                                                                                                  +- ChildOutputHolder
!            +- ChildOutputHolder                                                                                                                                                        +- Project
!               +- Project                                                                                                                                                                  +- Filter (part#9 = part-1)
!                  +- Project                                                                                                                                                                  +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]
!                     +- Filter (part#9 = part-1)                                                                                                                            
!                        +- HiveTableRelation [`default`.`test`, org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Data Cols: [id#7, scope#8], Partition Cols: [part#9]]   
```

At the same time, a `ChildOutputHolder` node that fixes the original child node output can prevent `ColumnPruning` from continuously adding empty `Project` nodes, these empty `Project` nodes will eliminate some original nodes that may carry useful outputs.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
LennonChin: > Does this affect children that are not `ChildOutputHolder` ?

This check is to allow `PrivilegeBuilder.buildQuery` method continue drill down to the child nodes of the `ChildOutputHolder` node. Without this check, the `PrivilegeBuilder.buildQuery` method will terminate before hitting `ChildOutputHolder`. Since the outputs held by `ChildOutputHolder` are all useful, combined with `columnPrune` method, I think other child nodes will not be affected.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/plan/ChildOutputHolder.scala
wForget: Thank you for the detailed explanation, that makes a lot of sense.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/PrivilegesBuilder.scala
wForget: This logic is in `for (child <- p.children) {`, which will be applied to all children of `p`. Do we need it for children  of `p` that are not `ChildOutputHolder`?

LennonChin: > This logic is in `for (child <- p.children) {`, which will be applied to all children of `p`. Do we need it for children of `p` that are not `ChildOutputHolder`?

Normally, recursive checking of deeper nodes is necessary, but the presence of an `Aggregate(count(*))` node is an exception, as it blocks recursion. In current case, `ChildOutputHolder` is only added when encountered `Aggregate(count(*))` node. When a `ChildOutputHolder` node exists in the plan tree, it indicates that nodes deeper than `ChildOutputHolder` holds useful output information, we need to continue checking deeper nodes. At the same time, when recursing to the node deeper than `ChildOutputHolder`, we regress to normal judgment logic. Therefore, I think the judgment here is reasonable.

Issue #7206, opened 2025-09-17
[Bug] [Lineage]


lyne7-sc:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When a SQL query contains a subquery in the WHERE clause, the table referenced within the subquery are not included in the extracted upstream table lineage.

For example,
```sql
insert overwrite v2_catalog.db.tb3
select *
from v2_catalog.db.tb1 t1
where exists (select 1 from v2_catalog.db.tb2 t2 where t2.col1 = t1.col1);
```
the current result is:
```scala
Lineage(
        List("v2_catalog.db.tb1"),
        List("v2_catalog.db.tb3"),
        List(
          ("v2_catalog.db.tb3.col1", Set("v2_catalog.db.tb1.col1")),
          ("v2_catalog.db.tb3.col2", Set("v2_catalog.db.tb1.col2")),
          ("v2_catalog.db.tb3.col3", Set("v2_catalog.db.tb1.col3")))))
```
the output omits table `v2_catalog.db.tb2`, which is referenced in the filter condition.

So I propose to add a new a configuration to control whether to collect the tables referenced in filter conditions as lineage input tables

### Affects Version(s)

1.11.0

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #7207, opened 2025-09-18
[KYUUBI #7206] [Lineage] support collecting filter condition tables

lyne7-sc:

The current lineage analysis misses tables that are only referenced in filter condition subqueries (e.g., in `WHERE EXISTS` or `WHERE IN` clauses). This PR adds an option to include these upstream dependencies.

Fixes https://github.com/apache/kyuubi/issues/7206

UT added

No

### DIFF
--- a/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
+++ b/extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
@@ -580,8 +580,7 @@ trait LineageParser {
               extractColumnsLineage(
                 expression.plan,
                 ListMap[Attribute, AttributeSet](),
-                inputTablesByPlan
-              )
+                inputTablesByPlan)
             case _ =>
           }
         }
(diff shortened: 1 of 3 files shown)


extensions/spark/kyuubi-spark-lineage/src/main/scala/org/apache/kyuubi/plugin/lineage/helper/SparkSQLLineageParseHelper.scala
```diff
@@ -572,6 +572,24 @@ trait LineageParser {
               p.cacheBuilder.tableName.toSeq)
         }
 
+      case p: Filter =>
+        if (SparkContextHelper.getConf(
+            LineageConf.COLLECT_FILTER_CONDITION_TABLES_ENABLED)) {
```
yabola, line 577: I am not sure if we need this conf, it isn't a major behavior change and are always needed. We also found this issue before and have fixed it similar to this, but forgot to report it.   Others LGTM.

In addition: this change will not affect column lineage, only affect the input tables, because we set `parentColumnsLineage` as `ListMap[Attribute, AttributeSet]()` and doesn't `mergeColumnsLineage`

lyne7-sc: I'm still not certain whether filter tables should be treated as upstream tables in all scenarios. Since the existing tests did not account for the existence of these tables, I added this conf to avoid breaking them.

PR #7217, opened 2025-09-29
Support customized session protocol version to support binary type

turboFei:

For RESTful api, the session protocol version is set to V1 before and not open to users.
https://github.com/apache/kyuubi/blob/aadf86683554ba6932126c80c9e33bf84cea8dee/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala#L419

And before V6, the result is RowBased, since V6, it is ColumnBased.
https://github.com/apache/kyuubi/blob/aadf86683554ba6932126c80c9e33bf84cea8dee/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/cli/RowSetFactory.java#L29-L33

And with `HIVE_CLI_SERVICE_PROTOCOL_V1`, for `BINARY` type column, it is converted as STRING type for RESTful API.

In this PR, we make the session protocol open to users and support to transfer binary type values.

Why we need to support binary type? Because some bytes can not be converted to UTF-8 String.

UT.

No.

### DIFF
--- a/docs/client/rest/rest_api.md
+++ b/docs/client/rest/rest_api.md
@@ -99,9 +99,10 @@ Create a session
 
 #### Request Parameters
 
-| Name    | Description                      | Type |
-|:--------|:---------------------------------|:-----|
-| configs | The configuration of the session | Map  |
+| Name            | Description                                                  | Type |
+|:----------------|:-------------------------------------------------------------|:-----|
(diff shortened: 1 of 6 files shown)


pan3793 (approved):

turboFei:
thanks, merged to 1.11.0

PR #7220, opened 2025-10-02
Proactive cleanup kubernetes terminated application info

turboFei:

Proactive cleanup kubernetes terminated application info instead of rely on the guava cache internal expiration.

GA.

No.


turboFei:
FROM GPT:

cache.cleanUp() forces the cache to process pending evictions (expired entries, over-capacity entries, etc.) right away instead of waiting for normal access patterns to trigger cleanup.

turboFei:
https://github.com/apache/kyuubi/blob/e613d7f6e4e0b4143dde8bb6ae23d1252698e639/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala#L179-L181

Here we respect the `retainPeriod` for `expireAfterWrite`.

turboFei:
thanks, merged to master(1.11.0)

Issue #7226, opened 2025-10-21
[Bug] Kyuubi OOM when polling for status of spark driver for Simultaneous Large Number of Batch Jobs


JoonPark1:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

This occurs for Kubernetes Cluster deployment mode for kyuubi over k8 cluster. When multiple large number of spark batch jobs are submitted to kyuubi, the kyuubi attempts to spin up large number of spark drivers to handle per batch job. However, when there is heavy load, it can cause kyuubi to store records via MetadataManager about the state of each batch job as "PENDING" and repeated polling about each batch job's status until it runs out of memory. Then, upon next restart of kyuubi pod container, it again repeatedly polls as the records never gets updated as all the spark drivers to handle batch jobs never gets created and scheduled in first place. This causes the records in Metadata Store for kyuubi to persist regarding batch jobs as "state" field of value "PENDING" and "engine_state" field of value "UNKNOWN". The records can never get resolved and the repeated polling continues causing subsequent restarts of kyuubi to run out of memory. 

### Affects Version(s)

v1.10.2

### Kyuubi Server Log Output

```logtalk
2025-10-21 16:03:48.837 WARN org.apache.kyuubi.engine.KubernetesApplicationOperation: Waiting for driver pod with label: kyuubi-unique-tag=406b8bc1-457e-4115-ae55-50a0d39c061c to be created, elapsed time: 92106ms, return UNKNOWN status
2025-10-21 16:03:48.929 WARN org.apache.kyuubi.engine.KubernetesApplicationOperation: Waiting for driver pod with label: kyuubi-unique-tag=6d439666-0484-4902-9dab-39ad39f96b3e to be created, elapsed time: 92291ms, return UNKNOWN status
2025-10-21 16:03:48.929 WARN org.apache.kyuubi.engine.KubernetesApplicationOperation: Waiting for driver pod with label: kyuubi-unique-tag=fc5979a0-5599-4088-90ee-c5e995e0fca7 to be created, elapsed time: 92365ms, return UNKNOWN status
[...]

PR #7227, opened 2025-10-21
Fix Kyuubi OOM bug when multiple batch jobs are submitted at once in large amount

JoonPark1:
…cordingly once engine submit timeout is reached - prevent subsequent kyuubi OOM

This PR change is to address bug #7226. It updates the behavior of updating metadata store accordingly for batch jobs that have timed out due to waiting for available spark driver engine. This will fix the subsequent restarted kyuubi server from repeatedly polling for the spark application status of each and every batch job, which can cause consecutive OOM errors under k8 cluster deployment mode for kyuubi. 

This patch was tested through integration test that was added to test suite class called "SparkOnKubernetesTestsSuite.scala". 

No!

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
@@ -181,36 +181,51 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
   @VisibleForTesting
   private[kyuubi] def recoverBatchSessions(): Unit = withBatchRecoveryLockRequired {
     val recoveryNumThreads = conf.get(METADATA_RECOVERY_THREADS)
+    val recoveryBatchSize: Int = conf.get(BATCH_SESSIONS_RECOVERY_SIZE)
     val batchRecoveryExecutor =
       ThreadUtils.newDaemonFixedThreadPool(recoveryNumThreads, "batch-recovery-executor")
     try {
-      val batchSessionsToRecover = sessionManager.getBatchSessionsToRecover(connectionUrl)
-      val pendingRecoveryTasksCount = new AtomicInteger(0)
-      val tasks = batchSessionsToRecover.flatMap { batchSession =>
-        val batchId = batchSession.batchJobSubmissionOp.batchId
-        try {
-          val task: Future[Unit] = batchRecoveryExecutor.submit(() =>
-            Utils.tryLogNonFatalError(sessionManager.openBatchSession(batchSession)))
-          Some(task -> batchId)
-        } catch {
-          case e: Throwable =>
-            error(s"Error while submitting batch[$batchId] for recovery", e)
-            None
-        }
-      }
+      val offset: Int = 0
+      val shouldFetchRemainingBatchSessions: Boolean = true
+      val totalBatchRecovered: Int = 0
+      while(shouldFetchRemainingBatchSessions) {
+        val batchSessionsToRecover = sessionManager.getBatchSessionsToRecover(connectionUrl, offset, recoveryBatchSize)
+        if(batchSessionsToRecover.length > 0){
+          val pendingRecoveryTasksCount = new AtomicInteger(0)
+          val tasks = batchSessionsToRecover.flatMap { batchSession =>
+            val batchId = batchSession.batchJobSubmissionOp.batchId
+            try {
+              val task: Future[Unit] = batchRecoveryExecutor.submit(() =>
+                Utils.tryLogNonFatalError(sessionManager.openBatchSession(batchSession)))
+              Some(task -> batchId)
+            } catch {
+              case e: Throwable =>
+                error(s"Error while submitting batch[$batchId] for recovery", e)
+                None
+            }
+          }
 
-      pendingRecoveryTasksCount.addAndGet(tasks.size)
+          pendingRecoveryTasksCount.addAndGet(tasks.size)
 
-      tasks.foreach { case (task, batchId) =>
-        try {
-          task.get()
-        } catch {
-          case e: Throwable =>
-            error(s"Error while recovering batch[$batchId]", e)
-        } finally {
-          val pendingTasks = pendingRecoveryTasksCount.decrementAndGet()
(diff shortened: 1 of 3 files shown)


turboFei:
Hi @JoonPark1 
Thanks for the contribution.

For this issue, does it has chance to update metadata in
https://github.com/apache/kyuubi/blob/e8bbf5204ad9c875ca863a30168b4604491ec182/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala#L169-L189

JoonPark1:
Hey @turboFei. I believe the spark driver engine state and spark app state will be updated for metadata store...

JoonPark1:
@turboFei Sure! Once the kyuubi batch job times out because the elapsed time exceeds the configured submitTimeout property value (no spark driver is instantiated and in running state to handle the submitted batch job), the metadata about the spark application and the spark driver engine state is updated accordingly via "org.apache.kyuubi.server.metadata.MetadataManager" class' updateMetadata method which takes in the new up-to-date Metadata construct object instance (which is instance of class "org.apache.kyuubi.server.metadata.api.Metadata"). Then, internally within the manager class, the method calls upon the "org.apache.kyuubi.server.metadata.MetadataStore" class' updateMetadata method, which keeps the data regarding the state of each submitted kyuubi batch jobs utilizing spark compute engine as in-sync with the state of kyuubi's metadata store in relationalDB. As you can see, the whole flow does not need to invoke the BatchJobSubmission:: updateBatchMetadata to update the kyuubi's metadata store instance.

JoonPark1:
@turboFei. Hey Fei! You are right about the flow of the call-site... basically, the polling for status of submitted batch job originates from BatchJobSubmission:currentApplicationInfo... However, I don't think my change at the level of KubernetesApplicationOperation class calls the method BatchJobSumission::updateBatchMetadata. Do you suggest I refactor my change so that the update to kyuubi metadata happens through the BatchJobSubmission internal method? The alternative call-site flow will be BatchJobSumission::updateBatchMetadata() -> KyuubiSessionManager::updateMetadata(). I think this way also works as you proposed because it also works with relevant fields for Metadata instance associated with the specific batch job submission on kyuubi side... Let me know what you think.

turboFei:
@JoonPark1 
What is the kyuubi version you use?

I have not met this issue ever with the latest Kyuubi code base.

turboFei:
I am using the latest kyuubi based on master branch, and do not meet this issue.

JoonPark1:
Hey @turboFei... I know you said the current latest version of kyuubi does not face this issue. Have you tried replicating it and checking the state of metadata store to see if the polling for the batch job that timed out is updated appropriately?

JoonPark1:
My bad @turboFei. The kyuubi version we are using that's being affected by the issue is v1.10.2!

turboFei:
I think we can limit the number to recover in one batch,

https://github.com/apache/kyuubi/blob/53074ec1ab0c88a2eba3af622ed382602a20ca40/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala#L182

Do not recover all batches together.


For example, there are 400 batches to recover on restarting.

We can add a config, to recover 50 batches one time, and wait the batches engineId ready or wait a maximum interval before next recovery. how do you think about?

@JoonPark1

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiRestFrontendService.scala
```diff
@@ -181,36 +181,51 @@ class KyuubiRestFrontendService(override val serverable: Serverable)
   @VisibleForTesting
   private[kyuubi] def recoverBatchSessions(): Unit = withBatchRecoveryLockRequired {
     val recoveryNumThreads = conf.get(METADATA_RECOVERY_THREADS)
+    val recoveryBatchSize: Int = conf.get(BATCH_SESSIONS_RECOVERY_SIZE)
```
ruanwenjun, line 184: How about directly use `recoveryNumThreads` as the batch size? Then we can avoid import a new config here, and if the batch size larger than `recoveryNumThreads`, seems meaningless.

turboFei: its default values is only 10 by defaults

ruanwenjun: If we want greater concurrency, we should be able to increase this value. 
My concern is that we've added an extra configuration parameter, but this parameter is actually only used to control the batch size for queries from the database. Query results are also subject to `recoveryNumThreads ` here.
If we use `recoveryNumThreads` as the batch size, it would be simpler to use, we only need to consider concurrency.

JoonPark1: Now that I think about it, if we want the option to fetch greater batch size from the db for batch-job submission records, we can use reuse existing `recoveryNumThreads `config as the default is 10 but this is a config-value that could accept custom-value to recovery more batches at a time... @ruanwenjun

ruanwenjun: OK, I'm not sure if I understand this scene. In my view, what needs to be done here is to query the tasks requiring recovery from the database and then commit them. If the number of tasks queried exceeds the number of concurrent commits, it seems pointless.

JoonPark1: Basically, the method` sessionManager.getBatchSessionsToRecover(connectionUrl, offset, recoveryNumThreads)` recovers `recoveryNumThreads` batches from metadata store whose batches didn't reach terminal FINISHED state and reattempts it upon kyuubi server restart and then each recovered session is opened via kyuubi session manager: `sessionManager.openBatchSession(batchSession)` But, on second thought, I think it's better to control behavior of how recovery thread for each batch is done concurrently instead of constantly opening up new batch sessions for batches that need recovery without confirming prior opened batch sessions' associated batch job if it actually began execution (due to spinned up compute engine by kyuubi server) @ruanwenjun

PR #7229, opened 2025-10-23
Fix countMetadata if the params is empty

turboFei:

Without this PR.
It will meet issue if the params is empty. 
```
jdbcMetadataStore.countMetadata(MetadataFilter())
```
image

image

UT.

No.


turboFei:
thanks, merged to 1.11.0

PR #7232, opened 2025-10-26
Support to get kyuubi server event with RESTful api

turboFei:

Support to get kyuubi server event with RESTful api

UT.

No.


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/ApiUtils.scala
```diff
@@ -136,6 +137,22 @@ object ApiUtils extends Logging {
...
+  def serverEvent(serverEvent: KyuubiServerInfoEvent): KyuubiServerEvent = {
+    if (serverEvent == null) return new KyuubiServerEvent()
+    new KyuubiServerEvent(
+      serverEvent.serverName,
+      serverEvent.startTime,
+      serverEvent.eventTime,
+      serverEvent.state,
+      serverEvent.serverIP,
+      serverEvent.serverConf.asJava,
+      serverEvent.serverEnv.asJava,
```
pan3793, line 151: for a security perspective, we should redact sensitive info by default, maybe we can have a server side config to control the behavior, the candidate value can be `ORIGINAL`, `REDACTED`, `NONE`

pan3793: BTW, I remember there was a similar discussion about retrieving batch info

turboFei: addressed

PR #7250, opened 2025-11-18
Fix JDBCMetadataStore Scala 2.13 compile error

ruanwenjun:
Fix CI failed https://github.com/apache/kyuubi/actions/runs/19451268258/job/55656320605?pr=7249

Test by CI compiled

No


ruanwenjun:
@turboFei Hi Fei, please help me retry the failed test case.

pan3793:
@ruanwenjun, thanks, merged to master.

BTW, we don't use tags like `[Chore]` in PR title, but prefer using concise and accurate short sentences. For this case, it would be great to find which PR introduces the issue and fix it as a follow-up instead of a standalone fix.

PR #7272, opened 2025-12-14
Improve `build/dependency.sh` script

pan3793:

This PR improves `build/dependency.sh` script inspired by Spark `dev/test-dependencies.sh`

Main advantage - the updated script uses `mvn jar:jar jar:test-jar install:install clean -q` to perform a noop install, which is super faster than making a normal `mvn install` because it triggers the dep resolution without compiling the source code.

Pass GHA.

No.


turboFei (approved):

pan3793:
merged to master/1.10

PR #7273, opened 2025-12-14
Bump Netty 4.2.7

pan3793:

Upgrade Netty 4.2, which is a blocker for Spark 4.1 support.

https://github.com/netty/netty/wiki/Netty-4.2-Migration-Guide

Pass GHA.

No.


kyuubi-ha/pom.xml
```diff
@@ -72,6 +72,11 @@
             <artifactId>grpc-util</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-classes-epoll</artifactId>
```
pan3793, line 77: it's not a new dep, just declare it explicitly

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -255,6 +255,15 @@
...
                         </includes>
+                        <excludes>
+                            <exclude>io.netty:netty-codec-*-quic</exclude>
+                            <exclude>io.netty:netty-resolver-dns-*-macos</exclude>
+                            <exclude>io.netty:netty-tcnative-boringssl-static:*:osx-*</exclude>
+                            <exclude>io.netty:netty-tcnative-boringssl-static:*:windows-*</exclude>
+                            <exclude>io.netty:netty-transport-*-kqueue</exclude>
+                            <exclude>io.netty:netty-transport-*-io_uring</exclude>
+                            <exclude>io.netty:netty-transport-native-epoll:*:linux-riscv64</exclude>
+                        </excludes>
```
pan3793, line 266: weirdly, maven resolves those transitive deps of `spark-sql`(provided scope) as runtime scope, so I must exclude them ...

turboFei: can we exclude them from spark-sql depencny?

turboFei: Same as https://github.com/apache/celeborn/pull/3559/changes#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8

pan3793: @turboFei we can't do that.

currently, kyuubi pulls a few netty deps from jetcd, which uses gRPC and thus depends on grpc-netty. spark uses more netty deps for internal rpc

PR #7274, opened 2025-12-15
Bump Scala 2.13.17

pan3793:

Align Scala version with Spark 4.1.

Exclude `@SparkLocalClusterTest` due to Scala 2.13.17 changes `serialVersionUID` of `scala.collection.immutable.ArraySeq`. This should only be a test issue, I suppose we still preserve binary compatibility for the Spark engine jar - which means, the `kyuubi-spark-sql-engine_2.13-<version>.jar` compile against Spark 3.5, still works well for Spark 4.0, and the upcoming Spark 4.1

GHA.

No.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2053,7 +2053,8 @@
                 <!-- TODO: update once Paimon support Spark 4.0.
                            paimon-spark-3.5 contains Scala 2.12 classes cause conflicts with Scala 2.13 -->
                 <paimon.artifact>paimon-common</paimon.artifact>
-                <maven.plugin.scalatest.exclude.tags>org.scalatest.tags.Slow,org.apache.kyuubi.tags.PaimonTest,org.apache.kyuubi.tags.HudiTest</maven.plugin.scalatest.exclude.tags>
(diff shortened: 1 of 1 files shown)


pan3793:
> Why are PR's not mereged and just closed?

@Chaho12 we use a script instead of GitHub's "merge button" to merge PR, so the final status is inconsistent, this patch has been included in v1.11.0.


> BTW kyuubi 1.11.1 seems that it is not built with scala 2.13, it just supports 2.13 for now right?

The pre-built binary distribution mainly uses Scala 2.12, with additional `externals/engines/spark/kyuubi-spark-sql-engine_2.13-1.11.1.jar`, Kyuubi will pick the right engine jar according to your `SPARK_HOME`. let's say you configure `SPARK_HOME` in `conf/kyuubi-env.sh` to `/path/of/spark-3.5`(built with scala 2.12), it picks `kyuubi-spark-sql-engine_2.12-1.11.1.jar`, and if `SPARK_HOME` points to `/path/of/spark-4.1`(built with scala 2.13), it picks `kyuubi-spark-sql-engine_2.13-1.11.1.jar`

Issue #7270, opened 2025-12-11
[FEATURE] # Enable BasicAuth for Prometheus metrics endpoint


Yao-MR:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

For now, there is no auth for the metrics exporter such as Prometheus,

Would like to support for the basic auth for the Prometheus metrics endpoint 

### Motivation

in our customer Scenario,  Security is very important part for the it security, and the auth for the metrics is 
necessary

### Describe the solution

we suggest introduce one auth module to ingest in to the Prometheus metrics endpoint, which is just add one http request filter to add this module

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #7276, opened 2025-12-17
[KYUUBI #7270]Enable BasicAuth for Prometheus metrics endpoint

Yao-MR:
Add the BasicAuth For the Prometheus exporter

Test in UT

No

### DIFF
--- a/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/BasicAuthFilter.scala
+++ b/kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/BasicAuthFilter.scala
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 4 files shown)


kyuubi-metrics/src/main/scala/org/apache/kyuubi/metrics/BasicAuthFilter.scala
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Base64
+import javax.servlet._
+import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
+
+import org.apache.kyuubi.Logging
+
+/**
+ * A servlet filter that implements HTTP Basic Authentication
+ */
+class BasicAuthFilter(username: String, password: String)
```
pan3793, line 30: can we reuse the infrastructure of the RESTful API?

Yao-MR: sure, will refactor the auth module for further reuse

PR #7277, opened 2025-12-17
Add UUID v7 generator

pan3793:

https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-7

https://uuidv7.org/

> The Key Benefits of UUIDv7
UUIDv7 is rapidly gaining traction for various use cases across many industries. Here are some of the primary benefits:
>
> 1. Time-Ordered and Sortable: UUIDv7 encodes the time of creation as part of the identifier, making it naturally sortable. This is especially valuable for systems where data needs to be ordered by time without needing an additional timestamp field.
> ...

I think it's very useful - e.g., use UUIDv7 as session ID, then we can infer the session creation time from the UUID itself, or use UUIDv7 as a staging dir name, then we can easily clean dangling staging folders (for example, created 3 months ago) with a prefix.

Implementation is inspired by https://github.com/apache/iceberg/pull/14700

New UTs are added.

No.

### DIFF
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/UuidUtils.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/UuidUtils.java
@@ -29,9 +29,9 @@ public class UuidUtils {
 
   private static final SecureRandom SECURE_RANDOM = new SecureRandom();
 
-  public static UUID generateUuidV7() {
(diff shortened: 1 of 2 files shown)


cxzl25 (approved):

pan3793:
thanks, merged to master

PR #7289, opened 2025-12-26
Release requires Java 17 or 21

pan3793:

I hit an issue while preparing 1.11.0 RC0, which indicates I need to use a higher version of JDK for releasing.

```
[INFO] --- antlr4:4.13.1:antlr4 (default) @ kyuubi-extension-spark-4-0_2.13 ---
[WARNING] Error injecting: org.antlr.mojo.antlr4.Antlr4Mojo
java.lang.UnsupportedClassVersionError: org/antlr/v4/Tool has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
    at java.lang.ClassLoader.defineClass1 (Native Method)
    at java.lang.ClassLoader.defineClass (ClassLoader.java:756)
    at java.security.SecureClassLoader.defineClass (SecureClassLoader.java:142)
```

In https://github.com/apache/kyuubi/pull/7144, we made it always use `-release:8` for `jdk9+`, so we still produce Java 8 compatible bytecode even using JDK 17, for all modules except Spark 4.0/4.1 extension modules.

Will verify this in the next RC of 1.11.0

No.

### DIFF
--- a/build/release/release.sh
+++ b/build/release/release.sh
@@ -62,8 +62,8 @@ else
 fi
 
 JAVA_VERSION=$($JAVA -version 2>&1 | awk -F '"' '/version/ {print $2}')
-if [[ $JAVA_VERSION != 1.8.* ]]; then
-  echo "Unexpected Java version: $JAVA_VERSION. Java 8 is required for release."
+if [[ $JAVA_VERSION != 17.* && $JAVA_VERSION != 21.* ]]; then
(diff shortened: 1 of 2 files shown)


pan3793:
cc @wForget

also cc @aajisaka, this only affects 1.11.0, for 1.10.3, release still uses Java 8

pan3793:
thanks, merged to master/1.11, will continue the 1.11.0 release

PR #7295, opened 2026-01-04
Upgrade components of play ground to recent versions

pan3793:

Upgrade components of play ground to recent versions, in addition, Kyuubi and Spark switch to JDK 17, while other components like Hadoop and Hive, remamin using JDK 8.

Tested locally by building images and run demo, the updated docker images will be available on DockerHub soon, reviewer can test it too.

No.


docker/playground/image/kyuubi-playground-base.Dockerfile
```diff
@@ -10,13 +10,23 @@
...
     ln -snf /usr/bin/bash /usr/bin/sh && \
     apt-get update -q && \
-    apt-get install -yq retry busybox && \
+    apt-get install -yq \
+      retry \
+      busybox \
+      openjdk-8-jdk-headless \
+      openjdk-17-jdk-headless && \
     rm -rf /var/lib/apt/lists/* && \
+    update-java-alternatives --set $(update-java-alternatives --list | grep java-1.8.0-openjdk | awk '{print $NF}') || \
```
pan3793, line 31: use `||` to ignore the error code returned by `update-java-alternatives` command, as JDK 8 lacks some commands provided by mordern JDKs

docker/playground/.env
```diff
@@ -18,15 +18,15 @@
...
+KYUUBI_VERSION=1.11.0
 KYUUBI_HADOOP_VERSION=3.3.6
 POSTGRES_VERSION=12
 POSTGRES_JDBC_VERSION=42.3.4
 SCALA_BINARY_VERSION=2.12
-SPARK_VERSION=3.4.4
-SPARK_BINARY_VERSION=3.4
+SPARK_VERSION=3.5.7
+SPARK_BINARY_VERSION=3.5
 SPARK_HADOOP_VERSION=3.3.4
```
pan3793, line 29: Spark 4 uses Hadoop client 3.4, which switches to AWS SDK 2.x, requires more work, so let's keep using Spark 3.5 for now. this also matches the current state of Kyuubi project - default Spakr version is 3.5

docker/playground/README.md
```diff
@@ -10,24 +10,35 @@ Playground
 
 ### Play
 
-1. Connect using `beeline`
+1. Connect using `kyuubi-beeline`
 
-`docker exec -it kyuubi /opt/kyuubi/bin/beeline -u 'jdbc:hive2://0.0.0.0:10009/tpcds/tiny'`;
+```
+docker exec -it kyuubi /opt/kyuubi/bin/kyuubi-beeline -u 'jdbc:hive2://0.0.0.0:10009/tpcds/tiny'
```
pan3793: we recommend using `kyuubi-beeline` instead of `beeline`, to distinguish from Hive/Spark's `beeline`

docker/playground/image/kyuubi-playground-hadoop.Dockerfile
```diff
@@ -29,9 +28,7 @@ RUN set -x && \
     tar -xzf ${HADOOP_TAR_NAME}.tar.gz -C /opt && \
     ln -s /opt/hadoop-${HADOOP_VERSION} ${HADOOP_HOME} && \
     rm ${HADOOP_TAR_NAME}.tar.gz && \
-    HADOOP_CLOUD_STORAGE_JAR_NAME=hadoop-cloud-storage && \
-    wget -q ${MAVEN_MIRROR}/org/apache/hadoop/${HADOOP_CLOUD_STORAGE_JAR_NAME}/${HADOOP_VERSION}/${HADOOP_CLOUD_STORAGE_JAR_NAME}-${HADOOP_VERSION}.jar -P ${HADOOP_HOME}/share/hadoop/hdfs/lib && \
```
pan3793, line 33: `hadoop-cloud-storage` is a package for assembling, has no classes

docker/playground/image/kyuubi-playground-base.Dockerfile
```diff
@@ -10,13 +10,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM eclipse-temurin:8-focal
+FROM ubuntu:focal
```
aajisaka, line 13: Ubuntu 20.04 (Focal Fossa) is already end of standard support.

https://ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare

Let's upgrade the OS version in this PR or a separate PR.

pan3793: yes, we should move forward.

one additional consideration, we'd better align it with hadoop dev container, otherwise there might be some issues when using hadoop native libs, especially when users play with security configs. e.g., ubuntu focal is the latest version that provides openssl 1.x, the hadoop native libs shipped by official release compile against ubuntu focal with openssl 1.x, a runtime linkage issue will be thrown if we try to enable kerberos on ubuntu jammy or noble.

but I think it has no issues for SIMPLE mode.

pan3793: not related to this PR, another issue related to Hadoop and Ubuntu, the APT repo's `jsvc` is too old to support modern JDK, as Hadoop trunk is moving to JDK 17+, this could be another noisy for users to run kerberized Hadoop with JDK 17+ on Ubuntu, maybe we should contact Debian or Ubuntu Java team to upgrade it ...

aajisaka: > APT repo's jsvc is too old

Found very old issue https://bugs.launchpad.net/ubuntu/+source/commons-daemon/+bug/1788154

aajisaka: FYI: Filed https://issues.apache.org/jira/browse/HADOOP-19774 to use Ubuntu 24.04 in Hadoop

PR #7299, opened 2026-01-08
[AUTHZ] Remove redundant concatenation with the list of conditions

attilapiros:

This concatenation with the list of conditions `conditionList` is redundant as the `columnPrune` method already contains it.

Existing unit tests.

No.


pan3793:
thanks, merged to master/1.11.1/1.10.4

PR #7300, opened 2026-01-11
Add pyhive connection timeout

ryanbordo:
Adds an optional parameter to pyhive hive connection class init to allow setting the socket connection timeout. By default, thrift socket connections have no timeout which causes connections to hang indefinitely when something goes wrong. This helps us in [dbt ](https://github.com/dbt-labs/dbt-adapters/blob/1657e9084fcafa0a1bee0b7362c870851c2d173c/dbt-spark/src/dbt/adapters/spark/connections.py#L542 )by allowing for multi-threaded management of connections while continuing to rely on pyhive thrift setup.

Reference thrift lines:
https://github.com/apache/thrift/blob/0d18fb2e97a00fc56997fa059d6819e54cdff64b/lib/py/src/transport/THttpClient.py#L130
https://github.com/apache/thrift/blob/0d18fb2e97a00fc56997fa059d6819e54cdff64b/lib/py/src/transport/TSocket.py#L111

Manually on TSocket path and added unit test

No

### DIFF
--- a/python/pyhive/hive.py
+++ b/python/pyhive/hive.py
@@ -161,7 +161,7 @@ class Connection(object):
         ssl_cert=None,
         thrift_transport=None,
         ssl_context=None,
-        socket_timeout=None
+        connection_timeout=None,
     ):
         """Connect to HiveServer2
 
@@ -176,7 +176,7 @@ class Connection(object):
             Incompatible with host, port, auth, kerberos_service_name, and password.
         :param ssl_context: A custom SSL context to use for HTTPS connections. If provided,
             this overrides check_hostname and ssl_cert parameters.
-        :param socket_timeout: Millisecond timeout for the Thrift socket connections.
+        :param connection_timeout: Millisecond timeout for Thrift connections.
         The way to support LDAP and GSSAPI is originated from cloudera/Impyla:
         https://github.com/cloudera/impyla/blob/255b07ed973d47a3395214ed92d35ec0615ebf62
         /impala/_thrift_api.py#L152-L160
@@ -195,6 +195,8 @@ class Connection(object):
                 ),
                 ssl_context=ssl_context,
             )
+            if connection_timeout:
+                thrift_transport.setTimeout(connection_timeout)
 
             if auth in ("BASIC", "NOSASL", "NONE", None):
                 # Always needs the Authorization header
@@ -238,8 +240,8 @@ class Connection(object):
             if auth is None:
                 auth = 'NONE'
             socket = thrift.transport.TSocket.TSocket(host, port)
-            if socket_timeout:
-                socket.setTimeout(socket_timeout)
+            if connection_timeout:
+                socket.setTimeout(connection_timeout)
             if auth == 'NOSASL':
                 # NOSASL corresponds to hive.server2.authentication=NOSASL in hive-site.xml
                 self._transport = thrift.transport.TTransport.TBufferedTransport(socket)
--- a/python/pyhive/hive.py
+++ b/python/pyhive/hive.py
@@ -195,7 +195,7 @@ class Connection(object):
                 ),
                 ssl_context=ssl_context,
             )
-            if connection_timeout:
+            if connection_timeout is not None:
                 thrift_transport.setTimeout(connection_timeout)
 
             if auth in ("BASIC", "NOSASL", "NONE", None):
@@ -240,7 +240,7 @@ class Connection(object):
             if auth is None:
                 auth = 'NONE'
             socket = thrift.transport.TSocket.TSocket(host, port)
-            if connection_timeout:
+            if connection_timeout is not None:
                 socket.setTimeout(connection_timeout)
             if auth == 'NOSASL':
                 # NOSASL corresponds to hive.server2.authentication=NOSASL in hive-site.xml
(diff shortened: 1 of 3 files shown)


python/pyhive/hive.py
```diff
@@ -236,6 +240,8 @@ def __init__(
             if auth is None:
                 auth = 'NONE'
             socket = thrift.transport.TSocket.TSocket(host, port)
+            if connection_timeout:
+                socket.setTimeout(connection_timeout)
```
fbertsch, line 244: FYI this sets both connection and socket timeout. There's a nice description of the difference [here](https://stackoverflow.com/questions/7360520/connectiontimeout-versus-sockettimeout).
```suggestion
                socket.setConnectTimeout(connection_timeout)
```

python/pyhive/hive.py
```diff
@@ -236,6 +240,8 @@ def __init__(
             if auth is None:
                 auth = 'NONE'
             socket = thrift.transport.TSocket.TSocket(host, port)
+            if connection_timeout:
```
fbertsch: This will not set the timeout if it's equal to 0, but that is [interpreted by the underlying Socket class as an "infinite wait" duration](https://docs.oracle.com/javase/8/docs/api/java/net/Socket.html#connect-java.net.SocketAddress-int-). It might be worth letting users set this to 0
```suggestion
            if connection_timeout is not None:
```

python/pyhive/hive.py
```diff
@@ -160,7 +160,8 @@ def __init__(
         check_hostname=None,
         ssl_cert=None,
         thrift_transport=None,
-        ssl_context=None
+        ssl_context=None,
+        connection_timeout=None,
```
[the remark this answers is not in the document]
ryanbordo: Omitting since this is not commonplace here

python/pyhive/hive.py
ryanbordo: Good call, just added

python/pyhive/hive.py
ryanbordo: That's a good mention. My goal with this was originally to manage the socket timeout. However, I think I will leave this way to not get too granular.

pan3793: In Java, a socket can have different `connectTimeout`, `readTimeout`, and `writeTimeout` values. Does Python have an equivalent? Generally, we want a small  `connectTimeout`, and long `readTimeout` and `writeTimeout`, so it can fail fast on connecting dead server and failover to health one, if it has multiple server instances

ryanbordo: From my short research, it looks like the short answer is no, but possible with an explicit `create_connection` At least, python thrift is only using supporting one timeout and not doing this.
https://github.com/apache/thrift/blob/c99d09a231648d72e05a89d80281b38c9d0d1b9a/lib/py/src/transport/TSocket.py#L145-L147

pan3793: we can call `socket.setTimeout` after the socket is connected, right? if so, it's easy to implement `connectTimeout`, `socketTimeout`

ryanbordo: That would be possible. Are you suggesting to add in here or in the Thrift library? I also want to call out that a user could choose to do this already by passing a thrift transport object.

pan3793: > user could choose to do this already by passing a thrift transport object.

the user passed `thrift_transport` should already be established? if so, the user takes responsibility to set `connectTimeout`.

I think what we should do here is

1. add two args `connect_timeout` and `socket_timeout`
2. for `thrift_transport`, suppose it is already established, just set `socket_timeout`
3. for host/port case, use `connect_timeout` for the connecting phase, and change it to `socket_timeout` after the connection is established

ryanbordo: The connection from the `thrift_transport` actually doesn't need to be established. And in our dbt code, we have a [path](https://github.com/dbt-labs/dbt-adapters/blob/b9ebd240e39882a8c43ed659de423c7504d4642a/dbt-spark/src/dbt/adapters/spark/connections.py#L533) that relies on this connection being opened in `hive.connect` function. Unfortunately, the case for passing a transport object is ambiguous since thrift python only has one `setTimeout` for both. It might make sense to skip both timeouts when a transport is passed? Regardless, I'll get started on the proposal you sent but let me know if you have any additional thoughts.

pan3793: okay, I read the code, `self._transport.open()` happens in Connection `__init__`, so we can call `setTimeout(connect_time)` before it, and call `setTimeout(socket_time)`, same for both host/port and thrift_transport cases

ryanbordo: Got it. Thanks for the feedback.

ryanbordo: Pushed up a solution for this, don't love it but should work. We can't guarantee a public API for setting a socket timeout so I believe it's this or reverting to only setting a timeout once, for connection and socket. Let me know what you think

ryanbordo: I'm leaning towards reverting to the connection timeout only, for both connections and socket, plus only on host, port combos for tradeoffs in simplicity. A user with a provided thrift object could fine tune themselves

ryanbordo: Reverting to the previous PR for these reasons. Plus, the pyhive connection wrapper provides polling and ability to run async, which lessons the need for a separate socket timeout. We can always add onto this given user feedback and a user can use a custom thrift transport to fine tune this still

https://github.com/apache/kyuubi/blob/c8a0ecbc5f60564054e603c670cf7b3baf52c947/python/pyhive/hive.py#L483
https://github.com/apache/kyuubi/blob/c8a0ecbc5f60564054e603c670cf7b3baf52c947/python/pyhive/hive.py#L524

Issue #7305, opened 2026-01-15
[Bug] DBeaver returned all tables in all databases when connecting.


zhzhenqin:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

DBeaver returned all tables in all databases when connecting.

<img width="2012" height="724" alt="Image" src="https://github.com/user-attachments/assets/2319c5ca-92ec-4fbf-9dbc-5fa8a7f62cd2" />

### Affects Version(s)

1.11.0

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

zhzhenqin:
as you see:

My StarRocks database "dataeye_starpony" only has 2 tables. However, DBeaver shows many more tables, and these extra tables cannot be queried.

wangzhigang1999:
Hi, I'm interested in fixing this issue. I've analyzed the root cause and have a working fix locally — I'm planning to open a new PR soon.

PR #7307, opened 2026-01-15
[KYUUBI #7305] Fix DBeaver connect returned all tables. return tables…

zhzhenqin:
return tables from current database now! doris/starrocks like mysql

### DIFF
--- a/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/MySQLDialect.scala
+++ b/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/MySQLDialect.scala
@@ -36,6 +36,10 @@ class MySQLDialect extends JdbcDialect {
     statement
   }
 
+  override def getSchemasOperation(catalog: String, schema: String): String = {
+    return "select database()"
+  }
+
   override def getTablesQuery(
       catalog: String,
       schema: String,
@@ -60,8 +64,11 @@ class MySQLDialect extends JdbcDialect {
       filters += s"$TABLE_CATALOG = '$catalog'"
     }
 
-    if (StringUtils.isNotBlank(schema)) {
+    // when DBeaver connect use %, kyuubi to return all tables from all other databases.
+    if (StringUtils.isNotBlank(schema) && !"%".equals(schema)) {
       filters += s"$TABLE_SCHEMA LIKE '$schema'"
+    } else {
+      filters += s"$TABLE_SCHEMA = database()"
     }
 
     if (StringUtils.isNotBlank(tableName)) {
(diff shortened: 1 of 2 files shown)


externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/MySQLDialect.scala
```diff
@@ -60,8 +64,11 @@ class MySQLDialect extends JdbcDialect {
       filters += s"$TABLE_CATALOG = '$catalog'"
     }
 
-    if (StringUtils.isNotBlank(schema)) {
+    // when DBeaver connect use %, kyuubi to return all tables from all other databases.
+    if (StringUtils.isNotBlank(schema) && !"%".equals(schema)) {
```
pan3793, line 68: I don't understand this change, I suppose `WHERE $TABLE_SCHEMA LIKE '%'` has the same effect as an empty condition

zhzhenqin: like MySQLDialect(mysql, doris, starrocks) use:

SELECT TABLE_CATALOG, TABLE_SCHEMA, TABLE_NAME, TABLE_TYPE, ENGINE,
TABLE_ROWS, AVG_ROW_LENGTH, DATA_LENGTH,
CREATE_TIME, UPDATE_TIME, TABLE_COLLATION, TABLE_COMMENT
FROM INFORMATION_SCHEMA.TABLES

当 Dbeaver 连接时，会采用 % 作为参数，这样或造成返回给 Dbeaver 所有库下的所有的表。
1. 和session配置的DB 不符，用户看到了其他数据库下的表（所有表）；
2. Dbeaver 在执行查询时，schema 不对，查询报错；

When connecting with Dbeaver, it uses '%' as a parameter, which may cause Dbeaver to return tables from all databases.

1. The database configuration may differ from the session's database, resulting in tables from other databases being displayed;
2. Dbeaver may encounter query errors due to an incorrect schema.

zhzhenqin: you can see issue: https://github.com/apache/kyuubi/issues/7305

My StarRocks database "dataeye_starpony" only has 2 tables. However, DBeaver shows many more tables, and these extra tables cannot be queried.

wangzhigang1999:
Hi @zhzhenqin, thanks for working on this! 
I noticed this PR has been inactive for about 3 months and is approaching the 100-day stale auto-close threshold. I've been looking into this issue as well and have a slightly different approach. 
Would you mind if I open a new PR to continue the work? Happy to credit your original contribution in the description.

pan3793:
@wangzhigang1999, please open a PR to demonstrate your solution. if it is based on others' work or has some reference, add the `Co-authored-by: Foo Bar <foobar@abc.org>` in your PR description

wangzhigang1999:
closed via https://github.com/apache/kyuubi/commit/98f81f2deab63b1b835be8ac9710f1290cca5f8f

PR #7317, opened 2026-01-29
[KYUUBI-7316] Fix parsePropertyFromUrl to remove query string from parsed values

wangyum:

Fixed the `parsePropertyFromUrl` method in `Utils.java` to properly remove query string parts (everything after ?) from parsed JDBC URL property values, making it consistent with the `extractURLComponents` method.

When parsing JDBC URLs with query strings like: `jdbc:hive2://host:10012/db;auth=JWT?kyuubi.session.cluster=clusterA`.
The `parsePropertyFromUrl` method was incorrectly returning `JWT?kyuubi.session.cluster=clusterA` for the `auth` property, while `extractURLComponents` correctly returned only `JWT`. This inconsistency could cause authentication failures in JWT auth scenarios with additional query parameters.

1. URLs with query strings parse properties correctly without the query string
2. URLs without query strings continue to work as expected
3. Both parsing methods return consistent results

No.

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
@@ -565,16 +565,14 @@ public class Utils {
   }
 
   public static String parsePropertyFromUrl(final String url, final String key) {
-    String clientPropertiesSection = url.split("[?#]")[0];
-    int semiColonIndex = clientPropertiesSection.indexOf(';');
-    if (semiColonIndex < 0) {
-      return null;
+    String urlClientPropertiesSection = url.split("[?#]")[0];
+    String[] tokens = urlClientPropertiesSection.split(";");
+    for (String token : tokens) {
+      if (token.trim().startsWith(key.trim() + "=")) {
+        return token.trim().substring((key.trim() + "=").length());
+      }
     }
-
-    String clientProperties = clientPropertiesSection.substring(semiColonIndex + 1);
-    Matcher matcher =
-        Pattern.compile("(?:^|;)" + Pattern.quote(key) + "=([^;]*)").matcher(clientProperties);
-    return matcher.find() ? matcher.group(1) : null;
+    return null;
   }
 
   /**


kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
```diff
@@ -568,7 +568,13 @@ public static String parsePropertyFromUrl(final String url, final String key) {
     String[] tokens = url.split(";");
     for (String token : tokens) {
       if (token.trim().startsWith(key.trim() + "=")) {
-        return token.trim().substring((key.trim() + "=").length());
+        String value = token.trim().substring((key.trim() + "=").length());
+        // Remove query string part (anything after ?) to match extractURLComponents behavior
+        int queryIndex = value.indexOf('?');
```
turboFei: https://github.com/apache/kyuubi/blob/dfd9b4935378c79d3c1f99e344273ff71e21765b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java#L482-L514

turboFei: see https://kyuubi.readthedocs.io/en/master/client/jdbc/kyuubi_jdbc.html#basic-connection-url-format

```
jdbc:subprotocol://host:port[/catalog]/[schema];<clientProperties;><[#|?]sessionProperties>
```

We can parse the clientProperties only to get auth property

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
```diff
@@ -568,7 +568,13 @@ public static String parsePropertyFromUrl(final String url, final String key) {
...
+        if (queryIndex > 0) {
```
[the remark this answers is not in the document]
wangyum: @copilot-pull-request-reviewer could you also add a test ?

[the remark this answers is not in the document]
wangyum: @Copilot could you also add a test ?

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
```diff
@@ -565,13 +565,16 @@ static int getVersionPart(String fullVersion, int position) {
...
+    int semiColonIndex = clientPropertiesSection.indexOf(';');
+    if (semiColonIndex < 0) {
+      return null;
     }
-    return null;
+
+    String clientProperties = clientPropertiesSection.substring(semiColonIndex + 1);
+    Matcher matcher =
+        Pattern.compile("(?:^|;)" + Pattern.quote(key) + "=([^;]*)").matcher(clientProperties);
+    return matcher.find() ? matcher.group(1) : null;
```
turboFei: 1. per the code, I can see that, this method is only used to parse clientPropertiesSection just likes this PR
Image

we can add some comments to make this method clear

2. we can make this PR simple (two lines change)
```
    String urlClientPropertiesSection = url.split("[?#]")[0];
    String[] tokens = urlClientPropertiesSection.split(";");
```

turboFei (approved):
LGTM, cc @pan3793

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java
```diff
@@ -565,7 +565,8 @@ static int getVersionPart(String fullVersion, int position) {
   }
 
   public static String parsePropertyFromUrl(final String url, final String key) {
-    String[] tokens = url.split(";");
+    String urlClientPropertiesSection = url.split("[?#]")[0];
```
pan3793, line 568: I can't recall if we support escaped char in URL, if not, LGTM

turboFei: image

turboFei:
thanks, merged to 1.10.4

PR #7319, opened 2026-01-30
Fix kyuubi sessions to return 404 for non-existent sessions

Yutong-1424:

This PR addresses below issue with the Kyuubi sessions API:

1**Incorrect HTTP status codes**: When attempting to delete a non-existent session, the API currently returns HTTP 500 (Internal Server Error), which is misleading. A non-existent resource should return HTTP 404 (Not Found) to properly indicate the resource doesn't exist.

 **Test for 404 response on delete non-existent session**: Added `delete_non-existent_admin_session_returns_404` test in `AdminResourceSuite.scala` that verifies deleting a non-existent session returns HTTP 404 instead of 500.

No

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
@@ -52,15 +52,11 @@ private[v1] class SessionsResource extends ApiRequestContext with Logging {
     content = Array(new Content(
       mediaType = MediaType.APPLICATION_JSON,
       array = new ArraySchema(schema = new Schema(implementation = classOf[SessionData])))),
-    description = "get the list of live sessions for the current user")
+    description = "get the list of all live sessions")
   @GET
   def sessions(): Seq[SessionData] = {
-    val userName = fe.getSessionUser(Map.empty[String, String])
-    sessionManager
-      .allSessions()
-      .filter(session => session.user == userName)
-      .map(session => ApiUtils.sessionData(session.asInstanceOf[KyuubiSession]))
-      .toSeq
+    sessionManager.allSessions()
+      .map(session => ApiUtils.sessionData(session.asInstanceOf[KyuubiSession])).toSeq
   }
 
   @ApiResponse(
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/ApiUtils.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/ApiUtils.scala
@@ -22,16 +22,13 @@ import scala.collection.JavaConverters._
 import org.apache.kyuubi.{Logging, Utils}
 import org.apache.kyuubi.client.api.v1.dto
 import org.apache.kyuubi.client.api.v1.dto.{OperationData, OperationProgress, ServerData, SessionData}
-import org.apache.kyuubi.config.KyuubiConf.SERVER_SECRET_REDACTION_PATTERN
 import org.apache.kyuubi.ha.client.ServiceNodeInfo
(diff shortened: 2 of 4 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/AdminResource.scala
```diff
@@ -205,8 +205,15 @@ private[v1] class AdminResource extends ApiRequestContext with Logging {
...
         s"$userName is not allowed to close the session $sessionHandleStr")
     }
-    fe.be.closeSession(SessionHandle.fromUUID(sessionHandleStr))
-    Response.ok(s"Session $sessionHandleStr is closed successfully.").build()
+    try {
+      fe.be.closeSession(SessionHandle.fromUUID(sessionHandleStr))
+      Response.ok(s"Session $sessionHandleStr is closed successfully.").build()
+    } catch {
+      case e: org.apache.kyuubi.KyuubiSQLException
+          if e.getMessage != null && e.getMessage.startsWith("Invalid ") =>
```
turboFei: The implementation is tricky.
Maybe you can check session exists directly, 
`fe.be.sessionManager.getSessionOption()`

Yutong-1424: Thanks for the review @turboFei ! That's good point, using getSessionOption() is much cleaner than catching the exception and matching on the message string. I'll update both AdminResource and SessionsResource to check session existence directly before calling closeSession. Will push the fix shortly.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -157,8 +161,15 @@ private[v1] class SessionsResource extends ApiRequestContext with Logging {
...
     info(s"Received request of closing $sessionHandleStr")
-    fe.be.closeSession(sessionHandleStr)
-    Response.ok().build()
+    try {
+      fe.be.closeSession(sessionHandleStr)
+      Response.ok().build()
+    } catch {
+      case e: org.apache.kyuubi.KyuubiSQLException
+          if e.getMessage != null && e.getMessage.startsWith("Invalid ") =>
+        throw new NotFoundException(
```
turboFei: please apply same change here

Yutong-1424: Done, Thanks!

Yutong-1424: Could you please review these changes again, to see if everything is good. @turboFei

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -52,11 +52,15 @@ private[v1] class SessionsResource extends ApiRequestContext with Logging {
     content = Array(new Content(
       mediaType = MediaType.APPLICATION_JSON,
       array = new ArraySchema(schema = new Schema(implementation = classOf[SessionData])))),
-    description = "get the list of all live sessions")
+    description = "get the list of live sessions for the current user")
```
turboFei: how about only fix the 404 issue in this PR?

Yutong-1424: sure, I will only include 404 fix and revert others

Yutong-1424:
Thanks! I would like to make another pr to the part that make Kyuubi sessions only return sessions belong to the calling user and redact credential data, and I would like to know if any concerns for this change

PR #7313, opened 2026-01-24
If the k8s application is pending, audit the status as well

turboFei:
### Why are the changes needed?

I saw that, a pod pending for 3 hours.
```
/var/log/hadoop/kyuubi/audit/k8s-audit.log.2026-01-23-08.gz::2026-01-23 08:09:31.356 INFO [-935552520-pool-3-thread-2861] org.apache.kyuubi.engine.KubernetesApplicationAuditLogger: eventType=UPDATE	label=cb370310-9c4c-4982-817e-85f4f6c031a9	context=kube-apiserver-a-spark	namespace=dls-prod	pod=kyuubi-hadp-adi-hadp-w-md-tbl-location-w-muso-0-ae7bd346-20260123-stm-cb370310-9c4c-4982-817e-85f4f6c031a9-driver	podState=Pending	containers=[]	appId=spark-f67cc33544654d828d36cc9553404a3e	appName=hadp-adi-hadp-w-md-t-818abc6d23397c33859c76dff9aa82e7244a92ba	appState=PENDING	appError=''
/var/log/hadoop/kyuubi/audit/k8s-audit.log.2026-01-23-11.gz::2026-01-23 11:09:50.121 INFO [-935552520-pool-3-thread-2902] org.apache.kyuubi.engine.KubernetesApplicationAuditLogger: eventType=UPDATE	label=cb370310-9c4c-4982-817e-85f4f6c031a9	context=kube-apiserver-a-spark	namespace=dls-prod	pod=kyuubi-hadp-adi-hadp-w-md-tbl-location-w-muso-0-ae7bd346-20260123-stm-cb370310-9c4c-4982-817e-85f4f6c031a9-driver	podState=Pending	containers=[]	appId=spark-f67cc33544654d828d36cc9553404a3e	appName=hadp-adi-hadp-w-md-t-818abc6d23397c33859c76dff9aa82e7244a92ba	appState=PENDING	appError=''
```

However, the audit log did not record that why the pod pending for so long time.

So, I think we should check the pod state if pod is pending.

Also, for yarn app, it would also return the app pending reason for example waiting for am allocation.
### How was this patch tested?

Code review.

### Was this patch authored or co-authored using generative AI tooling?

No.

PR #7320, opened 2026-02-06
Update engineError conditionally based on engineError or engineId presence

turboFei:

In the previous commit [KYUUBI #7313], when a k8s application is pending, we audit the app diagnostics to record the pending reason. This information is eventually updated to the `engineError` field in the metadata store via the `updateMetadata` method.

However, when the application transitions from pending to finished successfully, if there is no error (i.e., `engineError` is `None`), the old pending reason remains in the metadata, causing confusion.

### What are the changes?

Modified `JDBCMetadataStore.updateMetadata()` to always update the `engineError` field:
- Previously: Only updated when `engineError` was `Some(error)`
- Now: Always updates, setting to `null` when `engineError` is `None`

This ensures that pending reasons are properly cleared when the application state changes.

- Added unit test `update engineError to null when it becomes None` in `JDBCMetadataStoreSuite`
- The test verifies that:
  1. Setting engineError to a pending reason works correctly
  2. Updating engineError to None clears the value in the database

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
@@ -376,9 +376,11 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
       setClauses += "engine_state = ?"
       params += metadata.engineState
     }
-    metadata.engineError.foreach { error =>
+    // Update engineError when it's defined or when engineId is defined
+    // This ensures pending reasons are cleared when app transitions to success
+    if (metadata.engineError.isDefined || Option(metadata.engineId).isDefined) {
       setClauses += "engine_error = ?"
-      params += error
+      params += metadata.engineError.orNull
     }
     if (metadata.peerInstanceClosed) {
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -376,9 +376,11 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
       setClauses += "engine_state = ?"
       params += metadata.engineState
     }
-    metadata.engineError.foreach { error =>
+    // Update engineError when it's defined or when engineId is defined
+    // This ensures pending reasons are cleared when app transitions to success
```
pan3793, line 380: > ... when app transitions to success

the code does not reflect the condition. is there a chance that the batch transfers to another negative state and clears the error message?

turboFei: K8S App:

Pending (appError=Some("reason")) -> FINISHED (appError = None), and then the pending reason is stored in the metadata store.

pan3793 (approved):

PR #7324, opened 2026-02-11
[KYUUBI #7323] Support timeout at PlanOnlyMode

### LINKED ISSUE #7323: [FEATURE] Support timeout at `PlanOnlyMode`
### Code of Conduct

### Search before asking

### Describe the feature

At some case, due to the complexity of SQL execution plans, we want to be able to fail quickly during analysis mode.

### Motivation

_No response_

### Describe the solution

Support timeout at`PlanOnlyStatement`.

### Additional context

_No response_

ruanwenjun:
close #7323

Support to cancel the `PlanOnlyStatement` when timeout.

Test by manual, the job group has been canceled, and the driver memory stop increase.
```
26/02/12 10:54:11 INFO PlanOnlyStatement: Processing hive's query[bc48bfb9-2294-4987-aa81-c5ae3f3b39b5]: RUNNING_STATE -> TIMEDOUT_STATE, time taken: 10.007 seconds
26/02/12 10:54:11 INFO PlanOnlyStatement: statementId=bc48bfb9-2294-4987-aa81-c5ae3f3b39b5, operationRunTime=0 ms, operationCpuTime=0 ms
26/02/12 10:54:11 INFO DAGScheduler: Asked to cancel job group bc48bfb9-2294-4987-aa81-c5ae3f3b39b5
26/02/12 10:54:11 INFO DAGScheduler: Asked to cancel job group bc48bfb9-2294-4987-aa81-c5ae3f3b39b5
```

NO

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
@@ -18,6 +18,7 @@
 package org.apache.kyuubi.engine.spark.operation
 
 import java.lang.{Boolean => JBoolean}
+import java.util.concurrent.{TimeoutException, TimeUnit}
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
@@ -33,11 +34,12 @@ import org.apache.kyuubi.KyuubiSQLException
 import org.apache.kyuubi.config.KyuubiConf.{LINEAGE_PARSER_PLUGIN_PROVIDER, OPERATION_PLAN_ONLY_EXCLUDES, OPERATION_PLAN_ONLY_OUT_STYLE}
 import org.apache.kyuubi.engine.spark.KyuubiSparkUtil.getSessionConf
 import org.apache.kyuubi.engine.spark.operation.PlanOnlyStatement._
-import org.apache.kyuubi.operation.{AnalyzeMode, ArrayFetchIterator, ExecutionMode, IterableFetchIterator, JsonStyle, LineageMode, OperationHandle, OptimizeMode, OptimizeWithStatsMode, ParseMode, PhysicalMode, PlainStyle, PlanOnlyMode, PlanOnlyStyle, UnknownMode, UnknownStyle}
+import org.apache.kyuubi.operation.{AnalyzeMode, ArrayFetchIterator, ExecutionMode, IterableFetchIterator, JsonStyle, LineageMode, OperationHandle, OperationState, OptimizeMode, OptimizeWithStatsMode, ParseMode, PhysicalMode, PlainStyle, PlanOnlyMode, PlanOnlyStyle, UnknownMode, UnknownStyle}
 import org.apache.kyuubi.operation.PlanOnlyMode.{notSupportedModeError, unknownModeError}
 import org.apache.kyuubi.operation.PlanOnlyStyle.{notSupportedStyleError, unknownStyleError}
 import org.apache.kyuubi.operation.log.OperationLog
 import org.apache.kyuubi.session.Session
+import org.apache.kyuubi.util.ThreadUtils
 import org.apache.kyuubi.util.reflect.DynMethods
 
 /**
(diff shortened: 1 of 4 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
```diff
@@ -76,27 +79,48 @@ class PlanOnlyStatement(
...
-            case plan => style match {
-                case PlainStyle => explainWithPlainStyle(plan)
-                case JsonStyle => explainWithJsonStyle(plan)
-                case UnknownStyle => unknownStyleError(style)
-                case other => throw notSupportedStyleError(other, "Spark SQL")
-              }
-          }
+      if (queryTimeout > 0) {
+        val timeoutExecutor =
+          ThreadUtils.newDaemonSingleThreadScheduledExecutor("query-timeout-thread", false)
```
pan3793: this creates a new thread for each operation, it's too expensive. let's follow the idea of https://github.com/apache/kyuubi/pull/7121 to create a global ThreadScheduledExecutor shared by all operations.

Q: why plan-only stmt was not covered by that? I suppose it should be, though I didn't take a deeper look yet

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/spark/SparkSqlEngineSuite.scala
```diff
@@ -149,5 +151,38 @@ class SparkSqlEngineSuite extends WithKyuubiServer with HiveJDBCTestHelper {
     }
   }
 
+  test("KYUUBI-7323: Support timeout at PlanOnlyMode") {
```
pan3793: Kyuubi uses GitHub Issues instead of JIRA, use `KYUUBI #XXXX` instead of `KYUUBI-XXXX` for issue/PR reference.

Also, we don't require creating an issue before sending a PR, GitHub Issues and PRs share the sequence number, and the link can jump to each other, so `XXXX` can either be an issue number or a PR number. What really matters here, clear description of the context of your change - especially WHY, and TEST. WHAT is also important in most cases, but for cases where the code itself is self-explanatory, WHAT is optional.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/PlanOnlyStatement.scala
ruanwenjun: Tes, this is expensive. I will add a new method 
```java
def runWithTimeoutFallback(task: Runnable,
                             onTimeout: Runnable,
                             timeoutSeconds: Long): Unit = {
    val future = timeoutScheduler.submit(task)
    try {
      future.get(timeoutSeconds, TimeUnit.SECONDS)
    } catch {
      case _: TimeoutException =>
        future.cancel(true)
        onTimeout.run()
    }
  }
```
at `OperationManager` to deal with this.

And I had also assumed that plan-only statements could respond to timeouts, but I discovered that they cannot.

ruanwenjun: In our environment, we discovered that in the current 1.9.x version, when setting a timeout in plan-only mode, although the session in the Kyuubi server has timed out and closed, the driver in the Spark engine continues to parse and execute the plan. This causes memory to remain unavailable.

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/spark/SparkSqlEngineSuite.scala
ruanwenjun: Thanks, please forgive me for mistakenly copying the template from another test method above.

And I removed this test case here, since I find that the added test method actually fails to verify whether the engine has truly cancel operation.

ruanwenjun: Has added the solution at issue.

However, I don't fully understand why interrupting the thread can stop the analysis. There might be thread state detection logic during plan analysis, but I haven't found the specific code. 
During testing, I discovered that canceling the jobgroup via cleanup didn't halt the plan analy.

PR #7385, opened 2026-04-04
[KYUUBI #7379][1/4] Data Agent Engine: module skeleton, configuration, and engine core

wangzhigang1999:

Part 1 of 4 for the Data Agent Engine ([umbrella](https://github.com/apache/kyuubi/issues/7379), [KPIP-7373](https://github.com/apache/kyuubi/discussions/7373)).

This PR adds a new `DATA_AGENT` engine type with the module skeleton fully runnable via Echo provider. It establishes the foundation for subsequent PRs (tool system, agent runtime, REST API, and Web UI).

Changes include:
- Module `pom.xml`, `build/dist` packaging
- All `kyuubi.engine.data.agent.*` configuration entries
- `DATA_AGENT` engine type, `EngineRef` branch, `DataAgentProcessBuilder` with API key redaction
- Engine core: `DataAgentEngine`, `BackendService`, `TBinaryFrontendService`, `SessionManager/Impl`, `OperationManager`
- `DataAgentOperation`, `ExecuteStatement` — async agent execution with event-to-JSON conversion
- `IncrementalFetchIterator` — thread-safe producer/consumer streaming with compaction
- Event system: `AgentEvent`, `EventType`, all lifecycle/content/tool events
- `EchoProvider` — test provider that echoes input with proper event sequence

- `IncrementalFetchIteratorSuite` — basic ops, position tracking, concurrent producer/consumer safety
- `IncrementalFetchIteratorCompactionSuite` — compaction trigger, position correctness, fetchAbsolute after compaction
- `DataAgentOperationSuite` — JDBC round-trip with Echo provider, multiple queries in same session
- `EchoProviderTest` — event sequence validation and content echo
- `EventTest` — immutability, truncation, null handling, SSE name uniqueness
- `DataAgentProcessBuilderSuite` — main class, API key redaction, memory, classpath, java options

Partially assisted by Claude Code (Claude Opus 4.6) for test deduplication, code style fixes, and PR splitting. Core design and implementation are human-authored.


externals/kyuubi-data-agent-engine/src/main/scala/org/apache/kyuubi/engine/dataagent/operation/DataAgentOperation.scala
```diff
@@ -0,0 +1,111 @@
...
+      case FETCH_FIRST =>
+        iter.fetchAbsolute(0)
+    }
+
+    val taken = iter.take(rowSetSize).map(_.toSeq)
+    val resultRowSet = new DataAgentTRowSetGenerator().toTRowSet(
+      taken.toSeq,
+      Seq(COL_STRING_TYPE),
+      getProtocolVersion)
+    resultRowSet.setStartRowOffset(iter.getPosition)
```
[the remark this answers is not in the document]
wangzhigang1999: The setStartRowOffset(iter.getPosition) after take() is the established pattern across all Kyuubi engines —
  SparkOperation, TrinoOperation, JdbcOperation, ChatOperation, and RunnableCommand all use the exact same idiom.

externals/kyuubi-data-agent-engine/src/main/scala/org/apache/kyuubi/engine/dataagent/operation/ExecuteStatement.scala
```diff
@@ -0,0 +1,174 @@
...
+              n.put("args", toolCall.toolArgs().toString)
+            }))
+          case EventType.TOOL_RESULT =>
+            val toolResult = event.asInstanceOf[ToolResult]
+            incrementalIter.append(Array(toJson { n =>
+              n.put("type", sseType)
+              n.put("id", toolResult.toolCallId())
+              n.put("name", toolResult.toolName())
+              n.put("output", toolResult.output())
+            }))
```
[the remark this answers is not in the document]
wangzhigang1999: The toolArgs field was originally intended just for display, but serializing it as proper JSON is cleaner — will fix. 

And toolResult.isError() is already there on the class, makes sense to include it. 

Will update both.

build/dist
```diff
@@ -363,6 +364,18 @@ for jar in $(ls "$DISTDIR/jars/"); do
...
 
+# Copy data-agent engines
+cp "$KYUUBI_HOME/externals/kyuubi-data-agent-engine/target/kyuubi-data-agent-engine_${SCALA_VERSION}-${VERSION}.jar" "$DISTDIR/externals/engines/data-agent/"
+cp -r "$KYUUBI_HOME"/externals/kyuubi-data-agent-engine/target/scala-$SCALA_VERSION/jars/*.jar "$DISTDIR/externals/engines/data-agent/"
+
+# Share the jars w/ server to reduce binary size
+# shellcheck disable=SC2045
+for jar in $(ls "$DISTDIR/jars/"); do
+  if [[ -f "$DISTDIR/externals/engines/data-agent/$jar" ]]; then
+    (cd $DISTDIR/externals/engines/data-agent; ln -snf "../../../jars/$jar" "$DISTDIR/externals/engines/data-agent/$jar")
```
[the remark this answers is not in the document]
wangzhigang1999: the unquoted $DISTDIR in cd is a fair point, but this follows the existing pattern used by all other engines

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/dataagent/DataAgentProcessBuilder.scala
```diff
@@ -0,0 +1,111 @@
...
+
+    buffer ++= confKeyValue(KYUUBI_SESSION_USER_KEY, proxyUser)
+    buffer ++= confKeyValue(KYUUBI_ENGINE_ID, engineRefId)
+
+    buffer ++= confKeyValues(conf.getAll)
+
+    buffer
+  }
+
+  override def toString: String = {
```
pan3793, line 99: could you please paste a demo command for the data agent engine launch?

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
```diff
@@ -113,6 +116,17 @@ private[kyuubi] class EngineRef(
 
   @VisibleForTesting
   private[kyuubi] val subdomain: String = conf.get(ENGINE_SHARE_LEVEL_SUBDOMAIN) match {
+    // DATA_AGENT engines must be isolated by datasource (JDBC URL). This takes precedence over
```
pan3793: "... must be ..." due to what?

externals/kyuubi-data-agent-engine/src/main/scala/org/apache/kyuubi/engine/dataagent/operation/ApproveToolCall.scala
```diff
@@ -0,0 +1,82 @@
...
+}
+
+object ApproveToolCall {
+  private val JSON = new ObjectMapper()
+  val APPROVE_PREFIX = "__approve:"
+  val DENY_PREFIX = "__deny:"
+
+  def isApprovalCommand(statement: String): Boolean = {
+    val trimmed = statement.trim
+    trimmed.startsWith(APPROVE_PREFIX) || trimmed.startsWith(DENY_PREFIX)
```
pan3793, line 80: can you share a little bit more about the design details here? why `isApproval*`  returns true for `DENY_PREFIX`?

externals/kyuubi-data-agent-engine/src/main/java/org/apache/kyuubi/engine/dataagent/runtime/event/EventType.java
```diff
@@ -0,0 +1,67 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.engine.dataagent.runtime.event;
+
+/**
+ * Enumerates the types of events emitted by the ReAct agent loop. Each value maps to a
```
pan3793, line 21: can you add a reference link to the ReAct agent loop?

pan3793:
@wangzhigang1999 good job! thank you for kicking off the data agent engine development work!

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/dataagent/DataAgentProcessBuilder.scala
wangzhigang1999: Here's the real launch command from a local run (API key redacted by `redactConfValues`):

```
2026-04-10 14:58:04.968 INFO KyuubiSessionManager-exec-pool: Thread-86 org.apache.kyuubi.engine.EngineRef: Launching engine:
/Users/zhigang/Library/Java/JavaVirtualMachines/azul-11.0.23/Contents/Home/bin/java \
    -Xmx1g \
    -Dfile.encoding=UTF-8 \
    -Dkyuubi.frontend.thrift.binary.bind.host=127.0.0.1 \
    -cp /Users/zhigang/IdeaProjects/kyuubi/externals/kyuubi-data-agent-engine/target/kyuubi-data-agent-engine_2.12-1.12.0-SNAPSHOT.jar:/Users/zhigang/IdeaProjects/kyuubi/externals/kyuubi-data-agent-engine/target/scala-2.12/jars/* \
    org.apache.kyuubi.engine.dataagent.DataAgentEngine \
    --conf kyuubi.session.user=anonymous \
    --conf kyuubi.engine.id=d8d3181a-ba44-4335-a788-f475330caf21 \
    --conf kyuubi.client.ipAddress=127.0.0.1 \
    --conf kyuubi.engine.appMgrInfo=eyJyZXNvdXJjZU1hbmFnZXIiOm51bGwsImt1YmVybmV0ZXNJbmZvIjp7ImNvbnRleHQiOm51bGwsIm5hbWVzcGFjZSI6bnVsbH19 \
    --conf kyuubi.engine.data.agent.java.options=-Dkyuubi.frontend.thrift.binary.bind.host=127.0.0.1 \
    --conf kyuubi.engine.data.agent.jdbc.url=jdbc:sqlite:/Users/zhigang/PycharmProjects/data-insight/benchmarks/bird/minidev/MINIDEV/dev_databases/california_schools/california_schools.sqlite \
    --conf kyuubi.engine.data.agent.llm.api.key=*********(redacted) \
    --conf kyuubi.engine.data.agent.llm.api.url=https://dashscope.aliyuncs.com/compatible-mode/v1 \
    --conf kyuubi.engine.data.agent.llm.model=qwen3.5-plus-2026-02-15 \
    --conf kyuubi.engine.data.agent.max.iterations=100 \
    --conf kyuubi.engine.data.agent.provider=openai_compatible \
    --conf kyuubi.engine.submit.time=1775804284939 \
    --conf kyuubi.engine.type=DATA_AGENT \
    --conf kyuubi.ha.addresses=30.246.181.34:2181 \
    --conf kyuubi.ha.engine.ref.id=d8d3181a-ba44-4335-a788-f475330caf21 \
    --conf kyuubi.ha.namespace=/kyuubi_1.12.0-SNAPSHOT_USER_DATA_AGENT/anonymous/ds-2f23fd339a886fe9 \
    --conf kyuubi.ha.zookeeper.auth.type=NONE \
    --conf kyuubi.server.ipAddress=0.0.0.0 \
    --conf kyuubi.session.connection.url=0.0.0.0:10099 \
    --conf kyuubi.session.real.user=anonymous
```

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
wangzhigang1999: Thanks for pushing on this — "must" really was under-motivated, let me walk through the thinking.

Starting from what's actually in this PR, the provider SPI is single-datasource by contract:

- `DataAgentProvider.load(KyuubiConf)` is called once per engine from engine-level conf.
- `ENGINE_DATA_AGENT_JDBC_URL` is a single `OptionalConfigEntry[String]`.
- `ProviderRunRequest` carries `question` / `modelName` / `approvalMode` — no datasource field.

So a provider instance is bound to exactly one JDBC URL when it's loaded, and a caller has no way to dispatch a request against a different URL. The `EngineRef.subdomain` branch is just the routing side of that same contract, keeping the two halves in sync.

As for why the SPI is shaped that way: I tried the multi-datasource version in a previous project and ran into real pain — SQL dialect divergence (tools built around one dialect don't transfer cleanly to another), connection-pool and credential isolation, schema introspection cost, and the general awkwardness of one process quietly talking to several production databases. The takeaway for me was that binding a process to a single datasource makes tool design, prompt construction, and credential handling a lot simpler, so I carried that lesson into the SPI here.

Multi-datasource isn't ruled out forever — it would just mean rethinking the SPI (per-request datasource, or a provider keyed by datasource), the config surface, and the routing together, which feels like its own discussion. Happy to revisit if a concrete use case comes up.

On the comment itself, I'll tighten it to say the *why* directly:

```scala
// A DATA_AGENT engine is bound 1:1 to a JDBC datasource because the provider SPI
// is constructed once per engine from a single kyuubi.engine.data.agent.jdbc.url,
// and ProviderRunRequest carries no datasource field — there is no way to dispatch
// a request against a different JDBC URL than the one the provider was loaded
// with. Sessions targeting different datasources must therefore route to distinct
// engines.
```

WDYT?

externals/kyuubi-data-agent-engine/src/main/scala/org/apache/kyuubi/engine/dataagent/operation/ApproveToolCall.scala
wangzhigang1999: Sure. The goal is to let a human approve or deny risky tool calls (for example a destructive SQL statement) without adding any new RPC — everything stays on the normal `executeStatement` / `fetchResults` path, so plain JDBC and Thrift clients work as-is. Here's the full round trip:

```mermaid
sequenceDiagram
    autonumber
    participant C as Client
    participant OM as DataAgentOperationManager
    participant ES as ExecuteStatement<br/>(async)
    participant P as Provider<br/>(agent + approval)
    participant AT as ApproveToolCall<br/>(sync)

    C->>OM: executeStatement("what is ...")
    OM->>ES: new ExecuteStatement
    ES->>P: run(request, onEvent)
    P-->>ES: content_delta events
    ES-->>C: fetchResults (JSON rows)

    Note over P: agent hits a risky tool call<br/>put(requestId, future)
    P-->>ES: approval_request(requestId)
    ES-->>C: fetchResults (approval row)
    Note over P: agent thread blocks on<br/>future.get(timeout)

    C->>OM: executeStatement("__approve:<requestId>")
    OM->>AT: isApprovalCommand ⇒ route to ApproveToolCall
    AT->>P: resolveApproval(requestId, true)
    P-->>P: future.complete(true)
    AT-->>C: {status:"ok", action, requestId}
    Note over P: agent thread resumes

    P-->>ES: tool_result / content_delta / finished
    ES-->>C: fetchResults (remaining rows)
```

A couple of choices worth calling out:

`ApproveToolCall` is synchronous (`shouldRunAsync = false`) because it's just a control-plane ping — there's nothing to stream, so async would only add an extra state transition for no benefit. Staying on the `executeStatement` path means we don't touch the Thrift IDL and every existing client keeps working. The `__approve:` / `__deny:` prefixes are an easy sentinel since natural-language questions don't start with `__`.

The pending-request map lives inside the provider, not in the operation manager, because the agent runtime is what actually blocks on the tool call — the future belongs next to the thing that's waiting on it. The operation layer just forwards `resolveApproval` and doesn't need to know about tool-call semantics or timeouts, which keeps the `DataAgentOperation*` classes free of agent-specific state.

Scope for this PR: what lands here is the SPI (`DataAgentProvider.resolveApproval`), the routing, and `ApproveToolCall` itself. The actual parking mechanism inside a real provider comes with the LLM-backed provider in the follow-up PR. Right now the default `resolveApproval` returns `false`, so an unknown or expired `requestId` surfaces as `{"status":"not_found"}` to the client.

externals/kyuubi-data-agent-engine/src/main/java/org/apache/kyuubi/engine/dataagent/runtime/event/EventType.java
wangzhigang1999: Good catch — the concrete `ReactAgent` class that actually drives these events lives in the LLM-backed provider, which lands in the follow-up PR (2/4 of the series). I'll add the `{@link ReactAgent}` / `@see` from `EventType` in that PR, so the enum and the loop implementation live next to each other and stay in sync. Keeping this PR's Javadoc untouched for now to avoid a dangling reference.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala
pan3793: let's update it as your proposed comment for now, and then I will merge this PR to unblock your subsequent PRs, we can revise this later if we find any issues.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/dataagent/DataAgentProcessBuilder.scala
pan3793: BTW, it seems to have a bug in command redaction in the current versions of Kyuubi, reported by https://github.com/apache/kyuubi/issues/7387

externals/kyuubi-data-agent-engine/src/main/scala/org/apache/kyuubi/engine/dataagent/operation/ApproveToolCall.scala
wangzhigang1999: > can you share a little bit more about the design details here? why `isApproval*` returns true for `DENY_PREFIX`?

isApprovalCommand is a routing predicate — "is this statement part of the approval control channel?", not "did the user approve?".  __approve: and __deny: are two outcomes on the same channel, so both route to ApproveToolCall,  which then decodes the prefix into a boolean and passes it to resolveApproval(requestId, approved) to unblock the parked agent thread.

pan3793 (approved):

pan3793:
thanks, merged to master

PR #7386, opened 2026-04-06
Remove axios dependency and replace with native fetch in web-ui

aajisaka:

Replace axios with the native Fetch API in request.ts, removing the axios dependency from package.json. Reducing the dependent libraries to reduce the possibility of supply chain attack and to reduce future maintenance cost. Enforce Node.js >=18 via the engines field and update README accordingly.

Covered by the existing unit tests.

Generated-by: Claude Code (Claude Opus 4.6)


aajisaka:
Thank you @pan3793. Addressed the review comments.

yaooqinn (approved):

pan3793:
verified locally, thanks, merged to master

aajisaka:
Thanks @yaooqinn and @pan3793

PR #7390, opened 2026-04-07
[KYUUBI #7389] Fix wrong parameter binding order in JDBCMetadataStore#transformMetadataState

wangzhigang1999:
Closes #7389

The SQL placeholders expect (targetState, identifier, fromState) but the code passed (fromState, identifier, targetState), causing the UPDATE to set state back to fromState and match on targetState instead. This made cancelUnscheduledBatch silently fail.

In `JDBCMetadataStore#transformMetadataState`, the SQL is:
```sql
UPDATE metadata SET state = ? WHERE identifier = ? AND state = ?
```
The placeholders expect `(targetState, identifier, fromState)`, but the code passes `(fromState, identifier, targetState)`:
```scala
withUpdateCount(connection, query, fromState, identifier, targetState)
```

This causes `cancelUnscheduledBatch("INITIALIZED" -> "CANCELED")` to execute:
```sql
UPDATE metadata SET state = 'INITIALIZED' WHERE identifier = ? AND state = 'CANCELED'
```
Which matches no rows and silently returns `false`. The bug has existed since v1.8.0 ([KYUUBI #4790]).

Added a unit test `transformMetadataState should transition state correctly` in `JDBCMetadataStoreSuite` that:
1. Inserts a metadata record with `state = "INITIALIZED"`
2. Calls `transformMetadataState(id, "INITIALIZED", "CANCELED")`
3. Asserts the method returns `true`
4. Asserts the persisted state is `"CANCELED"`

No.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
@@ -235,7 +235,7 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
       targetState: String): Boolean = {
     val query = s"UPDATE $METADATA_TABLE SET state = ? WHERE identifier = ? AND state = ?"
     JdbcUtils.withConnection { connection =>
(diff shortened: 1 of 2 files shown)


pan3793:
a stupid mistake ... thanks for fixing

pan3793 (approved):

pan3793:

PR #7391, opened 2026-04-08
Improve session REST security 

Yutong-1424:
1. Sensitive settings in session APIs
Session APIs were returning full session configs as-is. Some keys (passwords, tokens, etc.) should not be shown to clients or in traces. This patch applies the same redaction pattern the server already uses elsewhere (SERVER_SECRET_REDACTION_PATTERN + Utils.redact) so list/detail responses don’t expose those values in plain text.

2. Over-broad session listing
GET /api/v1/sessions returned every session on the server. With auth turned on, it should only return sessions for the logged-in user, not everyone else’s

Added a REST test that opens a session with a sensitive Spark key, calls GET /api/v1/sessions, and checks the value is redacted.
Updated the session control CLI test so the session is opened with the same user the test client uses for auth, so the “list sessions” output matches real usage.
Verified tests run locally

No


wangzhigang1999:
One concern on the user-based filtering for GET /api/v1/sessions — we already have GET /api/v1/admin/sessions in AdminResource with user filtering support, and BatchesResource doesn't enforce user filtering either, so this would be inconsistent. Also without authentication enabled, anyone can claim any identity, so the filtering provides no real security.

pan3793:
can we make the returned configs content configurable? e.g., `ORIGINAL`, `REDACTED`, `NONE`

Yutong-1424:
> can we make the returned configs content configurable? e.g., `ORIGINAL`, `REDACTED`, `NONE`

Are you suggesting the conf display behavior should depend on the caller's role? For example, admins see the full config while regular users see redacted or nothing? or uses a single global mode kyuubi.session.conf.display.mode to set defult is REDACTED

pan3793:
@Yutong-1424, does the global server-side config meet your requirements? if so, this is fine. if you want fine-grained control capabilities, we can allow client-side config override and check it on the server-side, the check rules might be flexible, and we can discuss details during implementation.

Yutong-1424:
I implemented `kyuubi.session.conf.display.mode` with REDACTED, ORIGINAL, and NONE as discussed, and covered by SessionsResourceSuite tests. Happy to adjust if you had something else in mind. @pan3793

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -3336,6 +3336,21 @@ object KyuubiConf {
       .regexConf
       .createOptional
 
+  val SESSION_CONF_DISPLAY_MODE: ConfigEntry[String] =
+    buildConf("kyuubi.session.conf.display.mode")
```
pan3793: how about "kyuubi.server.conf.retrieveMode"? we might extend the effective scope to batch, operation, server conf display, etc.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/ApiUtils.scala
```diff
@@ -22,13 +22,28 @@ import scala.collection.JavaConverters._
...
 import org.apache.kyuubi.session.KyuubiSession
 
 object ApiUtils extends Logging {
+
+  private def buildConf(
+      rawConf: Map[String, String],
+      session: KyuubiSession): java.util.Map[String, String] = {
+    session.sessionManager.getConf.get(SESSION_CONF_DISPLAY_MODE) match {
+      case "NONE" => Map.empty[String, String].asJava
+      case "ORIGINAL" => rawConf.asJava
```
pan3793: use enum

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/v1/SessionsResource.scala
```diff
@@ -52,11 +52,15 @@ private[v1] class SessionsResource extends ApiRequestContext with Logging {
...
-    description = "get the list of all live sessions")
+    description = "get the list of live sessions for the current user")
   @GET
   def sessions(): Seq[SessionData] = {
-    sessionManager.allSessions()
-      .map(session => ApiUtils.sessionData(session.asInstanceOf[KyuubiSession])).toSeq
+    val userName = fe.getSessionUser(Map.empty[String, String])
+    sessionManager
+      .allSessions()
+      .filter(session => session.user == userName)
```
pan3793: it's reasonable for security purposes, but it's a breaking change.

I would suggest making this change independent, with an internal legacy config switch (something like `kyuubi.frontend.rest.legacy.v1.sessionsReturnAllUsers`), and adding an item in the migration guide.

Yutong-1424: Thanks for your comments, all comments addressed:
- Config renamed to `kyuubi.server.conf.retrieveMode`
- Using enum
- Added legacy session list config with migration guide
- Regenerated settings.md

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -3336,6 +3336,32 @@ object KyuubiConf {
...
+        "Supported values: " +
+        "<ul>" +
+        "<li>REDACTED: Mask values that match kyuubi.server.redaction.regex (default).</li>" +
+        "<li>ORIGINAL: Return the raw config values as-is.</li>" +
+        "<li>NONE: Omit the conf map from responses entirely.</li>" +
+        "</ul>")
+      .version("1.12.0")
+      .stringConf
+      .checkValues(Set("REDACTED", "ORIGINAL", "NONE"))
+      .createWithDefault("REDACTED")
```
pan3793, line 3352: this is a user-facing change, also needs to be mentioned in migration guide

pan3793 (approved):
LGTM, one nit

pan3793:
Wrong click the merge button, but let's wait for CI results ...

Yutong-1424:
> Wrong click the merge button, but let's wait for CI results ...

we should be good? the failed one likely a connection errors...

pan3793:
yeah, it's good, thanks for contribution

PR #7393, opened 2026-04-09
[Bug] Fix mapping for DECIMAL type to BigDecimal in JdbcColumn

kezhuoquan:

In JDBC/Java, SQL DECIMAL/NUMERIC must map to `java.math.BigDecimal` to preserve precision and scale. The current implementation maps DECIMAL to `java.math.BigInteger`, which drops fractional parts and causes type mismatches.

Existing tests passing indicates no regressions.

Generated-by: Trae AI (GPT-5-high)

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/JdbcColumn.java
@@ -19,7 +19,7 @@ package org.apache.kyuubi.jdbc.hive;
 
 import static java.sql.Types.*;
 
-import java.math.BigInteger;
+import java.math.BigDecimal;
 import java.sql.Date;
 import java.sql.SQLException;
 import java.sql.Timestamp;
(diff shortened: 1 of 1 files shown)


pan3793:
@kezhuoquan, this seems to be a bug inherited from upstream Hive

https://github.com/apache/hive/blob/branch-4.2/jdbc/src/java/org/apache/hive/jdbc/JdbcColumn.java#L100

Side notes: review your AI-generated PR description before sending publicly, such a small change is not worth such a long article to describe.

kezhuoquan:
> @kezhuoquan, this seems to be a bug inherited from upstream Hive
> 
> https://github.com/apache/hive/blob/branch-4.2/jdbc/src/java/org/apache/hive/jdbc/JdbcColumn.java#L100
> 
> Side notes: review your AI-generated PR description before sending publicly, such a small change is not worth such a long article to describe.

So I can only fix this bug in Hive?

pan3793:
@kezhuoquan, we usually follow the upstream-first strategy.

I suggest you report and fix this bug in the Hive project first, then port the fix here.

kezhuoquan:
> @kezhuoquan, we usually follow the upstream-first strategy.
> 
> I suggest you report and fix this bug in the Hive project first, then port the fix here.

ok

Issue #7407, opened 2026-04-16
[Bug] Continuous Heap Usage Increase


oh0873:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Kyuubi Heap usage is keep increasing eventually causing a Out Of Memory Kill.

Based on Heapdump ST Token Cache never gets freed up causing continuous increase.
```Total of 14971243 instances occupying 898274580 bytes.```

Here's a graph showing kyuubi heap usage over few days. 
<img width="1337" height="634" alt="Image" src="https://github.com/user-attachments/assets/162a1b8b-2805-444f-8f82-98ea6951a56a" />

ST Token is used during LDAP Authentication. Any application that uses LDAP is vulnerable to this Memory Leak behavior.

### Affects Version(s)

1.10.3

### Kyuubi Server Log Output

```logtalk
Total of 14971243 instances occupying 898274580 bytes.
```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #7408, opened 2026-04-16
[KYUUBI #7407] STGroup free to avoid OOM Kill

oh0873:

When LDAP Authentication is used ST Token is created and saved in the cache, but it is never freed up.
This is causing continuous increase in heap usage, eventually causing out-of-memory for kyuubi server pods.

This changes is added to clear ST Tokens. Also ST Group is added to avoid any race condition during clean up.

This patch was tested in our customer environment. We observed there's no more continuous heap increase after the fix.

Cursor auto-complete feature was used.

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
@@ -128,7 +128,7 @@ object Query {
     def build: Query = {
       validate()
       val filter: String = createFilter
-      // Unload template cache after render to avoid CompliedST/STToken retntion
+      // Unload template cache after render to avoid CompiledST/STToken retention
       Option(filterTemplate.groupThatCreatedThisInstance).foreach(_.unload())
       updateControls()
       new Query(filter, controls)
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
@@ -25,6 +25,9 @@ import org.stringtemplate.v4.{ST, STGroup}
 /**
  * The object that encompasses all components of a Directory Service search query.
  *
+ * The caller must and can only call each [[filter]] and [[build]] once,
+ * otherwise [[ST]] internal cache may leak and cause heap OOM.
+ *
  * @see [[LdapSearch]]
  */
 object Query {
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
@@ -40,6 +40,7 @@ object Query {
    * A builder of the [[Query]].
    */
   final class QueryBuilder {
+
     /** The [[STGroup]] used only for this builder's filter template; unloaded in [[build]]. */
     private var filterTemplateGroup: Option[STGroup] = None
     private var filterTemplate: ST = _
@@ -116,7 +117,7 @@ object Query {
       require(filterTemplate != null, "filter is required for LDAP search query")
     }
 
-    private def createFilter: String = filterTemplate.render
+    private def createFilter(): String = filterTemplate.render()
 
     private def updateControls(): Unit = {
       if (!returningAttributes.isEmpty) controls.setReturningAttributes(
@@ -130,7 +131,7 @@ object Query {
      */
     def build: Query = {
       validate()
-      val filter: String = createFilter
+      val filter: String = createFilter()
       // Unload template cache after render to avoid CompiledST/STToken retention
       filterTemplateGroup.foreach(_.unload())
       updateControls()
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
@@ -40,6 +40,8 @@ object Query {
    * A builder of the [[Query]].
    */
   final class QueryBuilder {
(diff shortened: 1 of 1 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
```diff
@@ -127,6 +128,8 @@ object Query {
     def build: Query = {
       validate()
       val filter: String = createFilter
+      // Unload template cache after render to avoid CompliedST/STToken retntion
+      Option(filterTemplate.groupThatCreatedThisInstance).foreach(_.unload())
```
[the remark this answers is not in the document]
oh0873: Added filterTemplateGroup to handle this

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
```diff
@@ -56,7 +58,9 @@ object Query {
      * @return the current instance of the builder
      */
     def filter(filterTemplate: String): Query.QueryBuilder = {
-      this.filterTemplate = new ST(filterTemplate)
+      val group = new STGroup()
+      this.filterTemplateGroup = Some(group)
+      this.filterTemplate = new ST(group, filterTemplate)
```
pan3793, line 67: does cache only happen in `filterTemplate.render`?

what will happen if we call `def filter` multi times?

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
```diff
@@ -127,6 +131,8 @@ object Query {
     def build: Query = {
       validate()
       val filter: String = createFilter
```
pan3793: could you update the `def createFilter` to add a `()` to the signature? because it actually has a side-effect.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
pan3793: add `()` for impure function
```suggestion
    private def createFilter(): String = filterTemplate.render()
```

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
oh0873: Done

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
oh0873: Done

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
oh0873: Cache is created in `def filter`, but it won't be needed after `createFilter` that's why I added unload right after `createFilter`. I'm not sure if this is the best place to unload. 

At least in our use case it seems to unload most if not all STToken used by LDAP auth process.

pan3793: > Cache is created in `def filter`

so cache still leaks if we call `def filter` multi times?

oh0873: Yes. IF `def filter` is called but `build` never called those STToken will stay in heap forever.

pan3793: I mean

```
builder
  .filter("exp1")
  .filter("exp2") // this replaces exp1
  .build()
```

will this case cause a leak?

oh0873: Yes, I think this would cause a leak.

pan3793: okay, as this is only used internally, we don't have such anti-patterns yet, let's somehow document those dangerous behaviors

BTW, this part of the code is ported from Apache Hive, it should have the same issue

https://github.com/apache/hive/blob/branch-4.2/service/src/java/org/apache/hive/service/auth/ldap/Query.java#L90

pan3793:
just out of curiosity, how much heap memory do you configure for kyuubi server when observing this issue? we have some customers who have used LDAP for many years, but have not received such reports.

oh0873:
> just out of curiosity, how much heap memory do you configure for kyuubi server when observing this issue? we have some customers who have used LDAP for many years, but have not received such reports.

We set 2G for kyuubi heap memory. Kyuubi server pod would get OOM kill in 2-7 days.

pan3793:
ah, it's relatively rare to set such a small heap for a big data component 🤣

oh0873:
Oh okay. After this STToken fixes, we generally see less than 25% of our heap getting used. Should we need to increase our heap size? (2G is in our testing environment)

kyuubi-common/src/main/scala/org/apache/kyuubi/service/authentication/ldap/Query.scala
pan3793, line 31:
```suggestion
 * The caller must and can only call each [[filter]] and [[build]] once,
 * otherwise [[ST]] internal cache may leak and cause heap OOM.
 *
 * @see [[LdapSearch]]
```

pan3793:
> Should we need to increase our heap size? (2G is in our testing environment)

2g is fine if it works well on your workload. it's intended to shift heavy workloads from server to engine so that server should be light and stable.

pan3793 (approved):

pan3793:

PR #7409, opened 2026-04-16
Added try-catch block in REST batch v2 openBatchSession

oh0873:

In Kyuubi Batch v2 version, Batch Service creates `batchExecutor` to submit a batch job. However, when `openBatchSession` fails, the executor dies and never recovers.

This results in a jobs stuck at PENDING state and no executor to submit any jobs at INITIALIZED state.

This PR added error handling logic to handle openBatchSession.

This also fixes `withUpdateCount` call changes fromState to targetState.

Tested in our environment, openBatchSession exceptions like connections per user or DB connection error no longer leaves jobs stuck at PENDING state. Instead all those jobs are labeled as ERROR.

Test case and bug finding were assisted with Cursor agent.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
@@ -149,7 +149,7 @@ class KyuubiBatchService(
                     ex)
               }
             }
-            // sleep 1 second to avoid excessive retries during transient network/DB failures 
+            // sleep 1 second to avoid excessive retries during transient network/DB failures
             Thread.sleep(1000)
         }
       }
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
@@ -60,57 +60,79 @@ class KyuubiBatchService(
   }
 
   override def start(): Unit = {
+    val UNINITIALIZED_BATCH_ID = "UNINITIALIZED_BATCH_ID"
     assert(running.compareAndSet(false, true))
     val submitTask: Runnable = () => {
       restFrontend.waitForServerStarted()
       while (running.get) {
-        metadataManager.pickBatchForSubmitting(kyuubiInstance) match {
-          case None => Thread.sleep(1000)
-          case Some(metadata) =>
-            val batchId = metadata.identifier
-            info(s"$batchId is picked for submission.")
-            val batchSession = sessionManager.createBatchSession(
-              metadata.username,
-              "anonymous",
-              metadata.ipAddress,
-              metadata.requestConf,
-              metadata.engineType,
-              Option(metadata.requestName),
-              metadata.resource,
-              metadata.className,
-              metadata.requestArgs,
-              Some(metadata),
-              fromRecovery = false)
-            sessionManager.openBatchSession(batchSession)
-            var submitted = false
-            while (!submitted) { // block until batch job submitted
-              submitted = metadataManager.getBatchSessionMetadata(batchId) match {
-                case Some(metadata) if OperationState.isTerminal(metadata.opState) =>
-                  true
-                case Some(metadata) if metadata.opState == OperationState.RUNNING =>
-                  metadata.appState match {
-                    // app that is not submitted to resource manager
-                    case None | Some(ApplicationState.NOT_FOUND) => false
-                    // app that is pending in resource manager while the local startup
-                    // process is alive. For example, in Spark YARN cluster mode, if set
-                    // spark.yarn.submit.waitAppCompletion=false, the local spark-submit
-                    // process exits immediately once Application goes ACCEPTED status,
-                    // even no resource could be allocated for the AM container.
-                    case Some(ApplicationState.PENDING) if batchSession.startupProcessAlive =>
-                      false
-                    // not sure, added for safe
-                    case Some(ApplicationState.UNKNOWN) => false
-                    case _ => true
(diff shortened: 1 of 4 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
```diff
@@ -60,57 +60,77 @@ class KyuubiBatchService(
...
+                metadataManager.failScheduledBatch(batchId)
+              } catch {
+                case ex: Exception =>
+                  error(s"Unable to modify metadata for $batchId to ERROR", ex)
               }
-              if (!submitted) Thread.sleep(1000)
             }
-            info(s"$batchId is submitted or finished.")
+            Thread.sleep(1000)
         }
```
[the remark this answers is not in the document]
oh0873: Added InteruptException handling

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
pan3793, line 238: I remember this has been fixed in master branch, did you create this PR from a stale commit?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
```diff
@@ -60,57 +60,90 @@ class KyuubiBatchService(
...
+          case e: Exception =>
+            if (batchId == UNINITIALIZED_BATCH_ID) {
+              error(s"Error picking batch for submission", e)
+            } else {
+              error(s"Error opening batch session for $batchId", e)
+              try {
+                metadataManager.failScheduledBatch(batchId)
+              } catch {
+                case ex: Exception =>
+                  error(s"Unable to modify metadata for $batchId to ERROR", ex)
```
pan3793: > Tested in our environment, openBatchSession exceptions like connections per user or DB connection error no longer leaves jobs stuck at PENDING state. Instead all those jobs are labeled as ERROR.

if there are transient network issues, and code reaches here, will job still be stuck at the PENDING state?

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
```diff
@@ -60,57 +60,90 @@ class KyuubiBatchService(
...
               }
-              if (!submitted) Thread.sleep(1000)
             }
-            info(s"$batchId is submitted or finished.")
+            Thread.sleep(1000)
```
pan3793, line 153: what's the intention of this sleep

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
oh0873: oh.. I made commit off of master. I can check again.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
oh0873: I can remove the sleep. We mostly observe the issue when there's server-wide network issue. Slowing down the submit process would reduce error rate. However, I do see this is not the only failure.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
oh0873: Yes they would still stuck at PENDING state. However, since the thread doesn't die, submitter will continue to pick up another job.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
oh0873: Oh I see what happen. I made a commit from 1.10.3 branch. Then cherry pick that commit from master. to make this branch. I see that this matches the master.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
pan3793: so it requires the admin to reset the state manually? if so, let's say that in the error message

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
pan3793: yeah, I guess this is mostly caused by transient network issues, then sleep makes sense, let's keep it and add a comment to explain the intention.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
oh0873: Thank you, error message is modified.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
oh0873: Thank you, Comment and sleep is added.

pan3793:
Side notes: link your email used to commit (e.g., https://github.com/apache/kyuubi/commit/80ff8b35232220af9a833456ed70a94721bfa883) to your GitHub account, then you don't need to wait for approval to run CI on pushing new changes

PR #7416, opened 2026-04-22
[KYUUBI #7415] Add a configuration to include additional Spark keys as path validation targets

### LINKED ISSUE #7415: [FEATURE] Enable adding Spark parameters that Kyuubi validate without rebuilding, similar to Livy
### Code of Conduct

### Search before asking

### Describe the feature

We can add Spark parameters that Kyuubi validate through configuration, without rebuilding.

### Motivation

Currently, the list of Spark parameters whose values Kyuubi validates is hard-coded in https://github.com/apache/kyuubi/blob/ed5d8b6af46d0ad04b533a1dc30230b0041cce8c/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/spark/SparkProcessBuilder.scala#L421-L441, so expanding the list requires rebuilding Kyuubi.

Apache Livy has similar features.

- https://github.com/apache/incubator-livy/blob/fd21e1dbd03636ec152780bb7caf1029eb2d6df4/server/src/main/scala/org/apache/livy/LivyConf.scala#L377-L386

### Describe the solution

_No response_

### Additional context

_No response_

LamiumAmplexicaule:

To add Spark configuration keys for path validation without rebuilding.
Close: https://github.com/apache/kyuubi/issues/7415

Unit tests.

```shell
$ ./build/mvn clean install -Dtest=none -DwildcardSuites=org.apache.kyuubi.engine.KyuubiApplicationManagerSuite
```

No.


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1873,6 +1873,16 @@ object KyuubiConf {
       .toSet()
       .createWithDefault(Set.empty)
 
+  val SPARK_FILE_CONFIG_LIST: ConfigEntry[Set[String]] =
+    buildConf("kyuubi.spark.file.config.list")
```
aajisaka: Maybe `kyuubi.session.engine.spark.file.config.list`?

There're several parameters start with `kyuubi.session.engine.spark.` but there's no parameter starts with `kyuubi.spark.` Also, the new parameter is related to `kyuubi.session.local.dir.allow.list`, and therefore I think it should be categorized as session parameter.

LamiumAmplexicaule: Thank you for your review.

My reasoning is that including `.engine.` makes it sound like an engine-side setting. 
However, this feature simply adds Spark keys to the list that Kyuubi checks before submitting an engine.

Given that, I think `kyuubi.session.spark.file.config.list` would be a better name.

aajisaka: > `kyuubi.session.spark.file.config.list` would be a better name.

Thanks. Make sense to me.

LamiumAmplexicaule: I renamed the config to `kyuubi.session.spark.file.config.list` in https://github.com/apache/kyuubi/pull/7416/changes/61fc167de36c0736995a9351584de16533f895a5.

pan3793:
is this config allow the user to override with the session conf? or is it a server-side static config? - unfortunately, our codebase does not use a canonical name to distinguish `serverConf`, `sessionConf`, `opConf`, etc., and this has been causing confusion and bugs

LamiumAmplexicaule:
Thank you for your question.
> is this config allow the user to override with the session conf? or is it a server-side static config?

Since it must not be overwritten, I think it should be a static server-side config.
Also, the conf used in this PR is the `KyuubiSessionManager` config, so I consider it a server-side static config.

https://github.com/apache/kyuubi/blob/8bed9de894f1b4ae0f510cc58daf7199555f6e98/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala#L99-L104
https://github.com/apache/kyuubi/blob/61fc167de36c0736995a9351584de16533f895a5/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala#L219-L228

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1873,6 +1873,16 @@ object KyuubiConf {
       .toSet()
       .createWithDefault(Set.empty)
 
+  val SESSION_SPARK_FILE_CONFIG_LIST: ConfigEntry[Set[String]] =
+    buildConf("kyuubi.session.spark.file.config.list")
```
pan3793: let's put it under `kyuubi.server.` namespace as it's a server-side static config.
```suggestion
    buildConf("kyuubi.server.spark.file.config.list")
```

pan3793 (approved):

pan3793:
change the config namespace, otherwise lgtm

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
LamiumAmplexicaule: I renamed the config to `kyuubi.server.spark.file.config.list` in https://github.com/apache/kyuubi/pull/7416/changes/2f8450c18442dced6bf7667cfb1779d908ad6305.

aajisaka (approved):

aajisaka:
Merged the PR into master. Thank you @LamiumAmplexicaule @pan3793 !

PR #7418, opened 2026-04-24
Rename kyuubi.engine.security.* parameters to kyuubi.internal.security.* and document them

aajisaka:

`kyuubi.engine.security.*` parameters are misleading because they affect not only Kyuubi engines but also Kyuubi Servers' internal communication. Rename these parameters to `kyuubi.internal.security.*` with a fallback and document them. Also add a new document to explain how to configure secure internal communication.

Covered by CI

No


pan3793:
I remember this config name causes some ambiguity, and some features are enforced to only be available when enabling this config ... we indeed need to document that, I will likely loop this part of code this weekend or early next week

aajisaka:
Thanks. Documented that the features are available only when kyuubi.engine.security.enabled is set to true. Also enhanced the doc for kyuubi.engine.security.secret.provider.

pan3793:
@aajisaka I recall the background. The security feature is requested and implemented by @turboFei from eBay. 

At the beginning, Kyuubi only supports interactive mode, there is no communication between Kyuubi server instances, so security applies to `server <=> engine` communication, and the config is named as `kyuubi.engine.security.enabled`.

Later, Kyuubi added support for batch mode, and the HA implementation contains the HTTP request between Kyuubi server instances, it reuses `kyuubi.engine.security.enabled` config, and expands the scope to all internal communications - both `server <=> server` and `server <=> engine`.

What's confusing?

1. the config name, obviously, `kyuubi.engine.security.*` is misleading, we might change it to `kyuubi.internal.security.*` and add a fallback to avoid breaking changes.
2. `kyuubi.engine.security.enabled` must be set to `true` to enable batch HA, this was strongly insisted on by the feature author @turboFei, for security reasons. But I think we should remove such a requirement. (as a workaround, I introduced `SimpleEngineSecuritySecretProviderImpl` to allow admin to configure pre-share secret key on `kyuubi-defaults.conf`, to avoid introducing a zk dependency to store the secret key). This should be documented and simplified, otherwise, users are not able to enable the batch HA without looking at the code in depth.

What should we do?

1. rename the config to `kyuubi.internal.security.*` with a fallback, and make those configs publicly.
2. make a decision on whether we can remove the internal security-enabled requirement for batch HA.
3. write a document to briefly explain how internal security works, how to enable/configure it, and how to extend it (e.g., how to implement a custom secret provider)

aajisaka:
Thank you @pan3793 for the detailed context.

Agreed to rename the config. Renamed them in the latest commit.

> make a decision on whether we can remove the internal security-enabled requirement for batch HA.

Before removing the requirement, we need to ask admin to secretly store the pre-share secret key. It must not be visible to users. Probably admin are required to configure `kyuubi.server.redaction.regex` as well. Is it sufficient, or do you have any idea? My preference is always redact the secret key regardless of the regex config.

> write a document to briefly explain how internal security works, how to enable/configure it, and how to extend it (e.g., how to implement a custom secret provider)

Totally agreed. I'll write a doc later.

pan3793:
@aajisaka, so let's do 1 and 3 first, and discuss 2 later.

aajisaka:
Renamed the config and wrote a doc for internal security.

docs/security/internal_secure_access.md
```diff
@@ -0,0 +1,58 @@
...
+Kyuubi supports secure communication channels for all internal interactions, both:
+
+* Between the Kyuubi server and Kyuubi engines.
+* Between Kyuubi server instances (High availability mode).
+
+If `kyuubi.internal.security.enabled` is set to `true`, all internal communication are encrypted.
+
+## Managing encryption secrets
+
+Encryption secrets are managed by a subclass of `org.apache.kyuubi.service.authentication.EngineSecuritySecretProvider`,
```
pan3793, line 31: now the trait name becomes strange, maybe we should provide a new Java-based interface in the `kyuubi-server-plugin` module and deprecate this one ...

docs/security/internal_secure_access.md
```diff
@@ -0,0 +1,58 @@
...
+which can be configured via `kyuubi.internal.security.secret.provider`.
+
+Kyuubi provides the following built-in implementations:
+
+* simple: Use the secret configured by `kyuubi.internal.security.secret.provider.simple.secret`.
```
pan3793: "... the secret ..." => "... the pre-shared secret ..."

pan3793 (approved):
sorry for being late, overall lgtm with small nits. 

For `EngineSecuritySecretProvider`, it's up to you to replace it or not.

docs/security/internal_secure_access.md
```diff
@@ -0,0 +1,58 @@
...
+- See the License for the specific language governing permissions and
+- limitations under the License.
+-->
+
+# Internal Secure Access
+
+Kyuubi supports secure communication channels for all internal interactions, both:
+
+* Between the Kyuubi server and Kyuubi engines.
+* Between Kyuubi server instances (High availability mode).
```
pan3793:
```suggestion
* Between Kyuubi server instances (REST API high availability mode).
```

docs/security/internal_secure_access.md
```diff
@@ -0,0 +1,58 @@
...
+
+If `kyuubi.internal.security.enabled` is set to `true`, all internal communication are encrypted.
```
pan3793, line 25: Also need to mention: currently, it's enforced to set `kyuubi.internal.security.enabled` to `true` if you want to use REST API high availability mode.

turboFei (approved):

aajisaka:
@pan3793 @turboFei Thank you for your review! I'll address the review comments before merging

docs/security/internal_secure_access.md
aajisaka: Thank you! Fixed by https://github.com/apache/kyuubi/pull/7418/commits/861cea95fba513c789068f13c87e9a7ccdf7be80

docs/security/internal_secure_access.md
aajisaka: Fixed by https://github.com/apache/kyuubi/pull/7418/commits/861cea95fba513c789068f13c87e9a7ccdf7be80

docs/security/internal_secure_access.md
aajisaka: Fixed by https://github.com/apache/kyuubi/pull/7418/commits/861cea95fba513c789068f13c87e9a7ccdf7be80

docs/security/internal_secure_access.md
aajisaka: Sounds good. Let's open a separate issue to provide a new one.

pan3793 (approved):

pan3793:
thanks for updating, lgtm. don't forget to modify the PR title to reflect the eventual change before merging.

docs/security/internal_secure_access.md
aajisaka: Created https://github.com/apache/kyuubi/issues/7448

PR #7426, opened 2026-04-28
Try Catch logic for openBatchSession

oh0873:

In Kyuubi Batch v2 version, Batch Service creates `batchExecutor` to submit a batch job. However, when openBatchSession fails, the executor dies and never recovers.

This results in a jobs stuck at PENDING state and no executor to submit any jobs at INITIALIZED state.

This PR added error handling logic to handle openBatchSession.

This also fixes withUpdateCount call changes fromState to targetState.

Tested in our environment, openBatchSession exceptions like connections per user or DB connection error no longer leaves jobs stuck at PENDING state. Instead all those jobs are labeled as ERROR.

Test case and bug finding were assisted with Cursor agent.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
@@ -60,57 +60,97 @@ class KyuubiBatchService(
   }
 
   override def start(): Unit = {
+    val UNINITIALIZED_BATCH_ID = "UNINITIALIZED_BATCH_ID"
     assert(running.compareAndSet(false, true))
     val submitTask: Runnable = () => {
       restFrontend.waitForServerStarted()
       while (running.get) {
-        metadataManager.pickBatchForSubmitting(kyuubiInstance) match {
-          case None => Thread.sleep(1000)
-          case Some(metadata) =>
-            val batchId = metadata.identifier
-            info(s"$batchId is picked for submission.")
-            val batchSession = sessionManager.createBatchSession(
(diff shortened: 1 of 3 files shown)


oh0873:
@pan3793 Can I get a review on this PR?

pan3793 (approved):

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
```diff
@@ -60,57 +60,97 @@ class KyuubiBatchService(
   }
 
   override def start(): Unit = {
+    val UNINITIALIZED_BATCH_ID = "UNINITIALIZED_BATCH_ID"
```
pan3793: nit: create an `object KyuubiBatchService`, and move const to there

kyuubi-server/src/main/scala/org/apache/kyuubi/server/KyuubiBatchService.scala
```diff
@@ -60,57 +60,97 @@ class KyuubiBatchService(
...
+              error(s"Interrupted while opening batch session for $batchId", e)
+              try {
+                metadataManager.failScheduledBatch(batchId)
+              } catch {
+                case ex: Exception =>
+                  error(
+                    s"Unable to modify metadata for $batchId to ERROR; " +
+                      "an administrator may need to reset the batch state manually.",
+                    ex)
+              }
```
pan3793: consider extracting this part to a method as it repeats twice

pan3793:
thanks, fix makes sense, just two style comments

PR #7428, opened 2026-05-02
Support fetchOrientation parameter in OperationRestApi.getOperationLog 

ychris78:

  The current getOperationLog method in OperationRestApi does not support the fetchOrientation parameter, which limits the client's ability to control the fetch direction (e.g., FETCH_NEXT, FETCH_FIRST) when retrieving operation logs. This change adds an overloaded method that accepts
  fetchOrientation, while keeping the original method signature for backward compatibility.

  - Existing tests should pass as the original method signature is preserved via method overloading.
  - Manual verification by calling the new API with and without fetchOrientation.

No


pan3793 (approved):

pan3793:
thanks, merged to master

PR #7437, opened 2026-05-08
[AUTHZ] Upgrade Apache Ranger to 2.9.0 and migrate Spark authz plugin to Ranger PDP mode

aajisaka:

Upgrade Apache Ranger from 2.6.0 to 2.9.0 in the Spark AuthZ plugin to fix
GHSA-c87w-642h-m97h, and migrate the plugin to the new Ranger authorization API
introduced in Ranger 2.9.0 (`ranger-authz-api` / `authz-remote` / `authz-embedded`):

- `SparkRangerAdminPlugin` delegates to `org.apache.ranger.authz.api.RangerAuthorizer`.
  The default is the Ranger PDP mode (`RangerRemoteAuthorizer`, a thin client sending
  REST requests to the Ranger PDP server); the embedded mode (pull policies from the
  Ranger admin and evaluate locally, as the previous plugin did) is kept as an option
  via `ranger.authorizer.impl.class`. The plugin now requires Apache Ranger 2.9.0 or
  above; the migration guide and the docs are updated.
- Security hardening found while reviewing the migration:
  - escape RRN metacharacters (backslash, slash) in resource names, and deny blank
    resource names instead of matching wildcard policies with them;
  - authorize a uri as alternative resource variants (exact path and trailing-slash
    path): the RRN request evaluates sub-resources as conjunctive requirements, which
    denied access to users granted exact-path url policies;
  - deny accesses when the Ranger authorizer returns no result, instead of skipping
    the checks or leaving the row filter / data mask unapplied;
  - read the security-critical configurations (the authorizer implementation, the
    Ranger PDP server address, the Ranger service name, and the Ranger PDP client
    authentication and encryption settings) from `ranger-spark-security.xml` only:
    JVM system properties can neither set nor override them.

- New and updated unit tests:
  - `AccessResourceSuite`: RRN metacharacter escaping (round-trip through the real
    `RangerResourceNameParser`), blank resource name denial, uri resource variants.
  - `SparkRangerAdminPluginSuite`: accesses denied when the authorizer returns no
    result (per-request and single-call paths), security-critical configurations
    cannot be overridden by system properties, fail fast on a missing Ranger PDP url.
  - `RangerRemoteAuthorizerSuite` (+ new `MockPdpServer`): the thin client against a
    mock PDP server.
- Unit suites (all pass):
  build/mvn -nsu test -pl extensions/spark/kyuubi-spark-authz -am \
    -Dtest=none \
    -DwildcardSuites=org.apache.kyuubi.plugin.spark.authz.ranger.RangerRemoteAuthorizerSuite,org.apache.kyuubi.plugin.spark.authz.ranger.SparkRangerAdminPluginSuite,org.apache.kyuubi.plugin.spark.authz.ranger.AccessResourceSuite
- Data masking and row filtering end-to-end suites (26 tests) pass.
- Built Apache Ranger 2.9.0 server via docker-compose locally and validated against it with:
  build/mvn -nsu test -pl extensions/spark/kyuubi-spark-authz -am \
    -DforkMode=never \
    -DwildcardSuites=org.apache.kyuubi.plugin.spark.authz.ranger.RangerRemoteAuthorizerSuite \
    -Dranger.authz.remote.pdp.url=http://localhost:6500 \
    -Dranger.authz.remote.authn.type=header \
    -Dranger.authz.remote.authn.header.X-Forwarded-User=ranger
Also verified the uri variant fix on the real PDP with both an exact-path url
policy and a trailing-slash url policy.

Assisted-by: OpenCode (GLM-5.3-Flash)


wForget:
@aajisaka Thank you for this work. Did you verify `kyuubi-spark-authz-shaded` module? Do the maven-shade-plugin configurations need to be changed?

https://github.com/apache/kyuubi/blob/be35121c5d6d53150459ad14afe661cb3ce3e69c/extensions/spark/kyuubi-spark-authz-shaded/pom.xml#L46-L112

aajisaka:
> @aajisaka Thank you for this work. Did you verify `kyuubi-spark-authz-shaded` module? Do the maven-shade-plugin configurations need to be changed?
> 
> https://github.com/apache/kyuubi/blob/be35121c5d6d53150459ad14afe661cb3ce3e69c/extensions/spark/kyuubi-spark-authz-shaded/pom.xml#L46-L112

Sorry I forgot to verify. I removed Jackson 1 config because Ranger 2.8.0+ doesn't have its dependency. Also, updated the LICENSE file of the shaded jar.

pan3793:
If we have to make a breaking change here eventually, I wonder if we should do more things to move to Ranger 2.9 and [PDP mode](https://ranger.apache.org/blogs/integrating_applications.html#s5).

The current architecture- pulling all policies from the Ranger Admin Server on Spark application start, and evaluating authz requests on the client side (Spark driver), does not fit short-lived Spark ETL cases, and another disadvantage is that we have to bundle a heavy Ranger client that pulls in many dependencies. (easy to conflict, varies across versions, vulnerable)

In PDP mode, the Ranger client becomes lightweight and is almost an HTTP client, also no need to download the whole policies to the client. There are many organizations have similar enhancements to make the Ranger Spark Plugin fit large-scale lakehouses; now this is supported by the upstream Ranger community, we should migrate to it ASAP.

UT infrastructure might require a significant adjustment to allow us to adopt the Ranger PDP ...

aajisaka:
Thank you @pan3793

While it's a significant change, it really makes sense to me. I'll try to move to Ranger 2.9 and support PDP mode.

aajisaka:
Updated to support Ranger 2.9.0 and also updated the PR description.

Issue #7438, opened 2026-05-09
[FEATURE] Make JSON event logger permission management configurable


kongjiexin888-source:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Add a configuration to control whether the built-in JSON event logger actively manages event log directory and file permissions.

Proposed configuration:

```properties
kyuubi.event.json.log.manage.permissions.enabled=true
```

The default value should be true to keep the existing behavior unchanged.

When set to false, the JSON event logger should still create and write event log paths, but it should not actively set fixed permissions on event log directories or files.

### Motivation

In some production deployments, event log directories are not managed only by POSIX-style HDFS permissions. They may be governed by Apache Ranger, inherited HDFS ACLs, default ACL inheritance, or other external authorization systems.

In these environments, the engine user may have enough permission to create and write JSON event logs, but may not be allowed to change permissions on the target path. The current JSON event logger behavior can therefore fail during initialization or log file creation because it actively sets permissions.

This makes multi-tenant Kyuubi deployments harder to manage when a shared JSON event log root is controlled by Ranger or another external permission system.

Making this behavior configurable allows administrators to keep the current default behavior, or disable Kyuubi-managed JSON event log permissions and let the external permission system manage access control.

### Describe the solution

Introduce a new boolean configuration:

```properties
kyuubi.event.json.log.manage.permissions.enabled=true
```

When the value is true, the JSON event logger keeps the current behavior:
1. create event log directories with the existing fixed directory permission
2. set event log files to the existing fixed file permission

When the value is false, the JSON event logger:
1. creates event log directories without passing fixed permissions
2. does not call setPermission for JSON event log files
3. still validates that the configured log root is a directory
4. still writes JSON event logs normally

This keeps backward compatibility while allowing deployments with Ranger, HDFS ACLs, or other external permission systems to manage access control outside Kyuubi.

### Additional context

This is useful for multi-tenant HDFS deployments where `kyuubi.engine.event.json.log.path` points to a shared event log root.

Example runtime configuration:

```properties
kyuubi.engine.event.loggers=SPARK,JSON
kyuubi.engine.event.json.log.path=hdfs:///path/to/kyuubi/engine-events
kyuubi.event.json.log.manage.permissions.enabled=false
```

[...]

PR #7439, opened 2026-05-09
[KYUUBI #7438] Make JSON event logger permission management configurable

kongjiexin888-source:
The built-in JSON event logger currently creates event log directories and actively sets permissions on directories and log files.

This default behavior is useful for the built-in permission model, but it can conflict with deployments where access control is managed by Apache Ranger, inherited HDFS ACLs, or other external authorization systems. In those environments, an engine user may have permission to create and write event logs, while still not being allowed to call `setPermission` on the target path.

This patch adds `kyuubi.event.json.log.manage.permissions.enabled`, defaulting to `true`, to keep the existing behavior unchanged. When set to `false`, the JSON event logger no longer actively manages directory or file permissions, leaving access control to Ranger, HDFS ACLs, filesystem defaults, or other external permission systems.

This helps multi-tenant Kyuubi deployments use a shared JSON event log root without requiring
each engine user to own or chmod the target HDFS path.

- [x] Added tests for default JSON event log file permission management.
- [x] Added tests for writing JSON event logs when permission management is disabled.
- [x] Ran:
```bash
build/mvn -pl kyuubi-events -am -DwildcardSuites=org.apache.kyuubi.events.handler.JsonLoggingEventHandlerSuite test
```

This patch was prepared with assistance from OpenAI Codex; the contributor reviewed and validated the changes.


pan3793:
@kongjiexin888-source thanks for contributing, but generally patches should go to master, not a released branch.

kongjiexin888-source:
> @kongjiexin888-source thanks for contributing, but generally patches should go to master, not a released branch.

Thanks for the guidance. I have retargeted the patch to master.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2649,6 +2649,16 @@ object KyuubiConf {
       .stringConf
       .createWithDefault("file:///tmp/kyuubi/events")
 
+  val EVENT_JSON_LOG_MANAGE_PERMISSIONS_ENABLED: ConfigEntry[Boolean] =
+    buildConf("kyuubi.event.json.log.manage.permissions.enabled")
```
pan3793: use dot for namespace, camelCase for words
```suggestion
    buildConf("kyuubi.event.json.log.managePermissions.enabled")
```

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -2649,6 +2649,16 @@ object KyuubiConf {
...
+      .doc("Whether the built-in JSON event logger manages the permissions of event log " +
+        "directories and files. When disabled, Kyuubi only creates and writes event log paths, " +
+        "leaving permissions to the underlying file system defaults, inherited ACLs, or " +
+        "external authorization systems.")
```
pan3793: say which permission will be set if enabled.

pan3793:
CI pipeline was not correctly triggered, not sure why

kongjiexin888-source:
Addressed. The config key has been renamed to `kyuubi.event.json.log.managePermissions.enabled`, and the config description now documents the managed directory/file permissions: directories `770` and files `660`.

wangzhigang1999:
Sharing some tips: before pushing your code, it's generally recommended to run `mvn spotless` and Scala style check. This can help you avoid many style-related checker failures. If you've modified any configuration options, you can run `dev/gen/gen_all_config_docs.sh` to regenerate the documentation. You can tell your Codex to automatically run these after making code changes. CI is quite expensive and slow, so it's better to catch these issues locally.

pan3793 (approved):

pan3793:
thanks, merged to master

PR #7442, opened 2026-05-11
[KYUUBI #7305] Route JDBC engine metadata calls through DatabaseMetaData

### LINKED ISSUE #7305: [Bug] DBeaver returned all tables in all databases when connecting.

wangzhigang1999:

The JDBC engine used to answer metadata requests with dialect-specific, hand-written `INFORMATION_SCHEMA` SQL for `getCatalogs`, `getSchemas`, `getTables`, `getColumns`, `getTypeInfo`, `getFunctions`, `getPrimaryKeys`, and `getCrossReference`.

That path was hard to keep aligned with the JDBC contract. Clients could see swapped or missing catalog/schema columns, backend-specific labels such as Impala's `TABLE_MD`, and inconsistent metadata filter handling across dialects.

This PR replaces those per-dialect SQL builders with the standard JDBC `DatabaseMetaData.getXxx(...)` APIs from the backend driver, while keeping catalog-vs-schema conventions and metadata-only label normalization in `JdbcDialect`.

Concretely this is the cause of #7305 ("DBeaver returned all tables in all databases when connecting"): the default `JdbcDialect` threw `featureNotSupported` for `getCatalogsOperation` / `getSchemasOperation`, so DBeaver could not obtain a database list and fell back to `getTables(null, null, ...)`, which the Hive client rewrites to schemaPattern=`"%"`. Routing `getCatalogs` / `getSchemas` through `DatabaseMetaData` lets the backend driver answer them directly, so DBeaver gets a real catalog/schema list and the all-databases fallback no longer fires. This is the direction @pan3793 suggested on #7419, which this PR supersedes.

### What changes?

- Replace the hand-written metadata SQL path with backend `DatabaseMetaData` calls.
- Add dialect support for database-term handling and metadata column-label normalization.
- Pass the resolved session dialect consistently through metadata operations.
- Apply the URL-level database before running per-session initialize SQL.
- Add `ResultSetFetchIterator` for correct metadata streaming, partial fetch, and idempotent close behavior.

### Compatibility notes

Metadata result sets now follow the JDBC spec more closely. Clients using spec column names should benefit; clients depending on the old non-spec labels or catalog/schema layout may need updates.

When the JDBC URL carries a database, `JdbcSessionImpl` now switches to it before executing the per-session initialize SQL, so init SQL runs in the URL-specified database rather than the driver's default.

- Updated existing per-dialect `OperationSuite` coverage for MySQL, PostgreSQL, Oracle, Phoenix, ClickHouse, Doris, StarRocks, and Impala.
- Added backend metadata contract suites that verify the Hive JDBC client over Thrift sees JDBC-spec column names, key fields, and metadata filters.
- Added `ResultSetFetchIteratorSuite` for prefetch, partial fetch, and close behavior.
- Verified end-to-end via beeline against all eight dialects (MySQL, PostgreSQL, Oracle, Phoenix, ClickHouse, Doris, StarRocks, Impala): metadata operations return JDBC-spec column layouts and `SELECT` against existing tables succeeds.
- Full DDL/DML cycles (`DROP IF EXISTS` -> `CREATE` -> `INSERT` -> `SELECT` -> `DROP`) verified on StarRocks, Doris, MySQL, PostgreSQL, and ClickHouse; Oracle / Phoenix / Impala excluded since their DDL grammar diverges.

Yes. Claude Opus 4.7 helped draft parts of the test cases and local test-environment setup. The implementation, test results, and final changes were reviewed and validated by the author.

---

Closes #7305. Supersedes #7419.


externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/session/JdbcSessionImpl.scala
```diff
@@ -63,6 +64,16 @@ class JdbcSessionImpl(
       sessionConnection = ConnectionProvider.create(sessionConf)
       databaseMetaData = sessionConnection.getMetaData
     }
+    // Apply URL database before init SQL so DDL runs in the user-requested database.
+    // "default" is a stub Hive clients send when the URL has no `/<db>` - skip it, since
+    // forwarding it would fail on backends (PG, ClickHouse) lacking that database.
+    conf.get(USE_DATABASE) match {
```
pan3793, line 78: also handle `USE_CATALOG`

wangzhigang1999: Done

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/DatabaseTermSupport.scala
```diff
@@ -0,0 +1,77 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kyuubi.engine.jdbc.dialect
+
+import java.sql.{Connection, SQLFeatureNotSupportedException}
+
+/**
+ * Mixin for dialects whose driver exposes the database as either the JDBC catalog or the
```
pan3793: can we replace this with a normal hierarchical inheritance?

JdbcDialect => MySQLDialect => DorisDialect/StarRocksDialect

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/DatabaseTermSupport.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  // Suppress only the "setter not implemented" signal; real failures (missing db, permission
+  // denied) propagate. If neither setter accepts, throw rather than silently no-op.
+  protected def setDatabase(conn: Connection, database: String): Unit = {
+    var accepted = false
+    try {
+      conn.setCatalog(database)
+      accepted = true
+    } catch {
+      case _: SQLFeatureNotSupportedException =>
```
pan3793: when code goes here? which driver lib and which version

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/impala/ImpalaSchemaHelper.scala
```diff
@@ -27,4 +27,13 @@ class ImpalaSchemaHelper extends SchemaHelper {
   override protected def realToTTypeId: TTypeId = {
     TTypeId.DOUBLE_TYPE
   }
+
+  // Apache Impala HS2 GetColumns labels the schema column "TABLE_MD" instead of the
+  // JDBC-spec "TABLE_SCHEM" (Impala upstream issue, not Kyuubi's). Rewrite the label so
```
pan3793: any jira tickets or discussion links?

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteMetaDataOperation.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.kyuubi.session.Session
+import org.apache.kyuubi.shaded.hive.service.rpc.thrift.TFetchResultsResp
+
+/**
+ * Runs a `DatabaseMetaData.getXxx(...)` call against the session connection and exposes the
+ * resulting [[ResultSet]] as the operation output.
+ */
+class ExecuteMetaDataOperation(
+    session: Session,
+    metadataCall: ExecuteMetaDataOperation.MetaDataCall)
```
pan3793: I would replace this with normal hierarchical inheritance instead of higher-order functions. the current way will be more flexible if we have many different actions. but in fact, there are only a few, create concrete classes to match each operation to fit the style of other engines.

pan3793 (commented):
thanks for making this big change, it overall looks good. Have you tried with DBeaver, is problem solved?

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/impala/ImpalaSchemaHelper.scala
wangzhigang1999: https://github.com/apache/impala/blob/master/fe/src/main/java/org/apache/impala/service/MetadataOp.java#L114
I discovered this issue during testing. I think it might be a typo. You can refer to the source code above.
The row-population site at L494 even comments // TABLE_SCHEM.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/DatabaseTermSupport.scala
wangzhigang1999: This is just a fallback approach. During my verification, this branch was never actually triggered. If you think it's unnecessary, it can certainly be simplified/removed.

wangzhigang1999:
> thanks for making this big change, it overall looks good. Have you tried with DBeaver, is problem solved?

Verified end-to-end on Impala 4.5.0 — DBeaver via Kyuubi shows the full column tree: 

image

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/DatabaseTermSupport.scala
pan3793: let's remove it to keep things simple

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/impala/ImpalaSchemaHelper.scala
pan3793: add a source code link (with the latest release tag) to the comment (also better to search Jira to see if it has been tracked)

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/DatabaseTermSupport.scala
wangzhigang1999: Done in 15bdd03ef. Removed the trait.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/DatabaseTermSupport.scala
wangzhigang1999: Done in 15bdd03ef. Dropped the `SQLFeatureNotSupportedException` try/catch — JDBC spec says these methods should silently no-op when unsupported, never fired in our 8-dialect coverage.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/impala/ImpalaSchemaHelper.scala
wangzhigang1999: Done in 94883e4ae. Permalink anchored to 4.5.0 release tag added in the comment.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteMetaDataOperation.scala
wangzhigang1999: Done in d92796691.

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/dialect/ImpalaDialect.scala
```diff
@@ -16,80 +16,40 @@
...
+  // Symmetric to `setSchema`: `KyuubiConnection#getSchema` ships a Kyuubi-private session
+  // config (`kyuubi.operation.get.current.database`) that Impalad rejects. Read the current
+  // database via plain SQL.
+  override def getCurrentSchema(conn: Connection): String = {
+    val stmt = conn.createStatement()
+    try {
+      val rs = stmt.executeQuery("SELECT current_database()")
+      try if (rs.next()) rs.getString(1) else null
+      finally rs.close()
+    } finally stmt.close()
```
pan3793: there are a few utilities method in `JdbcUtils` can simplify the code

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteMetaDataOperation.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.kyuubi.operation.OperationState.OperationState
+import org.apache.kyuubi.operation.log.OperationLog
+import org.apache.kyuubi.session.Session
+import org.apache.kyuubi.shaded.hive.service.rpc.thrift.TFetchResultsResp
+
+/**
+ * Base class that streams the result of a `DatabaseMetaData.getXxx(...)` call against the
+ * session connection. Subclasses fix the particular metadata RPC via [[runMetaDataCall]].
+ */
+abstract class ExecuteMetaDataOperation(session: Session) extends JdbcOperation(session) {
```
pan3793: ExecuteMetaDataOperation => MetaDataOperation

externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/operation/ExecuteMetaDataOperation.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  protected def runMetaDataCall(dialect: JdbcDialect, conn: Connection): ResultSet
```
pan3793: runMetaDataCall => fetchMetaData

pan3793 (approved):
LGTM, a few nits.

pan3793:
thanks, merged to master

PR #7054, opened 2025-05-03
Add server only flag for more server/credentials/frontend/metadata configs 

turboFei:
### Why are the changes needed?

Reduce the kyuubi server end configs involved into engine end.

### How was this patch tested?

UT and  code review.
### Was this patch authored or co-authored using generative AI tooling?

No.

PR #7451, opened 2026-05-15
[KYUUBI #7387] Replace serverOnly with audience and immutable on ConfigEntry

LamiumAmplexicaule:
This implementation is based on the following suggestion.
- https://github.com/apache/kyuubi/pull/7451#issuecomment-4587409731

`ConfigEntry.serverOnly` conflated two concerns, "don't send to engines" and "users can't override", by stripping entries from the config in `getUserDefaults()`.
This caused `kyuubi.server.redaction.regex` to be removed from the session config before `ProcBuilder.toString` could use it for redacting sensitive values in engine command lines (#7387).

Additionally, `conf.getAll` sends all configs to every engine regardless of engine type, causing Flink-specific configs to be sent to Spark, Trino configs to Hive, etc.

Splitting these concepts into audience and immutable provides the following benefits:
- Server-side code to always see all config values (fixing redaction — Closes #7387)
- Engine process builders to receive only audience-appropriate configs via `getEngineConf`

---

[redactCommandLineArgs](https://github.com/apache/kyuubi/blob/7066271318b2a1d81e4a3e7b7c89f622b14107ad/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala#L329-L348) previously only redacted `key=value` arguments that followed a `--conf` marker. 
Since arguments like `-Dsome.secret=value` are not preceded by `--conf`, they were not redacted even when matching `SERVER_SECRET_REDACTION_PATTERN`.
The `--conf` guard is removed so that all arguments containing `=` are subject to redaction.

Unit tests.

```shell
$ ./build/mvn test -pl kyuubi-server -am -Dtest=none -DwildcardSuites=org.apache.kyuubi.operation.KyuubiServerInfoProviderSuite
$ ./build/mvn test -pl kyuubi-common -am -Dtest=none -DwildcardSuites=org.apache.kyuubi.config,org.apache.kyuubi.session.SessionManagerValidationSuite
$ ./build/mvn test -pl kyuubi-server -am -Dtest=none -DwildcardSuites=org.apache.kyuubi.engine.dataagent.DataAgentProcessBuilderSuite,org.apache.kyuubi.engine.flink.FlinkProcessBuilderSuite,org.apache.kyuubi.engine.trino.TrinoProcessBuilderSuite,org.apache.kyuubi.engine.jdbc.JdbcYarnModeProcessBuilderSuite,org.apache.kyuubi.engine.jdbc.JdbcProcessBuilderSuite,org.apache.kyuubi.engine.hive.HiveProcessBuilderSuite,org.apache.kyuubi.engine.hive.HiveYarnModeProcessBuilderSuite,org.apache.kyuubi.engine.spark.SparkBatchProcessBuilderSuite,org.apache.kyuubi.engine.spark.SparkProcessBuilderSuite
```

Assisted-by: Claude:claude-opus-4-6


kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -325,17 +325,10 @@ object Utils extends Logging {
...
-          case PATTERN_FOR_KEY_VALUE_ARG(key, value) if nextKV =>
+          case PATTERN_FOR_KEY_VALUE_ARG(key, value) =>
             val (_, newValue) = redact(redactionPattern, Seq((key, value))).head
-            nextKV = false
             genKeyValuePair(key, newValue)
-
-          case cmd if cmd == CONF =>
-            nextKV = true
-            cmd
-
```
LamiumAmplexicaule, line 338: In the test that checks whether the logs in FlinkProcessBuilder are redacted, I noticed that values in the `-D` format don’t get redacted, so I removed this filter.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/dataagent/DataAgentProcessBuilder.scala
```diff
@@ -34,6 +34,7 @@ import org.apache.kyuubi.operation.log.OperationLog
 import org.apache.kyuubi.util.command.CommandLineUtils._
 
 class DataAgentProcessBuilder(
+    override val serverConf: KyuubiConf,
```
wangzhigang1999: I don't think adding an extra `serverConf` parameter is a good idea. This would result in two configurations that appear identical except for their names, which could cause significant confusion during usage—users might wonder, "Which configuration should I actually use?"

LamiumAmplexicaule: Thank you for your comments.

My understanding is that the reason redaction no longer happens after #7054 is that, in
https://github.com/apache/kyuubi/blob/e769f4239891536e9850c11b9661ea65d82a01f9/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala#L95, the sessionConf passed in has the serverOnly configs stripped out.

If we want to obtain `kyuubi.server.redaction.regex` from ProcessBuilder without passing serverConf, we need to explicitly set it somewhere.
However, it feels wrong to propagate server-side config into the session/engine configs, so I chose the approach of passing serverConf instead.

When passing conf as serverConf, if we strip out everything except serverOnly, we might avoid accidentally referencing serverConf.

LamiumAmplexicaule: I am planning to update this PR to pass `Option[Regex]` instead of serverConf.

wangzhigang1999:
Thanks for tracking this down!

I'd like to talk about how the fix is done before it lands. Not because the direction is wrong — I'm just not sure that passing the pattern into every `ProcBuilder` is the best long-term approach. I might be missing something, so feel free to disagree.

**What I agree with:** redaction is done on the server side, so the pattern has to be read from the server config, not from the config that goes to the engine. That part is correct and necessary.

**My concern is the mechanism.** `conf.get(...)` returned `None` because `serverOnly` is implemented by *removing* the entry from the per-session config in `getUserDefaults`. `ProcBuilder` runs on the server but holds that stripped config, so redaction quietly turned off. Passing `Option[Regex]` down fixes redaction, but it only fixes this one symptom:

- It doesn't scale. The next `serverOnly` config that some server-side code needs while launching an engine will hit the same problem, and we'd add another parameter for it.
- An earlier version of this PR passed `serverConf: KyuubiConf` instead. I think that's worse: now there are two configs of the same type, told apart only by their variable name — which is exactly the kind of mix-up that caused this bug.

So the real problem is that `serverOnly` mixes two unrelated things into one "remove": "don't send this to the engine" and "users can't change this". Removing the entry does both, but it also means any server-side code that legitimately needs the value can no longer read it.

**There's a related problem with the same cause**, which is why I'd rather fix the model than patch it: the launch path does `conf.getAll` with no per-engine filtering (`SparkProcessBuilder`), prefixes everything with `spark.`, and ships it. So a Flink tuning key like `kyuubi.engine.flink.memory`, set in a shared `kyuubi-defaults.conf`, ends up on the Spark engine's command line as `spark.kyuubi.engine.flink.memory=...` — harmless noise for tuning keys, but any sensitive value in there rides along too. Both problems are "send everything, then remove a few".

**The direction I'd suggest** is to split `serverOnly` into two separate, independent flags, and stop using "remove" to control where a config goes:

- `audience` — which engines a config is allowed to reach. Instead of removing it from the config, we filter it out when building the engine's startup command. The value is figured out from the key prefix by default (`kyuubi.engine.spark.*` → Spark, `kyuubi.server.*` → server-only), **and** can also be set by hand on the entry when the prefix isn't enough. So most configs need no marking at all, but you can still set it manually when you need to.
- `immutable` — users can't override it. This is checked when Kyuubi merges the user's config. It's set by hand, on purpose, because "can a user change this?" is a security decision worth deciding per config.

Then `serverOnly` is just `audience(server-only) + immutable`, and we can drop it. Two benefits: (1) the value stays in the server config and is fixed, so redaction goes back to a plain `conf.get` and the `Option[Regex]` passing here goes away completely; (2) filtering by `audience` also stops sending one engine's configs to other engines. One model, both problems, no per-config wiring.

```mermaid
flowchart TB
    MASTER["server config (loaded on server, the source of truth)"]
    UO["user config: connection / SET / user-defaults"]
    UO --> IMM{"immutable key?"}
    IMM -->|yes| PIN["user value ignored, server value kept"]
    IMM -->|no| APPLY["user value applied"]
    MASTER --> CONF
    PIN --> CONF
    APPLY --> CONF
    CONF["full config on the server — nothing removed"]
    CONF -->|"conf.get(...)"| READ["server-side reads — redaction sees the real value"]
    CONF --> PROJ{"build config for target"}
    PROJ -->|"engine = Spark"| SPARK["send to Spark: audience is Spark or all-engines"]
    PROJ -->|"engine = Flink"| FLINK["send to Flink: audience is Flink or all-engines"]
    PROJ -->|"logs / REST"| LOG["sent out, masked if sensitive"]
```

In short: `immutable` is checked when a user config comes in, `audience` is applied when a config goes out, and because nothing is removed from the server config, server-side reads (including redaction) always see the real value — which is exactly what broke here.

Curious to hear your thoughts! 

CC @pan3793

LamiumAmplexicaule:
Thank you for the suggestion.
I think it's a very good idea.
It also reduces the number of items we need to configure in `kyuubi.session.conf.restrict.list`, which should make it easier to manage.

LamiumAmplexicaule:
I'm working on implementing the suggestion.

LamiumAmplexicaule:
Looking at the failing tests, it seems that `SERVER_INFO_PROVIDER` needs to be mutable.
- https://github.com/apache/kyuubi/blob/dfd9b4935378c79d3c1f99e344273ff71e21765b/integration-tests/kyuubi-hive-it/src/test/scala/org/apache/kyuubi/it/hive/operation/KyuubiOperationHiveEnginePerUserSuite.scala#L43-L52

Related: https://github.com/apache/kyuubi/issues/7435

I checked that KyuubiServerInfoProviderSuite passes on this branch.

```
$ ./build/mvn test -pl kyuubi-server -Dtest=none -DwildcardSuites=org.apache.kyuubi.operation.KyuubiServerInfoProviderSuite
```

wangzhigang1999:
Thank you for your work! I’ll review it as soon as possible!

kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigAudience.scala
```diff
@@ -0,0 +1,73 @@
...
+  def isAudienceMatch(
+      configAudience: Set[ConfigAudience],
+      target: ConfigAudience): Boolean = {
+    configAudience.contains(ANY) ||
+    (configAudience.contains(ALL_ENGINES) && ENGINE_VALUES.contains(target)) ||
+    configAudience.contains(target)
+  }
+
+  def inferAudience(key: String): Set[ConfigAudience] = {
+    if (key.startsWith("kyuubi.server.") || key.startsWith("kyuubi.frontend.") ||
```
wangzhigang1999, line 48: This inference strips config that the engine actually reads at runtime. The engine runs its own thrift frontend (`SparkTBinaryFrontendService extends TBinaryFrontendService`, built on the shared kyuubi-common code) and reads at least:

- `kyuubi.frontend.connection.url.use.hostname` (`TFrontendService#connectionUrl` — decides whether the engine registers to service discovery by hostname or IP; critical on K8s)
- `kyuubi.frontend.thrift.min.worker.threads` / `.max.worker.threads` / `.worker.keepalive.time` (`TBinaryFrontendService`)

None of these are `serverOnly` on master (they're `fallbackConf` entries), so today they're passed to the engine on the command line via `conf.getAll`. After this change they're inferred as `SERVER` and filtered out, and the engine silently falls back to defaults. Their fallback keys are also `kyuubi.frontend.*` → also inferred `SERVER`, so the fallback materialization in `getEngineConf` can't recover them either.

Could we give these entries an explicit `.audience(...)` that includes the engines (the manual-override escape hatch this PR introduces), plus a test asserting they survive in `getEngineConf` for each engine type? Since these are registered entries, inference keeps applying to them unless they're annotated explicitly.

> Note: `kyuubi.frontend.advertised.host` is also read by `connectionUrl`, but it's already `serverOnly` on master (not forwarded today), so it's not a regression and needs no change.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigAudience.scala
```diff
@@ -0,0 +1,73 @@
...
+      key.startsWith("kyuubi.session.engine.hive.") || key.startsWith("hive.")) {
+      Set(HIVE)
+    } else if (key.startsWith("kyuubi.engine.trino.") ||
+      key.startsWith("kyuubi.session.engine.trino.") || key.startsWith("trino.")) {
+      Set(TRINO)
+    } else if (key.startsWith("kyuubi.engine.jdbc.") ||
+      key.startsWith("kyuubi.session.engine.jdbc.")) {
+      Set(JDBC)
+    } else if (key.startsWith("kyuubi.engine.data.agent.") ||
+      key.startsWith("kyuubi.session.engine.data-agent.")) {
```
wangzhigang1999, line 67: This branch matches no key and hard-codes a separator inconsistent with the rest of data-agent. Real keys are dot-separated (`kyuubi.engine.data.agent.*`), but this branch — and the `ConfigAudienceSuite` assertion on `kyuubi.session.engine.data-agent.main.resource` — uses a hyphen for a key that has no `buildConf` anywhere. Drop the branch and its test; add it back with the real (dot) naming if a session-level data-agent key lands.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -214,11 +205,47 @@ case class KyuubiConf(loadSysDefault: Boolean = true) extends Logging {
...
+    val extraServerPrefixes = get(KyuubiConf.SERVER_ONLY_PREFIXES)
+
+    def resolveAudience(key: String): Set[ConfigAudience.Value] = {
+      if (KyuubiReservedKeys.reservedKeys.contains(key)) {
+        Set(ConfigAudience.ANY)
+      } else {
+        val entry = KyuubiConf.getConfigEntryOrAlternative(key)
+        if (entry != null && entry.audience.isDefined) {
+          entry.audience.get
+        } else if (extraServerPrefixes.exists(prefix => key.startsWith(prefix))) {
```
wangzhigang1999, line 222: Behavior is fine, but the doc is now stale: `SERVER_ONLY_PREFIXES` still says "it only take affects for the configs that is not defined as a Kyuubi ConfigEntry," while this branch now also applies to registered entries without an explicit audience (master excluded registered entries via `getConfigEntry(key) == null`). Please update that doc line.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
wangzhigang1999, line 1: `KyuubiReservedKeys.reservedKeys` is hand-maintained — a future reserved key forgotten here that matches a server prefix would be silently dropped from engine config. A reflection test asserting all constants are in the set would lock it down.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigAudience.scala
LamiumAmplexicaule: Thank you for your review.
I added this because it’s referenced here. 

- https://github.com/apache/kyuubi/blob/dfd9b4935378c79d3c1f99e344273ff71e21765b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/ProcBuilder.scala#L67
- https://github.com/apache/kyuubi/blob/f06bf5ffc4a0f0dab0db725e71bda9c857d92ac8/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/dataagent/DataAgentProcessBuilder.scala#L49

If we remove it, then when `kyuubi.session.engine.data-agent.main.resource` is set, it ends up being sent to other engines as well.
Should we infer that the intended audience of `kyuubi.session.engine.*.main.resource` is the server?

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiReservedKeys.scala
LamiumAmplexicaule: Thank you for your review.

I added it in https://github.com/apache/kyuubi/pull/7451/changes/4bd82cf64127f36642487859be07dda994f85f6c

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
LamiumAmplexicaule: Thank you for your review.

I fixed it in https://github.com/apache/kyuubi/pull/7451/changes/59f09d984519b6274db05288557bd0a64afad222.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/ConfigAudience.scala
LamiumAmplexicaule: Thank you for your review.

I fixed it in https://github.com/apache/kyuubi/pull/7451/changes/9aec393139c7f63a964d75827030aca534e65458.

wangzhigang1999:
@pan3793 Could you take a look when you have time? 

This PR replaces the `serverOnly` mechanism you signed off on in #3879 — would appreciate your input on the new audience/immutable model.

wangzhigang1999:
I tested this end-to-end on a real Kyuubi + Spark deployment, comparing the current master behavior with this PR using the same JDBC configurations and `kyuubi-defaults.conf`.

The current implementation changes the semantics of unregistered native configurations. For example, with:

`hive.metastore.uris=thrift://127.0.0.1:65531`

current master passes it to the Spark engine launch command as:

`--conf spark.hive.metastore.uris=thrift://127.0.0.1:65531`

and the engine fails during initialization as expected because the metastore is unreachable.

With this PR, `hive.metastore.uris` is inferred as `HIVE` audience and removed from the Spark launch configuration. The Spark engine therefore initializes successfully using its existing metastore configuration.

When the configuration comes from the JDBC session conf, it is forwarded later through `OpenSession`, so `SET hive.metastore.uris` still returns the supplied value. However, that is too late for configurations consumed while initializing Spark, the Hive client, catalog, extensions, or Hadoop subsystems. Seeing the value through `SET` is therefore not equivalent to preserving the previous startup behavior.

There is an additional difference for configurations in `kyuubi-defaults.conf`: an unregistered `hive.*` entry is filtered from the Spark launch configuration and is not forwarded later through the session path either. In the tested engine it was simply `<undefined>`.

I also tested `spark.*`, `flink.*`, arbitrary custom keys, registered Kyuubi configs, and redaction. `spark.*`, custom keys, registered configs, and redaction behaved as intended. The regression is specifically caused by applying native-prefix audience inference to unregistered configuration keys.

I think prefix inference for unregistered native configurations needs to preserve the existing cross-engine/startup behavior. Audience filtering should primarily be applied to registered Kyuubi configuration entries, or unregistered configs should otherwise remain `ANY`, unless there is an explicit and backward-compatible mechanism for selecting their engine audience.

LamiumAmplexicaule:
Thank you for testing this end-to-end.
In my test environment, `hive.metastore.uris` is defined in hive-site.xml, so I didn’t notice this issue.

My understanding is that among the native prefixes covered by ConfigAudience, only `hive.*` may actually be required by the Spark SQL Engine.

* Spark
    * Consumes `hive.*` configurations such as `hive.metastore.uris`, as you pointed out.
* Trino
    * Uses `trino.*` only to pass session properties to the engine and does not reference other native prefixes.
* Flink
    * Although it may use Hive configurations, Flink SQL specifies the location of hive-site.xml through the catalog, so it does not rely on forwarding `hive.*` configurations.
* JDBC / Data Agent
    * Do not reference native prefixes.

With that in mind, instead of treating all unregistered native configurations as ANY, what do you think about introducing a backward-compatible mechanism for selecting the engine audience, such as the following configuration?

```scala
val ENGINE_SPARK_EXTRA_CONF_PREFIXES: ConfigEntry[Seq[String]] =
  buildConf("kyuubi.config.engine.spark.extra.prefixes")
    .doc("A comma-separated list of config key prefixes that should be forwarded to the " +
      "Spark engine in addition to spark.*, kyuubi.engine.spark.* and " +
      "kyuubi.session.engine.spark.* prefixes. " +
      "This is useful for native engine configs like hive.* that Spark also consumes.")
    .audience(ConfigAudience.SERVER)
    .immutable
    .version("1.11.0")
    .stringConf
    .toSequence()
    .createWithDefault(Seq("hive."))
```

wangzhigang1999:
> Thank you for testing this end-to-end. In my test environment, `hive.metastore.uris` is defined in hive-site.xml, so I didn’t notice this issue.
> 
> My understanding is that among the native prefixes covered by ConfigAudience, only `hive.*` may actually be required by the Spark SQL Engine.
> 
> * Spark
>   
>   * Consumes `hive.*` configurations such as `hive.metastore.uris`, as you pointed out.
> * Trino
>   
>   * Uses `trino.*` only to pass session properties to the engine and does not reference other native prefixes.
> * Flink
>   
>   * Although it may use Hive configurations, Flink SQL specifies the location of hive-site.xml through the catalog, so it does not rely on forwarding `hive.*` configurations.
> * JDBC / Data Agent
>   
>   * Do not reference native prefixes.
> 
> With that in mind, instead of treating all unregistered native configurations as ANY, what do you think about introducing a backward-compatible mechanism for selecting the engine audience, such as the following configuration?
> 
> ```scala
> val ENGINE_SPARK_EXTRA_CONF_PREFIXES: ConfigEntry[Seq[String]] =
>   buildConf("kyuubi.config.engine.spark.extra.prefixes")
>     .doc("A comma-separated list of config key prefixes that should be forwarded to the " +
>       "Spark engine in addition to spark.*, kyuubi.engine.spark.* and " +
>       "kyuubi.session.engine.spark.* prefixes. " +
>       "This is useful for native engine configs like hive.* that Spark also consumes.")
>     .audience(ConfigAudience.SERVER)
>     .immutable
>     .version("1.11.0")
>     .stringConf
>     .toSequence()
>     .createWithDefault(Seq("hive."))
> ```

I still prefer keeping unregistered configurations as ANY and applying audience filtering only to registered ConfigEntrys.

The extra-prefix configuration adds complexity and requires users to maintain engine-specific allowlists. Defaulting it to hive. fixes this case, but does not preserve compatibility: any existing unregistered prefix missing from the list would be silently filtered after an upgrade.

Kyuubi knows the audience of registered entries, but cannot know how plugins or downstream extensions consume unregistered configurations. Keeping those as ANY is safer and preserves the current behavior.

LamiumAmplexicaule:
> I still prefer keeping unregistered configurations as ANY and applying audience filtering only to registered ConfigEntrys.

I made that change in https://github.com/apache/kyuubi/pull/7451/changes/1099ba78b8a7ef9a844cbb016346a7874e827784.

wangzhigang1999:
LGTM from my side. Thanks for addressing the compatibility concerns. 

Two minor cleanup items remain: the PR description should refer to `kyuubi.server.redaction.regex`, and the stale `serverOnly` comments can be updated. Neither is blocking. Since this PR changes how configuration is validated and propagated to every engine, I would still like another reviewer to take a close look before merging. I’ll also run a few more tests on my side.

AGENTS.md
```diff
@@ -130,7 +130,11 @@ Runtime dependency changes may also require `LICENSE-binary` and `NOTICE` update
 
 Configs are `ConfigEntry` instances in `KyuubiConf` or per-engine config classes.
 
-- Server-only entries call `.serverOnly` on the builder; `KyuubiConf` tracks them in `serverOnlyConfEntries` so they are stripped before engine-side propagation.
+- Each config entry has two independent access-control properties:
+  - **`audience`** — which consumers should receive the config. Set via `.audience(SERVER)`, `.audience(SPARK)`, etc. on the builder. Entries without explicit audience use prefix-based inference (`inferAudience`).
+  - **`immutable`** — whether users can override the config at session time. Set via `.immutable` on the builder. Admin-controlled sources (user defaults, session conf advisors) bypass immutable checks.
+- Server-only entries use `.audience(SERVER)` and `.immutable` on the builder.
```
pan3793: seems we can eliminate the server-only concept now.

AGENTS.md
```diff
@@ -130,7 +130,11 @@ Runtime dependency changes may also require `LICENSE-binary` and `NOTICE` update
 
 Configs are `ConfigEntry` instances in `KyuubiConf` or per-engine config classes.
 
-- Server-only entries call `.serverOnly` on the builder; `KyuubiConf` tracks them in `serverOnlyConfEntries` so they are stripped before engine-side propagation.
+- Each config entry has two independent access-control properties:
+  - **`audience`** — which consumers should receive the config. Set via `.audience(SERVER)`, `.audience(SPARK)`, etc. on the builder. Entries without explicit audience use prefix-based inference (`inferAudience`).
```
pan3793:
```patch 
- ... use prefix-based inference (`inferAudience`)
+ ... use prefix-based inference (`ConfigAudience.inferAudience`)
```

pan3793 (approved):
LGTM, a few minor issues

### Issue 1: `redactCommandLineArgs` drops the `--conf` guard — widens redaction scope

`kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala`

The previous implementation only redacted a `key=value` argument when it immediately
followed a `--conf` marker (guarded by `if nextKV`). This PR removes the `nextKV`
tracking entirely:

```scala
-        var nextKV = false
         commands.map {
-          case PATTERN_FOR_KEY_VALUE_ARG(key, value) if nextKV =>
+          case PATTERN_FOR_KEY_VALUE_ARG(key, value) =>
             val (_, newValue) = redact(redactionPattern, Seq((key, value))).head
-            nextKV = false
             genKeyValuePair(key, newValue)
-
-          case cmd if cmd == CONF =>
-            nextKV = true
-            cmd
-
           case cmd =>
             cmd
         }
```

`PATTERN_FOR_KEY_VALUE_ARG` is `(.+?)=(.+)`, which matches **any** argument containing
`=`. With the guard removed, every KV-shaped argument on the command line is now a
redaction candidate, not just the value following `--conf`.

For configs passed as `--conf key=value` (where `--conf` is a separate array element)
the behavior is unchanged. But non-`--conf` arguments such as `-Dsome.jvm.prop=value`
or classpath/path entries containing `=` are now parsed as KV pairs and run through
`redact()`. If such an argument matches the configured `SERVER_SECRET_REDACTION_PATTERN`,
it will now be redacted where it previously would not have been.

This appears intentional (redact secrets regardless of argument prefix), but it is a
behavioral scope change and should be called out explicitly in the PR description.
### Issue 2: Stale `serverOnly` comments in `KyuubiServerInfoProviderSuite`

`kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiServerInfoProviderSuite.scala` (lines 29, 38)

These comments still describe the removed `serverOnly` stripping behavior:

```scala
// kyuubi-defaults.conf entry, so the test exercises the serverOnly fallback path.
...
// Without the fix the serverOnly value is stripped from the session conf and getInfo
```

There is no longer a `serverOnly` flag; the test now exercises the `audience` +
`immutable` behavior. The test itself still passes, but the comments are outdated and
could mislead future readers. Consider updating them to reflect the new model.

AGENTS.md
LamiumAmplexicaule: I fixed it in https://github.com/apache/kyuubi/pull/7451/changes/b6ba20873e944410c72d4412abd7c89f8dc5559b.

AGENTS.md
LamiumAmplexicaule: I fixed it in https://github.com/apache/kyuubi/pull/7451/changes/b6ba20873e944410c72d4412abd7c89f8dc5559b.

LamiumAmplexicaule:
@wangzhigang1999
Thank you for your review.
> the PR description should refer to kyuubi.server.redaction.regex

I updated the PR description.

> the stale serverOnly comments can be updated.

I fixed it in https://github.com/apache/kyuubi/pull/7451/changes/cbf70e153acf3dc5c3e0e03f88569bd10158ac87.

@pan3793 
Thank you for your review.

> Issue 1: redactCommandLineArgs drops the --conf guard — widens redaction scope

I added an explanation to the PR description.

> Issue 2: Stale serverOnly comments in KyuubiServerInfoProviderSuite

I fixed it in https://github.com/apache/kyuubi/pull/7451/changes/cbf70e153acf3dc5c3e0e03f88569bd10158ac87.

> There is no longer a serverOnly flag; the test now exercises the audience + immutable behavior. 

As mentioned below, `SERVER_INFO_PROVIDER` needs to remain mutable, so I haven’t made it immutable.

- https://github.com/apache/kyuubi/pull/7451#issuecomment-4767528203

pan3793:
thanks, merged to master/1.12

PR #7455, opened 2026-05-17
[KYUUBI #6832] Initial impl Spark DSv2 YARN Connector that supports reading YARN aggregated logs

### LINKED ISSUE #6832: [FEATURE] Impl Spark DSv2 YARN Connector that supports reading YARN aggregation logs
### Code of Conduct

### Search before asking

### Describe the feature

Leverage the Spark DSv2 API to implement a connector that provides a SQL interface to access the YARN agg logs, and maybe other YARN resources in the future.

### Motivation

For large-scale Spark on YARN deployments, there are dozens or even hundreds of thousands of Spark applications submitted to the cluster per day, and the app logs are collected and aggregated by YARN stored on HDFS, sometimes we might want to analyze the logs to identify some cluster-level issues, for example, some machine might have hardware issues that frequently produce disk/network exceptions, it's straightforward to leverage Spark to analyze those logs in parallel.

### Describe the solution

the usage might be like

```
$ spark-sql --conf spark.sql.catalog.yarn=org.apache.kyuubi.spark.connector.yarn.YarnCatalog
> SELECT
    app_id, app_attempt_id,
    app_start_time, app_end_time,
    container_id, host,
    file_name, line_num, message
  FROM yarn.agg_logs
  WHERE app_id = 'application_1234'
    AND container_id='container_12345'
    AND host = 'hadoop123.example.com'
```

### Additional context

_No response_

pan3793:

Close #6832. This connector gives the Hadoop administrator a chance to analyze YARN aggregated logs at the cluster level, for example, aggregate container logs across applications by host to diagnose potential host hardware issues

The current initial implementation has several limitations:
- feat: only support `TFile`, but not `IFile`
- perf: only supports pushing down `app_id`, `user`, `host` filters, does not support pushing down `container_id`, `log_type`, though they are supposed to be selective
- perf: listing aggregated log files runs in a single thread during the planning phase - for a large cluster, it should run in parallel on the driver side or launch a job to do that on the executor side.
- etc. 

UT is not added yet, as this requires a real YARN cluster with history agg logs.

```
$ kyuubi-beeline -u 'jdbc:kyuubi://spark-dev1.foo.bar:10009/default' \
  --conf spark.jars=/tmp/kyuubi-spark-connector-yarn_2.12-1.12.0-SNAPSHOT.jar \
  --conf spark.sql.catalog.yarn=org.apache.kyuubi.spark.connector.yarn.YarnCatalog
0: > select
. .>   mtime, app_id, container_id, host, log_type, message
. .> from yarn.app_logs
. .> where
. .>   user = 'hadoop'
. .>   and host = 'spark-dev2.foo.bar'
. .>   and message like '%ERROR%'
. .>   and message not like '%RECEIVED SIGNAL TERM%'
. .>   and message not like '%Aborting task%'
. .> limit 2;
...
+--------------------------+---------------------------------+-----------------------------------------+---------------------+-----------+----------------------------------------------------+
|          mtime           |             app_id              |              container_id               |        host         | log_type  |                      message                       |
+--------------------------+---------------------------------+-----------------------------------------+---------------------+-----------+----------------------------------------------------+
| 2025-04-03 18:07:18.893  | application_1743671377509_0001  | container_1743671377509_0001_01_000001  | spark-dev2.foo.bar  | stdout    | 25/04/03 18:07:15 ERROR ApplicationMaster$AMEndpoint: Driver terminated with exit code 1! Shutting down. spark-dev1.foo.bar:16601 |
| 2025-04-03 18:07:18.893  | application_1743671377509_0001  | container_1743671377509_0001_01_000001  | spark-dev2.foo.bar  | stdout    | 25/04/03 18:07:15 ERROR ApplicationMaster$AMEndpoint: Driver terminated with exit code 1! Shutting down. spark-dev1.foo.bar:16601 |
+--------------------------+---------------------------------+-----------------------------------------+---------------------+-----------+----------------------------------------------------+
2 rows selected (0.648 seconds)
```

Assisted-by: Claude Opus 4.7.


extensions/spark/kyuubi-spark-connector-yarn/pom.xml
```diff
@@ -0,0 +1,275 @@
...
+    <parent>
+        <groupId>org.apache.kyuubi</groupId>
+        <artifactId>kyuubi-parent</artifactId>
+        <version>1.12.0-SNAPSHOT</version>
+        <relativePath>../../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>kyuubi-spark-connector-yarn_${scala.binary.version}</artifactId>
+    <packaging>jar</packaging>
+    <name>Kyuubi Spark Hadoop YARN Connector</name>
```
pan3793, line 30: Use a generic name - we may extend it to include other interactions with YARN, for example, use `YarnClient` to retrieve app list from RM, implement `STORE PROCEDURE`s which are equivalent to the `yarn` commands

pan3793:
This is a longstanding missing feature for Hadoop. YARN-1440 was raised in 2013 - Yarn aggregated logs are difficult for external tools to understand. Large-scale log processing is a typical use case for Hadoop, I'm surprised there is not an out-of-the-box solution in 2026 to analyze logs of applications run in Hadoop YARN in batch. 

@aajisaka @wForget @cxzl25, could you please take a look? and also would like to know if you have any better ideas to process yarn aggregate logs across the application

aajisaka:
I agreed it's a long standing missing feature. I'll take a look further.

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
```diff
@@ -0,0 +1,116 @@
...
+    val files = fs.globStatus(path)
+    val taken = System.currentTimeMillis() - start
+    logInfo(s"Listing $pathStr takes $taken ms, found ${files.length} files")
+    files
+  }
+
+  private def listFilesWithFilters(): Array[FileStatus] = {
+    val baseDir = remoteAppLogDir
+    val bucketDir = s"bucket-$remoteAppLogDirSuffix-tfile"
+    var path = s"$baseDir/{{USER}}/$bucketDir/{{BUCKET}}/{{APP_ID}}/{{HOST}}_*"
```
cxzl25, line 80: In the future, we should be able to support paths without YARN-6929, which does not have buckets.

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
```diff
@@ -0,0 +1,116 @@
...
+        path = path.replace("{{APP_ID}}", appId)
+      case EqualTo("user", user: String) =>
+        path = path.replace("{{USER}}", user)
+      case EqualTo("host", host: String) =>
+        path = path.replace("{{HOST}}", host)
+      case f =>
+        logWarning(s"Unsupported filter: $f")
+    }
+    val globPath = path
+      .replace("{{BUCKET}}", "*") // TODO parallize bucket listing
```
cxzl25, line 96: If an app_id is entered, the bucket can also be calculated in advance to improve list performance.

org.apache.hadoop.yarn.logaggregation.LogAggregationUtils#getRemoteBucketDir
```java
    int bucket = appId.getId() % 10000;
    String bucketDir = String.format("%04d", bucket);
```

cxzl25 (approved):

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
```diff
@@ -0,0 +1,116 @@
...
+    YarnConfiguration.NM_REMOTE_APP_LOG_DIR_SUFFIX,
+    YarnConfiguration.DEFAULT_NM_REMOTE_APP_LOG_DIR_SUFFIX)
+
+  // given a path in hdfs, then get all files under it, supports *
+  private def listFiles(pathStr: String): Array[FileStatus] = {
+    val path = new Path(pathStr)
+    val fs = path.getFileSystem(hadoopConf)
+
+    val start = System.currentTimeMillis()
+    val files = fs.globStatus(path)
```
aajisaka, line 70: In case of Observer NameNode setup, would you call `fs.msync()` before listing files to avoid stale reads?

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnCatalog.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  override def initialize(
+      name: String,
+      caseInsensitiveStringMap: CaseInsensitiveStringMap): Unit = {
+    catalogName = name
+  }
+
+  override def listTables(namespace: Array[String]): Array[Identifier] = {
+    Array(Identifier.of(namespace, "app_logs"))
+  }
```
aajisaka: This catalog is not aware of namespace and it makes `catalog.any_namespace.app_logs` resolve to the same table. We need to at least document the limitation.

aajisaka (commented):
Can we create a sample YARN aggregated log and create a unit test to verify?

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
pan3793: thank you for pointing this out! YARN-6929 landed in Hadoop 3.3.0, I'm reading the branch-3.3 code, so missing this part, will add a TODO here for now.

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
pan3793: good point, thanks for pointing this out

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
pan3793: haven't experienced Observer NameNode ...

according to the API javadocs

> Synchronize client metadata state.
> In some FileSystem implementations such as HDFS metadata synchronization is essential to guarantee consistency of read requests particularly in HA setting.

and seems non-HDFS will throw `UnsupportedOperationException`?

BTW, is `fs.msync()` also expected to be called for listing data files on scanning Hive tables? I think we don't do that ...

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnCatalog.scala
pan3793: it's not intended, will limit it to `catalog.app_logs`

pan3793:
> Can we create a sample YARN aggregated log and create a unit test to verify?

yeah, I plan to do that.

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
aajisaka: > non-HDFS will throw UnsupportedOperationException?

Yes. We need to ignore if it's unsupported. Example in Delta Lake: https://github.com/delta-io/delta/blob/b5e5aecb69cfba6d5a1f4b1ecce81681414051ac/storage/src/main/java/io/delta/storage/HDFSLogStore.java#L171

> BTW, is fs.msync() also expected to be called for listing data files on scanning Hive tables? I think we don't do that ...

We don't usually do because it's new and minor setup, but yes, fs.msync() is also expected to be called.

wForget (approved):
Thanks @pan3793 , LGTM

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
pan3793: I feel this is not much necessary as we don't require strict consistency for log analysis cases, instead, log files may expire after listing, then executors may hit `FileNotFoundException`, I make it to respect `spark.sql.files.ignoreMissingFiles`. I leave a TODO here to mention the `fs.msync()` suggestion for now, we can add it if we see real problems. WDYT? @aajisaka

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
pan3793: added in line 85-86

pan3793:
> Can we create a sample YARN aggregated log and create a unit test to verify?

@aajisaka I collect a set of app logs by setting up a Hadoop cluster and runs 3 Spark applications, and then add some basic unit tests to verify the `app_logs` reading, with and without filters pushdown.

extensions/spark/kyuubi-spark-connector-common/src/test/scala/org/apache/kyuubi/spark/connector/common/LocalSparkSession.scala
```diff
@@ -28,12 +28,12 @@ object LocalSparkSession {
     System.clearProperty("spark.driver.port")
   }
 
-  /** Runs `f` by passing in `sc` and ensures that `sc` is stopped. */
-  def withSparkSession[T](sc: SparkSession)(f: SparkSession => T): T = {
+  /** Runs `f` by passing in `spark` and ensures that `spark` is stopped. */
+  def withSparkSession[T](spark: SparkSession)(f: SparkSession => T): T = {
```
pan3793, line 32: it's naming fix. we should follow the naming policy defined in `spark-shell`, `sc` refers to the `SparkContext`, `spark` refers to `SparkSession`

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
pan3793: the bucket folder layout seems to fit exactly `TABLESAMPLE SYSTEM` introduced by SPARK-55978

extensions/spark/kyuubi-spark-connector-yarn/src/main/scala/org/apache/kyuubi/spark/connector/yarn/YarnLogBatchScan.scala
aajisaka: Agreed. Make sense to me.

aajisaka (approved):

pan3793:
thanks all, merging to master

Issue #7456, opened 2026-05-17
[Umbrella] Ensure GitHub Actions compliance with ASF Policy


dnskr:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues) and found no similar issues.

### Describe the proposal

The Apache Software Foundation has a specific policy for using GitHub Actions, which can be found at [GitHub Actions Policy](https://infra.apache.org/github-actions-policy.html). We need to review existing CI workflows and make the necessary changes to ensure the project remains compliant and avoids potential risks.

#### Task list

- [x] **Review `pull_request_target` usage**
Evaluate the risk in `.github/workflows/labeler.yml`, which uses `pull_request_target` trigger and accesses `GITHUB_TOKEN`.
While this is common for labelers, the ASF policy is strict about "ANY confidential credentials". So, we need to confirm if the labeler can be moved to a safer trigger or if its current usage is acceptable under the "very limited circumstances" mentioned in the policy.

- [x] **Ensure job concurrency levels**
Make sure the job concurrency level is less than or equal to 20 (or 15). Configuring an explicit `max-parallel` value for all workflows with a matrix can be a good solution to respect these concurrency limits.

- [x] **Enable automatic dependency management for GitHub Actions**
All repositories using GitHub Actions must have automatic dependency management in place.
Create `.github/dependabot.yml` to manage updates for GitHub Actions, which will also maintain the SHA hashes required/recommended by the pinning policy.

- [x] **Ensure GitHub Actions in ASF org-level allowlist**
Add [ASF Allowlist Check](https://github.com/apache/infrastructure-actions/blob/d3e898ebd3169ee0a078359ff8434a09cd081375/allowlist-check/README.md) to verify all `uses:` refs in the project workflows are on the ASF Infrastructure [approved allowlist](https://github.com/apache/infrastructure-actions/blob/main/approved_patterns.yml).

dnskr:
### Review `pull_request_target` usage
I have reviewed the [.github/workflows/labeler.yml](https://github.com/apache/kyuubi/blob/de20c20323edeec8e5251873cd7769c775fba706/.github/workflows/labeler.yml) workflow to evaluate its risks against the policy regarding confidential credentials and the `pull_request_target` trigger.

### Conclusion
The current configuration is secure and acceptable.
[...]

PR #7461, opened 2026-05-20
[CI] Set default max-parallel value for all matrices

dnskr:
These changes are needed to ensure compliancy with [ASF GitHub Actions Policy](https://infra.apache.org/github-actions-policy.html).
The default value is set to `15`, which is the recommended job concurrency level.
All other matrices already have specified `max-parallel` value.

Similar PR: https://github.com/apache/airflow/pull/61954

Related to [[Umbrella] Ensure GitHub Actions compliance with ASF Policy #7456](https://github.com/apache/kyuubi/issues/7456).

Review.

No

### DIFF
--- a/.github/workflows/gluten.yml
+++ b/.github/workflows/gluten.yml
@@ -96,6 +96,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
+      max-parallel: 15
       matrix:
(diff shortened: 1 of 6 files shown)


pan3793 (approved):

dnskr:
Thank you for the review!

Merged to master.

PR #6815, opened 2024-11-19
JDBC Engine supports Oracle

naive-zhang:
# :mag: Description
## Issue References 🔗

This pull request fixes #

## Describe Your Solution 🔧

Currently, Kyuubi supports JDBC engines with limited dialects, and I extend the dialects to support Oracle.
* Introduce Oracle support in JDBC Engine
* Adding dialects and tests for Oracle

## Types of changes :bookmark:

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Behavior With This Pull Request :tada:
Supported Oracle in JDBC engine.

#### Related Unit Tests
Add tests of `OperationWithOracleEngineSuite`, `OracleOperationSuite`, `OracleSessionSuite` and `OracleStatementSuite`.

---

# Checklist 📝

- [x] This patch was not authored or co-authored using [Generative Tooling](https://www.apache.org/legal/generative-tooling.html)

**Be nice. Be informative.**

PR #7452, opened 2026-05-15
[KYUUBI #7445] Add AGENTS.md to guide AI coding agents

wangzhigang1999:
### Why are the changes needed?

Resolves #7445. Adds `AGENTS.md` at the repo root — a terse, command-oriented guide for AI coding agents contributing to Kyuubi. Complements `CONTRIBUTING.md`. Content is distilled from review feedback on landed PRs and aligned with sibling Apache projects (Spark, Iceberg, Airflow, DataFusion); the methodology and underlying rule set are linked in #7445.

### How was this patch tested?

Documentation-only change. Every referenced command, Maven profile, script path, class name, and file path was verified against the current `master`. Maven invocations (`-Pfast`, `-DwildcardSuites`, `-Dtest`, `dev/reformat` via Spotless) were run on tiny modules to confirm the documented usage actually executes.

### Was this patch authored or co-authored using generative AI tooling?

Assisted-by: Claude:claude-opus-4-7

PR #7463, opened 2026-05-21
[KYUUBI #6815] fix jdbc oracle engine NUMBER(20,0) type problem

ylq5126:
when oracle column type is NUMBER(20,0), and value is 530924057810571264， query tips Integer cast exception
This PR adds Column precision length check，only length <= 9 can be cast to Integer.

Deployed and Tested in our environment. use kyuubi 1.11.1 jdbc engine proxy oracle, data can be queried normally.

No.

### DIFF
--- a/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/oracle/OracleSchemaHelper.scala
+++ b/externals/kyuubi-jdbc-engine/src/main/scala/org/apache/kyuubi/engine/jdbc/oracle/OracleSchemaHelper.scala
@@ -25,7 +25,7 @@ class OracleSchemaHelper extends SchemaHelper {
   override protected def toTTypeDesc(sqlType: Int, precision: Int, scale: Int): TTypeDesc = {
     sqlType match {
       // case for int, returns NUMERIC type in Oracle JDBC
-      case Types.NUMERIC if scale == 0 =>
+      case Types.NUMERIC if scale == 0 && (precision > 0 && precision <= 9) =>
         super.toTTypeDesc(Types.INTEGER, precision, scale)
       // except for int
       case Types.NUMERIC =>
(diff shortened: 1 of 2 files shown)


wangzhigang1999:
Thanks for the fix. Before diving into the code, could you please tidy up the PR a bit? Since we already have a PR template, it’s best to follow it. Also, if you used any AI tools to assist with development, could you please perform an initial review following the guidelines in #7452?

ylq5126:
> Thanks for the fix. Before diving into the code, could you please tidy up the PR a bit? Since we already have a PR template, it’s best to follow it. Also, if you used any AI tools to assist with development, could you please perform an initial review following the guidelines in #7452?

ok, changed

wangzhigang1999:
I noticed the existing Oracle tests are failing after this change. Oracle `INTEGER` seems to be exposed by ojdbc as `NUMBER(*,0)` / `NUMERIC` with scale 0, and its reported precision may not match `precision <= 9`, so it now falls through to DECIMAL. That changes `resultSet.getObject()` from an integer-like value to `BigDecimal(1)`, causing the assertions in `OracleStatementSuite` to fail.

Would it make sense to add regression tests for the boundary cases, e.g. `INTEGER`, `NUMBER(9,0)`, `NUMBER(10,0)`, and `NUMBER(20,0)`? Maybe we can also consider whether `NUMBER(10..18,0)` should map to BIGINT rather than DECIMAL.

ylq5126:
> I noticed the existing Oracle tests are failing after this change. Oracle `INTEGER` seems to be exposed by ojdbc as `NUMBER(*,0)` / `NUMERIC` with scale 0, and its reported precision may not match `precision <= 9`, so it now falls through to DECIMAL. That changes `resultSet.getObject()` from an integer-like value to `BigDecimal(1)`, causing the assertions in `OracleStatementSuite` to fail.
> 
> Would it make sense to add regression tests for the boundary cases, e.g. `INTEGER`, `NUMBER(9,0)`, `NUMBER(10,0)`, and `NUMBER(20,0)`? Maybe we can also consider whether `NUMBER(10..18,0)` should map to BIGINT rather than DECIMAL.

I will create a new merge request later.

PR #7475, opened 2026-05-25
[CI] Add ASF allowlist check workflow

dnskr:
These changes are needed to ensure compliancy with [ASF GitHub Actions Policy](https://infra.apache.org/github-actions-policy.html).

The workflow verifies all `uses:` refs in a project's workflow files are on the ASF Infrastructure [approved allowlist](https://github.com/apache/infrastructure-actions/blob/d3e898ebd3169ee0a078359ff8434a09cd081375/approved_patterns.yml).
It catches violations before merge, preventing the silent CI failures that occur when an action is not on the org-level allowlist.
See more details [here](https://github.com/apache/infrastructure-actions/blob/d3e898ebd3169ee0a078359ff8434a09cd081375/allowlist-check/README.md).

Related to [[Umbrella] Ensure GitHub Actions compliance with ASF Policy #7456](https://github.com/apache/kyuubi/issues/7456).

Similar PRs:
 - https://github.com/apache/airflow/pull/64450
 - https://github.com/apache/iceberg/pull/15797
 - https://github.com/apache/hive/pull/6417

Review.

No

### DIFF
--- a/.github/workflows/asf-allowlist-check.yml
+++ b/.github/workflows/asf-allowlist-check.yml
@@ -37,7 +37,7 @@ permissions:
 
 jobs:
   asf-allowlist-check:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-slim
     steps:
       - uses: actions/checkout@v6
         with:
--- a/.github/workflows/asf-allowlist-check.yml
+++ b/.github/workflows/asf-allowlist-check.yml
@@ -1,0 +1,45 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


.github/workflows/asf-allowlist-check.yml
```diff
@@ -0,0 +1,45 @@
...
+      - branch-*
+    paths:
+      - ".github/**"
+
+permissions:
+  contents: read
+
+jobs:
+  asf-allowlist-check:
+    runs-on: ubuntu-24.04
```
pan3793: does ubuntu-slim work? it's cheaper

dnskr: Thanks for pointing that out!
I wasn't aware of the `ubuntu-slim` option and copied the runner used in all other workflows.

Regarding pricing, my understanding is that it should be [free for public repositories](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories), but please let me know if I'm missing something here. If you think we should address it for this workflow, then it would be great to check the other workflows like `stale.yml` and `greetings.yml` as well.

pan3793: All Apache projects share a GitHub Actions quota limit. They have asked some projects' PMC that consume a large amount of quota to reduce their usage, and have also encouraged all projects to do so. And yes, we can move some other light workflows to `ubuntu-slim` too. I'm also investigating if we can speed up our CI by caching the engine tgz in a smarter way to improve the cache hit rate instead of triggering download from the asf site, especially from archive.apache.org, which is quite slow and fragile.

dnskr: 👍 Thank you for the detailed explanation. I've changed the runner to `ubuntu-slim`.

pan3793 (approved):

dnskr:
Thank you for the review!

Merged to master.

PR #7483, opened 2026-05-29
Restrict Engine UI proxy targets

aajisaka:

Allow Kyuubi server admin to explicitly configure the proxy destinations or even disable the proxy.

Add the below 2 parameters:

- kyuubi.frontend.rest.engine.ui.proxy.enabled (default: false): Whether to route
  Engine UI traffic via Kyuubi REST frontend proxy. When disabled, the Web UI links
  directly to the Kyuubi engine URL. 
- kyuubi.frontend.rest.engine.ui.proxy.hosts (default: none): A comma-separated list
  of hosts that Engine UI proxy requests can route to when
  `kyuubi.frontend.rest.engine.ui.proxy.enabled` is enabled. Wildcard is allowed.

Built Kyuubi and verified locally
```
# Build Kyuubi
./build/dist --web-ui
# Run Kyuubi Server
./dist/bin/kyuubi start
# Start Kyuubi Engine
./dist/bin/kyuubi-beeline -u 'jdbc:hive2://<your local IP address>:10009/' -n anonymous
# Browse Web UI (http://localhost:10009)
```
Also, added unit tests.

Assisted-by: OpenCode (GPT-5.5)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/EngineUIProxyServlet.scala
```diff
@@ -52,17 +70,125 @@ private[api] class EngineUIProxyServlet extends ProxyServlet with Logging {
...
+      () => registeredEngineUrls(fe))
+  }
+
+  private[api] def registeredEngineUrls(fe: KyuubiRestFrontendService): Set[(String, Int)] = {
+    DiscoveryClientProvider.withDiscoveryClient(fe.getConf) { discoveryClient =>
+      registeredEngineUrls(discoveryClient, fe.getConf.get(HA_NAMESPACE))
+    }
+  }
+
+  private[api] def registeredEngineUrls(
```
pan3793: So the idea is to store the registered engine URL on Zookeeper? Sounds too heavy, maybe have a configurable allow list? something similar to `hadoop.proxyuser.foo.hosts`

aajisaka: > So the idea is to store the registered engine URL on Zookeeper?

In HA setup, yes.

> Sounds too heavy, maybe have a configurable allow list? something similar to hadoop.proxyuser.foo.hosts

Agreed. I'll update the patch to pass a configurable allowlist.

aajisaka: Probably we might need to set `*` by default for compatibility, but I feel it doesn't change the situation. Probably almost all the users rely on the default value.

Instead, can we set the default to none? and if the host is not in the allowlist, just redirect it. The traffic directly goes to the engine URL. What do you think?

pan3793: I think we should have two configs, breaking changes are acceptable as long as we document them in the migration guide.
```
kyuubi.frontend.rest.engine.ui.proxy.enabled  -- disabled by default
kyuubi.frontend.rest.engine.ui.proxy.hosts    -- none by default
```

1. do you think host restriction is sufficient? are port, path necessary?
2. we should at least allow a wildcard for values, e.g., `*`(admins take their own risk to set it), `hadoop*.example.org`, not sure if regex should be used (not much readable, but much more powerful)

aajisaka: > breaking changes are acceptable as long as we document them in the migration guide.

Thanks. Good to know.

> do you think host restriction is sufficient? are port, path necessary?

I think host restriction is sufficient.

> we should at least allow a wildcard for values

+1

kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/EngineUIProxyServlet.scala
```diff
@@ -18,51 +18,92 @@
...
+  private def isAllowedHost(host: String): Boolean =
+    proxyEnabled && {
+      val normalizedHost = EngineUIProxyServlet.normalizeHost(host)
+      allowedHostPatterns.exists(_.matcher(normalizedHost).matches())
+    }
+
+  private def allowedTarget(requestURI: String): Option[(String, Int)] =
+    extractTargetAddress(requestURI).filter(target => isAllowedHost(target._1))
+
   private val r = "^/engine-ui/([^/:]+):(\\d+)/?.*".r
```
wangzhigang1999: Parsing host and port directly with this regex is insecure — the host can be spoofed. I did a quick verification, and perhaps we should use a more robust method to parse the host and port.

```java
import java.net.URI;
import java.net.URL;
import java.util.regex.Matcher;
import java.util.regex.Pattern;

public class EngineUIProxyRegexPoc {
    public static void main(String[] a) throws Exception {
        // same regex as EngineUIProxyServlet.scala:84
        Pattern r = Pattern.compile("^/engine-ui/([^/:]+):(\\d+)/?.*");

        // attacker URI: smuggle the real target after '@'
        String uri = "/engine-ui/spark-x@100.100.100.200:80/latest/meta-data/";

        Matcher m = r.matcher(uri);
        m.matches();
        String host = m.group(1);                                   // host checked against allowlist
        String target = new URL("http", host, 80, "/").toString();  // same as PR
        String realHost = URI.create(target).getHost();             // host Jetty actually connects to

        System.out.println("checked host : " + host);               // spark-x@100.100.100.200
        System.out.println("connect host : " + realHost);           // 100.100.100.200
        System.out.println("bypass?      : " + !host.equals(realHost)); // true
    }
}
```

wangzhigang1999: >   private val r = "^/engine-ui/([^/:]+):(\\d+)/?.*".r

In fact, I want to refer to this line of code.

wangzhigang1999: Idea: drop the regex and let `URI` own the parse, so the validated host is exactly the one Jetty connects to. Roughly:

```scala
val rest = requestURI.stripPrefix("/engine-ui/")
val authority = rest.indexOf('/') match { case -1 => rest; case i => rest.substring(0, i) }
val u = URI.create(s"http://$authority")            // may throw -> deny
// reject smuggling / invalid targets
if (u.getUserInfo != null || u.getHost == null || u.getPort < 0) return denied
// build the target ONLY from the validated host/port, never from the raw string
// (getHost already brackets IPv6 literals; avoids the deprecated java.net.URL ctor)
val target = s"http://${u.getHost}:${u.getPort}$path$query"
```

aajisaka: Nice catch! I dropped the regex pattern and let `URI` own the parse.

wangzhigang1999:
Looks good! Since this flips the proxy default, might be worth a quick migration-guide note for existing users.

aajisaka:
Thank you! Added a note in the migration guide.

wangzhigang1999 (approved):

aajisaka:
Merged. Thank you @pan3793 and @wangzhigang1999 !

PR #7485, opened 2026-05-31
[CI] Run "Docs" workflow for documentation changes only

dnskr:
These changes are needed to prevent unnecessary `Docs` workflow runs by ensuring they only trigger when files in `docs/**` or the workflow itself are changed.

Also updated `.github/workflows/master.yml` and `.github/workflows/python.yml` to ignore changes in `.github/workflows/docs.yml`, preventing redundant triggers of the main test suites when only documentation workflow updated.

Review/CI

No

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,6 +21,9 @@ on:
   pull_request:
     branches:
       - master
+    paths:
+      - 'docs/**'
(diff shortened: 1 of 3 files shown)


pan3793 (approved):

dnskr:
Thank you for the review!

Merged to master.

PR #7489, opened 2026-06-02
[KYUUBI #7488][SERVER] Fix canceled batch reporting stale RUNNING application state

### LINKED ISSUE #7488: [Bug] Canceled batch keeps reporting a stale RUNNING application state
### Code of Conduct

### Search before asking

### Describe the bug

When a **RUNNING** batch is canceled via `DELETE /api/v1/batches/{batchId}` (or the cancel button in the Web UI), the batch operation state correctly transitions to `CANCELED` and the Spark application is actually killed — yet the batch keeps reporting `appState = RUNNING` forever. The REST API permanently returns `state=CANCELED` together with `appState=RUNNING`, which is contradictory and misleading (e.g. the Web UI shows a canceled batch whose application still looks alive).

The application really is gone: the driver Pod and driver Service are deleted, and the `engine-ui` reverse proxy to the batch's `appUrl` returns HTTP 502. So this is purely stale metadata, not an orphaned/leaked application.

**Steps to reproduce**

1. Submit a long-running Spark batch in cluster mode and wait until it reaches `RUNNING` (so `appState=RUNNING` and `appUrl` are populated).
2. `DELETE /api/v1/batches/{batchId}` — the response is `CloseBatchResponse(success=true)`.
3. `GET /api/v1/batches/{batchId}`.

**Actual result** — `appState` stays `RUNNING` indefinitely, even though the driver Pod/Service are already gone:

```json
{
  "id": "723d0f90-8a0d-49eb-9b5d-8a7a84e35787",
  "name": "claude-sparkpi-resident",
  "state": "CANCELED",
  "appState": "RUNNING",
  "appId": "spark-d1a76dfdc1cd4836aaabebc8b09d616d",
  "appUrl": "http://...-driver-svc.<ns>.svc:4040",
  "endTime": 1780411775395
}
```

**Expected result** — once a batch is canceled, its application has been killed, so `appState` should report a terminated state (e.g. `KILLED`) rather than a stale `RUNNING`.

### Affects Version(s)

master (verified on commit `7067644e1`) and 1.10.1. The affected code path exists in all released versions; no branch currently fixes it.

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

**Root cause**

In `BatchJobSubmission.close()`, the kill-success branch sets the operation state to `CANCELED` and persists metadata via `updateBatchMetadata()` without re-polling the application info. `updateBatchMetadata()` writes `engineState = getAppState(state, appInfo.state)`, where `appInfo.state` is the last-polled value (`RUNNING`). `getAppState(...)` only coerces a non-terminated application state to `UNKNOWN` when `opState == ERROR`; it leaves `CANCELED` untouched and returns `RUNNING` as-is. Because status polling stops once the operation is closed, the stale `RUNNING` is persisted and reported forever.

The same gap applies to the peer-instance-closed cancellation path (`peerInstanceClosed`), which also reaches `CANCELED` carrying the last-polled application state. The kill-*failed* branch is unaffected, since it re-polls and maps `KILLED`/`NOT_FOUND` appropriately. The earlier related fixes #6722 and #6668 do not cover this cancel path.

**Proposed fix**

When the operation state is `CANCELED` and the application state is not yet terminated, reconcile it to `KILLED` (the application has been killed), mirroring the existing `ERROR -> UNKNOWN` handling in `getAppState`.

wangzhigang1999:

Closes #7488.

Canceling a RUNNING batch leaves its reported `appState` stuck at `RUNNING` forever: the operation state correctly becomes `CANCELED` and the application is killed, but `getAppState` only reconciles a non-terminal application state for `ERROR` (to `UNKNOWN`), not `CANCELED`, so the stale `RUNNING` is persisted and reported by the REST API. This reconciles `CANCELED` + non-terminal application state to `KILLED`, mirroring the existing `ERROR` handling. Affects master and 1.10.1.

Added an assertion to the existing `BatchesResourceSuite` metrics test (v1 and v2): it cancels a RUNNING batch and now asserts the reported `appState` is terminated, which fails on the old behavior (where it stayed `RUNNING`). 

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
@@ -437,6 +437,17 @@ class BatchJobSubmission(
             // kill success and we can change state safely
             // note that, the batch operation state should never be closed
             setState(OperationState.CANCELED)
+            // re-poll the resource manager so the persisted application state reflects the kill
+            // rather than the last-seen (e.g. RUNNING) state; otherwise polling stops after close
+            // and the stale state would be reported by the REST API forever.
+            _applicationInfo = currentApplicationInfo()
+            _applicationInfo.filterNot(app => ApplicationState.isTerminated(app.state)).foreach {
+              app =>
+                withOperationLog(info(s"Batch $batchId state is $state," +
+                  s" but the application state is ${app.state}" +
+                  " and not terminated after kill succeeds, set to KILLED."))
+                _applicationInfo = Some(app.copy(state = ApplicationState.KILLED))
+            }
             updateBatchMetadata()
           } else if (killMessage._1) {
             // we can not change state safely
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/BatchesResourceSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/server/api/v1/BatchesResourceSuite.scala
@@ -37,7 +37,7 @@ import org.apache.kyuubi.client.util.BatchUtils
 import org.apache.kyuubi.client.util.BatchUtils._
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.config.KyuubiConf._
-import org.apache.kyuubi.engine.{ApplicationInfo, ApplicationManagerInfo, KyuubiApplicationManager}
+import org.apache.kyuubi.engine.{ApplicationInfo, ApplicationManagerInfo, ApplicationState, KyuubiApplicationManager}
 import org.apache.kyuubi.engine.spark.SparkBatchProcessBuilder
 import org.apache.kyuubi.metrics.{MetricsConstants, MetricsSystem}
 import org.apache.kyuubi.operation.{BatchJobSubmission, OperationState}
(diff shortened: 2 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/operation/BatchJobSubmission.scala
```diff
@@ -195,6 +195,13 @@ class BatchJobSubmission(
       withOperationLog(error(s"Batch $batchId state is $opState," +
         s" but the application state is $appState and not terminated, set to UNKNOWN."))
       ApplicationState.UNKNOWN
+    } else if (opState == OperationState.CANCELED && !ApplicationState.isTerminated(appState)) {
+      // A canceled batch has had its application killed, so the last-polled application state
+      // (e.g. RUNNING) is stale. Without reconciling it, this stale state would be persisted and
+      // reported by the REST API forever, since polling stops once the operation is closed.
+      withOperationLog(info(s"Batch $batchId state is $opState," +
+        s" but the application state is $appState and not terminated, set to KILLED."))
+      ApplicationState.KILLED
```
pan3793: the ERROR => UNKNOWN is a defensive inference, how do we confidently infer KILLED  for CANCELED op? is it possible to query it from the resource manager?

wangzhigang1999: Good point. Querying the RM helps, but it does not fully fix the stale metadata issue by itself: application kill is asynchronous, so an immediate RM query may still report RUNNING/PENDING. Since Kyuubi stops polling once the batch operation is closed, persisting that non-terminal state would make the REST API report it forever.

I think UNKNOWN is not a good fallback here either, because UNKNOWN is not a terminal ApplicationState in Kyuubi, so it would still leave clients with a non-terminal state that will never converge.

So the intended logic is: after killApplication succeeds, prefer the RM state if it has already converged to a terminal state; otherwise, for a CANCELED operation whose last observed app state is still non-terminal, reconcile it to KILLED instead of persisting stale RUNNING/PENDING. The KILLED fallback is only used after the kill-success path, not for arbitrary canceled metadata. I can re-poll once in the kill-success branch before applying the fallback and clarify the comment.

WDYT？

wangzhigang1999:
Discovered while building the batch management UI: the UI faithfully renders this state, so a canceled batch would show RUNNING forever.

pan3793 (approved):

wangzhigang1999:
thanks, merged to master

PR #7496, opened 2026-06-04
[KYUUBI #7495][KUBERNETES] Avoid overlong pod log directory names

### LINKED ISSUE #7495: [Bug] Spark on Kubernetes generated pod names can exceed kubelet pod log directory name length
### Code of Conduct

### Search before asking

### Describe the bug

When Kyuubi submits Spark applications to Kubernetes, `KubernetesUtils` generates the Spark driver pod name and executor pod name prefix from the Spark application name and engine reference id.

The current length check only considers the Kubernetes pod name limit. For example, the driver pod name is allowed up to 253 characters, and the executor pod name prefix reserves only Spark's executor suffix.

However, kubelet creates the pod log directory with the following format:

```text
namespace_podName_podUID
```

See Kubernetes' `BuildPodLogsDirectory` implementation:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/helpers.go#L223

As a result, a pod name that is valid for Kubernetes can still make the kubelet log directory name too long after the namespace, separators, and pod UID are appended. In that case, Spark submission can succeed far enough to create the pod, but the kubelet fails when creating the pod log directory on the node because the directory name exceeds the filesystem name length limit.

Kyuubi should take the kubelet pod log directory format into account when generating Spark driver pod names and executor pod name prefixes for Kubernetes.

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk
N/A
```

### Kyuubi Engine Log Output

```logtalk
N/A
```

```yaml
spark.master: k8s://...
spark.submit.deployMode: cluster
# The issue is easier to reproduce with a long spark.app.name and/or a long Kubernetes namespace.
```

```yaml
N/A
```

### Additional context

The fix should preserve user-specified `spark.kubernetes.driver.pod.name` and `spark.kubernetes.executor.podNamePrefix`, but generated names should leave enough room for the namespace, the two `_` separators, the pod UID, and Spark's executor suffix where applicable.

ruanwenjun:

Kyuubi currently limits generated Spark driver pod names and executor pod name prefixes only against the Kubernetes pod name limit.

However, kubelet creates pod log directories with the `namespace_podName_podUID` format. When the generated pod name is close to the Kubernetes pod name limit, the log directory name can exceed the Linux path component limit and kubelet fails to create it.

This patch makes the generated driver pod name and executor pod name prefix reserve the kubelet pod log directory budget. Closes #7495.

- `env PATH="/tmp/kyuubi-black-22.3.0/bin:$PATH" dev/reformat`
- `git diff --check HEAD~1..HEAD`
- `build/mvn test -pl kyuubi-server -am -Dtest=none -DwildcardSuites=org.apache.kyuubi.util.KubernetesPodNameSuite`
- `build/mvn test -pl externals/kyuubi-spark-sql-engine -am -Dtest=none -DwildcardSuites=org.apache.kyuubi.engine.spark.SparkSQLEngineSuite`

Assisted-by: OpenAI Codex (GPT-5)

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -230,9 +230,14 @@ object SparkSQLEngine extends Logging {
 
   private val sparkSessionCreated = new AtomicBoolean(false)
 
-  // Kubernetes pod name max length - '-exec-' - Int.MAX_VALUE.length
-  // 253 - 10 - 6
-  val EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH = 237
+  final private val POD_NAME_MAX_LENGTH = 253
+  final private val POD_UID_MAX_LENGTH = 36
+  final private val POD_LOGS_DIRECTORY_SEPARATOR_LENGTH = 2
+  final private val EXECUTOR_POD_NAME_RESERVED_LENGTH =
+    "-exec-".length + Int.MaxValue.toString.length
+
+  val EXECUTOR_POD_NAME_PREFIX_MAX_LENGTH =
+    executorPodNamePrefixMaxLength(KUBERNETES_NAMESPACE.defaultValStr)
 
   SignalRegister.registerLogger(logger)
   setupConf()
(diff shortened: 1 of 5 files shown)


pan3793:
this makes sense, and maybe we want a config to always rewrite the name for safety (this can be implemented in a dedicated PR)

ruanwenjun:
@pan3793 Yes, Kyuubi provides the following configurations, and they work as expected:

```
kyuubi.kubernetes.spark.forciblyRewriteDriverPodName.enabled
kyuubi.kubernetes.spark.forciblyRewriteExecutorPodNamePrefix.enabled
```

When enabled, Kyuubi rewrites all pod names and therefore avoids the problem. However, when these options are disabled, there is a bug in the current pod name length validation logic that can result in invalid pod names being accepted.

pan3793:
oh, we already have this feature.

pan3793 (approved):

ruanwenjun:
Is this PR ready to merge? cc@pan3793

pan3793:
thanks, merged for 1.12.0/1.11.2/1.10.4

PR #7503, opened 2026-06-08
[DOC] Document kyuubi.session.local.dir.allow.list is only for Spark engine

aajisaka:

Document kyuubi.session.local.dir.allow.list is currently Spark engine only.

Doc only change

No


cxzl25 (approved):

aajisaka:
Merged. Thank you @cxzl25 !

PR #7516, opened 2026-06-16
[KYUUBI #7515] Handle Spark variant type in metadata

### LINKED ISSUE #7515: [Bug] Spark VARIANT result metadata fails with "Unrecognized type name: variant"
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct).

### Search before asking

- [x] I have searched the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

Kyuubi fails when returning result set metadata for a Spark SQL result containing a `VARIANT` column.

The SQL operation itself completes successfully, and row fetching can succeed, but the REST metadata endpoint fails when Kyuubi converts the Spark result schema to HiveServer2/Thrift metadata.

The failure comes from:

```text
org.apache.kyuubi.engine.spark.schema.SchemaHelper$.toTTypeId(SchemaHelper.scala:58)
java.lang.IllegalArgumentException: Unrecognized type name: variant
```

`SchemaHelper.toTTypeId` handles common Spark SQL types but does not appear to handle Spark 4 `VariantType`. Spark 4.0+ supports `VariantType`, and `parse_json(...)` returns a `VARIANT`.

### Steps to reproduce

Run Kyuubi with a Spark 4.x engine, then execute a query that returns a `VARIANT` column:

```sql
SELECT parse_json('{"a":1,"b":0.8}') AS v;
```

Then request result set metadata for the completed operation, for example through the Kyuubi Web UI or REST API:

```text
GET /api/v1/operations/<operationHandle>/resultsetmetadata
```

### Expected behavior

Kyuubi should return result set metadata successfully.

If HiveServer2/Thrift has no native `VARIANT` type, Kyuubi should expose Spark `VariantType` as a compatible metadata type, such as `STRING_TYPE` or another appropriate fallback, instead of failing the metadata request.

### Actual behavior

The query reaches `FINISHED_STATE`, but metadata retrieval fails:

```text
Error getting result set metadata for operation handle <operationHandle>
org.apache.kyuubi.KyuubiSQLException: Unrecognized type name: variant
Caused by: java.lang.IllegalArgumentException: Unrecognized type name: variant
    at org.apache.kyuubi.engine.spark.schema.SchemaHelper$.toTTypeId(SchemaHelper.scala:58)
    at org.apache.kyuubi.engine.spark.schema.SchemaHelper$.toTTypeDesc(SchemaHelper.scala:77)
    at org.apache.kyuubi.engine.spark.schema.SchemaHelper$.toTColumnDesc(SchemaHelper.scala:87)
    at org.apache.kyuubi.engine.spark.schema.SchemaHelper$.toTTableSchema(SchemaHelper.scala:95)
```

Image

### Affects Version(s)

Observed on Kyuubi `1.11.1` with Spark `4.1.1`.

The same `SchemaHelper.toTTypeId` mapping appears to be present on current `master`, so `master` may also be affected.

### Kyuubi Server Log Output

```logtalk
Processing query[<operationHandle>]: RUNNING_STATE -> FINISHED_STATE

GET /api/v1/operations/<operationHandle>/rowset ... status=200

ERROR org.apache.kyuubi.server.api.ApiUtils:
Error getting result set metadata for operation handle <operationHandle>
org.apache.kyuubi.KyuubiSQLException: Unrecognized type name: variant
Caused by: java.lang.IllegalArgumentException: Unrecognized type name: variant
    at org.apache.kyuubi.engine.spark.schema.SchemaHelper$.toTTypeId(SchemaHelper.scala:58)
```

### Kyuubi Engine Log Output

No Spark query execution failure was observed. The query completed successfully; the failure occurs during Kyuubi result set metadata conversion.

```yaml
kyuubi.engine.type: SPARK_SQL
kyuubi.frontend.protocols: THRIFT_BINARY,REST
```

The issue should reproduce without extra engine configuration using:

```sql
SELECT parse_json('{"a":1,"b":0.8}') AS v;
```

### Additional context

Spark 4.0 introduced `VariantType`. Kyuubi already stringifies non-primitive row values in rowset generation, but result metadata conversion currently throws before clients can render metadata for a result containing `VARIANT`.

A likely fix is to add Spark `VariantType` handling in Kyuubi's Spark schema conversion path.

ericyangliu:

Fixes #7515.

Spark 4 introduced `VariantType`. Kyuubi's Spark schema conversion did not handle it when building HiveServer2 result-set metadata, so `GetResultSetMetadata` could fail with `Unrecognized type name: variant` even when the query itself completed.

This patch maps Spark `VariantType` to the existing HiveServer2 `STRING_TYPE` representation for query result-set metadata, and reports `java.sql.Types.OTHER` for `DatabaseMetaData.getColumns` `DATA_TYPE`.

- Added `SchemaHelperSuite` coverage for `VariantType` mapping to `TTypeId.STRING_TYPE` and `java.sql.Types.OTHER`.
- Added Spark 4 regression coverage for `SELECT parse_json('{"a":1}') AS v` result-set metadata.
- Added Spark 4 regression coverage for `SELECT *` from a dummy table with a `VARIANT` column populated via `parse_json`.
- Added Spark `getColumns` coverage for reporting `VARIANT` as `java.sql.Types.OTHER`.
- `dev/reformat`
- `env JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk@17/bin:$PATH ./build/mvn test -pl externals/kyuubi-spark-sql-engine -Dtest=none -DwildcardSuites=org.apache.kyuubi.engine.spark.schema.SchemaHelperSuite -Pspark-4.1 -Pscala-2.13`
- `env JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk@17/bin:$PATH ./build/mvn test -pl externals/kyuubi-spark-sql-engine -Dtest=none -Dsuites=org.apache.kyuubi.engine.spark.operation.SparkOperationSuite -Pspark-4.1 -Pscala-2.13`

Assisted-by: OpenAI Codex

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
@@ -32,6 +32,11 @@ object SchemaHelper {
    */
   final val TIMESTAMP_NTZ = "TimestampNTZType$"
 
+  /**
+   * Spark 4.0.0 DataType VariantType's class name.
+   */
+  final val VARIANT = "VariantType$"
+
   def toTTypeId(typ: DataType): TTypeId = typ match {
     case NullType => TTypeId.NULL_TYPE
     case BooleanType => TTypeId.BOOLEAN_TYPE
@@ -46,6 +51,7 @@ object SchemaHelper {
     case DateType => TTypeId.DATE_TYPE
     case TimestampType => TTypeId.TIMESTAMP_TYPE
     case ntz if ntz.getClass.getSimpleName.equals(TIMESTAMP_NTZ) => TTypeId.TIMESTAMP_TYPE
+    case variant if variant.getClass.getSimpleName.equals(VARIANT) => TTypeId.STRING_TYPE
     case BinaryType => TTypeId.BINARY_TYPE
     case CalendarIntervalType => TTypeId.STRING_TYPE
     case _: DayTimeIntervalType => TTypeId.INTERVAL_DAY_TIME_TYPE
@@ -112,6 +118,7 @@ object SchemaHelper {
     case DateType => java.sql.Types.DATE
     case TimestampType => java.sql.Types.TIMESTAMP
     case ntz if ntz.getClass.getSimpleName.equals(TIMESTAMP_NTZ) => java.sql.Types.TIMESTAMP
+    case variant if variant.getClass.getSimpleName.equals(VARIANT) => java.sql.Types.OTHER
     case BinaryType => java.sql.Types.BINARY
     case _: ArrayType => java.sql.Types.ARRAY
     case _: MapType => java.sql.Types.JAVA_OBJECT
--- a/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
+++ b/externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkOperationSuite.scala
@@ -32,7 +32,7 @@ import org.apache.spark.sql.types._
 
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.engine.spark.WithSparkSQLEngine
-import org.apache.kyuubi.engine.spark.schema.SchemaHelper.TIMESTAMP_NTZ
+import org.apache.kyuubi.engine.spark.schema.SchemaHelper.{TIMESTAMP_NTZ, VARIANT}
 import org.apache.kyuubi.engine.spark.util.SparkCatalogUtils
 import org.apache.kyuubi.jdbc.hive.KyuubiStatement
 import org.apache.kyuubi.operation.{HiveMetadataTests, SparkQueryTests}
@@ -97,6 +97,9 @@ class SparkOperationSuite extends WithSparkSQLEngine with HiveMetadataTests with
     if (SPARK_ENGINE_RUNTIME_VERSION >= "3.4") {
       schema = schema.add("c20", "timestamp_ntz", nullable = true, "20")
     }
+    if (SPARK_ENGINE_RUNTIME_VERSION >= "4.0") {
+      schema = schema.add("c21", "variant", nullable = true, "21")
+    }
 
     val ddl =
       s"""
@@ -135,7 +138,8 @@ class SparkOperationSuite extends WithSparkSQLEngine with HiveMetadataTests with
           STRUCT,
           OTHER,
           OTHER,
-          TIMESTAMP)
+          TIMESTAMP,
+          OTHER)
 
         var pos = 0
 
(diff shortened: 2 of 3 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -112,6 +118,7 @@ object SchemaHelper {
     case DateType => java.sql.Types.DATE
     case TimestampType => java.sql.Types.TIMESTAMP
     case ntz if ntz.getClass.getSimpleName.equals(TIMESTAMP_NTZ) => java.sql.Types.TIMESTAMP
+    case variant if variant.getClass.getSimpleName.equals(VARIANT) => java.sql.Types.VARCHAR
```
pan3793: is this consistent with other product implementations? e.g., upstream Hive JDBC driver, Databricks JDBC driver, Trino JDBC driver.

ericyangliu: @pan3793 Thanks for the quick review!

I checked the JDBC metadata behavior against other implementations and updated this path.

  For `DatabaseMetaData.getColumns`, `VariantType` now reports `java.sql.Types.OTHER`, which is
  consistent with Databricks JDBC’s `VARIANT` mapping and avoids presenting semi-structured data as
  plain `VARCHAR`. Trino’s JSON-like types also use non-`VARCHAR` JDBC type codes. Hive JDBC does not
  have a native `VARIANT` mapping.

  I kept `TTypeId.STRING_TYPE` for HiveServer2 result-set metadata because the Thrift enum has no
  native `VARIANT`/`OTHER` type, and Kyuubi still needs a concrete type that clients can fetch/render.

Please let me know if you have other concerns.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/SchemaHelper.scala
```diff
@@ -112,6 +118,7 @@ object SchemaHelper {
     case DateType => java.sql.Types.DATE
     case TimestampType => java.sql.Types.TIMESTAMP
     case ntz if ntz.getClass.getSimpleName.equals(TIMESTAMP_NTZ) => java.sql.Types.TIMESTAMP
+    case variant if variant.getClass.getSimpleName.equals(VARIANT) => java.sql.Types.OTHER
```
pan3793, line 121: VARIANT goes to the default branch even without this change, but it's good to write it explicitly.

pan3793 (approved):

pan3793:

ericyangliu:
@pan3793 Thank you for the review and guidance!

Issue #1091, opened 2021-09-13
[FEATURE] Support renew delegation tokens of multiple Hive metastore clusters


zhouyifan279:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Enhance `HiveDelegationTokenProvider` to renew delegation tokens of multiple Hive metastore clusters.

### Motivation

Kyuubi supports [setting Hive metastore uris in JDBC connection URL](https://kyuubi.readthedocs.io/en/latest/deployment/hive_metastore.html#via-jdbc-connection-url).
But current `HiveDelegationTokenProvider` implemented in #1009 can only renew delegation tokens of Hive metastore defined in hive-site.xml.

For users who frequently access different Hive metastore clusters, the Kyuubi engines visiting non default Hive metastore server are still limited by Hive token's expiration time.

### Describe the solution

Introduce a Kyuubi configuration `kyuuib.credentials.hive.metastore.uris` to specify extra Hive metastore URIs.  
URIs should be specified in format "thrift://host01:port01,thrift://host02:port02;thrift://host11:port11,thrift://host12:port12", where "," separates URIs belonging to same metastore cluster, ";" separates URIs of different metastore clusters.

 `HiveDelegationTokenProvider`  will create a `HiveMetaStoreClient` for each Hive metastore URIs and periodically renew the delegation tokens.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

Narcasserun:
hi Brother, how is this going? I have a similar problem

PR #7518, opened 2026-06-18
[KYUUBI #7433][SPARK] Preserve Hive delegation tokens with non-empty service

### LINKED ISSUE #7433: [Bug] SparkTBinaryFrontendService#addHiveToken silently drops Hive delegation tokens with non-empty service field (multi-HMS scenario)
### Code of Conduct

### Search before asking

### Describe the bug

In environments with **multiple Hive Metastore (HMS) instances backed by different Kerberos principals**, Spark engines launched by Kyuubi cannot authenticate against non-default HMS instances. Catalog operations against those HMSes fail with `DIGEST-MD5: IO error acquiring password`.

The root cause is in the engine-side credential push path, `SparkTBinaryFrontendService#addHiveToken`. The current logic assumes a **single HMS** and selects only tokens whose `getService()` is empty:

```scala
val newToken = newTokens
  .find { case (uris, token) =>
    val matched = uris.toString.split(",").exists(uriSet.contains) &&
      token.getService == new Text()
    ...
  }
```

Any Hive delegation token that has its `service` field populated — the natural way to disambiguate per-HMS tokens that share a single `Credentials` — is **silently dropped**. As a result, tokens for non-default HMSes never reach the engine UGI even when a (custom) provider has issued them correctly.

### Affects Version(s)

1.11.1

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk
Authentication fails on the engine side because the token for HMS B never reached the engine UGI:


Caused by: MetaException(message:Could not connect to meta store using any of the URIs provided.
  Most recent failure: org.apache.thrift.transport.TTransportException:
  Peer indicated failure: DIGEST-MD5: IO error acquiring password
    at org.apache.thrift.transport.TSaslTransport.receiveSaslMessage(...)
    at org.apache.thrift.transport.TSaslTransport.open(...)
    at org.apache.thrift.transport.TSaslClientTransport.open(...)
    at org.apache.hadoop.hive.thrift.client.TUGIAssumingTransport$1.run(...)
    at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.open(...)
    at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.<init>(...)
    at org.apache.hadoop.hive.metastore.RetryingMetaStoreClient.<init>(...)
```

```yaml

```

```yaml
A Spark session configured with **two Iceberg catalogs, each backed by a separate HMS with different Kerberos principals** — a common setup for domain-partitioned warehouses:


# Catalog A — HMS A
spark.sql.catalog.warehouse_a                                          org.apache.iceberg.spark.SparkCatalog
spark.sql.catalog.warehouse_a.type                                     hive
spark.sql.catalog.warehouse_a.uri                                      thrift://hms-a.example.com:9083
spark.sql.catalog.warehouse_a.warehouse                                s3a://warehouse-a/warehouse
spark.sql.catalog.warehouse_a.io-impl                                  org.apache.iceberg.aws.s3.S3FileIO
spark.sql.catalog.warehouse_a.hadoop.hive.metastore.uris               thrift://hms-a.example.com:9083
spark.sql.catalog.warehouse_a.hadoop.hive.metastore.sasl.enabled       true
spark.sql.catalog.warehouse_a.hadoop.hive.metastore.kerberos.principal hive-a/_HOST@EXAMPLE.COM
# Tells the HMS client which delegation token to pick from UGI credentials.
# Must match the alias used when the token was added to Credentials (= instance name).
spark.sql.catalog.warehouse_a.hadoop.hive.metastore.token.signature    warehouse_a

# Catalog B — HMS B (different Kerberos principal)
spark.sql.catalog.warehouse_b                                          org.apache.iceberg.spark.SparkCatalog
spark.sql.catalog.warehouse_b.type                                     hive
spark.sql.catalog.warehouse_b.uri                                      thrift://hms-b.example.com:9083
spark.sql.catalog.warehouse_b.warehouse                                s3a://warehouse-b/warehouse
spark.sql.catalog.warehouse_b.io-impl                                  org.apache.iceberg.aws.s3.S3FileIO
spark.sql.catalog.warehouse_b.hadoop.hive.metastore.uris               thrift://hms-b.example.com:9083
spark.sql.catalog.warehouse_b.hadoop.hive.metastore.sasl.enabled       true
spark.sql.catalog.warehouse_b.hadoop.hive.metastore.kerberos.principal hive-b/_HOST@EXAMPLE.COM
spark.sql.catalog.warehouse_b.hadoop.hive.metastore.token.signature    warehouse_b
```

### Additional context

- This bug is layered *under* #1091 — fixing the provider does not eliminate the engine-side drop.

Sunwoo-Shin:

Closes #7433.

`SparkTBinaryFrontendService#addHiveToken` keeps only the Hive delegation token whose `service` field is empty (the one `HiveMetaStoreClient` selects by default) and silently drops every Hive token whose `service` is non-empty.

A Hive delegation token gets a non-empty `service` when it is bound to a specific metastore via `hive.metastore.token.signature` (the signature is stored in the token service). When an engine talks to multiple Hive metastores that use different Kerberos principals — e.g. two Iceberg catalogs, each backed by its own HMS — each metastore produces its own signature-bound token. Because these tokens are dropped before reaching the engine UGI, the engine fails to authenticate against the non-default metastore with `DIGEST-MD5: IO error acquiring password`.

This is the engine-side counterpart of #1091 (renewing delegation tokens for multiple Hive metastore clusters): even when the server pushes per-metastore tokens, the engine drops them.

Affected versions: 1.11.1.

This change partitions the incoming Hive tokens by their `service` field:

- Tokens with a non-empty `service` are added to the engine credentials keyed by their alias, reusing the same issue-date downgrade guard the default path already applies.
- The existing single-metastore URI matching now runs only over the default (empty-`service`) tokens, so behavior for the common single-HMS case is unchanged.
- The `No matching Hive token found ...` warning is emitted only when a default-`service` token was actually expected, to avoid noise in metastore deployments that rely solely on signature-bound tokens.

Added unit tests in `SparkTBinaryFrontendServiceSuite` (the token-merging logic was extracted into `mergeHiveTokens` so it can be exercised without a `SparkContext`):

- signature-bound tokens for multiple metastores are preserved, keyed by alias;
- a signature-bound token with an earlier issue date is ignored, and a later one replaces the existing token;
- the existing single-metastore matching for default-`service` tokens is unchanged;
- signature-bound tokens are added without disturbing the default-`service` path.

```
build/mvn test -pl externals/kyuubi-spark-sql-engine -am \
  -Dtest=none \
  -DwildcardSuites=org.apache.kyuubi.engine.spark.SparkTBinaryFrontendServiceSuite
```

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkTBinaryFrontendService.scala
@@ -139,6 +139,31 @@ object SparkTBinaryFrontendService extends Logging {
       oldCreds: Credentials,
       updateCreds: Credentials): Unit = {
     val metastoreUris = hiveConf(sc.hadoopConfiguration).getTrimmed("hive.metastore.uris", "")
+    mergeHiveTokens(metastoreUris, newTokens, oldCreds, updateCreds)
+  }
+
+  private[spark] def mergeHiveTokens(
+      metastoreUris: String,
+      newTokens: Map[Text, Token[_ <: TokenIdentifier]],
+      oldCreds: Credentials,
+      updateCreds: Credentials): Unit = {
+    // A Hive delegation token whose service field is non-empty is bound to a specific metastore
+    // via `hive.metastore.token.signature`, which is stored in the token service. When an engine
+    // accesses several metastores with different Kerberos principals, each metastore yields one
+    // such token. The single-metastore matching below only keeps the token whose service is empty,
+    // so these signature-bound tokens were silently dropped and never reached the engine UGI. Add
+    // them keyed by their alias so they survive.
+    val (signatureBoundTokens, defaultServiceTokens) = newTokens.partition {
+      case (_, token) => token.getService != new Text()
+    }
+    signatureBoundTokens.foreach { case (alias, token) =>
+      val oldToken = oldCreds.getToken(alias)
+      if (oldToken == null || KyuubiHadoopUtils.compareIssueDate(token, oldToken) > 0) {
+        updateCreds.addToken(alias, token)
+      } else {
+        warn(s"Ignore Hive token with earlier issue date: $token")
+      }
+    }
 
     // `HiveMetaStoreClient` selects the first token whose service is "" and kind is
     // "HIVE_DELEGATION_TOKEN" to authenticate.
@@ -148,11 +173,11 @@ object SparkTBinaryFrontendService extends Logging {
       }
 
     if (metastoreUris.nonEmpty && oldAliasAndToken.isDefined) {
-      // Each entry of `newTokens` is a <uris, token> pair for a metastore cluster.
+      // Each entry of `defaultServiceTokens` is a <uris, token> pair for a metastore cluster.
       // If entry's uris and engine's metastore uris have at least 1 same uri, we presume they
       // represent the same metastore cluster.
       val uriSet = metastoreUris.split(",").filter(_.nonEmpty).toSet
-      val newToken = newTokens
+      val newToken = defaultServiceTokens
         .find { case (uris, token) =>
           val matched = uris.toString.split(",").exists(uriSet.contains) &&
             token.getService == new Text()
@@ -169,7 +194,7 @@ object SparkTBinaryFrontendService extends Logging {
           warn(s"Ignore Hive token with earlier issue date: $token")
         }
       }
-      if (newToken.isEmpty) {
+      if (newToken.isEmpty && defaultServiceTokens.nonEmpty) {
         warn(s"No matching Hive token found for engine metastore uris $metastoreUris")
       }
     } else if (metastoreUris.isEmpty) {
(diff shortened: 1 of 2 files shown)


pan3793:
Code LGTM, but as you know, the Kerberos/UGI/DT part is always tricky stuff, have you tested it in a real cluster in addition to the UT?

pan3793:
also cc @zhouyifan279 and @cxzl25, who have better knowledge of this part

Sunwoo-Shin:
Thanks for the review @pan3793!

Yes — besides the UTs, this has been running on a production Kerberized cluster for a while now.

For context on our setup: the built-in [HiveDelegationTokenProvider](https://github.com/apache/kyuubi/blob/master/kyuubi-server/src/main/scala/org/apache/kyuubi/credentials/HiveDelegationTokenProvider.scala) only supports a single Hive metastore, so we additionally implemented a custom provider through the DelegationTokenProvider SPI that fetches delegation tokens from multiple Hive metastores. Combined with this patch — which preserves the tokens carrying a non-empty service field — the per-metastore (signature-bound) tokens are now propagated correctly to the executors, while the single-metastore path keeps working as before.

pan3793 (approved):

cxzl25 (approved):

pan3793:
thanks, merged to master

PR #7519, opened 2026-06-18
[KYUUBI #7435][SERVER] Honor server-level kyuubi.server.info.provider in getInfo

### LINKED ISSUE #7435: [Bug] `kyuubi.server.info.provider` set in `kyuubi-defaults.conf` is silently ignored
### Code of Conduct

### Search before asking

### Describe the bug

`kyuubi.server.info.provider` set in `kyuubi-defaults.conf` is silently ignored — `KyuubiSessionImpl#getInfo` always behaves as if the value were the default `ENGINE`; every `TGetInfo` call therefore waits for an engine to launch and is forwarded to it.

### Root cause

`SERVER_INFO_PROVIDER` is marked `.serverOnly`. `KyuubiSessionManager#openSession` builds the per-session conf via `getConf.getUserDefaults(user)`, which **unsets every `serverOnly` entry** before returning. As a result, `sessionConf.get(SERVER_INFO_PROVIDER)` in `KyuubiSessionImpl#getInfo` always returns the hard-coded default `"ENGINE"`, regardless of `kyuubi-defaults.conf`.

It only appears to work in tests because they pass the value as a per-session overlay (`withSessionConf(...)`), which is re-applied after `getUserDefaults`.

### Steps to reproduce

1. Set `kyuubi.server.info.provider=SERVER` in `kyuubi-defaults.conf`.
2. Connect via JDBC/beeline without overriding it per-session.
3. A `GetInfo` call still waits for an engine and the call is forwarded to it, instead of the server returning its own info directly.

### Expected vs Actual

- Expected: server-level `SERVER` setting takes effect; `getInfo` returns Kyuubi server info immediately.
- Actual: server-level setting is dropped; `getInfo` always takes the `ENGINE` branch.

### Suggested fix

Read `SERVER_INFO_PROVIDER` from the server-level conf, while still letting an explicit per-session override win:

```scala
val provider = sessionConf.getOption(SERVER_INFO_PROVIDER.key)
  .getOrElse(sessionManager.getConf.get(SERVER_INFO_PROVIDER))
```

Making the entry non-`serverOnly` would also fix it, but would change the semantics for all other usages and would let clients flip the mode, so the targeted fix is preferred.

### Affects Version(s)

1.11.1

### Kyuubi Server Log Output

```logtalk
N/A — no error is logged. The bug is a silent fallback to the default value.
```

### Kyuubi Engine Log Output

```logtalk
N/A
```

```yaml
kyuubi.server.info.provider=SERVER
```

```yaml
# N/A — engine-side config does not affect this bug.
```

### Additional context

_No response_

Sunwoo-Shin:

Closes #7435.

`kyuubi.server.info.provider` was declared `.serverOnly`, so `KyuubiConf#getUserDefaults` stripped it from the per-session conf that `KyuubiSessionManager#openSession` builds. As a result `KyuubiSessionImpl#getInfo` read it through `sessionConf.get(SERVER_INFO_PROVIDER)`, which never saw the server-level value and always returned the hard-coded default `ENGINE`. Setting `kyuubi.server.info.provider=SERVER` in `kyuubi-defaults.conf` was therefore silently ignored, and every `GetInfo` request waited for the engine to launch and forwarded to it instead of being answered by the server directly.

`SERVER_INFO_PROVIDER` is only read on the server frontend (`KyuubiSessionImpl#getInfo`); the engine never reads it. Dropping the `serverOnly` flag lets the server-level value flow into the session conf so `getInfo` honors it, while an explicit per-session value still wins as before.

Affected versions: 1.11.1.

Added `KyuubiServerInfoProviderSuite`, which boots a server with `kyuubi.server.info.provider=SERVER` set only at the server level (no per-session override) and asserts `GetInfo(CLI_DBMS_NAME)` returns `Apache Kyuubi`. Before the fix it returns the engine name (`Spark SQL`).

```
  -Dtest=none \
  -DwildcardSuites=org.apache.kyuubi.operation.KyuubiServerInfoProviderSuite
```

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
@@ -3077,7 +3077,6 @@ object KyuubiConf {
 
   val SERVER_INFO_PROVIDER: ConfigEntry[String] =
     buildConf("kyuubi.server.info.provider")
-      .serverOnly
       .doc("The server information provider name, some clients may rely on this information" +
         " to check the server compatibilities and functionalities." +
         " <li>SERVER: Return Kyuubi server information.</li>" +
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiServerInfoProviderSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiServerInfoProviderSuite.scala
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+import org.apache.kyuubi.WithKyuubiServer
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.shaded.hive.service.rpc.thrift.{TGetInfoReq, TGetInfoType}
+
+class KyuubiServerInfoProviderSuite extends WithKyuubiServer with HiveJDBCTestHelper {
+
+  override protected def jdbcUrl: String = getJdbcUrl
+
+  // `kyuubi.server.info.provider` is set only at the server level, mimicking a
+  // kyuubi-defaults.conf entry, so the test exercises the serverOnly fallback path.
+  override protected val conf: KyuubiConf =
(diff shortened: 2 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionImpl.scala
```diff
@@ -298,7 +298,12 @@ class KyuubiSessionImpl(
   }
 
   override def getInfo(infoType: TGetInfoType): TGetInfoValue = {
-    sessionConf.get(SERVER_INFO_PROVIDER) match {
+    // SERVER_INFO_PROVIDER is `.serverOnly`, so KyuubiConf.getUserDefaults strips it from
+    // sessionConf. Unless the client set it explicitly for this session, fall back to the
+    // server-level conf so the operator's kyuubi-defaults.conf setting is honored.
+    val provider = sessionConf.getOption(SERVER_INFO_PROVIDER.key)
+      .getOrElse(sessionManager.getConf.get(SERVER_INFO_PROVIDER))
+    provider match {
```
ruanwenjun: Would removing serverOnly from SERVER_INFO_PROVIDER be sufficient?

Sunwoo-Shin: Agreed — cleaner. `kyuubi.server.info.provider` is only consumed by KyuubiSessionImpl#getInfo, never by the engine, so dropping serverOnly is safe.

Reverted the getInfo change, removed the flag, and added KyuubiServerInfoProviderSuite.

ruanwenjun (approved):

pan3793 (approved):

pan3793:
thanks, merged to master/1.11

Issue #7244, opened 2025-11-16
[Bug]  batch recovery crashes startup on null request_conf (KyuubiConf.set)


moelhoussein:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

When Batch API v2 is enabled, Kyuubi persists batch requests’ conf map to the metadata store without null validation. On server restart, recovery loads these rows and constructs KyuubiBatchSession from metadata.request_conf. If any key has a null value (e.g., `"sfPassword": null`), `KyuubiConf.set` throws IllegalArgumentException (“value cannot be null for key: …”) during session construction in getBatchSessionsToRecover, which propagates and causes KyuubiRestFrontendService startup to fail. 

**Expected**
the offending batch should be skipped/marked ERROR and startup should continue; 

**Actual** 
server startup aborts on the uncaught exception.

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk
2025-11-15 02:53:58.945 ERROR org.apache.kyuubi.server.KyuubiServer: Error starting service KyuubiRestFrontendService
org.apache.kyuubi.KyuubiException: Cannot start KyuubiRestFrontendService
        at org.apache.kyuubi.server.KyuubiRestFrontendService.start(KyuubiRestFrontendService.scala:239)
        at org.apache.kyuubi.service.CompositeService.$anonfun$start$1(CompositeService.scala:47)
        at org.apache.kyuubi.service.CompositeService.$anonfun$start$1$adapted(CompositeService.scala:45)
        at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
        at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
        at org.apache.kyuubi.service.CompositeService.start(CompositeService.scala:45)
        at org.apache.kyuubi.service.Serverable.start(Serverable.scala:53)
        at org.apache.kyuubi.server.KyuubiServer.start(KyuubiServer.scala:225)
        at org.apache.kyuubi.server.KyuubiServer$.startServer(KyuubiServer.scala:69)
[...]

PR #7521, opened 2026-06-23
[KYUUBI #7244][SERVER] Skip invalid batch metadata during recovery

hutiefang76:

Null-valued batch configuration entries can reach persisted batch metadata and later fail when `KyuubiBatchSession` copies `requestConf` into `KyuubiConf`, whose `set(key, value)` rejects null values.

The fix now covers the full lifecycle instead of turning a recovery-time validation error into a terminal batch state:

- filter null-valued configuration at the REST creation boundary, before user/batch-id lookup and V2 metadata persistence;
- sanitize configuration and the attached metadata at the shared `createBatchSession` boundary, covering V1 creation, V2 `INITIALIZED` submission, and `PENDING`/`RUNNING` recovery;
- keep unrelated recovery failures on their existing retry/error path rather than broadly converting them to `ERROR`.

The shared sanitizer also provides compatibility for legacy metadata that already contains a null value.

- Added `BatchesResourceSuite` coverage proving V1 and V2 requests persist metadata without the null-valued entry.
- Added `KyuubiSessionManagerSuite` coverage for both `INITIALIZED` submission and `PENDING`/`RUNNING` recovery; the batch remains recoverable and the null key is absent from the session configuration.
- Targeted lifecycle suite: 2 tests passed, 0 failed.
- REST V1/V2 suites: the 2 new tests and 26 other tests passed. Two existing `open batch session` log assertions failed because the local Spark fixture did not emit `Submitted application: Spark Pi`; the same two failures were reproduced on a detached clean `origin/master` worktree with the same `SPARK_HOME`.
- `dev/reformat`
- `git diff --check`

Assisted-by: OpenAI Codex

### Bug history and affected versions

- General batch-session recovery was introduced by KYUUBI #2373 and is present from the 1.6.0-incubating release line onward.
- Batch API V2 was introduced by KYUUBI #4790 and is present from the 1.8.0 release line onward. The reported V2 path is therefore confirmed for 1.8.0+ and current master.
- Current master also has the explicit batch reassignment recovery path added by KYUUBI #6884; the shared session-creation sanitizer covers both normal startup recovery and explicit batch-id recovery.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
@@ -158,11 +158,13 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
     // scalastyle:on
     val username = Option(user).filter(_.nonEmpty).getOrElse("anonymous")
     val sessionConf = this.getConf.getUserDefaults(user)
+    val sanitizedConf = Metadata.sanitizeRequestConf(conf)
+    val sanitizedMetadata = metadata.map(_.withSanitizedRequestConf)
     new KyuubiBatchSession(
       username,
       password,
       ipAddress,
-      conf,
+      sanitizedConf,
       this,
       sessionConf,
       batchType,
@@ -170,7 +172,7 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       resource,
       className,
       batchArgs,
-      metadata,
+      sanitizedMetadata,
       fromRecovery)
   }
 
@@ -242,7 +244,7 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
       resource = batchRequest.getResource,
       className = batchRequest.getClassName,
       requestName = batchRequest.getName,
-      requestConf = conf,
+      requestConf = Metadata.sanitizeRequestConf(conf),
       requestArgs = batchRequest.getArgs.asScala.toSeq,
       createTime = System.currentTimeMillis(),
       engineType = batchRequest.getBatchType,
@@ -319,9 +321,7 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
         stateToRecover.toString,
         kyuubiInstance,
         0,
-        Int.MaxValue).map { metadata =>
-        createBatchSessionFromRecovery(metadata)
-      }).getOrElse(Seq.empty)
+        Int.MaxValue).map(createBatchSessionFromRecovery)).getOrElse(Seq.empty)
     }
   }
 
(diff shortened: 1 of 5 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
```diff
@@ -338,11 +337,38 @@ class KyuubiSessionManager private (name: String) extends SessionManager(name) {
...
               m.kyuubiInstance == kyuubiInstance && batchStatesToRecovery.contains(m.opState))
-            .flatMap { metadata => Some(createBatchSessionFromRecovery(metadata)) }
+            .flatMap(createBatchSessionFromRecoverySafely)
       }
     }
   }
 
+  private def createBatchSessionFromRecoverySafely(
+      metadata: Metadata): Option[KyuubiBatchSession] = {
+    metadata.requestConf.collectFirst { case (key, null) => key } match {
```
pan3793: has this bug been fixed? if so, document the history and the affected version

hutiefang76:
Thanks for the reminder. I updated the PR description with the history and affected-version scope: general batch-session recovery exists from 1.6.0-incubating+, Batch API V2 was introduced in 1.8.0+, and this issue is confirmed for the reported Batch V2 path on 1.8.0+ and current master. The patch also guards the current master explicit batch reassignment recovery path added later.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
hutiefang76: Yes. I documented the history and affected versions in the PR description and verified them from the repository tags: batch-session recovery was introduced by KYUUBI #2373 (first released in 1.6.0-incubating), while Batch API V2 was introduced by #4790 (first released in 1.8.0). The reported V2 recovery failure is therefore confirmed for 1.8.0+ and current master. #6884 later added the explicit batch-reassignment recovery path; the helper in this PR guards both normal startup recovery and that reassignment path.

pan3793: what I mean bug fix is, filtering out the null-valued config when creating the batch ...

I remember having fixed it, seems not to have been pushed to upstream ...

wangzhigang1999:
Thanks for the fix. It addresses the recovery failure, but I’m still concerned about the full lifecycle of invalid metadata, so I’d prefer to hold off on merging for now.

kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiSessionManager.scala
hutiefang76: Thanks for clarifying. I reworked the patch around creation-time filtering in 34dcff8a0:

- The REST boundary removes null-valued entries before user/batch-id lookup and V2 metadata persistence.
- The shared createBatchSession boundary sanitizes both the configuration and attached metadata, so legacy INITIALIZED submission and PENDING/RUNNING recovery are covered as well.
- The previous broad recovery catch/ERROR transition was removed; unrelated recovery failures keep their existing behavior.

I added separate tests for REST persistence, INITIALIZED submission, and recovery. The targeted lifecycle suite passes 2/2; the new REST test passes in both V1 and V2 suites. I also documented the full test evidence, including the two unrelated Spark log-fixture failures reproduced on clean origin/master, in the PR description.

PR #7527, opened 2026-06-26
[JDBC] Respect interrupt while waiting for engine launch

ruanwenjun:

JDBC clients can block inside `KyuubiConnection` construction while waiting for engine launch. If the caller cancels that connection attempt with `Thread.interrupt()`, the wait loop previously kept polling and the caller still had no returned `Connection` object to close.

This patch makes the launch wait path honor interruption by converting it to a typed JDBC exception, `KyuubiInterruptedException`. The interrupt status is intentionally consumed before the exception reaches the caller, and comments explain that the caller receives the typed exception while the current thread is not interrupted.

The patch also keeps `SQLException` subclasses intact when `KyuubiDataSource` creates a connection, so callers can distinguish the interrupted launch path.

- `./build/mvn -pl kyuubi-hive-jdbc -am spotless:apply`
- `./build/mvn -pl kyuubi-hive-jdbc -am -Dtest=KyuubiConnectionTest,KyuubiStatementTest,TestJdbcDriver,TestKyuubiPreparedStatement,UtilsTest,ZooKeeperHiveClientHelperTest -DwildcardSuites=none test`
- `git diff --check`

Also attempted `dev/reformat`, but the full repo reformat is blocked in this local environment because Spotless expects `black 22.3.0` while the installed executable is `black 25.1.0`.

Assisted-by: OpenAI Codex: GPT-5

### DIFF
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
@@ -79,6 +79,7 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
   public static final Logger LOG = LoggerFactory.getLogger(KyuubiConnection.class.getName());
   public static final String BEELINE_MODE_PROPERTY = "BEELINE_MODE";
   public static final String HS2_PROXY_USER = "hive.server2.proxy.user";
+  private static final String SQL_STATE_LAUNCH_ENGINE_CANCELED = "01000";
   public static int DEFAULT_ENGINE_LOG_THREAD_TIMEOUT = 10 * 1000;
 
   private String jdbcUriString;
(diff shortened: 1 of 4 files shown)


ruanwenjun:
PTAL cc @pan3793 @cxzl25

zhouyifan279:
LGTM

wangzhigang1999:
LGTM 👍

Just a nit (non-blocking): isInterrupted(Throwable) and the test's hasCause(...) both hand-roll a cause-chain walk. Since commons-lang3 is already a dependency here (this file already imports StringUtils/ClassUtils from it), we could reuse ExceptionUtils and drop both helpers:

```java
  // import org.apache.commons.lang3.exception.ExceptionUtils;
  ExceptionUtils.indexOfType(e, InterruptedException.class) >= 0
```

Same semantics, ~20 fewer lines, removes the duplicated logic between main and test.

ruanwenjun:
Addressed in 6795917ac: replaced both cause-chain walks with `ExceptionUtils.indexOfType(..., InterruptedException.class)` and removed the duplicated helpers.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/KyuubiConnection.java
```diff
@@ -79,6 +80,7 @@ public class KyuubiConnection implements SQLConnection, KyuubiLoggable {
   public static final Logger LOG = LoggerFactory.getLogger(KyuubiConnection.class.getName());
   public static final String BEELINE_MODE_PROPERTY = "BEELINE_MODE";
   public static final String HS2_PROXY_USER = "hive.server2.proxy.user";
+  private static final String SQL_STATE_LAUNCH_ENGINE_CANCELED = "01000";
```
pan3793: The SQLSTATE "01000" is used for engine launch cancellation, but class "01" is "success-with-warning." Any caller distinguishing errors from warnings via `getSQLState()` will misclassify it. Recommend "57014" (query cancelled) or "HY008" (operation cancelled).

ruanwenjun: Addressed in 6ff3e96ef. Changed launch-engine cancellation from warning-class `01000` to error-class `57014`, and added a regression test for `CANCELED_STATE` in `waitLaunchEngineToComplete`.

pan3793 (approved):

cxzl25 (approved):

pan3793:
thanks, merged to master

PR #7528, opened 2026-06-27
Remove MySQL frontend protocol

pan3793:

The MySQL frontend was added as an experimental feature for a long time, but it never got improved after that. 

Pass GHA.

Assisted-by: GLM 5.2


dnskr (changes requested):
This looks great, thanks!

Just one minor thing I noticed while reviewing: it looks like we also should delete the references to the MySQL frontend from a few of the documentation pages:

1. https://github.com/apache/kyuubi/blob/0d7abecba11e4226f4ed4cbc5276a1416df42275/docs/index.rst?plain=1#L82-L83
2. https://github.com/apache/kyuubi/blob/0d7abecba11e4226f4ed4cbc5276a1416df42275/docs/client/jdbc/index.rst?plain=1#L24
3. https://github.com/apache/kyuubi/blob/0d7abecba11e4226f4ed4cbc5276a1416df42275/docs/client/jdbc/mysql_jdbc.rst
4. https://github.com/apache/kyuubi/blob/0d7abecba11e4226f4ed4cbc5276a1416df42275/docs/configuration/settings.md?plain=1#L167

pan3793:
@dnskr thanks for checking, I addressed 1, 2, 3, left 4 as-is because it's a different thing.

dnskr (approved):

pan3793:
thanks, merged to master

PR #7531, opened 2026-06-30
Add rebalance advisoryPartitionSizeInBytes config

ulysses-you:

Starting from Spark 3.5, the RebalancePartitions operator supports an optAdvisoryPartitionSize parameter to specify the desired partition size after shuffle. This allows Kyuubi to directly control the target output size of the rebalance shuffle, helping AQE's CoalesceShufflePartitions to produce appropriately sized output files and avoid small file issues.

Currently, RebalanceBeforeWriting inserts RebalancePartitions without any advisory size, leaving AQE to use the global spark.sql.adaptive.advisoryPartitionSizeInBytes which may not be suitable for the final write stage.

This patch introduces a new Kyuubi configuration key spark.sql.adaptive.rebalancePartitionsAdvisoryPartitionSizeInBytes that takes precedence over the final-stage config spark.sql.finalStage.adaptive.advisoryPartitionSizeInBytes as a fallback.

Added 5 unit tests in RebalanceBeforeWritingSuite covering all scenarios:
- Final-stage config takes effect
- Kyuubi config fallback
- Kyuubi config takes precedence over final-stage when both are set
- Neither config is set (optAdvisoryPartitionSize is None)
- Raw getAdvisoryPartitionSize(conf) config resolution

Claude Code (Anthropic Claude Opus 4.7)


pan3793 (approved):

wForget (approved):
Thanks @ulysses-you .

Slightly off topic, but do you think it would be possible to introduce a REBALANCE_WITH_SIZE hint in upstream spark? That would provide a simple way to control small file generation when the kyuubi spark extension is not installed.

extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
```diff
@@ -254,4 +267,16 @@ object KyuubiSQLConf {
       .version("1.9.0")
       .booleanConf
       .createWithDefault(true)
+
+  def getAdvisoryPartitionSize(conf: SQLConf): Option[Long] = {
+    conf.getConf(REBALANCE_PARTITIONS_ADVISORY_PARTITION_SIZE).orElse {
+      if (conf.contains(FINAL_STAGE_ADVISORY_PARTITION_SIZE_KEY)) {
```
wForget, line 273: should also check `spark.sql.optimizer.finalStageConfigIsolation.enabled`

wForget:
We may also need to update the documentation here: https://github.com/apache/kyuubi/blob/master/docs/extensions/engines/spark/rules.md

extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
ulysses-you: In general, the finalStage related configs should only appear with finalStage enabled.

ulysses-you:
> do you think it would be possible to introduce a REBALANCE_WITH_SIZE hint in upstream spark?

@wForget  good idea, I think it's a useful hint

extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
ulysses-you: addressed

ulysses-you:
thanks, merging to master!

PR #7539, opened 2026-07-03
[Spark]  Remove rebalance shuffle before writing at AQE time when not worthwhile

ulysses-you:
## Why are the changes needed?

`RebalanceBeforeWriting` injects a `RebalancePartitions` operator before writes (physically a
`ShuffleExchangeExec` with `shuffleOrigin = REBALANCE_PARTITIONS_BY_NONE`) to merge small files and
avoid the small-file problem. However, in some cases this rebalance shuffle is pure overhead: once
the upstream data is materialized and its real size is known, the shuffle brings no benefit and only
adds an extra round of shuffle read/write.

This PR adds an AQE runtime optimizer rule `RemoveRebalanceShuffle` that, at AQE time (when the
upstream query stage is materialized and real statistics are available), decides whether such a
not-worthwhile rebalance shuffle can be removed, saving one shuffle.

A new rule `RemoveRebalanceShuffle`, injected into the `AQEOptimizer` via
`injectRuntimeOptimizerRule`, runs on the **logical plan**: it replaces `RebalancePartitions` with
its child, and the physical planner then re-plans without the shuffle.

**Scope** — only the rebalance feeding the final write is considered: starting from `WriteFiles`,
the rule walks down through at most `Project` and local `Sort`; any other operator in between leaves
the rebalance untouched.

**Removal conditions** (all must hold):

1. `spark.sql.adaptive.removeRebalanceShuffle.enabled` is on;
2. the `RebalancePartitions` has no partition expressions (i.e. `REBALANCE_PARTITIONS_BY_NONE`);
3. its `optAdvisoryPartitionSize` is present and **larger than**
   `spark.sql.adaptive.advisoryPartitionSizeInBytes` (a smaller advisory size sugge
   intentionally wants finer partitions, so we respect that).

**Size decision** — the rebalance input is split into groups of sub-plans whose leaves are all
materialized `LogicalQueryStage` exchange stages (a `Union` yields one group per br
measured by its runtime `sizeInBytes`. The shuffle is removed only when every group is removable,
where a group is removable under either:

- **Large data**: the input has no data-reducing operator and the group's represent
  `max(maxStageSize, groupSum / 2)` is larger than
  `spark.sql.shuffle.partitions * removeRebalanceShuffle.smallPartitionSize`; or
- **Small data**: the input has no data-expanding operator and the representative size is smaller
  than `advisoryPartitionSizeInBytes * removeRebalanceShuffle.tolerableSmallFileNum

Operator classification (logical operators):

- reducing: `Aggregate` / `Filter` / `Sample` / `Offset` / `GlobalLimit` / `LocalLi
  `WindowGroupLimit` / `Inner` join / semi-anti join;
- expanding: `Generate` / `Expand` / non-semi-anti `Join`.

**New configs** (`KyuubiSQLConf`, version 1.12.0):

| Config | Default |
|---|---|
| `spark.sql.adaptive.removeRebalanceShuffle.enabled` | `false` |
| `spark.sql.adaptive.removeRebalanceShuffle.smallPartitionSize` | `96MB` |
| `spark.sql.adaptive.removeRebalanceShuffle.tolerableSmallFileNum` | `5` | 

Modules changed (identical across 3.5 / 4.0 / 4.1):

- Add `RemoveRebalanceShuffle.scala` and `RemoveRebalanceShuffleSuite.scala`;
- Add 3 configs in `KyuubiSQLConf.scala`;
- Register the rule in `KyuubiSparkSQLExtension.scala`;
- Document the feature and configs in `docs/extensions/engines/spark/rules.md`.

## How was this patch tested?

Added `RemoveRebalanceShuffleSuite` (3.5 / 4.0 / 4.1), which asserts on the presence of
`REBALANCE_PARTITIONS_BY_NONE` shuffle in the executed plan, covering:

- semi join (reducing), small data → removed;
- outer join (expanding, non-reducing), large data with tiny / huge `smallPartitionSize` → removed /
  kept respectively;
- rebalance advisory size ≤ session advisory size → kept;
- inner join (both reducing and expanding) → kept;
- semi join with `tolerableSmallFileNum=0` → kept;
- switch off → kept.

All pass on Spark 3.5:

```
build/mvn -Pspark-3.5 -pl extensions/spark/kyuubi-extension-spark-3-5 -am test \
  -Dtest=none -DwildcardSuites=org.apache.spark.sql.RemoveRebalanceShuffleSuite
```

No regression in `RebalanceBeforeWritingSuite`. Formatted with `spotless:apply` across all Spark
profiles.

co-authored: Claude Opus 4.8


ulysses-you:
cc @pan3793 @wForget

extensions/spark/kyuubi-extension-spark-4-1/src/main/scala/org/apache/kyuubi/sql/RemoveRebalanceShuffle.scala
```diff
@@ -0,0 +1,192 @@
...
+      case _: Filter => true
+      case _: Sample => true
+      case _: Offset => true
+      case _: GlobalLimit => true
+      case _: LocalLimit => true
+      case _: WindowGroupLimit => true
+      case Join(_, _, joinType, _, _) =>
+        joinType match {
+          case LeftExistence(_) => true
+          case Inner => true
```
pan3793, line 197: why classify inner join as a reducing op? what's the side effect if the decision is wrong here?

pan3793 (approved):
LGTM, one question

wForget:
>  (!hasExpandingOperators && maxStageSize < sessionAdvisorySize * tolerableSmallFileNum)

If I understand correctly, this is because CoalescePartitions will cause the downstream shuffle to always approach sessionAdvisorySize, so we can remove this rebalance. (should we check the `spark.sql.adaptive.coalescePartitions.enabled` configuration?)
    
> (!hasReducingOperators && maxStageSize > shufflePartitions * smallPartitionSizeThreshold)

Currently, only the rebalance and join operators split skewed partitions. If there are skewed partitions upstream and they are not rebalance/join operators, will this lead to excessively large partitions?

ulysses-you:
>  (should we check the spark.sql.adaptive.coalescePartitions.enabled configuration?)

I'm fine to check this config althrough it should always be true

> If there are skewed partitions upstream and they are not rebalance/join operators, will this lead to excessively large partitions

the behavior is same as before, we only remove the rebalace which is the parent of the skewed operator, e.g.,

```
   Exchange
      |
     Sort
      |
Window (skewed)
      |
Rebalance Exchange (this pr wants to remove)
      |
  WriteFiles
```

extensions/spark/kyuubi-extension-spark-4-1/src/main/scala/org/apache/kyuubi/sql/RemoveRebalanceShuffle.scala
ulysses-you: > why classify inner join as a reducing op?

the inner join have a chance reduce data size if no keys matching

> what's the side effect if the decision is wrong here?

if the inner join does not reduce data size, the rebalance is redundant which is same with current behavior (without this rule)

wForget (approved):
Thanks @ulysses-you , LGTM

ulysses-you:
thanks, merged to master

PR #7540, opened 2026-07-03
KSHC supports Spark 4.1

pan3793:

This PR makes KSHC (Kyuubi Spark Hive Connector) support Spark 4.1

Main idea: get rid of `spark.sql.legacy.v1IdentifierNoCatalog` (which does not work well in Spark 4.1) and use `attachCatalogName` to adapt to SPARK-46283

Update GHA workflow to enable KSCH tests for Spark 4.1, also update KSHC cross-version tests - build with Spark 3.5, test on Spark 4.1

Assisted-by: GLM 5.2

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -217,6 +217,11 @@ jobs:
             spark-compile: "3.5"
             spark-runtime: "4.0"
             comment: "normal"
+          - java: 21
+            scala: "2.13"
+            spark-compile: "3.5"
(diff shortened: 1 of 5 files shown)


pan3793:
to reviwers, suggest checking hide whilespace on viewing diffs

image

cc @ulysses-you @wForget

ulysses-you (approved):

pan3793:
thanks, merged to master

PR #7542, opened 2026-07-06
Validate filename in Utils

aajisaka:

Validate the file path is in the target directory before copying the file

Added unit tests

Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
@@ -219,9 +219,6 @@ object Utils extends Logging {
       if (source == null) {
         throw new IOException("the source inputstream is null")
       }
-      if (!dir.toFile.exists()) {
-        dir.toFile.mkdirs()
-      }
       val (prefix, suffix) = fileName.lastIndexOf(".") match {
         case i if i > 0 => (fileName.substring(0, i), fileName.substring(i))
         case _ => (fileName, "")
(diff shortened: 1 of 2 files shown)


kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
```diff
@@ -229,6 +229,10 @@ object Utils extends Logging {
       val currentTime = new SimpleDateFormat("yyyyMMddHHmmss").format(new Date())
       val identifier = s"$currentTime-${tempFileIdCounter.incrementAndGet()}"
       val filePath = Paths.get(dir.toString, s"$prefix-$identifier$suffix")
+      if (!filePath.normalize().startsWith(dir.normalize())) {
+        throw new IOException(
```
pan3793: `source` leaks if it goes here

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
pan3793: `mkdirs` runs before validation, so a failed check leaves an empty directory behind.

pan3793 (commented):
fix makes sense, two other issues

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
aajisaka: Thanks. Moved into try-finally

kyuubi-common/src/main/scala/org/apache/kyuubi/Utils.scala
aajisaka: Thank you! Moved `mkdirs` after the validation

pan3793 (approved):

aajisaka:
Merged into master and branch-1.11. Thank you @pan3793

PR #7553, opened 2026-07-10
Bump Flink 1.20.5

pan3793:

As title.

Pass GHA.

Assisted-by: GLM 5.2


pan3793:
merged to master/1.11

Issue #7230, opened 2025-10-23
[Bug] DataSource V2 table extraction fails for MongoDB


moelhoussein:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

For MongoDB, the table extractor fails to produce meaningful (database, table) identifiers. It relies on `DataSourceV2Relation.identifier` or `table.name`, which are empty, path-like, or non-semantic for these connectors. As a result, Ranger resources are misidentified and policies do not match.

DataSource V2 connectors encode logical identifiers in `table.properties`, not in `identifier/table.name`. The extractor didn’t read connector-specific properties.

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk
N/A
```

### Kyuubi Engine Log Output

```logtalk
Exception in thread "main" org.apache.kyuubi.plugin.spark.authz.AccessControlException: Permission denied: user [JOB-USER] does not have [update] privilege on [MongoTable()]
```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

j1wonpark:
Hit the same class of problem with the ClickHouse Spark connector (its native `TableProvider` landed in 0.10.0).

@pan3793 re: whether `table.properties` is a contract — it isn't. `Table#properties()` is a `default` method returning `Collections.emptyMap()`, and the reserved keys `TableCatalog` defines (`location`, `provider`, `owner`, ...) carry metadata, not identity; identity lives in the separate `Identifier` type. Reading connector-specific property keys means one extraction rule per connector — tried for Paimon in #5592, closed unmerged once Paimon exposed a proper identifier.

Where the empty identifier comes from: only `SupportsCatalogOptions` providers get a catalog/identifier pair. On the read side `DataSourceV2Utils.loadV2Source` falls back to

```scala
case _ =>
  // TODO: Non-catalog paths for DSV2 are currently not well defined.
  ...
  (tbl, None, None, None)
```

[...]

PR #6544, opened 2024-07-17
[KYUUBI #6541] [AUTHZ] Fix DataSourceV2RelationTableExtractor can't get the 'database' attribute if it's a Paimon plan.

promising-forever:
# :mag: Description
## Issue References 🔗

This pull request fixes #6541

## Describe Your Solution 🔧
Fix an issue where DataSourceV2RelationTableExtractor#table could not fetch the ‘database’ attribute causing the Ranger checks to fail when using the Paimon Catalog.
If the database attribute is not resolved, use DataSourceV2RelationTableExtractor#identifier to complete it.

## Types of changes :bookmark:

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Related Unit Tests

---

# Checklist 📝

- [ ] This patch was not authored or co-authored using [Generative Tooling](https://www.apache.org/legal/generative-tooling.html)

**Be nice. Be informative.**

PR #7557, opened 2026-07-11
[KYUUBI #7230][AUTHZ] Support skipping privilege check for DataSourceV2Relation without catalog and identifier

j1wonpark:

DSv2 `TableProvider`s that do not implement `SupportsCatalogOptions` (the MongoDB connector, the ClickHouse native connector, etc.), when used directly via `format(...).load()` / `.save()`, produce a `DataSourceV2Relation` with neither `catalog` nor `identifier` — Spark's non-catalog fallback in `DataSourceV2Utils.loadV2Source` / `DataFrameWriter` passes `(None, None)`, and the read-side fallback carries Spark's own TODO note that "Non-catalog paths for DSV2 are currently not well defined".

For such relations `DataSourceV2RelationTableExtractor` falls back to `table.name()`, which is connector-defined and usually synthetic (e.g. `MongoTable()`). The resulting Ranger resource has no `database` element, and a resource missing a parent level of the `database → table → column` hierarchy matches no policy at all — not even `database=* / table=*` (`RangerDefaultPolicyResourceMatcher.isHierarchyValidForResources`, Ranger 2.6.0). So the current behavior is an unconditional deny that operators cannot lift with any policy, reported in #7230. Meanwhile the v1 path is already skipped: a `LogicalRelation` without `catalogTable` produces no privilege object.

This PR adds an operator-level opt-out, default `false` so default behavior is unchanged:

- `spark.kyuubi.authz.skip.catalogless.v2.relation.enabled` (default `false`): when enabled, the extractor returns `None` for a `DataSourceV2Relation` whose catalog and identifier are both empty, restoring parity with the v1 path.
- The key is added to `AuthzConfigurationChecker`'s built-in restricted list (like `spark.sql.runSQLOnFiles`), so end users cannot flip it with `SET`. Operators should also consider listing it in `kyuubi.session.conf.restrict.list`.

Note: whether skipping is safe is deployment-specific — fine when credentials for the external system are supplied per query by the end user, risky when the shared Spark principal holds ambient credentials. Hence opt-in rather than unconditional. Skipped relations produce no Ranger audit events.

- New `DataSourceV2RelationTableExtractorSuite`:
  - default behavior preserved: the synthetic `table.name()` is still extracted verbatim when the conf is off;
  - the relation is skipped when the conf is enabled;
  - relations carrying an identifier are unaffected even with the conf enabled (the database fallback from #6544 still applies).
- Extended `AuthzConfigurationCheckerSuite`: `SET` on the new key throws `AccessControlException`.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/serde/DataSourceV2RelationTableExtractorSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/serde/DataSourceV2RelationTableExtractorSuite.scala
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.plugin.spark.authz.serde
+
+import java.util
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.connector.catalog.{Identifier, Table => V2Table, TableCapability}
+import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.scalatest.BeforeAndAfterAll
+// scalastyle:off
+import org.scalatest.funsuite.AnyFunSuite
+
+class DataSourceV2RelationTableExtractorSuite extends AnyFunSuite with BeforeAndAfterAll {
+// scalastyle:on
+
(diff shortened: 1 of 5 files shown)


extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/serde/DataSourceV2RelationTableExtractorSuite.scala
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.scalatest.BeforeAndAfterAll
+// scalastyle:off
+import org.scalatest.funsuite.AnyFunSuite
+
+class DataSourceV2RelationTableExtractorSuite extends AnyFunSuite with BeforeAndAfterAll {
+// scalastyle:on
+
+  private val skipConfKey = "spark.kyuubi.authz.skip.catalogless.v2.relation.enabled"
```
pan3793:
```suggestion
  private val skipConfKey = "spark.kyuubi.authz.skipCataloglessV2Relation.enabled"
```

pan3793: dot for namespace, camelCase for words

pan3793 (approved):
LGTM, except for conf name

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/serde/DataSourceV2RelationTableExtractorSuite.scala
j1wonpark: Renamed to `spark.kyuubi.authz.skipCataloglessV2Relation.enabled` — dot for namespace, camelCase for words, as suggested. Thanks for the review!

pan3793:
thanks, merged to master/1.12

Issue #7529, opened 2026-06-29
[FEATURE][UMBRELLA] Data Agent Engine — isolation, identity, and security hardening


wangzhigang1999:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Umbrella for the security-hardening / isolation follow-ups deferred while landing the Data Agent engine (#7379, last PR #7431). Sub-tasks are tracked inline, not split into separate issues.

**Root cause behind most of these:** `DataAgentProcessBuilder extends ProcBuilder`, so the engine runs as a local child process on the Kyuubi server host (no k8s/YARN launch path) and shares the server's Kerberos credential cache, network identity, and filesystem. That sharing is why Kerberos is currently banned and why credentials must go in the JDBC URL. Giving the engine its own isolated runtime + identity unblocks the rest.

### Sub-tasks

- [ ] **①a Cross-server Data Agent session routing.** Make Kyuubi servers stateless proxies for Data Agent REST sessions. Register a discovery mapping from the external session UUID to its Data Agent engine, then let any server attach to the existing engine-side session handle instead of depending on the creating server's in-memory `KyuubiSessionImpl`. The conversation, approval, and tool state remains in the engine. This fixes requests being routed to another server in a multi-server deployment; it does not preserve a local child engine when its launching server fails.
- [ ] **①b Isolated runtime with its own identity (keystone).** Add a remote launch path so `DATA_AGENT` runs in its own container with a distinct identity instead of as a local child of the server — k8s first, YARN later; keep local-process as a dev fallback. This decouples engine lifetime from a server instance and unblocks ② and ③.
- [ ] **② Kerberos support** *(depends on ①b)*. Engine logs in with its own principal/keytab and impersonates the session user via `hadoop.proxyuser` ACLs, lifting the current ban. Removes the two TODOs in `EngineRef#deriveDataAgentJdbcUrl` and `ChatCompletionProvider`.
- [ ] **③ Credential propagation** *(design depends on ①b)*. Dedicated config / UI inputs + a `propagateCredential` opt-in so secrets aren't embedded in `kyuubi.engine.data.agent.jdbc.url`. Align with the JDBC engine's `propagateCredential` and Trino's `connection.user` fallback. Raised in #7431.
[...]

PR #7559, opened 2026-07-13
[KYUUBI #7529][DATA-AGENT] Support cross-server session routing

wangzhigang1999:

In a multi-server deployment, Data Agent requests may reach a server that does not own the local session and fail with `session not found`.

This patch stores the session-to-engine route in the HA backend, allowing any server to attach to the existing engine session without resetting its state.

Related to #7529.

- HA and REST test suites: 39 tests passed.
- All 17 affected Maven modules built successfully.
- Tested two Kyuubi Servers with ZooKeeper, covering cross-server chat, approval, close, concurrent access, state retention, authorization, and stale route cleanup.


wangzhigang1999:
Manually tested on a two-server Kubernetes deployment:

- Opened a Data Agent session on one server, then accessed, chatted, approved tool calls, and closed it through the other server.
- Verified conversation state, user isolation, stale-route cleanup, and real model/SQL execution.
- Ran four concurrent sessions and 500 cross-server requests without failures or resource leaks.

Server and local Engine failure recovery remains outside this PR.

PR #7561, opened 2026-07-13
[INFRA] Cache CI engine archives per URL with a reusable action

pan3793:

The old `cache-engine-archives` cache is one blob under a constant key, so bumping any
engine invalidates all of them. This provisions engines explicitly instead:

- New reusable action `.github/actions/download-archive` downloads a tarball and caches it
  with `actions/cache` **keyed by the URL** (atomic download), so each engine caches
  independently and stable versions restore instantly.
- `.github/scripts/resolve-engine-archive-urls.sh` derives each URL from the Maven
  properties (honoring the build's profiles/`-D` overrides); `expose-engine-homes.sh`
  publishes `SPARK_HOME`/`HIVE_HOME`/`FLINK_HOME` from `/opt`.
- Every `master.yml` job now resolves → downloads → exposes only the engines it needs, and
  the build runs with `-P{spark,flink,hive}-provided`. The now-orphaned
  `cache-engine-archives` action is removed.
- Flink test helpers now honor `FLINK_HOME` first (fallback to the download dir for local
  runs), matching Spark/Hive.
- Spark extension modules gain a `spark-home-from-archive` profile, auto-activated only when
  `SPARK_HOME` is unset, so the Standalone-cluster tests use the downloaded archive locally
  but a provided `SPARK_HOME` is no longer overridden.
- The `hive-it` avatica/CALCITE-1224 hack is now a single `zip -d`.

Run by this PR's CI.

Locally checked: URL resolution across all matrix cases (Scala 2.13, Spark/Flink/Hive
version overrides, CDH6 empty-query), `*_HOME` resolution, the `closer.lua` redirect for
`curl -fSL`, the `zip -d` patch on a mock jar, and the extension `SPARK_HOME` profile via
the effective POM (inherits the env when set, falls back to the archive otherwise).

Assisted-by: Claude Opus 4.8


pan3793:
now the engine binary tarball takes less than 10s to download when hitting cache

wForget (approved):
Thanks @pan3793

wangzhigang1999 (approved):
Nice work， thanks！

bowenliang123 (approved):
LGTM. Thanks!

pan3793:
thanks, merged to master/1.11

pan3793:
backported to branch-1.10 too

PR #7564, opened 2026-07-14
[INFRA] Bump GHA workflows to latest action versions

pan3793:

To keep our CI on the latest GitHub Actions versions, per the [apache/infrastructure-actions allowlist](https://github.com/apache/infrastructure-actions/blob/main/actions.yml).

- For the GitHub-maintained actions (`actions/*`, `github/*`) and allowlisted `*` actions, the allowlist permits any version, so this PR tracks the latest **major** version tags (`@vN`) rather than specific patches.
- For hash-pinned allowlisted third-party actions, the convention is to pin to the allowlisted commit hash with a `# vX.Y.Z` comment. The Docker actions used here were already at the latest allowlisted hashes, so they are unchanged.

Changes (patch tag -> latest major):

- `actions/checkout`      `v7.0.0`  -> `v7`
- `actions/setup-java`    `v5.4.0`  -> `v5`
- `actions/setup-python`  `v6.3.0`  -> `v6`
- `actions/cache`         `v6.1.0`  -> `v6`  (and `v4` -> `v6` in `.github/actions/setup-maven/action.yaml`)
- `actions/stale`         `v10.3.0` -> `v10`

YAML-only change; no unit tests apply. Verified by:

- Cross-checked each action's target against the ASF allowlist and the latest release tag on GitHub.
- Ran `./dev/reformat` (no source changes; Spotless does not manage YAML).
- The `ASF Allowlist Check` workflow in this repo enforces the allowlist on this PR.

Assisted-by: codemaker with glm-5.2

### DIFF
--- a/.github/actions/setup-maven/action.yaml
+++ b/.github/actions/setup-maven/action.yaml
@@ -21,7 +21,7 @@ runs:
   using: composite
   steps:
     - name: Restore cached Maven
-      uses: actions/cache@v4
+      uses: actions/cache@v6
       with:
         path: build/apache-maven-*
         key: setup-maven-${{ hashFiles('pom.xml') }}
--- a/.github/workflows/asf-allowlist-check.yml
+++ b/.github/workflows/asf-allowlist-check.yml
@@ -39,7 +39,7 @@ jobs:
   asf-allowlist-check:
     runs-on: ubuntu-slim
     steps:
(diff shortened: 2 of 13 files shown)


pan3793:
merged to master, will port to branch-1.11 and branch-1.10 too

PR #7567, opened 2026-07-14
[SERVER] Improve Engine UI proxy 403 response

wangzhigang1999:

The Engine UI proxy currently falls back to Jetty's generic 403 page when proxying is disabled, the target host is not allowlisted, or the URL is invalid. The page does not tell users which Kyuubi settings need to be changed.

This patch keeps the existing 403 status and host validation, but returns an actionable Kyuubi-styled page that shows the target engine and the relevant proxy and host allowlist settings.

- `dev/reformat`
- `build/mvn -pl kyuubi-server -am -Pspark-provided,hive-provided,flink-provided scalastyle:check`
- `build/mvn test -pl kyuubi-server -am -rf :kyuubi-server_2.12 -Pspark-provided,hive-provided,flink-provided -Dtest=none -DwildcardSuites=org.apache.kyuubi.server.api.EngineUIProxyServletSuite`
- Compared the previous and updated 403 responses through real Jetty servlets in a browser.

### Visual comparison

Before:

Jetty default 403 response

After:

Actionable Kyuubi Engine UI 403 response

Assisted-by: Codex with GPT-5

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/EngineUIProxyServlet.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/api/EngineUIProxyServlet.scala
@@ -20,7 +20,7 @@ package org.apache.kyuubi.server.api
 import java.net.{URI, URL}
 import java.util.Locale
 import java.util.regex.Pattern
-import javax.servlet.http.HttpServletRequest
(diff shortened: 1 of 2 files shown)


aajisaka (approved):
+1, thanks.

Built web-ui and verified the change locally.

pan3793:
thanks, merged to master

PR #7569, opened 2026-07-14
[KYUUBI #7568][AUTHZ] Eliminate TypeOfPlaceHolder across the whole plan

### LINKED ISSUE #7568: [Bug] typeof() fails with CLASS_NOT_OVERRIDE_EXPECTED_METHOD when it is not in the root plan node
### Code of Conduct

### Search before asking

### Describe the bug

With `RangerSparkExtension` enabled, `typeof()` fails whenever it is not in the root plan node:

```sql
CREATE TABLE t1 (id int, day string);

SELECT typeof(id) FROM t1;                                       -- works (root is Project)
SELECT typeof(id) AS t FROM t1 ORDER BY t;                       -- fails
SELECT typeof(id) FROM t1 UNION ALL SELECT typeof(day) FROM t1;  -- fails
SELECT typeof(id), count(*) FROM t1 GROUP BY typeof(id);         -- fails
```

```
[CLASS_NOT_OVERRIDE_EXPECTED_METHOD] org.apache.kyuubi.plugin.spark.authz.rule.expression.TypeOfPlaceHolder
must override either eval or nullSafeEval. SQLSTATE: 38000
```

`RuleApplyTypeOfMarker` marks every `TypeOf` in the whole plan (`transformAllExpressions`), but
since #5997 `RuleEliminateTypeOf` reverts them with `transformExpressionsUp`, which only visits the
root node's expressions — the outer `transformUp` removed there as an unnecessary loop was what
walked the plan. Any placeholder below the root survives into execution and fails, because
`TypeOfPlaceHolder` implements `doGenCode` but no `eval`.

Existing tests only cover queries whose root node is a `Project`, which is why this went unnoticed.

### Affects Version(s)

1.9.0 and later, master

j1wonpark:

Fixes #7568.

`RuleApplyTypeOfMarker` marks every `TypeOf` in the whole plan (`transformAllExpressions`), but since #5997 `RuleEliminateTypeOf` reverts them with `transformExpressionsUp`, which only visits the root node's expressions. A `TypeOfPlaceHolder` below the root — under `Sort`, `Union`, `Aggregate`, or a subquery — survives into execution and fails with `CLASS_NOT_OVERRIDE_EXPECTED_METHOD`, since it implements `doGenCode` but no `eval`.

This change uses `transformAllExpressions` on the eliminating side as well, so both rules cover the same scope in a single pass.

Affects 1.9.0 onwards, including master.

Added a test to `RangerSparkExtensionSuite` covering the root shapes the existing `Project`-root-only test misses: `Sort`, `Union`, `Aggregate`, and a subquery. It fails on master and passes with this change. Full `extensions/spark/kyuubi-spark-authz` suite: 657 passed, 0 failed.

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
@@ -1413,14 +1413,16 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
           s"SELECT typeof(id) AS t FROM $db1.$table1 ORDER BY t",
           Seq(Row("int")))
 
-        // root is Union
-        checkAnswer(
+        // root is Union. UNION ALL has no order guarantee, and an ORDER BY would put a Sort
+        // above the Union, so sort the collected rows instead to keep Union as the root node.
+        doAs(
           admin,
-          s"""
-             |SELECT typeof(id) FROM $db1.$table1
-             |UNION ALL
-             |SELECT typeof(day) FROM $db1.$table1""".stripMargin,
-          Seq(Row("int"), Row("string")))
+          assert(sql(
+            s"""
+               |SELECT typeof(id) FROM $db1.$table1
+               |UNION ALL
+               |SELECT typeof(day) FROM $db1.$table1""".stripMargin)
+            .collect().sortBy(_.getString(0)) === Seq(Row("int"), Row("string"))))
 
         // root is Aggregate
         checkAnswer(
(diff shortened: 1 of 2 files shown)


extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
```diff
@@ -1387,6 +1387,56 @@ class HiveCatalogRangerSparkExtensionSuite extends RangerSparkExtensionSuite {
...
+          Seq(Row("int")))
+
+        // root is Union
+        checkAnswer(
+          admin,
+          s"""
+             |SELECT typeof(id) FROM $db1.$table1
+             |UNION ALL
+             |SELECT typeof(day) FROM $db1.$table1""".stripMargin,
+          Seq(Row("int"), Row("string")))
```
pan3793: UNION ALL has no order guarantee, consider adding an ORDER BY to make it robust

pan3793 (approved):

pan3793 (approved):

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/RangerSparkExtensionSuite.scala
j1wonpark: Fixed in 38464f17f — sorted the collected rows on the Scala side to keep Union as the root node.

pan3793:
thanks, merged to master/1.11/1.10

PR #7575, opened 2026-07-17
Improve access path validation

aajisaka:

Normalize config keys before access path validation

Added unit tests

Assisted by: GPT-5.5


pan3793 (approved):
thank you @aajisaka, the fix is sound and correctly closes the bypass.

aajisaka:
Thank you! Merged into master and branch-1.12.

Issue #3581, opened 2022-09-29
[FEATURE] [Authz] support row filter and data masking on permanent views


bowenliang123:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

support row filter and data masking on permanent views.

### Are you willing to submit PR?

- [X] Yes. I can submit a PR independently to improve.
- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #7577, opened 2026-07-17
[KYUUBI #7576][AUTHZ] Fix data masking failure on a UNION-ALL view over a masked view

### LINKED ISSUE #7576: [Bug][AUTHZ] Data masking fails with MISSING_ATTRIBUTES on a UNION-ALL view over a masked view
### Code of Conduct

### Search before asking

### Describe the bug

Reading a permanent view that UNION-ALLs a masked permanent view fails analysis, while reading the masked view directly works (#3581):

```
[MISSING_ATTRIBUTES.RESOLVED_ATTRIBUTE_APPEAR_IN_OPERATION] Resolved attribute(s) "value2"
missing from "value2" in operator !Project [cast(value2#613 as string) AS value2#611].
```

With a view-level, type-preserving MASK policy on `perm_view_masked.value2`:

```sql
CREATE OR REPLACE VIEW default.perm_view_masked AS SELECT key, value2 FROM default.src;
CREATE OR REPLACE VIEW default.perm_view_union AS
  SELECT key, value2 FROM default.perm_view_masked WHERE key = 1
  UNION ALL
  SELECT key, value2 FROM default.perm_view_masked WHERE key = 10;

-- as the masked user
SELECT key, value2 FROM default.perm_view_union WHERE key = 1;   -- FAIL
SELECT key, value2 FROM default.perm_view_masked WHERE key = 1;  -- OK
```

The same view structure without a masking policy resolves fine, so the union/view shape itself is handled by Spark — only the masking rewrite breaks it.

Root cause (confirmed with `spark.sql.planChangeLog`): both UNION branches expand the masked view with identical output exprIds. After `RuleApplyDataMaskingStage0` marks both branches, `DeduplicateRelations` re-instances the second branch's `PermanentViewMarker`, but `DataMaskingStage0Marker.scan` is not a child and keeps the pre-dedup ids, so the second branch's `exprToMaskers()` becomes keyed on ids now owned by the *first* branch. `RuleApplyDataMaskingStage1` then merges all marker maps into one `toMap` (duplicate keys silently last-wins) and substitutes unconditionally, rewiring the outer view's schema compensation Project — above the Union — to attributes the Union never outputs.

Fix: scope the Stage1 substitution — rewrite the children first, then only substitute maskers the rewritten children actually expose, merging the per-marker maps after that filtering. Branch-local healing keeps working; the cross-branch leak is blocked at the Union boundary. All five data masking suites and the full kyuubi-spark-authz module pass with the fix.

### Affects Version(s)

master

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

_No response_

_No response_

_No response_

### Additional context

_No response_

j1wonpark:

Close #7576.

Reading a permanent view that UNION-ALLs a masked permanent view fails with `MISSING_ATTRIBUTES`, while reading the masked view directly works (#3581).

Both branches expand the masked view with identical output exprIds. After `RuleApplyDataMaskingStage0` marks both branches, `DeduplicateRelations` re-instances one branch, so that branch's `exprToMaskers()` ends up keyed on ids now owned by the other branch. `RuleApplyDataMaskingStage1` merges all marker maps into one `Map` (duplicate keys silently last-wins) and substitutes unconditionally, rewiring the outer view's schema compensation Project — above the Union — to attributes the Union never outputs. See #7576 for the full analysis.

This change scopes the Stage1 substitution: rewrite the children first, keep only the maskers the rewritten children actually expose (matched by exprId, since `SubqueryAlias` may requalify outputs), and merge the per-marker maps after that filtering. When a child is not resolved yet (a type-changing masker leaves Union branches incompatible until type coercion runs, and `Union.output` fails on them), the previous unfiltered substitution is kept. Branch-local substitution keeps working; the cross-branch leak is blocked at the Union boundary.

Two tests added to `DataMaskingTestBase` (run by all data masking suites): a reproduction that failed with `MISSING_ATTRIBUTES` before this fix, and a control proving the same view shape resolves fine without a masking policy. A view-level type-preserving MASK policy is added to the fixture, since a type-changing mask dies earlier with `CANNOT_UP_CAST` and would hide the bug.

```
build/mvn test -pl extensions/spark/kyuubi-spark-authz -Dtest=none \
  -DwildcardSuites=org.apache.kyuubi.plugin.spark.authz.ranger.datamasking.DataMaskingForHiveParquetSuite
```

All five data masking suites pass, and the full kyuubi-spark-authz module is green (662 succeeded / 0 failed / 42 pre-existing skips).

Assisted-by: Claude:claude-fable-5

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/datamasking/RuleApplyDataMaskingStage1.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/datamasking/RuleApplyDataMaskingStage1.scala
@@ -61,16 +61,36 @@ case class RuleApplyDataMaskingStage1(spark: SparkSession) extends RuleHelper {
       case m: DataMaskingStage0Marker => m
       case m: DataMaskingStage1Marker => m
       case p =>
-        val maskerExprs = p.collect {
+        val maskerExprsPerMarker = p.collect {
           case marker: DataMaskingStage0Marker if marker.resolved => marker.exprToMaskers()
-        }.flatten.toMap
-        if (maskerExprs.isEmpty) {
+        }
+        if (maskerExprsPerMarker.forall(_.isEmpty)) {
           p
         } else {
-          val t = p.transformExpressionsUp {
+          // KYUUBI #7576: marker maps are keyed by pre-masking exprIds, which can collide
+          // across UNION branches once DeduplicateRelations re-instances a repeated masked
+          // scan. Rewrite the children first, keep only the maskers the rewritten children
+          // actually expose, and merge the per-marker maps after that filtering so an
+          // out-of-scope entry cannot shadow the in-scope one on a colliding key. Maskers are
+          // matched by exprId since SubqueryAlias may requalify child outputs.
+          val planWithNewChildren =
+            p.withNewChildren(p.children.map(replaceOriginExprWithMasker))
+          // A type-changing masker (e.g. MASK_HASH) leaves a Union's branches incompatible
+          // until type coercion runs and Union.output fails on them, so the masker visibility
+          // is judged only against resolved children; otherwise keep the unfiltered
+          // substitution.
+          val maskerExprs = if (planWithNewChildren.children.forall(_.resolved)) {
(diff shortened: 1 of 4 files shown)


pan3793 (approved):

pan3793:
thanks, merged to master/1.12

PR #7581, opened 2026-07-22
[SPARK] Use the larger advisory partition size when both configs are set

ulysses-you:

When both `spark.sql.adaptive.rebalancePartitionsAdvisoryPartitionSizeInBytes`
and `spark.sql.finalStage.adaptive.advisoryPartitionSizeInBytes` are set,
`KyuubiSQLConf.getAdvisoryPartitionSize` previously let the rebalance config
unconditionally take precedence. This could pick a smaller advisory partition
size than the finalStage config even when the finalStage value is larger.

This changes the resolution so the larger of the two values is used when both
are set, which better reflects the intent of both knobs. The doc for the
rebalance config is updated accordingly. The change is applied to the
spark-3.5, 4.0 and 4.1 extension modules.

Updated `RebalanceBeforeWritingSuite` in all three modules to cover:
- both set, rebalance larger -> rebalance wins
- both set, finalStage larger -> finalStage wins
- both set to the same value

Ran the suite locally:

```
build/mvn test -pl extensions/spark/kyuubi-extension-spark-3-5 -am -Pspark-3.5 \
  -Dtest=none -DwildcardSuites=org.apache.spark.sql.RebalanceBeforeWritingSuite   # 14 passed
build/mvn test -pl extensions/spark/kyuubi-extension-spark-4-0 -am -Pspark-4.0 -Pscala-2.13 \
  -Dtest=none -DwildcardSuites=org.apache.spark.sql.RebalanceBeforeWritingSuite   # 12 passed
build/mvn test -pl extensions/spark/kyuubi-extension-spark-4-1 -am -Pspark-4.1 -Pscala-2.13 \
  -Dtest=none -DwildcardSuites=org.apache.spark.sql.RebalanceBeforeWritingSuite   # 12 passed
```

Yes.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
+++ b/extensions/spark/kyuubi-extension-spark-3-5/src/main/scala/org/apache/kyuubi/sql/KyuubiSQLConf.scala
@@ -259,7 +259,8 @@ object KyuubiSQLConf {
     buildConf("spark.sql.adaptive.rebalancePartitionsAdvisoryPartitionSizeInBytes")
       .doc("The advisory partition size for RebalancePartitions operator. When set, it will be " +
         "passed to `optAdvisoryPartitionSize` of the RebalancePartitions node. " +
-        "It takes precedence over `spark.sql.finalStage.adaptive.advisoryPartitionSizeInBytes`.")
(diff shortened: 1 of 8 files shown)


pan3793:
4.2 is missing?

pan3793 (approved):

ulysses-you:
thanks, merging to master

Issue #7589, opened 2026-07-27
[Bug][AUTHZ] Rejecting RESET on the restricted configs parameters


attilapiros:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The `AuthzConfigurationChecker` only validates setting configuration parameters to explicit values (`SET <key>=<value>`). It currently fails to check actions that restore default values or remove configurations entirely, such as `RESET <key>` or `RESET` (which reset all the config parameters).

For more details, see the: https://spark.apache.org/docs/latest/sql-ref-syntax-aux-conf-mgmt-reset.html

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

### Kyuubi Server Configurations

```yaml

```

### Kyuubi Engine Configurations

```yaml

```

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

PR #7591, opened 2026-07-27
[KYUUBI #7589][AUTHZ] Rejecting RESET on the restricted configs parameters

attilapiros:

The `AuthzConfigurationChecker` only validates setting configuration parameters to explicit values (`SET <key>=<value>`). It currently fails to check actions that restore default values or remove configurations entirely, such as `RESET <key>` or `RESET` (which reset all the config parameters).

For more details, see the: https://spark.apache.org/docs/latest/sql-ref-syntax-aux-conf-mgmt-reset.html

A unit test was extended.

Yes.

Assisted-by: Claude:claude-sonnet


pan3793 (approved):

pan3793:
thanks, merged to master

Issue #3515, opened 2022-09-19
[Bug] [authZ]can't update iceberg table


lordk911:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/incubator-kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

env: spark:3.2.1 \ authz:master

1、complie authz from master branch
clean package -pl :kyuubi-spark-authz_2.12 -DskipTests -Dspark.version=3.2.1 -Dranger.version=1.2.0
2、put kyuubi-spark-authz_2.12-1.7.0-SNAPSHOT.jar to $SPARK_HOME/jars
3、config spark
spark.sql.extensions org.apache.kyuubi.plugin.spark.authz.ranger.RangerSparkExtension,org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,org.apache.kyuubi.sql.KyuubiSparkSQLExtension

spark.sql.catalog.iceberg_catalog org.apache.iceberg.spark.SparkCatalog
spark.sql.catalog.iceberg_catalog.type hive

spark.sql.catalog.spark_catalog org.apache.iceberg.spark.SparkSessionCatalog
spark.sql.catalog.spark_catalog.type hive

4、use spark-shell
5、spark.sql("create table test.iceberg919(id bigint, name string) USING iceberg")
6、spark.sql("INSERT INTO test.iceberg919 VALUES (1, 'a'), (2, 'b'), (3, 'c')")
7、spark.sql("select * from test.iceberg919").show
```
+---+----+                                                                      
| id|name|
+---+----+
|  1|   a|
|  2|   b|
|  3|   c|
+---+----+
```
8、
```
spark.sql("update test.iceberg919 set name='aa' where id=1")
spark.sql("update iceberg_catalog.test.iceberg919 set name='aa' where id=1")
spark.sql("update spark_catalog.test.iceberg919 set name='aa' where id=1")
``` 

**all got error with : `is not an Iceberg table`**

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk
spark.sql("update test.iceberg919 set name='aa' where id=1")
org.apache.spark.sql.AnalysisException: Project [id#25L, name#26]
+- RowFilterAndDataMaskingMarker
   +- RelationV2[id#25L, name#26] spark_catalog.test.iceberg919
[...]

PR #7594, opened 2026-07-28
[KYUUBI #7593][FEATURE][AUTHZ] "Paranoid mode", try to catch missing authz checks

### LINKED ISSUE #7593: [FEATURE][AUTHZ] "Paranoid mode", try to catch missing authorization checks
### Code of Conduct

### Search before asking

### Describe the feature

Add an opt-in, _runtime_ enforcement mode to the Spark authz plugin that makes unrecognized plan nodes fail closed (i.e. throw an authz exception) instead of falling through the match and silently contributing no access request. 

Plus, at _build time_, catch classification drift by checking that every plan node class found in the classpath is one of:
 * Explicitly allow-listed as "known harmless"
 * Durably asserted by FQCN to be checked properly, in each Spark version -- in text files, not easily-ignored tests
 * Not a `Command` or `LeafNode`

Any newly introduced plan node type would need to be allow-listed, or manually asserted as safe (hopefully with careful PR review!)

### Runtime config:
`spark.kyuubi.authz.unclassifiedNode.behavior` = `allow` | `warn` | `deny`   (default: warn)

 * `allow`: today's behavior; violations counted and visible at DEBUG.
 * `warn`: log once per (class name, violation kind) per JVM and count occurrences. The accretion mode for building an allowlist from real workloads.
 * `deny`: throw `AccessControlException` naming the unclassified class and the config key.

The enforced invariant is deliberately narrower than "flag every fallthrough", which would be unusably noisy since intermediate operators legitimately recurse:

Every `Command`, every `LeafNode`, every node that can execute or mutate state outside the checked path (e.g. Spark 4's `ExecutableDuringAnalysis`), and every node whose class name has a spec that dispatch did not consult — including in subtrees pruned by constant-projection elimination — must be either matched by a spec or present on an explicit allowlist. Ordinary non-leaf query operators recurse freely.

### Motivation

In the authz plugin, recognition of plan nodes is the first and mandatory step to forming a security boundary, and non-recognition (e.g. a `match` that falls through) fails open. Privilege building walks the Catalyst plan and matches nodes against the JSON spec files; any node that falls through is implicitly treated as not-authz-relevant and contributes no access request. Spark's plan space is open, meaning that new commands appear in every minor release, and third-party catalogs inject their own nodes at user runtime, so the unrecognized set grows silently.

A green test suite is unfortunately insufficient to bound this risk: every test was written for a node type someone had already classified, so the tests certify that the previously known sample still authorizes correctly, but can't verify nodes that nobody has written a test for yet.

Unfortunately, this is not hypothetical. Running the existing suites in deny mode against master immediately surfaced two real fail-opens, both fixed and regression-tested in the patch:

1. Iceberg metadata tables (`SELECT * FROM t.snapshots`) were never authorized: the 4-part `name()` threw a `MatchError` that fail-open swallowed.
2. Iceberg `MERGE INTO`: the rewrite embeds an already-planned `DataSourceV2ScanRelation` for its read of the target table, which `buildQuery` skipped entirely.

The sharpest structural case is `CALL` on Spark 4. On Spark 3.x with Iceberg it resolves to `o.a.s.sql.catalyst.plans.logical.Call`, an Iceberg-injected `Command`, so dispatch reaches its spec. Spark 4 ships its own class under the identical fully qualified name with a different hierarchy — a `UnaryNode` implementing `ExecutableDuringAnalysis`, not a `Command`.

Dispatch never reached the spec lookup and the procedure executed during analysis, before any optimizer-phase check could run. The spec entry still existed and still named the right class; it was simply unreachable. Neither a class-name check nor the presence of a spec entry detects this.

When the build-time enumeration first ran, it counted 136 unclassified authz-relevant plan classes on Spark 3.5 and 195 on Spark 4.1 — the population fail-open had been hiding.

The gap between "policies are enforced on every operation the plugin recognizes" and "policies are enforced on every operation" is something very few users can audit from the outside.

### Describe the solution

Four fail-open layers; the fourth was found during implementation.

1. Unknown commands: `buildCommand`'s fallthrough returned `QUERY` with zero privilege objects, so `RuleAuthorization` never called `verify`.
2. Unknown leaf relations: an unmatched leaf (or a known scan node not resolved) fell into the generic recursive arm and contributed nothing.
3. Extractor drift on known commands: extraction was wrapped in `catch { case e: Exception => LOG.debug(...); Nil }`, so a spec'd command failed open when its extractors broke against a new Spark version.
4. Constant-projection pruning: `buildQuery` skips the subtree of a constant Project, but the subtree still executes (`SELECT 'x' FROM t`).

Supporting pieces: an allowlist with a required `reason` per entry (`Command`s take a second colocated review); per-command extraction-failure semantics, so single descriptor failures stay expected variance rather than drift; dispatch hardening so a spec'd class routes to buildCommand even when it isn't a `Command` on this version; `verifiedSparkVersions` as an explicit enumeration of exact major.minor pairs, never a range — gating for allowlist entries, advisory for specs; and build-time coverage checks that enumerate every concrete `LogicalPlan` descendant into exactly one of four buckets, failing the build on any class new to the diff.

These do not replace the runtime mode, for two independent reasons: build-time enumeration cannot see third-party catalog plugins loaded only in the user's environment, and the set of dangerous node shapes is open, so the only sound default is "unrecognized ⇒ deny".

### Additional context

Implemented and green in deny mode on all four profiles (3.5, 4.0, 4.1, 4.2); backlogs 135/182/195/208. Design doc ships in the patch.

Known limitations stated up front: 
 * analysis-time execution is unreachable at runtime (`CALL` on Spark 4 is an acknowledged gap, not a fix)
 * partial extractor drift isn't reported at runtime
 * row-filtering and data-masking traversals aren't covered and want the same treatment as a follow-up
 * a startup Spark-major assertion is worth considering separately.

Three questions for the community: 
1. is `warn` the right default or should the first release ship `allow`?; 
2. should the allowlist ship in-repo or be user-supplied?
3. should the per-profile backlog files be committed or generated on demand?

acruise:
Fixes https://github.com/apache/kyuubi/issues/7593.

Privilege building walks the logical plan and classifies each node: it has a command spec, a scan spec, or it is a pass-through. A node matching none of these is currently silent: the walk continues and the statement executes with whatever privileges the rest of the plan happened to require. Nothing logs at a level an operator would see, so an unauthorized plan shape is indistinguishable from an authorized one.

That silence is load-bearing in two ways. A plan node can change shape under the same fully qualified name between Spark releases (`CALL` changed supertype between Spark 3 and 4 and stopped matching its dispatch arm), and a third-party catalog can introduce nodes the spec set has never seen. In both cases authorization stops happening rather than fails.

Running the existing suites with unclassified nodes set to `deny` surfaced two live fail-opens on master. Both are fixed here with regression tests:

1. **Iceberg metadata tables were never authorized.** `SELECT * FROM db.t.snapshots` (and `.history`, `.manifests`, ...) reports a four-part `name()`. `StringTableExtractor` matches only one-, two- and three-part names, so a four-part name throws `MatchError`; `CommandSpec` catches `case e: Exception`, logs at `debug`, and returns `None`. The table therefore carries no access request and the read is not authorized at all. Iceberg metadata tables expose table history, manifest listings and file paths. `TableTableExtractor` now reflectively unwraps `BaseMetadataTable` to its base table, so the read is authorized as a read of the table it exposes.
2. **Iceberg `MERGE INTO` skipped its read of the target.** The rewrite embeds an already-planned `DataSourceV2ScanRelation` for that read, which the plan builder passed over silently. It now has its own scan spec.

`TableTableExtractor` is byte-identical in `v1.10.3`, `v1.11.1` and current `master`, so both gaps are present in released versions, not just on master.

The main change adds `spark.kyuubi.authz.unclassifiedNode.behavior` (`allow` | `warn` | `deny`, default `warn`) so an operator can choose to fail closed on an unclassified node, plus the build-time accounting that makes the classification set auditable rather than assumed: every concrete `LogicalPlan` descendant on the classpath is placed in exactly one of four buckets — spec'd, allowlisted, pass-through, or a per-Spark-minor backlog file that is
checked in and reviewed at PR time.

Two design points worth reviewer attention:

- Allowlist entries require a written `reason`. An entry grants silence, so it should record a reviewed decision rather than a reflexive silencing.
- Every entry names the exact Spark `major.minor` versions its review covers, as an explicit enumeration rather than a range. Ranges invite boundary misreadings ("less than 4.0,
  exclusive" read as inclusive); an enumeration has no boundary to misread and makes a new Spark minor unverified by default, which is exactly when re-review is due. Allowlist entries *gate* on this — unverified means inert, so the node counts as unclassified and fails closed. Command and scan specs are advisory, because a spec imposes checks and staying active on an unaudited version is the safe direction.

Specs added before the Spark 4 port take their versions from a frozen ledger (`spec_verified_spark_versions.txt`) that records an inherited `3.3/3.4/3.5` baseline rather than a per-minor audit; the ledger header says so plainly and is closed, so a new spec must declare its own versions rather than inherit that baseline.

`docs/paranoid-mode.md` carries the full design, including the known limitation that `RuleAuthorization` is an optimizer rule and so cannot stop an `ExecutableDuringAnalysis` node (Spark 4 `CALL`) that has already run during analysis — which is what the build-time check exists to cover.

Default behaviour is unchanged: `warn` logs and proceeds.

`ParanoidModeSuite` (15 tests) covers the policy: the `warn` default, case-insensitive parsing, an invalid value rejected loudly, and each behaviour end to end — `deny` failing closed on an unclassified leaf relation and on an unclassified command, `warn` passing but counting every occurrence, `allow` passing silently. It also covers the shapes that made this worth doing: an unclassified node cannot hide under a constant projection, scan spec extraction failures are surfaced rather than swallowed, allowlisted nodes still pass under `deny`, and ordinary multi-operator queries recurse freely. Allowlist hygiene is tested too — entries require a reason and explicitly enumerated versions rather than ranges, and an entry applies only to the Spark minors it was reviewed against.

`ClassificationCoverageSuite` (5 tests) does the build-time accounting: every plan node class on the classpath is accounted for, every command spec entry is routable to `buildCommand`, allowlisted classes are not Commands in disguise, allowlist entries are shapes the classifier would actually flag (so an entry for a pass-through shape is rejected as dead weight that reads as coverage), and no class has both a command spec and an allowlist entry.

For the two fail-opens:

- The Iceberg metadata table gap has a dedicated regression test, `selecting an Iceberg metadata table requires select on the base table`, asserting both directions — denied for an unprivileged user, allowed for one with select on the base table. It fails without the `TableTableExtractor` change.
- The `MERGE INTO` gap has no new test of its own; it is covered by the existing `[KYUUBI #3515] MERGE INTO` test now running under `deny`. Without the new `DataSourceV2ScanRelation` scan spec the embedded relation is unclassified, so that test fails. This holds on the 3.5, 4.0 and 4.1 profiles; Iceberg is tag-excluded on 4.2.

The ledger's two guards — a spec that neither declares `verifiedSparkVersions` nor appears in the ledger, and a ledger entry whose spec no longer exists — are assertions inside the existing `check spec json files` generator test rather than standalone tests. Both fail generation.

Existing tests: all suites now run with sessions in `deny` mode (`SparkSessionProvider`), so any unclassified node reached by an existing test fails that test rather than passing silently. That is the main verification vehicle for the classification set, and it is how both fail-opens above were found.

```bash
# full module, default profile (Spark 3.5)
build/mvn test -pl extensions/spark/kyuubi-spark-authz -am

# the new suites alone
build/mvn test -pl extensions/spark/kyuubi-spark-authz -am -Dtest=none \
  -DwildcardSuites=org.apache.kyuubi.plugin.spark.authz.ParanoidModeSuite,org.apache.kyuubi.plugin.spark.authz.ClassificationCoverageSuite

# other Spark profiles
build/mvn test -pl extensions/spark/kyuubi-spark-authz -am -Pspark-4.0 -Pscala-2.13
build/mvn test -pl extensions/spark/kyuubi-spark-authz -am -Pspark-4.1 -Pscala-2.13
build/mvn test -pl extensions/spark/kyuubi-spark-authz -am -Pspark-4.2 -Pscala-2.13

# regenerate the spec JSON and backlog files after changing a spec
dev/gen/gen_ranger_spec_json.sh
```

Results: the default profile passes on this branch (697 tests). The one failure in that run is unrelated to this change and reproduces on `master` — Hudi's embedded timeline server timing out to a LAN address in
`DeleteHoodieTableCommand/UpdateHoodieTableCommand/MergeIntoHoodieTableCommand`. The 4.0, 4.1
and 4.2 profiles were green in `deny` mode before this branch was rebased onto current master and are being re-run against it; I will confirm here before asking for a merge.

Assisted-by: Claude:claude-opus-5

Used for code review, rebasing onto master, and drafting; every line reviewed by me.


acruise:
cc: @alexandrefimov @hit2024 :)

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ParanoidMode.scala
```diff
@@ -0,0 +1,155 @@
...
+     * (fail-open layer 3, typically Spark version drift).
+     */
+    val EXTRACTION_FAILURE: Value = Value("spec matched but extraction failed")
+  }
+
+  import Behavior._
+  import ViolationKind.ViolationKind
+
+  def behavior(spark: SparkSession): Behavior = {
+    val raw = spark.conf.getOption(UNCLASSIFIED_NODE_BEHAVIOR_KEY).getOrElse(WARN.toString)
```
alexandrefimov: This makes `deny` controllable by the client because the value is read from the session-level `spark.conf`. A user can set `spark.kyuubi.authz.unclassifiedNode.behavior=allow` before running the query, so an operator cannot use `deny` as an authorization boundary.

Could we read this setting only from the application/engine configuration, for example `SparkContext.getConf`, and add negative tests for attempts to change it through SQL `SET` and the Spark Connect config API?

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/gen/KnownHarmlessNodes.scala
```diff
@@ -0,0 +1,145 @@
...
+      "Wraps a session-created RDD/local collection (e.g. spark.createDataset, and Delta's" +
+        " internal VACUUM plumbing); RDD-level access is outside the plugin's scope and is" +
+        " an existing, separate concern",
+      spark3xAnd4x),
+    HarmlessNodeSpec(
+      "org.apache.spark.sql.execution.LogicalRDD",
+      "Wraps a pre-existing RDD; RDD-level access is outside the plugin's scope and is an" +
+        " existing, separate concern",
+      spark3xAnd4x),
+    HarmlessNodeSpec(
```
alexandrefimov: I do not think `InMemoryRelation` is harmless across users. Spark sessions can share the cache, while authorization must be evaluated for the current user. In that case the optimized plan is an `InMemoryRelation` leaf and no table privilege is produced.

The existing cache test confirms that another session gets `InMemoryRelation`, but both sessions run inside the same `doAs(admin, ...)` block and the test only checks the authz tag.

Could we add a regression test where an admin populates the cache and an unprivileged user reads the same table from a new session? The read should still be denied. This probably requires authorizing the original relation instead of allowlisting `InMemoryRelation`.

alexandrefimov (commented):
Thanks for working on this. I am not a Kyuubi maintainer, so please treat this as an external security review. I reviewed the current head with focus on the fail-closed behavior.

I found two cases that look like authorization bypasses and left inline comments:

1. The client can change the paranoid-mode behavior through the session configuration.
2. A cached `InMemoryRelation` can lose the original table authorization when the cache is reused by another user.

I think these should be fixed before `deny` mode is presented as a reliable authorization boundary.

A few other points also need attention:

- The classification backlog is keyed only by the Spark minor version. The Spark 3.5 / Scala 2.13 CI job already gets a different set of classes than Spark 3.5 / Scala 2.12, so one golden file cannot represent both profiles. The backlog should include the relevant build profile, or the enumeration should be limited to a stable classpath.
- The Iceberg metadata-table test expects a privilege on `base_table/committed_at`, although `committed_at` is not a column of the base table. The positive case uses only `admin`. Please add a non-admin positive test for a user who has normal SELECT access to the base table and define how metadata columns should map to Ranger privileges.
- The document correctly says that Spark 4 `CALL` can execute during analysis, before the authorization rule runs. Unwrapping `CommandResult` cannot prevent a side effect that already happened. This needs an analysis-time check or a hard block; otherwise `CALL` should be explicitly excluded from the fail-closed guarantee.
- The classification-only sweep finds a table behind a constant projection, but it does not create a privilege object. Therefore `SELECT 1 FROM protected_table` can still produce no table authorization request. The documentation should not say that this gap is fully addressed unless table-level authorization is added and tested.

The current branch also conflicts with master, which has dropped Spark 3.3 and 3.4. The old failures for those profiles are no longer relevant, but the Spark 3.5 / Scala 2.13 backlog failure and the RAT failure for `spec_verified_spark_versions.txt` still need fixes. After rebasing, the full Spark and Scala profile matrix should be rerun.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/ParanoidMode.scala
acruise: thanks, 32a79a9625e1e7350621d663ea29435e47e70eeb :)

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/gen/KnownHarmlessNodes.scala
acruise: thanks, 32a79a9625e1e7350621d663ea29435e47e70eeb :)

PR #6677, opened 2024-09-06
Bump scala-collection-compat to 2.12.0

pionCham:
# :mag: Description
## Describe Your Solution 🔧
-   Bump `scala-collection-compat` to 2.12.0 
    - `scala-collection-compat` is used by the authz plugin testing Iceberg
    - Currently Iceberg 1.6.1 uses `scala-collection-compat` 2.12.0,
according to: https://github.com/apache/iceberg/blob/apache-iceberg-1.6.1/gradle/libs.versions.toml#L78
    -  release note: https://github.com/scala/scala-collection-compat/releases/tag/v2.12.0

## Types of changes :bookmark:

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Related Unit Tests
- pass the Authz tests

---

# Checklist 📝

- [x] This patch was not authored or co-authored using [Generative Tooling](https://www.apache.org/legal/generative-tooling.html)

**Be nice. Be informative.**

PR #7600, opened 2026-07-29
Drop unused scala-collection-compat dependency

bowenliang123:

`scala-collection-compat` is no longer used anywhere in Kyuubi:

- **No source code references it**: `grep -rn 'import scala.collection.compat' --include='*.scala'` returns zero hits across the entire repo.
- **No transitive dep needs it from us**: Iceberg `1.11.0` (current `iceberg.version` per `pom.xml:167`) no longer ships `scala-collection-compat` in its runtime jar; checked `iceberg-spark-runtime-3.5_2.12-1.11.0.jar` and the `scala/collection/compat` package is absent.
- **The 2024 pin is stale**: the original rationale (commit `550f1fce8`, #6677) was that Iceberg 1.6.1 transitively pulled `scala-collection-compat:2.12.0`. Iceberg has since moved on.
- **Scala 2.13 path is also unaffected**: `pom.xml:124` `<scala-collection-compat.version>` lives outside any profile, so both `scala-2.12` and `scala-2.13` (lines 2003/2011) shared the same pin. Neither path needs it.

3 hunks across 2 files:

- `pom.xml`: drop `<scala-collection-compat.version>` property
- `pom.xml`: drop the corresponding `<dependencyManagement>` entry
- `extensions/spark/kyuubi-spark-authz/pom.xml`: drop the `<scope>test</scope>` dependency

- [x] `mvn -Pfast clean package -DskipTests -pl extensions/spark/kyuubi-spark-authz -am` — 7 modules build cleanly (kyuubi-parent, kyuubi-util, kyuubi-util-scala, kyuubi-common, kyuubi-events, kyuubi-spark-lineage, kyuubi-spark-authz)
- [x] `mvn -pl extensions/spark/kyuubi-spark-authz dependency:tree` confirms `scala-collection-compat_2.12:2.7.0` still lands on the classpath transitively (via Spark SQL `provided` scope) — classpath behaviour unchanged, just let Maven dependency mediation pick the version Spark wants instead of pinning an older 2.12.0
- [ ] full unit-test pass via `mvn test` to follow before merge

Yes.
Assisted-by: OpenCode with MiniMax-M3

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/pom.xml
+++ b/extensions/spark/kyuubi-spark-authz/pom.xml
@@ -370,12 +370,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.scala-lang.modules</groupId>
-            <artifactId>scala-collection-compat_${scala.binary.version}</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.paimon</groupId>
             <artifactId>${paimon.artifact}</artifactId>
             <scope>test</scope>
--- a/pom.xml
+++ b/pom.xml
@@ -121,7 +121,6 @@
         <enforcer.maxJdkVersion>8</enforcer.maxJdkVersion>
         <scala.version>2.12.19</scala.version>
         <scala.binary.version>2.12</scala.binary.version>
-        <scala-collection-compat.version>2.12.0</scala-collection-compat.version>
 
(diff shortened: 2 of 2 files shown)


pan3793:
spark 3.3 still uses old iceberg 1.6 for testing ...

bowenliang123:
Sorry that it's missed out. Closing this PR.

PR #7601, opened 2026-07-29
[Web] Bump Node.js to v24 LTS and pnpm to v11

bowenliang123:

Stay on a supported LTS line and pick up the latest pnpm stable. Node v22 has moved to Maintenance LTS (security fixes only, EOL 2027-04-30); v24 is the current Active LTS and is supported until 2028-04-30. pnpm v12 is still beta, so the latest v11 stable is the right target.

| Tool    | Before   | After            | Released   |
| ------- | -------- | ---------------- | ---------- |
| Node.js | v22.23.1 | **v24.11.1** (Active LTS) | 2025-11-11 |
| pnpm    | v9.11.0  | **v11.18.0**     | 2026-07-29 |

Migration notes:

- pnpm v11 no longer reads the `pnpm` field in `package.json`; moved the security overrides to a new `kyuubi-server/web-ui/pnpm-workspace.yaml`.
- pnpm v11's `strictDepBuilds` defaults to true; explicitly allow `@parcel/watcher` and `vue-demi` via `allowBuilds` since Vite's toolchain depends on their install/postinstall scripts.
- `engines.node` simplified to `>=24.0.0`.

- Local `pnpm install` against the updated toolchain succeeded with no resolution drift (lockfile content byte-identical to v9.11.0).
- `pnpm run build` and `pnpm run lint` pass locally.
- GitHub `web-ui` and `style` workflows both read `node.version` / `pnpm.version` from `pom.xml` and will exercise the new versions end-to-end.

Yes. Assisted-by: OpenCode with MiniMax-M3

### DIFF
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -49,6 +49,7 @@ build/scala-*/**
 **/web-ui/dist/**
 **/web-ui/coverage/**
 **/pnpm-lock.yaml
+**/pnpm-workspace.yaml
 **/node_modules/**
 **/gen/*
 **/*.tokens
--- a/kyuubi-server/web-ui/README.md
+++ b/kyuubi-server/web-ui/README.md
@@ -2,7 +2,7 @@
 
 ### Start Using
 
-Node.js 18 or higher is required.
+Node.js 24 or higher is required.
 You can learn how to install the corresponding version from its official website.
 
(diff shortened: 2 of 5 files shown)


pan3793 (approved):

pan3793:
thanks, merged to master/1.12

PR #7607, opened 2026-08-03
Support configurable trust certificate for LDAPS authenticator

zhaohehuhu:

Kyuubi already supports LDAPS when the JVM global truststore is configured. This pr adds native LDAP SSL truststore configurations so deployments can provide custom trusted certificates without relying on global Java SSL properties.

build/mvn test -pl kyuubi-common -Dtest=none -DwildcardSuites=org.apache.kyuubi.service.authentication.ldap.LdapSearchFactorySuite


zhaohehuhu:
@pan3793 @wForget  Could you plz help review this PR ? Thanks!

PR #7609, opened 2026-08-04
Performance and allocations improvement in TRowSet generation

bowenliang123:

`TRowSet` generation is on the hot path of every result fetch. This patch removes pure-overhead allocations, with no wire-protocol/API change and byte-identical output:

- `getColumnToList`: `foreach` closure -> `while` loop (row index and null bitmap stay in registers instead of captured heap `IntRef`s), hoist the `convertFunc == null` check out of the loop, pre-size the null `BitSet` to `rowSize` (was default 64-bit, paying growth copies).
- `asFloatTColumn` / `asByteArrayTColumn`: two passes -> one. The old code built an intermediate boxed list (`JList[JFloat]` / `JList[Array[Byte]]`) and re-mapped it via JavaConverters; now `convertFunc` produces `JList[JDouble]` / `JList[ByteBuffer]` directly. The float->string->double conversion (float display semantics) is unchanged.
- `toColumnBasedSet`: the `rows` list is always empty in columnar layout — create it with zero capacity instead of `rowSize`.
- Spark engine `RowSet.toHiveString`: the `DynMethods` chain (two reflective method-table scans) was rebuilt on every cell of a complex-typed column (Date/Timestamp/Decimal/Array/Map/Struct). Build the bound method once in a `private val`; only `invokeChecked` runs per cell.

No new tests: behavior-preserving refactor; existing engine `RowSetSuite`s (Spark/Flink/Trino) cover these paths in CI.

Verified with a side-by-side harness: old vs new implementations over 100k-row columns, asserting value lists and null bitmaps are equal (string/integer/float/binary, with/without `convertFunc`, 0%/10% nulls). Numbers below: JDK 17, 50 interleaved samples, median; allocation via `ThreadMXBean.getCurrentThreadAllocatedBytes`.

| Scenario | Time (old -> new) | Allocation (old -> new) |
|---|---|---|
| String column, 10% nulls, no convertFunc | 0.41 -> 0.41 ms (~1.0x) | 435 -> 415 KB/op (-20 KB, BitSet pre-sizing) |
| Integer column (boxed), 10% nulls, no convertFunc | 0.38 -> 0.38 ms (~1.0x) | 435 -> 415 KB/op (-20 KB) |
| Float column (`asFloatTColumn`), 10% nulls | 8.39 -> 7.51 ms (**1.12x**) | 14.6 -> 12.6 MB/op (**1.16x**, intermediate `JFloat` list eliminated) |
| Binary column (`asByteArrayTColumn`), 10% nulls | 1.41 -> 0.90 ms (**1.57x**) | 6.9 -> 5.3 MB/op (**1.30x**) |
| `toColumnBasedSet` rows list (per result set) | — | 391 KB saved per 100k rows, 3.9 MB per 1M rows (retained, not garbage) |
| Spark `RowSet.toHiveString`, DateType cell | 1.05 -> 0.51 µs (**2.0x**) | 920 -> 608 B/cell (**1.5x**) |
| Spark `RowSet.toHiveString`, TimestampType cell | 0.57 -> 0.36 µs (**1.6x**) | 949 -> 635 B/cell (**1.5x**) |
| Spark `RowSet.toHiveString`, Decimal cell | 0.18 -> 0.11 µs (**1.7x**) | 478 -> 184 B/cell (**2.6x**) |
| Spark `RowSet.toHiveString`, ArrayType cell | 0.63 -> 0.36 µs (**1.8x**) | 1232 -> 944 B/cell (**1.3x**) |

The `toHiveString` numbers were measured against the real Spark 3.5.8 `HiveResult` (not a stub); outputs of old and new implementations are asserted equal per case.

Also: `build/mvn -Pfast clean install -pl kyuubi-common -am -DskipTests`, `spotless:check`, `dev/reformat`.

Assisted-by: Kimi Code CLI (Moonshot AI)

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/schema/RowSet.scala
@@ -37,11 +37,7 @@ object RowSet {
       .buildChecked(HiveResult)
       .invokeChecked[BinaryFormatter]()
 
-  def toHiveString(
-      valueAndType: (Any, DataType),
-      nested: JBoolean = false,
-      timeFormatters: TimeFormatters,
-      binaryFormatter: BinaryFormatter): String =
+  private val toHiveStringMethod =
     DynMethods.builder("toHiveString")
       .impl( // for Spark 3.5 and before
         HiveResult.getClass,
@@ -55,5 +51,11 @@ object RowSet {
         classOf[TimeFormatters],
         classOf[BinaryFormatter])
       .buildChecked(HiveResult)
-      .invokeChecked[String](valueAndType, nested, timeFormatters, binaryFormatter)
+
+  def toHiveString(
+      valueAndType: (Any, DataType),
+      nested: JBoolean = false,
+      timeFormatters: TimeFormatters,
+      binaryFormatter: BinaryFormatter): String =
+    toHiveStringMethod.invokeChecked[String](valueAndType, nested, timeFormatters, binaryFormatter)
 }
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TColumnGenerator.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/engine/result/TColumnGenerator.scala
@@ -20,8 +20,6 @@ import java.lang.{Boolean => JBoolean, Byte => JByte, Double => JDouble, Float =
 import java.nio.ByteBuffer
 import java.util.{ArrayList => JArrayList, BitSet => JBitSet, List => JList}
 
-import scala.collection.JavaConverters._
-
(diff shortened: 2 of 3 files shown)


pan3793:
Be careful with Scala 2.12/2.13 collection api changes, we experienced several O(n) => O(n^2) regressions due to them.

bowenliang123:
@pan3793 Good point — that's exactly why the patch removes JavaConverters altogether and sticks to iterator-based loops over java.util collections. No Scala collection conversions or indexed Seq access remain, so behavior and complexity are the same on 2.12 and 2.13.

bowenliang123:
The performance improvments are achieved mostly by skipping repeated method refelection in Spark and the intermediate transformation for float and byte columns.

aajisaka (approved):
LGTM, thanks.

pan3793:
sorry, forget this one ... I will take a detailed look today

bowenliang123:
Thanks for the review. It's no hurry. Let's have further test and insights of it for iterators, collection apis.

PR #7622, opened 2026-08-11
[KYUUBI #6403][KSHC] Fix write into partitioned table with non-last partition column

### LINKED ISSUE #6403: [Bug] [spark hive connector] failed to create table is partition column is not at the end
### Code of Conduct

### Search before asking

### Describe the bug

create table failed if the partition column is not last column. could be reproduced

```scala
import org.apache.spark.sql.{Row, SparkSession}
import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
import scala.collection.mutable.ListBuffer

    val schema = StructType(Array(
      StructField("name", DataTypes.StringType, nullable = false),
      StructField("favorite_color", DataTypes.StringType, nullable = false),
      StructField("favorite_numbers", DataTypes.StringType, nullable = false)
    ))

    val data = ListBuffer[Row]()
    data += Row("Alyssa", "blue", "1")
    data += Row("Ben", "red", "2")

    val usersDF = spark.createDataFrame(spark.sparkContext.parallelize(data), schema)

   // "favorite_color" is not last column
   usersDF.write.partitionBy("favorite_color").saveAsTable("users3")
```

### Affects Version(s)

1.8.1

### Kyuubi Server Log Output

_No response_

### Kyuubi Engine Log Output

```logtalk
Caused by: org.apache.hadoop.hive.ql.metadata.HiveException: Exception when loading 2 in table users4 with loadPath=hdfs://localhost:9000/user/hive/warehouse/users4/.hive-staging_hive_2024-05-21_18-42-55_548_3233995390238788603-1/-ext-10000
	at org.apache.hadoop.hive.ql.metadata.Hive.loadDynamicPartitions(Hive.java:1970)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.spark.sql.hive.client.Shim_v2_1.loadDynamicPartitions(HiveShim.scala:1605)
	at org.apache.spark.sql.hive.client.HiveClientImpl.$anonfun$loadDynamicPartitions$1(HiveClientImpl.scala:977)
	at org.apache.spark.sql.hive.client.HiveClientImpl.$anonfun$withHiveState$1(HiveClientImpl.scala:303)
	at org.apache.spark.sql.hive.client.HiveClientImpl.liftedTree1$1(HiveClientImpl.scala:234)
	at org.apache.spark.sql.hive.client.HiveClientImpl.retryLocked(HiveClientImpl.scala:233)
	at org.apache.spark.sql.hive.client.HiveClientImpl.withHiveState(HiveClientImpl.scala:283)
	at org.apache.spark.sql.hive.client.HiveClientImpl.loadDynamicPartitions(HiveClientImpl.scala:968)
	at org.apache.spark.sql.hive.HiveExternalCatalog.$anonfun$loadDynamicPartitions$1(HiveExternalCatalog.scala:966)
	at org.apache.spark.sql.hive.HiveExternalCatalog.withClient(HiveExternalCatalog.scala:101)
	... 87 more
Caused by: java.util.concurrent.ExecutionException: org.apache.hadoop.hive.ql.metadata.Table$ValidationFailureSemanticException: Partition spec {favorite_color=, favorite_numbers=2} contains non-partition columns
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
	at org.apache.hadoop.hive.ql.metadata.Hive.loadDynamicPartitions(Hive.java:1961)
	... 102 more
Caused by: org.apache.hadoop.hive.ql.metadata.Table$ValidationFailureSemanticException: Partition spec {favorite_color=, favorite_numbers=2} contains non-partition columns
	at org.apache.hadoop.hive.ql.metadata.Table.validatePartColumnNames(Table.java:384)
	at org.apache.hadoop.hive.ql.metadata.Hive.getPartition(Hive.java:2232)
	at org.apache.hadoop.hive.ql.metadata.Hive.getPartition(Hive.java:2188)
	at org.apache.hadoop.hive.ql.metadata.Hive.loadPartition(Hive.java:1618)
	at org.apache.hadoop.hive.ql.metadata.Hive$3.call(Hive.java:1929)
	at org.apache.hadoop.hive.ql.metadata.Hive$3.call(Hive.java:1920)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
org.apache.spark.sql.AnalysisException: org.apache.hadoop.hive.ql.metadata.HiveException: Exception when loading 2 in table users4 with loadPath=hdfs://localhost:9000/user/hive/warehouse/users4/.hive-staging_hive_2024-05-21_18-42-55_548_3233995390238788603-1/-ext-10000
  at org.apache.spark.sql.hive.HiveExternalCatalog.withClient(HiveExternalCatalog.scala:110)
  at org.apache.spark.sql.hive.HiveExternalCatalog.loadDynamicPartitions(HiveExternalCatalog.scala:946)
  at org.apache.spark.sql.catalyst.catalog.ExternalCatalogWithListener.loadDynamicPartitions(ExternalCatalogWithListener.scala:189)
  at org.apache.kyuubi.spark.connector.hive.write.HiveBatchWrite.commitToMetastore(HiveBatchWrite.scala:188)
  at org.apache.kyuubi.spark.connector.hive.write.HiveBatchWrite.commit(HiveBatchWrite.scala:63)
  at org.apache.spark.sql.execution.datasources.v2.V2TableWriteExec.writeWithV2(WriteToDataSourceV2Exec.scala:422)
  at org.apache.spark.sql.execution.datasources.v2.V2TableWriteExec.writeWithV2$(WriteToDataSourceV2Exec.scala:382)
  at org.apache.spark.sql.execution.datasources.v2.CreateTableAsSelectExec.writeWithV2(WriteToDataSourceV2Exec.scala:68)
  at org.apache.spark.sql.execution.datasources.v2.TableWriteExecHelper.$anonfun$writeToTable$1(WriteToDataSourceV2Exec.scala:599)
  at org.apache.spark.util.Utils$.tryWithSafeFinallyAndFailureCallbacks(Utils.scala:1563)
  at org.apache.spark.sql.execution.datasources.v2.TableWriteExecHelper.writeToTable(WriteToDataSourceV2Exec.scala:587)
  at org.apache.spark.sql.execution.datasources.v2.TableWriteExecHelper.writeToTable$(WriteToDataSourceV2Exec.scala:582)
  at org.apache.spark.sql.execution.datasources.v2.CreateTableAsSelectExec.writeToTable(WriteToDataSourceV2Exec.scala:68)
  at org.apache.spark.sql.execution.datasources.v2.CreateTableAsSelectExec.run(WriteToDataSourceV2Exec.scala:93)
  at org.apache.spark.sql.execution.datasources.v2.V2CommandExec.result$lzycompute(V2CommandExec.scala:43)
  at org.apache.spark.sql.execution.datasources.v2.V2CommandExec.result(V2CommandExec.scala:43)
  at org.apache.spark.sql.execution.datasources.v2.V2CommandExec.executeCollect(V2CommandExec.scala:49)
  at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.$anonfun$applyOrElse$1(QueryExecution.scala:98)
  at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$6(SQLExecution.scala:118)
  at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:195)
  at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$1(SQLExecution.scala:103)
  at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:827)
  at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:65)
  at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.applyOrElse(QueryExecution.scala:98)
  at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.applyOrElse(QueryExecution.scala:94)
  at org.apache.spark.sql.catalyst.trees.TreeNode.$anonfun$transformDownWithPruning$1(TreeNode.scala:512)
  at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(TreeNode.scala:104)
  at org.apache.spark.sql.catalyst.trees.TreeNode.transformDownWithPruning(TreeNode.scala:512)
  at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.org$apache$spark$sql$catalyst$plans$logical$AnalysisHelper$$super$transformDownWithPruning(LogicalPlan.scala:31)
  at org.apache.spark.sql.catalyst.plans.logical.AnalysisHelper.transformDownWithPruning(AnalysisHelper.scala:267)
  at org.apache.spark.sql.catalyst.plans.logical.AnalysisHelper.transformDownWithPruning$(AnalysisHelper.scala:263)
  at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.transformDownWithPruning(LogicalPlan.scala:31)
  at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.transformDownWithPruning(LogicalPlan.scala:31)
  at org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:488)
  at org.apache.spark.sql.execution.QueryExecution.eagerlyExecuteCommands(QueryExecution.scala:94)
  at org.apache.spark.sql.execution.QueryExecution.commandExecuted$lzycompute(QueryExecution.scala:81)
  at org.apache.spark.sql.execution.QueryExecution.commandExecuted(QueryExecution.scala:79)
  at org.apache.spark.sql.execution.QueryExecution.assertCommandExecuted(QueryExecution.scala:133)
  at org.apache.spark.sql.DataFrameWriter.runCommand(DataFrameWriter.scala:856)
  at org.apache.spark.sql.DataFrameWriter.saveAsTable(DataFrameWriter.scala:633)
  at org.apache.spark.sql.DataFrameWriter.saveAsTable(DataFrameWriter.scala:563)
  ... 47 elided
```
```


```yaml
no
```

```yaml
no
```

### Additional context

the direct cause:
```
  private val allColumns = info.schema().toAttributes
  private val dataColumns = allColumns.take(allColumns.length - hiveTable.getPartCols.size())
  private val partColumns = allColumns.takeRight(hiveTable.getPartCols.size())
```

yikf:

Fixes [#6403](https://github.com/apache/kyuubi/issues/6403).

Writing a partitioned Hive table via the Kyuubi Spark Hive connector fails when the partition column is **not the last column** of the schema, e.g. `df.write.partitionBy("mid_col").saveAsTable(...)`.

There are two root causes:

1. `HiveWrite` split data/partition columns **positionally** with `take`/`takeRight`, assuming partition columns are always trailing. When a partition column is in the middle, columns are mis-classified and the metastore rejects the written partition spec:
   ```
   Table$ValidationFailureSemanticException: Partition spec {favorite_color=, favorite_numbers=2} contains non-partition columns
   ```
2. The Hive metastore reorders partition columns to the end of the schema, but the CTAS query output keeps the original column order. Spark's `ResolveOutputRelation` resolves output columns **by position**, so a non-last partition column is mis-cast onto a trailing data column (`INCOMPATIBLE_DATA_FOR_TABLE.CANNOT_SAFELY_CAST`).

This fixes both:
- `HiveWrite` now splits data/partition columns **by name** against `table.partitionColumnNames` (also for the write `dataSchema` derivation).
- `HiveTableCatalog.createTable` preserves the requested schema order on the returned table, so the subsequent CTAS write resolves output columns against the original order.

Affects 1.8.1 and current `master`.

Added a regression test in `HiveQuerySuite` that writes a table whose partition column is in the middle of the schema, using **mixed column types** (`String`/`String`/`Int`) so the schema-order bug is genuinely exercised (an all-`String` case would not surface the cast failure).

Verified locally with `-Pspark-3.5`:
- `HiveQuerySuite`: 21 tests passed (incl. the new case)
- `HiveCatalogSuite`: 24 tests passed
- `dev/reformat` (Spotless + Scalastyle) passed

### DIFF
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWrite.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWrite.scala
@@ -32,6 +32,7 @@ import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.catalog.CatalogTable
 import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
+import org.apache.spark.sql.catalyst.expressions.AttributeReference
 import org.apache.spark.sql.connector.distributions.{Distribution, Distributions}
 import org.apache.spark.sql.connector.expressions.{Expressions, SortDirection, SortOrder}
 import org.apache.spark.sql.connector.write.{BatchWrite, LogicalWriteInfo, RequiresDistributionAndOrdering}
@@ -66,8 +67,27 @@ case class HiveWrite(
   private val tableLocation = hiveTable.getDataLocation
 
   private val allColumns = info.schema().toAttributes
-  private val dataColumns = allColumns.take(allColumns.length - hiveTable.getPartCols.size())
-  private val partColumns = allColumns.takeRight(hiveTable.getPartCols.size())
+  // Split data and partition columns by name rather than by position. Users can create a
+  // partitioned table whose partition columns are not the trailing columns of the schema
+  // (e.g. df.write.partitionBy("mid_col")), so a positional take/takeRight would mis-classify
+  // columns and lead the metastore to reject the written partition spec. Match names with the
+  // session resolver because the Hive metastore lower-cases partition column names while the
+  // write schema preserves the user's original case. See KYUUBI #6403.
+  private val resolver = sparkSession.sessionState.conf.resolver
+  private def isPartitionColumn(name: String): Boolean =
+    table.partitionColumnNames.exists(resolver(_, name))
+  // partColumns keeps the query's original column case so requiredOrdering() resolves against the
+  // query.
+  private val (partColumns, dataColumns) =
+    allColumns.partition(col => isPartitionColumn(col.name))
+
+  // SPARK-28054: the Hive metastore keeps partition columns with lower-cased names, so the written
+  // partition directories must use lower-cased names; otherwise loadDynamicPartitions rejects the
+  // spec with "contains non-partition columns" when the schema preserves the user's original case.
+  // Rename via withName to keep the same exprId, so the value projection against allColumns is
+  // unaffected. See KYUUBI #6403.
+  private val outputPartColumns: Seq[AttributeReference] =
+    partColumns.map(col => col.withName(col.name.toLowerCase(Locale.ROOT)))
 
   lazy val tableDesc: TableDesc = new TableDesc(
     hiveTable.getInputFormatClass,
@@ -131,7 +151,8 @@ case class HiveWrite(
       customPartitionLocations: Map[TablePartitionSpec, String],
       options: Map[String, String]): WriteJobDescription = {
     val hiveFileFormat = getHiveFileFormat(fileSinkConf)
-    val dataSchema = StructType(info.schema().fields.take(dataColumns.length))
+    val dataSchema = StructType(
+      info.schema().filterNot(field => isPartitionColumn(field.name)))
     val outputWriterFactory = hiveFileFormat.prepareWrite(sparkSession, job, options, dataSchema)
     val metrics: Map[String, SQLMetric] = BasicWriteJobStatsTracker.metrics
     val serializableHadoopConf = new SerializableConfiguration(hadoopConf)
@@ -143,7 +164,7 @@ case class HiveWrite(
       outputWriterFactory = outputWriterFactory,
       allColumns = allColumns,
       dataColumns = dataColumns,
-      partitionColumns = partColumns,
+      partitionColumns = outputPartColumns,
       bucketSpec = None,
       path = pathName,
       customPartitionLocations = customPartitionLocations,
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTable.scala
@@ -46,7 +46,8 @@ import org.apache.kyuubi.spark.connector.hive.write.HiveWriteBuilder
 case class HiveTable(
     sparkSession: SparkSession,
     catalogTable: CatalogTable,
-    hiveTableCatalog: HiveTableCatalog)
+    hiveTableCatalog: HiveTableCatalog,
+    requestedSchema: Option[StructType] = None)
   extends Table with SupportsRead with SupportsWrite with Logging {
 
   lazy val dataSchema: StructType = catalogTable.dataSchema
@@ -86,7 +87,11 @@ case class HiveTable(
     }
   }
 
-  override def schema(): StructType = catalogTable.schema
+  // The metastore reorders partition columns to the end, but Spark's ResolveOutputRelation
+  // resolves output columns by position, so a CTAS with a non-last partition column mis-casts
+  // (CANNOT_SAFELY_CAST). Expose the requested order only via the V2 Table.schema() that feeds
+  // output resolution, keeping catalogTable (partition columns trailing) untouched. KYUUBI #6403.
+  override def schema(): StructType = requestedSchema.getOrElse(catalogTable.schema)
 
   override def properties(): util.Map[String, String] = catalogTable.properties.asJava
 
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/HiveTableCatalog.scala
@@ -398,7 +398,11 @@ class HiveTableCatalog(sparkSession: SparkSession)
         throw new TableAlreadyExistsException(ident)
     }
 
-    loadTable(ident)
+    // Preserve the requested schema order so the subsequent CTAS write resolves output columns
+    // against the original order. Override only the V2 Table.schema() rather than rebuilding the
+    // CatalogTable, which would break the trailing-partition invariant and drop fields
+    // newCatalogTable does not forward. See KYUUBI #6403.
+    loadTable(ident).asInstanceOf[HiveTable].copy(requestedSchema = Some(schema))
   }
 
   override def alterTable(ident: Identifier, changes: TableChange*): Table = {
--- a/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveBatchWrite.scala
+++ b/extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveBatchWrite.scala
@@ -17,6 +17,8 @@
 
 package org.apache.kyuubi.spark.connector.hive.write
 
+import java.util.Locale
+
 import scala.util.control.NonFatal
 
 import org.apache.hadoop.conf.Configuration
@@ -152,17 +154,23 @@ class HiveBatchWrite(
 
         val caseInsensitiveDpMap = CaseInsensitiveMap(dpMap)
 
+        // Lower-case the partition column names when reconstructing the path: the written
+        // directory is lower-cased while the restored table.partitionColumnNames may keep the
+        // original case, which would miss the directory on a case-sensitive FS. This alignment is
+        // not observable with bundled Hive 2.3.10 (moveFile(replace=true) already deletes the
+        // destination), so it is not covered by a test. See KYUUBI #6403.
         val updatedPartitionSpec = dynamicPartition.map {
-          case (key, Some(null)) => key -> ExternalCatalogUtils.DEFAULT_PARTITION_NAME
-          case (key, Some(value)) => key -> value
+          case (key, Some(null)) =>
+            key.toLowerCase(Locale.ROOT) -> ExternalCatalogUtils.DEFAULT_PARTITION_NAME
+          case (key, Some(value)) => key.toLowerCase(Locale.ROOT) -> value
           case (key, None) if caseInsensitiveDpMap.contains(key) =>
-            key -> caseInsensitiveDpMap(key)
+            key.toLowerCase(Locale.ROOT) -> caseInsensitiveDpMap(key)
           case (key, _) =>
             throw KyuubiHiveConnectorException(
               s"Dynamic partition key ${toSQLValue(key, StringType)} " +
                 "is not among written partition paths.")
         }
-        val partitionColumnNames = table.partitionColumnNames
+        val partitionColumnNames = table.partitionColumnNames.map(_.toLowerCase(Locale.ROOT))
         val tablePath = new Path(table.location)
         val partitionPath = ExternalCatalogUtils.generatePartitionPath(
           updatedPartitionSpec,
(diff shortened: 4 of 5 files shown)


extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWrite.scala
```diff
@@ -66,8 +67,14 @@ case class HiveWrite(
...
-  private val dataColumns = allColumns.take(allColumns.length - hiveTable.getPartCols.size())
-  private val partColumns = allColumns.takeRight(hiveTable.getPartCols.size())
+  // Split data and partition columns by name rather than by position. Users can create a
+  // partitioned table whose partition columns are not the trailing columns of the schema
+  // (e.g. df.write.partitionBy("mid_col")), so a positional take/takeRight would mis-classify
+  // columns and lead the metastore to reject the written partition spec. See KYUUBI #6403.
+  private val dataColumns: Seq[AttributeReference] =
+    allColumns.filterNot(col => table.partitionColumnNames.contains(col.name))
+  private val partColumns: Seq[AttributeReference] =
+    allColumns.filter(col => table.partitionColumnNames.contains(col.name))
```
maomaodev: Thanks for the fix, does this handle the case where the partition column name contains upper-case letters

yikf: Good catch -  it didn't. 

- Fixed by matching partition columns with the session resolver and lower-casing the written partition dir names, same as Spark's `V1WritesHiveUtils` (SPARK-28054).
- Added a `partitionBy("FavoriteColor")` regression test.

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWrite.scala
```diff
@@ -66,8 +67,28 @@ case class HiveWrite(
...
+  // session resolver because the Hive metastore lower-cases partition column names while the
+  // write schema preserves the user's original case. See KYUUBI #6403.
+  private val resolver = sparkSession.sessionState.conf.resolver
+  private def isPartitionColumn(name: String): Boolean =
+    table.partitionColumnNames.exists(resolver(_, name))
+  private val dataColumns: Seq[AttributeReference] =
+    allColumns.filterNot(col => isPartitionColumn(col.name))
+  // Keep the query's original column case here so requiredOrdering() resolves against the query.
+  private val partColumns: Seq[AttributeReference] =
+    allColumns.filter(col => isPartitionColumn(col.name))
```
maomaodev: Overall lgtm, nit: could we collapse the two passes into a single `allColumns.partition(col => isPartitionColumn(col.name))`?

yikf:
@pan3793 could you please take a look if you have a time, thans!

extensions/spark/kyuubi-spark-connector-hive/src/main/scala/org/apache/kyuubi/spark/connector/hive/write/HiveWrite.scala
```diff
@@ -66,8 +67,27 @@ case class HiveWrite(
...
+  // query.
+  private val (partColumns, dataColumns) =
+    allColumns.partition(col => isPartitionColumn(col.name))
+
+  // SPARK-28054: the Hive metastore keeps partition columns with lower-cased names, so the written
+  // partition directories must use lower-cased names; otherwise loadDynamicPartitions rejects the
+  // spec with "contains non-partition columns" when the schema preserves the user's original case.
+  // Rename via withName to keep the same exprId, so the value projection against allColumns is
+  // unaffected. See KYUUBI #6403.
+  private val outputPartColumns: Seq[AttributeReference] =
```
pan3793, line 89: [P1] Keep external-overwrite cleanup aligned with the lowercased write path

For an external table partitioned by `FavoriteColor`, this writes `favoritecolor=...`, but `HiveBatchWrite.commitToMetastore` constructs the stale partition path from the restored `table.partitionColumnNames` (`FavoriteColor`). The existence check then misses the old directory, bypassing its SPARK-29295 cleanup and causing `INSERT OVERWRITE` to retain old files. Normalize the names in the cleanup path as well (or delete the actual written paths), and cover a second overwrite.

pan3793: Correction on my P1: the mechanism holds, the impact does not.

On ext4, with the reconstruction reverted, the guard resolves `.../FavoriteColor=blue` (`exists=false`) while the data lives in `favoritecolor=blue`, so it never fires. But the stale file is still removed: Hive `moveFile(..., replace=true)` deletes the destination directory before moving (`Hive.java`, branch-2.3), including when the partition is missing from the metastore. So this is a dead guard, not data retention - downgrading from P1. The lower-casing is still the right alignment, keep it.

pan3793:
The external overwrite test does not cover the change it ships with: reverting only `HiveBatchWrite.scala` to master leaves `HiveQuerySuite` at 24/24 on both a case-insensitive FS (APFS) and a case-sensitive one (ext4). Reason in the inline thread. Suggest dropping the test and noting in the description that the alignment is not observable with bundled Hive 2.3.10.

The other three hunks are load-bearing, each verified by reverting it alone:

- by-name split -> both new tests fail
- lower-cased `outputPartColumns` -> uppercase test fails with `Partition spec {favoritecolor=, FavoriteColor=blue} contains non-partition columns`
- `createTable` column order -> both new tests fail with `CANNOT_SAFELY_CAST`

For the last one, consider scoping the order to the V2 table instead of rebuilding the `CatalogTable`. A reordered `CatalogTable` breaks the invariant that partition columns are trailing, which `partitionSchema` asserts and `HiveTable.dataSchema`/`partitionSchema` inherit, and `newCatalogTable` silently drops fields it does not forward, e.g. `collation` on Spark 4.2. Only `Table.schema()` feeds by-position output resolution:

```scala
case class HiveTable(
    sparkSession: SparkSession,
    catalogTable: CatalogTable,
    hiveTableCatalog: HiveTableCatalog,
    requestedSchema: Option[StructType] = None)

override def schema(): StructType = requestedSchema.getOrElse(catalogTable.schema)

// HiveTableCatalog.createTable
loadTable(ident).asInstanceOf[HiveTable].copy(requestedSchema = Some(schema))
```

Full proposal, 23 insertions / 80 deletions against your head, in https://github.com/pan3793/kyuubi/tree/branch-kyuubi-6403, CI at https://github.com/pan3793/kyuubi/actions/runs/31581643348. `HiveQuerySuite` + `HiveCatalogSuite` 46/46 with `-Pspark-3.5`, `dev/reformat` clean.

yikf:
Thanks @pan3793, you're right — Hive's `moveFile(..., replace=true)` already covers this, so the guard never fires with bundled Hive 2.3.10.

My intent was to make the alignment self-sufficient rather than rely on that fallback, and since it's harmless I kept it. But you're right the behavior isn't observable, so the test can't cover it — I've dropped it.

I also took your `createTable` suggestion: expose the requested order only through the V2 `Table.schema()` via `requestedSchema` on `HiveTable`, leaving `catalogTable` untouched.

Thanks for the detailed review!

pan3793 (approved):

pan3793:
thanks, merged to master/1.12

PR #7628, opened 2026-08-12
[KYUUBI #6995] Support Flink 2.0, 2.1, 2.2 and 2.3

### LINKED ISSUE #6995: [FEATURE] Support Flink 2.0
### Code of Conduct

### Search before asking

### Describe the feature

Flink has already release 2.0.0, which refers to [release-2.0.0](https://github.com/apache/flink/tree/release-2.0.0). Kyuubi could Flink 2.0 including integration with application mode.

### Motivation

Kyuubi could Flink 2.0 including integration with application mode.

### Describe the solution

Support Flink 2.0.

### Additional context

_No response_

pan3793:

Closes #6995.

Extend the Flink SQL engine to support Flink 2.0, 2.1, 2.2 and 2.3, while keeping the existing
1.17 to 1.20 support. The engine keeps shipping a single jar that runs against all supported
Flink versions, so the version differences are bridged by reflection rather than by per-version
source trees.

Flink 2.x changes that the engine has to adapt to:

- FLINK-14068 (2.0.0) removed `org.apache.flink.api.common.time.Time` in favor of `Duration`,
  which is a parameter of `EmbeddedJobClient`'s constructor.
- FLINK-33212 (2.0.0) added a `Configuration` parameter to `EmbeddedExecutor`'s constructor.
- FLINK-38974 (2.3.0) split `EmbeddedExecutorFactory`'s and `EmbeddedExecutor`'s job ids into the
  application, suspended and terminal ones, and made the dispatcher reject jobs that carry no
  registered application id. Flink resolves the factory constructor at compile time, so Kyuubi's
  copy of `EmbeddedExecutorFactory` declares both constructors and stamps the application id
  captured from the bootstrap job on every submitted `StreamGraph`, which enables application
  mode on Flink 2.3.
- FLINK-35625 and FLINK-36310 (2.0.0) merged `flink run-application` into `flink run` and removed
  the former, so `FlinkProcessBuilder` picks the action by the version detected from the
  `flink-dist` jar under `$FLINK_HOME/lib`.
- FLINK-36760 (2.0.0) removed `CliOptionsParser#checkFilePath`, which is now inlined into
  `FlinkEngineUtils`. Note that `DefaultContext#load` needs no adaption, its dependency list
  changed from `List<URL>` to `List<URI>` but both erase to the same descriptor, and Flink still
  reads the elements as `URL` in `SessionContext#create`.
- `Configuration#getString(ConfigOption)` was removed, `Configuration#get(ConfigOption)` is used
  instead, which exists in all supported versions.

Flink 2.x requires Java 11 or higher and is compiled with Java 11 target, so the `flink-2.x`
profiles raise `enforcer.maxJdkVersion` to 11, and the Flink 2.x CI jobs run on Java 17.

New UTs cover the Flink version detection and the application mode command selection in
`FlinkProcessBuilderSuite`.

`FlinkOperationSuite` asserts the result of `ENCODE` by version, since FLINK-38062 (2.2.0)
corrected its return type from `BINARY` to `VARBINARY`, before that the result was silently
truncated to the first byte.

`externals/kyuubi-flink-sql-engine` tests were run against each Flink binary distribution, with
`FLINK_HOME` pinned to the matching distribution:

| Flink | Result |
| --- | --- |
| 1.20.5 | 111 passed |
| 2.0.2 | 111 passed |
| 2.1.3 | 111 passed |
| 2.2.1 | 111 passed |
| 2.3.0 | 111 passed |

`FlinkOperationOnYarnSuite` covers application mode on Flink 2.3, and the cross-version job
compiles the engine against Flink 1.20 and runs it on the Flink 2.3 distribution.

The GA `flink-it` matrix now always compiles against the default Flink version and varies only
the runtime distribution, covering 1.17, 1.18, 1.19 on Java 8 and 2.0, 2.1, 2.2, 2.3 on Java 17.

Assisted-by: Claude Opus 5, Codex

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -334,6 +334,22 @@ jobs:
             flink: '1.20'
             flink-archive: '-Dflink.archive.mirror=https://www.apache.org/dyn/closer.lua/flink/flink-1.19.3 -Dflink.archive.name=flink-1.19.3-bin-scala_2.12.tgz'
             comment: 'verify-on-flink-1.19-binary'
+          - java: 17
+            flink: '1.20'
+            flink-archive: '-Dflink.archive.mirror=https://www.apache.org/dyn/closer.lua/flink/flink-2.0.2 -Dflink.archive.name=flink-2.0.2-bin-scala_2.12.tgz'
+            comment: 'verify-on-flink-2.0-binary'
+          - java: 17
+            flink: '1.20'
+            flink-archive: '-Dflink.archive.mirror=https://www.apache.org/dyn/closer.lua/flink/flink-2.1.3 -Dflink.archive.name=flink-2.1.3-bin-scala_2.12.tgz'
+            comment: 'verify-on-flink-2.1-binary'
+          - java: 17
+            flink: '1.20'
+            flink-archive: '-Dflink.archive.mirror=https://www.apache.org/dyn/closer.lua/flink/flink-2.2.1 -Dflink.archive.name=flink-2.2.1-bin-scala_2.12.tgz'
+            comment: 'verify-on-flink-2.2-binary'
+          - java: 17
+            flink: '1.20'
+            flink-archive: '-Dflink.archive.mirror=https://www.apache.org/dyn/closer.lua/flink/flink-2.3.0 -Dflink.archive.name=flink-2.3.0-bin-scala_2.12.tgz'
+            comment: 'verify-on-flink-2.3-binary'
     steps:
       - uses: actions/checkout@v7
       - name: Free up disk space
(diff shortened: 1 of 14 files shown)


pan3793:
After this, I'm going to drop support for Flink 1.17, 1.18, and 1.19, then mark Flink 2.0 as deprecated because it has been EOL

wForget (approved):
Thanks @pan3793 , LGTM

pan3793:
Pushed a new commit to resolve the application mode with Flink 2.3

pan3793:
thanks, merged to master

PR #7101, opened 2025-06-17
Load the existing pods when initializing kubernetes client to cleanup terminated app pods

turboFei:
### Why are the changes needed?

To prevent the terminated app pods leak if the events missed during kyuubi server restart. 

### How was this patch tested?

Manual test.

```
:2025-06-17 17:50:37.275 INFO [main] org.apache.kyuubi.engine.KubernetesApplicationOperation: [KubernetesInfo(Some(28),Some(dls-prod))] Found existing pod kyuubi-xb406fc5-7b0b-4fdf-8531-929ed2ae250d-8998-5b406fc5-7b0b-4fdf-8531-929ed2ae250d-8998-90c0b328-930f-11ed-a1eb-0242ac120002-0-20250423211008-grectg-stm-17da59fe-caf4-41e4-a12f-6c1ed9a293f9-driver with label: kyuubi-unique-tag=17da59fe-caf4-41e4-a12f-6c1ed9a293f9 in app state FINISHED, marking it as terminated
2025-06-17 17:50:37.278 INFO [main] org.apache.kyuubi.engine.KubernetesApplicationOperation: [KubernetesInfo(Some(28),Some(dls-prod))] Found existing pod kyuubi-xb406fc5-7b0b-4fdf-8531-929ed2ae250d-8998-5b406fc5-7b0b-4fdf-8531-929ed2ae250d-8998-90c0b328-930f-11ed-a1eb-0242ac120002-0-20250423212011-gpdtsi-stm-6a23000f-10be-4a42-ae62-4fa2da8fac07-driver with label: kyuubi-unique-tag=6a23000f-10be-4a42-ae62-4fa2da8fac07 in app state FINISHED, marking it as terminated
```
The pods are cleaned up eventually.
image

### Was this patch authored or co-authored using generative AI tooling?

No.

PR #7632, opened 2026-08-14
[K8s] Clean up terminated driver pods on informer add

ruanwenjun:

At startup, Kyuubi uses a separate one-time LIST to find and clean terminated driver pods. A pod can be missed if this LIST fails, or if it completes after the cleanup LIST observes it as running but is first reported by the informer as an ADD event. Since `onAdd` only records state and does not schedule cleanup, the pod may receive no further update and therefore remain in memory and not be deleted even when the cleanup strategy is `ALL`.

The informer already lists existing pods during initialization. Handling terminated pods in `onAdd` closes both gaps and makes the separate cleanup LIST unnecessary.

Added `mark terminated application received from pod add event` to verify that a succeeded Spark driver pod received through ADD is registered in the terminated-application cleanup trigger. The test fails without the production change and passes with it.

Yes, using Codex: GPT-5

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KubernetesApplicationOperationSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KubernetesApplicationOperationSuite.scala
@@ -17,13 +17,62 @@
 
 package org.apache.kyuubi.engine
 
-import io.fabric8.kubernetes.api.model.{ContainerState, ContainerStateWaiting}
+import com.google.common.cache.Cache
+import io.fabric8.kubernetes.api.model.{ContainerState, ContainerStateWaiting, Pod, PodBuilder}
+import io.fabric8.kubernetes.client.informers.ResourceEventHandler
 
 import org.apache.kyuubi.{KyuubiException, KyuubiFunSuite}
 import org.apache.kyuubi.config.KyuubiConf
-import org.apache.kyuubi.engine.ApplicationState.{FAILED, PENDING}
+import org.apache.kyuubi.engine.ApplicationState.{ApplicationState, FAILED, FINISHED, PENDING}
+import org.apache.kyuubi.engine.KubernetesApplicationOperation.LABEL_KYUUBI_UNIQUE_KEY
 
 class KubernetesApplicationOperationSuite extends KyuubiFunSuite {
+
+  private def podEventHandler(
+      operation: KubernetesApplicationOperation,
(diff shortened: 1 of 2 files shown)


ruanwenjun:
@pan3793 PTAL

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KubernetesApplicationOperationSuite.scala
```diff
@@ -17,14 +17,63 @@
...
+      .find(_.getSimpleName == "SparkEnginePodEventHandler")
+      .getOrElse(fail("SparkEnginePodEventHandler not found"))
+    val constructor = handlerClass.getDeclaredConstructors.head
+    constructor.setAccessible(true)
+    constructor
+      .newInstance(operation, kubernetesInfo)
+      .asInstanceOf[ResourceEventHandler[Pod]]
+  }
+
+  private def cleanupTrigger(
```
pan3793: The suite is in the same package, so reflection is avoidable: mark `SparkEnginePodEventHandler` and `cleanupTerminatedAppInfoTrigger` as `private[engine]` (with a `// Visible for testing` note, consistent with `checkKubernetesInfo`) and construct/access them directly. This also drops the name-mangling workaround `endsWith("cleanupTerminatedAppInfoTrigger")`.

pan3793 (approved):
Overall LGTM. The informer's initial LIST delivers existing pods as ADD events with built-in retry, so handling terminated pods in `onAdd` subsumes the one-shot cleanup LIST from KYUUBI #7101 (1.11.0) and closes the ADD-after-LIST race. Verified ADD/UPDATE map identically in `toApplicationStateAndError`, and `markApplicationTerminated` is idempotent, so no behavior loss.

kyuubi-server/src/test/scala/org/apache/kyuubi/engine/KubernetesApplicationOperationSuite.scala
ruanwenjun: done

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/1f04800014073998b296943da63bd71960a44652
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/1adb8c2ba6d3fb1696d23c7dc229c4b5127d9e65
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/aaa495ca8048b327f1b059f11f9b0da17c3b5e21

Issue #7635, opened 2026-08-14
[FEATURE] Support Zstd compression for Arrow IPC query results


byronwang123:
### Code of Conduct

* [x] I agree to follow this project's [[Code of Conduct](https://www.apache.org/foundation/policies/conduct)](https://www.apache.org/foundation/policies/conduct)

### Search before asking

* [x] I have searched in the [[issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue)](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the feature

Kyuubi supports Arrow-based query results to reduce serialization and deserialization overhead. However, the Arrow IPC batches are currently transferred without compression, so large result sets can still incur significant network transfer costs.

This follows up on #3877 (related issue #3865), which was deferred in 2022 while the community waited for an upstream Spark direction for Arrow compression.

Spark 4.1.0 has since introduced the upstream Arrow compression configurations:

* `spark.sql.execution.arrow.compression.codec`
* `spark.sql.execution.arrow.compression.zstd.level`

This provides an upstream-aligned approach that Kyuubi can reuse instead of introducing a Kyuubi-specific compression configuration or wire protocol.

### Motivation

In a cross-subnet cluster benchmark, enabling Zstd significantly reduced the Arrow IPC payload size and result fetch time:

* **8 numeric columns:** 1,167,279,640 B -> 247,522,728 B (**78.8% fewer bytes**); fetch time 19.3 s -> 7.5 s.
* **34 columns:** 4,674,336,136 B -> 849,920,016 B (**81.8% fewer bytes**); fetch time 205.8 s -> 94.1 s.

The exact timing depends on the network and workload, while the byte counts above are the Arrow IPC payload bytes returned by the server.

### Describe the solution

Reuse Spark upstream Arrow compression semantics:

- Reuse `spark.sql.execution.arrow.compression.codec` and
  `spark.sql.execution.arrow.compression.zstd.level`.
- Keep `none` as the default, preserving the existing behavior.
- Support Zstd compression initially.
- Use standard Arrow IPC body compression metadata instead of introducing
  Kyuubi-specific wire metadata.
- Let the JDBC client decode compressed batches through Arrow's
  `CompressionCodec.Factory`.
- Preserve compression when Arrow batches are sliced and re-serialized.
- Package the required Arrow compression implementation with the Spark engine
  for runtimes that do not provide it, so users do not need to configure
  additional runtime jars.

[...]

PR #7636, opened 2026-08-14
[KYUUBI #7635][SPARK] Support Zstd compression for Arrow IPC query results

byronwang123:

Kyuubi's Arrow result format reduces serialization/deserialization overhead, but Arrow IPC results are currently transferred without compression. For large result sets, network transfer can become a significant part of fetch latency.

This follows up on #3877 (related issue #3865). Spark 4.1 introduced the upstream Arrow compression configurations (`SPARK-54134`), so this PR follows Spark's configuration and Arrow IPC compression semantics instead of introducing a Kyuubi-specific protocol.

The main changes are:

- Support `none` (default) and `zstd` for Arrow IPC query results, including ZSTD compression-level configuration.
- Preserve compression when Arrow batches are sliced, e.g. when `LIMIT` crosses a batch boundary.
- Decode compressed batches in the JDBC client using standard Arrow IPC compression metadata.
- Keep `arrow-compression` optional on the Spark engine side. The default `none` path requires no additional dependency; when zstd is enabled, Kyuubi checks the required Arrow compression capability at runtime and reports a clear error if it is unavailable.
- Update `kyuubi-hive-jdbc-shaded` to include/relocate `arrow-compression`, while excluding `zstd-jni` and declaring it optional.
- Exclude `commons-compress`, since LZ4 is not supported.

For Spark runtimes that do not provide a compatible `arrow-compression` implementation, users enabling zstd need to add one to the Spark classpath. The default `none` behavior remains unchanged.

A cluster benchmark with Kyuubi 1.13.0-SNAPSHOT + Spark 3.5.9 showed 78.8%-81.8% fewer Arrow IPC bytes, with fetch time reduced from 19.3s to 7.5s in the 8-column case and from 205.8s to 94.1s in the 34-column case.

Added/updated coverage for:

- ZSTD compression/decompression and compression-level behavior.
- Compressed batch slicing across `LIMIT` boundaries.
- Full JDBC client <=> Kyuubi server <=> Spark engine ZSTD round trip.
- The default `none` path with no `arrow-compression` on the engine classpath.
- Runtime capability checking when zstd is enabled without the required dependency.
- Session-level `none -> zstd -> none` switching within the same engine.
- Unsupported LZ4 configuration.

Targeted engine tests and `KyuubiOperationPerUserSuite` pass on JDK 17 / Spark 3.5.8. The shaded artifacts were also checked to verify the intended `arrow-compression` / `zstd-jni` packaging.

Assisted-by: DeepSeek Harness:deepseek-v4-pro

### DIFF
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -257,6 +257,7 @@
                             <include>io.netty:*</include>
                             <include>io.perfmark:perfmark-api</include>
                             <include>io.vertx:*</include>
+                            <include>org.apache.arrow:arrow-compression</include>
                             <include>org.apache.kyuubi:*</include>
                             <include>org.checkerframework:checker-qual</include>
                             <include>org.codehaus.mojo:animal-sniffer-annotations</include>
@@ -269,6 +270,23 @@
                             <exclude>io.netty:netty-transport-*-kqueue</exclude>
                             <exclude>io.netty:netty-transport-*-io_uring</exclude>
                             <exclude>io.netty:netty-transport-native-epoll:*:linux-riscv64</exclude>
+                            <!--
+                              The Arrow vector/memory/format modules and the zstd/commons libraries
+                              used by arrow-compression are all provided by the Spark runtime on
+                              every supported Spark version, so they are excluded from the engine
+                              jar to avoid classpath conflicts and native library relocation issues.
+                            -->
+                            <exclude>org.apache.arrow:arrow-format</exclude>
+                            <exclude>org.apache.arrow:arrow-memory-core</exclude>
+                            <exclude>org.apache.arrow:arrow-memory-netty</exclude>
+                            <exclude>org.apache.arrow:arrow-memory-netty-buffer-patch</exclude>
+                            <exclude>org.apache.arrow:arrow-vector</exclude>
+                            <exclude>com.github.luben:zstd-jni</exclude>
+                            <exclude>commons-codec:commons-codec</exclude>
+                            <exclude>commons-io:commons-io</exclude>
+                            <exclude>org.apache.commons:commons-compress</exclude>
+                            <exclude>org.apache.commons:commons-lang3</exclude>
+                            <exclude>org.immutables:value</exclude>
                         </excludes>
                     </artifactSet>
                     <filters>
@@ -289,6 +307,7 @@
                                 <exclude>NOTICE.txt</exclude>
                                 <exclude>mozilla/**</exclude>
                                 <exclude>**/module-info.class</exclude>
+                                <exclude>arrow-git.properties</exclude>
                             </excludes>
                         </filter>
                     </filters>
@@ -410,6 +429,17 @@
                         <relocation>
                             <pattern>net.bytebuddy</pattern>
                             <shadedPattern>${kyuubi.shade.packageName}.net.bytebuddy</shadedPattern>
+                        </relocation>
+                        <!--
+                          Relocate the bundled arrow-compression classes so that Kyuubi's copy
+                          never conflicts with the arrow-compression jar shipped by Spark 4.1+.
+                          The relocated codec classes still resolve the unrelocated
+                          org.apache.arrow.vector.* and com.github.luben.zstd.* types, which are
+                          provided by the Spark runtime on all supported Spark versions.
+                        -->
+                        <relocation>
+                            <pattern>org.apache.arrow.compression</pattern>
+                            <shadedPattern>${kyuubi.shade.packageName}.org.apache.arrow.compression</shadedPattern>
                         </relocation>
                     </relocations>
                     <transformers>
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -37,10 +37,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <!-- Compile-time only: not bundled into the engine jar; provided by Spark 4.1+ or the user. -->
         <dependency>
             <groupId>org.apache.arrow</groupId>
             <artifactId>arrow-compression</artifactId>
             <version>${arrow.version}</version>
+            <optional>true</optional>
         </dependency>
 
         <dependency>
@@ -257,7 +259,6 @@
                             <include>io.netty:*</include>
                             <include>io.perfmark:perfmark-api</include>
                             <include>io.vertx:*</include>
-                            <include>org.apache.arrow:arrow-compression</include>
                             <include>org.apache.kyuubi:*</include>
                             <include>org.checkerframework:checker-qual</include>
                             <include>org.codehaus.mojo:animal-sniffer-annotations</include>
@@ -427,17 +428,6 @@
                         <relocation>
                             <pattern>net.bytebuddy</pattern>
                             <shadedPattern>${kyuubi.shade.packageName}.net.bytebuddy</shadedPattern>
-                        </relocation>
-                        <!--
-                          Relocate the bundled arrow-compression classes so that Kyuubi's copy
-                          never conflicts with the arrow-compression jar shipped by Spark 4.1+.
-                          The relocated codec classes still resolve the unrelocated
-                          org.apache.arrow.vector.* and com.github.luben.zstd.* types, which are
-                          provided by the Spark runtime on all supported Spark versions.
-                        -->
-                        <relocation>
-                            <pattern>org.apache.arrow.compression</pattern>
-                            <shadedPattern>${kyuubi.shade.packageName}.org.apache.arrow.compression</shadedPattern>
                         </relocation>
                     </relocations>
                     <transformers>
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -271,10 +271,10 @@
                             <exclude>io.netty:netty-transport-*-io_uring</exclude>
                             <exclude>io.netty:netty-transport-native-epoll:*:linux-riscv64</exclude>
                             <!--
-                              The Arrow vector/memory/format modules and the zstd/commons libraries
-                              used by arrow-compression are all provided by the Spark runtime on
-                              every supported Spark version, so they are excluded from the engine
-                              jar to avoid classpath conflicts and native library relocation issues.
+                              The Arrow vector/memory/format modules and the zstd library used by
+                              arrow-compression are all provided by the Spark runtime on every
+                              supported Spark version, so they are excluded from the engine jar
+                              to avoid classpath conflicts and native library relocation issues.
                             -->
                             <exclude>org.apache.arrow:arrow-format</exclude>
                             <exclude>org.apache.arrow:arrow-memory-core</exclude>
@@ -284,7 +284,6 @@
                             <exclude>com.github.luben:zstd-jni</exclude>
                             <exclude>commons-codec:commons-codec</exclude>
                             <exclude>commons-io:commons-io</exclude>
-                            <exclude>org.apache.commons:commons-compress</exclude>
                             <exclude>org.apache.commons:commons-lang3</exclude>
                             <exclude>org.immutables:value</exclude>
                         </excludes>
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -307,7 +307,6 @@
                                 <exclude>NOTICE.txt</exclude>
                                 <exclude>mozilla/**</exclude>
                                 <exclude>**/module-info.class</exclude>
-                                <exclude>arrow-git.properties</exclude>
                             </excludes>
                         </filter>
                     </filters>
--- a/externals/kyuubi-spark-sql-engine/pom.xml
+++ b/externals/kyuubi-spark-sql-engine/pom.xml
@@ -38,6 +38,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.arrow</groupId>
+            <artifactId>arrow-compression</artifactId>
+            <version>${arrow.version}</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.kyuubi</groupId>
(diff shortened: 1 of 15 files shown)


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -77,6 +79,28 @@ object SparkDatasetHelper extends Logging {
     toArrowBatchRddMethod.bind(ds).invoke()
   }
 
+  /**
+   * Read the arrow compression codec config from the session conf, reusing the Spark upstream
+   * configuration keys (`spark.sql.execution.arrow.compression.codec` and
+   * `spark.sql.execution.arrow.compression.zstd.level`, introduced in Spark 4.1). On Spark < 4.1
```
pan3793: Add the JIRA ticket when referring to upstream Spark changes, e.g., `SPARK-XXXXX (4.1.0) do sth`

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -77,6 +79,28 @@ object SparkDatasetHelper extends Logging {
...
+   * these keys are not registered as typed SQLConf entries, so they are read as raw strings: a
+   * session-level SET lands in the SQLConf settings map and takes precedence, otherwise the
+   * engine-level kyuubi-defaults.conf value applies, otherwise the default.
```
pan3793: that's too verbose, you can just say: Read the session-level arrow compression codec configs ...

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -77,6 +79,28 @@ object SparkDatasetHelper extends Logging {
...
+   * Returns (codecName, zstdLevel).
+   */
+  private def arrowCompressionConf(spark: SparkSession): (String, Int) = {
+    // Lower-case the codec name to match Spark upstream, whose typed entry applies
+    // `.transform(_.toLowerCase(Locale.ROOT))`; reading the key as a raw string here would
+    // otherwise make `ZSTD`/`Zstd` reject on Spark < 4.1 while accepted natively on 4.1+.
```
pan3793:
```suggestion
    // Lowercase the codec name to match Spark upstream.
```

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkArrowbasedOperationSuite.scala
```diff
@@ -380,6 +380,46 @@ class SparkArrowbasedOperationSuite extends WithSparkSQLEngine with SparkDataTyp
     }
   }
 
+  test("arrow zstd compression round-trips through the JDBC client") {
+    // End-to-end coverage of the compressed Arrow IPC path: the engine compresses batches with
+    // the upstream-named codec config, and the JDBC client transparently decompresses them from
+    // the batch body compression metadata (no server-side codec hint on the wire).
```
pan3793, line 386: the "end-to-end coverage" is inaccurate, it's just the engine side coverage, you still need to add the server side coverage to verify the: jdbc client <==> kyuubi server <==> spark engine

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/execution/arrow/KyuubiArrowConverters.scala
```diff
@@ -39,6 +41,28 @@ object KyuubiArrowConverters extends SQLConfHelper with Logging {
...
+   * user-configured level. This follows the latest Spark upstream behavior
+   * (ArrowCompressionUtils).
+   */
+  private def createCodec(codecName: String, zstdLevel: Int): CompressionCodec = {
+    if (codecName == null || codecName == "none") {
+      NoCompressionCodec.INSTANCE
+    } else {
+      codecName match {
+        case "zstd" => new ZstdCompressionCodec(zstdLevel)
+        case other =>
```
pan3793: Why not merge the `null` and `"none"` branches into the pattern-match body? For `lz4`, we need to either fail with a clear error message, or fall back to null with a warning message

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
```diff
@@ -89,20 +113,26 @@ object SparkDatasetHelper extends Logging {
...
     }
   }
 
   def toArrowBatchLocalIterator(df: DataFrame): Iterator[Array[Byte]] = {
     withNewExecutionId(df) {
-      toArrowBatchRdd(df).toLocalIterator
+      // use the plan-based toArrowBatchRdd so that the arrow compression codec takes effect;
+      // the Dataset#toArrowBatchRdd (reflective) path uses the vanilla Spark ArrowConverters
+      // which does not apply the Kyuubi compression codec.
+      toArrowBatchRdd(df.queryExecution.executedPlan).toLocalIterator
```
pan3793, line 113: do we still need `def toArrowBatchRdd[T](ds: Dataset[T])` then?

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -283,6 +307,7 @@
                                 <exclude>NOTICE.txt</exclude>
                                 <exclude>mozilla/**</exclude>
                                 <exclude>**/module-info.class</exclude>
+                                <exclude>arrow-git.properties</exclude>
```
pan3793: will this cause any side-effect if it is unavailable at runtime?

externals/kyuubi-spark-sql-engine/pom.xml
```diff
@@ -405,6 +430,17 @@
                             <pattern>net.bytebuddy</pattern>
                             <shadedPattern>${kyuubi.shade.packageName}.net.bytebuddy</shadedPattern>
                         </relocation>
+                        <!--
+                          Relocate the bundled arrow-compression classes so that Kyuubi's copy
+                          never conflicts with the arrow-compression jar shipped by Spark 4.1+.
+                          The relocated codec classes still resolve the unrelocated
+                          org.apache.arrow.vector.* and com.github.luben.zstd.* types, which are
```
pan3793: I'm worried about the half-shading decision. Some questions:

- if we don't enable this feature, and don't bundle this depenedency, does it work fine or throw a runtime class not found error?
- if the answer to the above question is "work fine", we can leave it and ask the user to put the `arrow-compression` dependency in their Spark classpath manually

pom.xml
```diff
@@ -348,6 +348,11 @@
                 <artifactId>arrow-vector</artifactId>
                 <version>${arrow.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.arrow</groupId>
+                <artifactId>arrow-compression</artifactId>
+                <version>${arrow.version}</version>
```
pan3793, line 355: actually, we can exclude `commons-compress` since we don't support lz4.

note: things have changed since I suggested `none`, `gzip`, `lz4`, and `zstd` options, lz4-java was abandoned by the original author, and another guy took over but focused only on security patches, so I would suggest `none` and `zstd` only now. (arrow compression does not support `gzip` so exclude it too)

kyuubi-hive-jdbc/pom.xml
```diff
@@ -46,6 +46,11 @@
             <artifactId>arrow-vector</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.arrow</groupId>
+            <artifactId>arrow-compression</artifactId>
+        </dependency>
```
pan3793, line 52: we should tune the `kyuubi-hive-jdbc-shaded`'s `pom.xml` too, include and relocate `arrow-compression`, but exclude `zstd-jni` and declare it as an optional dependency

pan3793:
the decision to choose "Arrow IPC compression" looks good to me, Arrow Java has implemented it since 8.0.0, and Spark 3.5 ships Arrow 12.0.1, so that covers all supported Spark versions.

cc @cfmcgrady, could you take a look since you are the author of this feature?

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
byronwang123: Addressed.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
byronwang123: Addressed

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
byronwang123: Addressed.

externals/kyuubi-spark-sql-engine/src/test/scala/org/apache/kyuubi/engine/spark/operation/SparkArrowbasedOperationSuite.scala
byronwang123: Addressed. I renamed the existing test to engine-side coverage and added a server-side test covering JDBC client <=> Kyuubi server <=> Spark engine.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/execution/arrow/KyuubiArrowConverters.scala
byronwang123: Addressed. Merged the branches and made unsupported codecs fail with a clear error.

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/kyuubi/SparkDatasetHelper.scala
byronwang123: Addressed. Removed the unused Dataset overload.

externals/kyuubi-spark-sql-engine/pom.xml
byronwang123: I checked this further. With the current implementation, the none path still references CommonsCompressionFactory in the slice path, so simply removing the bundled arrow-compression would cause a runtime class-not-found issue on Spark 3.5/4.0.

One alternative is to keep the original one-argument VectorLoader/VectorUnloader for none, and only use the compression-aware path for zstd. That would make arrow-compression truly optional on the engine side and avoid the half-shading.

Do you prefer this approach over keeping the relocated bundled dependency?

pan3793: How about checking the existence of `CommonsCompressionFactory` at runtime? You can refer to the implementation of Spark's `enableHiveSupport`.

pan3793: > One alternative is to keep the original one-argument VectorLoader/VectorUnloader for none, and only use the compression-aware path for zstd.

~it's does not work if it can be changed at the session level~

withdraw the assertion, I wrongly assumed it was a static var

externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/spark/sql/execution/arrow/KyuubiArrowCompressionSupport.scala
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.arrow
+
+import org.apache.arrow.compression.{CommonsCompressionFactory, ZstdCompressionCodec}
+import org.apache.arrow.vector.{VectorLoader, VectorSchemaRoot, VectorUnloader}
+
+/** Isolates the optional arrow-compression dependency so the uncompressed path never loads it. */
+private[sql] object ArrowCompressionSupport {
```
pan3793: Add a `Kyuubi` prefix to the class name to avoid potential class collision

pan3793:
code change looks almost good to me, and I need to check the packaging result and verify it on a real setup before approval (hope I can find time to finish it in one or two days)

pan3793:
And seems you deleted this part from the PR description? But this is a legally mandated item.

> ### Was this patch authored or co-authored using generative AI tooling?
>
> If generative AI tooling has assisted in authoring or reviewing this patch, please include
> an 'Assisted-by: ' trailer that identifies the agent and model.
> For example: 'Assisted-by: Claude:claude-opus-4-7' or 'Assisted-by: Claude Opus 4.7'.
> If no, write 'No'.
> Please refer to the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) for details.

byronwang123:
> code change looks almost good to me, and I need to check the packaging result and verify it on a real setup before approval (hope I can find time to finish it in one or two days)

Thanks for taking another look. I've addressed the remaining naming comment and updated the PR description to reflect the final runtime/packaging behavior, including the generative AI disclosure. Appreciate you checking the packaging and real setup.

pan3793:
CI failure is related

pan3793 (approved):
LGTM, I verified this feature locally.

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/cfbbeec429db0115ce42fbaf91f13c6eb9092762

PR #7637, opened 2026-08-18
[KYUUBI #7623][AUTHZ] Check the effective value of spark.sql.optimizer.excludedRules

### LINKED ISSUE #7623: [FEATURE][AUTHZ] AuthzConfigurationChecker should validate the effective value of spark.sql.optimizer.excludedRules, not the SET syntax
### Code of Conduct

### Search before asking

### Describe the feature

`AuthzConfigurationChecker` matches the shape of the plan — `SetCommand` and `ResetCommand` — and throws when a `SET` carries `spark.sql.optimizer.excludedRules` with a value from `org.apache.kyuubi.plugin.spark.authz.ranger.*` (`AuthzConfigurationChecker.scala:42-53`). Every path that writes the same key without producing a plan is unaffected, because `RuleAuthorization` is an injected optimizer rule (`RangerSparkExtension.scala:65`) and extension rules are not part of `SparkOptimizer.nonExcludableRules`. I would like the check rule to read the effective value on every plan instead, so that enforcement no longer depends on which API set the config.

### Motivation

Kyuubi 1.13.0-SNAPSHOT built for Spark 4.0.3, engine in `local[*]` with `RangerSparkExtension` and no policies, share level USER, stock server configuration. Two JDBC connections of the same user, so the second one reuses the engine started by the first:

```
A1 create table, plain session    -> Permission denied: user [probeuser] does not have [create] privilege on [default/jdbc_a51791]
A2 SET excludedRules via SQL      -> Excluding Authz security rules is not allowed
B1 create table, conf in JDBC url -> OK
```

B differs from A only by `;#spark.sql.optimizer.excludedRules=org.apache.kyuubi.plugin.spark.authz.ranger.RuleAuthorization` in the JDBC URL. Session confs are applied through `SparkSessionImpl.open()` → `spark.conf.set` (`SparkSessionImpl.scala:55-61`, `:101-103`), which never builds a plan.

This is not an undocumented hole. `docs/security/authorization/spark/overview.rst:85-93` prescribes `kyuubi.session.conf.restrict.list = spark.driver.memory,spark.sql.optimizer.excludedRules`, and with that set the connection is rejected at open — I verified that as well. The gap is that enforcement lives in two places that have to agree, the safe one is off by default, and it only covers the channels that pass through the server's `SessionManager`.

### Describe the solution

Have `AuthzConfigurationChecker` read `spark.sql.optimizer.excludedRules` from the session config on every plan and throw if it names an authz rule, in addition to the existing `SetCommand` and `ResetCommand` cases. Check rules are not subject to `excludedRules` — that config filters optimizer batches only — so the check itself cannot be removed by the same trick. The restrict list then becomes defence in depth rather than the only line, and any client writing the key through `spark.conf.set` is covered too.

### Additional context

Related to #7454, and to the fail-open theme in #7593.

alexandrefimov:

Closes #7623.

`AuthzConfigurationChecker` guards the exclusion by matching a `SetCommand` in the logical plan (`AuthzConfigurationChecker.scala:42-45`), so the protection covers the `SET` syntax only - which is also how the docs describe it (`docs/security/authorization/spark/overview.rst:106`, "A set statement with key equal to ..."). Every channel that writes the config without producing a plan keeps working: `spark.conf.set`, the Spark Connect Config RPC (`SparkConnectConfigHandler.handleSet` calls `conf.set` directly), or the key passed in a JDBC connection string. Once `RuleAuthorization` is named there, `Optimizer.batches` drops it - extension rules are not in `SparkOptimizer.nonExcludableRules` - and the rest of the session runs unauthorized.

Reproduced on Spark 4.0.3 with `kyuubi-spark-authz` and a Ranger plugin that denies by default: `create` denied, `SET spark.sql.optimizer.excludedRules=...RuleAuthorization` rejected by the checker, the same key accepted over the Connect Config RPC, the next `create` allowed.

The documented mitigation, `kyuubi.session.conf.restrict.list` (`docs/security/authorization/spark/overview.rst:85-93`), does reject such a JDBC connection - I checked that too - but it is enforced in the server's `SessionManager`, so it does not reach a client that talks to the engine directly.

This patch reads the value in effect on every plan instead of matching the statement. Check rules are not filtered by `excludedRules`, which only applies to optimizer batches, so this check cannot be removed the same way.

Two points a reviewer may want to decide differently:

- The value check matches `org.apache.kyuubi.plugin.spark.authz.ranger`, the prefix the existing `SET` case uses. The plugin also injects optimizer rules from `org.apache.kyuubi.plugin.spark.authz.rule` (`RuleEliminateMarker` and its neighbours), which neither the old nor the new check covers. I left the prefix as is rather than widen the scope here.
- The check is fail closed for the whole session: with the exclusion already in the session conf, every plan is rejected, not just the `SET`. That is the intent, but it is a visible behaviour change for a session that set the key before this patch.

New test in `AuthzConfigurationCheckerSuite`: the config is written through `spark.conf.set` - the same write path the Connect Config RPC takes - and the next plan is rejected, while excluding a non-authz rule (`ConstantFolding`) stays allowed.

```
build/mvn test -pl extensions/spark/kyuubi-spark-authz -Dtest=none \
    -DwildcardSuites=org.apache.kyuubi.plugin.spark.authz.rule.AuthzConfigurationCheckerSuite
```

All three tests in the suite pass; with the new check removed from `apply`, exactly the new test fails.

The existing test needed one line: `sql("set spark.sql.optimizer.excludedRules=...")` applies the value to the shared session before the rule is invoked by hand, so the test now unsets it - otherwise the effective-value check rejects every later plan in that session.

Assisted-by: Claude:claude-opus-5

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/config/AuthzConfigurationChecker.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/config/AuthzConfigurationChecker.scala
@@ -30,6 +30,9 @@ import org.apache.kyuubi.plugin.spark.authz.util.AuthZUtils.SKIP_CATALOGLESS_V2_
 case class AuthzConfigurationChecker(spark: SparkSession) extends (LogicalPlan => Unit) {
 
   final val RESTRICT_LIST_KEY = "spark.kyuubi.conf.restricted.list"
+  final val EXCLUDED_RULES_KEY = "spark.sql.optimizer.excludedRules"
+
+  final private val AUTHZ_RANGER_RULE_PACKAGE = "org.apache.kyuubi.plugin.spark.authz.ranger"
 
   private val restrictedConfList: Set[String] =
     Set(
@@ -39,17 +42,27 @@ case class AuthzConfigurationChecker(spark: SparkSession) extends (LogicalPlan =
       SKIP_CATALOGLESS_V2_RELATION_ENABLED_KEY) ++
       spark.conf.getOption(RESTRICT_LIST_KEY).map(_.split(',').toSet).getOrElse(Set.empty)
 
-  override def apply(plan: LogicalPlan): Unit = plan match {
-    case SetCommand(Some((
-          "spark.sql.optimizer.excludedRules",
-          Some(v)))) if v.contains("org.apache.kyuubi.plugin.spark.authz.ranger") =>
+  override def apply(plan: LogicalPlan): Unit = {
+    // SET is not the only way the exclusion can be set: spark.conf.set and the Spark
+    // Connect Config RPC write the value with no logical plan at all, so the SetCommand
+    // case below never sees them. Check the value that is actually in effect on every
+    // plan instead - check rules are not affected by spark.sql.optimizer.excludedRules,
+    // which only filters optimizer batches, so this check cannot be turned off the same way.
+    if (spark.conf.getOption(EXCLUDED_RULES_KEY).exists(_.contains(AUTHZ_RANGER_RULE_PACKAGE))) {
       throw new AccessControlException("Excluding Authz security rules is not allowed")
-    case SetCommand(Some((k, Some(_)))) if restrictedConfList.contains(k) =>
-      throw new AccessControlException(s"Modifying config $k is not allowed")
-    case ResetCommand(Some(k)) if restrictedConfList.contains(k) =>
-      throw new AccessControlException(s"Resetting config $k is not allowed")
-    case ResetCommand(None) =>
-      throw new AccessControlException("Resetting all configs is not allowed")
-    case _ =>
+    }
+    plan match {
+      case SetCommand(Some((
+            EXCLUDED_RULES_KEY,
+            Some(v)))) if v.contains(AUTHZ_RANGER_RULE_PACKAGE) =>
+        throw new AccessControlException("Excluding Authz security rules is not allowed")
+      case SetCommand(Some((k, Some(_)))) if restrictedConfList.contains(k) =>
+        throw new AccessControlException(s"Modifying config $k is not allowed")
+      case ResetCommand(Some(k)) if restrictedConfList.contains(k) =>
+        throw new AccessControlException(s"Resetting config $k is not allowed")
+      case ResetCommand(None) =>
+        throw new AccessControlException("Resetting all configs is not allowed")
+      case _ =>
+    }
   }
 }
--- a/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/rule/AuthzConfigurationCheckerSuite.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/rule/AuthzConfigurationCheckerSuite.scala
@@ -50,9 +50,38 @@ class AuthzConfigurationCheckerSuite extends KyuubiFunSuite with SparkSessionPro
     val p8 = sql(s"set spark.sql.optimizer.excludedRules=${classOf[RuleAuthorization].getName}")
       .queryExecution.analyzed
     intercept[AccessControlException](extension.apply(p8))
+    // sql() has already applied the SET before the rule is invoked here, and the value it
+    // leaves behind is what the effective-value check rejects on every plan of this session
+    spark.conf.unset(extension.EXCLUDED_RULES_KEY)
     val p9 = sql(
       "set spark.kyuubi.authz.skipCataloglessV2Relation.enabled=true").queryExecution.analyzed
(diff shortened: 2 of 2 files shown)


alexandrefimov:
The one red check here, `Flink Test (8, 1.20, normal)`, looks unrelated to this patch. The failure is `execute statement - select column name with dots` in `FlinkOperationSuiteOnYarn`, and the throw comes from Flink's own job submission — `ArrayIndexOutOfBoundsException` in `EmbeddedExecutor.submitAndGetJobClientFuture`, below the engine's `ExecuteStatement`. This PR only touches `extensions/spark/kyuubi-spark-authz`, which that job does not exercise.

It does not look like a master-side break either: the same job passed on #7636 against the same master. This branch is level with master, so there is nothing to rebase onto that would re-trigger the run — could someone re-run that job?

wForget:
@alexandrefimov Thanks. The failure of `Flink Test (8, 1.20, normal)` is unrelated to the current PR; I filed it in #7644.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/config/AuthzConfigurationChecker.scala
```diff
@@ -30,6 +30,9 @@ import org.apache.kyuubi.plugin.spark.authz.util.AuthZUtils.SKIP_CATALOGLESS_V2_
 case class AuthzConfigurationChecker(spark: SparkSession) extends (LogicalPlan => Unit) {
 
   final val RESTRICT_LIST_KEY = "spark.kyuubi.conf.restricted.list"
+  final val EXCLUDED_RULES_KEY = "spark.sql.optimizer.excludedRules"
+
+  final private val AUTHZ_RANGER_RULE_PACKAGE = "org.apache.kyuubi.plugin.spark.authz.ranger"
```
wForget, line 35: `org.apache.kyuubi.plugin.spark.authz.ranger` should be changed to `org.apache.kyuubi.plugin.spark.authz`. Currently, most rules are located at https://github.com/apache/kyuubi/tree/master/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule cc @cfmcgrady @pan3793

wForget (approved):
Thanks @alexandrefimov ，LGTM.

I added a comment that may not be relevant to this PR; we can fix it in a new PR.

extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/config/AuthzConfigurationChecker.scala
alexandrefimov: Happy to do that in a follow-up.

One thing I checked while looking at it: `excludedRules` only filters optimizer batches, and Spark has no analyzer equivalent, so the rules that apply masking and row filters can't be excluded at all. Of the rules the extension injects, four are reachable this way — `RuleAuthorization` in `…authz.ranger`, and `RuleEliminateMarker`, `RuleEliminatePermanentViewMarker`, `RuleEliminateTypeOf` in `…authz.rule`. Reading those three, excluding one looks like it breaks the query rather than lifting a check: the markers survive into planning and `TypeOfPlaceHolder` has no `eval`. So widening the prefix looks like future-proofing to me — new rules land under `…authz.rule` — rather than closing a live bypass. That's code reading, not a run.

It does flip behaviour for anyone excluding one of those cleanup rules today, so waiting for @cfmcgrady / @pan3793 seems right. Should I open an issue and a PR for it once they weigh in?

alexandrefimov: Ran it since, and the reading holds. With the real extension and the test policies, excluding `RuleEliminateMarker` fails planning for both a masked and a row-filtered query — `No plan for RowFilterMarker +- DataMaskingStage0Marker …`; `RuleEliminatePermanentViewMarker` fails the same way (`No plan for PermanentViewMarker …`), and `RuleEliminateTypeOf` fails at runtime with `UnaryExpressions must override either eval or nullSafeEval`. The masking projection is still in the failing plan (`md5(cast(cast(value1 as string) as binary))`), which is the point: masking and row filtering come from resolution rules that can't be excluded at all, and those three only strip the markers afterwards. A user with no privilege on the table is denied exactly as before with all three excluded.

So nothing is reachable through the narrower prefix today, and the wider one covers rules added under `…authz.rule` later. Glad to carry these as tests in the follow-up if that's the direction you land on.

alexandrefimov:
@wForget this is approved and green. The prefix widening is a separate change — it rejects configs that are accepted today — so I kept it out of this PR and will open the follow-up once it has an issue. Anything left before this can go in?

wForget:
Thanks, merged to master

PR #7638, opened 2026-08-19
Minor improvement of PR template

pan3793:

Small words tune, please review the inline change.

Review.

No.


wForget (approved):
Thanks @pan3793

PR #3230, opened 2022-08-14
Flink SQL engine supports run across versions

pan3793:
### _Why are the changes needed?_

To make sure flink sql engine build against flink-1.15 can run on flink-1.14.

### _How was this patch tested?_
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] [Run test](https://kyuubi.apache.org/docs/latest/develop_tools/testing.html#running-tests) locally before make a pull request

PR #7641, opened 2026-08-21
[KYUUBI #7640] Fix exception double-unwrap in DynMethods.UnboundMethod.invoke

### LINKED ISSUE #7640: DynMethods.UnboundMethod.invoke loses or replaces exceptions thrown by invoked methods
### Description

`DynMethods.UnboundMethod.invoke` unwraps exceptions twice. `invokeChecked` already unwraps the `InvocationTargetException` and throws the real exception produced by the invoked method, but `invoke` then catches that exception and re-throws `e.getCause()` as if `e` were still the wrapper:

```java
public <R> R invoke(Object target, Object... args) {
  try {
    return this.invokeChecked(target, args);
  } catch (Exception e) {
    if (e.getCause() instanceof RuntimeException) {
      throw (RuntimeException) e.getCause();
    }
    throw new RuntimeException(e.getCause());
  }
}
```

Two consequences for callers:

- An exception without a cause degrades to `new RuntimeException(null)` — the type, message and stack trace of the real failure are all lost. For example, a target method throwing `new IOException("connection refused")` surfaces as a blank `RuntimeException` with a null message.
- A `RuntimeException` carrying a cause is silently replaced by its cause: `IllegalStateException("...", inner)` surfaces as `inner`, so `catch (IllegalStateException)` at the call site misses it.

Both shapes are reachable wherever `invoke` is used across the codebase (e.g. `KyuubiBeeLine` error reporting, `EngineTab` install-error reporting, `ApplicationMaster`'s exit message), which today can print `null` or a misattributed failure instead of the real error.

### Provenance

The vendored copy diverges from both upstreams: parquet-common (`org.apache.parquet.util.DynMethods`, unchanged since PARQUET-777 in 2017) and Iceberg (`org.apache.iceberg.common.DynMethods`) both re-throw the already-unwrapped `e` itself (`throwIfInstance(e, RuntimeException.class); throw new RuntimeException(e);`). The double-unwrap was introduced by the vendoring commit db0047d51 ([KYUUBI #3230]) when the upstream helper calls were hand-translated into instanceof checks on `e.getCause()`. parquet's own `TestDynMethods#testExceptionThrown` pins the correct semantics, which the current code fails.

The sibling `DynConstructors.Ctor.newInstance` in the same package already implements the correct pattern.

### Fix direction

Replace the second unwrap with: re-throw `RuntimeException` as-is and wrap checked exceptions with the real exception as cause, matching `DynConstructors.Ctor.newInstance` and upstream. Add unit tests for `org.apache.kyuubi.util.reflect`, which currently has none.

LuciferYang:

Fix #7640.

`UnboundMethod.invoke` unwrapped exceptions twice: `invokeChecked` already throws the real exception produced by the invoked method, but `invoke` re-threw `e.getCause()` as if `e` were still the `InvocationTargetException`. Consequences:

- an exception without a cause degraded to `new RuntimeException(null)` — the type, message and stack trace of the real failure were all lost;
- a `RuntimeException` carrying a cause was silently replaced by that cause, so `catch (IllegalStateException)` at a call site missed an `IllegalStateException` thrown by the target.

The fix re-throws `RuntimeException` as-is and wraps checked exceptions with the real exception as cause, matching the sibling `DynConstructors.Ctor.newInstance` and restoring the semantics both upstreams always had (parquet-common, unchanged since PARQUET-777; Iceberg) and that were lost during vendoring in db0047d51 ([KYUUBI #3230]).

All in-repo `.invoke`/`.invokeChecked` call sites were audited: none depends on the old behavior. Callers going through `ReflectUtils.invokeAs` keep their top-level message but now carry the real cause; direct users of the wrappers (`EmbeddedExecutorFactory`, `KyuubiSparkUtil`, `EngineTab`, `RowSet`) now see the real exception type and message.

Added `DynMethodsTest`, the first unit tests for `org.apache.kyuubi.util.reflect`: 9 cases covering both failure modes of the old code, a causeless-`RuntimeException` identity pin (`assertSame`), a checked-exception-carrying-runtime-cause pin, the `BoundMethod`/`StaticMethod` wrappers, fixed-arity and varargs happy paths, and the `invokeChecked` contract.

Assisted-by: GLM-5.3
Assisted-by: Claude Opus

### DIFF
--- a/kyuubi-util/src/test/java/org/apache/kyuubi/util/reflect/DynMethodsTest.java
+++ b/kyuubi-util/src/test/java/org/apache/kyuubi/util/reflect/DynMethodsTest.java
@@ -29,7 +29,6 @@ public class DynMethodsTest {
 
   public static class ReflectionTarget {
 
-    /** Kept as a field so a test can assert the very same instance came back out of invoke. */
     final IllegalStateException causelessFailure = new IllegalStateException();
 
     public Object throwCheckedWithoutCause() throws IOException {
@@ -86,10 +85,6 @@ public class DynMethodsTest {
     RuntimeException thrown =
         assertThrows(RuntimeException.class, () -> method.invoke(new ReflectionTarget()));
 
-    // Before the fix, invoke unwrapped a second time and threw the IllegalArgumentException
-    // bare, dropping the target's IOException: a plausible-looking wrong type in place of the
-    // real failure. The IOException must now arrive as the wrapper's cause, with its own cause
-    // still attached.
     assertEquals(RuntimeException.class, thrown.getClass());
     assertTrue(thrown.getCause() instanceof IOException);
     assertEquals("checked failure", thrown.getCause().getMessage());
--- a/kyuubi-util/src/test/java/org/apache/kyuubi/util/reflect/DynMethodsTest.java
+++ b/kyuubi-util/src/test/java/org/apache/kyuubi/util/reflect/DynMethodsTest.java
@@ -1,0 +1,183 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.util.reflect;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+public class DynMethodsTest {
+
+  public static class ReflectionTarget {
+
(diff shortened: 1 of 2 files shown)


LuciferYang:
cc @pan3793

wForget (approved):
Thanks @LuciferYang

kyuubi-util/src/test/java/org/apache/kyuubi/util/reflect/DynMethodsTest.java
```diff
@@ -0,0 +1,183 @@
...
+    DynMethods.UnboundMethod method =
+        DynMethods.builder("throwCheckedWithRuntimeCause").impl(ReflectionTarget.class).build();
+
+    RuntimeException thrown =
+        assertThrows(RuntimeException.class, () -> method.invoke(new ReflectionTarget()));
+
+    // Before the fix, invoke unwrapped a second time and threw the IllegalArgumentException
+    // bare, dropping the target's IOException: a plausible-looking wrong type in place of the
+    // real failure. The IOException must now arrive as the wrapper's cause, with its own cause
+    // still attached.
```
pan3793: I think it's unnecessary to write the bug history in the code comments (it should stay in the PR description and commit message instead), and since the test logic is simple, let's remove the comment

pan3793 (approved):

kyuubi-util/src/test/java/org/apache/kyuubi/util/reflect/DynMethodsTest.java
LuciferYang: Removed. I also dropped the javadoc on the `causelessFailure` field, for the same reason: it narrated the test's intent, which `assertSame` already shows. I kept the note on the varargs `invoke` call because it documents `Method.invoke` behavior rather than this bug, but I can remove that one too if you'd rather.

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/f316cd38175839394b4b996b29c72614ef76bfc8
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/c6e24e85241dbd708c8c31643189b0cca7cb135c
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/cca7a66e4e87319b1e315198e5645cfe0b78ed80

LuciferYang:
Thank you @pan3793 and @wForget

PR #7642, opened 2026-08-21
[KYUUBI #7645][SERVER] Fix operation state metrics transition

HanShuhang:

`KyuubiOperation#setState` updated operation state metrics before delegating to `AbstractOperation#setState`, where the state transition is validated and the actual operation state is changed. It also did not protect the whole transition-and-metrics-update sequence with the operation lock.

This can cause inaccurate operation state metrics in two cases:

1. A stale or invalid state transition may update metrics before being rejected.

    For example, an operation may have already moved from `RUNNING` to `FINISHED`, while a delayed cancel/close path still tries to update it to another terminal state such as `CANCELED`. In the old implementation, the target state metric was updated before `OperationState#validateTransition` rejected the transition, so the operation remained in the original state but the metrics had already been changed.

2. Concurrent state transitions for the same operation may update metrics based on the same stale old state.

    For example, one thread may observe the remote query as finished and call `KyuubiOperation#setState(FINISHED)`, while another thread concurrently handles cancel/close/timeout/error and calls `KyuubiOperation#setState(CANCELED)` or another terminal state. Since the old implementation read `state` and updated metrics outside a common operation lock, both paths could observe the old state as `RUNNING` and both update the `RUNNING` metric, making `kyuubi.operation.state.ExecuteStatement.running` inaccurate.

This patch makes `KyuubiOperation#setState` execute under the operation lock, captures the old state once, delegates to `AbstractOperation#setState` first, and updates operation state metrics only after the state transition succeeds.

Added a unit test in `KyuubiOperationSuite` to verify that a stale terminal transition does not update operation state metrics.
```
  -Pspark-provided -Pflink-provided -Phive-provided \
  -Dtest=none \
  -DwildcardSuites=org.apache.kyuubi.operation.KyuubiOperationSuite
```

Assisted-by: TraeCode with GPT-5

### DIFF
--- a/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationSuite.scala
+++ b/kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationSuite.scala
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.operation
+
+import com.codahale.metrics.MetricRegistry
+import org.mockito.Mockito.when
+import org.scalatestplus.mockito.MockitoSugar.mock
+
+import org.apache.kyuubi.{KyuubiFunSuite, KyuubiSQLException}
+import org.apache.kyuubi.config.KyuubiConf
+import org.apache.kyuubi.metrics.{MetricsConf, MetricsConstants, MetricsSystem}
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/test/scala/org/apache/kyuubi/operation/KyuubiOperationSuite.scala
```diff
@@ -0,0 +1,118 @@
...
+    val session = mock[Session]
+    when(session.protocol).thenReturn(HIVE_CLI_SERVICE_PROTOCOL_V10)
+    when(session.handle).thenReturn(SessionHandle())
+    when(session.user).thenReturn("kyuubi")
+    when(session.sessionManager).thenReturn(sessionManager)
+    session
+  }
+}
+
+private class TestKyuubiOperation(session: Session) extends KyuubiOperation(session) {
```
yikf: Hi @HanShuhang , thanks for working on this, left two minor points, let me know what you think.

- Please file/link an issue and add '[KYUUBI #NNNN]' to the title.

- `TestKyuubiOperation` is over-implemented — `KyuubiOperation` already implements everything except `runInternal`. Trim to:
```suggestion
private class TestKyuubiOperation(session: Session) extends KyuubiOperation(session) {
      def transitState(newState: OperationState): Unit = setState(newState)
      override protected def runInternal(): Unit = {}
}
```

HanShuhang: Thanks! I agree with your suggestions and have updated the PR accordingly.

wangzhigang1999 (approved):
LGTM

yikf (approved):

wangzhigang1999:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/c24e02a65d73a411cb08fd69b6ca279b41fd269a

PR #7648, opened 2026-08-22
[KYUUBI #7647] Stop truncating varargs constructor arguments in DynConstructors

### LINKED ISSUE #7647: DynConstructors silently truncates extra arguments to varargs constructors
### Description

`DynConstructors.Ctor.newInstanceChecked` truncates arguments to the resolved constructor's parameter count unconditionally:

```java
if (args.length > ctor.getParameterCount()) {
  return ctor.newInstance(Arrays.copyOfRange(args, 0, ctor.getParameterCount()));
}
```

`Constructor.newInstance` — like `Method.invoke` — never packs loose varargs arguments: the only valid way to pass varargs through reflection is the packed array as the last argument (loose trailing args throw `IllegalArgumentException: wrong number of arguments`). So for a **varargs** constructor the truncation can never produce a correct call; it has two failure shapes:

- if the truncated prefix happens to type-match the formal parameters, the trailing arguments are **silently dropped** and a wrong instance is constructed. Example: for `Holder(String first, Object... rest)` (2 params), `ctor.newInstance("one", new Object[]{"2a","2b"}, "three")` constructs `Holder("one", ["2a","2b"])` — `"three"` vanishes with no error;
- otherwise the caller gets a misleading `argument type mismatch` (an artifact of the truncated call) instead of the honest `wrong number of arguments`.

### Upstream lineage

Although the class header says "Copied from parquet-common", the truncation logic matches Apache Iceberg's copy, not parquet's: parquet-mr's `newInstanceChecked` has never truncated (it passes `args` straight through, and its `TestDynConstructors#testFirstImplReturned` pins extra-arguments-must-throw), while Iceberg's copy carries the identical unconditional truncation with no `isVarArgs` guard, no fixing commit in its history, and no test covering extra or varargs arguments. The fixed-arity truncation is long-standing in our copy and is load-bearing (cross-version constructor compatibility in `TPCDSTable`, `HiveSessionManager`, `HiveTableCatalog` and others), so it must be kept.

### Fix direction

Skip the truncation when `ctor.isVarArgs()`: varargs constructors' arguments pass through unchanged, so callers get either a working packed-array invocation or the same honest `IllegalArgumentException: wrong number of arguments` that parquet-mr specifies upstream. Add unit tests for `DynConstructors` (none exist today). Possibly worth reporting upstream against Iceberg's `org.apache.iceberg.common.DynConstructors` separately.

LuciferYang:

Fix #7647.

`Ctor.newInstanceChecked` truncated arguments to the resolved constructor's parameter count unconditionally. `Constructor.newInstance` — like `Method.invoke` — never packs loose varargs arguments (the only valid reflection form is the packed array as the last argument), so for a **varargs** constructor the truncation could never produce a correct call:

- when the truncated prefix happened to type-match, trailing arguments were **silently dropped** and a wrong instance was constructed — e.g. for `Holder(String, Object...)`, `ctor.newInstance("one", new Object[]{"2a","2b"}, "three")` constructed `Holder("one", ["2a","2b"])` with `"three"` vanishing;
- otherwise the caller got a misleading `argument type mismatch` (an artifact of the truncated call) instead of the honest `wrong number of arguments`.

The fix skips the truncation when `ctor.isVarArgs()`: varargs constructors' arguments pass through unchanged, so callers get either a working packed-array invocation or the same fail-loudly contract parquet-mr specifies upstream.

This aligns `DynConstructors` with the sibling class in the same package rather than introducing a local special case: `DynMethods.UnboundMethod` keeps `argLength` at -1 for a varargs method and so skips the identical `Arrays.copyOfRange` truncation, both in this copy and upstream, and `KyuubiSparkUtil`'s `UriBuilder.build(Object...)` lookups depend on that.

Upstream lineage: this class is adapted from iceberg-common (itself derived from parquet-common), and the truncation logic matches Iceberg's copy — parquet-mr's `newInstanceChecked` has never truncated (its `TestDynConstructors#testFirstImplReturned` pins extra-arguments-must-throw), while Iceberg's copy carries the identical unconditional truncation, unfixed and untested for extra/varargs arguments. This is therefore a Kyuubi-local fix, recorded on `newInstanceChecked`; it may be worth reporting upstream against Iceberg separately.

Scope: the fixed-arity truncation is long-standing and load-bearing — 7 in-repo sites rely on it for cross-version constructor compatibility (`TPCDSTable` across Spark versions, `HiveSessionManager` on Hive 2.3, and optional-conf provider patterns such as `EngineSecuritySecretProvider`, which binds a no-arg constructor and passes `conf` on every call) — so it is deliberately preserved. All in-repo `DynConstructors` call sites were audited: none passes loose arguments to a varargs constructor (structurally impossible to bind one without an explicit array-typed `impl` chain, of which the repo has none for constructors), so no caller's behavior changes except the intended silent-drop → loud-failure correction.

Added `DynConstructorsTest`, the first unit tests for `DynConstructors`:

- rejection of loose varargs arguments in both forms (packed-plus-extras and all-loose, asserting the honest `wrong number of arguments` message by prefix, since the JDK 18+ reflection accessor appends `: 3 expected: 2`) plus the under-arity call (reflection synthesizes no empty varargs array either);
- the non-null-target guards of `invoke`/`invokeChecked`;
- the packed-array happy path through all four entry points (`newInstance`, `newInstanceChecked`, `invoke`, `invokeChecked`);
- the preserved fixed-arity truncation boundary.

Assisted-by: Claude Code Opus 5

### DIFF
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynConstructors.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynConstructors.java
@@ -27,12 +27,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
-/**
- * Adapted from iceberg-common, which is itself derived from parquet-common.
- *
- * <p>Diverges from both upstreams: extra arguments are truncated only for non-varargs constructors;
- * varargs constructors' arguments are passed through to {@link Constructor#newInstance} unchanged.
- */
+/** Adapted from iceberg-common, which is itself derived from parquet-common. */
 public class DynConstructors {
 
   private DynConstructors() {}
@@ -51,6 +46,17 @@ public class DynConstructors {
       return constructed;
     }
 
+    /**
+     * Creates an instance, truncating extra arguments only for a non-varargs constructor.
+     *
+     * <p>{@link Constructor#newInstance} never packs loose varargs arguments into the trailing
+     * array, so truncating such a call could only drop arguments or fail with a misleading message.
+     * {@link DynMethods.UnboundMethod} skips the same truncation for varargs methods, by keeping
+     * {@code argLength} at -1.
+     *
+     * <p>Both upstream copies of this class differ here: iceberg-common truncates unconditionally,
+     * parquet-common does not truncate at all.
+     */
     public C newInstanceChecked(Object... args) throws Exception {
       try {
         if (!ctor.isVarArgs() && args.length > ctor.getParameterCount()) {
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynConstructors.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynConstructors.java
@@ -27,7 +27,12 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
-/** Copied from parquet-common */
+/**
+ * Adapted from iceberg-common, which is itself derived from parquet-common.
+ *
+ * <p>Diverges from both upstreams: extra arguments are truncated only for non-varargs constructors;
+ * varargs constructors' arguments are passed through to {@link Constructor#newInstance} unchanged.
+ */
 public class DynConstructors {
 
   private DynConstructors() {}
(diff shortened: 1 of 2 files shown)


LuciferYang:
cc @pan3793

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/797ead5dfb2e3b14cb51ce75b20b0bf700984cf9
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/0d7fa6e54ff86af94d99c2ada75b63fc6e4d4634

LuciferYang:
Thank you @pan3793

Issue #7659, opened 2026-08-25
[BUG] Code coverage reports are MIA: the test argLine override drops the JaCoCo agent


acruise:
## Describe the bug

`-Pcodecov` doesn't seem to produce any coverage for any module, and hasn't since 1.8.0. The build stays green, so nothing signals the failure — the report is simply empty.

Both `maven-surefire-plugin` and `scalatest-maven-plugin` default their `argLine` to the `${argLine}` property, resolved when the mojo runs — late enough to pick up what `jacoco:prepare-agent` publishes there. KYUUBI #4849 gave both plugins an explicit `<argLine>${extraJavaTestArgs}</argLine>` to open JDK modules. Plugin configuration is interpolated during model building, before the agent mojo runs, so the override wins, the agent option is dropped, and the forked test JVM runs uninstrumented:

```
[INFO] --- jacoco:0.8.11:prepare-agent (pre-test) @ kyuubi-util-scala_2.12 ---
[INFO] argLine set to -javaagent:.../org.jacoco.agent-0.8.11-runtime.jar=destfile=.../jacoco.exec
[INFO] --- jacoco:0.8.11:report (report) @ kyuubi-util-scala_2.12 ---
[INFO] Skipping JaCoCo execution due to missing execution data file.
```

To reproduce: `build/mvn test -Pcodecov -pl kyuubi-util-scala` — BUILD SUCCESS, 21 tests pass, no `target/jacoco.exec`, no report.

This is a repeat of KYUUBI #756 ("Recover CODECOV", 2021), which fixed the same thing by deleting an `argLine` element.

## Affects Version(s)

1.8.0 and later, including master. `v1.7.4` is the last release without the override: branch-0.2 through branch-1.7 have zero `<argLine>` elements in `pom.xml`; branch-1.8 through branch-1.12 and master have two.

## Additional context

Fixable in three lines by moving the JDK options into the `argLine` property itself and dropping both plugin overrides, so the defaults apply again — `prepare-agent` prepends the agent to the property's existing value, so `--add-opens` survives. Surefire's `@{argLine}` late replacement is not sufficient on its own, since `scalatest-maven-plugin` has no equivalent and that is where most of Kyuubi's tests run. PR to follow.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

[...]

PR #7661, opened 2026-08-25
[KYUUBI #7659][INFRA] Restore code coverage 

acruise:
...by not overriding the test argLine!

Code coverage has reported nothing since 1.8.0. Both maven-surefire-plugin and scalatest-maven-plugin default their `argLine` to the `${argLine}` property, which is resolved when the mojo runs -- late enough to see the value `jacoco:prepare-agent` publishes there. KYUUBI #4849 gave both plugins an explicit `<argLine>${extraJavaTestArgs}</argLine>` to open JDK modules, and plugin configuration is interpolated during model building, long before the agent mojo runs. The agent option is therefore dropped and the forked test JVM runs uninstrumented:

    [INFO] --- jacoco:0.8.11:prepare-agent (pre-test) @ kyuubi-util-scala_2.12 ---
    [INFO] argLine set to -javaagent:.../org.jacoco.agent-0.8.11-runtime.jar=...
    [INFO] --- jacoco:0.8.11:report (report) @ kyuubi-util-scala_2.12 ---
    [INFO] Skipping JaCoCo execution due to missing execution data file.

The build stays green and every module reports zero. KYUUBI #756 ("Recover CODECOV") fixed the same thing in 2021 by deleting an `argLine` element.

Move the JDK options into the `argLine` property itself and drop both overrides, so the plugin defaults apply again. `prepare-agent` prepends the agent to the property's existing value rather than replacing it, so `--add-opens` survives:

    argLine set to -javaagent:...=destfile=... -XX:+IgnoreUnrecognizedVMOptions ...

Surefire's `@{argLine}` late replacement is not enough on its own here, since scalatest-maven-plugin has no equivalent and that is where most of Kyuubi's tests run.

Verified locally:

  - `test -Pcodecov -pl kyuubi-util-scala` (scalatest) analyzes 8 classes and writes target/jacoco.exec, where before it skipped for missing data.
  - `test -Pcodecov -pl kyuubi-rest-client` (surefire) analyzes 63 classes.
  - `test -pl kyuubi-util-scala` without the profile is green, argLine unchanged.

Yes, Claude Opus 5

### DIFF
--- a/externals/kyuubi-data-agent-engine/pom.xml
+++ b/externals/kyuubi-data-agent-engine/pom.xml
@@ -187,7 +187,6 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <skipTests>${skipTests}</skipTests>
-                    <argLine>${extraJavaTestArgs}</argLine>
                 </configuration>
             </plugin>
             <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -289,6 +289,10 @@
         <distMgmtSnapshotsName>Apache Development Snapshot Repository</distMgmtSnapshotsName>
         <distMgmtSnapshotsUrl>https://repository.apache.org/content/repositories/snapshots</distMgmtSnapshotsUrl>
 
+        <!-- Both test plugins default their argLine to ${argLine}; jacoco:prepare-agent
+             prepends the agent to it under -Pcodecov. Overriding argLine in plugin config
+             instead breaks that, since plugin config is interpolated before the agent runs. -->
+        <argLine>${extraJavaTestArgs}</argLine>
         <extraJavaTestArgs>-XX:+IgnoreUnrecognizedVMOptions
             --add-opens=java.base/java.lang=ALL-UNNAMED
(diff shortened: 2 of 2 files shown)


wangzhigang1999:
Thanks for fixing this. I found one remaining `argLine` override in `externals/kyuubi-data-agent-engine/pom.xml`:

```xml
<argLine>${extraJavaTestArgs}</argLine>
```

It has the same effect on the Data Agent JUnit tests. With the current patch, `TableRefTest` runs 17 tests successfully, but JaCoCo reports zero covered lines for the module. After removing this line, the same tests pass and JaCoCo records 21 covered lines for `TableRef`.

Could you remove this override as part of this PR?

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/a162cde27e43cb19f1df30ed596db743682218e1
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/e959286bfefc70a8dbbceb409cb2225e308a2793

Issue #7662, opened 2026-08-26
[FEATURE] Extend virtual thread support to I/O-bound Kyuubi Server executors


wangzhigang1999:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct).

### Search before asking

- [x] I have searched the existing issues and found no similar issue.

### Describe the feature

PR #7656 added optional virtual-thread support to the Thrift Binary frontend. This issue proposes extending it to other I/O-bound Kyuubi Server executors, including HTTP frontends, operation execution, Server-to-Engine RPC, engine lifecycle helpers, metadata workers, batch submission, Data Agent, and Kubernetes operations.

Engine-side executors remain out of scope.

### Motivation

Kyuubi Server is an I/O gateway. These executors spend much of their time waiting for network calls, engine responses, subprocess output, or cluster APIs, which makes them suitable for virtual threads.

Benchmark TL;DR:

We compared platform threads with All Virtual Threads (All VT) on Java Development Kit (JDK) 25.0.3 and Spark 4.2 in Kubernetes cluster mode. Each cell used a 60-second warm-up and a 300-second measurement window. Server and Engine ran in separate Pods.

Throughput is reported as completed operations per second. The p50 value is median latency. Native Memory Tracking (NMT) is the JVM facility used to measure committed native thread memory.

| Workload and metric | Platform threads | All VT | Difference |
|---|---:|---:|---:|
| Connection churn throughput | 558.7 ops/s | 745.0 ops/s | **+33.4%** |
| Connection churn p50 | 681.8 ms | 411.4 ms | **-39.7%** |
| Connection churn NMT thread memory | 120.8 MiB | 4.6 MiB | **-96.2%** |
| Operation lifecycle throughput | 108.0 ops/s | 104.2 ops/s | -3.5% |
| Operation lifecycle NMT thread memory | 124.1 MiB | 5.6 MiB | **-95.5%** |
| Metadata throughput | 5428.4 ops/s | 5430.0 ops/s | Flat |
| Metadata Server CPU | 4.58 cores | 4.11 cores | -10.2% |
| Metadata container memory | 3322.6 MiB | 3156.8 MiB | -5.0% |
| Short Spark SQL throughput | 13.230 ops/s | 13.227 ops/s | Flat |
| 10 GB Spark SQL throughput | 0.337 ops/s | 0.347 ops/s | +3.0% |

[...]

PR #7663, opened 2026-08-26
[KYUUBI #7662][SERVER] Extend virtual thread support to server executors

wangzhigang1999:

PR #7656 added virtual-thread support to the Binary frontend. This follow-up extends the same optional model to other I/O-bound Kyuubi Server executors and adds one global opt-in switch with component-level overrides.

The feature remains disabled by default. Engine-side executors are unchanged. See #7662 for the scope and benchmark results.

- Targeted configuration, thread utility, session manager, and Jetty suites on JDK 17 and JDK 21.
- Kyuubi Common test suite on JDK 21: 295 tests passed.
- Kyuubi Server fast package build.
- `dev/reformat`
- `dev/gen/gen_all_config_docs.sh`
- `git diff --check`
- Five-minute JDK 25 A/B benchmark with Spark 4.2 and real Spark SQL. See #7662.

Assisted-by: OpenAI Codex (GPT-5)


kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -635,6 +635,19 @@ object KyuubiConf {
       .immutable
       .fallbackConf(FRONTEND_BIND_PORT)
 
+  val SERVER_ALL_VIRTUAL_THREADS_ENABLED: ConfigEntry[Boolean] =
+    buildConf("kyuubi.server.virtualThreads.all.enabled")
```
pan3793:
```suggestion
    buildConf("kyuubi.server.virtualThreads.enabled")
```

wangzhigang1999: SGTM, thanks!

wangzhigang1999:
Hi @pan3793, this PR has changed quite a bit since your earlier review. I adopted your suggested `kyuubi.server.virtualThreads.enabled` umbrella switch, extended virtual-thread support to selected server-side blocking executors, and removed the Jetty/HTTP frontend changes after further review to avoid changing Jetty’s queueing semantics.

The scope is now stable. All virtual-thread options remain disabled by default, and explicit component-level settings take precedence over the umbrella switch. Could you take another look when you have time? Thanks!

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -635,6 +635,19 @@ object KyuubiConf {
       .immutable
       .fallbackConf(FRONTEND_BIND_PORT)
 
+  val SERVER_VIRTUAL_THREADS_ENABLED: ConfigEntry[Boolean] =
+    buildConf("kyuubi.server.virtualThreads.enabled")
+      .doc("Whether to use virtual threads by default for all supported executors in the " +
+        "Kyuubi server. This requires Java 21 or later. Enabling this option is recommended for " +
+        "I/O-bound Kyuubi Server deployments on Java 25 or later. On Java 21, enable virtual " +
+        "threads selectively after validating synchronized blocking paths for carrier-thread " +
+        "pinning. An explicitly configured component-level virtual thread option takes precedence.")
```
pan3793: shorten the words for the Java version requirement to:

Requires at least Java 21; Java 25 or later is recommended.

also apply this to all component-level configs docs

pan3793 (commented):
Thanks for the PR. The overall shape looks good: opt-in global switch via `fallbackConf`, `audience(SERVER)` + `immutable` on every entry, engine side untouched, and solid test coverage including the JDK < 21 negative path. A few comments:

**1. Misconfiguration surfaces too late and too obscurely on JDK < 21**

Only `SessionManager.initialize` and the binary frontend fail at startup; the other converted paths fail lazily (`ProcBuilder`'s `lazy val` factory at first engine launch, `KyuubiSyncThriftClient` at first session RPC, `MetadataManager` / `DataAgentResource` at first use), with a reflective `IllegalStateException` deep in the stack.

Suggest failing fast at the config layer: keep a list of all VT entries in `object KyuubiConf` right after their definitions (essentially the `serverVirtualThreadConfigs` list that `KyuubiConfSuite` already builds — move it to main code), and add a check method in `case class KyuubiConf` that reads each entry and throws if any resolves to `true` while `!ThreadUtils.isVirtualThreadSupported`, naming the offending key. Invoke it once at server startup, after the conf is fully loaded — note it cannot run at `KyuubiConf` construction time because `loadFileDefaults()` populates values after the constructor. This turns every lazy runtime failure into one clear boot-time error, and the list keeps future VT entries from being forgotten.

**2. Virtual threads lose the Kyuubi uncaught exception handler**

`NamedThreadFactory` installs `kyuubiUncaughtExceptionHandler`; `ThreadUtils.newVirtualThreadFactory` does not. For executor-managed tasks this rarely matters (`FutureTask` captures exceptions), but `ProcBuilder` starts a raw thread from the factory, so the log-capture path regresses from Kyuubi-logged to stderr-only uncaught errors. `Thread.Builder.uncaughtExceptionHandler(...)` exists on JDK 21 and can be wired through the same reflection.

**3. `getPoolSize` double-counts against the queue gauge**

`BoundedQueuedExecutorService.getPoolSize` returns `activeCount + queueSize`, while `KyuubiSessionManager` registers `EXEC_POOL_ALIVE` and `EXEC_POOL_WORK_QUEUE_SIZE` as separate gauges (also exposed via `SessionsResource`). With platform pools, pool size excludes queued tasks, so queued tasks now show up in both gauges. Since there is no idle-worker concept in the VT model, `getPoolSize = getActiveCount` would keep dashboards comparable.

**4. FIFO ordering is no longer guaranteed — worth a code comment**

The fair `runningTasks` semaphore orders by arrival at `acquire()`, and VT start order is not guaranteed to match submission order. Current call sites are safe (I checked `withLockAcquiredAsyncRequest`: it submits and blocks on `task.get()` while holding `lock`, so at most one task is ever in flight), but a short comment on `BoundedQueuedExecutorService` would prevent a future call site from relying on FIFO. Related nit: given that serialization, the `Int.MaxValue - 1` queue for `asyncRequestExecutor` can never hold more than one task; a smaller value or a comment would state intent better.

**5. Component docs omit the inheritance**

The generated `settings.md` renders a literal `false` default for every component entry, and none of the `.doc()` texts mention they inherit `kyuubi.server.virtualThreads.enabled`. Adding "Defaults to `kyuubi.server.virtualThreads.enabled`." to each component doc would make the rendered docs self-contained.

kyuubi-common/src/main/scala/org/apache/kyuubi/util/ThreadUtils.scala
```diff
@@ -26,29 +27,70 @@ import org.apache.kyuubi.{KyuubiException, Logging}
...
-    new BoundedExecutorService(
-      newVirtualThreadPerTaskExecutor(threadNamePrefix),
-      maxConcurrentTasks)
+  trait ExecutorServiceWithMetrics extends ExecutorService {
+    def getPoolSize: Int
+    def getActiveCount: Int
+    def getQueueSize: Int
+  }
+
+  def isVirtualThreadSupported: Boolean = {
```
pan3793: lazy val ?

pan3793 (approved):
LGTM, one minor suggestion.

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/888bf0e0ef2c23ea5fec707ed815de239c5f6933

Issue #7434, opened 2026-05-07
:memo: Consolidate a documentation format to a single approach


dnskr:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### What type of changes will we make to the documentation?

Refactoring

### Affects Version(s)

master/1.11

### **Description**
Currently, the Apache Kyuubi documentation uses a hybrid approach, mixing **reStructuredText (.rst)** and **Markdown (.md)** files, while being processed by **Sphinx**. 

While this hybrid approach provides flexibility (using `myst-parser` for Markdown), it introduces several challenges:
1. Contributors need to be familiar with both syntaxes.
2. Cross-referencing between RST and MD files can be error-prone and hard to detect.
3. The build pipeline relies on multiple Sphinx extensions (like `myst-parser`, `sphinx-markdown-tables`) to bridge the gap.
4. Formatting features available in RST might not be easily replicable in Markdown (or vice-versa).

**I propose migrating to a single technology to streamline the contribution process and improve documentation quality.**

---

### **Technology Comparison**

Below is an analysis of the two primary paths for consolidation:

#### **Option 1: Migrate to Markdown (using MyST-Parser)**
*   **Pros:**
    *   Markdown is the de facto standard for most developers, so it is easier for new contributors.
    *   Better preview support in GitHub, IDEs, and other tools.
    *   MyST (Markedly Structured Text) provides almost all the power of RST (directives, roles) while remaining valid Markdown.
*   **Cons:**
    *   Once we start using advanced features, the Markdown "cleanliness" can get cluttered with curly braces and colons.
    *   Some advanced Sphinx features or specialized extensions may still favor RST's native structure.

#### **Option 2: Migrate to reStructuredText (RST)**
*   **Pros:**
    *   RST is the native format for Sphinx, offering the most robust support for technical documentation features.
    *   Eliminates the need for "bridge" plugins like `myst-parser`.
*   **Cons:**
    *   RST syntax is more verbose and less intuitive for casual contributors compared to Markdown.
    *   Limited rendering outside of Sphinx-based environments.

---

### **Action Plan**
1.  Decide on the target format (Markdown/MyST vs. RST).
2.  Convert a few effortless directories to the target format using automated tools or AI agents to verify the approach.
[...]

PR #7666, opened 2026-08-26
[KYUUBI #7434][DOC] Reformat quick start docs from RST to Markdown

dnskr:
The changes are needed to unify the format used for documentation, as described in the issue https://github.com/apache/kyuubi/issues/7434.

The PR also:
- Removes `sphinx_markdown_tables` extension to fix variable substitution in tables.
- Fixes cross-references to `quick_start` pages.
- Fixes `building kyuubi` link.

There is a slight difference between the configuration table representation on the `Getting Started` page and the Maven dependency declaration on the `Getting Started with Hive JDBC` page because variable substitution doesn't work inside code blocks, but it shouldn't affect user experience.

This PR **does not** change page contents to ensure an easier review of the migration.

Tested by building the documentation and comparing the pages against the live site:
```
make clean html


open https://kyuubi.readthedocs.io/en/master/quick_start/index.html
open _build/html/quick_start/index.html

open https://kyuubi.readthedocs.io/en/master/quick_start/quick_start.html
open _build/html/quick_start/quick_start.html

open https://kyuubi.readthedocs.io/en/master/quick_start/quick_start_with_jdbc.html
open _build/html/quick_start/quick_start_with_jdbc.html
```

Assisted-by: Qwen Coder


docs/quick_start/quick_start_with_jdbc.md
```diff
@@ -0,0 +1,98 @@
...
+forked from Hive 3.1.x JDBC driver, aims to support some missing functionalities
+of the original Hive JDBC driver.
+
+The driver is available from Maven Central:
+
+```xml
+<dependency>
+    <groupId>org.apache.kyuubi</groupId>
+    <artifactId>kyuubi-hive-jdbc-shaded</artifactId>
+    <version>${kyuubi.version}</version>
```
pan3793, line 34: > variable substitution doesn't work inside code blocks

maybe we can ask AI to write a simple extension or pre-process it with a python function to make it work? but the current approach is also fine.

dnskr: I researched this while trying to make it work, and from what I understand, there's a design reason why substitutions don't work inside code blocks in Sphinx and MyST parser.
One solution is to substitute the entire code block, but that adds logic to `conf.py` and makes the Markdown file less friendly to read.
For now, I'm leaning toward avoiding the release value in code blocks when possible, in favor of something more general like this workaround. But I'll continue working on format migration, and I'll try to find more elegant solution.

dnskr: A quick update regarding the issue.

I've used `parsed-literal` code-block to render the following code and it works as intended:
```{parsed-literal}
$ tar zxf apache-kyuubi-{{ release }}-bin.tgz
```
image



However, in the current MyST implementation `parsed-literal` directive removes leading and trailing spaces (PR https://github.com/executablebooks/MyST-Parser/pull/1176 available), and doesn't work properly with XML content, so the discussed code does not rendered correctly:
```{parsed-literal}
<dependency>
    <artifactId>kyuubi-hive-jdbc-shaded</artifactId>
    <version>{{ release }}</version>
</dependency>
```
image

pan3793 (approved):
the current state is good enough to go. and we can re-evalute `parsed-literal` once the patched version is available.

dnskr:
Thank you for the review!

Merged to `master` and `branch-1.12`.

Issue #7664, opened 2026-08-26
[Bug] After Kerberos is enabled, no extra custom parameters can be passed to the Hive‑connector


CrazyBeeline:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

1. kerberos enabled (hive spark hadoop kyuubi)
2. set spark.sql.catalog.hivex=org.apache.kyuubi.spark.connector.hive.HiveTableCatalog;
use hivex;
create table test_bb (id int); 
drop table test_bb;
ok
3.set spark.sql.catalog.hivex2=org.apache.kyuubi.spark.connector.hive.HiveTableCatalog;
set spark.sql.catalog.hivex2.hive.metastore.uris=thrift://hadoopoe24x8602.qsdi.bigdata:19083; 
use hivex2;
create table test_bb (id int); 
error find 

javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)]
	at com.sun.security.sasl.gsskerb.GssKrb5Client.evaluateChallenge(GssKrb5Client.java:211)
	at org.apache.thrift.transport.TSaslClientTransport.handleSaslStartMessage(TSaslClientTransport.java:95)
	at org.apache.thrift.transport.TSaslTransport.open(TSaslTransport.java:271)
	at org.apache.thrift.transport.TSaslClientTransport.open(TSaslClientTransport.java:38)
	at org.apache.hadoop.hive.thrift.client.TUGIAssumingTransport$1.run(TUGIAssumingTransport.java:52)
	at org.apache.hadoop.hive.thrift.client.TUGIAssumingTransport$1.run(TUGIAssumingTransport.java:49)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1899)
	at org.apache.hadoop.hive.thrift.client.TUGIAssumingTransport.open(TUGIAssumingTransport.java:49)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.open(HiveMetaStoreClient.java:478)
[...]

PR #7667, opened 2026-08-27
[KYUUBI #7664][DOC] Reformat and improve docs for KSHC

maomaodev:

1. Reformat KSHC docs from RST to Markdown, as described in the issue https://github.com/apache/kyuubi/issues/7434.
2. Enrich the description for KSCH.

Tested by building the documentation and comparing the pages against the live site:
```
open https://kyuubi.readthedocs.io/en/master/connector/spark/hive.html
open _build/html/connector/spark/hive.html
```

Assisted-by: DeepSeek-V4-Pro


docs/connector/spark/hive.md
```diff
@@ -0,0 +1,166 @@
...
+implementation inside `spark-sql`, you should not expect KSHC works properly with `spark-sql`, and
+any issues caused by such a combination usage won't be considered at this time. Instead, it's recommended
+using BeeLine with Kyuubi as a drop-in replacement for `spark-sql`, or switching to `spark-shell`.
+
+KSHC supports accessing Kerberized Hive Metastore and HDFS, by using keytab, or TGT cache, or Delegation Token.
+It's not expected to work properly with multiple KDC instances, the limitation comes from JDK Krb5LoginModule,
+for such cases, consider setting up Cross-Realm Kerberos trusts, then you just need to talk with one KDC.
+
+For HMS Thrift API used by Spark, it's known that Hive 2.3.9 client is compatible with HMS from 2.1 to 4.0, and
+Hive 2.3.10 client is compatible with HMS from 1.1 to 4.0, such version combinations should cover the most cases.
```
pan3793: the claim is no longer valid; HIVE-26537 (4.1.0, 4.0.1) actually broke the Thrift wire protocol by removing the Thrift API used by Spark (the `get_table` method).

docs/connector/spark/hive.md
```diff
@@ -0,0 +1,166 @@
...
+SerDe.
+
+Since v1.12.0, KSHC supports the `PURGE TABLE` command. By default, `DROP TABLE` moves the table
+data to the HDFS trash. Set `spark.sql.kyuubi.hive.connector.dropTableAsPurgeTable` to `true` to
+make `DROP TABLE` skip the HDFS trash and completely remove its data, behaving like `PURGE TABLE`.
+
+Since v1.13.0, KSHC supports Dynamic Partition Pruning (DPP) for partitioned Hive tables, which
+significantly reduces the amount of data scanned when joining against large partitioned tables.
+DPP is implemented by wrapping Spark's built-in `ParquetScan` / `OrcScan`. As a result, native
+engines such as Gluten and Comet cannot recognize these scans by class name, so KSHC-converted
```
pan3793: we can promise nothing for those ecosystem projects because they use the spark private api, I would drop this statement

docs/connector/spark/hive.md
pan3793: so let's be conservative, replace the 4.0 with 3.1

docs/connector/spark/hive.md
```diff
@@ -0,0 +1,166 @@
...
+Hive 2.3.10 client is compatible with HMS from 1.1 to 4.0, such version combinations should cover the most cases.
+For other corner cases, KSHC also supports `spark.sql.catalog.<catalog_name>.spark.sql.hive.metastore.jars` and
+`spark.sql.catalog.<catalog_name>.spark.sql.hive.metastore.version` as well as the Spark built-in Hive datasource
+does, you can refer to the Spark documentation for details.
+
+## Limitations
+
+Currently, KSHC has the following limitations:
+
+- KSHC does not support Hive functions / UDFs.
```
pan3793: this might confuse and scare users, since it can be interpreted as:

"all my existing Hive UDFs, including those `ADD JAR ...` and `CREATE TEMPORARY FUNCTION ...`, will stop working if I want to use KSHC"

docs/connector/spark/hive.md
```diff
@@ -0,0 +1,166 @@
...
+| `spark.sql.kyuubi.hive.connector.dropTableAsPurgeTable`        | `false`           | When enabled, `DROP TABLE` completely removes its data by skipping HDFS trash, equivalent to the `PURGE TABLE` command.                                                                                                    | boolean | 1.12.0 |
+| `spark.sql.catalog.<catalog>.hive.metastore.warehouse.dir`     | &lt;undefined&gt; | The default warehouse directory for the catalog, taking precedence over the global `spark.sql.warehouse.dir` when creating a database.                                                                                     | string  | 1.12.0 |
+
+```{note}
+Catalog-level configurations (`spark.sql.catalog.<catalog>.*`) are captured when the catalog is
+initialized, so they must be present at the Spark application bootstrap (in `spark-defaults.conf`
+or via `--conf`). Setting them through `SET` after the catalog has been created does not take
```
pan3793: > ... so they must be present at the Spark application bootstrap (in `spark-defaults.conf` or via `--conf`).

I don't think the "must" is correct, the V2 session catalog is lazily initialized and session-isolated, so technically configs will take effect as long as it set before the catalog is initialized, of course, for `ONE_FOR_ALL`, configs only apply to the first initialized Hive client instance.

What matters here is that, for the kerberized cluster mode, if you want to distribute HMS tokens (non-keytab mode), you must declare all catalogs with metastore.uris at bootstrap so that `spark-submit` can fetch the HMS tokens ahead.

The above words are just an explanation, which do not fit the docs standard and need to be reorganized and polished

docs/connector/spark/hive.md
maomaodev: Agreed, replaced 4.0 with 3.1 for both Hive client versions.

docs/connector/spark/hive.md
maomaodev: Dropped the statement entirely.

docs/connector/spark/hive.md
maomaodev: Rephrased to clarify that only persistent views/UDFs (`CREATE VIEW / CREATE FUNCTION / SHOW VIEWS / SHOW FUNCTIONS`) fail through a KSHC catalog, while temporary views and temporary functions are unaffected. Also added `spark_catalog` as the workaround.

docs/connector/spark/hive.md
maomaodev: Good catch, the V2 catalog is lazily initialized and session-isolated, so configs apply as long as they're set before first access. Only the Kerberized cluster mode (no keytab) genuinely needs `metastore.uris` declared at bootstrap.

docs/connector/spark/hive.md
```diff
@@ -0,0 +1,172 @@
...
+
+(kshc-dependencies)=
+
+### Dependencies
+
+The **classpath** of Kyuubi Spark SQL engine with Hive connector supported consists of
+
+1. kyuubi-spark-sql-engine-{{ release }}_2.12.jar, the engine jar deployed with a Kyuubi distribution
+2. a copy of Spark distribution
+3. kyuubi-spark-connector-hive_2.12-{{ release }}, which can be found in the [Maven Central](https://mvnrepository.com/artifact/org.apache.kyuubi/kyuubi-spark-connector-hive)
```
pan3793:
```suggestion
1. kyuubi-spark-sql-engine_2.12-{{ release }}.jar, the engine jar deployed with a Kyuubi distribution
2. a copy of Spark distribution
3. kyuubi-spark-connector-hive_2.12-{{ release }}.jar, which can be found in the [Maven Central](https://mvnrepository.com/artifact/org.apache.kyuubi/kyuubi-spark-connector-hive)
```

docs/connector/spark/hive.md
```diff
@@ -0,0 +1,172 @@
...
+
+In order to make the Hive connector packages visible for the runtime classpath of engines, we can use one of these methods:
+
+1. Put the Kyuubi Hive connector packages into `$SPARK_HOME/jars` directly
+2. Set `spark.jars=/path/to/kyuubi-hive-connector`
```
pan3793:
```suggestion
2. Set `spark.jars=/path/to/kyuubi-spark-connector-hive_2.12-<version>.jar`
```

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/8cafd873c11323310d306481368290780a37d5b7

Issue #7668, opened 2026-08-28
[Bug] Executor podNamePrefix may collide when multiple engines for the same user start within the same millisecond


littlexyw:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

On Kubernetes, SparkSQLEngine.generateExecutorPodNamePrefixForK8s builds the executor podNamePrefix as kyuubi-<user>-<epoch-millis>. When two engines for the same user (same resolvedUserName) are launched within the same millisecond in the same namespace — for example, a burst of short-lived engines from the same user, or two Kyuubi servers racing to launch engines for the same USER share level — the two engines get an identical spark.kubernetes.executor.podNamePrefix. Their executor pod names (<prefix>-exec-<id>) then collide inside the namespace, and one of the two apps fails to create executors until the other releases the names.

### Affects Version(s)

master

### Additional context

Reproducible by calling SparkSQLEngine.generateExecutorPodNamePrefixForK8s("same_user") in a tight loop — many calls land on the same millisecond and produce identical prefixes.

### Are you willing to submit PR?

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix.
- [ ] No. I cannot submit a PR at this time.

zhang-arvin:
I will work on this issue.

PR #7669, opened 2026-08-28
[KYUUBI #7668][SPARK] Use UUIDv7 for executor podNamePrefix to avoid same-millisecond collision

littlexyw:

On Kubernetes, `SparkSQLEngine.generateExecutorPodNamePrefixForK8s` built the executor podNamePrefix as `kyuubi-<user>-<epoch-millis>`. When two engines for the same user (same `resolvedUserName`) are launched within the same millisecond in the same namespace — e.g. a burst of short-lived engines from the same user, or two Kyuubi servers racing to launch engines for a `USER` share level — the two engines get an identical `spark.kubernetes.executor.podNamePrefix`. Their executor pod names (`<prefix>-exec-<id>`) then collide inside the namespace, and one of the two apps fails to create executors until the other releases the names.

This PR replaces the epoch-millis suffix with a UUIDv7 from the existing `UuidUtils.generateUUIDv7` helper. UUIDv7 keeps the 48-bit epoch millis in its leading bits so pod names remain roughly time-ordered when listed by name, while the trailing 74 bits of secure randomness make same-millisecond collisions essentially impossible. The oversized-user-name fallback path is switched to the same UUIDv7 for consistency.

- Added a concurrent unit test `generate executor pod name prefix should be unique for concurrent calls with the same user` in `SparkSQLEngineSuite` that fans out 1000 prefix generations across 10 threads for the same user and asserts every generated prefix is unique. Under the pre-fix epoch-millis implementation this test necessarily fails, since many calls land on the same millisecond and produce identical prefixes.
- The existing `[KYUUBI #3385] generate executor pod name prefix with user or UUID` test still passes (sanitization, length cap, and pod-log-directory length invariants unchanged).

Assisted-by: Claude Opus 4.7

### DIFF
--- a/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
+++ b/externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/SparkSQLEngine.scala
@@ -18,7 +18,7 @@
 package org.apache.kyuubi.engine.spark
 
 import java.time.Instant
-import java.util.{Locale, UUID}
+import java.util.Locale
 import java.util.concurrent.{CountDownLatch, ScheduledExecutorService, ThreadPoolExecutor, TimeUnit}
 import java.util.concurrent.atomic.AtomicBoolean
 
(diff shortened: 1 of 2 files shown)


pan3793 (approved):
LGTM. But please fill PR description seriously.

pan3793:
thanks, merged to master/1.12

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/f35e7df6c1197e21c6a1993eb7eff115f9dc8229
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/6ab4207f260916632cb82153198e208166e3f48b

Issue #7644, opened 2026-08-21
Flaky Test: FlinkOperationSuiteOnYarn: execute statement - select column name with dots


wForget:
### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

https://github.com/apache/kyuubi/actions/runs/32186943591/job/95943569442?pr=7637

```
- execute statement - select column name with dots *** FAILED ***
  org.apache.kyuubi.jdbc.hive.KyuubiSQLException: Error operating ExecuteStatement: org.apache.flink.table.api.TableException: Failed to execute sql
	at org.apache.flink.table.api.internal.TableEnvironmentImpl.executeQueryOperation(TableEnvironmentImpl.java:1094)
	at org.apache.flink.table.api.internal.TableEnvironmentImpl.executeInternal(TableEnvironmentImpl.java:1134)
	at org.apache.flink.table.gateway.service.operation.OperationExecutor.executeOperation(OperationExecutor.java:495)
	at org.apache.flink.table.gateway.service.operation.OperationExecutor.executeStatement(OperationExecutor.java:243)
	at org.apache.flink.table.gateway.service.operation.OperationExecutor.executeStatement(OperationExecutor.java:199)
	at org.apache.kyuubi.engine.flink.operation.ExecuteStatement.executeStatement(ExecuteStatement.scala:78)
	at org.apache.kyuubi.engine.flink.operation.ExecuteStatement.runInternal(ExecuteStatement.scala:62)
	at org.apache.kyuubi.operation.AbstractOperation.run(AbstractOperation.scala:197)
	at org.apache.kyuubi.session.AbstractSession.runOperation(AbstractSession.scala:114)
	at org.apache.kyuubi.session.AbstractSession.$anonfun$executeStatement$1(AbstractSession.scala:144)
	at org.apache.kyuubi.session.AbstractSession.withAcquireRelease(AbstractSession.scala:87)
	at org.apache.kyuubi.session.AbstractSession.executeStatement(AbstractSession.scala:141)
	at org.apache.kyuubi.service.AbstractBackendService.executeStatement(AbstractBackendService.scala:66)
	at org.apache.kyuubi.service.TFrontendService.ExecuteStatement(TFrontendService.scala:253)
[...]

PR #7672, opened 2026-08-28
[KYUUBI #7671][AUTHZ] Cover all injected authz rules in the excludedRules check

### LINKED ISSUE #7671: [FEATURE][AUTHZ] AuthzConfigurationChecker should cover every authz rule the extension injects
### Code of Conduct

### Search before asking

### Describe the feature

`AuthzConfigurationChecker` rejects a `spark.sql.optimizer.excludedRules` value that names `org.apache.kyuubi.plugin.spark.authz.ranger`, which was the whole extension when the check was written. Most rules live under `org.apache.kyuubi.plugin.spark.authz.rule` now — 21 files there against 9 in `ranger` — so a rule added there is outside the denylist.

Nothing is reachable through that gap today. `spark.sql.optimizer.excludedRules` filters optimizer batches only, and Spark has no analyzer counterpart, so the resolution rules that apply masking and row filtering cannot be excluded at all. Of the rules `RangerSparkExtension` injects, four are optimizer rules: `RuleAuthorization` in `…authz.ranger`, which the check already covers, and `RuleEliminateMarker`, `RuleEliminatePermanentViewMarker` and `RuleEliminateTypeOf` in `…authz.rule`, which only strip markers after the check has run. Excluding one of those three breaks the query instead of lifting a check: with the extension and the test policies loaded, excluding `RuleEliminateMarker` fails planning with `No plan for RowFilterMarker +- DataMaskingStage0Marker …`, `RuleEliminatePermanentViewMarker` fails the same way, `RuleEliminateTypeOf` fails at runtime, and a user without privileges on the table stays denied with all three excluded.

So this is future-proofing for rules that land under `…authz.rule` rather than closing a live bypass, and it does reject configurations that are accepted today — which is why it is separate from #7623 rather than part of it.

Raised by @wForget in review of #7637; cc @cfmcgrady @pan3793.

### Motivation

The denylist should not depend on where a rule happens to live, or on the accident that excluding the current cleanup rules breaks the query rather than lifting a check.

### Describe the solution

Widen the package the check matches from `org.apache.kyuubi.plugin.spark.authz.ranger` to `org.apache.kyuubi.plugin.spark.authz`, and cover the three `…authz.rule` rules in `AuthzConfigurationCheckerSuite`.

### Additional context

_No response_

alexandrefimov:

Closes #7671, the follow-up @wForget asked for while reviewing #7637.

`AuthzConfigurationChecker` matches `org.apache.kyuubi.plugin.spark.authz.ranger` in the effective `spark.sql.optimizer.excludedRules`, which was the whole extension when the check was written. Most rules live under `org.apache.kyuubi.plugin.spark.authz.rule` now, so a rule added there is outside the denylist. This matches the extension's own package instead.

It is future-proofing rather than a fix for a live bypass, and worth saying why. `spark.sql.optimizer.excludedRules` filters optimizer batches only and Spark has no analyzer counterpart, so the resolution rules that apply masking and row filtering cannot be excluded at all. Of the four optimizer rules `RangerSparkExtension` injects, `RuleAuthorization` is in `…authz.ranger` and already covered; the other three — `RuleEliminateMarker`, `RuleEliminatePermanentViewMarker`, `RuleEliminateTypeOf` — only strip markers after the check has run, and excluding one of them breaks the query instead of lifting a check.

The change does reject configurations that are accepted today, which is why it is separate from #7637 rather than part of it.

A new case in `AuthzConfigurationCheckerSuite` covers the three `…authz.rule` rules by their `ruleName`, through both paths the checker guards: the value in effect, and the `SET` syntax.

30 tests green across `AuthzConfigurationCheckerSuite`, `DataMaskingForInMemoryParquetSuite` and the row filtering suite. With the constant reverted to the ranger package the new case fails, so it tests the change rather than the code around it.

Assisted-by: Claude Opus 5

### DIFF
--- a/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/config/AuthzConfigurationChecker.scala
+++ b/extensions/spark/kyuubi-spark-authz/src/main/scala/org/apache/kyuubi/plugin/spark/authz/rule/config/AuthzConfigurationChecker.scala
@@ -32,7 +32,11 @@ case class AuthzConfigurationChecker(spark: SparkSession) extends (LogicalPlan =
   final val RESTRICT_LIST_KEY = "spark.kyuubi.conf.restricted.list"
   final val EXCLUDED_RULES_KEY = "spark.sql.optimizer.excludedRules"
 
-  final private val AUTHZ_RANGER_RULE_PACKAGE = "org.apache.kyuubi.plugin.spark.authz.ranger"
+  // Every rule the extension injects lives under this package, not only the ranger ones:
+  // the marker-eliminating rules sit in org.apache.kyuubi.plugin.spark.authz.rule, and new
+  // rules land there too. Excluding one of those breaks the query at planning today rather
+  // than lifting a check, but the denylist should not depend on that staying true.
+  final private val AUTHZ_RULE_PACKAGE = "org.apache.kyuubi.plugin.spark.authz"
 
   private val restrictedConfList: Set[String] =
     Set(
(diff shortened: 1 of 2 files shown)


alexandrefimov:
The one red check, `Flink Test (8, 1.20, normal)`, is `FlinkOperationOnYarnSuite: get catalogs` failing with `Time out retrieving Flink engine service url`. This PR only touches `extensions/spark/kyuubi-spark-authz`. It is not #7644 either — that one is fixed by `bf10595ba`, which this branch is based on — and master runs are green, so it looks like a one-off rather than something to file.

Could someone re-run that job? The `(17, 3.11, 3.5, …mirror…)` job was cancelled in the same run and needs the re-run too.

wForget (approved):
Thanks, LGTM

If we accept this PR, it may mean that we can no longer disable the data masking and row filter rules independently. However, I think a better way to support that use case would be to provide an additional `RangerSparkExtension` without the data masking and row filter rules, as discussed here: https://github.com/apache/kyuubi/pull/7082#issuecomment-2969153814.

So I'm fine with the current change.

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/330a6426d8126da8721bd288e7b1257193204149

PR #7677, opened 2026-08-28
[KYUUBI #7676][DOCS] Recommend multi-tenant session config safeguards

### LINKED ISSUE #7676: :memo: Recommend to configure `kyuubi.session.conf.restrict.list` or/and `kyuubi.session.conf.ignore.list` for multi-tenant Kyuubi cluster
### Code of Conduct

### Search before asking

### What type of changes will we make to the documentation?

Bugfixes

### Affects Version(s)

master

### Improving the documentation

`kyuubi.session.conf.restrict.list` or `kyuubi.session.conf.ignore.list` has no default value, therefore any user can overwrite the session conf.

We can provide a reasonable default values or/and improve documents to ask admin to configure them for multi-tenant Kyuubi use-case.

My opinion is to document it instead of updating the default. No default value is useful for development use case and there's no incompatible change for Kyuubi users.

nightcityblade:

The session configuration ignore and restrict lists are empty by default. In a
multi-tenant deployment, leaving both lists empty lets clients override
sensitive session-level configuration during engine bootstrap and connection
setup. This adds an explicit administrator recommendation to the existing
security documentation.

Closes #7676.

- `env PATH=/usr/bin:/bin:/usr/sbin:/sbin dev/reformat`
- `sphinx-build -W --keep-going -D suppress_warnings=myst.xref_missing,misc.highlighting_failure -b html docs /tmp/kyuubi-docs-build-7676-focused`
- Confirmed the generated `security/authorization/spark/overview.html` contains
  the recommendation.

The Sphinx suppression covers existing warnings in unrelated documentation;
the changed page builds without warnings.

Assisted-by: OpenAI Codex:GPT-5

### DIFF
--- a/docs/security/index.md
+++ b/docs/security/index.md
@@ -20,6 +20,29 @@
 Securing Kyuubi involves enabling authentication(authn), authorization(authz)
 and encryption, etc.
 
+## Protect Session Configurations
+
+For a multi-tenant deployment, configure at least one of
+`kyuubi.session.conf.ignore.list` or `kyuubi.session.conf.restrict.list`. Both
+lists are empty by default, so clients can otherwise override sensitive
+session-level configurations during engine bootstrap and connection setup.
+
+One conservative starting point is to keep server-defined resource sizing while
+rejecting client changes to deployment and authorization settings:
+
+```properties
+kyuubi.session.conf.ignore.list=spark.driver.memory,spark.executor.memory
+kyuubi.session.conf.restrict.list=spark.master,spark.submit.deployMode,spark.sql.extensions,spark.sql.optimizer.excludedRules
+```
+
+Tailor these lists to the deployment. The ignore list silently drops matching
+client values, while the restrict list rejects the connection. They do not
+prevent later changes through `SET` statements; configure the engine's
+operation-level restrictions separately when that protection is required.
+
+See the [session configuration settings](../configuration/settings.md#session)
+for details.
+
 ```{toctree}
 :maxdepth: 2
 
@@ -29,4 +52,3 @@ kinit
 hadoop_credentials_manager
 internal_secure_access
 ```
-
--- a/docs/security/index.md
+++ b/docs/security/index.md
@@ -20,34 +20,12 @@
 Securing Kyuubi involves enabling authentication(authn), authorization(authz)
 and encryption, etc.
 
-## Protect Session Configurations
-
-For a multi-tenant deployment, configure at least one of
-`kyuubi.session.conf.ignore.list` or `kyuubi.session.conf.restrict.list`. Both
-lists are empty by default, so clients can otherwise override sensitive
-session-level configurations during engine bootstrap and connection setup.
-
-One conservative starting point is to keep server-defined resource sizing while
-rejecting client changes to deployment and authorization settings:
-
-```properties
-kyuubi.session.conf.ignore.list=spark.driver.memory,spark.executor.memory
-kyuubi.session.conf.restrict.list=spark.master,spark.submit.deployMode,spark.sql.extensions,spark.sql.optimizer.excludedRules
-```
-
-Tailor these lists to the deployment. The ignore list silently drops matching
-client values, while the restrict list rejects the connection. They do not
-prevent later changes through `SET` statements; configure the engine's
-operation-level restrictions separately when that protection is required.
-
-See the [session configuration settings](../configuration/settings.md#session)
-for details.
-
 ```{toctree}
 :maxdepth: 2
 
 Authentication <authentication>
 Authorization <authorization/index>
+Session Configuration <session_configuration>
 kinit
 hadoop_credentials_manager
 internal_secure_access
--- a/docs/security/index.md
+++ b/docs/security/index.md
@@ -30,3 +30,4 @@ kinit
 hadoop_credentials_manager
 internal_secure_access
 ```
+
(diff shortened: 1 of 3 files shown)


aajisaka:
Greeting is somehow broken. Sending a message by hand

>Hello @nightcityblade,
> Thanks for finding the time to submit the Pull request, the Kyuubi community thanks you !
> We really appreciate the community's efforts to improve Apache Kyuubi.
> If you are new to Apache Kyuubi for code changes, please take some time to read our [contributing guide](https://kyuubi.apache.org/contributing.html).

aajisaka:
@nightcityblade Thank you for the PR!
I feel it would be better to document sample parameter list to configure, so that Kyuubi admin can start from there. Do you have any idea?

nightcityblade:
Thanks — I added a concrete starting policy in 44994a2. It silently keeps server-defined driver/executor memory while explicitly rejecting client changes to `spark.master`, `spark.submit.deployMode`, `spark.sql.extensions`, and `spark.sql.optimizer.excludedRules`.

The text also explains when to use the ignore versus restrict list, tells administrators to tailor the baseline, and clarifies that operation-level `SET` protection is configured separately below. The full Sphinx HTML build succeeds (with the same six unrelated existing warnings), and `git diff --check` passes.

wForget (changes requested):
@nightcityblade  Thanks for your contribution. However, the documentation you modified is for the `Kyuubi AuthZ Plugin For Spark SQL`, which is unrelated to the configuration mentioned in this PR.

nightcityblade:
Thanks for catching the scope mismatch. I moved the multi-tenant session-configuration guidance out of the Spark authorization plugin page and into the general Kyuubi security overview. The full Sphinx HTML build succeeds with the same six unrelated existing warnings, and `git diff --check` passes.

docs/security/index.md
```diff
@@ -20,6 +20,29 @@
...
+`kyuubi.session.conf.ignore.list` or `kyuubi.session.conf.restrict.list`. Both
+lists are empty by default, so clients can otherwise override sensitive
+session-level configurations during engine bootstrap and connection setup.
+
+One conservative starting point is to keep server-defined resource sizing while
+rejecting client changes to deployment and authorization settings:
+
+```properties
+kyuubi.session.conf.ignore.list=spark.driver.memory,spark.executor.memory
+kyuubi.session.conf.restrict.list=spark.master,spark.submit.deployMode,spark.sql.extensions,spark.sql.optimizer.excludedRules
```
aajisaka: These parameters are not a good example for security. For example, I would like to add `kyuubi.session.engine.spark.main.resource` to restrict overriding the main class and add `kyuubi.engine.share.level` to restrict overriding the share level, also `spark.submit.deployMode` to ensure client use the expected deploy mode.

Also please mention this example is not a comprehensive list.

docs/security/index.md
```diff
@@ -20,6 +20,29 @@
...
+```
+
+Tailor these lists to the deployment. The ignore list silently drops matching
+client values, while the restrict list rejects the connection. They do not
+prevent later changes through `SET` statements; configure the engine's
+operation-level restrictions separately when that protection is required.
```
aajisaka: > They do not prevent later changes through `SET` statements; configure the engine's operation-level restrictions separately when that protection is required.

This is not always true. If the config is static or used in other module (e.g. yarn/core), it's never updated.

Reference: https://kyuubi.readthedocs.io/en/master/monitor/troubleshooting.html#failed-to-start-spark-3-1-or-above-version-with-error-msg-cannot-modify-the-value-of-a-spark-config

aajisaka (changes requested):

aajisaka:
Also, would you create a new doc under docs/security for this guidance? Adding to index.md is not a good idea. I would like to keep the index.md only for index

nightcityblade:
Addressed the latest feedback in 95c480e:

- moved the guidance into a dedicated `docs/security/session_configuration.md` page and linked it from the security index
- added `kyuubi.session.engine.spark.main.resource` and `kyuubi.engine.share.level` to the restrictive example
- clarified that the example is not comprehensive
- corrected the `SET` discussion to distinguish static/startup-consumed settings from settings that need operation-level restrictions

Validation: the full Sphinx HTML build succeeds with the same six unrelated existing warnings, and `git diff --check` passes.

nightcityblade:
Addressed: the guidance now lives in a dedicated docs/security/session_configuration.md page, includes the requested main-resource, engine-share-level, and deploy-mode safeguards, marks the list as non-comprehensive, and clarifies that only some settings remain mutable via SET.

aajisaka (approved):
LGTM

wForget (approved):
Thanks @nightcityblade

docs/security/index.md
```diff
@@ -25,8 +25,8 @@ and encryption, etc.
 
 Authentication <authentication>
 Authorization <authorization/index>
+Session Configuration <session_configuration>
 kinit
 hadoop_credentials_manager
 internal_secure_access
 ```
-
```
wForget, line 32: Related failure

```
Caused by: org.apache.maven.plugin.MojoExecutionException: The following files had format violations:
    docs/security/index.md
        @@ -30,3 +30,4 @@
         hadoop_credentials_manager
         internal_secure_access
         ```
        +
Run 'mvn spotless:apply' to fix these violations.
```

nightcityblade:
Fixed the Spotless Markdown failure in a861a77 by adding the required trailing blank line to docs/security/index.md. Local build/mvn spotless:check now passes across all 41 reactor modules.

aajisaka (approved):

aajisaka:
Merged to master and branch-1.12. Thank you @nightcityblade @wForget

PR #7683, opened 2026-08-29
[KYUUBI-7659][INFRA] Restore code coverage by not overriding the test argLine

### LINKED ISSUE #7659: [BUG] Code coverage reports are MIA: the test argLine override drops the JaCoCo agent

zhang-arvin:

`-Pcodecov` hasn't produced any coverage since 1.8.0 because both `maven-surefire-plugin` and `scalatest-maven-plugin` override their `argLine` with `${extraJavaTestArgs}`, which drops the JaCoCo agent that `jacoco:prepare-agent` publishes to the `${argLine}` property.

This PR moves the JDK `--add-opens` options into the `argLine` property itself (referencing `extraJavaTestArgs`) and removes the `argLine` overrides from both plugin configs. This allows `jacoco:prepare-agent` to prepend its agent to the property's existing value.

## How was this patch tested?

Manual verification:
```
build/mvn test -Pcodecov -pl kyuubi-util-scala
```
Should produce `target/jacoco.exec` and a coverage report.

Closes #7659

### DIFF
--- a/externals/kyuubi-data-agent-engine/pom.xml
+++ b/externals/kyuubi-data-agent-engine/pom.xml
@@ -187,7 +187,6 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <skipTests>${skipTests}</skipTests>
-                    <argLine>${extraJavaTestArgs}</argLine>
                 </configuration>
             </plugin>
             <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -289,6 +289,10 @@
         <distMgmtSnapshotsName>Apache Development Snapshot Repository</distMgmtSnapshotsName>
         <distMgmtSnapshotsUrl>https://repository.apache.org/content/repositories/snapshots</distMgmtSnapshotsUrl>
 
+        <!-- Both test plugins default their argLine to ${argLine}; jacoco:prepare-agent
+             prepends the agent to it under -Pcodecov. Overriding argLine in plugin config
+             instead breaks that, since plugin config is interpolated before the agent runs. -->
+        <argLine>${extraJavaTestArgs}</argLine>
         <extraJavaTestArgs>-XX:+IgnoreUnrecognizedVMOptions
             --add-opens=java.base/java.lang=ALL-UNNAMED
             --add-opens=java.base/java.lang.invoke=ALL-UNNAMED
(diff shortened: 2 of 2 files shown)


wangzhigang1999:
@zhang-arvin,

I am closing [[PR 7680](https://github.com/apache/kyuubi/pull/7680)](https://github.com/apache/kyuubi/pull/7680) and [[PR 7681](https://github.com/apache/kyuubi/pull/7681)](https://github.com/apache/kyuubi/pull/7681) because the proposed fixes are not suitable in their current form.

[[PR 7682](https://github.com/apache/kyuubi/pull/7682)](https://github.com/apache/kyuubi/pull/7682) duplicates the approved [[PR 7669](https://github.com/apache/kyuubi/pull/7669)](https://github.com/apache/kyuubi/pull/7669), and [[PR 7683](https://github.com/apache/kyuubi/pull/7683)](https://github.com/apache/kyuubi/pull/7683) duplicates the active [[PR 7661](https://github.com/apache/kyuubi/pull/7661)](https://github.com/apache/kyuubi/pull/7661).

Please check for existing work and coordinate with other contributors before opening a PR. Also follow the PR template and include test results.

PR #7685, opened 2026-08-30
[KYUUBI #7684][SPARK] Keep the python operation context private to the operation

### LINKED ISSUE #7684: [Bug] Concurrent operations on one session corrupt kyuubi.session.user in the Spark Python engine
### Code of Conduct

### Search before asking

### Describe the bug

Python operations of one session all run on a single pinned Py4J thread, and Spark local properties
are thread local, so the per-operation context in `ExecutePython.withLocalProperties` is not private
to the operation the way it is for SQL. A concurrent operation of the same session can run after
another operation's teardown has already cleared `kyuubi.session.user`, and Python clears it with
the empty string, which is fatal.

Two defects in `ExecutePython` combine to produce it.

1. The teardown clears with `""`, while `SparkOperation.withLocalProperties` (SQL and Scala) clears
   with `null`. The empty string is the outlier inside its own method: the same `finally` block
   already clears the signing keys with `null` through `clearSessionUserSign`.

```scala
setSparkLocalProperty(KYUUBI_SESSION_USER_KEY, "")
setSparkLocalProperty(KYUUBI_STATEMENT_ID_KEY, "")
setSparkLocalProperty(SPARK_SCHEDULER_POOL_KEY, "")
```

2. `SessionPythonWorker.runCode` takes the worker lock per call, so the 7 to 12 internal round-trips
   that apply and clear the context are not atomic. The lock serializes round-trips, not operations.

```scala
def runCode(code: String, internal: Boolean = false): Option[PythonResponse] = withLockRequired {
```

```
Z: runCode(user code)                              <- holds the lock, long running
A: setJobGroup                                     <- blocks, A stays PENDING
Z: (releases lock)
A: setJobGroup
A: setLocalProperty('kyuubi.session.user', 'alice')
Z: setLocalProperty('kyuubi.session.user', '')     <- Z's teardown
A: setLocalProperty('kyuubi.statement.id', ...)
A: runCode(user code)                              <- runs with an empty user
```

An empty user is fatal because `getAuthzUgi` guards only against `null`:

```scala
val user = spark.getLocalProperty(KYUUBI_SESSION_USER)
if (isSessionUserVerifyEnabled) {
  verifyKyuubiSessionUser(spark, user)
}
if (user != null && user != UserGroupInformation.getCurrentUser.getShortUserName) {
  UserGroupInformation.createRemoteUser(user)
} else {
  UserGroupInformation.getCurrentUser
}
```

`""` passes `user != null` and reaches `createRemoteUser`, which rejects it with
`IllegalArgumentException: Null user`, verified on `hadoop-client-api` 3.3.4 and 3.4.3. With
`kyuubi.session.user.sign.enabled` set to true, `verifyKyuubiSessionUser` rejects it first with
`AccessControlException: Invalid user identifier []`. Both defects are needed for the `Null user`
variant. With signing enabled the missing atomicity alone is enough, since `StringUtils.isAnyBlank`
rejects null too.

Two more consequences of the same non-atomic context. An operation waiting on the worker lock is
blocked inside its own setup, before the `setState(OperationState.RUNNING)` that sits in the block
`withLocalProperties` evaluates as `f`, so it stays in `PENDING_STATE` for the whole wait. And job
groups collide: if A's `setJobGroup(A)` lands after Z's, Z's jobs are tagged with group A and A's
`cancelJobGroup(A)` cancels Z's jobs.

`ExecutePython` does not override `getNextRowSetInternal`, so a fetch runs another full apply and
clear cycle through the worker, which widens the window for clients that pipeline execute and fetch
calls. One worker serves one session, so the corrupted value is the session user or empty, never a
different user.

### Affects Version(s)

1.7.0 and later, including master. The empty-string clear arrived with KYUUBI #3839.

### Kyuubi Server Log Output

```logtalk
INFO  o.a.k.operation.ExecuteStatement: Query[<operation-id>] in PENDING_STATE
        (repeated every poll, about 11s apart, for roughly 42 minutes)
INFO  o.a.k.operation.ExecuteStatement: Processing user_a's query[<operation-id>]:
        RUNNING_STATE -> ERROR_STATE, time taken: 2560.826 seconds
ERROR o.a.k.operation.ExecuteStatement: Error operating ExecuteStatement:
        org.apache.kyuubi.KyuubiSQLException: Error operating ExecutePython:
        org.apache.kyuubi.KyuubiSQLException: Interpret error
```

About 2525 of those 2560 seconds were spent in `PENDING_STATE` behind another Python statement on
the same session. A long PENDING followed by a short `RUNNING -> ERROR` carrying `Null user` is the
signature, and it is checkable in existing logs.

### Kyuubi Engine Log Output

```logtalk
org.apache.kyuubi.KyuubiSQLException: Interpret error:
{"response":{"content":{"ename":"IllegalArgumentException","evalue":"Null user",
 "traceback":["  File \"<stdin>\", line N, in <module>\n",
              "  File \"/opt/spark/python/pyspark/sql/session.py\", line 1631, in sql\n",
              "pyspark.errors.exceptions.captured.IllegalArgumentException: Null user\n"],
 "status":"error"}}}
    at o.a.k.engine.spark.operation.ExecutePython.withLocalProperties(ExecutePython.scala:166)
    at o.a.k.engine.spark.operation.ExecutePython.executePython(ExecutePython.scala:85)
    at o.a.k.engine.spark.operation.ExecutePython$$anon$1.run(ExecutePython.scala:116)
```

Line 166 is the `f` call site, reached after this operation's own setup had already set
`kyuubi.session.user` correctly. It was set and then overwritten.

```yaml
kyuubi.operation.language: PYTHON
kyuubi.session.user.sign.enabled: false   # true gives the AccessControlException variant instead
# the failure needs the authz plugin, the only reader of kyuubi.session.user that rejects it
spark.sql.extensions: org.apache.kyuubi.plugin.spark.authz.ranger.RangerSparkExtension
```

### Additional context

The empty-user rejection reproduces deterministically with no concurrency: set
`kyuubi.session.user` to `""` and call `AuthZUtils.getAuthzUgi`, which is the state
$$

moelhoussein:
Fixes #7684.

Python operations of one session all run on a single pinned Py4J thread, and Spark local properties are thread local, so the per-operation context in `ExecutePython.withLocalProperties` is not private to the operation the way it is for SQL. Two problems follow from that.

The teardown clears `kyuubi.session.user` with the empty string, while `SparkOperation.withLocalProperties` clears with `null`. `AuthZUtils.getAuthzUgi` skips the property only when it is null, so an empty user reaches `UserGroupInformation.createRemoteUser` and fails with `IllegalArgumentException: Null user`. With `kyuubi.session.user.sign.enabled` it fails earlier, as `AccessControlException: Invalid user identifier []`. The empty string was already the outlier in its own method, since the same `finally` block clears the signing keys with `null` through `clearSessionUserSign`. Null is not what saves the signed path, where `verifyKyuubiSessionUser` treats null as blank and reports `Invalid user identifier [null]` just the same. There the lock is what keeps a concurrent operation from reading a cleared context.

`SessionPythonWorker.runCode` takes the worker lock per call, so the 7 to 12 internal round-trips that apply and clear the context are not atomic. Another operation of the same session can set or clear these properties in between, which loses the session user and also lets one operation's `setJobGroup` tag another operation's jobs, so its `cancelJobGroup` cancels them.

This patch clears with `null` and holds the worker lock across the whole body of `withLocalProperties`, which substitutes for the isolation the SQL path gets from running each operation on its own thread. Throughput should be unaffected, because `runCode(statement)` already holds the lock for the whole execution and same-session operations are already serialized behind it. Cancellation is unaffected, because `ExecutePython.cleanup` interrupts the worker with a signal and never takes this lock.

Two new tests in `ExecutePythonSuite`, both of which fail on master. The first runs an operation and then probes the worker thread with a bare `runCode`, which observes exactly what the teardown left behind. Master leaves `''`, so it fails with `"['']" did not equal "[None]"`. The second holds an operation's context and checks that a second thread cannot reach the worker. On master it can, so it fails with `otherOperationRanCode.get() was true`.

Both tests start the python worker inside the test JVM, where its PySpark reaches this JVM's Spark classes over Py4J, so they cancel when the PySpark under `SPARK_HOME` does not match the engine's Spark. That is the case in the jobs that verify a Spark 3.5 build on a Spark 4.x binary, which carry no 3.5 PySpark to point the worker at. `PySparkTests` still covers python in those jobs, since it drives an engine from `SPARK_HOME` and both sides match there.

Local results are `ExecutePythonSuite` 2 of 2 and `PySparkTests` 7 of 8. The single failure is `executePython support timeout`, which also fails on unmodified master in the same environment (one pass in six runs on master, zero in six on this branch). It fails in the second half of that test, where `bad_code` after a timeout returns no error, which points at response handling on the worker stream after the interrupt rather than at the context.

Assisted-by: Claude Opus 5


externals/kyuubi-spark-sql-engine/src/main/scala/org/apache/kyuubi/engine/spark/operation/ExecutePython.scala
```diff
@@ -143,35 +143,44 @@ class ExecutePython(
...
-      setSparkLocalProperty(SPARK_SCHEDULER_POOL_KEY, "")
-      // using cancelJobGroup for pyspark, see details in pyspark/context.py
-      worker.runCode(s"spark.sparkContext.cancelJobGroup('$statementId')", internal = true)
-      if (isSessionUserSignEnabled) {
-        clearSessionUserSign()
+        f
+      } finally {
+        // Clear with null, not the empty string. AuthZUtils.getAuthzUgi only skips the property
+        // when it is null, so an empty user reaches UserGroupInformation.createRemoteUser and
+        // fails with `Null user`, and fails session user signing as a blank identifier.
```
wangzhigang1999: This is only true when session-user signing is disabled. When signing is enabled, verifyKyuubiSessionUser runs first and rejects both null and empty values. Could you clarify that the outer lock, rather than the null value, protects the signing-enabled path?

extensions/spark/kyuubi-spark-authz/src/test/scala/org/apache/kyuubi/plugin/spark/authz/ranger/AuthzSessionUserSuite.scala
wangzhigang1999, line 1: This test verifies existing `AuthZUtils` behavior, but it does not exercise the `ExecutePython` change and still passes when the patch is reverted. Since `ExecutePythonSuite` already verifies that the property is cleared with null, could we remove this suite or fold the relevant assertion into that regression test?

wangzhigang1999:
Both failures in this job come from the newly added `ExecutePythonSuite`, but the log shows that the Python worker exits before either test body runs.

This CI job compiles the embedded test JVM against Spark 3.5.8 while `SPARK_HOME` and PySpark point to Spark 4.1.2. The suite creates a `SessionPythonWorker` directly inside that embedded JVM, so PySpark 4.1.2 tries to initialize a `SparkSession` against Spark 3.5 classes and exits with:

```
TypeError: 'JavaPackage' object is not callable
```

The later `setLocalProperty(..., None) failure: None` is a secondary cleanup failure after the worker has exited. It does not show that clearing the property with `None` is invalid.

Could you update the tests to work with the existing cross-binary CI matrix, either by using matching Spark/PySpark versions for the embedded worker or by testing through the actual Spark binary, and rerun the checks?

moelhoussein:
Thanks, the diagnosis is right on all three points.

**Cross-binary jobs.** The worker imports PySpark from `SPARK_HOME`, since `execute_python.py` puts `$SPARK_HOME/python` on `sys.path`, so in those jobs PySpark 4.x initializes a `SparkSession` against the Spark 3.5 classes held by the test JVM and exits at import. The `setLocalProperty(..., None) failure: None` is the `Option` from `stdout.readLine()` once the worker is gone, as you said, not a verdict on clearing with `None`.

`withSessionAndWorker` now reads the PySpark version under `SPARK_HOME`, resolved the way `createSessionPythonWorker` resolves it, and cancels when it does not match `SPARK_ENGINE_RUNTIME_VERSION`. Locally that is 2 canceled with a 4.1.2 `SPARK_HOME` and 2 of 2 with a matching one.

On the two options you offered: those jobs unpack only the 4.x distribution, so there is no 3.5 PySpark on disk to point the worker at without adding a second Spark download to the matrix. Testing through the Spark binary does not reach either behavior, because the cleared context is only observable between operations on the pinned thread, and every statement over JDBC re-applies the context before user code runs, while racing two operations to catch the non-atomic teardown would be flaky. `PySparkTests` keeps python covered in those jobs, since its engine comes from `SPARK_HOME`, so both sides match there.

**Signing.** Agreed. `verifyKyuubiSessionUser` runs before the null check and `isAnyBlank` counts null as blank, so it rejects with `Invalid user identifier [null]`. The comment and the description now say that null only fixes the unsigned path and the lock is what protects the signed one.

**AuthzSessionUserSuite.** Removed. Folding the assertion into `ExecutePythonSuite` is not possible, because `kyuubi-spark-sql-engine` does not depend on `kyuubi-spark-authz`, so the reason now lives in the `ExecutePython` comment instead.

Checks are rerunning on the new commit.

moelhoussein:
Thanks, the diagnosis is right on all three points.

**Cross-binary jobs.** The worker imports PySpark from `SPARK_HOME`, since `execute_python.py` puts `$SPARK_HOME/python` on `sys.path`, so in those jobs PySpark 4.x initializes a `SparkSession` against the Spark 3.5 classes held by the test JVM and exits at import. The `setLocalProperty(..., None) failure: None` is the `Option` from `stdout.readLine()` once the worker is gone, as you said, not a verdict on clearing with `None`.

`withSessionAndWorker` now reads the PySpark version under `SPARK_HOME`, resolved the way `createSessionPythonWorker` resolves it, and cancels when it does not match `SPARK_ENGINE_RUNTIME_VERSION`. Locally that is 2 canceled with a 4.1.2 `SPARK_HOME` and 2 of 2 with a matching one.

On the two options you offered: those jobs unpack only the 4.x distribution, so there is no 3.5 PySpark on disk to point the worker at without adding a second Spark download to the matrix. Testing through the Spark binary does not reach either behavior, because the cleared context is only observable between operations on the pinned thread, and every statement over JDBC re-applies the context before user code runs, while racing two operations to catch the non-atomic teardown would be flaky. `PySparkTests` keeps python covered in those jobs, since its engine comes from `SPARK_HOME`, so both sides match there.

**Signing.** Agreed. `verifyKyuubiSessionUser` runs before the null check and `isAnyBlank` counts null as blank, so it rejects with `Invalid user identifier [null]`. The comment and the description now say that null only fixes the unsigned path and the lock is what protects the signed one.

**AuthzSessionUserSuite.** Removed. Folding the assertion into `ExecutePythonSuite` is not possible, because `kyuubi-spark-sql-engine` does not depend on `kyuubi-spark-authz`, so the reason now lives in the `ExecutePython` comment instead.

Checks are rerunning on the new commit.

moelhoussein:
Small correction to my last note: the five workflow runs for `da9e7d9e2` are not running yet, they are queued in `action_required` and need a maintainer to approve them before they start.

moelhoussein:
CI on `da9e7d9e2`: the three `verify-on-spark-4.x-binary` jobs pass now, with both new tests cancelled there, and both run and pass in the four matched jobs (3.5, 4.0, 4.1, 4.2).

The one red job is unrelated to this change. `BatchesResourceSuite` fails at line 251 with `"CANCELED" equaled "CANCELED"`, which is the `else` branch of `if (closeBatchResponse.isSuccess)` asserting that the batch is not `CANCELED`. The close reported failure while the state had already moved to `CANCELED`, and the log shows `kill: (23957): No such process` just before it, so the kill had already landed. That job passed on the previous commit, and this round only added the test gate, reworded a comment and deleted the authz suite, none of which reach the batch REST path.

I do not have rights to re-run it. Could you kick that job when you get a chance?

wangzhigang1999 (approved):
LGTM，Thanks！

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/084a14f9d4c0f0a5e017753143ca96514058cba2
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/7dcc5d0c21731ddff2b67466f10c28fe9cbb03e9
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/f67b3d91ea4de1ab1559a2f17ec13b8543dc3402
- merged into branch-1.10 https://github.com/apache/kyuubi/commit/6f43a5aad5cc5b5e974ef0e29fb457dd3ff86274

PR #7687, opened 2026-08-30
[KYUUBI #7686][UTIL] Treat strongly-encapsulated members as candidate misses in hiddenImpl

### LINKED ISSUE #7686: DynMethods, DynFields and DynConstructors abort hiddenImpl fallback chains on JDK 16+
### Description

The three `hiddenImpl` builders in `kyuubi-util`'s vendored `reflect` package call `setAccessible` on the member they just looked up:

```java
try {
  Method hidden = targetClass.getDeclaredMethod(methodName, argClasses);
  AccessController.doPrivileged(new MakeAccessible(hidden));
  this.method = new UnboundMethod(hidden, name);
} catch (SecurityException | NoSuchMethodException e) {
  // unusable or not the right implementation
}
```

`setAccessible` on a member of a package that is not open throws `java.lang.reflect.InaccessibleObjectException` — since JDK 9 for named modules, and by default from the unnamed module since JDK 16. It is a `RuntimeException`, and none of the three catch clauses match it, so it escapes the builder. Every candidate registered after the inaccessible one is never tried, and the public-lookup `.impl(...)` fallback that usually follows never runs.

The shape that breaks is the one `ReflectUtils` uses throughout the engines:

```scala
// kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
val field = DynFields.builder
  .hiddenImpl(clz, fieldName)
  .impl(clz, fieldName)      // never reached when hiddenImpl throws
```

Reproducer on JDK 16+ without `--add-opens`:

```java
DynFields.builder()
    .hiddenImpl("java.math.BigInteger", "signum")
    .impl(String.class, "value")
    .build();

// java.lang.reflect.InaccessibleObjectException: Unable to make field final int
//   java.math.BigInteger.signum accessible: module java.base does not "opens java.math"
//   to unnamed module @1b069fdf
```

Expected: the inaccessible candidate counts as a miss and the chain continues to the next implementation.

### Affected code

- `DynMethods.Builder.hiddenImpl(Class, String, Class...)`
- `DynFields.Builder.hiddenImpl(Class, String)`
- `DynConstructors.Builder.hiddenImpl(Class, Class...)`

### Upstream

Neither upstream copy catches it either: parquet-common's `DynMethods`/`DynConstructors` and iceberg-common's `DynMethods`/`DynFields`/`DynConstructors` all catch only `SecurityException` plus the `NoSuchMethod`/`NoSuchField` variant, on current master as well as on the releases these files were copied from. So this is a divergence Kyuubi has to carry rather than something a newer upstream release would bring in.

LuciferYang:

Closes #7686.

`setAccessible` on a member of a package that is not open throws `java.lang.reflect.InaccessibleObjectException`. That happens since JDK 9 for named modules, and by default from the unnamed module since JDK 16. It is a `RuntimeException` that none of the three `hiddenImpl` builders caught, so one inaccessible candidate escaped the builder and aborted the whole lookup chain instead of falling through to the next implementation. That breaks the `hiddenImpl(...).impl(...)` pattern `ReflectUtils` uses throughout the engines, where the public-lookup fallback exists for exactly the case where the hidden lookup does not work.

The exception type does not exist on Java 8, which this module compiles against, so the catch matches it by class name and rethrows anything else unchanged.

`DynFields` also records the exception with the candidate, so the `module ... does not "opens ..."` text that names the package to open survives into the `Cannot find field from candidates: ...` failure. Single-candidate call sites have no fallback chain to save and would otherwise be left with a bare cannot-find error that reads like a missing field; `ExecutePython`'s `ProcessImpl.pid` lookup is one of them. `DynMethods.Builder` has no per-candidate structure to record into, which its javadoc now states. Aggregating diagnostics there is left to a follow-up.

Neither upstream copy catches this exception either. parquet-common and iceberg-common both catch only `SecurityException` plus the `NoSuchMethod`/`NoSuchField` variant, on the releases these files were copied from and on current master, so this is a divergence Kyuubi carries deliberately and each `hiddenImpl` javadoc records it.

New tests in all three test classes, each branching on the running JDK because the behavior is version-dependent.

`testHiddenImplHandlesStronglyEncapsulatedMethods`, `testHiddenImplTreatsStronglyEncapsulatedFieldsAsMisses` and `testHiddenImplHandlesStronglyEncapsulatedConstructors`: on JDK 16+ the hidden lookup of a `java.math` member must fail as a candidate miss instead of letting the exception escape; on JDK 8 through 15 the same lookup still succeeds. `DynConstructorsTest` also asserts the `InaccessibleObjectException` is among the suppressed problems, and `DynFieldsTest` asserts the recorded candidate carries it.

`testHiddenImplPropagatesUnrelatedFailures` in all three: a `RuntimeException` that is not `InaccessibleObjectException` still escapes the builder. These pass before and after the change. They are here to stop the new catch from being widened into a catch-all, not to cover the fix.

`java.math` is the probe package because the surefire JVM opens `java.lang` but leaves `java.math` closed.

Ran on Zulu 17.0.18: `build/mvn -o test -pl kyuubi-util -am -DwildcardSuites=none` gives 39/39 green, and `build/mvn -o spotless:check -pl kyuubi-util` is clean. Reverting the `DynFields` candidate recording turns the `InaccessibleObjectException` assertion in `DynFieldsTest` red, so that assertion fails if the recording goes away. The JDK 8 leg of these tests runs in CI's `scala-test` job.

Assisted-by: Claude Opus 5

### DIFF
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynConstructors.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynConstructors.java
@@ -32,6 +32,11 @@ public class DynConstructors {
 
   private DynConstructors() {}
 
+  // The type is absent from Java 8, which this module compiles against, so it is matched
+  // by name where it is caught.
+  private static final String INACCESSIBLE_OBJECT_EXCEPTION =
+      "java.lang.reflect.InaccessibleObjectException";
+
   public static class Ctor<C> extends DynMethods.UnboundMethod {
     private final Constructor<C> ctor;
     private final Class<? extends C> constructed;
@@ -211,6 +216,18 @@ public class DynConstructors {
       return this;
     }
 
+    /**
+     * Checks for a hidden implementation.
+     *
+     * <p>Neither upstream copy catches InaccessibleObjectException from {@code setAccessible}, so
+     * under strong encapsulation one inaccessible constructor aborts the whole fallback chain,
+     * while this copy counts it as a miss like the other lookup failures.
+     *
+     * @param targetClass a class instance
+     * @param types argument classes for the constructor
+     * @return this Builder for method chaining
+     * @see Class#getDeclaredConstructor(Class[])
+     */
     public <T> Builder hiddenImpl(Class<T> targetClass, Class<?>... types) {
       // don't do any work if an implementation has been found
       if (ctor != null) {
@@ -226,6 +243,15 @@ public class DynConstructors {
         problems.put(methodName(targetClass, types), e);
       } catch (NoSuchMethodException e) {
         // not the right implementation
+        problems.put(methodName(targetClass, types), e);
+      } catch (RuntimeException e) {
+        // setAccessible on a member of a package that is not open reports
+        // InaccessibleObjectException: since JDK 9 for named modules, since JDK 16 by
+        // default from the unnamed module; count it as a candidate miss like the
+        // failures above instead of letting it escape the fallback chain.
+        if (!INACCESSIBLE_OBJECT_EXCEPTION.equals(e.getClass().getName())) {
+          throw e;
+        }
         problems.put(methodName(targetClass, types), e);
       }
       return this;
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynFields.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynFields.java
@@ -31,6 +31,11 @@ public class DynFields {
 
   private DynFields() {}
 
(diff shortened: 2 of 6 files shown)


pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/6ac41a4c7ea281c1a2a8bbb26b64a82a829ca9dc
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/c58bfad6363a4ae90135ae14099891b2ff7838e5

LuciferYang:
Thank you @pan3793

PR #7689, opened 2026-08-30
[KYUUBI #7688][UTIL] Resolve ctorImpl classes through the builder's configured loader

### LINKED ISSUE #7688: DynMethods.Builder.ctorImpl ignores the builder's configured ClassLoader
### Description

`DynMethods.Builder` lets a caller choose the loader used for name-based lookups:

```java
public Builder loader(ClassLoader newLoader) {
  this.loader = newLoader;
  return this;
}
```

`impl(String className, ...)` honours it:

```java
Class<?> targetClass = Class.forName(className, true, loader);
```

`ctorImpl(String className, ...)` does not. It builds a fresh inner `DynConstructors.Builder`, whose own `loader` field defaults to `Thread.currentThread().getContextClassLoader()`, and never forwards the one configured on the outer builder:

```java
this.method = new DynConstructors.Builder().impl(className, argClasses).buildChecked();
```

So `.loader(x).ctorImpl("some.Class")` resolves `some.Class` through the thread context loader rather than through `x`. When the context loader cannot see the class, `DynConstructors.Builder.impl(String, ...)` swallows the `ClassNotFoundException` as a candidate miss, and the outer `build()` / `buildChecked()` then fails with a generic cannot-find error. The configured loader is ignored without any diagnostic saying so.

This matters in Kyuubi because engine code reflects into classes that are only visible to a particular loader (a Spark or Flink user classloader, an isolated engine classloader), and `.loader(...)` is the documented way to name that loader.

Reproducer:

```java
ClassLoader appLoader = MyClass.class.getClassLoader();
ClassLoader blinded = new URLClassLoader(new URL[0], null);
Thread.currentThread().setContextClassLoader(blinded);

DynMethods.builder("MyClass")
    .loader(appLoader)                  // ignored
    .ctorImpl(MyClass.class.getName())
    .buildChecked();

// java.lang.NoSuchMethodException: Cannot find method: MyClass
```

Expected: the constructor is found, because `appLoader` can see `MyClass`.

### Affected code

`DynMethods.Builder.ctorImpl(String, Class...)`. The `ctorImpl(Class, Class...)` overload takes an already-loaded class and its inner lookup only calls `Class.getConstructor`, so it never reads a loader and is unaffected.

### Upstream

Neither upstream copy forwards the loader either: parquet-common has the same shape on 1.12.x and on current master, and iceberg-common had it on 0.12.1 before removing `ctorImpl` altogether. Note that `ctorImpl` currently has no callers inside Kyuubi, so an alternative to fixing it is to follow iceberg and drop both overloads; that is a maintainer call.

LuciferYang:

Closes #7688.

`ctorImpl(String, ...)` built its inner `DynConstructors.Builder` without forwarding the loader configured on the outer builder, so the name resolved through the thread context loader no matter what the caller passed to `.loader(...)`. `impl(String, ...)` has always honoured it. When the context loader cannot see the class, the `ClassNotFoundException` is swallowed as a candidate miss, so the failure reads as a generic cannot-find with nothing pointing at the ignored loader.

The fix also changes when the default loader is captured: it is now the one read at builder construction, matching `impl(String, ...)`, rather than the context loader read at `ctorImpl` call time. A caller that mutates the context loader in between gets the earlier snapshot.

`ctorImpl` has no caller in Kyuubi outside the new test, so this is hygiene on a retained upstream API rather than a field bug. iceberg-common removed the method in 1.7.0 (apache/iceberg#10818); dropping both overloads instead of fixing them is a maintainer call rather than mine.

`testCtorImplUsesTheConfiguredLoader` in `DynMethodsTest`, in two halves that fail in opposite directions without the fix. With the context loader blinded (a `URLClassLoader` with no URLs and a null parent) and the builder given the app loader, the lookup must succeed; reverting the fix makes `buildChecked()` throw. Swapped, the lookup must fail, which is what pins the absence of a silent fallback; reverting makes `assertThrows` see nothing.

```
build/mvn test -pl kyuubi-util -am -DwildcardSuites=none
```

45 of 45 green on Zulu 17.0.18, spotless clean.

Assisted-by: Claude Opus 5


kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynMethods.java
```diff
@@ -343,14 +343,29 @@ public Builder ctorImpl(Class<?> targetClass, Class<?>... argClasses) {
       return this;
     }
 
+    /**
+     * Checks for a constructor implementation, first finding the given class by name using this
+     * builder's {@link ClassLoader}.
+     *
+     * <p>Neither upstream copy forwarded the loader to the inner constructor builder, so the name
+     * resolved through the thread context loader regardless of what the caller configured.
+     * parquet-common still resolves it that way; iceberg removed {@code ctorImpl} altogether in
```
pan3793: > parquet-common still resolves it that way; 

state might change over time, and these words will go stale silently, let's drop it, and the iceberg change link is sufficient for us to track the upstream change.

pan3793 (approved):

pan3793:
PR description is too long for such a change, can you make it shorter?

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/9cd300b87919e65799910244ebb666a3555be53f
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/f93d229ce53436d482cabe0212b60e69f234786f

kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynMethods.java
LuciferYang: Dropped, the iceberg link stays. Same reason applies to a clause in #7706's class javadoc that claimed this copy reports candidates unlike either upstream, so I removed that one too.

LuciferYang:
Trimmed to about half the length: the loader bug, the change in when the default loader is captured, and that `ctorImpl` has no caller here so deleting both overloads is an option. Dropped the `orNoop()` flip that the capture change can cause and the note about the generic `Cannot find method` message, which #7706 covers.

PR #7691, opened 2026-08-30
[KYUUBI #7690] Fix DynFields javadoc copied from the method variants

### LINKED ISSUE #7690: DynFields javadoc describes fields as methods
### Description

`kyuubi-util`'s `DynFields` is a vendored copy of iceberg-common's class, and its javadoc was carried over from the `DynMethods` variants without being adapted. Fourteen statements are wrong: most describe a field API in terms of methods, and two name an exception the method cannot throw.

`bind` is documented in terms of a type that does not exist in this file:

```java
    /**
     * Returns this method as a BoundMethod for the given receiver.
     *
     * @param target an Object on which to get or set this field
     * @return a {@link BoundField} for this field and the target
     * @throws IllegalStateException if the method is static
     * @throws IllegalArgumentException if the receiver's class is incompatible
     */
    public BoundField<T> bind(Object target) {
```

There is no `BoundMethod` in `DynFields`, and the condition is `isStatic()` on the wrapped `Field`, not on a method.

`build()` promises an exception it cannot throw:

```java
    /**
     * Returns the first valid implementation as a UnboundField or throws a NoSuchFieldException if
     * there is none.
     *
     * @param <T> Java class stored in the field
     * @return a {@link UnboundField} with a valid implementation
     * @throws RuntimeException if no implementation was found
     */
    @SuppressWarnings("unchecked")
    public <T> UnboundField<T> build() {
```

It has no `throws` clause and raises `RuntimeException`, so `NoSuchFieldException` is impossible there, and the `@throws RuntimeException` tag in the same comment already said so. `buildChecked(Object)` has the mirror-image problem: it was documented as throwing `NoSuchMethodException` when its signature declares `throws NoSuchFieldException`.

The full set: the `UnboundField` class doc ("invoke the wrapped method"), `bind`, `asStatic`, `buildChecked(Object)`, `build()`, `build(Object)`, `buildStaticChecked()`, `buildStatic()`, and three `@param fieldName` tags carrying `(different from constructor)`, a parenthetical that separates methods from constructors and has nothing to separate on a field.

The class doc has a second problem. It claims "all Exceptions wrapped by RuntimeException, or with a single Exception catch block", but `UnboundField.get` and `set` catch only `IllegalAccessException`. An `IllegalArgumentException` for an incompatible target, a `NullPointerException` for a null target on an instance field, or an `ExceptionInInitializerError` through `StaticField` all propagate unwrapped, and there is no checked-throwing accessor here for the "single Exception catch block" to refer to.

Both kinds of error cost a reader time rather than breaking anything at runtime. `BoundMethod` cannot be resolved at all, and the documented checked exceptions are ones the compiler will not let a caller catch, so the failure path the javadoc describes turns out to be unwritable. The wrapping claim is the one that can produce a real bug: code written as `catch (RuntimeException e) { log(e.getCause().getMessage()); }` on the strength of it throws a `NullPointerException` of its own when the actual failure was an unwrapped `IllegalArgumentException` with no cause.

### Upstream

Upstream iceberg-common has the same wrong docs, so this is not something a newer upstream release would bring in. Thirteen of the fourteen lines have a direct upstream counterpart and are worth sending to iceberg separately. The remaining one is `buildStaticChecked()`'s `@throws`, and that method exists only in Kyuubi's copy because iceberg removed it along with `ctorImpl`.

LuciferYang:

Closes #7690.

`DynFields` is vendored from iceberg-common and its javadoc came across with the `DynMethods` wording still in it, so a field API ends up documented in method vocabulary. `bind` said it returns a `BoundMethod`, a type this file does not contain. Six `@throws` tags said "if the method is static" or "if the method is not static" for a check that reads `Modifier.isStatic` on a `Field`. Three `@param fieldName` tags carried `(different from constructor)`, which separates a method from a constructor and has nothing to separate on a field.

Two of the fourteen corrections change an exception type. `build()` was documented as throwing `NoSuchFieldException` when it has no `throws` clause and raises `RuntimeException`; the `@throws RuntimeException` tag in the same comment already contradicted that summary. `buildChecked(Object)` was documented as returning a `BoundMethod` and throwing `NoSuchMethodException` while its signature declares `throws NoSuchFieldException`. In both cases the documented exception is one the compiler will not let a caller catch, so the failure path the javadoc described could not be written at all.

The `UnboundField` class doc claimed "all Exceptions wrapped by RuntimeException, or with a single Exception catch block". `get` and `set` catch only `IllegalAccessException`, and this class has no checked-throwing accessor for the second clause to refer to; that clause describes `DynMethods.UnboundMethod.invokeChecked`. The class doc now names the one throwable that is wrapped and says the rest are not.

Upstream iceberg-common carries the same wrong text, so this is not local drift that a resync would repair. Thirteen of the fourteen lines have a direct upstream counterpart and are worth sending to iceberg separately.

Left for a follow-up: the builder's `@throws` tags are written unconditionally and stop holding once `defaultAlwaysNull()` is set, because the sentinel is returned instead of an exception and its `bind` override skips the static and receiver checks. Both `hiddenImpl` overloads also point `@see` at `Class#getField(String)` while the code calls `getDeclaredField`. `DynMethods` has both defects too, so the two vendored copies should change together.

Every changed line sits inside a `/** */` block, so there is no behaviour to test, and a javadoc assertion would pass with the patch reverted.

`build/mvn -o test -pl kyuubi-util -am -DwildcardSuites=none` gives 33/33 green. `build/mvn -o spotless:check -pl kyuubi-util` is clean.

Existing tests in `DynFieldsTest` already cover the behaviour these sentences describe: `testBuildWithoutAlwaysNullFallbackThrows` for `build()` raising `RuntimeException`, `testBindRejectsStaticField` and `testBindRejectsIncompatibleTarget` for `bind`'s two conditions, and `testAlwaysNullBindsThroughEveryEntryPoint` for the sentinel path behind the deferred wording. Two of the behavioural claims have no test and were checked by reading the code: the class doc's wrapping sentence, and `asStatic`'s "if the field is not static" path.

Assisted-by: Claude Opus 5

### DIFF
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynFields.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/reflect/DynFields.java
@@ -34,8 +34,9 @@ public class DynFields {
   /**
    * Convenience wrapper class around {@link Field}.
    *
-   * <p>Allows callers to invoke the wrapped method with all Exceptions wrapped by RuntimeException,
-   * or with a single Exception catch block.
+   * <p>Allows callers to access the wrapped field with {@code IllegalAccessException} wrapped by
+   * RuntimeException. Other throwables {@link Field} raises, such as the {@code
+   * IllegalArgumentException} for an incompatible target, are not wrapped.
    */
   public static class UnboundField<T> {
     private final Field field;
@@ -70,11 +71,11 @@ public class DynFields {
     }
 
     /**
-     * Returns this method as a BoundMethod for the given receiver.
+     * Returns this field as a BoundField for the given receiver.
      *
      * @param target an Object on which to get or set this field
      * @return a {@link BoundField} for this field and the target
-     * @throws IllegalStateException if the method is static
+     * @throws IllegalStateException if the field is static
      * @throws IllegalArgumentException if the receiver's class is incompatible
      */
     public BoundField<T> bind(Object target) {
@@ -94,7 +95,7 @@ public class DynFields {
      * Returns this field as a StaticField.
      *
      * @return a {@link StaticField} for this field
-     * @throws IllegalStateException if the method is not static
+     * @throws IllegalStateException if the field is not static
      */
     public StaticField<T> asStatic() {
       if (!isStatic()) {
@@ -262,7 +263,7 @@ public class DynFields {
      * Checks for an implementation.
      *
      * @param targetClass a class instance
-     * @param fieldName name of a field (different from constructor)
+     * @param fieldName name of a field
      * @return this Builder for method chaining
      * @see Class#forName(String)
      * @see Class#getField(String)
@@ -286,7 +287,7 @@ public class DynFields {
      * Checks for a hidden implementation, first finding the class by name.
      *
      * @param className name of a class
-     * @param fieldName name of a field (different from constructor)
+     * @param fieldName name of a field
      * @return this Builder for method chaining
      * @see Class#forName(String)
      * @see Class#getField(String)
@@ -311,7 +312,7 @@ public class DynFields {
      * Checks for a hidden implementation.
      *
      * @param targetClass a class instance
-     * @param fieldName name of a field (different from constructor)
+     * @param fieldName name of a field
      * @return this Builder for method chaining
      * @see Class#forName(String)
      * @see Class#getField(String)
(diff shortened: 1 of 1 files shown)


LuciferYang:
cc @pan3793

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/797ba2bd43bf44dd2db94ac35e756c8d69abbce6
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/cbb31bc2d483e83b04b2af6c531478b113dbac27

LuciferYang:
Thank you  @pan3793

PR #6499, opened 2024-06-24
Rewrite some utility methods in Java

pan3793:
# :mag: Description

This PR rewrites some utility methods in Java, specifically, 
```
Utils.isWindows
Utils.isMac
Utils.findLocalInetAddress
```

and moves them from `kyuubi-common`'s `Utils` to the `kyuubi-util`'s `JavaUtils`, so that they could be used in other modules that do not depend on `kyuubi-common`.

## Types of changes :bookmark:

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Test Plan 🧪

Pass GHA.

---

# Checklist 📝

- [x] This patch was not authored or co-authored using [Generative Tooling](https://www.apache.org/legal/generative-tooling.html)

**Be nice. Be informative.**

PR #7702, opened 2026-08-31
[KYUUBI #7701] Log the resolved host in the findLocalInetAddress loopback warning

### LINKED ISSUE #7701: JavaUtils.findLocalInetAddress logs the substitute address instead of the host that resolved to loopback
### Describe the bug

`JavaUtils.findLocalInetAddress()` returns `InetAddress.getLocalHost()` unchanged unless it is loopback. When it is, the method scans network interfaces for an address that is neither loopback nor link-local and warns that it is substituting one. That warning reads its host name and loopback address from the substitute rather than from the address that resolved to loopback:

```java
          LOG.warn(
              "{} was resolved to a loopback address: {}, using {}",
              addr.getHostName(),
              addr.getHostAddress(),
              strippedAddress.getHostAddress());
```

`addr` is the interface address just picked. On a host with no PTR record for it, `addr.getHostName()` returns the numeric form and all three slots print the same string:

```
10.0.0.5 was resolved to a loopback address: 10.0.0.5, using 10.0.0.5
```

So the line never says which host was resolved or what it resolved to. The fallback warning in the same method, taken when no interface has a usable address, reads from the original address, so the two warnings disagree about their subject.

There is a second cost. `addr` carries no cached host name, so `addr.getHostName()` goes to the resolver, and the call sits in the argument list of the `warn` invocation, so it runs whether or not WARN is enabled. Timings for that one call on a single host at different times: 0.9 ms, 1.8 ms, 6.5 ms, 324 ms, 1027 ms, 3147 ms, each returning the numeric form. The tail is the problem, not the median. `InetAddress.getLocalHost()` already carries its name, so reading from it costs nothing.

The branch is only entered on a host whose name resolves to loopback, which is the Debian and Ubuntu default (`127.0.1.1` in `/etc/hosts`) and common on developer machines. Hosts with a routable name never see it.

### Affects Version(s)

master (1.13.0-SNAPSHOT), 1.12, 1.11, 1.10. Introduced by the Java rewrite in #6499; the Scala predecessor used the original address in both slots.

Yes.

LuciferYang:

Closes #7701.

`findLocalInetAddress` warns when the host name resolves to loopback and it substitutes an address taken from a network interface. The warning filled its host name and loopback slots from the substitute rather than from the address that actually resolved to loopback, so on a host with no PTR record for the substitute all three slots printed the same string: `10.0.0.5 was resolved to a loopback address: 10.0.0.5, using 10.0.0.5`. The fallback warning a few lines down already reads from the original address, so the two warnings in one method had been describing different things.

Naming the wrong subject is also what sent the thread to the resolver. An address from `NetworkInterface.getInetAddresses()` carries no cached host name, so `addr.getHostName()` blocks on a lookup, and the argument is evaluated at the call site, so the lookup happens whether or not WARN is enabled. The same call measured 0.9 ms, 1.8 ms, 6.5 ms, 324 ms, 1027 ms and 3147 ms on one host at different times, every one returning the numeric form for want of a PTR record. `InetAddress.getLocalHost()` already carries its name, so the corrected slots cost nothing. This removes that one lookup and no other: callers that ask the returned address for `getCanonicalHostName()`, such as `KyuubiRestFrontendService` and `EmbeddedZookeeper`, do their own regardless of what this method returns.

The Scala code that #6499 replaced used the original address in both slots, so the Java rewrite moved the receiver. The line is unchanged on `branch-1.12`, `branch-1.11` and `branch-1.10`, so a backport is possible if maintainers want one.

No unit test. `findLocalInetAddress` takes no arguments and reads `InetAddress.getLocalHost()` and `NetworkInterface.getNetworkInterfaces()` as statics, so a test cannot steer it into the loopback branch on an ordinary host. The returned address is identical before and after; only the log text differs, so an assertion over the return value would pass with this patch reverted. Asserting on the text would mean adding a logging backend or a static mocking library to `kyuubi-util`, whose test scope today is `junit-jupiter` alone. The missing coverage of these branches is tracked separately in #7692.

```
build/mvn test -pl kyuubi-util -am -DwildcardSuites=none -Dtest=JavaUtilsTest
```

The argument rendering was checked out of tree on a host that takes the branch, where `InetAddress.getLocalHost()` returns `<name>/127.0.0.1` and an interface carries a routable address. The old arguments render as `<lan-address> was resolved to a loopback address: <lan-address>, using <lan-address>`; the new ones as `<name> was resolved to a loopback address: 127.0.0.1, using <lan-address>`.

Assisted-by: Claude Opus 5

### DIFF
--- a/kyuubi-util/src/main/java/org/apache/kyuubi/util/JavaUtils.java
+++ b/kyuubi-util/src/main/java/org/apache/kyuubi/util/JavaUtils.java
@@ -67,13 +67,15 @@ public class JavaUtils {
                   .findFirst()
                   .orElse(addresses.get(0));
 
+          // Rebuild from the raw bytes to drop the IPv6 zone suffix that getHostAddress would
+          // otherwise append.
           InetAddress strippedAddress = InetAddress.getByAddress(addr.getAddress());
 
           // We've found an address that looks reasonable!
           LOG.warn(
               "{} was resolved to a loopback address: {}, using {}",
-              addr.getHostName(),
-              addr.getHostAddress(),
+              address.getHostName(),
+              address.getHostAddress(),
               strippedAddress.getHostAddress());
           return strippedAddress;
         }


pan3793 (approved):

LuciferYang:
Thank you @pan3793

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/6dd1d73639d1c9e334ddd984e79f33d5eb0a27ae
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/93140e41430b708d0a1f41aa914e4a6d87d75aa2
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/2813b381150d1d83e2338ef7b03635ade8ce2b1e
- merged into branch-1.10 https://github.com/apache/kyuubi/commit/2b567f14eff31ed80c9ff643911dedea411fdf1c

PR #7706, opened 2026-08-31
[KYUUBI #7705][UTIL] Report the candidates a failed DynMethods lookup tried

### LINKED ISSUE #7705: DynMethods build failures report only the method name, discarding every candidate that was tried
### Describe the bug

`DynMethods.Builder` probes a chain of candidate classes and signatures, swallowing each lookup failure so the next candidate gets a turn. When none matches, `build()` and `buildChecked()` throw with the method name alone:

```java
    public UnboundMethod build() {
      if (method != null) {
        return method;
      } else {
        throw new RuntimeException("Cannot find method: " + name);
      }
    }
```

Every `ClassNotFoundException` and `NoSuchMethodException` collected on the way is gone, so the message names neither the classes that were tried nor the signatures they were tried with.

`HiveConnectorUtils.storageFormatApply` shows what that costs. It probes the seven-argument `CatalogStorageFormat.apply` that SPARK-55645 added in Spark 4.2, then the six-argument one:

```scala
    DynMethods.builder("apply")
      .impl( // SPARK-55645 (4.2.0): 7-arg apply with serdeName
        classOf[CatalogStorageFormat],
        ...
      .impl( // Spark < 4.2.0: 6-arg apply without serdeName
```

On a Spark whose signature matches neither, the whole diagnosis is `Cannot find method: apply`. Nothing says which two arities were tried, so there is no way to tell a Spark version this code has never seen from a genuine bug in the probe. `RowSet.toHiveString` (four-argument then five-argument) and `SparkUtilsHelper.getLocalDir` have the same shape.

`DynConstructors.Builder` in the same package does not have this problem: it records each miss in a `problems` map and attaches the entries to the thrown exception, both in the message and as suppressed throwables.

It does have a related one, which surfaces as soon as `DynMethods.ctorImpl` starts reporting what it delegates. `DynConstructors.methodName` renders a candidate by calling `getName()` on every argument type, so a null element throws `NullPointerException` out of the `catch` that was supposed to record an ordinary miss:

```java
      sb.append(type.getName());
```

A null element is what a caller passes when it probes an optional dependency through `DynClasses.builder()...orNull().build()`, and `getConstructor` treats it as a plain no-match rather than an error. `EmbeddedExecutorFactory` writes that shape with `LEGACY_TIME_CLASS`, which is null on Flink 2.x, and is saved only by an earlier candidate matching first.

### Affects Version(s)

master (1.13.0-SNAPSHOT), and every earlier release carrying this file.

Yes.

LuciferYang:

Closes #7705.

`DynMethods.Builder` swallows per-candidate lookup failures and, when none matches, `build()`/`buildChecked()` threw with only the method name, e.g. `Cannot find method: apply`. Now the failure names every candidate signature it tried, each paired with its `NoSuchMethodException`/`ClassNotFoundException`:

```
Cannot find method: apply
	Missing org.apache.spark.sql.catalyst.catalog.CatalogStorageFormat#apply(...) [java.lang.NoSuchMethodException: ...]
	Missing ...#apply(...) [...]
```

This matches what `DynConstructors.Builder` already does. Entries are a list, not a name-keyed map, since two lookups can render the same name for unrelated reasons (e.g. `ClassNotFoundException` then `NoSuchMethodException`); declared-member candidates are marked so `hiddenImpl`/`impl` pairs read distinctly.

Two gaps fixed along the way:
- A null argument class in a candidate (as `SparkUtilsHelper` passes below Spark 4.1) must count as an ordinary miss, not a throw; `DynConstructors.methodName` had the same gap.
- #7687's `catch (RuntimeException e)` in `hiddenImpl` now records a candidate too, preserving the "does not `opens`" text instead of discarding it.

Also: javadoc corrections (parquet -> iceberg provenance, `RuntimeError` -> `RuntimeException`). `NoClassDefFoundError` handling and `DynFields` cause recording are left for follow-ups.

20 tests (19 `DynMethodsTest` + 1 `DynConstructorsTest`) cover each recording site, message layout/order, throwable pairing, suppressed order, null argument class/array on both paths, and `orNoop` with prior misses. Each was verified by mutation testing (mutate, recompile, rerun, restore). Module 65/65 green on Zulu 17.0.18; `spotless:check` passes.

```
build/mvn test -pl kyuubi-util -am -DwildcardSuites=none -Dtest='DynMethodsTest,DynConstructorsTest'
```

Assisted-by: Claude Opus 5


pan3793:
@LuciferYang could you pleasa rebase this one as there are conflicts?

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/e0522cd3205855f2806e2389626cdac780611aaa
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/72c70597084f2914f2dbace8321213b1f23da24f

PR #7709, opened 2026-09-02
[KYUUBI #7708][SERVER] Reconcile zero-count metadata updates

### LINKED ISSUE #7708: [Bug] Persistent metadata update failures can block batch session timeout cleanup
### Code of Conduct

### Search before asking

- [x] I have searched the existing open and closed issues and found no similar issue.

### Describe the bug

A persistent JDBC metadata update failure can remain in the asynchronous retry queue indefinitely and block batch session cleanup.

The relevant sequence is:

1. `JDBCMetadataStore.updateMetadata` throws when `PreparedStatement.executeUpdate()` returns `0`.
2. With `kyuubi.metadata.request.async.retry.enabled=true`, `MetadataManager.updateMetadata` enqueues the failed update.
3. The retry worker removes the queue head only after a successful update. A deterministic failure therefore remains at the head and is retried indefinitely.
4. `KyuubiBatchSession.close` waits without a timeout until the metadata retry queue is empty.
5. The session timeout checker is a single-thread scheduled executor and calls `closeSession` synchronously. One affected batch session can therefore prevent later sessions from being checked and removed. Old sessions may continue to appear in `/api/v1/sessions` even while their metadata rows are independently removed by the metadata cleaner.

In the observed sequence, a batch session was closed while its application monitor was still running. The monitor later observed the driver as `FAILED`, updated the application information, and the asynchronous operation's `finally` block attempted another terminal metadata update. This explains repeated terminal updates, but it has **not** been established that repetition alone caused Connector/J to return `0`.

The exact reason for the first zero update count is still under investigation. This report is about the deterministic failure amplification after any persistent zero-count update, not an assumption about its original cause.

#### Reproduction outline

1. Configure the MySQL JDBC metadata store with asynchronous retry enabled.
2. Submit a batch and make a terminal metadata update return `0`, for example by deleting the row before the update or by using a test metadata store that produces a deterministic zero-count update.
3. Close the corresponding batch session.
4. Observe that the same update remains at the retry queue head.
5. Observe that `KyuubiBatchSession.close` does not return and that the single session timeout checker cannot process later sessions.

#### Expected behavior

- Metadata persistence failures must not block session lifecycle cleanup indefinitely.
- A zero-count update should be reconciled using its postcondition instead of being treated as unconditional success or unconditional transient failure:
  - if the row exists and all columns included in the update already match the intended values, the operation is idempotently complete;
  - if the row is missing or its values do not match, retain a classified failure without silently recreating deleted metadata;
  - preserve ordering when an earlier INSERT for the same identifier is pending.
- Persistent failures should use observable and bounded/rate-limited handling, and normal session close should not wait indefinitely for the asynchronous retry queue.

### Affects Version(s)

Observed on a deployment based on 1.9.1. The relevant code paths are also present on master at `804c00239815d3b7eece5c4c8bb8f6263063beff`.

### Kyuubi Server Log Output

```logtalk
INFO  ... KyuubiBatchSession: Mark session SessionHandle [<batch-id>] closed
INFO  ... BatchJobSubmission: Batch report for <batch-id>, Some(ApplicationInfo(...,FAILED,...))
ERROR ... MetadataManager: Error updating metadata for session <batch-id>
org.apache.kyuubi.KyuubiException: Error updating metadata for <batch-id> by SQL: UPDATE metadata SET ... WHERE identifier = ?
INFO  ... MetadataManager: Retrying metadata requests for <batch-id>
ERROR ... MetadataManager: Error retrying metadata requests for <batch-id>
INFO  ... KyuubiBatchSession: There are still remaining metadata store requests for batch[<batch-id>]
```

### Kyuubi Engine Log Output

Not required to reproduce the metadata retry and session-close behavior.

```yaml
kyuubi.metadata.store.jdbc.database.type: MYSQL
kyuubi.metadata.request.async.retry.enabled: true
kyuubi.metadata.request.retry.interval: 5s
```

The deployment is intended to use MySQL Connector/J 5.x and the JDBC URL does not explicitly set `useAffectedRows=true`; the effective runtime driver and connection properties are still being verified.

### Additional context

Relevant current-master code:

- [`JDBCMetadataStore.updateMetadata`](https://github.com/apache/kyuubi/blob/804c00239815d3b7eece5c4c8bb8f6263063beff/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala#L393-L401)
- [metadata asynchronous retry loop](https://github.com/apache/kyuubi/blob/804c00239815d3b7eece5c4c8bb8f6263063beff/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala#L322-L380)
- [`KyuubiBatchSession` waiting for retry completion](https://github.com/apache/kyuubi/blob/804c00239815d3b7eece5c4c8bb8f6263063beff/kyuubi-server/src/main/scala/org/apache/kyuubi/session/KyuubiBatchSession.scala#L127-L136)

- [x] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to fix this.

ruanwenjun:

A JDBC metadata `UPDATE` may report an update count of zero when the row is missing or when the requested values are already present. Kyuubi currently treats every zero count as a retryable failure. A persistent failure therefore remains in the retry queue and can block batch-session timeout cleanup while it waits for metadata retries to finish.

When an `UPDATE` returns zero, this patch verifies only the updated fields on the same JDBC connection:

- An exact match is treated as an idempotent success.
- A missing or mismatched row is retried once through the existing FIFO queue, then discarded if the postcondition still fails. This preserves recovery for an earlier queued `INSERT` without retrying a permanently missing row forever.
- Database access failures remain retryable, and the existing close-time retry wait is preserved.

Closes #7708.

- Added a SQLite trigger-based regression test that forces `executeUpdate()` to return zero for both matching and mismatching rows.
- Added retry coverage showing that a persistently missing row is removed from the retry queue.
- Kept the existing FIFO `INSERT`-then-`UPDATE` retry coverage passing.
- Ran:

  ```shell
    -Pspark-provided,flink-provided,hive-provided \
    -Dtest=none \
    -DwildcardSuites=org.apache.kyuubi.server.metadata.jdbc.JDBCMetadataStoreSuite,org.apache.kyuubi.server.metadata.MetadataManagerSuite
  ```

  All 13 tests passed.

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataStore.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataStore.scala
@@ -19,6 +19,7 @@ package org.apache.kyuubi.server.metadata
 
 import java.io.Closeable
 
+import org.apache.kyuubi.KyuubiException
 import org.apache.kyuubi.server.metadata.api.{KubernetesEngineInfo, Metadata, MetadataFilter}
 
 trait MetadataStore extends Closeable {
@@ -118,3 +119,17 @@ trait MetadataStore extends Closeable {
    */
   def cleanupKubernetesEngineInfoByAge(maxAge: Long, limit: Int): Int
 }
+
+sealed abstract private[metadata] class MetadataUpdatePostconditionException(message: String)
+  extends KyuubiException(message)
+
+private[metadata] class MetadataRowNotFoundException(identifier: String)
+  extends MetadataUpdatePostconditionException(
+    s"Metadata row $identifier was not found after an update returned 0")
+
+private[metadata] class MetadataUpdateMismatchException(
+    identifier: String,
+    val mismatchedColumns: Seq[String])
+  extends MetadataUpdatePostconditionException(
+    s"Metadata row $identifier does not match the requested update for columns: " +
+      mismatchedColumns.mkString(", "))
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
@@ -34,7 +34,7 @@ import com.zaxxer.hikari.{HikariConfig, HikariDataSource}
 import org.apache.kyuubi.{KyuubiException, Logging, Utils}
 import org.apache.kyuubi.config.KyuubiConf
 import org.apache.kyuubi.operation.OperationState
-import org.apache.kyuubi.server.metadata.MetadataStore
+import org.apache.kyuubi.server.metadata.{MetadataRowNotFoundException, MetadataStore, MetadataUpdateMismatchException}
 import org.apache.kyuubi.server.metadata.api.{KubernetesEngineInfo, Metadata, MetadataFilter}
 import org.apache.kyuubi.server.metadata.jdbc.DatabaseType._
 import org.apache.kyuubi.server.metadata.jdbc.JDBCMetadataStoreConf._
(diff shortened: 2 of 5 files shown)


pan3793:
the change lgtm, cc @turboFei, who authored this feature, to take a look

pan3793 (commented):
Not for this PR: the queued `InsertMetadata` retry path has a similar amplification pattern. A queued insert that hits a duplicate-key error rethrows to the outer catch and stays at the queue head forever - the foreground path fails fast on `unrecoverableDBErr`, but the retry loop does not apply the same check. Worth a follow-up issue.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/jdbc/JDBCMetadataStore.scala
```diff
@@ -332,76 +332,97 @@ class JDBCMetadataStore(conf: KyuubiConf) extends MetadataStore with Logging {
...
-          throw new KyuubiException(
-            s"Error updating metadata for ${metadata.identifier} by SQL: $query, " +
-              s"with params: ${params.mkString(", ")}")
+          verifyMetadataUpdatePostcondition(connection, metadata.identifier, updateFields.toSeq)
         }
       }
     }
   }
 
+  private def verifyMetadataUpdatePostcondition(
```
pan3793, line 392: Since #7708 says the original cause of the zero update count is still under investigation, please log a `warn` here when the row already matches, e.g. "update for {} returned 0 but all requested values already match, treating as idempotent success". Otherwise this case becomes invisible, while the old behavior at least surfaced an exception.

kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataStore.scala
```diff
@@ -118,3 +119,17 @@ trait MetadataStore extends Closeable {
...
 }
+
+sealed abstract private[metadata] class MetadataUpdatePostconditionException(message: String)
+  extends KyuubiException(message)
+
+private[metadata] class MetadataRowNotFoundException(identifier: String)
+  extends MetadataUpdatePostconditionException(
+    s"Metadata row $identifier was not found after an update returned 0")
+
+private[metadata] class MetadataUpdateMismatchException(
```
pan3793, line 130: Nit: consider including expected vs actual values in the message, not just column names. When the discard is logged in `MetadataManager`, that makes the error actionable without reproducing the DB state.

PR #7027, opened 2025-04-16
refresh

turboFei:

Issue #4843, opened 2023-05-16
Support multiple kubernetes contexts and namespaces


turboFei:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues) and found no similar issues.

### Describe the proposal

Support to submit spark engine to multiple kubernetes contexts and namespaces.

- KubernetesApplicationOperation support multiple kubernetes contexts and namespaces
- detect the context and space that engine used.

### Task list

TBD

### Are you willing to submit PR?

- [ ] Yes. I would be willing to submit a PR with guidance from the Kyuubi community to improve.
- [ ] No. I cannot submit a PR at this time.

PR #7710, opened 2026-09-02
[KYUUBI #7293][KUBERNETES] Initialize in-cluster client automatically

### LINKED ISSUE #7293: [Bug] Kyuubi Server might not delete the driver pod after completed
### Code of Conduct

### Search before asking

### Describe the bug

Hi Kyuubi Team,

I'm running Kyuubi on k8s and setting `kyuubi.kubernetes.spark.cleanupTerminatedDriverPod.kind=ALL` in Kyuubi Server but it seems not work as expected. After reviewing the source code in version 1.10.2, specifically the Kubernetes informer initialization logic [here](https://github.com/apache/kyuubi/blob/v1.10.2/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala#L186), I noticed that the informer is only initialized when configurations like the following are provided:
```
kyuubi.kubernetes.client.initialize.list=context1:namespace1
kyuubi.kubernetes.authenticate.context1.oauthToken=
kyuubi.kubernetes.authenticate.context1.caCertFile=
...
```
According to the documentation, `kyuubi.kubernetes.client.initialize.list` is only documented as available starting from 1.11.0, yet the corresponding initialization code already exists in 1.10.2.

Additionally, this setup feels unnecessarily complex for cases where Kyuubi is running inside Kubernetes and uses the ServiceAccount credentials that was attached to Pod. In such scenarios, users would expect Kyuubi to automatically initialize the Kubernetes informer based on the in-cluster configuration. However, without explicitly setting `kyuubi.kubernetes.client.initialize.list`, no informer is created, which prevents features like terminated driver pod cleanup from working.

Could you please clarify:
- Whether this behavior is intended in 1.10.2 ?
- Kyuubi is expected to work without explicitly setting `kyuubi.kubernetes.client.initialize.list` (using CA cert, token from pod filesystem) ?

### Affects Version(s)

1.10.2

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

```yaml
kyuubi.server.administrators=admin
# Kyuubi Metrics
kyuubi.metrics.enabled=true
kyuubi.metrics.reporters=PROMETHEUS
kyuubi.metrics.prometheus.port=10019

## User provided Kyuubi configurations
kyuubi.kubernetes.namespace=kyuubi
kyuubi.kubernetes.spark.cleanupTerminatedDriverPod.kind=ALL
kyuubi.kubernetes.terminatedApplicationRetainPeriod=PT1M
kyuubi.ha.addresses=zookeeper.zookeeper.svc:2181
kyuubi.ha.namespace=kyuubi
kyuubi.ha.zookeeper.auth.digest=admin:password
kyuubi.frontend.rest.bind.host=0.0.0.0
kyuubi.frontend.connection.url.use.hostname=false
kyuubi.frontend.thrift.binary.bind.port=10009
kyuubi.frontend.thrift.http.bind.port=10010
kyuubi.frontend.rest.bind.port=10099
kyuubi.frontend.mysql.bind.port=3309
kyuubi.frontend.protocols=MYSQL,REST,THRIFT_BINARY

## User provided Kyuubi configurations
kyuubi.engine.pool.size=-1
kyuubi.engine.share.level=CONNECTION
kyuubi.session.close.on.disconnect=true
kyuubi.engine.kubernetes.submit.timeout=PT90S
kyuubi.session.engine.spark.main.resource=local:///opt/kyuubi/externals/engines/spark/kyuubi-spark-sql-engine_2.12-1.10.2.jar
```

```yaml

```

### Additional context

_No response_

miaht94:

When Kyuubi Server runs inside Kubernetes without
`kyuubi.kubernetes.client.initialize.list`, no Kubernetes client or informer is initialized
at startup. As a result, terminated Spark driver pods can remain after their retention period
even though cleanup is enabled.

Initialize the configured namespace automatically from the in-cluster environment while
preserving the explicit initialization list as the higher-priority configuration. Outside
Kubernetes, the existing empty-list behavior is unchanged.

Closes #7293.

- `dev/gen/gen_all_config_docs.sh`
- `dev/reformat`
- `build/mvn test -pl kyuubi-server -am -Pspark-provided,hive-provided,flink-provided -Dtest=none -DwildcardSuites=org.apache.kyuubi.engine.KubernetesApplicationOperationSuite`
- `build/mvn scalastyle:check -pl kyuubi-common,kyuubi-server -am -Pflink-provided,hive-provided,spark-provided,spark-3.5,tpcds`
- `helm lint charts/kyuubi`
- `helm template review-test charts/kyuubi --show-only templates/kyuubi-role.yaml`

The focused suite covers no automatic initialization outside Kubernetes, automatic
in-cluster initialization with non-empty Kubernetes service environment variables, and
explicit-list precedence. The rendered Role grants pod and service informer permissions.

Assisted-by: OpenAI Codex


miaht94:
@pan3793 Hi, would you mind reviewing this PR?

wangzhigang1999 (changes requested):
Could you update the default Helm RBAC rules as part of this change?

When the initialization list is empty, the official Helm deployment will now enter this branch and call `buildKubernetesClient`. That method starts a pod informer, which requires `list` and `watch` permissions on pods. Since the default `kyuubi.kubernetes.spark.appUrlPattern` uses `{{SPARK_DRIVER_SVC}}`, it also starts a service informer, which requires `list` and `watch` permissions on services.

The current `charts/kyuubi/values.yaml` only grants `create`, `list`, and `delete` on pods. It does not grant `watch` on pods or any permissions on services. A default Helm installation would therefore receive 403 responses and would not observe pod termination events, so the cleanup described in this PR would still not work.

The Kubernetes integration test does not catch this because the workflow binds the Kyuubi service account to `clusterrole=edit`. Please add the least-privilege pod and service informer permissions to the default Helm Role, and preferably cover the rendered Role in a test.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -217,14 +217,26 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
...
+          case Array(ctx) => (Some(ctx).filterNot(_.isEmpty), None)
+          case _ => (None, None)
+        }
+        KubernetesInfo(context, namespace)
       }
-      KubernetesInfo(context, namespace)
+    if (configuredInitializeInfo.nonEmpty) {
+      configuredInitializeInfo
+    } else if (environment.contains(KUBERNETES_SERVICE_HOST) &&
+      environment.contains(KUBERNETES_SERVICE_PORT)) {
```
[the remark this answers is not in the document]
miaht94: Addressed in 5181e49. The fallback now requires non-empty Kubernetes service environment variables and readable service account token and CA files. The unit test also covers the missing-credential case.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -217,14 +217,26 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
...
+      Seq(KubernetesInfo(
+        None,
+        Some(kyuubiConf.get(KyuubiConf.KUBERNETES_NAMESPACE))))
```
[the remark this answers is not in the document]
miaht94: Addressed in 5181e49. The configured namespace is now filtered consistently, so an empty value produces no namespace override.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1427,7 +1427,10 @@ object KyuubiConf {
       .doc("The kubernetes client initialize list to register kubernetes resource informers" +
         " during Kyuubi server startup. This ensure the Kyuubi server is promptly informed for" +
         " any Kubernetes resource changes after startup. It is highly recommend to set it for" +
```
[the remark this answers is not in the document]
miaht94: Addressed in 5181e49, and the generated configuration documentation has been refreshed.

miaht94:
Thanks for the review. Addressed in 5181e49: the default Helm Role now grants pod create/list/watch/delete and service list/watch permissions. I validated the chart with Helm lint and inspected the rendered Role. I also strengthened in-cluster detection to require readable service account credentials, handled an empty namespace consistently, and refreshed the configuration documentation.

wangzhigang1999:
Thanks for the quick update, and sorry for the churn. I should have caught these issues before your update.

The RBAC and grammar changes look good. Could you revert the other two changes?

- Requiring the standard token and CA files is too restrictive. `KubernetesUtils` already supports configured authentication, and this check would silently skip initialization for those setups. Keeping the non-empty environment checks is enough.
- Filtering an empty namespace to `None` does not fix it because `getKubernetesConf` retains the original empty value. Namespace validation should be handled in the config entry if needed.

Sorry again for the extra work.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1425,9 +1425,12 @@ object KyuubiConf {
...
       .doc("The kubernetes client initialize list to register kubernetes resource informers" +
-        " during Kyuubi server startup. This ensure the Kyuubi server is promptly informed for" +
-        " any Kubernetes resource changes after startup. It is highly recommend to set it for" +
-        " multiple Kyuubi instances mode. The format is `context1:namespace1,context2:namespace2`.")
+        " during Kyuubi server startup. This ensures the Kyuubi server is promptly informed for" +
+        " any Kubernetes resource changes after startup. It is highly recommended to set it for" +
+        " multiple Kyuubi instances mode. The format is" +
+        " `context1:namespace1,context2:namespace2`." +
+        " When the list is empty and Kyuubi runs in Kubernetes, the client for the configured" +
+        " namespace is initialized automatically with the in-cluster configuration.")
```
pan3793, line 1436: Since `kyuubi.kubernetes.namespace` now also selects the namespace watched by the auto-initialized in-cluster client, please fix its doc (`KUBERNETES_NAMESPACE`, above) in this PR - the current text "running the kyuubi pods" is misleading, as it has never controlled where the Kyuubi server runs. Suggestion:

```
The default namespace used by the Kyuubi server's Kubernetes client to discover and
manage engine pods, when the engine submission does not specify one (e.g.
`spark.kubernetes.namespace`). It does not control the namespace where the Kyuubi
server itself runs.
```

Remember to regenerate `settings.md` after the change.

kyuubi-common/src/main/scala/org/apache/kyuubi/config/KyuubiConf.scala
```diff
@@ -1425,9 +1425,12 @@ object KyuubiConf {
...
       .version("1.11.0")
```
pan3793, line 1437: Note for other reviewers: this entry shipped in v1.10.2 via the branch-1.10 backport of KYUUBI #7027, though `version()` says 1.11.0 - pre-existing, not related to this PR.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -217,14 +218,31 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
...
       }
-      KubernetesInfo(context, namespace)
+    if (configuredInitializeInfo.nonEmpty) {
+      configuredInitializeInfo
+    } else if (environment.get(KUBERNETES_SERVICE_HOST).exists(_.nonEmpty) &&
+      environment.get(KUBERNETES_SERVICE_PORT).exists(_.nonEmpty) &&
+      Files.isReadable(serviceAccountTokenPath) &&
+      Files.isReadable(serviceAccountCaCertPath)) {
+      Seq(KubernetesInfo(
+        None,
```
pan3793, line 236: Please add a brief comment that `context` is intentionally `None` here, since kube context is meaningless for in-cluster configuration.

pan3793 (commented):
LGTM overall. This effectively restores the pre-1.8.0 behavior, where the server built the in-cluster client and informer unconditionally at startup, before KYUUBI #4843 (1.8.0) switched to lazy per-KubernetesInfo client creation. The credential-file check nicely mirrors what `KubernetesUtils.buildKubernetesClient` consumes, and the Helm RBAC update covers the informer verbs.

A few comments inline.

kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KubernetesApplicationOperation.scala
```diff
@@ -217,14 +217,27 @@ class KubernetesApplicationOperation extends ApplicationOperation with Logging {
...
+          case Array(ctx, ns) => (Some(ctx).filterNot(_.isEmpty), Some(ns).filterNot(_.isEmpty))
+          case Array(ctx) => (Some(ctx).filterNot(_.isEmpty), None)
+          case _ => (None, None)
+        }
+        KubernetesInfo(context, namespace)
       }
-      KubernetesInfo(context, namespace)
+    if (configuredInitializeInfo.nonEmpty) {
+      configuredInitializeInfo
+    } else if (environment.get(KUBERNETES_SERVICE_HOST).exists(_.nonEmpty) &&
```
pan3793: One more thought: `kyuubi-common` already has `Utils.isOnK8s`, but it checks `/var/run/secrets/kubernetes.io`, which under-detects pods running with `automountServiceAccountToken: false`. The env-var check you use here matches client-go's `rest.InClusterConfig()` semantics (kubelet always injects `KUBERNETES_SERVICE_HOST`/`PORT`, even with `enableServiceLinks: false`), so it is the more authoritative signal.

Could you fix `Utils.isOnK8s` to use the env-var check and reuse it here? For testability, add an env parameter:

```scala
def isOnK8s(env: Map[String, String] = sys.env): Boolean =
  env.get("KUBERNETES_SERVICE_HOST").exists(_.nonEmpty) &&
    env.get("KUBERNETES_SERVICE_PORT").exists(_.nonEmpty)
```

This also makes the existing caller in `SparkSQLEngine` (prefer-IP registration, KYUUBI #3597 (1.7.0)) behave correctly when automount is disabled.

miaht94: Thanks for pointing it out, i've fixed all the chores above in: https://github.com/apache/kyuubi/pull/7710/changes/46f4649af3fd291c7da27c2efa48efeaba0c3a78
Also changed + reused `Utils.isOnK8s` in: https://github.com/apache/kyuubi/pull/7710/changes/cc3e487994367a704e6c951c0e069d1aef19e1fc

miaht94: I've corrected the syntax errors in the `isOnK8s` call site; could you please trigger the integration CI?

pan3793:
LGTM

wangzhigang1999 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/5c80e74bcef6727356d2996de6f4bf792bdff1cc

PR #7713, opened 2026-09-04
[KYUUBI #7712] Support IPv6 `host:port` parsing in Kyuubi service discovery

### LINKED ISSUE #7712: [Bug] Kyuubi service discovery cannot parse IPv6 `host:port` endpoints
### Code of Conduct

### Search before asking

### Describe the bug

**1. Describe the Bug**

Kyuubi fails to discover or connect to server/engine endpoints when running in an IPv6-only cluster. The server side (`DiscoveryClient.parseInstanceHostPort`) and the standalone JDBC client side (`ZooKeeperHiveClientHelper`) both parse a `host:port` string by splitting on `":"`. This is incorrect for IPv6 because IPv6 addresses themselves contain colons.

For example, an endpoint registered to ZooKeeper as: `fc00:172:0:0:0:0:0:47a:10009` (IPv6 without brackets) 
is split into 8 segments instead of `host` + `port`, so the host is wrongly resolved as `fc00` and the port as `172`. This results in either a `NumberFormatException` or a `UnknownHostException: fc00` when the client tries to connect.

**2. Root cause**

Host/port parsing relies on naive `split(":")`, which is only correct for IPv4 and hostnames. Hive already solved this in `org.apache.hadoop.hive.common.IPStackUtils` by using `lastIndexOf(':')` and handling square brackets around IPv6 addresses. Kyuubi has no equivalent utility, so IPv6-only deployments break.

**3. Expected behavior**

Kyuubi should correctly parse IPv4, IPv6 (with or without square brackets), and hostname `host:port` strings on both the server side and the standalone JDBC client side, so that service discovery works in IPv6-only clusters.

### Affects Version(s)

master

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

```yaml

```

```yaml

```

### Additional context

_No response_

maomaodev:

Fix https://github.com/apache/kyuubi/issues/7712. Kyuubi currently cannot discover or connect to server/engine endpoints in IPv6-only clusters. Both the server side (`DiscoveryClient.parseInstanceHostPort`) and the standalone JDBC client side (`ZooKeeperHiveClientHelper`) parse a `host:port` string by naively splitting on `":"`. This breaks for IPv6 because IPv6 addresses themselves contain `:`.

The correct parsing behavior already exists upstream in Hive (`org.apache.hadoop.hive.common.IPStackUtils`). This patch mirrors that logic in a new lightweight `IPStackUtils` utility under `kyuubi-util`, so it can be shared between the server side and the standalone JDBC client (which only depends on `kyuubi-util`).

- Added `IPStackUtilsTest` (11 test cases) aligned with Hive's own [IPStackUtilsTest](https://github.com/apache/hive/blob/master/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/common/IPStackUtils.java), covering IPv4, IPv6 with brackets, IPv6 without brackets, hostname, invalid port, missing port, missing host and port range validation.
- Extended `DiscoveryClientSuite` with IPv6 cases (with and without brackets) and converted it from an orphan trait to a runnable class.

Assisted-by: DeepSeek-V4-Pro

### DIFF
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
@@ -120,7 +120,7 @@ abstract class TFrontendService(name: String)
       case (None, None) => serverAddr.getHostAddress
     }
 
-    IPStackUtils.formatIPAddressForURL(host) + ":" + actualPort
+    IPStackUtils.concatHostPort(host, actualPort)
   }
 
   protected def getProxyUser(
--- a/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
+++ b/kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
@@ -36,7 +36,7 @@ import org.apache.kyuubi.shaded.hive.service.rpc.thrift._
 import org.apache.kyuubi.shaded.thrift.protocol.TProtocol
 import org.apache.kyuubi.shaded.thrift.server.{ServerContext, TServerEventHandler}
 import org.apache.kyuubi.shaded.thrift.transport.TTransport
-import org.apache.kyuubi.util.{JavaUtils, KyuubiHadoopUtils, NamedThreadFactory}
+import org.apache.kyuubi.util.{IPStackUtils, JavaUtils, KyuubiHadoopUtils, NamedThreadFactory}
 
 /**
  * Apache Thrift based hive-service-rpc base class
@@ -120,7 +120,7 @@ abstract class TFrontendService(name: String)
       case (None, None) => serverAddr.getHostAddress
     }
 
-    host + ":" + actualPort
+    IPStackUtils.formatIPAddressForURL(host) + ":" + actualPort
   }
 
   protected def getProxyUser(
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
@@ -30,6 +30,7 @@ import org.apache.kyuubi.jdbc.hive.strategy.zk.RandomSelectStrategy;
 import org.apache.kyuubi.shaded.curator.framework.CuratorFramework;
 import org.apache.kyuubi.shaded.curator.framework.CuratorFrameworkFactory;
 import org.apache.kyuubi.shaded.curator.retry.ExponentialBackoffRetry;
+import org.apache.kyuubi.util.IPStackUtils;
 
 class ZooKeeperHiveClientHelper {
   // Pattern for key1=value1;key2=value2
@@ -96,13 +97,14 @@ class ZooKeeperHiveClientHelper {
     // it must be the server uri added by an older version HS2
     Matcher matcher = kvPattern.matcher(dataStr);
     if (!matcher.find()) {
-      String[] split = dataStr.split(":");
-      if (split.length != 2) {
+      try {
+        IPStackUtils.HostPort hostPort = IPStackUtils.getHostAndPort(dataStr);
+        connParams.setHost(hostPort.getHostname());
+        connParams.setPort(hostPort.getPort());
+      } catch (Exception e) {
         throw new ZooKeeperHiveClientException(
-            "Unable to read HiveServer2 uri from ZooKeeper: " + dataStr);
+            "Unable to parse HiveServer2 uri from ZooKeeper: " + dataStr, e);
       }
-      connParams.setHost(split[0]);
-      connParams.setPort(Integer.parseInt(split[1]));
     } else {
       applyConfs(dataStr, connParams);
     }
--- a/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
+++ b/kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
@@ -101,7 +101,7 @@ class ZooKeeperHiveClientHelper {
         IPStackUtils.HostPort hostPort = IPStackUtils.getHostAndPort(dataStr);
         connParams.setHost(hostPort.getHostname());
         connParams.setPort(hostPort.getPort());
-      } catch (Exception e) {
+      } catch (IllegalArgumentException e) {
         throw new ZooKeeperHiveClientException(
             "Unable to parse HiveServer2 uri from ZooKeeper: " + dataStr, e);
       }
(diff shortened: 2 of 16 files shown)


kyuubi-util/src/main/java/org/apache/kyuubi/util/IPStackUtils.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Returns an integer representation of the port number. Also validates whether the given string
+   * represents a valid port number. A valid port number is an integer between 0 and 65535
+   * inclusive.
+   *
+   * @param portString The string representing the port number.
+   * @return {@code int} the port number.
+   */
```
pan3793, line 77: add `@throws IllegalArgumentException ...`

kyuubi-util/src/main/java/org/apache/kyuubi/util/IPStackUtils.java
```diff
@@ -0,0 +1,138 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kyuubi.util;
+
+/**
+ * Utility methods for handling IPv4/IPv6 host:port strings. The logic mirrors Hive's {@code
+ * org.apache.hadoop.hive.common.IPStackUtils}.
```
pan3793: also say the referenced hive version

pan3793 (commented):
Please prefix the title with `[KYUUBI #7712]`.

A few more places have the same issue and should be fixed in this PR:

- `kyuubi-ha/src/main/scala/org/apache/kyuubi/ha/client/zookeeper/ZookeeperDiscoveryClient.scala` L319: when `kyuubi.ha.zookeeper.publish.configs=true`, an IPv6 instance is split on the first colon and publishes a broken host/port. Please use `IPStackUtils.getHostAndPort` too.
- `kyuubi-server/src/main/scala/org/apache/kyuubi/server/ui/JettyServer.scala` L59: `getServerUri` (and the `advertisedHost` branch in `KyuubiRestFrontendService.connectionUrl`) emits unbracketed IPv6, which breaks the batch `kyuubiInstance` / `InternalRestClient` path in IPv6-only clusters. Please fix these with `IPStackUtils` as well.
- `kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/Utils.java` L409 (also L422, L536): authority strings concatenate host without brackets, so the rewritten `jdbcUriString` is malformed for IPv6. Please use `IPStackUtils.formatIPAddressForURL` when building them.

kyuubi-hive-jdbc/src/main/java/org/apache/kyuubi/jdbc/hive/ZooKeeperHiveClientHelper.java
```diff
@@ -96,13 +97,14 @@ private static void updateParamsWithZKServerNode(
     // it must be the server uri added by an older version HS2
     Matcher matcher = kvPattern.matcher(dataStr);
     if (!matcher.find()) {
-      String[] split = dataStr.split(":");
-      if (split.length != 2) {
+      try {
+        IPStackUtils.HostPort hostPort = IPStackUtils.getHostAndPort(dataStr);
+        connParams.setHost(hostPort.getHostname());
+        connParams.setPort(hostPort.getPort());
+      } catch (Exception e) {
```
pan3793: `getHostAndPort` throws `IllegalArgumentException` only, catch it instead of `Exception`.

kyuubi-common/src/main/scala/org/apache/kyuubi/service/TFrontendService.scala
```diff
@@ -120,7 +120,7 @@ abstract class TFrontendService(name: String)
       case (None, None) => serverAddr.getHostAddress
     }
 
-    host + ":" + actualPort
+    IPStackUtils.formatIPAddressForURL(host) + ":" + actualPort
```
pan3793: Port Hive's `concatHostPort` and use it here, it also validates host and port.

kyuubi-util/src/main/java/org/apache/kyuubi/util/IPStackUtils.java
pan3793: HIVE-28782 (4.1.0) introduced `org.apache.hadoop.hive.common.IPStackUtils`.

pan3793 (approved):

maomaodev:
@pan3793 Good catch! Fixed all three, and swept the codebase for the same pattern.

**Your three:**

- `ZookeeperDiscoveryClient.addConfsToPublish`: `split(":", 2)` → `IPStackUtils.getHostAndPort`.
- `JettyServer` (`start` / `getServerUri`) & `KyuubiRestFrontendService.connectionUrl` (advertisedHost): → `IPStackUtils.concatHostPort`.
- `Utils.java` L409/L422/L536: authority building → `IPStackUtils.concatHostPort` (which internally calls `formatIPAddressForURL` + validates host/port).

**Also fixed (same bug elsewhere):**

- `KyuubiTrinoFrontendService.connectionUrl` (advertisedHost branch).
- `ServiceNodeInfo`: the `instance` string published to service discovery.
- `EngineRef`: Data Agent JDBC URL.
- `KyuubiConnection.getServerHttpUrl`: HTTP URL construction.
- `KyuubiConfFileParser.addHosts`: `hosts` property fed to beeline JDBC URL builder.

**Related:**

- Ported Hive's `concatHostPort` (validates host/port) into `IPStackUtils` with UT mirroring Hive.
- Narrowed `catch (Exception)` in `DiscoveryClient` to `catch (IllegalArgumentException)` per your earlier point.

**Intentionally left alone** (all non-functional, happy to revisit if you disagree):

- `KyuubiSessionImpl.scala`: 5 `s"$host:$port"` occurrences, all in `warn` / `info` / `error` / `logSessionInfo` messages — display only, no parsing downstream.
- `KyuubiConnection.java`: `LOG.warn("Failed to connect to " + host + ":" + port)` and `logZkDiscoveryMessage("Connected to " + host + ":" + port)` — same, log-only.
- `TBinaryFrontendService.scala` / `KyuubiTHttpFrontendService.scala`: `s"Failed to initialize frontend service on $serverAddr:$portNum"` — exception message string, not consumed by any code path.
- `EmbeddedZookeeper.scala`: `s"...binding to $host:$clientPort"` — exception message, real ZK connect string comes from `getConnectString`.
- `EngineUIProxyServlet.scala` L87: `X-Forwarded-Context` header value is a path prefix, not a URL authority.
- `HiveSiteHS2ConnectionFileParser.java` L215: Hive-side compatibility layer using `getHostName()` (returns hostname string, not a numeric address).

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/b090cc28eb55c0f0d47189a6952dec66c3c474bb

PR #7714, opened 2026-09-06
[SPARK] Update PATH_CONFIGS and validate main-resource path

aajisaka:

Add more configs into Spark `PATH_CONFIGS`. Also, add path validation for `KyuubiConf.ENGINE_SPARK_MAIN_RESOURCE`.

This can fail sessions that previously worked because of the validation.

Added regression tests for `KyuubiConf.ENGINE_SPARK_MAIN_RESOURCE`.

Assisted-by: OpenCode (GLM-5.3)

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/KyuubiApplicationManager.scala
@@ -186,6 +186,11 @@ object KyuubiApplicationManager {
           }
         }
       }
+      // the main resource is not a spark.* conf but spark-submit uploads it from the server
+      // host as the primary resource of the engine application
+      appConf.get(KyuubiConf.ENGINE_SPARK_MAIN_RESOURCE.key).filter(_.nonEmpty).foreach { path =>
+        checkApplicationAccessPath(path, kyuubiConf)
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


pan3793 (approved):
+1. Verified the newly policed keys against Spark 3.5.8 bytecode - all three are read and
distributed by the submission client, and the main-resource check is placed where `ProcBuilder`
actually reads it. Tests look adequate.

Two asks:

1. Please also add
   `spark.kubernetes.authenticate.driver.{oauthTokenFile,clientCertFile,clientKeyFile,caCertFile}`
   to `PATH_CONFIGS` in this PR - same class of configs (happy to discuss details privately).
2. PR description: worth noting this can fail sessions that previously worked, and the title
   undersells the main-resource fix.

Nit: the `PATH_CONFIGS` doc comment source list is no longer exhaustive.

aajisaka:
Thank you @pan3793!

Included the configs into `PATH_CONFIGS`. Also updated the PR title and description.

aajisaka:
@pan3793 would you take a look at the latest commit?

pan3793:
Re-checked the new revision, CI is green.

Two small things on the Kubernetes additions:

1. Please drop `spark.kubernetes.authenticate.driver.oauthTokenFile`: it is not read by Spark
   (checked 3.5, 4.3 and master). The driver only supports the inline
   `spark.kubernetes.authenticate.driver.oauthToken`; a file form exists only as
   `...driver.mounted.oauthTokenFile`, which is a path inside the pod. My earlier suggestion was
   imprecise on this one - sorry. The other three (`caCertFile`, `clientKeyFile`, `clientCertFile`)
   are exactly right.
2. The new source bullet does not match these entries: `SparkKubernetesClientFactory
   ::createKubernetesClient` reads the files with the prefix passed by the caller (submission:
   `spark.kubernetes.authenticate.submission`, driver: `...driver.mounted` / client mode:
   `spark.kubernetes.authenticate`). The non-mounted `driver.*File` keys added here are read by
   `DriverKubernetesCredentialsFeatureStep::configurePod` - please cite that instead.

aajisaka:
Thank you @pan3793 ! Updated

pan3793 (approved):

aajisaka:
Merged into master and branch-1.12.

PR #7721, opened 2026-09-10
[KYUUBI #7720][SERVER] Reconcile duplicate-key insert retries

### LINKED ISSUE #7720: [Bug] Queued InsertMetadata retry can block metadata retry queue forever on duplicate-key error
### Code of Conduct

### Search before asking

### Describe the bug

MetadataManager.insertMetadata intentionally fails fast when it hits a duplicate-key error (unrecoverableDBErr), rethrowing rather than queuing a retry. However, when an InsertMetadata request that is already in the async retry queue is replayed by startMetadataRequestsAsyncRetryTrigger, it calls insertMetadata(metadata, asyncRetryOnError = false). If that replay hits a duplicate-key error (e.g. because the original insert actually succeeded and only the acknowledgment was lost), the exception is not caught by any case in insertMetadata and propagates out of the retry loop's while block before ref.metadataRequests.remove(request) runs. The request is left at the head of the queue and is retried indefinitely, blocking all later queued requests for that session — the same amplification pattern reported in #7708 for updates, but for inserts.

Additional context:
Raised during review of #7709 by @pan3793: "the queued InsertMetadata retry path has a similar amplification pattern... Worth a follow-up issue."

Affects Version(s): master

Are you willing to submit a PR? Yes.

### Affects Version(s)

Master

### Kyuubi Server Log Output

```logtalk

```

### Kyuubi Engine Log Output

```logtalk

```

```yaml

```

```yaml

```

### Additional context

_No response_

nnvscholar-collab:

When a queued `InsertMetadata` retry is replayed by the async retry trigger, it can hit a duplicate-key error if the original insert actually succeeded but its acknowledgement was lost. `insertMetadata` intentionally fails fast on duplicate-key errors, and since the retry loop only caught exceptions around the whole `while` block, this exception escaped before the request could be removed from the queue — leaving it stuck at the head and blocking all later requests for that session indefinitely. This is the same amplification pattern reported in #7708 for updates, but for inserts (raised by @pan3793 during review of #7709).

This patch verifies the postcondition inside the retry loop specifically: if a duplicate-key error occurs while retrying an insert, and a row matching the identifier, session type, and create time already exists, it's treated as an idempotent success and removed from the queue. The fail-fast behavior of `insertMetadata()` itself is unchanged for direct/foreground calls.

Closes #7720.

Added a regression test verifying that a queued insert retry which hits a duplicate-key error is removed from the queue instead of blocking subsequent requests for the same session.

Ran:
build/mvn test -pl kyuubi-server -am -Pspark-provided,flink-provided,hive-provided -Dtest=none -DwildcardSuites=org.apache.kyuubi.server.metadata.MetadataManagerSuite

Assisted-by: Claude

### DIFF
--- a/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
+++ b/kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
@@ -135,6 +135,18 @@ class MetadataManager extends AbstractService("MetadataManager") {
     }
   }
 
+  // Verifies whether a duplicate-key error on a queued insert retry actually means the
+  // insert already succeeded earlier (e.g. the acknowledgement was lost), by checking
+  // whether a matching row already exists. Used only by the retry loop below, so the
+  // fail-fast contract of insertMetadata() itself is unchanged for direct/foreground calls.
+  private def verifyInsertPostcondition(metadata: Metadata): Boolean = {
+    Option(_metadataStore.getMetadata(metadata.identifier)).exists { existing =>
+      existing.identifier == metadata.identifier &&
+      existing.sessionType == metadata.sessionType &&
+      existing.createTime == metadata.createTime
+    }
+  }
+
   def getBatch(batchId: String): Option[Batch] = {
     getBatchSessionMetadata(batchId).map(buildBatch)
   }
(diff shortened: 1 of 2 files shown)


kyuubi-server/src/main/scala/org/apache/kyuubi/server/metadata/MetadataManager.scala
```diff
@@ -344,11 +356,30 @@ class MetadataManager extends AbstractService("MetadataManager") {
...
+                        case rethrow: Throwable
+                            if request.isInstanceOf[InsertMetadata] &&
+                              unrecoverableDBErr(rethrow) &&
+                              verifyInsertPostcondition(request.metadata) =>
+                          warn(
+                            s"Insert for ${request.metadata.identifier} returned a " +
+                              "duplicate-key error on retry but a matching row already " +
+                              "exists; treating as idempotent success and removing it " +
+                              "from the retry queue.",
+                            rethrow)
```
ruanwenjun, line 382: Could you handle the exception that occurs when trying to update a nonexistent record here? That way, we can resolve this type of issue within a single RP. Then I can close #7709

PR #7724, opened 2026-09-10
[KYUUBI #7723][UTIL] Answer false from isClassLoadable when the class fails to link or initialize

### LINKED ISSUE #7723: ReflectUtils.isClassLoadable throws LinkageError instead of answering false for classes that fail to link or initialize
### Describe the bug

`ReflectUtils.isClassLoadable` probes a class through `DynClasses`, wrapped in `scala.util.Try`:

```scala
def isClassLoadable(
    className: String,
    cl: ClassLoader = Thread.currentThread().getContextClassLoader): Boolean =
  Try {
    DynClasses.builder().loader(cl).impl(className).buildChecked()
  }.isSuccess
```

`DynClasses.impl` resolves the name with `Class.forName(className, true, loader)`. `initialize = true` means loading the class also links it and runs its static initializer, so a class that is present but fails at that stage raises a `LinkageError`, not a `ClassNotFoundException`: `NoClassDefFoundError` for a missing superclass or dependency, `ExceptionInInitializerError` when a static initializer throws.

`Try` only catches `NonFatal`, and `LinkageError` is excluded from `NonFatal`. So a class that fails to link or initialize never reaches `.isSuccess`; the `LinkageError` escapes `isClassLoadable` and crashes the caller. A loadability probe should answer `false` for such a class, not throw.

Both call sites treat the result as a feature-detection boolean and never expect an exception:

- `Logging.scala:152`: `if (ReflectUtils.isClassLoadable("org.slf4j.bridge.SLF4JBridgeHandler"))`
- `JDBCMetadataStore.scala:51`: `if (ReflectUtils.isClassLoadable("com.mysql.cj.jdbc.Driver"))`

A probed class that is on the classpath but whose static initializer throws (a misconfigured dependency, say) aborts the check instead of being reported as not loadable.

The `Try` wrapping came in with the reflection-utility consolidation (#4887); nothing there depended on a `LinkageError` propagating.

### Affects Version(s)

Yes.

LuciferYang:

Closes #7723.

`ReflectUtils.isClassLoadable` wrapped its `DynClasses` probe in `scala.util.Try`, which catches only `NonFatal`. The probe resolves the name with `Class.forName(className, true, loader)`, so a class that is present but fails to link or initialize raises a `LinkageError` (`NoClassDefFoundError`, `ExceptionInInitializerError`) rather than a `ClassNotFoundException`. `LinkageError` is not `NonFatal`, so it escaped the probe and crashed the caller instead of answering `false`.

This replaces the `Try` with an explicit `try`/`catch` that answers `false` on `ClassNotFoundException` and `LinkageError`. The two call sites (`Logging` probing the SLF4J bridge, `JDBCMetadataStore` probing the MySQL driver) read the result as a boolean and are only made safer. `OutOfMemoryError`/`StackOverflowError` are `VirtualMachineError`, not `LinkageError`, so they stay fatal exactly as they did under `Try`.

Added a `ReflectUtilsSuite` case that probes a top-level object whose static initializer throws, by its binary name, and asserts `isClassLoadable` returns `false`. Reverting the production change makes the suite abort as the `ExceptionInInitializerError` escapes, which pins the regression.

```
```

Module green on Zulu 17.0.18 (22/22), scalastyle 0 errors, spotless clean.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
+++ b/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/reflect/ReflectUtils.scala
@@ -22,12 +22,12 @@ import java.util.ServiceLoader
 import scala.collection.JavaConverters._
 import scala.language.existentials
 import scala.reflect.ClassTag
-import scala.util.Try
 
 object ReflectUtils {
 
   /**
-   * Determines whether the provided class is loadable
+   * Determines whether the provided class is loadable, answering false when it cannot
+   * be found, linked, or initialized.
    * @param className the class name
    * @param cl the class loader
    * @return is the class name loadable with the class loader
@@ -35,9 +35,15 @@ object ReflectUtils {
   def isClassLoadable(
       className: String,
       cl: ClassLoader = Thread.currentThread().getContextClassLoader): Boolean =
-    Try {
+    try {
       DynClasses.builder().loader(cl).impl(className).buildChecked()
-    }.isSuccess
+      true
+    } catch {
+      // scala.util.Try does not catch LinkageError, so a class that fails to link or
+      // initialize would escape the probe instead of answering false
+      case _: ClassNotFoundException => false
+      case _: LinkageError => false
+    }
 
   /**
    * get the field value of the given object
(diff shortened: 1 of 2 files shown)


pan3793 (approved):

LuciferYang:
can we merge this one? @pan3793

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/c002a42b3b8cb649e892e38b58fe93e0a29df8e9
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/5b3a10d2311dd573114c3b2956f1ef4c8aa6cf82
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/6eafef5bdec0052126316e9c0808b9d63fe03685
- merged into branch-1.10 https://github.com/apache/kyuubi/commit/a1ec06300d878de8a4381242b9044317c6ae1081

LuciferYang:
Thank you @pan3793

PR #7726, opened 2026-09-10
[KYUUBI #7725][UTIL] Parse two-segment snapshot versions like 1.2-SNAPSHOT correctly

### LINKED ISSUE #7725: SemanticVersion drops the minor of X.Y-SUFFIX version strings like 1.2-SNAPSHOT
### Describe the bug

`SemanticVersion` parses a version string with:

```scala
private val semanticVersionRegex = """^(\d+)(?:\.(\d+))?(\..*)?$""".r
```

The minor group `(?:\.(\d+))?` is optional and the trailing group `(\..*)?` also begins with a literal `.`. On `"1.2-SNAPSHOT"` the trailing group cannot match the `-`, so the engine backtracks: it drops the minor group, and the trailing group then matches `.2-SNAPSHOT` whole. The parse succeeds as `(major=1, minor=0)`, so the minor `.2` is silently lost with no parse error.

`isVersionAtLeast`/`isVersionAtMost` then compare against that wrong `(1, 0)`, so `SemanticVersion("1.2-SNAPSHOT").isVersionAtLeast("1.1")` returns `false`.

This is reachable for Flink development builds. `EnvironmentInformation.getVersion` returns strings like `X.Y-SNAPSHOT`, and `FlinkEngineUtils.checkFlinkVersion` runs them through `SemanticVersion`. A supported `1.20-SNAPSHOT` parses as `(1, 0)` and fails the version gate, while an unsupported `2.4-SNAPSHOT` parses as `(2, 0)` and passes as if it were `2.0`.

The regex first made the minor group optional in #5039 (to accept a bare major such as `"3"`); the backtracking side effect on `X.Y-SUFFIX` came with it and was never intended, since nothing pins `(1, 0)` as the expected parse of such input.

### Affects Version(s)

Yes.

LuciferYang:

Closes #7725.

`SemanticVersion`'s regex made the minor group optional while keeping a trailing group that also starts with `.`:

```
^(\d+)(?:\.(\d+))?(\..*)?$
```

On `"1.2-SNAPSHOT"` the trailing group cannot match the `-`, so the engine backtracks past the minor group and lets the trailing group swallow `.2-SNAPSHOT`. The parse succeeds as `(major=1, minor=0)` with the minor silently dropped, so `isVersionAtLeast`/`isVersionAtMost` compare against the wrong version. This is reachable for Flink development builds (`X.Y-SNAPSHOT` from `EnvironmentInformation.getVersion`, checked in `FlinkEngineUtils.checkFlinkVersion`): a supported `1.20-SNAPSHOT` is rejected with `UnsupportedOperationException`, and an unsupported `2.4-SNAPSHOT` slips through as `2.0`.

Letting the trailing group start with `.` or `-` (`(?:[.-].*)?`) keeps the minor group and lets the suffix be consumed after it, matching how `.4-SNAPSHOT` was already handled following a minor. Capture groups 1 and 2 are unchanged, and the trailing group was never read, so making it non-capturing is safe.

Added a `SemanticVersionSuite` case asserting `"1.2-SNAPSHOT"` parses to `(1, 2)` and satisfies the `1.1`/`1.3` bounds; it fails on the old regex (minor 0). The existing cases parse identically under both regexes.

```
```

Module green on Zulu 17.0.18, scalastyle 0 errors, spotless clean.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/SemanticVersion.scala
+++ b/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/SemanticVersion.scala
@@ -65,7 +65,7 @@ case class SemanticVersion(majorVersion: Int, minorVersion: Int)
 
 object SemanticVersion {
 
-  private val semanticVersionRegex = """^(\d+)(?:\.(\d+))?(\..*)?$""".r
+  private val semanticVersionRegex = """^(\d+)(?:\.(\d+))?(?:[.-].*)?$""".r
 
   def apply(versionString: String): SemanticVersion = {
     semanticVersionRegex.findFirstMatchIn(versionString) match {
--- a/kyuubi-util-scala/src/test/scala/org/apache/kyuubi/util/SemanticVersionSuite.scala
+++ b/kyuubi-util-scala/src/test/scala/org/apache/kyuubi/util/SemanticVersionSuite.scala
@@ -38,6 +38,14 @@ class SemanticVersionSuite extends AnyFunSuite {
     assert(version.minorVersion === 14)
   }
 
+  test("parse two-segment snapshot version") {
+    val version = SemanticVersion("1.2-SNAPSHOT")
+    assert(version.majorVersion === 1)
+    assert(version.minorVersion === 2)
+    assert(version.isVersionAtLeast("1.1"))
+    assert(version.isVersionAtMost("1.3"))
+  }
+
   test("parse binary version") {
     val version = SemanticVersion("0.9")
     assert(version.majorVersion === 0)


pan3793 (approved):

LuciferYang:
can we merge this one @pan3793 ?

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/1885ea1e1c85c30bf59a09de4a729e428163e21a
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/347ce1ea94f8c8a247139ac619db1bdb80eef81b
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/3d9d6baf5894ca425b88292ef58a6679b06d960b
- merged into branch-1.10 https://github.com/apache/kyuubi/commit/b1f4de9bb5f29e474de6e0a6873d91f1ce8f30ee

LuciferYang:
Thank you @pan3793

PR #7728, opened 2026-09-10
[KYUUBI #7727][UTIL] Trim each side of the pair in genKeyValuePair

### LINKED ISSUE #7727: Spaced conf keys escape exact-key redaction in logged engine commands
### Describe the bug

`genKeyValuePair` assembles a `key=value` command-line argument by trimming only the concatenated string:

```scala
def genKeyValuePair(key: String, value: String): String = s"$key=$value".trim
```

`.trim` on the whole string strips the outer ends but leaves any whitespace between the key and the value. `genKeyValuePair("abc ", "123")` returns `"abc =123"`, with the space kept before the `=`, so the trim the method advertises is only half applied.

This has a concrete consequence for secret redaction. `redactConfValues` splits each assembled argument on `=` and matches the key against an exact key set. When a conf key carries stray whitespace, the assembled argument is `"<key> =<value>"`, so the extracted key `"<key> "` misses the exact set and the value is not redacted.

The gap is reachable for the Data Agent engine's API key. `kyuubi.engine.data.agent.openai.api.key` is not matched by the default pattern-based redaction (`secret|password|token|access.key`), so exact-key redaction is the only stage that would mask it. Session conf keys are not trimmed on the way through, so a key submitted with a trailing space escapes that stage and the API key value lands in the engine launch command written to the server log.

### Affects Version(s)

Yes.

LuciferYang:

Closes #7727.

`genKeyValuePair` trimmed only the concatenated string (`s"$key=$value".trim`), so whitespace between the key and value survived: `genKeyValuePair("abc ", "123")` returned `"abc =123"`. `redactConfValues` splits an assembled argument on the first `=` and matches the key against an exact key set, so a spaced key (`"<key> ="`) misses the set and the value is not redacted.

This matters for the Data Agent engine's API key. `kyuubi.engine.data.agent.openai.api.key` is not matched by the default pattern-based redaction, so exact-key redaction is its only guard, and session conf keys are not trimmed on the way to the engine command. A key carrying a trailing space slipped both stages and logged the API key value. Trimming each side (`s"${key.trim}=${value.trim}"`) makes the assembled pair carry no inner whitespace, so the exact-key stage sees the clean key. `Utils.redactCommandLineArgs` also assembles through `genKeyValuePair`, so the same fix closes the hole at both assembly points.

Added `CommandUtilsSuite` cases: symmetric-trim assertions (`genKeyValuePair("abc ", " 123")` -> `"abc=123"`, including a tab) and a redaction round-trip that assembles a spaced Data Agent api-key argument and asserts `redactConfValues` masks it. Both fail on the old `.trim`.

```
build/mvn test -pl kyuubi-common -am -DwildcardSuites=org.apache.kyuubi.UtilsSuite
```

`kyuubi-util-scala` green on Zulu 17.0.18 (22/22), scalastyle 0 errors, spotless clean; `kyuubi-common` `UtilsSuite` (which exercises `redactCommandLineArgs`) 14/14.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/command/CommandLineUtils.scala
+++ b/kyuubi-util-scala/src/main/scala/org/apache/kyuubi/util/command/CommandLineUtils.scala
@@ -35,9 +35,9 @@ object CommandLineUtils {
   val CP = "-cp"
 
   /**
-   * Assemble key value pair with "=" seperator
+   * Assemble key value pair with "=" seperator, trimming each side of the pair
    */
-  def genKeyValuePair(key: String, value: String): String = s"$key=$value".trim
+  def genKeyValuePair(key: String, value: String): String = s"${key.trim}=${value.trim}"
 
   /**
    * Assemble key value pair with config option prefix
--- a/kyuubi-util-scala/src/test/scala/org/apache/kyuubi/util/command/CommandUtilsSuite.scala
+++ b/kyuubi-util-scala/src/test/scala/org/apache/kyuubi/util/command/CommandUtilsSuite.scala
@@ -30,9 +30,27 @@ class CommandUtilsSuite extends AnyFunSuite {
     assertResult("abc=123")(genKeyValuePair("   abc", "123   "))
     assertResult("abc.def=xyz.123")(genKeyValuePair("abc.def", "xyz.123"))
 
+    // trim each side of the pair separately, so a dirty key or value cannot leave
+    // whitespace inside the assembled argument
+    assertResult("abc=123")(genKeyValuePair("abc ", " 123"))
+    assertResult("abc=123")(genKeyValuePair("abc\t", "123"))
+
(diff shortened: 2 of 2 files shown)


pan3793 (approved):

LuciferYang:
can we merge this one ? @pan3793

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/fb4ad6c8b1818cc0381368a056a8b17e2b198b0f
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/1e0acf9b0983405b9ead2cf26343aac4c794898d
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/05aef35edd417f14cc41f652a8a9d7e4b16ab0dd
- merged into branch-1.10 https://github.com/apache/kyuubi/commit/f6e4e633d91def4a1f7db94d4d76f4630541141f

LuciferYang:
Thank you @pan3793

PR #7744, opened 2026-09-14
[KYUUBI #7434][DOC] Reformat connector docs from RST to Markdown

dnskr:
The changes are needed to unify the format used for documentation, as described in the issue https://github.com/apache/kyuubi/issues/7434.

The PR **does not** change page contents (only very minimal adjustments) to ensure an easier review of the migration.

The pages contain out-of-date information and broken links, requiring a comprehensive review and updates in a separate PRs.

Tested by building the documentation and comparing the pages against the live site.

Assisted-by: Qwen Coder


docs/connector/spark/tpch.md
```diff
@@ -0,0 +1,95 @@
...
+- Referencing the TPC-H connector [dependencies](#dependencies)
+- Setting the Spark catalog [configurations](#configurations)
+
+### Dependencies
+
+The **classpath** of Kyuubi Spark SQL engine with TPC-H supported consists of:
+
+1. kyuubi-spark-sql-engine-{{ release }}_2.12.jar, the engine jar deployed with a Kyuubi distribution
+2. A copy of Spark distribution
+3. `kyuubi-spark-connector-tpch-<release>_2.12.jar`, which can be found in the [Maven Central](https://repo1.maven.org/maven2/org/apache/kyuubi/kyuubi-spark-connector-tpch_2.12/)
```
wangzhigang1999: Could we preserve {{ release }} here and in the spark.jars example below, as in tpcds.md? The [1.12.0 documentation](https://kyuubi.readthedocs.io/en/v1.12.0/connector/spark/tpch.html#dependencies) renders the matching version in both places. Keeping the substitution would preserve this behavior after the format migration.

wangzhigang1999 (approved):
Non-blocking; otherwise looks good. Built the documentation locally and checked the rendered pages.

docs/connector/spark/tpch.md
dnskr: Nice catch, thanks!

I've fixed it:

Screenshot 2026-09-20 at 20 11 14

maomaodev: In fact, we should use `kyuubi-spark-sql-engine_2.12-1.13.0-SNAPSHOT.jar` instead of `kyuubi-spark-sql-engine-1.13.0-SNAPSHOT_2.12.jar`. There are similar issues elsewhere in the doc, should we update them all together? Note: the PMC changed this once before, see: https://github.com/apache/kyuubi/commit/c3edb02cd36c1bd29c77631eb4c6b0d47a66661e

dnskr: Good catch! That definitely needs to be fixed.

As noted in the PR description, these docs have out-of-date info and broken links that need a broader review. To keep this PR's scope focused, I'm going to merge it as-is. Feel free to open a follow-up PR for the artifact name fixes if you'd like, that would be greatly appreciated!

dnskr:
Thank you for the review!

Merged to `master` and `branch-1.12`.

PR #7747, opened 2026-09-15
[KYUUBI #7746][AUTHZ] Report nondeterministic permanent views as nondeterministic to InlineCTE

### LINKED ISSUE #7746: [Bug] A nondeterministic permanent view is recomputed for every CTE reference under authz
### Code of Conduct

### Search before asking

### Describe the bug

With the authz extension enabled, a permanent view over a nondeterministic expression is
recomputed for every reference when it is referenced more than once through a CTE. Without the
extension Spark computes it once and shares the result through a `ReusedExchange`.

```sql
CREATE VIEW v AS SELECT id, reflect('java.util.UUID', 'randomUUID') AS r FROM range(20);
WITH c AS (SELECT * FROM v)
SELECT a.id, a.r, b.r FROM c a JOIN c b ON a.id = b.id;
-- authz enabled: a.r <> b.r on every row; authz disabled: a.r = b.r
```

`InlineCTE.shouldInline` inlines a CTE that is referenced more than once only when
`cteDef.deterministic` holds. `PermanentViewMarker` is a `LeafNode`, so `QueryPlan.deterministic`
sees no expressions and no children and reports `true` for any view, including one over
`rand()`, `uuid()` or `reflect()`. The `Inline CTE` batch runs before the extended optimizer
batch that removes the marker, so the wrong answer is what `InlineCTE` sees.

For seeded expressions such as `uuid()` and `rand()` the inlined copies happen to agree because
both copies share the seed and the partition layout, so the visible symptom is the duplicated
work. For anything that draws a fresh value per evaluation the two references disagree.

Reproduced on Spark 3.5, 4.0, 4.1 and 4.2 (the marker has been a `LeafNode` since #5780, released in 1.9.0).

### Affects Version(s)

master, 1.12.0, 1.11.x, 1.10.x, 1.9.x

### Additional context

`SubqueryExpression.deterministic` reads the same flag, so a scalar or IN subquery over such a
view is also treated as deterministic under authz (e.g. by `MergeScalarSubqueries`).

A PR with a one-line fix (`override lazy val deterministic = child.deterministic`) and tests follows.

j1wonpark:

Fixes #7746.

`PermanentViewMarker` is a `LeafNode`, so `QueryPlan.deterministic` reports `true` for any
permanent view, even one over `rand()`, `uuid()` or `reflect()`. `InlineCTE` reads that flag and
inlines a nondeterministic view CTE that is referenced more than once, so each reference
recomputes the view and, for expressions that draw a fresh value per evaluation, the references
disagree. Without authz Spark keeps the CTE out of line and shares one `ReusedExchange`.

This overrides `deterministic` to delegate to the wrapped plan. The same flag feeds
`SubqueryExpression.deterministic`, so subqueries over such a view are classified the same way
as without authz.

`override lazy val` compiles against both the `lazy val` in Spark 3.5 and the `def` in 4.0+.

Two tests in `HiveCatalogRangerSparkExtensionSuite`:

- a `uuid()` view referenced twice through a CTE keeps a `RepartitionByExpression` in the
  optimized plan and a `ReusedExchange` in the executed plan;
- a `reflect('java.util.UUID', 'randomUUID')` view referenced twice through a CTE yields the same
  value for both references.

Both fail with `PermanentViewMarker` reverted to master and pass with the change, on Spark 3.5.8,
4.0.3, 4.1.3 and 4.2.0.

Assisted-by: Claude Fable 5.1


wForget (approved):
Thanks @j1wonpark

pan3793 (approved):

pan3793:
**Merge Summary:**
- merged into master https://github.com/apache/kyuubi/commit/df5aa6c8b966d868f5ad2a0d4f5d148ebddc63ba
- merged into branch-1.12 https://github.com/apache/kyuubi/commit/fdb6c667deaac469243e83ccb480b5cbc90ca92d
- merged into branch-1.11 https://github.com/apache/kyuubi/commit/3ffd0ee9b1ea59242ac444aa3bf0cfaad1f0b6d5
- merged into branch-1.10 https://github.com/apache/kyuubi/commit/0cd1bff191d4c5e0b16688d54808f47a92839a77

Issue #4942, opened 2023-06-08
[Bug] Unable to create Spark executors, k8s label is too long, exceeding 63 characters


khwj:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### Search before asking

- [X] I have searched in the [issues](https://github.com/apache/kyuubi/issues?q=is%3Aissue) and found no similar issues.

### Describe the bug

The default Kubernetes driver pod name set by [EngineRef.scala](https://github.com/apache/kyuubi/blob/9ff46a3c633534c2266ad8e6316b9fddaa024a6c/kyuubi-server/src/main/scala/org/apache/kyuubi/engine/EngineRef.scala#LL129C32-L129C32) is longer than the maximum allowed length of 63 characters. This poses a problem as the driver pod name is subsequently used as a label in Spark executor pods, resulting in invalid label errors.

To mitigate this issue, I have resorted to configuring the `spark.app.name` to a shorter value. However, this workaround hampers our ability to identify specific Spark apps based on session, user, or group (Kyuubi currently does not set the Spark user or group as Kubernetes labels).

### Affects Version(s)

1.7.1

### Kyuubi Engine Log Output

```logtalk
++ id -u
+ myuid=999
++ id -g
+ mygid=1000
+ set +e
++ getent passwd 999
+ uidentry=hadoop:x:999:1000::/home/hadoop:/bin/bash
+ set -e
+ '[' -z hadoop:x:999:1000::/home/hadoop:/bin/bash ']'
+ '[' -n '' ']'
+ SPARK_K8S_CMD=driver
+ [[ driver == executor ]]
+ SPARK_CLASSPATH=':/usr/lib/spark/jars/*'
+ env
+ grep SPARK_JAVA_OPT_
+ sort -t_ -k4 -n
+ sed 's/[^=]*=\(.*\)/\1/g'
+ readarray -t SPARK_EXECUTOR_JAVA_OPTS
+ '[' -n '' ']'
+ '[' -z ']'
+ '[' -z ']'
+ '[' -n '' ']'
+ '[' -z x ']'
+ SPARK_CLASSPATH='/etc/hadoop/conf::/usr/lib/spark/jars/*'
+ '[' -z x ']'
+ SPARK_CLASSPATH='/usr/lib/spark/conf:/etc/hadoop/conf::/usr/lib/spark/jars/*'
+ '[' -n '' ']'
[...]

PR #7752, opened 2026-09-17
[KYUUBI #4942][K8S] Fall back to short driver pod name when it would exceed the 63-char K8s label length limit

HwangDongJun:

Kyuubi's SparkProcessBuilder.appendPodNameConf() generates a driver pod name that
preserves the app name when possible, falling back to a short kyuubi-$engineRefId-driver
name only if the app-name-based name would exceed the 253-char DNS subdomain limit that
applies to Kubernetes pod names (KubernetesUtils.generateDriverPodName).

However, this generated name is also propagated as the *value* of Kubernetes labels
(e.g. via spark.kubernetes.driver.label.*, which Spark in turn uses on executor pods),
and label values are restricted to the stricter 63-char DNS label limit, independent of
the 253-char pod name limit. A moderately long spark.app.name (well within the 253-char
budget) can still produce a driver pod name over 63 characters, causing executor pod
creation to fail with a Kubernetes API 422 error such as:

  Pod "..." is invalid: metadata.labels: Invalid value: "...-driver":
  must be no more than 63 characters

This adds a DRIVER_POD_NAME_AS_LABEL_MAX_LENGTH (63) check to generateDriverPodName(),
so it now falls back to the always-safe kyuubi-$engineRefId-driver name whenever either
the 253-char pod name limit or the 63-char label limit would be violated.

executor pod name prefixes are not affected by this change: they are only ever used as
a *name prefix* (with the executor hostname already truncated to 63 chars separately in
Spark's BasicExecutorFeatureStep), not propagated as a label value.

Added a new test to KubernetesPodNameSuite reproducing a realistic
spark.app.name (derived from Kyuubi's defaultEngineName pattern) that stays well
within the 253-char pod name budget but previously resolved to an 83-character driver
pod name, confirming it now falls back to the short, always-safe name and satisfies
the 63-char label length limit.

Ran the full kyuubi-server test suite locally (mvn test -pl kyuubi-server -am
-P spark-3.5) — all 4 tests in KubernetesPodNameSuite pass, including the 3
pre-existing tests (no regression).

Assisted-by: claude-sonnet-5


pan3793:
> However, this generated name is also propagated as the value of Kubernetes labels
(e.g. via `spark.kubernetes.driver.label.*`, which Spark in turn uses on executor pods)

can you describe how the chain happens in detail?

ruanwenjun:
We have encountered instances in our environment where a fallback occurred to `kyuubi-$engineRefId-driver`, but this should not result in a length exceeding that of the k8s label, since the engineId is 36 characters long, so the total length will not exceed 63 characters.

HwangDongJun:
@pan3793 @ruanwenjun Thanks both for the careful review — you're right to
push on this, and after digging further I don't think this PR should be
merged as-is. Closing it.

To answer @pan3793's question directly: I re-traced the propagation chain
in vanilla Apache Spark's K8s resource manager
(`KubernetesConf`/`BasicDriverFeatureStep`/`BasicExecutorFeatureStep`) and
I could not find any code path where the driver pod *name* is copied into
a Kubernetes *label* value:

- `spark.kubernetes.driver.pod.name` is only ever used to set the driver
  pod's `.metadata.name` (`BasicDriverFeatureStep`).
- The driver/executor `resourceNamePrefix` used for labels and the driver
  service name is computed independently from `spark.app.name` via
  `KubernetesConf.getResourceNamePrefix`, and never reads the configured
  driver pod name.
- `spark.kubernetes.driver.label.*` and `spark.kubernetes.executor.label.*`
  are parsed completely independently (`KubernetesUtils.parsePrefixedKeyValuePairs`)
  — there's no auto-propagation from driver labels to executor labels.
- The only custom driver label Kyuubi itself sets is
  `spark.kubernetes.driver.label.kyuubi-unique-tag=<engineRefId>`
  (`KyuubiApplicationManager`), and `engineRefId` is always a 36-char UUID,
  so it can never exceed the 63-char limit — which is exactly the point
  @ruanwenjun raised.
- The driver Service name (`DriverServiceFeatureStep`) already has its own
  independent 63-char check with its own fallback, unrelated to the driver
  pod name.

I went back to the original report (#4942) and noticed the failing label
key in that log is `spark-driver-pod-name`, together with an
EMR-internal label (`eks-subscription.amazonaws.com/emr.internal.id`) on
the same pod, and the reporter's Spark build is `3.3.1-amzn-0`. I can't
find `spark-driver-pod-name` anywhere in Apache Spark's history, so this
looks like an EMR-specific patch that copies the driver pod name into a
label — not something vanilla Apache Spark or Kyuubi does. That means the
underlying bug is real for that reporter's environment, but the fix
doesn't belong in Kyuubi's `KubernetesUtils.generateDriverPodName`, since
Kyuubi (on vanilla Spark) never actually produces a driver-pod-name-valued
label that could trip the 63-char limit.

Apologies for not verifying the actual label propagation path before
opening this — I should have confirmed where the value would actually end
up as a label before assuming the chain described in the issue applied
generally. Closing this PR; happy to revisit if someone can point to a
concrete vanilla-Spark/Kyuubi code path I missed.

PR #7760, opened 2026-09-22
[KYUUBI #7759] Reduce the connection limit count when session creation fails

### LINKED ISSUE #7759: [Bug] Session limiter count is not returned when session creation fails, locking the user out
### Code of Conduct

### Search before asking

### Describe the bug

`KyuubiSessionManager.openSession` takes a count from the session limiter before opening the
session, and `closeSession` is the only place that gives one back.

`SessionManager.openSession` calls `createSession` outside the `try` block whose `catch` calls
`closeSession`, and `AbstractSession.normalizedConf` is a `val`, so a session whose conf fails
validation dies inside the constructor and is never registered in `handleToSession`.
`closeSession` therefore never runs for it and the count is kept for good.

Once as many opens have failed as `kyuubi.server.limit.connections.per.user` allows, every further
open from that user is rejected by the limiter, with no open session behind the count, until the
server restarts. `kyuubi.server.limit.connections.per.ipaddress` leaks the same way.

This is reachable with ordinary client configuration on a server that follows the documented
`kyuubi.session.conf.restrict.list` guidance: a client that keeps sending a restricted key - for
example the `spark.sql.extensions` entry in the example list of
https://kyuubi.readthedocs.io/en/master/security/session_configuration.html - locks its own
account out.

### Affects Version(s)

master/1.11.1

### Kyuubi Server Log Output

```logtalk
org.apache.kyuubi.KyuubiSQLException: Error opening session for <user> client ip <ip>, due to
spark.sql.extensions is a restrict key according to the server-side configuration, please remove
it and retry if you want to proceed
        at org.apache.kyuubi.session.SessionManager.validateKey(SessionManager.scala:220)
        at org.apache.kyuubi.session.KyuubiSessionManager.openSession(KyuubiSessionManager.scala:125)

... after the limit is reached, the same client gets:

org.apache.kyuubi.KyuubiSQLException: Connection limit per user reached (user: <user> limit: 20)
        at org.apache.kyuubi.session.SessionLimiterImpl.incrLimitCount(SessionLimiter.scala:92)
        at org.apache.kyuubi.session.KyuubiSessionManager.openSession(KyuubiSessionManager.scala:115)
```

### Kyuubi Engine Log Output

```logtalk
> Fails before engine starts
```

```yaml
kyuubi.server.limit.connections.per.user  20
kyuubi.session.conf.restrict.list         spark.sql.extensions
```

```yaml
> Fails before engine starts
```

### Additional context

Observed on a 3-node deployment: each node showed exactly 20 rejections with the restrict-key
message, after which every open from that user failed with the connection-limit message while
`kyuubi_connection_opened` stayed 0. A rolling restart was the only way to recover.

Chaho12:

`KyuubiSessionManager.openSession` takes a count from the session limiter before the session is
opened, and `closeSession` is the only place that gives one back.

`SessionManager.openSession` calls `createSession` outside the `try` whose `catch` calls
`closeSession`, and `AbstractSession.normalizedConf` is a `val`, so a session whose conf fails
validation dies in the constructor and is never registered in `handleToSession`. `closeSession`
never runs for it, and the count is kept for good.

Once as many opens have failed as `kyuubi.server.limit.connections.per.user` allows, every further
open from that user is rejected by the limiter although the user holds no session, until the
server restarts. `kyuubi.server.limit.connections.per.ipaddress` leaks the same way.

Ordinary client configuration reaches this on a server that follows the documented
`kyuubi.session.conf.restrict.list` guidance: `spark.sql.extensions` is in the example restrict
list in the [session configuration docs](https://kyuubi.readthedocs.io/en/master/security/session_configuration.html),
and a client that keeps sending it locks its own account out. We hit this in production - three
server nodes each rejected 20 opens for the restricted key, after which every open from that user
failed with the connection-limit message while `kyuubi_connection_opened` stayed 0, and a rolling
restart was the only way to recover.

Affects master and 1.11.1. Batch sessions are not affected: `openBatchSession` takes its count
after the session is built and inside the `try` that closes it.

close #7759 

Added `KyuubiSessionManagerLimiterSuite`, which sets `kyuubi.server.limit.connections.per.user=2`
and a restrict list, then opens `limit + 2` sessions carrying the restricted key and asserts every
attempt is still rejected by the restrict list rather than by the limiter.

    build/mvn test -pl kyuubi-server -am -Dtest=none \
        -DwildcardSuites=org.apache.kyuubi.session.KyuubiSessionManagerLimiterSuite

Passes with the change; with the change reverted it fails on the third attempt with
`"Connection limit per user reached (user: user limit: 2)" did not contain
"spark.sql.extensions is a restrict key"`, which is the production symptom.

Assisted-by: Claude Opus 5


wForget:
Thanks for your contribution; I prefer handling exceptions within `openSession`, like this:

```
  override def openSession(
      protocol: TProtocolVersion,
      user: String,
      password: String,
      ipAddress: String,
      conf: Map[String, String]): SessionHandle = {
    val username = Option(user).filter(_.nonEmpty).getOrElse("anonymous")
    limiter.foreach(_.increment(UserIpAddress(username, ipAddress)))
    var handle: SessionHandle = null
    try {
      handle = super.openSession(protocol, username, password, ipAddress, conf)
      handle
    } catch {
      case e: Throwable =>
        MetricsSystem.tracing { ms =>
          ms.incCount(CONN_FAIL)
          ms.incCount(MetricRegistry.name(CONN_FAIL, user))
          ms.incCount(MetricRegistry.name(CONN_FAIL, SessionType.INTERACTIVE.toString))
        }
        // If the session handle is null, closeSession(handle) will not be called,
        // so we need to decrement the limiter here.
        if (handle == null) {
          limiter.foreach(_.decrement(UserIpAddress(username, ipAddress)))
        }
        throw KyuubiSQLException(
          s"Error opening session for $username client ip $ipAddress, due to ${e.getMessage}",
          e)
    }
  }
```

